PR #88, opened 2015-01-31
Adding a readme to the docs folder.

jackiekazil:


jackiekazil:
@dmasad -- can you test out these instructions to make sure that they make sense? (I know there is a lot more to add and work on, but I wanted to push this through.) Assuming that this makes sense, please merge and close this request.

PR #92, opened 2015-03-04
Added iterators in space.py. Used them in Schelling.py.

gcallah:
Added iterators in space.py. Used them in Schelling.py. I also fixed some bugs in get_neighborhood() in space.py.

### DIFF
--- a/examples/Schelling/Schelling.py
+++ b/examples/Schelling/Schelling.py
@@ -58,32 +58,23 @@ class SchellingModel(Model):
         self.datacollector = DataCollector(
             {"happy": lambda m: m.happy}, # Model-level count of happy agents
             # For testing purposes, agent's individual x and y
-            {"x": lambda a: a.x, "y": lambda a: a.y}) 
+            {"x": lambda a: a.x, "y": lambda a: a.y})
 
         self.running = True
 
         # Set up agents
-        for empty in self.grid.occupied_iter(occupied=False):
-            x = empty[1]
-            y = empty[2]
+        for cell in self.grid.coord_iter():
+            x = cell[1]
+            y = cell[2]
             if random.random() < self.density:
                 if random.random() < self.minority_pc:
                     agent_type = 1
                 else:
                     agent_type = 0
 
-                agent = SchellingAgent((x,y), x, y, agent_type)
+                agent = SchellingAgent((x, y), agent_type)
                 self.grid[y][x] = agent
                 self.schedule.add(agent)
-
-    def get_empty(self):
-        '''
-        Get a list of coordinate tuples of currently-empty cells.
-        '''
-        empty_cells = []
-        for empty in self.grid.occupied_iter(occupied=False):
-            empty_cells.append((empty[1], empty[2]))
-        return empty_cells
 
     def step(self):
         '''
@@ -100,7 +91,7 @@ class SchellingAgent(Agent):
     '''
     Schelling segregation agent
     '''
-    def __init__(self, unique_id, x, y, agent_type):
+    def __init__(self, coords, agent_type):
         '''
          Create a new Schelling agent.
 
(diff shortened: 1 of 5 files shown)


examples/Schelling/Schelling.py
```diff
@@ -17,6 +17,9 @@
 
 To advance the model by e.g. 10 steps and print the new state:
     viz.step_forward(10)
+
```
jackiekazil: @gcallah -- can you remove the notes & can you remove the second random?

mesa/space.py
```diff
@@ -34,12 +37,44 @@ class Grid(object):
...
     '''
 
-    width = None
-    height = None
-    torus = False
-    grid = None
     default_val = lambda s: None
 
+
+    class GridOccupiedIter:
```
jackiekazil: @gcallah - this is a much nicer way of doing this. 
`GridOccupiedIter` is not intuitive for non-programmers to use. Without understanding Python, you might not know what that is. Suggestions for another names? 
@dmasad - do you have any thoughts on this?

mesa/space.py
```diff
@@ -34,12 +37,44 @@ class Grid(object):
...
+        """
+        Return just the (un)occupied cells of the grid.
```
jackiekazil: @gcallah -- for someone reading this -- what form does the grid need to be passed in? Can you add that and/or an example grid to the doc string?

mesa/space.py
```diff
@@ -62,65 +97,75 @@ def __init__(self, height, width, torus):
...
-    def get_neighborhood(self, x, y, moore, include_center=False, radius=1):
-        '''
-        Return a list of cells that are in the neighborhood of a certain point.
+    def __iter__(self):
+        # create an iterator that chains the
+        #  rows of grid together as if one list:
+        return itertools.chain(*self.grid)
+
+
+    def occupied_iter(self, occupied=True):
```
jackiekazil: @gcallah please add doc string & maybe we should consider updating the name on this also? Thoughts?

jackiekazil:
@gcallah --- this is an awesome update! Thank you for submitting.  I only had minor updates for submitting. Let me know if you have questions. 

Side note... 
We (as a project) really need to add tests. I feel like I can't ask for that, because we don't have the framework in place for that.

mesa/space.py
gcallah: class GridOccupiedCells

That might work.

The whole idea of using iterators is to make it easier for novice programmers: instead of nested loops over y and x, you just call an iterator.

But I am happy to give it a better name!

jackiekazil: I like where you are going with that, but since this is a subclass of a class called `Grid`, that would ... 
`Grid.GridOccupiedCells`

Maybe -- `Grid.OccupiedCells`? 
How about ... 
`class OccupiedCells`
Yes? No? 
(If you agree -- update. if not, then let me know you ur thoughts.)

mesa/space.py
gcallah: The grid does not need to be passed in! Just call the function I added to the Grid class: occupied_iter()

That will create an iterator for you that will return either the occupied or the unoccupied cells, depending on the occupied flag.

mesa/space.py
gcallah: Yes, renaming it OccupiedCells would be fine.

jackiekazil: :+1:

mesa/space.py
jackiekazil: :+1:

examples/Schelling/Schelling.py
gcallah: Done.

gcallah:
More improvements to space.py; used in Schelling.py.

jackiekazil:
Updates look good -- if you have anymore updates, please file a separate merge question. 
One last thing -- since we don't have tests, what is the easiest way to make sure this works?

gcallah:
Jackie, I tested Schelling.py and it seemed fine. I will look at the other examples as well.

PR #96, opened 2015-03-26
Tested models with new space code.

gcallah:
Everything is working except WolfSheep, which I did not change. I run the server, but it never starts listening on port 8888. I don't think this is something my work could have broken, but... perhaps someone can test the existing code and see if it works OK. If it does, I will debug on my end. The other models all seem to work fine.


jackiekazil:
I updated our code base to start adding continuous integration and the bases for tests. One of the things that I found is that you are using Python <=2.7. We are using Python 3.4, which is why WolfSheep appears broken to you now, because of this change: https://github.com/projectmesa/mesa/commit/08f8dd6f71809be497e603e2632a27bf5903d576

Assuming you have updated from master, you will need to update to Python3. I setup Python3 by using homebrew, which you would have to install if you don't have it already, then running... 
```
brew install python3
```
Then using virtualenv, you can create a new environment using python3. First identify where it is... 
```
which python3
```
output: /usr/local/bin/python3

Then use that to create a new virtualenv to work out of instead of the one that you have been working out of. Something like following...  
```
mkvirtual mesa --python=/usr/local/bin/python3
```

jackiekazil:
Also make sure to update from master. The update from master will add the travis.yml file, which will allow the build for the software to *hopefully* build properly. See: https://travis-ci.org/projectmesa/mesa/builds/55981318

gcallah:
No, I use Python 3.4 for all my development. I definitely did not switch versions to run the Mesa code.

Regards,
Gene Callahan

gcallah:
Ah, I think the problem was this missing file: I recall seeing a message about travis!

Regards,
Gene Callahan

jackiekazil:
It looks like the travis build is not working yet for this PR. Can you update your master branch w/ the new additions in our master branch? (Looks like your branched off awhile ago and missed a few updates.)

gcallah:
I’d be happy to, but how do I do that?

gcallah:
In other words, I don’t know how to refresh my cloned copy from the master. I poked around the web interface, and didn’t see any links that appear to do this.

dmasad:
That's fine, it just means we haven't written any tests yet (if I
understand it correctly).

jackiekazil:
@dmasad Yes, that is correct. 
This looks good to me.

Issue #97, opened 2015-04-01
WolfSheep not running on my end


gcallah:
I seem to have the latest code, and am running Python 3.4 on a Mac running Yosemite. All of the other examples work, but this one launches, says it is listening on port 8888... and that's it. I cannot connect on the port, and I see no further messages on the server end. But the server does keep running until I "Ctrl-C" it.

dmasad:
I'll try to take a look today, that might just not work at the moment.

gcallah:
One thing that could make life easier for a newcomer to the project: consistent ways to run the models. My students and I have a convention that our model is always named X_model.py and the file that sets it up and runs it is X_run.py.

In Mesa, this isn’t consistent across examples. Sometimes you run the model file, sometimes the server file, sometimes a file with visualization in the name (as I recall).

Enforced consistency in this regard makes the system as a whole easier to pick up.

Gene

PS — If you’d like, I will try debugging WolfSheep here to see what is up with it. Let me know.

dmasad:
Working on this. It seems to be the case that the model is starting, then hanging at some point during the run. I'm getting similar behavior sometimes (but not always) with the Schelling model as well. Digging into it.

dmasad:
Looks like there are two separate issues: 

1. There was a bug introduced into the Canvas Server which made it disregard the max_steps parameter when running a model, leading some instantiations to run forever. That's been resolved, and will be merged in ASAP.
2. Even with that bug fixed, the WolfSheepPredation model takes too long to run. In practice, this means that it takes several minutes between when the script is launched and the server is actually ready to serve the visualization to the browser.

Related to 2., the model seems to be exhibiting some incorrect behavior, which will become its own ticket.

PR #98, opened 2015-04-10
Fixing Canvas Server bug

dmasad:


dmasad:
Fixes the primary issue in ticket #97.

Issue #35, opened 2014-10-31
Make model run a coroutine


dmasad:
Right now, the visualization server (*mesa/visualization/TextServer.py*) runs the entire model to completion prior (the *run_model(...)* method) to serving the visualization. The model run should if possible be concurrent. This is the "Kim make this thing concurrent" ticket.

dmasad:
Attempting to work on this. One issue coming up so far is that with models where the number of agents increases indefinitely (e.g. WolfSheep), the continuous running slows down enough to slow down the server itself. One possible solution might be to pause the model running if it gets too far ahead of the current viz state? Or else just remind users to set reasonable max_steps constraints and termination conditions.

datnamer:
Perhaps the [dask]( http://dask.pydata.org/en/latest/ ) async scheduler can help? 

Pure python dependency, can encode tasks as nodes in a dag using dicts, tuples, and functions.

Also may be helpful for a future  parallel HPC version of Mesa, along with the upcoming [class PR](https://github.com/numba/numba/pull/1023) in numba:

dmasad:
Interesting, hadn't seen that package before! In the mean time, Tornado's built-in coroutines seem to work, but this is good to keep in mind as well.

PR #104, opened 2015-05-10
Attempt to add concurrency

dmasad:
This seems to successfully makes the model running by the server concurrent (addressing #35 ), but I'm not 100% sure that it isn't doing something horribly wrong.


dmasad:
Consolidating the visualization modifications into a single branch.

Issue #83, opened 2015-01-30
Split visualization / templates into two or more folders


jackiekazil:
Split visualization / templates into two or more folders & create a separate folder for the javascript/static files. 

Make template and static paths settable in the settings file (that doesn't exist until # is done).

Whomever takes this ticket -- make sure to send a head ups to the group. This is because this could cause a lot of conflicts in code.

dmasad:
This was done.

Issue #49, opened 2014-11-22
Composite different server visualization elements


dmasad:
Currently there are two completely different servers: TextServer takes a TextVisualization and puts it in the browser, and CanvasServer can only draw a grid. They have many elements in common, though also some key differences. 

The way to probably refactor them is to let the user create a Visualization class, which can mix and match text, canvas, to-be-implemented charts, etc. Then this composite visualization object is given to a highly parameterized server, which assembles the corresponding web page (based on a highly parameterized template, or better yet via mixing-and-matching with Tornado's [UI objects](http://tornado.readthedocs.org/en/stable/guide/templates.html)), and figures out how to serve it appropriately.

dmasad:
Being tackled in dmasad.ticket49 branch.

PR #124, opened 2015-06-15
Merging the modular server

### LINKED ISSUE #79: Modular visualization JavaScript in the HTML file or separate?
Right now, the JavaScript to render each visualization module is included in the template HTML file as a script tag. This means that on the final page, each visualization div has its own separate script tag. This is ugly, and also not very standard for web development. But none of the alternatives are great:

- Include the code in a separate JavaScript template. This means that each module needs at least two external files: an HTML template, and a JavaScript template.

- Have users write template code directly in Python, as a string. This seems non-ideal, since it involves writing JavaScript in a Python environment, as a long and hard-to-debug string.

- Other ideas?

dmasad:
This is a major commit, merging in the modular visualization server. It includes the server itself, a few modules (especially the CanvasGrid), and rewrites the example model visualizations to work with it. It also fixes a few other miscellaneous bugs. In particular, this closes #79, #83, #49, #35, and possibly some other ones I'm missing.


jackiekazil:
Holy crap this is big.
@dmasad So, I was going through this and there are a lot of conflicts w/ master. 
Questions... 
* Can I assume that everything in this would trump over master?  
* Did you run the tests locally with these changes?

dmasad:
Yes, it passes all the tests. And it mostly breaks backwards-compatibility with the visualization -- it shouldn't have an effect on the models themselves.

I'm not seeing a way to view all the conflicts on GitHub, but if you want to go over them together for verification, we can do that.

jackiekazil:
I feel like I have never seen this file in a repo before that I can remember.... not to say that it is a problem, but what is it? Is is supposed to be checked in? https://github.com/projectmesa/mesa/tree/clean_modules.79/examples/EpsteinCivilViolence/.ipynb_checkpoints

dmasad:
Ugh those are the chaff that the IPython Notebook leaves behind, and we should exclude those from being checked in.

jackiekazil:
:+1:

jackiekazil:
I am going to close this in lieu of #126 since you check that one already.

PR #140, opened 2015-07-07
Creating simultaneous activation scheduler.

dmasad:
Creating the simultaneous activation scheduler, and demonstrating it with a spatial prisoner's dilemma model.


jackiekazil:
This is a great example! Ty for adding it!

PR #150, opened 2015-07-11
Updates to package metadata + setup.py processing

jackiekazil:
* Adding version, license, copyright, title to __init__.py
* Auto updating version in setup.py

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -6,7 +6,10 @@ Release History
 Next release
 ++++++++++++++++++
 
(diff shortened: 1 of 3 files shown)


jackiekazil:
This add metadata to the package itself, so you will be able to do.... 
```
mesa.__version__
mesa.__license__
etc
```

Also, the addition to setup.py will take the version # from the __init__.py file and use it in the release, so it will only have to be updated in one place.

Issue #100, opened 2015-04-12
Standardize coordinate notation


dmasad:
The current models and Grid methods use either coordinate tuples or pairs of x, y arguments / variables. We should standardize to one or the other.

dmasad:
Standardizing to coordinate tuples

jackiekazil:
@dmasad can this be closed?

jackiekazil:
YAY!

PR #155, opened 2015-07-11
Standardizing spatial arguments

dmasad:
Standardizes the arguments passed to spatial functions to only tuples, not separate x and y coordinates.


jackiekazil:
Fixes #100

Issue #34, opened 2014-10-30
Add grass to Wolf-Sheep Predation Model


dmasad:
The wolf-sheep predation model does not currently have the sheep eating grass. Add in the grass, and have it reduced by sheep eating it, and grow back at a fixed rate.

davpaez:
I'm working on this ;)

dotsdl:
@davpaez are you still working on this? Otherwise, happy to dig into it.

jackiekazil:
@dotsdl, I assume you were at the sprint? Would you still be interested if @davpaez does not chime in soon?

dotsdl:
Yup. Still interested. :D

davpaez:
Hi, I'm sorry for the delay. I just submitted a pull request. It seems to be working. I'd like to hear your toughts on the custom schedule that I added.

dmasad:
@davpaez are you still working on the changes to the pull request, and/or do you want someone else to jump in and work on it?

jackiekazil:
Let's move this convo to #169

PR #161, opened 2015-07-14
Adding grass patches to WolfSheep model

davpaez:
This solves #34 

I tested this against the netlogo model, and it seems to behave very similarly.

### DIFF
--- a/examples/WolfSheep/WolfSheep.py
+++ b/examples/WolfSheep/WolfSheep.py
@@ -15,7 +15,7 @@ import random
 
 from mesa import Model, Agent
 from mesa.space import MultiGrid
-from mesa.time import RandomActivation
+from mesa.time import BaseScheduler
 
 from RandomWalk import RandomWalker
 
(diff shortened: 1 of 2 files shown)


dmasad:
Looks pretty good! Just adding a couple notes. Do you want to take a shot at adding the grass to the visualization as well?

examples/WolfSheep/WolfSheep.py
```diff
@@ -86,11 +92,41 @@ def __init__(self, height=20, width=20,
...
+
+                patch = GrassPatch(fully_grown, countdown)
+                self.grid.place_agent(patch, (x, y))
+                self.schedule.add(patch)
+
         self.running = True
 
     def step(self):
         self.schedule.step()
+        print([self.schedule.time,
```
dmasad: Printing should be optional (e.g. with a `verbose` parameter) since otherwise it clutters up the command line during batch runs, visualization, etc. This is fine for debugging, but best practice is probably to use a data collector to keep track of these stats.

examples/WolfSheep/WolfSheep.py
```diff
@@ -86,11 +92,41 @@ def __init__(self, height=20, width=20,
...
+               self.schedule.get_breed_count(Wolf),
+               self.schedule.get_breed_count(Sheep)])
+
+    def run_model(self):
+
+        print('Initial number wolves: ', self.schedule.get_breed_count(Wolf))
+        print('Initial number sheep: ', self.schedule.get_breed_count(Sheep))
+
+        for i in range(200):
```
dmasad: The number of steps to run for should be parameterized.

examples/WolfSheep/WolfSheep.py
```diff
@@ -99,15 +135,42 @@ class Sheep(RandomWalker, Agent):
...
-        if random.random() < model.sheep_reproduce:
+        x, y = self.pos
+        living = True
+
+        if model.grass:
+            # Reduce energy
+            self.energy -= 1
+
+            # If there is grass available, eat it
+            this_cell = model.grid[y][x]
```
dmasad: Check out the new `get_cell_list_contents` method that's been added to the Grid class, which is a better way of doing this.

examples/WolfSheep/WolfSheep.py
```diff
@@ -99,15 +135,42 @@ class Sheep(RandomWalker, Agent):
...
+            grass_patch = [obj for obj in this_cell if isinstance(obj, GrassPatch)][0]
+            if grass_patch.fully_grown:
+                self.energy += model.sheep_gain_from_food
+                grass_patch.fully_grown = False
+
+            # Death
+            if self.energy < 0:
+                model.grid[y][x].remove(self)
```
dmasad: Probably better practice to use the `grid._remove(....)` method here.

examples/WolfSheep/WolfSheep.py
```diff
@@ -139,10 +202,91 @@ def step(self, model):
...
+        if not self.fully_grown:
+            if self.countdown <= 0:
+                # Set as fully grown
+                self.fully_grown = True
+                self.countdown = model.grass_regrowth_time
+            else:
+                self.countdown -= 1
+
+
+class RandomActivationByBreed(BaseScheduler):
```
dmasad: Suggestion: do you want to generalize this, so that instead of being model-specific it's generic enough to handle any model with multiple agent types? If the scheduler has multiple agent lists, one per 'breed', it could add and remove them in those methods, and save the time of sorting through the entire list with every step.

dmasad:
Additions made in #169.

Issue #180, opened 2016-01-02
WebSocket connection fail


VictorNorman:
When I start up ForestFireVisualization.py, I get this:

WebSocket connection to 'ws://127.0.0.1/ws' failed: Error during WebSocket handshake: Unexpected response code: 403

dmasad:
Fixed!

PR #181, opened 2016-01-05
Fix Issue 180: WebsocketHandler's check_origin() method returns True.…

VictorNorman:
…  (See https://groups.google.com/forum/#!topic/web2py/6rUe9EZBd58).


dmasad:
(Fixes #180)

PR #188, opened 2016-03-08
Fixed error on numpy update

jamjar919:
Numpy updated and caused errors in this example, due to vectors using on the fly conversion, causing the error: TypeError: Cannot cast ufunc add output from dtype('float64') to dtype('int64') with casting rule 'same_kind'.

This adds np.int64() functions around some variables to get rid of this error.
 .


jackiekazil:
Thank you for catching this. Looks good.

Issue #191, opened 2016-03-23
The parameters of CanvasGrid(): grid_row, grid_col and canvas_height, canvas_width are inverted somewhere downstream


ReblochonMasque:
As shown on the screenshots below, the parameters for CanvasGrid() seem to be inverted - they in fact are not, but there is (likely) an inversion problem downstream towards rendering on canvas.

GRID_COLS which represents the number of cols in the width comes first, but it is CANVAS_HEIGHT that comes first for the next pair of arguments… (the same goes for GRID_WIDTH and CANVAS_WIDTH)

It also appears that the number of white lines that separate the cells drawn on canvas is determined by the number of rows GRID_ROWS only, thus leading to messing up the display of a non square grid.

(These white lines should be optional, but that is a cosmetic issue.)

The parameter CELL_SIZE which represents the pixel size of a cell on canvas is not used as I think it should: the cell proportion is back-calculated from the ratio of each canvas dimension / number of cells in this dimension, leading to rectangular cells i/o square ones... I think this is okay, but maybe not, depending on the behavior we want.

![grid row cols vs canvas height width](https://cloud.githubusercontent.com/assets/6275531/13979177/8ad749d8-f111-11e5-9f3e-f1b7adb743a0.png)

#################################################################################

![grid row cols vs canvas height width2](https://cloud.githubusercontent.com/assets/6275531/13979184/92f14d9e-f111-11e5-83d0-0b9ae304e286.png)

I took a quick look at the CanvasGrid module and the signature of the initializer looks okay... the error is further down but I did not spot anything immediately obvious.

In reference to the conversation on the mailing list, renaming the variables _row & _col everywhere a reference to the number of cells on the height and the number of cells in the width is made, will improve the code clarity.

Ideas, suggestions, advice?

ReblochonMasque:
I think I found the culprit in `CanvasModule.js`, line 14:

`var canvasDraw = new GridVisualization(canvas_width, canvas_height, grid_height, grid_width, context);`

I'll see tomorrow if I can fix it, it is too late now, I need a fresh mind for this.

most of the js I saw has parameters `(width, height)` in this order, whereas the python code has parameters in `(row, col)`, that is `(height, width)` order... This is going to bite us in the butt one day!

ReblochonMasque:
Tests also reflect the inconsistency:

From `test_grid.py` :

> `        for agent in self.agents:
>             x, y = agent.pos
>             assert self.grid[y][x] == agent`

[...]

PR #192, opened 2016-03-24
Color Patches Model example

ReblochonMasque:
Wrote an example model.
Faced some API glitches:
* several row/col col/row issues as mentioned elsewhere.
* direct attributes access forcing a number of specific choice of attribute names
* canvas parameters (lines separating the cells on the canvas)

I documented these in the code in preparation to addressing the issues; however, I am unsure if this example should be published as-is, or if the comments added are a no go. (it probably depends on the timeline to correct the API)
With current settings of the example, the lines separating cells on canvas are only visible on half of it.

Request a pull for a **quick eyeball** on the code and **advice** on how to proceed.
This is not a WIP, it can be merged if the comments added and the canvas glitch are okay.

### DIFF
--- a/examples/ColorPatches/Readme.md
+++ b/examples/ColorPatches/Readme.md
@@ -6,15 +6,20 @@ At each time step, an agent's opinion is influenced by that of its neighbors, an
 As an agent adapts its thinking to that of its neighbors, the cell color changes.
 
 
-### Options you can play with:
+### Parameters you can play with (you must change the code at this stage):
 * Vary the number of opinions.
 * Vary the size of the grid
 * Change the grid from fixed borders to a torus continuum
 
 
+### Observe
+* how groups of like minded agents form and evolve
+* how sometimes a single opinion prevails
+* how some minority or fragmented opinions rapidly disappear
+
+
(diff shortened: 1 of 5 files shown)


jackiekazil:
@ReblochonMasque 
1. Waiting to merge until we have a convo about x,y issue -- I think this has come up before, but I don't remember the result. 
2. Why is the grid outline only spanning 1/2 the visualization? Is that an artifact of Mesa OR is that on purpose? (I am thinking Mesa, and maybe that is a bug to fix?)

ReblochonMasque:
The grid canvas drawing of the white lines that separate the cells has a problem for rectangular grids  issue #191. This problem is hidden for square grids which is most of the other examples.

jackiekazil:
No keep it as is. So we can make sure it is fixed. 
So, when #191 is fixed, then this model would be displaying properly?

ReblochonMasque:
Yes, when #191 is fixed, the display of rectangular grids will be good

jackiekazil:
That is what I thought, but I wanted to confirm. This all looks good then. Merging.

PR #197, opened 2016-03-29
Fix Schelling documentation

dmasad:
The documentation for the Schelling example provided instructions which were no longer compatible with the code; trying to follow them would just lead to errors. I largely just removed the out-of-date instructions, leaving only one which works.


jackiekazil:
Looks good to me. Merging!

PR #204, opened 2016-05-15
the x and y coordinates are switched

verginer:
If the grid is not square this would result in an error, it works for the dimensions used in the example but could be confusing.


dmasad:
Good catch! Fixed.

PR #207, opened 2016-06-01
Updating model example readmes

dmasad:
Updating the model example readmes, and model directory structures.


examples/Schelling/server.py
```diff
@@ -1,6 +1,35 @@
-from Schelling import SchellingModel
 from mesa.visualization.ModularVisualization import ModularServer
 from mesa.visualization.modules import CanvasGrid, ChartModule, TextElement
+from mesa.visualization.TextVisualization import (TextData, TextGrid,
+    TextVisualization)
+
+from model.world import SchellingModel
```
jackiekazil: This line needs to be changed to `from schelling import SchellingModel` OR we need to rename `schelling.py` to `model.py`. I believe I would prefer `model.py` rename. Thoughts? @dmasad

jackiekazil:
Rebase is going to cause lots of conflicts, so I am merging in as is.

PR #212, opened 2016-06-02
introduce a requirements.txt for the tutorial.

smacleod:
The tutorial contains dependencies which the mesa project does not
explicitly require itself (such as matplotlib). Even though most users
of the mesa library will have a full scientific python environment it's
good to call this out explicitly.

I've added a requirements.txt to list every dependency needed to run
the tutorial example, including mesa itself.

The tutorial documentation is also updated to include steps for
installing the dependencies using the requirements.txt

### DIFF
--- a/examples/Tutorial-Boltzmann_Wealth_Model/Readme.md
+++ b/examples/Tutorial-Boltzmann_Wealth_Model/Readme.md
@@ -8,4 +8,4 @@ This model is a tutorial contained in a Jupyter notebook
 
 ### To run this example
 
-You will need to open the file as a Jupyter (aka iPython) notebook with an iPython 3 kernel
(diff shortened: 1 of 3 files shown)


examples/Tutorial-Boltzmann_Wealth_Model/Readme.md
```diff
@@ -8,4 +8,4 @@ This model is a tutorial contained in a Jupyter notebook
 
 ### To run this example
 
-You will need to open the file as a Jupyter (aka iPython) notebook with an iPython 3 kernel
\ No newline at end of file
+You will need to open the file as a Jupyter (aka iPython) notebook with an iPython 3 kernel. Requirement dependencies are listed in the provided `requirements.txt` file which can be installed by running `pip install -r requirements.txt`
```
smacleod: "Requirement dependencies"  - ew... I'll fix this.

PR #222, opened 2016-06-02
add a requirements.txt for WolfSheep.

smacleod:


jackiekazil:
Yup yup... that looks good.

PR #228, opened 2016-06-03
Ignored _build sphinx docs still in repo

lowcloudnine:
Based on my research into the history the built Sphinx docs were
commited once before they were added to the .gitignore file.  I
deleted from the cache so they don't show as deleted when working
with sphinx from now on.


jackiekazil:
@lowcloudnine ty for that catch!!!! LGTM

PR #231, opened 2016-06-03
pass optional port parameter to ModularServer.launch

max-orhai:
Allows configuration of Tornado server's port at server launch time.


jackiekazil:
I'm okay with the .2% decrease .... for now.

PR #237, opened 2016-06-03
Gitignore

jackiekazil:
Adding .DS_Store to the gitignore so it stops being annoying.


dmasad:
The actual .gitignore file isn't showing up in the Files Changed tab. Is that normal for hidden files?

jackiekazil:
That is weird. I updated it and it seemed to work. After the CI build is done, it is ready.

jackiekazil:
@dmasad this is ready.

Issue #225, opened 2016-06-02
cookiecutter for mesa


strixcuriosus:
A [cookiecutter template](http://cookiecutter.readthedocs.io/en/latest/) could be an easy way to start scaffolding for a new mesa project. thoughts?

jackiekazil:
Closing this ticket in preference for PR #242

PR #242, opened 2016-06-03
Mesa new project scaffolding command

strixcuriosus:
This introduces a command `mesa-init` that can be used to start a new mesa project.

It's based on a cookiecutter template, which could potentially go into a separate repository in the future.

**Demo:**
```bash
python setup.py install
mesa-init
```
- Follow the interactive prompts to name the new project (just hit enter to use default names)

- You should now have a new subdirectory in your current working directory. The subdirectory will be named in title case based on the project name you provided.
```bash
cd MyNewProject
python run.py
```

https://github.com/projectmesa/mesa/issues/225

We can add to the template in the future and potentially make sub-templates for particular use cases. e.g. the interactive cookiecutter prompt could ask for a scheduler type and grid params.

We could also make license, readme, and version components optional.


mesa/cookiecutter-mesa/CHANGELOG.md
```diff
@@ -0,0 +1,8 @@
+# Change Log
+All notable changes to this project will be documented in this file.
```
drewrey: How does this differ from the repo's [history file](https://github.com/projectmesa/mesa/blob/master/HISTORY.rst), and should we maintain both?

jackiekazil: The change log is the same as the history file, just a different way of calling it. I am thinking this might be a changelog for the model if the cookiecutter template is for the model. Does it have to be changelog or can it be history? (It doesn't matter to me what it is as long as it is consistent with the other. So, if we do change log here, we should probably rename the history file to change log as well. Make sense?

strixcuriosus: @drewrey this changelog file is for the cookiecutter template itself. It is not necessary, though, since the template is encapsulated in the main mesa repo for now.

Note: We could make a standalone repo for the cookiecutter template, but then the mesa-init command would have this external repo as a dependency.

@jackiekazil we could also add a per-model history file to the cookiecutter template. I will defer that addition for now.

In summary there are at least 3 places to track history:
1. **mesa repo itself**, which is already covered by the [HISTORY.rst](https://github.com/projectmesa/mesa/blob/master/HISTORY.rst)
2. **cookiecutter template**, which is what this changelog at `mesa/cookiecutter-mesa/CHANGELOG.md` or `mesa/cookiecutter-mesa/HISTORY.rst` would track
3. **the new mesa project** created by the mesa-init command, i.e. the user's own model could have a history file. this would be inside of the `{{cookiecutter.title}}` directory

dmasad:
I wonder -- would this be a good candidate for a demo mesa-kit? On one hand, it's self-contained in terms of files and dependencies. On the other, it makes assumptions about the API that may be broken going forward.

@jackiekazil -- do you have any thoughts on structuring mesa-kits and how this fits in?

PR #243, opened 2016-06-03
Test batchrunner

drewrey:
- Added tests to check the shape of the model and agent variable dataframes.
- Calling: `self.collect_agent_vars.items()` was erroring, so I explicitly called `self.collect_agent_vars(model)`.


jackiekazil:
This LGTM -- Ty for bringing the test coverage up to 72%!!!!

PR #245, opened 2016-06-04
clean up TextVisualization

drewrey:
- remove unused method `step_forward`
- remove class variables
- fix type in docstring


jackiekazil:
Ty for this clean up! Good catch!

Issue #239, opened 2016-06-03
Separate the front-end visualization and back-end data generation for visualization from the web server


smacleod:
It would be nice to refactor both the front-end javascript and the back-end rendering code of a `VisualizationElement`, so that they aren't coupled as tightly to the Tornado server setup and its template.

## Motivation
The motivation for this issue is to allow the front-end visualizations written in javascript to be used in situations where the webserver is not needed (or wanted). Examples would include the inside of a Jupyter notebook output cell, or a static HTML page where the data for a model run could be saved statically and replayed.

## Current Code
Currently the javascript visualization code makes assumptions about the template it is loaded into,  injecting into the body tag. Some of the visualizations also send along canvas commands, rather than data which should be left to the javascript to make use of.

Each individual visualization also uses a `js_code` property to specify initialization code, which must deal with specifying arguments itself.

## Proposed Changes
- Make each visualization generate an element which can be injected into any container at the discretion of the caller. This will allow consumers like a jupyter notebook extension to inject into output cells.
- Replace `js_code` with an initialization function name e.g. `ChartModule`, and always pass a provided dictionary as the single argument containing any of the options. This will ensure new modules are written in a way which doesn't make assumptions about the type of front-end environment it will be run in and also allows the tornado webserver or notebook extension to change options as needed (possibly change dimensions?).

dmasad:
I'm definitely in favor of the motivation and concept. Two questions:

1. It seems that right now the websocket connection chokes and slows down if the JSON data being passed is too big; wouldn't that get even worse if we're sending entire serialized HTML elements?

2. One of the strengths of the current system is that it can use third-party JavaScript charting tools (e.g. Charts.js). Is that still possible under this change?

One interim solution is not to refactor the ModualServer architecture, but to create a parallel visualization architecture that does this.

smacleod:
To answer your questions:
1. I'm not actually proposing that any of the data being sent over the websocket needs to change, the JS side will still deal with generating the markup / updating the visualization. I might have been making some assumptions about *what* data was being sent before, but this change doesn't need to modify that at all.
2. Yup, that should all still work!

I've hacked together a `Visualizer` class which can use the current `VisualizationElement`s as is to inject them into the notebook and render using JS. The only change I needed to make to them was having the JS inject the canvas elements into a specific div I create, rather than the body. This can all be fixed though by the refactoring I layout here, and will only require a small update to the `ModualServer` template.

PR #247, opened 2016-06-04
WIP: introduce Visualizer for rendering in notebooks.

### LINKED ISSUE #246: Allow displaying mesa visualizations inline in jupyter notebooks
It would be nice to run the Model visualizations inside of a notebook.

smacleod:
This change is a Work In Progress.

It should be possible to render Mesa's javascript visualizations
directly in jupyter notebook cells, where the model has been defined in
the notebook itself.

This introduces a helper Visualizer class which will inject the required
javascript into the notebook, deal with running the model and, and
sending the updated state data to the notebook.

This change is dependent upon #239

fixes #246

### DIFF
--- a/mesa/visualization/ipython.py
+++ b/mesa/visualization/ipython.py
@@ -1,0 +1,137 @@
+import os
+import json
+import ipywidgets as widgets
+from IPython.display import (
+    display,
+    Javascript,
+    HTML,
+)
+MESA_VISUALIZER_ID_COUNT = 0
+
+
+class Visualizer(object):
+    """A helper for running Mesa visualizations inline in a notebook.
+
+    The instantiation is similar to the ModularVisualization server but instead
+    of starting the server the `display()` method should be called at the
+    end of a notebook cell.
+
+    ex:
+    ```
+        from mesa.visualization.ipython import (
+            Visualizer,
+        )
+        from mesa.visualization.modules import (
+            ChartModule,
+        )
+
+        chart = ChartModule(
+            [{"Label": "Gini", "Color": "Black"}],
+            data_collector_name='datacollector')
+        v = Visualizer(MoneyModel, [chart], "Money Model", 100, 10, 10)
+        v.display()
+    ```
+    """
+    __name__ = "Execute Model Step"
+
+    def __init__(self, model_cls, visualization_elements, name="Mesa Model",
+                 *args, **kwargs):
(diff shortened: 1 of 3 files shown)


dmasad:
Playing around with this, a few observations:

1. This will be an amazing new feature.

2. Turned out I didn't have ipywidgets installed, so I needed to install / enable that.

3.  At the moment, just importing the ```Visualizer``` class from ```mesa.visualization.ipython``` isn't enough, and ```MESA_VISUALIZER_ID_COUNT``` also needs to be imported. Is there a reason to have that as a global variable and not a Visualizer property?

4. The visualization doesn't render until the first time "Step the Model" is pressed.

5. Seriously, this will be an amazing new feature.

I'm going to dive a bit more into this code and into how widgets work, to get a better idea of what's going on and how to help. Re Issue #239 definitely agree that we should look to restructure visualization elements in a way that makes them more compatible with this.

smacleod:
> Turned out I didn't have ipywidgets installed, so I needed to install / enable that.

Ah, good call, we'll need to add some documentation around this.

> At the moment, just importing the `Visualizer` class from `mesa.visualization.ipython` isn't enough, and `MESA_VISUALIZER_ID_COUNT` also needs to be imported. Is there a reason to have that as a global variable and not a Visualizer property?

No good reason for any of this. I actually prototyped the `Visualizer` by writing it in a notebook cell itself rather as a module of mesa - I was running out of time at the sprint so this pull request was put together pretty haphazardly. Definitely will need cleanup.

> The visualization doesn't render until the first time "Step the Model" is pressed.

Ya, there are some UX warts. I was actually thinking the controls would be expanded giving the user the ability to run the model at some steps per second etc. - similar to the current visualization server UI. I'm no expert with in notebook animation or ipython widgets though, so we'll definitely need to do some exploring with respect to what's possible and makes sense.

> Seriously, this will be an amazing new feature.

Yay! 😀 

> I'm going to dive a bit more into this code and into how widgets work, to get a better idea of what's going on and how to help. 

Awesome. Just keep in mind this is more of a prototype with lots of room for improvement in the way we inject the JS and render things etc. So if you find something strange that's probably why.

jackiekazil:
Since this is a POC, I am going to close this in lieu of the tickets, knowing that if someone decides to dig into this, they can refer to the POC that @smacleod sketched out here.

PR #252, opened 2016-06-04
Pyconsprints

lowcloudnine:
Small adjustment to docs/conf.py -- changed back to default theme.  Also converted docstrings in visualizaiton module to double quotes and added # -*- coding: utf-8 -*- at the top off all py files.


jackiekazil:
LGTM

PR #259, opened 2016-06-14
Pyconsprints  - updating master. 

jackiekazil:
General update to master.


jackiekazil:
After this passes, I will merge into master.

jackiekazil:
@drewrey could the build be failing because there is an issue with the nightly Python build. (Before I go down a rabbit hole, I wanted to ask.)

drewrey:
Hm, it looks that way to me. Pandas [doesn't build](https://travis-ci.org/pydata/pandas) against the nightly Python build. I don't see any obvious results for the `sre_constants` error.

Since mesa is downstream of other popular scientific projects, perhaps we should remove `nightly` from the travis config?

jackiekazil:
@drewrey before I just remove nightly & 3.6... is there any other option that is > 3.5 as an option to build again or are we held up by pandas?

Issue #260, opened 2016-06-15
web visualization examples does not show grid in offline mode


jeetu7:
I installed mesa on a offline m/c and found that the grid is not visible in web version of visualization.
The reasons might be that the server is trying to fetch jquery from online (check the 4th line on following link)

https://github.com/projectmesa/mesa/blob/master/mesa/visualization/templates/modular_template.html#L4

Is it possible to include jquery in static part alongside dat.gui.min.js so that the examples/code work offline.

jackiekazil:
PR #261 takes care of this.

PR #261, opened 2016-06-15
Fixed issue #260 by adding a copy of jquery directly into the code.

jeetu7:
Added copy of jquery and modified the modular_template.html file appropriately.


jackiekazil:
RE: Failed Travis build -- that is being taken care of by another PR. 
I think this is good to go! Thank you for the contribution!

PR #262, opened 2016-06-20
Minor updates to how docs are build.

jackiekazil:


jackiekazil:
@dmasad this is a very minor update, but I don't want to merge myself for the sake that everything that is going into master gets another pair of eyes on it.

jackiekazil:
This also includes the version update that went out with the late release.... whoops. Sorry about that. It needs to go into master too.

Issue #248, opened 2016-06-04
Example models should call super in init


drewrey:
Some example models are not calling `super()` in their respective `__init__` methods.

nvrzm:
One detriment of this is that random.seed() isn't getting called for these examples, so they're not deterministic. Another is that each subclass must remember to set self.running = True. On the other hand, it's not clear that Model.__init__ is the best place for these things to be happening - maybe it would be better for the base class to have no __init__ and to add, for instance, set_seed and set_running methods.

dmasad:
The base classes are more there as placeholders, and the constructors aren't necessarily intended to be initialized via super() -- especially when in some cases (e.g. WolfSheep) there are multiple inheritances between the base class and the actual agent class. If we want to require agents to call ``super().__init__()``, we need to make sure that it actually contains something useful.

At the moment, I also don't think the examples need to be deterministic.

jeetu7:
This message is just to document the error message that the system will print once I run the web visualization and hit reset. (if i miss setting self.running = True in my model subclass)

```
  File "/home/user/learn/venvs/mesa3_venv/lib/python3.5/site-packages/tornado/gen.py", line 282, in wrapper
    yielded = next(result)
  File "/home/user/learn/venvs/mesa3_venv/lib/python3.5/site-packages/mesa/visualization/ModularVisualization.py", line 270, in run_model
    while self.model.schedule.steps < self.max_steps and self.model.running:
AttributeError: 'RoboModel' object has no attribute 'running'
```

dmasad:
@jeetu7 That issue is fixed in master, so the fix will be included in the next release.

jackiekazil:
And that release should have been released a few days ago.
Let me know if it still needs to be addressed.

On Tue, Jun 21, 2016 at 2:39 PM, David Masad <notifications@github.com>
wrote:

> @jeetu7 <https://github.com/jeetu7> That issue is fixed in master, so the
> fix will be included in the next release.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/projectmesa/mesa/issues/248#issuecomment-227580796>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAKLTr6WrIgEEbStYoTlztU7TKb1D15Lks5qOFoQgaJpZM4IuAh0>
> .
>



-- 
Jacqueline Kazil | @jackiekazil

jeetu7:
[...]

PR #275, opened 2016-07-16
Update intro_notebook and documents to include self.running = True in MoneyModel

tphoran:
This should close issue #248


jackiekazil:
LGTM

PR #288, opened 2016-07-17
Refactor forest fire example #223

evgraph:

PR #290, opened 2016-07-17
Updating kernel version on forest fire model.

jackiekazil:
When checking #288, I received an error with regards to which kernel jupyter uses when the notebook is running. The was resolved by setting the kernel to Python3.

To test this, the notebook should be runnable when you open it with not pop up prompts asking you which kernel to use. (This assumes that you have a Python3 kernel for jupyter to use.)


jackiekazil:
Tested / reviewed in room at Scipy.

PR #301, opened 2016-07-24
Fixed #295, #296, #300. 

jeetu7:

PR #303, opened 2016-07-28
Adding gradient support for agents in GridVisualization

yigitbey:
I've renamed the Color parameter to Colors and made it an array. If the list consists of only one color then the shape is filled only with that color, otherwise it fills it with a gradient created with that colors. 
I've also added an optional parameter called stroke_color. If it is not provided than it becomes the first given color in the colors array.

### DIFF
--- a/mesa/visualization/templates/GridDraw.js
+++ b/mesa/visualization/templates/GridDraw.js
@@ -104,7 +104,7 @@ var GridVisualization = function(width, height, gridWidth, gridHeight, context) 
             var grd=context.createRadialGradient(cx, cy, r, cx, cy, 0);
 
             for (i = 0; i < colors.length; i++) {
-                grd.addColorStop(i, colors[i]);
+                grd.addColorStop(i/colors.length, colors[i]);
             }
 
             // Fill with gradient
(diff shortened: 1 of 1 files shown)


jackiekazil:
FYI https://groups.google.com/forum/#!topic/projectmesa/x2En3z9KtWQ

jackiekazil:
@yigitbey so sorry for the delay on this. I was just about to review, but I think the changes from #301 caused a conflict in your PR. Can you merge or rebase those into this branch to resolve?

yigitbey:
Sorry, I was not able to check it yet due to some health issues. I'm recovering now but I won't be able to merge the changes for another week or two.

dmasad:
This looks pretty great! No rush, and get well soon! 

For when you're back -- My concern here is that it seems like ```Colors``` has to be an array here. Can we change it so that it can also accept just a single color parameter, without being wrapped in an array?

jackiekazil:
@yigitbey --- I was wondering what the status of this was & if you are okay? Will you update this?

jackiekazil:
Since there has been no movement on this PR, I am going to comment and close and open a ticket to update and resubmit. This is a great idea, but there is more work to be done on this.

PR #305, opened 2016-07-29
Scheduler keeps track of agents in dict

### LINKED ISSUE #302: Removing an agent interferes with RandomActivation step method
First of all, thanks so much for all the great work y'all've put into this. It's let me explore agent based modeling very quickly and easily.

I am implementing a model where agents have a finite lifespan, however, I noticed that my agents were persisting in the model (their AgentIDs were present in more Steps) than they should be for their lifespan. I tracked this down to a conflict between removing an agent and stepping through the model. If the agent's `step()` removes the agent from the model, this modifies the list of agents being iterated over in the scheduler's `step()` (starting at line 63 in [schedule.py](https://github.com/projectmesa/mesa/blob/master/examples/WolfSheep/wolf_sheep/schedule.py); `step_breed()` is called by `step()`):

``` python
def step_breed(self, breed):
    '''
    Shuffle order and run all agents of a given breed.
    Args:
        breed: Class object of the breed to run.
    '''
    agents = self.agents_by_breed[breed]
    random.shuffle(agents)
    for agent in agents: # <- If agent is removed, 'agents' is modified
        agent.step()
```

This leads to undesirable behavior (at least in the RandomActivation scheduler) where not all agents are called in each iteration if an agent is removed from the list. In the code below, all the agents should be present for only two steps maximum, however, some are present for more than that (the value varies by run presumably because of the shuffling of agents in the scheduler). Here's an example of what I mean:

``` python
from mesa import *
import numpy as np

class TestModel(Model):
    def __init__(self, agent_lifetime = 1, n_agents = 100):
        self.agent_lifetime = agent_lifetime
        self.n_agents = n_agents

        ## keep track of the the remaining life of an agent and
        ## how many ticks it has seen
            agent_reporters = {"remaining_life" : lambda a: a.remaining_life,
                               "steps" : lambda a: a.steps})

        self.current_ID = 0

        for _ in range(n_agents):
            self.schedule.add(FiniteLifeAgent(self.get_next_ID(),
                                              self.agent_lifetime))

    def get_next_ID(self):
        self.current_ID += 1
        return self.current_ID

    def step(self):
        '''Add agents back to n_agents in each step'''
        self.schedule.step()

        if len(self.schedule.agents) < self.n_agents:
            for _ in range(self.n_agents - len(self.schedule.agents)):
                self.schedule.add(FiniteLifeAgent(self.get_next_ID(),
                                                  self.agent_lifetime))

    def run_model(self, step_count = 100):
        for _ in range(step_count):
            self.step()

class FiniteLifeAgent(Agent):

    '''
    An agent that is supposed to live for a finite number of ticks.
    Also has a 10% chance of dying in each tick.
    '''

    def __init__(self, unique_id, lifetime):
        self.unique_id = unique_id
        self.remaining_life = lifetime
        self.steps = 0

    def step(self, model):
        inactivated = self.inactivate(model)
        self.steps += 1 # keep track of how many ticks are seen
        if np.random.binomial(1,0.1) is not 0: # 10% chance of dying
            model.schedule.remove(self)

    def inactivate(self, model):
        self.remaining_life -= 1
        if self.remaining_life < 0:
            model.schedule.remove(self)
            return True
        return False


model = TestModel()
model.run_model()

df = df.reset_index()

# Get the total number of ticks an agent is present in. 
lifetimes = df.groupby(["AgentID"]).Step.agg({"Step" : lambda x: len(x)})
# Varies between ~8 and 14 depending on the run.
print("Total number of ticks an agent is alive in should be 2. Actual value is: %i" % lifetimes.Step.max())

# actual value is 1 and 0, respectively, showing that all agents aren't called in each step of the scheduler
print("Total steps seen by agent should be 1. Actual value is: %i" % df.steps.max())
print("Minimum remaining life of agent should be 0. Actual value is: %i" % df.remaining_life.min())
```

I have fixed this for my model by having the scheduler keep track of a dictionary of agents keyed by a unique integer that I stored in the agent's `unique_id` slot rather than a list of agents:

``` python
class BaseScheduler:
    """ Simplest scheduler; activates agents one at a time, in the order
    they were added.

    Assumes that each agent added has a *step* method which takes no arguments.

    (This is explicitly meant to replicate the scheduler in MASON).

    """
    model = None
    steps = 0
    time = 0
    agents = {}

    def __init__(self, model):
        """ Create a new, empty BaseScheduler. """
        self.model = model
        self.steps = 0
        self.time = 0
        self.agents = {}

    def add(self, agent):
        """ Add an Agent object to the schedule.

        Args:
            agent: An Agent to be added to the schedule. NOTE: The agent must
            have a step() method.

        """
        self.agents[agent.unique_id] = agent

    def remove(self, agent):
        """ Remove all instances of a given agent from the schedule.

        Args:
            agent: An agent object.

        """
        del self.agents[agent.unique_id]

    def step(self):
        """ Execute the step of all the agents, one at a time. """
        agent_keys = list(self.agents.keys())
        for agent_key in agent_keys:
            self.agents[agent_key].step(self.model)
        self.steps += 1
        self.time += 1

    def get_agent_count(self):
        """ Returns the current number of agents in the queue. """
        return len(self.agents.keys())
```

By iterating over the dictionary keys rather than the agents themselves, if an agent is removed the other agents are still stepped by the scheduler correctly. 

I've made changes to the relevant classes in `time.py` in order to implement this, and they pass the unit tests. However, all classes that inherit the BaseScheduler class in example models would need to be examined and probably fixed. I also noticed that in some example models the `unique_id` slot is being used for other purposes than just identifying the agent. For example, in the WolfSheep model, it's being used to store the position of the agent. There could be other examples that would need updated, too.

I would be happy to fix up the WolfSheep model and submit a pull request if that would be helpful, but please advise this total newcomer to collaborative software how best to proceed!

best,
Colin

csmaxwell:
@jackiekazil 

Hello! I think this would resolve #302 . I've changed the scheduler to handle a dict rather than a list. I added a `next_id()` method to Model to dispense `unique_id`s for agents. I've made WolfSheep and the tests compatible with the change, and added a test for whether the lifespan of agents with a finite life is what they should be. I haven't messed with the docs or other examples, but I could work on that if y'all'd like. 

Since this is my first pull request, please let me know if I've messed anything up!

best,
Colin

### DIFF
--- a/examples/WolfSheep/wolf_sheep/agents.py
+++ b/examples/WolfSheep/wolf_sheep/agents.py
@@ -14,8 +14,8 @@ class Sheep(RandomWalker):
 
     energy = None
 
-    def __init__(self, pos, model, moore, energy=None):
-        super().__init__(pos, model, moore=moore)
+    def __init__(self, unique_id, pos, model, moore, energy=None):
+        super().__init__(unique_id, pos, model, moore=moore)
         self.energy = energy
 
     def step(self):
@@ -47,7 +47,8 @@ class Sheep(RandomWalker):
             # Create a new sheep:
             if self.model.grass:
                 self.energy /= 2
-            lamb = Sheep(self.pos, self.model, self.moore, self.energy)
+            lamb = Sheep(self.model.next_id(), self.pos, self.model,
+                         self.moore, self.energy)
             self.model.grid.place_agent(lamb, self.pos)
             self.model.schedule.add(lamb)
 
@@ -59,8 +60,8 @@ class Wolf(RandomWalker):
 
     energy = None
 
-    def __init__(self, pos, model, moore, energy=None):
-        super().__init__(pos, model, moore=moore)
+    def __init__(self, unique_id, pos, model, moore, energy=None):
+        super().__init__(unique_id, pos, model, moore=moore)
         self.energy = energy
 
     def step(self):
@@ -87,7 +88,8 @@ class Wolf(RandomWalker):
             if random.random() < self.model.wolf_reproduce:
                 # Create a new wolf cub
                 self.energy /= 2
-                cub = Wolf(self.pos, self.model, self.moore, self.energy)
+                cub = Wolf(self.model.next_id(), self.pos, self.model,
+                           self.moore, self.energy)
                 self.model.grid.place_agent(cub, cub.pos)
                 self.model.schedule.add(cub)
 
@@ -97,7 +99,7 @@ class GrassPatch(Agent):
     A patch of grass that grows at a fixed rate and it is eaten by sheep
     '''
 
-    def __init__(self, pos, model, fully_grown, countdown):
+    def __init__(self, unique_id, pos, model, fully_grown, countdown):
         '''
         Creates a new patch of grass
 
@@ -105,9 +107,10 @@ class GrassPatch(Agent):
             grown: (boolean) Whether the patch of grass is fully grown or not
             countdown: Time for the patch of grass to be fully grown again
         '''
-        super().__init__(pos, model)
+        super().__init__(unique_id, model)
         self.fully_grown = fully_grown
         self.countdown = countdown
+        self.pos = pos
 
     def step(self):
         if not self.fully_grown:
--- a/examples/WolfSheep/wolf_sheep/model.py
+++ b/examples/WolfSheep/wolf_sheep/model.py
@@ -60,7 +60,7 @@ class WolfSheepPredation(Model):
                                  once it is eaten
             sheep_gain_from_food: Energy sheep gain from grass, if enabled.
         '''
-
+        super().__init__()
         # Set parameters
         self.height = height
         self.width = width
@@ -84,7 +84,7 @@ class WolfSheepPredation(Model):
             x = random.randrange(self.width)
             y = random.randrange(self.height)
             energy = random.randrange(2 * self.sheep_gain_from_food)
-            sheep = Sheep((x, y), self, True, energy)
+            sheep = Sheep(self.next_id(), (x, y), self, True, energy)
             self.grid.place_agent(sheep, (x, y))
             self.schedule.add(sheep)
 
@@ -93,7 +93,7 @@ class WolfSheepPredation(Model):
             x = random.randrange(self.width)
             y = random.randrange(self.height)
             energy = random.randrange(2 * self.wolf_gain_from_food)
-            wolf = Wolf((x, y), self, True, energy)
+            wolf = Wolf(self.next_id(), (x, y), self, True, energy)
             self.grid.place_agent(wolf, (x, y))
             self.schedule.add(wolf)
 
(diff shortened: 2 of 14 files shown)


jackiekazil:
@csmaxwell if you run this line in your dev environment, it will provide you a list of all the pep8 issues, so you don't have to wait on travis to be mean. :-) 

`pip install flake8`
`flake8 . --ignore=F403,E501,E123,E128 --exclude=docs,build`

Let us know when this is read for review.

jackiekazil:
FYI https://groups.google.com/forum/#!topic/projectmesa/x2En3z9KtWQ

csmaxwell:
@jackiekazil Thanks for the tip on flake8! I think it's ready for review. Also, thanks for giving me the heads up on possible review delays. They won't affect my research or deadlines, so no worries from my end.

njvrzm:
(I commented initially on the issue but I think this is probably a better place)
Nice catch indeed, and I think the approach of keeping Agents in a dictionary seems like a good one, but there are some things to consider before adopting it:
* It will change the order in which Agents are called for step() in any models using the BaseScheduler's step() method. Currently they're called in the same order as they're added, but with this change they'll be called in an order that depends on the values of the Agents' unique_ids. This could be fixed by using an OrderedDict instead.
* More importantly, it will enforce the uniqueness of the unique_id each agent is created with. Currently, despite the name, nothing happens differently if unique_id is in fact not unique across the Agents of one's Model - indeed, the nature of each Agent's unique_id has no affect at all on the behavior of the model - so this change might break some existing models. I think there should at least be an exception raised if one tries to add an agent with a duplicate unique_id to a scheduler.

I think this raises interesting questions, though: why do Agents have unique_id at all? As far as I can tell, it's only currently used for logging. Is it useful to be able to specify the id, or should that be managed for the user?

jackiekazil:
I am going to try to do a review of this over the weekend.

jackiekazil:
(sorry for the delay of my chime in... I have one more component to my Ph.D exam that wraps up today --- I will catch up after that.)

pgervila:
Maybe I am missing something here and if so I apologize in advance, but in order to avoid iteration over a list that is being modified, wouldn't it suffice to iterate over a COPY of the agents list in the schedule step method, instead of the original ?

Something like this

    def step(self):
        """ Execute the step of all the agents, one at a time. """
        for agent in self.agents[:] :
            agent.step()

I don't expect any significant overhead from the copy operation

The dict implementation could still be a better idea ( I have to check performance-wise), but the impact of iterating over a copy would be minimum and I think it could fix the bug

pgervila:
In any case, even if there was no bug in the list implementation, I think dicts would be a better approach. Removing an agent from a list would be at least an order of magnitude slower than from a dict. Therefore, for models with some few thousand limited life-span agents , dicts would really make a difference

rht:
Memory consumption is reduced in the dict version, since the list version loops
over a list copy of all the agents, while the dict version loops over the list
of the shuffled keys (there shouldn't be any runtime perf penalty).

rht:
I have a local branch which uses @csmaxwell's PR. To speed up the process of this being merged, if it is okay, I have re-organized the commits + rebased.

dmasad:
This was refactored into #386

Issue #314, opened 2016-09-07
Feature request: Display progress of BatchRunner.run_all() 


rdotsch:
I have now run a couple of simulations using BatchRunner, taking several hours of computing time. It would be very helpful if the BatchRunner.run_all() method would display a text progress bar, updated after each single simulation run, preferably with remaining time estimation.

jackiekazil:
That is a great point! Any interest in developing that? :-D If not, we can add it our list. I love this feature.

rdotsch:
Absolutely, when I find time.

Ron Dotsch, PhD | Associate Professor | Department of Psychology | Utrecht University | Office: Langeveld H1.11, Heidelberglaan 1 | +31 30 253 4198 | http://ron.dotsch.org/

njvrzm:
[tqdm](https://pypi.python.org/pypi/tqdm) looks like a nice option here!

rdotsch:
Yes, that’s an excellent suggestion. I’ve implemented it just now and it looks great. I’ll test it a bit, and submit the new code soon.

Ron

dmasad:
Done!

PR #316, opened 2016-09-29
Ticket #314 - added progress bar to BatchRunner

rdotsch:
Introduces new dependency: tqdm. Default behaviour: progress bar is displayed automatically. This can be overridden by setting display_progress parameter to False when creating a new BatchRunner object. Code will work on any existing model and can be tested with any model using BatchRunner.


dmasad:
This is terrific, thanks!

PR #321, opened 2016-10-14
Changing to port 8521, fixes #320.

### LINKED ISSUE #320: Update default port number
After lecturing at a class @ GMU, when I was helping users run models, I noticed that some users were also running jupyter, which uses port 8888, our current default port. 

For this reason, I am going to update the default.

jackiekazil:
I choose 8000, because it is considered the "default" port to use for web applications. Technically, this project fits that. Let me know if you have other feels. 

fixes #320


njvrzm:
The ZIP code for downtown Mesa, AZ is 85211; we could use port 8521 :)

jackiekazil:
I love that... but I until we have #322 resolved, I dislike going with a funky number.

njvrzm:
Okay! Just made a PR for that.

jackiekazil:
@njvrzm this is ready for review. Updated to 8521.

jackiekazil:
After this is merged, I am going to do a PyPI push to 0.7.8

PR #323, opened 2016-10-14
Opens a browser window when launching the server

### LINKED ISSUE #322: When server is launched, browser window should open automagically
When guest lecturing at GMU, I observed that users did not read the instructions to visit 127.0.0.1... in their browser. I think jupyter spoiled them ;-). We should add functionality to have the browser auto launch when someone kicks off a model.

njvrzm:
There shouldn't be any race conditions here because we're
opening the browser after the server starts listening but
before the event loop runs.

Fixes #322.

### DIFF
--- a/mesa/visualization/ModularVisualization.py
+++ b/mesa/visualization/ModularVisualization.py
@@ -81,11 +81,12 @@ Client -> Server:
 import os
 
 import tornado.ioloop
-import tornado.template
 import tornado.web
 import tornado.websocket
 import tornado.escape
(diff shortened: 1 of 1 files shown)


njvrzm:
I hope the small drop in coverage is acceptable; I added a couple of lines to an untested function.

jackiekazil:
giphy

/me ignore drop, for sake of getting this in, begins review.

jackiekazil:
LGTM. Merging.

PR #326, opened 2016-10-27
Update space.py

ycaxgjd:
convert float to int


njvrzm:
In Python 3, which mesa requires, the math.ceil function returns an integer already, so this change shouldn't be necessary. Did you encounter a problem which this fixed?

njvrzm:
I'm going to close this one - if there is an issue, please re-open it!

PR #329, opened 2016-10-31
Auto update year for copyright.

jackiekazil:
This is a super small update that just auto increments the year.
Should we write a test for this? My hesitation is that it will break when the year changes or that we will write the exact same code as the commit itself. Thoughts?

To test that this works... 

```
import mesa
mesa.__copyright__
```


jackiekazil:
I have no idea how this increased coverage.

PR #339, opened 2016-11-11
Only start tornado ioloop if necessary

### LINKED ISSUE #337: Advanced tutorial chart
I have been unable to get the chart showing the gini coefficient to open up in  the advanced tutorial to open up. Everything else is working, the chart just does not appear. Please advise, 

Thank you.

Tom

etymancer:
Running the tutorial (or any of the examples) in jupyter would
result in errors when calling server.launch() because the tornado
ioloop would already be running.

Fixes #337.

### DIFF
--- a/mesa/visualization/ModularVisualization.py
+++ b/mesa/visualization/ModularVisualization.py
@@ -80,6 +80,7 @@ Client -> Server:
 """
 import os
 
+import tornado.autoreload
 import tornado.ioloop
 import tornado.web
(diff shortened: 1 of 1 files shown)


etymancer:
Same problem as before with coverage: the launch method isn't tested at all, so making it longer drops coverage. I'll see about creating a test for it.

dmasad:
Looks good to me, and seems to solve the problem!

PR #345, opened 2016-12-21
fix boundaries of ContinousSpace

jkerfs:
An IndexOutOfBounds Exception was thrown when an agent was placed at either the x_max or y_max in a ContinuousSpace.  My simple fix was to have the `out_of_bounds` function use a >= comparison instead of >.  I also added some tests of these boundaries in test_space.py


dmasad:
Good catch!

PR #346, opened 2016-12-27
implement "end" message in visualization

jkerfs:
It looks like support for terminating simulations was designed but not finished in the visualization code.  This PR sends the "end" message from the Tornado server to the browser as a websocket message when `model.running` is false.  The client-side javascript was also updated to stop the simulation when the "end" message is received.

The Forest Fire example can be used to demonstrate this change.


dmasad:
Looks good!

Issue #347, opened 2017-01-04
find_empty() and move_to_empty() for MultiGrid


Corvince:
Hello mesa,

is there any good reason not to implement find_empty() (and other empty-related functions) for a MultiGrid? Or is just not implemented yet? 

It would be useful for the model I am building currently and if there is demand I could try to implement it, but maybe I am missing a good point against those functions?

Best
Corvin

Corvince:
Hi dmasad,

I am running into a few more questions here: 
1) First I implemented the methods for MultiGrid, but than I thought I could as well implement it for the base Grid class. Only then did I realize that Grid is also frequently used in the examples, but its neither a SingleGrid nor a MultiGrid. I guess that is an own issue, but I am not sure of the expected behaviour of Grid (regarding empty cells). It also seems inconsistent: If I place multiple Agents on the same position in a Grid, only the last agent is visible by grid[x][y], but all the agents keep their agent.pos set. 
2) There is no remove_agent() function, can I add one? (calling _remove_agent() and setting agent.pos = None) 
3) Since I have never contributed to a repo before: There is also an easily resolved TODO snippet in space.py, can I resolve that in the same pull request? (with its own commit?) Sorry, but I don't now whats the best practice!

dmasad:
1. the base Grid class is intended to be used fairly minimally. Ideally, we should go through the examples and clean them up so that they're using SingleGrid or MultiGrid. However, you can probably implement the `empty` methods at the Grid level by checking whether the contents of a cell are the same as the grid's default value.

2. That seems like a good addition -- go ahead!

3. Good question! Yes, fixing the minor TODO at the same time should be fine. For large changes, there should probably be one pull request per feature, but fixing a small outstanding issue like that should be safely foldable into another PR.

dmasad:
Looks like this has been addressed!

PR #349, opened 2017-01-10
Move empty-cell functions to baseclass Grid

Corvince:
Following up on the discussions #347 

All the examples should work without any changes.


dmasad:
Looks good!

PR #354, opened 2017-02-03
User params 0.8

Corvince:
This is an updated branch for the user parameters. I did not change any of the actual code, but mainly rebased the code to the current master branch.

My personal additions are the updates of the newer examples and a first documentation, which needs some review! 

The jupyter notebook of the advanced tutorial would also need to get updated, but I want to wait for a review of the .rst file


Corvince:
I added an option to autoupdate model parameters, which is switchable via a checkbox

I have barely any knowledge about JavaScript, so I hope it is not too hackish. Please note that my editor automatically removed trailing whitespaces, so the changes for the modular_template.html appear larger than they are.

dmasad:
This looks really good (sorry for the very late response)! I'm not sure why the Travis check is failing, though. @jackiekazil any idea?

mesa/visualization/templates/modular_template.html
```diff
@@ -139,12 +164,30 @@
 <script>
 	// Set up the GUI
 	var gui = new dat.GUI();
-	var fps_control = gui.add(control, "fps", 0, 20).step(1); // fps slider, 0 to 20 in 1 increments 
+	var fps_control = gui.add(control, "fps", 0, 50).step(1); // fps slider, 0 to 20 in 1 increments
```
dmasad: Is there a reason to increase this to 50? That seems on the high end, and at least some models can't keep up with that frame-rate, which causes buggy behavior right now.

Corvince: Ah, no not really. It just felt too slow for me sometimes ;)

Haven't tested it though, I wasn't aware of the buggy behavior. But that should at some time be fixed anyway, right? When you run a complex model on a slow computer it can already be buggy at 20 fps.

But I have nothing against leaving it at 20 for now.

dmasad: Fair point. I'd lean towards a compromise (maybe 30fps) for now.

dmasad:
Cool, this looks just about ready! I'd still like to get some thoughts on whether to merge straight into master or a development branch.

Corvince:
Good question when to merge this. 

It is not a small API change, it will basically break all existing models using the built-in visualizations. 

On the other hand I think this can be a quite substantial feature, especially for beginners.

Corvince:
While looking into the flockers issue I realized that this one should be merged soon. Not only to set the parameters interactively (which could help with the bug-hunting), but especially because right now the examples are hard to understand. Currently you have to know all the model parameters and their order to understand the server.py code. By explicitly requiring a dictionary this becomes much easier to read. 

But could someone please review the documentation? I have never written any documentation and the jupyter notebook still needs to be updated.

dmasad:
Playing around with this just now, I hit what I think is a bug, where the visualization server doesn't work if there are no UserParams present. The simplest way to verify this is to run the Flockers example as-is. 

It should be an easy fix, so I'll try to add that ASAP.

TaylorMutch:
Hello! I pulled together a simple PR for looking at this topic, but then actually read through this issue (whoops). See #375. I'm guessing it would be better to build off of this PR rather than rework the wheel... Hope that this helps in some way though! I'm happy to contribute towards this effort.

jackiekazil:
Please hold on doing further work on this. @TaylorMutch & I have been chatting about general improvements. These are great, but because of some complexity in a series of changes, pieces from this and another one might have to be pulled together. 

More info coming from @TaylorMutch via the mailing list.

dmasad:
Superceded by #383; closing.

PR #355, opened 2017-02-10
Added function to use local images as shapes in GridDraw

Corvince:
Enables to draw a local imagefile as shape for agents. Example below

If there is support for this feature I can add documentation. 

download 1


dmasad:
This is really cool! Will merge as soon as the Travis build issue is fixed.

Issue #352, opened 2017-01-30
Removing agents is causing duplication issue


bianicap:
Hi,

As a disclaimer, I'm new to both Python and Mesa.

I'm working on an ABM of attrition. To debug my model, I've pared the code down to a really simple model. You have N number of agents, at each time step, agents have a certain probability of attriting. If the agent attrits, the agent is removed/killed from the simulation. For each agent, I track a Month variable (a tick in the simulation represents one month). In this very simple version, the Month should match the Step in the model. However, this is not always the case. For some reason, the Month value sometimes repeats. I see this in a csv output of the model results. So we might have an agent with Steps 0, 1, 2, 3, but then the corresponding Month might be 0, 1, 1, 2 (the Month that repeats is often not the month that the agent attrits). I'm using the following line of code to kill the agents when they attrit: model.schedule.remove(self). If I do not "kill" agents, the duplication issue with the Month variable is resolved. However, I will need a way to remove agents for my bigger model.

Any ideas? I'm happy to provide the complete code for my simple model. Thanks.

Corvince:
This is probably related to #302 

The issue is that model.schedule is a simple list of all agents that gets iterated over in each model step. 
What you are doing is modifying this list while being iterated. This can cause unforeseen consequences. 
There is already a PR #305 to solve this issue by replacing the list with a dictionary.

In the meantime, a simple workaround I use is to create a list of agents to be killed and only kill them after each agent got activated. Something along those lines:

```
# Agents step function
if self.gets_killed():
    self.model.kill_agents.append(self)

# Models step function
self.kill_agents = []
self.schedule.step()
self.schedule.remove(self.kill_agents)
```

Note that this of course could change model behavior if you need to kill your agents instantaneous. In that case follow the example of #302

bianicap:
Thanks! This workaround worked great.

Corvince:
Glad it worked. The one-liner was actually my error, I do the same loop in my model! 
schedule.remove() only accepts agents and no list so my code above can't work and you have to use a loop as you did.

djmitche:
Addressed in #356 if you want to close this up

PR #356, opened 2017-02-17
Modified loop over agents in schedule step method

pgervila:
This addresses both #352 and #302, basically the same issue
If agents are deleted while looping over agents list in the model schedule,
looping becomes unpredictable and agents step methods for some agents may not be called for a given model step
A simple solution seems to be to iterate over a shallow copy of the agent list by using [:]
A shallow copy constructs a new compound object and then (to the extent possible) inserts references into it to the objects found in the original. So the objects we get from iteration are referencing the originals, but the copied list is not modified if agents are removed

A dict implementation could still be a preferable solution in terms of performance ( see #305) , but this one should have less impact

### DIFF
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -72,7 +72,7 @@ class BaseScheduler:
 
     def step(self):
         """ Execute the step of all the agents, one at a time. """
-        for agent in self.agents:
+        for agent in self.agents[:]:
             agent.step()
         self.steps += 1
         self.time += 1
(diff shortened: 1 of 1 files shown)


dmasad:
This definitely seems like a solid placeholder solution. As a longer-term thing, we do need to decide what the correct / desired behavior is when an agent is removed mid-step -- should they still potentially be activated one last time (which I believe is the case here), or removed immediately even if they're still in the queue.

Will merge as soon as the Travis issue is figured out.

Issue #359, opened 2017-03-10
[New Feature request] Evolutionary models


bennati:
Hello everyone,

I have seen the sheep-wolf example and it looks quite complex.
I thought that one could simplify it by creating custom classes that take care of the evolutionary process.
If you are interested I can prepare a pull request, let me know!

Cheers,
Stefano

dmasad:
Hi Stefano,
That sounds like an interesting idea ! Can you explain what you mean by evolutionary process? I'm not sure I see how that would simplify the wolf-sheep example.

dmasad:
The model as it is now doesn't have any selection component, and is mostly focused on population size dynamics -- but if you're interested in implementing an extension that adds an evolutionary submodel, I think that could be a great additional example to have alongside the current wolf-sheep model!

jackiekazil:
Looking back at the notes in #360. I am going to close this ticket. 
However, after rereading this, I want to encourage you once again to create this as a separate library. If you need help figuring out how to set this up.

PR #360, opened 2017-03-19
Added support for evolutionary models

bennati:
As discussed in Issue #359 , I implemented some new classes that allow to simplify the creation of evolutionary models, i.e. where agents accumulate energy, reproduce, mutate and die.

In order to show how to use the model, I refactored the wolf-sheep example to use the new classes.
The example does not use all features of the new classes, so if you have another evolutionary model in mind I could implement it as an example.

This code is to be considered preliminary, I'm open to discussion about how to improve its design to make it more general and about implementation of new (selection) algorithms.

Cheers,
Stefano

### DIFF
--- a/examples/wolf_sheep/wolf_sheep/agents.py
+++ b/examples/wolf_sheep/wolf_sheep/agents.py
@@ -1,6 +1,5 @@
 import random
-
-from mesa import Agent
+from mesa.EvoAgent import *
 
 from wolf_sheep.random_walk import RandomWalker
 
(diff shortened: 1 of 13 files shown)


dmasad:
Hi Stefano,

This looks great! I think you've built a really cool set of components, and I want to get a chance to review further and suggest tweaks or improvements.

My main concern is that I'm not sure whether a specific evolutionary model is a good fit for adding to the Mesa core, since it does currently have a lot of hard-coded assumptions and a limited number of example cases. 

One thing we've talked about doing before is Mesa kits -- extensions to the core framework which add specific features like geography, or -- in this case -- evolution. What are your thoughts on using this as a first example of that, and spinning this off into a `mesa.evolution` or something like that sub-module?

bennati:
Sure, it sounds like a good idea! Will you take care of the changes?
Please discard my last commit as that is a change in the core module, I'll submit a separate pull request for that later

mesa/EvoAgent.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def step(self):
+        """
+        Performs one step of simulation
+
+        Side effect: age is increased by 1
+        """
+        self.age+=1
+
+    def duplicate(self):
```
rht: I couldn't tell why you chose not to use "clone", given the domain area of the model ;)

rht: Or rather, clone in the strictest sense is DNA->DNA copy, but duplicate is for molecule->molecule copy. The former is just the recipe, which doesn't even include the epigenome, the latter's definition includes all that is necessary.

examples/wolf_sheep/wolf_sheep/agents.py
```diff
@@ -1,6 +1,5 @@
 import random
-
-from mesa import Agent
+from mesa.EvoAgent import *
```
rht: `import *` makes it harder for `pylint`/`flake8` to statically analyze the code.

rht:
This fixes #359 if nailed (to keep track of the things...).

bennati:
Hi guys, sorry about the long absence.
Could you please explain what I have to do in order to make this a sub-module? Is there any example/guide?
Thank you!

jackiekazil:
@bennati -- you bring up a great point. It is not clear how this might move forward design wise. Since everyone doesn't follow every ticket, I have posted to the dev email list. Hopefully, we can get some folks to chime in there. See: https://groups.google.com/forum/#!topic/projectmesa-dev/4HThxwC5sGQ 

For now, I am going to consider this 'blocked'. Hopefully we can get you unblocked soon!

jackiekazil:
@bennati ... looks like we are narrowing in on a model for this. I will update when have something solidified. -- https://groups.google.com/forum/#!topic/projectmesa-dev/4HThxwC5sGQ

jackiekazil:
@bennati after meeting about this, we think the best way to approach this is to create a separate library that works with Mesa. For example - evolutionary-mesa or evol-mesa or mesa-evol or something like that. That library would have mesa w/ a specific version number as a requirement. This would be hosted under your account or another organization. We would point people to it as a resource. Over time, if it became a key component with wide adoption, we might discuss hosting it under the Mesa org, but still keeping it separate from core mesa. We will also be here for you a resource to help you structure the library correctly. When you start a repo, please post it here or in a ticket, so that we can poke around and help you get up and running.

PR #371, opened 2017-05-22
Add best-practices document and describe models.

### LINKED ISSUE #334: Write a section in the docs that explains the modeling standard
We have the following two standards...
#324
#223

But we don't have them documented anywhere. 
Maybe we should create a "best practices" doc?

djmitche:
Fixes #334


jackiekazil:
Great addition! Ty!

Issue #368, opened 2017-05-19
Seeding when using numpy


larsasp:
I am using np.random. a lot in my model, and I want to set a seed, np.random.seed to get consistent line of random numbers. This does not seem to work to great, and many people seem to dislike np.random vs just random. Has anyone had a similar problem?

larsasp:
Update: specified np.random.RandomState and passed it instead of numpy.random function. It works beautifully, however, I am still getting too much variation in modelled outcomes, and I suspect it is due to the activation of the agents not being seeded. Is there a plan to include such a function, or does anyone have any idea how to work around it in the mean time?

djmitche:
`RandomActivation` uses Python's random number generator, not numpy's, so you will need to seed that in order to get stable activation.

Note that the Model parent class seeds this generator by default.  I'm making a pull request to seed the numpy generator, too.

jackiekazil:
Fixed by #373 -- @larsasp, this fix will be in our next release, which will happen post PyCon... approximately next week. 

Ty @djmitche for the fix!

Issue #366, opened 2017-05-18
Have the schedulers use the model's randomizer


lmzintgraf:
Regarding this TODO in the documentation:

> _TODO: Have the schedulers use the model's randomizer, to keep random number seeds consistent and allow for replication._

How about letting the model have a random state (using numpy) which the scheduler can then call? I.e.,
- in the model init: `self.random_state = numpy.random.RandomState(seed)`
- in the schedule step: `self.model.random_state.shuffle(self.agents)`
- alternatively, in the schedule init: `self.random_state = self.model.random_state` (which I assume will make a copy; not important for reproducibility but it might be nicer to use the model's random state directly)

Some things I ran into while trying this:
- It would now be required for new models to call the `super().__init__()` method (see #248 ) to make sure that models have a random state. Alternatively, we could put a check in the scheduler if the model has a random state, and if not, create a local one. The problem is that then the seed can't be specified, or, if it can be passed to the scheduler, it will only have effect if the model doesn't have a random state, which seems weird.
- When testing this, test_time.py will crash:
    - First error is because `super().__init__()` is not called (see above), and the scheduler cannot call `self.model.random_state`
    - Another error occurs when `mesa.time.random.shuffle` is called in the test. It will either complain that the import isn't there (I removed it, `import random` in time.py is not necessary anymore), or the following happens:
    - `test_random_activation_step_shuffles` and `test_shuffle_shuffles_agents` crash. I guess it's because the way random calls are made changed (it's not `mesa.time.random.shuffle`, but via np.random). Not sure how to fix this.

djmitche:
The `Model` base class already seeds the global Python random number generator.  I think it can also seed the global Numpy random number generator with the same seed, and I'm making a PR to do so.  Then all uses of random values can just consult those global generators, avoiding the need to pass around random state instances.

lmzintgraf:
I think it might be safer to use random state instances, instead of setting the global seed. If someone uses the mesa framework and sets global random seeds in their own implementation, they will unknowingly influence the random generators in the mesa implementation, which might lead to unwanted behaviour. In addition, setting the global seed might not be thread-safe, and it could very well be that somebody works with multiple threads with mesa.

I can make a PR with the proposed changes. (Same thing goes for the space module, where also the global random generator was used.)

djmitche:
[...]

PR #373, opened 2017-05-22
Seed both global random number generators

### LINKED ISSUE #107: Allow fixed seed for replication
Figure out how to best handle running a model with a fixed seed, to ensure that it will produce the same result. and implement that.

djmitche:
This also includes a best-practices note encouraging use of these RNGs
and suggesting calling the Model parent constructor.

Fixes #107, #366, #368.


jackiekazil:
LGTM -- Ty for the fix.

PR #374, opened 2017-05-22
Explicit fixed and variable parameters for batchrunner

njvrzm:
A fix for #307. The batchrunner was getting confused if there was a parameter with a single value that was an iterable type, thinking it was getting a range of values instead. This change provides for explicit variable parameters, which take a sequence of values, and fixed parameters, which take a single value (which may itself be a sequence type).


jackiekazil:
This all looks good, except for the example models need to be updated... 
- mesa/examples/schelling notebook
- mesa/examples/forest_fire notebook
- mesa/docs/tutorials intro notebook

I *think* that is all of them.

njvrzm:
Good point! I'll work on that today.

dmasad:
@njvrzm I can tackle updating the examples if that would be helpful.

mesa/batchrunner.py
```diff
@@ -67,36 +87,42 @@ def __init__(self, model_cls, parameter_values, iterations=1,
 
         self.display_progress = display_progress
 
+    def _process_parameters(self, params):
+        params = copy.deepcopy(params)
+        bad_names = []
+        for name, values in params.items():
```
rht: This typecheck can be enforced with PEP484, will be much simpler. The version requirement will have to be bumped to be at least 3.5.

mesa/batchrunner.py
```diff
@@ -51,8 +71,8 @@ def __init__(self, model_cls, parameter_values, iterations=1,
 
         """
         self.model_cls = model_cls
-        self.parameter_values = {param: self.make_iterable(vals)
-                                 for param, vals in parameter_values.items()}
+        self.variable_parameters = self._process_parameters(variable_parameters)
+        self.fixed_parameters = fixed_parameters or {}
```
rht: This *reads* like `fixed_parameters || {}` ;) -- `{}` should have been a default value for fixed_parameters, instead of `None`.

mesa/batchrunner.py
```diff
@@ -67,36 +87,42 @@ def __init__(self, model_cls, parameter_values, iterations=1,
...
-            kwargs = dict(zip(params, param_values))
-            for _ in range(self.iterations):
+        param_names, param_ranges = zip(*self.variable_parameters.items())
+        run_count = count()
+        total_iterations = self.iterations
+        for param_range in param_ranges:
+            total_iterations *= len(param_range)
+        with tqdm(total_iterations, disable=not self.display_progress) as pbar:
+            for param_values in product(*param_ranges):
+                kwargs = dict(zip(param_names, param_values))
```
rht: This is basically `kwargs = self.variable_parameters`

rht: Ah, Cartesian `product()`, nvm.

mesa/batchrunner.py
```diff
@@ -67,36 +87,42 @@ def __init__(self, model_cls, parameter_values, iterations=1,
 
         self.display_progress = display_progress
 
+    def _process_parameters(self, params):
+        params = copy.deepcopy(params)
+        bad_names = []
+        for name, values in params.items():
+            if (isinstance(values, str) or
+                    not isinstance(values, collections.Sequence)):
```
dmasad: This fails with numpy sequences (e.g. in ``forest_fire/Forest Fire Model.ipynb``)

a4d442663580: I found an unresolved issue, but with a patch here https://github.com/numpy/numpy/issues/2776#issuecomment-10932295. `Sequence.register(np.ndarray)` has to be added at somewhere.

dmasad: I think it's also resolved with ``hasattr(values, "__iter__")``

dmasad:
It looks like the tests are failing due to changes in the visualization API. Can someone take a look and confirm / verify, and see whether it's something that needs to be addressed here, or will self-resolve with the merge?

a4d442663580:
It has to be addressed here, the error in https://travis-ci.org/projectmesa/mesa/jobs/249398056#L1038
is basically
`TypeError: __init__() missing 1 required positional argument: 'val'`.

a4d442663580:
The fix is that `MockAgent` takes 3 params, https://github.com/projectmesa/mesa/blob/bce83226f315c82fe6f6b1b214f1539e05d5fef4/tests/test_batchrunner.py#L17. The `MockModel` has to be inserted before `val`.

a4d442663580:
In addition to have that bug fixed, the branch also needs to be rebased.

rht:
Heads up: to reduce the number of roundtrips, I have fixed the errs, rebased in #393.

mesa/batchrunner.py
njvrzm: Oh, this absolutely a no-no in Python. Never use mutable objects as default parameters - they're shared between every invocation of the function, so if they're modified, terrible bugs can creep in. The way this works is quite standard.

rht: TIL

rht: This is pylint w0102[1].

[1]http://pylint-messages.wikidot.com/messages:w0102

Issue #223, opened 2016-06-02
Update all examples to follow model standard file setup


jackiekazil:
We created stand file setups for our models. You can see these in [Schelling](https://github.com/projectmesa/mesa/tree/pyconsprints/examples/Schelling) and [WolfSheep](https://github.com/projectmesa/mesa/tree/pyconsprints/examples/WolfSheep). 

Schelling is an example of a very simple model setup, while WolfSheep is an example of a more complex setup. To update you will need to do the following... 
- Update the model & server files
- Update the notebook to analysis and to use the new file structure
- Update the Readme

This is a multi-effort ticket, so feel free to pick one, claim it then check it off when you submit a PR.

It looks like the following projects need to be updated:
- [x] [Color patches](https://github.com/projectmesa/mesa/tree/master/examples/color_patches)
- [x] [Conways Game of Life](https://github.com/projectmesa/mesa/tree/master/examples/conways_game_of_life)
- [x] [Epstein Violence model](https://github.com/projectmesa/mesa/tree/master/examples/epstein_civil_violence) #255
- [x] [Flockers](https://github.com/projectmesa/mesa/tree/master/examples/Flockers) #254
- [x] [Forrest Fire](https://github.com/projectmesa/mesa/tree/master/examples/forest_fire)
- [x] [PD Grid](https://github.com/projectmesa/mesa/tree/master/examples/PD_Grid)
- [x]  [Schelling](https://github.com/projectmesa/mesa/tree/master/examples/Schelling)
- [x] [Boltzmann_Wealth_Model](https://github.com/projectmesa/mesa/tree/master/examples/boltzmann_wealth_model)
- [x] [WolfSheep](https://github.com/projectmesa/mesa/tree/master/examples/wolf_sheep)

strixcuriosus:
I'll take on updating color patches!

strixcuriosus:
I created a follow-up issue to consider making a cookiecutter template: https://github.com/projectmesa/mesa/issues/225

jackiekazil:
\o/ Color patches is done!!

nvrzm:
I've opened a PR for updating the Game of Life example.

nvrzm:
Flockers PR'd... on to Epstein.

nvrzm:
Aaaaand that's done. More soon!

jackiekazil:
Forest Fire was fixed w/ #288.

jackiekazil:
[...]

PR #377, opened 2017-05-23
Prisoner's Dilemma refactor to meet new model standard format.

tojo13:
Part of #223.

Notes:
- Right now the schedule type is set to `RandomActivation` via the `"Random"` argument in `server.py`. However, in the jupyter notebook example, the PD model behaves differently for different schedule types. `model.py` handles different schedule types, but there is currently no option in the front end visualization UI to change the schedule type.


jackiekazil:
LGTM

PR #375, opened 2017-05-22
Add simple, adjustable model parameters to client.

TaylorMutch:
Adds basic support for updating user-supplied parameters from model keyword arguments .

PR #383, opened 2017-05-29
Bootstrap UI starter

TaylorMutch:
This PR works towards moving the UI elements off of `dat.GUI` and utilize Bootstrap instead. Attached below is a screenshot of what the Wolf and Sheep example looks like. The run controls are moved into the top-right of the UI, boolean toggles and numerical inputs are on the left, and the drawing elements are on the right.

I think that this could be a good discussion or for building off of this PR. Referencing #375, this PR is built on the same logic implemented there, so #375 could be ignored in lieu of this PR.

It'd be great to get feedback on this!

image


Corvince:
I think this looks awesome!

From peeking at the code it also seems more structured and robust. I would love to see this merged. Love the idea to have parameters user-settable by default and only disable them with a exclude list. 

Some things I noticed:

- No "time" control, i.e. fps per second
- The viz looks odd on (actually not so) wide screens. Controls on the left, viz on the right and big blank in the center. 
- No range for parameters. Would be nice to have a meaningful range for parameters. 

I don't think these are big problems, but maybe its important to keep them in mind. Especially the last point, need to discuss how this could be implemented.

Also an idea for a future update: Allow text boxes in the viz, for example to include a small description of the model above the user parameters.

TaylorMutch:
Ah yes I left a `Todo` to reimplement the FPS control. I'll add that into this PR.

I will play with the Bootstrap column widths to see about squeezing things more in the center.

I was thinking about the range control and description a bit. It may be sufficient to implement two small classes that resolve to a dictionary​ representation. For ranges, they resolve to `{min: min_value, max: max_value, default: value}`, and for textboxes `{text: 'My text here's}`. Then in the frontend we know how to handle these two signatures.

TaylorMutch:
@Corvince not all that far off from #354.

TaylorMutch:
@jackiekazil  @Corvince I implemented more robust user options, implemented using a new class, `Option`, on the backend which allows the frontend to have more flexibility. This is also coupled with better handling in the frontend code for options created using the `Option` class. It is also backwards compatible with existing examples that do not implement `Option`s when specifying parameters.

brylie:
Wow, looks really nice! 😀

dmasad:
Wow, this looks excellent! 

My one comment so far is that I think the server API should still take an explicit `model_param` dictionary, as inferring them from the `**kwargs` inputs. 

Other than that, we should probably implement the new visualization across all the example models prior to merging? I'll go ahead and start working on that.

TaylorMutch:
@dmasad I agree that an explicit kwarg `model_param` would make things a bunch simpler, rather than the mess I added for inferring the arguments from the server `**kwargs` and matching those up to the model.

Let me know if you run into any issues.!

dmasad:
I've started working on this in what apparently is a new branch https://github.com/projectmesa/mesa/tree/pr/383 
So far, I think the main issue (from my perspective) that I'm seeing is that the front-end is populating interface elements even for parameters that aren't explicitly `Option`s, and that interacting with those elements changes the underlying parameter. (See the Schelling model, for example). My sense is that the desired behavior is that only `Options` create interactive elements, with all other arguments hidden or displayed statically.

TaylorMutch:
@dmasad My only reason for keeping functionality for non-`Option` elements was for backwards compatibility. If that's not as important we could handle that differently.

A way this could be handled is by extending the `Option` class to have a `static` input type. This could accept strings or numbers, and then display those in a static text widget. Then, see [this switch statement](https://github.com/projectmesa/mesa/blob/pr/383/mesa/visualization/templates/js/runcontrol.js#L141) where we can handle the specified `Option` type `static` in the JS code.

TaylorMutch:
@dmasad I pulled in your changes from `pr/383` into this branch. I also added the "About" modal box we discussed, (optional) tooltip descriptions for user-settable model parameters in the visualization, and a method for inputting a simple, static textbox. It was a pretty light lift, so I thought it would be simple enough to add it into this PR.

Probably best to merge this in after we change the other models to use the `Option` interface.

dmasad:
Updated my branch with all the examples. Also tweaked the About tab to pull from the model docstring by default if there is no `description` property.

Things left to do:

-  Update the documentation
- Appease the linter

TaylorMutch:
@jackiekazil I refactored `Option` to `UserSettableParam`, since it has more semantic meaning for this context. I also updated the documentation for the `UserSettableParam` to match.

@dmasad It looks like there would be a decent amount of refactoring for the main documentation w.r.t. images and such. I don't have much time in the next week to do this but could circle around to it the following week.

I also pulled in your changes from pr/383, this PR is up to date.

dmasad:
Documentation updated! Pull that into this PR, take a look, and I think this might be good to go.

TaylorMutch:
@dmasad looks great! Ready to pull in?

EwoutH:
@TaylorMutch Hi! I'm looking into updating Bootstrap to a more modern version (4.6 and later 5.1). I see for bootstrap, bootstrap-slider and bootstrap-switch both a non-minified version (`.js`) as wel a a minified version is included (`.min.js`). Are both actually used? In the code I can only find references to the minified versions.

The [Bootstrap JavaScript docs](https://getbootstrap.com/docs/4.6/getting-started/javascript/) state:
> Plugins can be included individually (using Bootstrap’s individual `js/dist/*.js`), or all at once using `bootstrap.js` or the minified `bootstrap.min.js` (don’t include both).

EwoutH:
I opened an initial PR to test Bootstrap 4 here:
- https://github.com/projectmesa/mesa/pull/1100

PR #385, opened 2017-06-17
Add Sugarscape Constant Growback example

rht:
NetLogo specifies the model relatively more succinctly than Mesa. I wonder if the patterns in Netlogo should be ported to Mesa? For instance, `move` was:
```netlogo
to turtle-move ;; turtle procedure
  ;; consider moving to unoccupied patches in our vision, as well as staying at the current patch
  let move-candidates (patch-set patch-here (patches at-points vision-points) with [not any? turtles-here])
  let possible-winners move-candidates with-max [psugar]
  if any? possible-winners [
    ;; if there are any such patches move to one of the patches that is closest
    move-to min-one-of possible-winners [distance myself]
  ]
end
```

I have more comments to further simplify the implementation:
- the get_distance method should be available from the `space.Grid` class, not just
  in `space.ContinuousSpace`.
- Removing an agent should be a one-line method from the model class, which
  wraps the internal call `grid._remove_agent` and `schedule.remove`.
- `RandomActivationByBreed` should be moved to `mesa.time`, since this is used a lot in ABM models.


rht:
@dmasad merci!

PR #386, opened 2017-06-17
Scheduler: track agents with dict instead of list

### LINKED ISSUE #302: Removing an agent interferes with RandomActivation step method

rht:
Based on #305. Fix #302.

### DIFF
--- a/mesa/batchrunner.py
+++ b/mesa/batchrunner.py
@@ -154,7 +154,7 @@ class BatchRunner:
     def collect_agent_vars(self, model):
         """ Run reporters and collect agent-level variables. """
         agent_vars = {}
-        for agent in model.schedule.agents:
+        for agent in model.schedule.agents.values():
             agent_record = {}
             for var, reporter in self.agent_reporters.items():
                 agent_record[var] = reporter(agent)
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -43,7 +43,7 @@ class BaseScheduler:
         self.model = model
         self.steps = 0
         self.time = 0
-        self.agents = []
+        self.agents = {}
 
     def add(self, agent):
         """ Add an Agent object to the schedule.
@@ -53,7 +53,7 @@ class BaseScheduler:
             have a step() method.
 
         """
-        self.agents.append(agent)
+        self.agents[agent.unique_id] = agent
 
     def remove(self, agent):
         """ Remove all instances of a given agent from the schedule.
@@ -62,19 +62,19 @@ class BaseScheduler:
             agent: An agent object.
 
         """
-        while agent in self.agents:
-            self.agents.remove(agent)
+        del self.agents[agent.unique_id]
 
     def step(self):
         """ Execute the step of all the agents, one at a time. """
-        for agent in self.agents[:]:
-            agent.step()
+        agent_keys = list(self.agents.keys())
+        for agent_key in agent_keys:
+            self.agents[agent_key].step()
         self.steps += 1
         self.time += 1
 
     def get_agent_count(self):
         """ Returns the current number of agents in the queue. """
-        return len(self.agents)
+        return len(self.agents.keys())
 
 
 class RandomActivation(BaseScheduler):
@@ -92,9 +92,11 @@ class RandomActivation(BaseScheduler):
         random order.
 
         """
-        random.shuffle(self.agents)
-        for agent in self.agents[:]:
-            agent.step()
+        agent_keys = list(self.agents.keys())
+        random.shuffle(agent_keys)
+
+        for agent_key in agent_keys:
+            self.agents[agent_key].step()
         self.steps += 1
         self.time += 1
 
@@ -109,10 +111,11 @@ class SimultaneousActivation(BaseScheduler):
     """
     def step(self):
         """ Step all agents, then advance them. """
-        for agent in self.agents[:]:
-            agent.step()
-        for agent in self.agents[:]:
-            agent.advance()
+        agent_keys = list(self.agents.keys())
+        for agent_key in agent_keys:
+            self.agents[agent_key].step()
+        for agent_key in agent_keys:
+            self.agents[agent_key].advance()
         self.steps += 1
         self.time += 1
 
@@ -151,13 +154,14 @@ class StagedActivation(BaseScheduler):
 
     def step(self):
         """ Executes all the stages for all agents. """
+        agent_keys = list(self.agents.keys())
         if self.shuffle:
-            random.shuffle(self.agents)
+            random.shuffle(agent_keys)
         for stage in self.stage_list:
-            for agent in self.agents[:]:
-                getattr(agent, stage)()  # Run stage
+            for agent_key in agent_keys:
+                getattr(self.agents[agent_key], stage)()  # Run stage
             if self.shuffle_between_stages:
-                random.shuffle(self.agents)
+                random.shuffle(agent_keys)
             self.time += self.stage_time
 
         self.steps += 1
--- a/examples/sugarscape_cg/sugarscape/schedule.py
+++ b/examples/sugarscape_cg/sugarscape/schedule.py
@@ -1,5 +1,5 @@
 import random
-from collections import defaultdict
+from collections import OrderedDict
 
 from mesa.time import RandomActivation
 
@@ -14,11 +14,11 @@ class RandomActivationByBreed(RandomActivation):
 
     Assumes that all agents have a step() method.
     '''
-    agents_by_breed = defaultdict(list)
+    agents_by_breed = OrderedDict(dict)
 
     def __init__(self, model):
         super().__init__(model)
-        self.agents_by_breed = defaultdict(list)
+        self.agents_by_breed = OrderedDict(dict)
 
     def add(self, agent):
         '''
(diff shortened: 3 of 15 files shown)


mesa/batchrunner.py
```diff
@@ -118,7 +118,7 @@ def collect_model_vars(self, model):
     def collect_agent_vars(self, model):
         """ Run reporters and collect agent-level variables. """
         agent_vars = {}
-        for agent in model.schedule.agents:
+        for agent in model.schedule.agents.values():
```
rht: IIRC the ordering is non-deterministic in `python<3.6`. Yet deterministic execution is crucial in scientific simulations.

dmasad: I agree that preserving order is important. I'm not sure whether in Python <3.6 iterating over `values()` is guaranteed to preserve order any better than just iterating over the keys, and/or for that order to be the same as the order in which the agents were added (which should be the default non-random behavior).

rht: A sure-fire fix is to used `collections.OrderedDict`. But this will be slower since in Python 3 it is not written in C. But correctness > speed in this case.

rht: Actually, it has to be a `defaultdict` + `collections.OrderedDict`.

mesa/time.py
```diff
@@ -67,19 +67,19 @@ def remove(self, agent):
...
         """
-        while agent in self.agents:
-            self.agents.remove(agent)
+        del self.agents[agent.unique_id]
 
     def step(self):
         """ Execute the step of all the agents, one at a time. """
-        for agent in self.agents[:]:
-            agent.step()
+        agent_keys = list(self.agents.keys())
```
rht: Actually, in all the places where `.values()` are used, the order does not matter since they are for data collection/record or asserting properties. This is the only place where the order matters.
But in one way or another, order preserving still has to be enforced so that users running a Mesa simulation don't accidentally fall into the pitfall -- principle of least surprise.

rht:
@dmasad, I just figured that the deterministic run problem can be resolved with `PYTHONHASHSEED=0` (>=3.2.3).
I tested
```
PYTHONHASHSEED=0 python3.5 -c 'print({"GO:0000054": "ribosomal subunit export from nucleus","GO:0000278": "mitotic cell cycle","GO:0000746": "conjugation","GO:0000902": "cell morphogenesis","GO:0000910": "cytokinesis"})'
{'GO:0000746': 'conjugation', 'GO:0000910': 'cytokinesis', 'GO:0000278': 'mitotic cell cycle', 'GO:0000902': 'cell morphogenesis', 'GO:0000054': 'ribosomal subunit export from nucleus'}
```
I repeated more than 30 times, with all the result having the same order. If this were to happen by chance, the probability would be (0.25**30 = 8.67e-19).

Since there is no way to set `PYTHONHASHSEED` from a library, the only way to do this is via `os.environ['PYTHONHASHSEED'] = 0`.

rht:
There are 3 ways to resolve this:
1. `os.environ['PYTHONHASHSEED'] = 0`.
2. Stick to list, add an attribute to the agent, `self.alive`. I have seen this being used in several ABM's. Caveat: if lots of agents are created/destroyed over time, the list grows, which results in memory leaks and slow loops.
3. python>=3.6.

I am inclined with 1, since 3 makes it harder to run Mesa. The version in Debian stable as of this timestamp is still 3.5.3 https://packages.debian.org/stretch/python3, but 3.6.1 in macOS/Homebrew https://github.com/Homebrew/homebrew-core/blob/dd7a66070b32a07b31ca2cbf4b3509cad5971226/Formula/python3.rb#L4.

dmasad:
Is there a reason not to just use OrderedDicts?

rht:
Yes, that will do, `OrderedDict` and `.get('key', default_value)`, with reduced speed as a trade-off (but this can always be fixed later, as long as the implementation is correct). I will push this.

TaylorMutch:
@rht What's the status on this one? Looks like there is a conflict (possibly just stale).

rht:
Yeah, I have just rebased this.

rht:
Rebased again.

dmasad:
Closing this one since it's incorporated into #510

PR #389, opened 2017-07-01
Slight clarification of licensing language.

jackiekazil:
This is a slight clarification of the contribution language. I realize there was a grammar error. Let me know if any adjustments need to be made.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -67,11 +67,10 @@ With respect to code standards, we follow `PEP8`_ and the `Google Style Guide`_.
 .. _`Google Style Guide` : https://google.github.io/styleguide/pyguide.html
 
 
-Licensing
+Licensing of contributions
 --------
 
-The License of this project is located in `LICENSE`_. By submitting a contribution to this project, you are agreeing that your contributions under this license and
-with this waiver of copyright interest.
+The License of this project is located in `LICENSE`_. By submitting a contribution to this project, you are agreeing that your contributions are under this license and with this you waiver of copyright interest.
 
 .. _`LICENSE` : https://github.com/projectmesa/mesa/blob/master/LICENSE
 


CONTRIBUTING.rst
```diff
@@ -67,11 +67,10 @@ With respect to code standards, we follow `PEP8`_ and the `Google Style Guide`_.
 .. _`Google Style Guide` : https://google.github.io/styleguide/pyguide.html
 
 
-Licensing
+Licensing of contributions
 --------
 
-The License of this project is located in `LICENSE`_. By submitting a contribution to this project, you are agreeing that your contributions under this license and
-with this waiver of copyright interest.
+By submitting a contribution to this project, you are granting the project maintainers the ability to release your contribution under the project license located in `LICENSE`_. Along with this, you also waiver copyright interest.
```
rht: Hmm... What does 'you also waive~~r~~ copyright interest' mean? I think just the first sentence is sufficient.

rht: (I am not a Lawyer) But again, the `CONTRIBUTING.rst` has its legal impact only when a contributor has been recorded (in a way that can be shown as an evidence to the court) to read it.

dmasad: @jackiekazil I actually think the original language is more correct, since (as I understand it) the copyright waiver is part of the Apache license itself; this wording makes it sound like the copyright waiver is a separate thing.

@rht that's an interesting question. I think American courts at least are inconsistent on that (e.g. software EULAs that you click 'agree' to are binding, but 'by using this website you agree to ...' type things are not). I don't know what the rules are with regard to open-source contributions.

CONTRIBUTING.rst
```diff
@@ -67,11 +67,10 @@ With respect to code standards, we follow `PEP8`_ and the `Google Style Guide`_.
 .. _`Google Style Guide` : https://google.github.io/styleguide/pyguide.html
 
 
-Licensing
+Licensing of contributions
 --------
 
-The License of this project is located in `LICENSE`_. By submitting a contribution to this project, you are agreeing that your contributions under this license and
-with this waiver of copyright interest.
+The License of this project is located in `LICENSE`_. By submitting a contribution to this project, you are agreeing that your contributions are under this license and with this you waiver of copyright interest.
```
jackiekazil: To clarify the differences... 

Changed "your contributions under" to "your contributions are under"

Changed "with this waiver" to "with this you waiver"

CONTRIBUTING.rst
```diff
@@ -67,11 +67,10 @@ With respect to code standards, we follow `PEP8`_ and the `Google Style Guide`_.
 .. _`Google Style Guide` : https://google.github.io/styleguide/pyguide.html
 
 
-Licensing
+Licensing of contributions
```
jackiekazil: @dmasad -- Do you prefer "Licensing" or "Licensing of contributions" ... I kind of prefer the latter to be explicit, but I am open .

CONTRIBUTING.rst
jackiekazil: @dmasad does this work for what we discussed? Let me know if you want any changes on this.

CONTRIBUTING.rst
jackiekazil: @rht This is not an area of expertise of mine, but it is my understanding that this is a implicit contributor's agreement rather than explicit, which is an industry standard until you have a code base that is well adopted enough, which you go explicit via explicit signing of documents to contribute.... if software projects did this in the early days of projects no one would contribute.

Putting that aside, @dmasad and I chatted about it during the meeting and decided to keep the same language just add a little clarification of wording that made it a little hard to read.

rht: (I actually went through the licenses and TOS (for rigor's sake) had wanted to convey this far earlier; there is section 5 of Apache 2.0 for submission of contributions as @dmasad stated, there is also section D.6 in Github TOS which was only recently added [this year](https://github.com/blog/2327-new-github-terms-of-service-are-in-effect) to ease this for a sane default / to reduce bureaucracy excess. And should the Github servers experience a downtime, `git` is always there, which is distributed.)

PR #393, opened 2017-07-07
Split parameter_values into fixed & variable parameters in batchrunner

### LINKED ISSUE #307: Passing dictionary argument to BatchRunner

rht:
Fix #307. Based on #374.


rht:
cc: @devforfu @njvrzm @dmasad ?

dmasad:
Looks good!

PR #394, opened 2017-07-10
space: Add Mypy type annotation

rht:
Mypy type annotation ensures correctness of types of the variables and return-values.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -19,6 +19,10 @@ import random
 import math
 import numpy as np
 
+from typing import Any, Iterator, List, Tuple, Union
+from .agent import Agent
+Coordinate = Tuple[int, int]
+
 
 def accept_tuple_argument(wrapped_function):
     """ Decorator to allow grid methods that take a list of (x, y) position tuples
@@ -67,7 +71,7 @@ class Grid:
         is_cell_empty: Returns a bool of the contents of a cell.
 
     """
-    def __init__(self, width, height, torus):
+    def __init__(self, width: float, height: float, torus: bool) -> None:
         """ Create a new grid.
 
         Args:
@@ -79,7 +83,7 @@ class Grid:
         self.width = width
         self.torus = torus
 
-        self.grid = []
+        self.grid = []  # type: List[List[Any]]
 
         for _ in range(self.width):
             col = []
@@ -92,25 +96,29 @@ class Grid:
             *(range(self.width), range(self.height))))
 
     @staticmethod
-    def default_val():
+    def default_val() -> None:
         """ Default value for new cell elements. """
         return None
 
-    def __getitem__(self, index):
+    def __getitem__(self, index: int) -> List[Any]:
         return self.grid[index]
 
-    def __iter__(self):
-        # create an iterator that chains the
-        #  rows of grid together as if one list:
+    def __iter__(self) -> Iterator[Any]:
+        """
+        create an iterator that chains the
+        rows of grid together as if one list:
+        """
         return itertools.chain(*self.grid)
 
-    def coord_iter(self):
+    def coord_iter(self) -> Iterator[Tuple[Any, int, int]]:
+        # TODO precisify the Mypy type annotation
         """ An iterator that returns coordinates as well as cell contents. """
         for row in range(self.width):
             for col in range(self.height):
                 yield self.grid[row][col], row, col    # agent, x, y
 
-    def neighbor_iter(self, pos, moore=True):
+    def neighbor_iter(self, pos: Coordinate, moore: bool=True) -> Any:
+        # TODO precisify the Mypy type annotation
         """ Iterate over position neighbors.
 
         Args:
@@ -122,8 +130,8 @@ class Grid:
         neighborhood = self.iter_neighborhood(pos, moore=moore)
         return self.iter_cell_list_contents(neighborhood)
 
-    def iter_neighborhood(self, pos, moore,
-                          include_center=False, radius=1):
+    def iter_neighborhood(self, pos: Coordinate, moore: bool,
+                          include_center: bool=False, radius: int=1) -> Iterator[Coordinate]:
         """ Return an iterator over cell coordinates that are in the
         neighborhood of a certain point.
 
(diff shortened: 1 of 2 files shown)


dmasad:
know Python 3 has type hinting, but I'm not familiar with Mypy. If I'm understanding correctly it seems to be a work-in-progress library. Can you explain why you think it's particularly valuable here?

rht:
Mypy is PEP 484, since Python 3.5. Even if it is still a WIP library, having it should help with improving readability, safe refactoring, ... [1]. For scientific code in particular, this always helps, doesn't it?

[1] http://blog.zulip.org/2016/10/13/static-types-in-python-oh-mypy/.

rht:
I think the F401 flake8 check MUST be enabled, not sacrificed for the type check.

Either:
1. Use ' # NOQA' to disable flake8 just for the typing line (this is the current state, which accommodates for older versions of Python 3, and also Python 2.7; ironically one of the pragmatic reason for Mypy is to ease the migration to Python 3 in the first place). Currently, Travis checks for both 3.3 and 3.4.
2. Use PEP484 syntax.

The current version in Debian Stable is 3.5.3[1].
I can't see the use case of having to use 3.3/3.4 unless you have to run ABM simulations on embedded systems (e.g. on an autonomous self-driving vehicle with an outdated os firmware).

[1] https://packages.debian.org/stable/python3 (can't figure how to timestamp this URL to make it immutable :/)

rht:
> I can't see the use case of having to use 3.3/3.4 unless you have to run ABM simulations on embedded systems (e.g. on an autonomous self-driving vehicle with an outdated os firmware).

Debian 8 (until 2020), Ubuntu 14.04 (until 2019), and any distributions, container/vm images based on them are still on 3.4. Given the timeline constraint, the default should be to accommodate 3.4.

rht:
My mistake! The `->` syntax has been around since 3.0. It is just that `typing` becomes the stdlib since 3.5.

rht:
I have replaced the annotation with `->`.

jackiekazil:
Asked for more feedback on this explicitly through the dev listserv -- https://groups.google.com/forum/#!topic/projectmesa-dev/dPKX5Pckmds

mesa/space.py
```diff
@@ -80,38 +85,43 @@ def __init__(self, width, height, torus):
...
 
-    def __getitem__(self, index):
+    def __getitem__(self, index: int) -> List[Any]:
         return self.grid[index]
 
-    def __iter__(self):
-        # create an iterator that chains the
-        #  rows of grid together as if one list:
+    def __iter__(self) -> Any:
+        # TODO precisify the Mypy type annotation
```
TaylorMutch: Precisify? In terms of the grid values? I'm not sure what this means.

requirements.txt
```diff
@@ -7,4 +7,5 @@ coverage==4.1
 nose==1.3.7
 tornado==4.0.2
 flake8==2.5.2
+typing==3.6.2
```
TaylorMutch: Does the `typing` library at all conflict with Python `typing` module (>=3.5)?

TaylorMutch:
I like that this pr focuses the type annotations on a core class that really does have well-defined types, in which case I feel like the annotations really help the developer know what they are working with.

mesa/space.py
rht: Yes, it should be more specific that a generic `Any`, it should be `Iterator[Set[Any]]` maybe.

requirements.txt
rht: I think it doesn't, in that the stdlib one will be imported instead of this one.

mesa/space.py
```diff
@@ -15,9 +15,15 @@
...
-import numpy as np
 import random
 import math
+import numpy as np
+
+# \begin{mypy stuff}
+from typing import Any, Iterator, List, Tuple, Union
+from .agent import Agent
+Coordinate = Tuple[int, int]
+# \end{mypy stuff}
```
TaylorMutch: Do these comments have any utility in mypy? If not could they be removed?

rht: Ok, removing.

mesa/space.py
TaylorMutch: That looks like a good annotation, though for the base `Grid` class I wonder that since `default_val` returns `None`, though others return `set` (and potentially something different than either of those), that maybe `Iterator[Any]` might suffice.

rht: Ok, will put that in.

jackiekazil:
@rht I really hate turning down a contribution. I don't think I have ever done this before. However, it sounds like we are not in consensus to accept this yet. However, I think it is important to acknowledge the work that is happening in this PR and as this becomes a more solidified standard revisit in the future.

rht:
Ok, I will attempt again by the time there are official stubs for numpy (should be by 2020 the latest).

rht:
I found another side-effect of having a Mypy annotation in addition to correctness: faster perf for attribute access via cython than even using `__slots__` (e.g. https://github.com/AB-CE/abce/pull/178). This is possible because cython's latest release finally supported PEP 484 and 526 type hints instead of having to write a separate .pyx or .pxd file.

rht:
Also, experimental but official stubs: https://github.com/numpy/numpy-stubs. Adding annotation doesn't affect performance, and so can be done gradually. Is now the correct time?

Corvince:
By now I would be in favor of having type annotations, if the types can be well defined

rht:
Given sufficient time, the types can be as well defined as they would in statically typed languages. Precision is not a problem.

If you find the annotations to be too distracting (though some people may find the annotations to be indispensable, as evident from the fact that each params and rtype are always specified in any good docstring) to see the control flow, again, the files can be annotated in separate cython .pxd files (something along the line of c header files).

Corvince:
Sorry my comment was a bit short. I was referring to the mesa types. So one of the good things of type hints is that we need to think about which types are needed and returned. In your PR you often just have the Any type which I think is insufficient (and clearly marked by yourself with TODO comments)

rht:
It's gradual typing annotation, just like how in docstrings the param types and rtypes are added not all at once. I'm not sure why it has to be completed all in one go?

rht:
The decision is on whether to add type annotation or not. The completeness of the implementation comes after. Whether my implementation is half-baked, it has nothing to do with the decision.

`Any` is the default type that is then precisified later on. There is no harm with starting with `Any`.

PR #399, opened 2017-07-18
Email list language updates and link updates

jackiekazil:
These are minor updates to update email list links in the readme and the contributing to mesa doc.... basically it adds in the new dev email list --- https://groups.google.com/forum/#!forum/projectmesa-dev and pushes dev convos off of the original list, which is not referred to as the user list. 

If any has feels about the language I used, please feel free to chime in if the language I used isn't clear.


Corvince:
Thanks for the update

Issue #311, opened 2016-08-30
Most class variables are unnecessary


njvrzm:
A handful of the class-level variables in the project are actual class-level configuration, but most of them are shadowed by the `__init__` methods of their classes. This is untidy and could lead to confusion. I'd be happy to clean them up, unless there are objections.

jackiekazil:
Can you give an brief example & how it would change?

njvrzm:
Sure! It shouldn't affect the behavior of the code at all. One example would be BaseScheduler. The beginning of that class looks like this:
```python
class BaseScheduler:
    """ Simplest scheduler; activates agents one at a time, in the order
    they were added.

    Assumes that each agent added has a *step* method which takes no arguments.

    (This is explicitly meant to replicate the scheduler in MASON).

    """
    model = None
    steps = 0
    time = 0
    agents = []

    def __init__(self, model):
        """ Create a new, empty BaseScheduler. """
        self.model = model
        self.steps = 0
        self.time = 0
        self.agents = []
```

The class attributes set before the `__init__` are clobbered. It's mostly an aesthetic thing, but it could be confusing, especially to someone relatively new to Python, that making changes to those values won't do anything.

TaylorMutch:
I added #381 to start resolving this. Definitely agree with @njvrzm, a newcomer to Python would easily question whether this was necessary across all implementations, especially if one is coming from something like C/C++.

TaylorMutch:
@njvrzm I believe the unused class variables have been removed. Closing this issue.

PR #400, opened 2017-07-18
Remove unused class parameters.

TaylorMutch:
This is a (single commit) change to remove unused class parameters that are overwritten (or rather, never used), resolving #311.

@rht @jackiekazil When you fail once, try and try again, right?

### DIFF
--- a/mesa/datacollection.py
+++ b/mesa/datacollection.py
@@ -47,12 +47,6 @@ class DataCollector:
     functions one by one and stores the results.
 
     """
-    model_reporters = {}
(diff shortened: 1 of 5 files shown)


rht:
LGTM

(The answer on *that* question depends on the context. I'd say given a goal, and for any reinforcement learning method, a failure, as long as its consequence is not dire (e.g. death / out-of-sugar in a hostile env), is necessary for a checkpoint)

jackiekazil:
@rht ty for the review. Sorry for the delay in this.

PR #402, opened 2017-07-28
Add 3D rendering for CanvasGrids

TaylorMutch:
This PR opens up opportunities for rendering out grids in 3D. For simplicity, it sub-classes the `CanvasGridVisualization` and adds frontend components using ThreeJS to render out cubes over a plane.

I mainly want to have this here as both a discussion topic as well as solicit suggestions for additions and revisions. I will likely add to this PR with a couple bug fixes, but this is a starter.

I want to add this for the Flocker's example (as it seems like the more applicable example), but I started with WolfSheep.

image


jackiekazil:
As I catch up on things, I just want to say that your contributions have brought me such joy! <3 <3 <3

jackiekazil:
RE: "I want to add this for the Flocker's example (as it seems like the more applicable example), but I started with WolfSheep." 
- I see Flockers run3D.py, but it doesn't seem to work for me. 
- Also, I don't see the wolf sheep files at all. Did you check them in?

jackiekazil:
RE: Testing going down... can you update your branch with what it is Master, so when I look at test coverage, I am only looking at your changes?

TaylorMutch:
Sure I'll do a rebase and push it out.

jackiekazil:
Let me know when you are done & I will look again.

TaylorMutch:
@jackiekazil I'll get back to this today or tomorrow. Been at IEEE VIS all week, haven't had a moment to spare!

TaylorMutch:
@jackiekazil Okay, I'm finally looking at this again.

Coverage went down because of changes to `ModularVisualization.py`. There is a change here that is of consequence when it comes to loading JavaScript packages that are included locally. Instead of using a `set`, using a `list` encourages ordering of `package_includes` from visualizations that inherit from `VisualizationElement`.

The reason it is needed for the `CanvasGrid3D` class is that ThreeJS, a WebGL package used to render the objects in a WebGL `canvas` tag, must be loaded before the corresponding mouse controls package can be loaded.

Also, I did not include the WolfSheep 3D visualization files, but the 3D Flocker example seems to work on my machine. If folks have issues, then we should revisit it.

jackiekazil:
Discussion during today's meeting.... 
- This PR will be closed. 
- Are there any good examples that make use of 3-D that we can use? 
  - We need good example before we implement. 
  - @tpike3 will get example model from Matt.
- There should be airplanes flying over cars or airplanes flying over airplanes.

Closing for now... Taylor will resubmit when he is ready. No rush on this. Let's make sure we have a good use case implemented for this.

PR #409, opened 2017-09-04
Extended Grid to support hexagonal grids

JamesHovet:
Extended Grid to support hexagonal grids following the same conventions as other grids

TODO: 
- ~~add example~~
- ~~create visualization module~~

PS. I have no idea why the last test keeps failing on Travis, so if somebody could help me out with that I'd appreciate it.


Corvince:
I think this looks awesome! Good job 👍 

I didn't reviewe your code, although from a glimpse I could tell your readme.md still refers to game_of_life instead of hex_snowflake.

Regarding the Travis fail, it looks like everything runs smoothly except for the `./tests/test_end_to_end_viz.sh`

Since this is not associated with your pull request in any way my guess would be that this test broke outside your PR.

rht:
> Regarding the Travis fail, it looks like everything runs smoothly except for the ./tests/test_end_to_end_viz.sh

The viz test runs `examples/Flockers`. Possibly the server is not returning any response to curl which stalls the test indefinitely.

TaylorMutch:
@JamesHovet Taking a quick look at this, but this is super awesome.

Corvince:
Haha, I just realized I previously merged the wrong way around. I don't even know how I did this. 

Thanks @JamesHovet  for the great contribution

Issue #412, opened 2017-09-22
Slow ContinuousSpace


JamesArruda:
When using Mesa for its `ContinuousSpace` simulations with many agents and many steps take a while to run. Reducing the run time is valuable to me for performing Monte Carlo analyses and the like. Looking at the profiling, it appears that most of the time is spent in `get_distance` (I ran this for 100 steps on the Boids example):

              3018122 function calls in 4.453 seconds  
    
       Ordered by: internal time
    
       ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       136798    1.017    0.000    2.259    0.000 space.py:561(get_distance)
       146798    0.670    0.000    1.186    0.000 linalg.py:2014(norm)
        10000    0.261    0.000    4.423    0.000 <ipython-input-20-2c3c5a099d5f>:86(step)
    ....
        10000    0.184    0.000    2.548    0.000 space.py:524(get_neighbors)

This issue is being submitted according to the contribution guidelines that asks for tickets to be made for items that will be a PR. I have a proposed solution that'll be up as a PR.

Corvince:
I think the speedup from your PR is quite impressive

dmasad:
Merged!

PR #413, opened 2017-09-22
Modified ContinuousSpace to use pure Numpy for speedup

JamesArruda:
… Added tests for the alternate implementation.

The PR is meant to address issue #412

The idea was to replace the grid implementation with a Numpy array and perform the distance calculations all at once, rather than iterating over each agent in the nearby cells. Agents are mapped to the space's array through dictionaries that map agent to index, and index to agent. Most of the methods are the same. The primary changes are in `remove_agent`, `place_agent`, `move_agent`, and `get_neighbors`.

One change to look over is the initialization of the `ContinuousSpace` class. It no longer needs the grid width variables, so they have been removed. I added a `*args, **kwargs` to not break any examples or other code. 

The result of this code is the following improvement in the runtime of the Boids example model; a speedup of nearly 2x:

image

I opted to not use `np.linalg.norm` for the distance calculation based on testing that showed that doing the sum of squares on its own ('raw') and comparing to the squared radius was generally faster, although most of these times are probably negligibly different. I also compared using `mod` for the toroid calculation vs. numpy's array `minimum` comparison.  The mod appears to be much faster.

image

image

I also compared the pure `get_neighbors` methods between the proposed class and the current one over a range of agent counts. The green lines are from the proposed class. The darker the blue lines, the smaller the grid width used. This was tested on a unit grid.

comparison_toroid


rht:
Neat!

JamesArruda:
Consider this 'paused' until #425 is resolved.

JamesArruda:
I'm going to make a new PR that has a cleaner history. I did a rebase to catch up to master and now the diff to master is too extreme to understand this specific PR.

PR #421, opened 2017-11-08
Remove localhost hardcoding + allow secure sockets

shippy:
When I deployed a pet model to Heroku, it failed to open a socket. This happened for two reasons:

1. My browser defaults to a secure connection with Heroku, which would block any connection made over `ws://` rather than `wss://`,
2. `runcontrol.js` assumes that the Tornado server is only served on 127.0.0.1 and requires an explicit port specification.

Consequently, even though the web app starts, the canvas doesn't show because the socket cannot be opened.

I took a hint from [Heroku's help page](https://devcenter.heroku.com/articles/python-websockets#front-end) and tried making this PR's edit on the fly via Chrome debugger. This successfully connected to the socket and restored the visualization functionality both on localhost and on the deployed Heroku instance.

### DIFF
--- a/mesa/visualization/templates/js/runcontrol.js
+++ b/mesa/visualization/templates/js/runcontrol.js
@@ -43,7 +43,15 @@ var fpsControl = $('#fps').slider({
 var sidebar = $("#sidebar");
 
 // WebSocket Stuff
-var ws = new WebSocket("ws://127.0.0.1:" + port + "/ws"); // Open the websocket connection
+// Support TLS-specific URLs, when appropriate
+if (window.location.protocol == "https:") {
+    var ws_scheme = "wss://";
+} else {
+    var ws_scheme = "ws://"
+};
+
+
+var ws = new WebSocket(ws_scheme + location.host + "/ws"); // Open the websocket connection
 ws.onopen = function() {
     console.log("Connection opened!");
     send({"type": "get_params"}); // Request model parameters when websocket is ready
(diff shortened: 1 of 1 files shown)


mesa/visualization/templates/js/runcontrol.js
```diff
@@ -43,7 +43,15 @@ var fpsControl = $('#fps').slider({
...
-var ws = new WebSocket("ws://127.0.0.1:" + port + "/ws"); // Open the websocket connection
+// Support TLS-specific URLs, when appropriate
+if (window.location.protocol == "https:") {
+    var ws_scheme = "wss://";
+} else {
+    var ws_scheme = "ws://"
+};
+
+
+var ws = new WebSocket(ws_scheme + location.host + "/ws"); // Open the websocket connection
```
TaylorMutch: @shippy Could this be converted into a ternary expression? Unless the scheme needs to be used elsewhere in the frontend code? So something like this:

```
// Open the websocket connection
var ws = new WebSocket((window.location.protocol == "https:" ? "wss://" : "ws://") + location.host + "/ws"); 
```

shippy:
@TaylorMutch Done. Out of interest, was this an elegance judgment call, or is there a JS style guide I should have followed and didn't?

TaylorMutch:
@shippy This is mostly because the `ws_scheme` variable was only being used in one place and the result of the if-else was fairly small (though crucial). It's also a neat use of JS and keeps the code tidy.

Thank you for your work on this!

Issue #422, opened 2017-11-09
[Enhancement] Include fixed parameters in Batchrunner.get_model_vars_dataframe()


Corvince:
Hey all,

I really come to like the mesa included Batchrunner. However, I also want to keep track of the fixed parameters I used for the simulation run. Maybe others feel the same so I added a `include_fixed` option to the Batchrunner.get_model_vars_dataframe() function. Feel free to use my branch or even merge it.

Best
Vincent

TaylorMutch:
Closing as #423 has been merged.

PR #423, opened 2017-11-09
Batchrunfixedparameters

Corvince:
Added ability to include fixed parameters in get_model_vars_dataframe() function. See #422


TaylorMutch:
@Corvince is there any inherit issue with the `fixed_parameters` being included by default, and have folks opt out of automatically including the fixed parameters? And if the someone doesn't include `fixed_parameters` as part of the `BatchRunner`'s constructor, then there's no difference.

Also, since `_prepare_report_table` is a 'private' method anyway, could we instead have the `BatchRunner` class instead have a method that knows whether to include the fixed parameters or not?

TaylorMutch:
@Corvince would the following approach achieve the same effect?

In `BatchRunner.__init__`:
```
...
self.fixed_parameters = fixed_parameters or {}
self._include_fixed = len(self.fixed_parameters.keys()) > 0
...
```
In `BatchRunner._prepare_report_table`:
```
...
if self._include_fixed:
        for param in self.fixed_parameters.keys():
        ordered[param] = self.fixed_parameters[param]
return ordered
...
```

Corvince:
Thanks @TaylorMutch. That seems to me like a better approach and achieves the same result. I think it makes sense to have fixed parameters in by default.

PR #425, opened 2017-11-11
[hotfix!] Fixes #410: Wrong result got from get_distance.

### LINKED ISSUE #410: [Bug Report] Wrong result got from get_distance method for torus continuous space
```
  def get_distance(self, pos_1, pos_2):
        """ Get the distance between two point, accounting for toroidal space.

        Args:
            pos_1, pos_2: Coordinate tuples for both points.

        """
        pos_1 = np.array(pos_1)
        pos_2 = np.array(pos_2)
        if self.torus:
            pos_1 = (pos_1 - self.center) % self.size
            pos_2 = (pos_2 - self.center) % self.size
        return np.linalg.norm(pos_1 - pos_2)
```
Above is the get_distance method for continuous space in mesa 0.8.1.

Although I'm not clear about the algorithm here, when my input is
```
pos_1 = (50, 40)
pos_2 = (50, 50)
size = (100, 100)
torus = True
```
I got the result 90 rather than 10. 
So I wonder if there is any problem about this algorithm.

jackiekazil:


jackiekazil:
Looking for someone to do a review.

JamesArruda:
Looks good. My only suggestion is to put the second distance calculation and the min statement inside the `if self.torus:` block to avoid an excess `linalg.norm` call.

Corvince:
Closing this in favor of #430

Issue #420, opened 2017-11-04
[Bug Report] Torus adjustment check in Grid class problem


Moody1021:
It seems like a torus adjustment is not being made in Grid class in move_agent or _place_agent. Stack trace and code that shows problem can be referenced in a [mesa googlegroups](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/projectmesa/KUltKSC3MKg) thread.

TaylorMutch:
Closing now that #429 is merged.

PR #429, opened 2017-11-16
[BUGFIX] Torus adjustment in Grid class

JamesArruda:
This is a PR in response to #420 

The `torus_adj` method was updated to have a similar signature to the one in `ContinuousGrid`. `move_agent` now does a position check for toroid.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -162,8 +162,7 @@ class Grid:
                                        not (0 <= dy + y < self.height)):
                     continue
 
-                px = self.torus_adj(x + dx, self.width)
-                py = self.torus_adj(y + dy, self.height)
+                px, py = self.torus_adj((x + dx, y + dy))
 
                 # Skip if new coords out of bounds.
                 if(self.out_of_bounds((px, py))):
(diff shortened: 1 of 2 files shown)


mesa/space.py
```diff
@@ -246,11 +245,15 @@ def get_neighbors(self, pos, moore,
...
-    def torus_adj(self, coord, dim_len):
+    def torus_adj(self, pos):
         """ Convert coordinate, handling torus looping. """
-        if self.torus:
-            coord %= dim_len
-        return coord
+        if not self.out_of_bounds(pos):
+            return pos
+        elif not self.torus:
+            raise Exception("Point out of bounds, and space non-toroidal.")
```
TaylorMutch: Raise a bare `Exception`, while valid, is not very descriptive, and requires writing a message within the specific implementation.

If would be great if we could add an `AgentOutOfBoundsError` exception. So adding something like this to `mesa/space.py`:
```
# New Exception class
class AgentOutOfBoundsError(Exception):
        def __init__(self, position):
                super().__init__(self, "Agent position {} is out of bounds.".format(position))

# Usage
def torus_adj(self, pos):
...
        elif not self.torus:
                raise AgentOutOfBoundsError(pos)

JamesArruda: I had considered this, but didn't do since it didn't directly address the bug.

TaylorMutch: @JamesArruda that is fair enough. I think it's more important at this point to get the bug fix in. I can submit a PR that adds the `AgentOutOfBoundsError`.

PR #434, opened 2017-11-25
Removed mutable defaults from DataCollector constructor

TaylorMutch:
Mutable defaults for class constructors can cause undesired effects. See the [Hitchhiker's guide](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments) on this issue.

### DIFF
--- a/mesa/datacollection.py
+++ b/mesa/datacollection.py
@@ -50,7 +50,7 @@ class DataCollector:
 
     model = None
 
-    def __init__(self, model_reporters={}, agent_reporters={}, tables={}):
+    def __init__(self, model_reporters=None, agent_reporters=None, tables=None):
         """ Instantiate a DataCollector with lists of model and agent reporters.
 
         Both model_reporters and agent_reporters accept a dictionary mapping a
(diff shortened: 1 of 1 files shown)


mesa/datacollection.py
```diff
@@ -80,14 +80,17 @@ def __init__(self, model_reporters={}, agent_reporters={}, tables={}):
         self.agent_vars = {}
         self.tables = {}
 
-        for name, func in model_reporters.items():
-            self._new_model_reporter(name, func)
+        if isinstance(model_reporters, dict):
```
rht: Would it be faster or more idiomatic with `if model_reporters is not None`?

TaylorMutch: TL;DR - I'll change it.

Long version - In this case, since the expected type is of type `dict`, it makes sense that we only want objects that are of that type. Though I suppose you'll just get errors if you pass in something that doesn't have a `__getitem__`  method. In think about this more, you're probably right that it should just be a check against `None`, and then the user can dig through the stack trace to discover that the object that they're wanting to use is a dictionary of some kind.

TaylorMutch:
@rht See changes.

mesa/datacollection.py
rht: Yeah, with `isinstance(model_reporters, dict)`, the error doesn't show up soon when it is not a dict.

PR #436, opened 2017-12-03
[BUGFIX] Fix sidebar visibility in Edge

TaylorMutch:
Uses less jQuery and more template-styled rendering.


TaylorMutch:
@Corvince thanks for looking it over! I'll go ahead and merge it.

PR #437, opened 2017-12-07
Update to model.py data collection

cgit2017:
I noticed that in model.py the DataCollector wasn't doing anything with the data, so I updated it to create data files at certain step counts during the run.

### DIFF
--- a/examples/bank_reserves/bank_reserves/model.py
+++ b/examples/bank_reserves/bank_reserves/model.py
@@ -1,3 +1,6 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
 from bank_reserves.agents import Bank, Person
 from mesa import Model
 from mesa.space import MultiGrid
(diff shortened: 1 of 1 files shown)


TaylorMutch:
@cgit2017 Aside, I am now a bit confused, and possibly should have addressed this in the first PR. In the `bank_reserves` example, there are two files that are nearly identical. `batch_run.py` and `bank_reserves/model.py`. Is there a need to have both? There is also a bunch of duplicated code there.

TaylorMutch:
@cgit2017 I think it makes sense to want to output data at intervals within a batch run, but having the model itself contain the logic for outputs doesn't seem like the right approach. However, your change might be better suited being applied more generally to the [BatchRunner](https://github.com/projectmesa/mesa/blob/master/mesa/batchrunner.py) class in some way!

cgit2017:
@TaylorMutch I think that's a better idea. I guess you'll reject the pull request? And I can start working on that idea for the BatchRunner.

TaylorMutch:
I'll go ahead and close it. Also, I would suggest the following when developing `mesa`:
- Create a Python virtual environment. This will help prevent issues with global Python packages being pulled into your development code. You can create a virtual environment as follows:
```
# You can substitute 'mesa-env' for whatever name you want to call your virtual environment
> python -m mesa-env 

# Now 'activate' your virtual environment
# On Windows, substitute 'Scripts' for 'bin'
> mesa-env/bin/activate

# Your terminal should now start with the name of your virtual environment
(mesa-env) > 
```
I hope this helps, and always feel free to send questions to the Mesa Google Group!

Issue #268, opened 2016-06-27
x-tick=1 corresponds to step=0 in Chart plotting in Canvas


jeetu7:
I tried to plot the chart wherein the y axis represents number of cells remains unexplored. Effectively this translates to requirement where output is non-zero(y=100 on y-axis) at step 0. But while plotting, the point corresponding to step 0 was plotted at tick 1 (on x-axis). Therefore a line was visible from (0,0) to (1,100) on (x, y) plot. But ideally the plot should start from (0,100). Seems some indexing is getting lost while calling Chart.js library.  This could be taken up with #77 issue.

jackiekazil:
👍

Issue #77, opened 2015-01-15
Make the charting visualization not suck


dmasad:
The charting module (currently using [Charts.js](http://www.chartjs.org/) is (barely) functional, but ugly. The defaults should be made nicer, and the configuration should be more Pythonic.

jeetu7:
When number of steps are large (say 100 or 1000) the ticks on x-asix looks very cluttered. The autoSkip option might help in this case. See http://www.chartjs.org/docs/#scales-ticks-configuration

jackiekazil:
@jeetu7 That sounds like a great idea!

jackiekazil:
This has made progress forward thanks to @cauemello!!
For now, I am closing this ticket lieu of more specific tickets.

PR #438, opened 2017-12-07
Corrects a bug in Chart.js Module and some other issues

cauemello:
This PR corrects:
1- a bug occurring in chart when the reset button was pressed.
2- the ticks visualization on x-axis when chart has too many values => #77 
3- the datacollector data with the corresponding model step * => #268 

* In order to work properly, we must collect data after defining the initial state of the model (step 0) and after each step. So, we must collect data as the last line of model __init__ and after each schedule.step().

### DIFF
--- a/examples/epstein_civil_violence/civil_violence/model.py
+++ b/examples/epstein_civil_violence/civil_violence/model.py
@@ -51,7 +51,6 @@ class CivilViolenceModel(Model):
         self.active_threshold = active_threshold
         self.arrest_prob_constant = arrest_prob_constant
         self.movement = movement
-        self.running = True
         self.max_iters = max_iters
         self.iteration = 0
         self.schedule = RandomActivation(self)
@@ -93,7 +92,7 @@ class CivilViolenceModel(Model):
                 self.grid[y][x] = citizen
                 self.schedule.add(citizen)
 
-        # collect initial data
+        self.running = True
         self.datacollector.collect(self)
 
     def step(self):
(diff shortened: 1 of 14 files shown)


mesa/visualization/templates/js/ChartModule.js
```diff
@@ -80,12 +83,10 @@ var ChartModule = function(series, canvas_width, canvas_height) {
...
 
     this.reset = function() {
-        chart.destroy();
-        chartData.labels = [];
-        chart = new Chart(context, {
-            type: 'line',
-            data: chartData,
-            options: chartOptions
+        while (chart.data.labels.length) { chart.data.labels.pop(); }
+        chart.data.datasets.forEach((dataset) => {
```
TaylorMutch: While I think it is by and large appropriate to use `=>` notation for declaring functions, since we are not using any sort of build process, it is best for the time to use the long-form declaration of functions.
```
chart.data.datasets.forEach(function(dataset) {
    while (dataset.data.length) {
        dataset.data.pop();
    }
})

File: examples/epstein_civil_violence/civil_violence/model.py [hunk 2]
```diff
@@ -93,12 +93,16 @@ def __init__(self, height, width, citizen_density, cop_density,
                 self.grid[y][x] = citizen
                 self.schedule.add(citizen)
 
+        # collect initial data
```
TaylorMutch: The duplicated `collection initial data` comments seem unnecessary given that we're calling `datacollector.collect`.

TaylorMutch:
@cauemello I will look over this tonight and see if we can get this taken care of.

cauemello:
Hi @TaylorMutch , is this PR out of Mesa 0.8.3 release? Did you find any problem with the code?

PR #440, opened 2017-12-21
[WIP] Interactively add visualizations

Corvince:
Dear all,

I am working on some changes to the visualization. The idea is to make the model reporter functions available as a selection and then create charts/visualizations on the fly. 

Here is a quick illustration of the current status:
https://drive.google.com/file/d/1qqimtKRjwkSQGSKIdudxTQKpKza-YX_h/view
(the screencast is kind of bad, because it not really shows the drag-and-drop part)

There are still a lot of bugs and probably unnecessary code since it is basically hacked together

I still post this already in case anyone wants to pick up the work, since I wont be able to work on this for the next 2 months.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -11,7 +11,7 @@ In no particular order, examples include:
 - Documentation improvements
 - Tutorials
 
-No contribution is too small. Although, contributions can be too big, so let's discuss via the `email list`_ OR `an issue`_.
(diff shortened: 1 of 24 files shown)


jackiekazil:
@Corvince -- This looks like a great idea, but it also looks like it has stalled. If you pick it up and finish it, feel free to resubmit as another PR. I would LOVE to see this added! I am moving this over to a ticket for tracking -- #537

PR #443, opened 2017-12-31
TODO: use model random for consistency

aadeshnpn:
Completed todo list of using model class random generator for all other classes


dmasad:
This looks good! Is there a reason you removed the `__init__.py` file? The original file contents should probably be restored there.

aadeshnpn:
I am not sure if I removed __init__,py. Travis-ci is failing. I made the changes related to random TODO list. Can anyone approve the changes?

dmasad:
One issue with this implementation: since `model.random` is now created in the `Model.__init__` method, models need to either call `super().__init__` or explicitly initialize their own `random` property.

(To see the issue, try to run e.g. the Schelling example; it raises an `AttributeError: 'SchellingModel' object has no attribute 'random'` error. )

There are a few ways I can see addressing this
1. Keeping it as-is, and refactoring the examples to confirm.
2. Having the Scheduler constructors check if the model seed has been set, and if not, setting it themselves.

Thoughts?

aadeshnpn:
Which approach among those 2 you think would be appropriate? I would go with approach 2.

aadeshnpn:
@dmasad, I have changed all the example to include super().__init__ the method call to initialize the model superclass.

dmasad:
This is something I'd like to get a little more discussion going on. Do we want calling `super().__init__` to be mandatory / highly encouraged?

jackiekazil:
Task for Jackie: Get opinions -- should be calling super().__init__ or is there is some better way of doing this?

jackiekazil:
I requested folks to chime in from Scipy... hopefully someone does. :-)

mesa/model.py
```diff
@@ -24,13 +23,13 @@ def __init__(self, seed=None):
             running: a bool indicating if the model should continue running
 
         """
-        # seed both the numpy and Python random number generators
+        # seed both the numpy random number generators
         if seed is None:
-            self.seed = dt.datetime.now()
+            self.seed = int(dt.datetime.timestamp(dt.datetime.now())) % 39916801
```
werwty: How about setting `seed` and `random` as class attributes instead?
```
class Model:
    seed =  int(dt.datetime.timestamp(dt.datetime.now())) % 39916801
    random =  numpy.random.RandomState(self.seed)
   
    def __init__(self, seed=None):
        if seed:
            self.seed=seed
            self.random=numpy.random.RandomState(self.seed)
```

This way anyone who subclasses `Model` can set their own instance attribute to override the class attribute. But if they don't want to override, there's still a good default they get without having to call  `super().__init__()` (and thing won't break if people forget to call `super().__init__()`

jackiekazil:
Discussion from tonight... @werwty @dmasad 
- assigning @ class time won't work. :-( 
- we are going to try experimenting with __new__ override.

werwty:
__new__ seems to work. 

```
>>> class A():
...     def __new__(cls):
...         cls.test=random.random()
...         return super(A, cls).__new__(cls)
... 
>>> b=A()
>>> b.test
0.25672892633328914
>>> c=A()
>>> c.test
0.4568693535366597
```

Corvince:
Closed in favor of #572 

Thanks again for this work!

PR #445, opened 2018-01-01
Datacollector fix

dmasad:
Simplify the `DataCollector` API by allowing the user-provided model- and agent-level reporters to be the names of attributes, in addition to methods. e.g. instead of needing to write

```
agent_reporters={"Wealth": lambda a: a.wealth}
```

you can write
```
agent_reporters={"Wealth":"wealth"}
```

This PR implements this feature, and updates the tests and documentation accordingly.


jackiekazil:
I just went through this. It all looks good! Ty for the update! I don't have anything specific that I can see that needs to be changed.

Issue #455, opened 2018-01-19
[Feature Request]: Multiprocessing BatchRunner


Maarten-vd-Sande:
The current implementation of BatchRunner.run_all() 'spawns' only a single process for all the different iterations and parameter combinations. By implementing the function in a multiprocessing approach, modern computers can fully leverage their multiple processors.

jackiekazil:
YAY!

Issue #454, opened 2018-01-19
[Bug report] Setting iterations greater than 1 in Batchrunner has no effect 


wesselklijnsma:
If the iteration parameter for a Batchrunner is greater than 1, the model will immediately reach max_steps after one run of a parameter combination. This is because the model is initialised once per parameter combination.  See code below (from run_all() in batchrunner.py):          

```
            for param_values in product(*param_ranges):
                kwargs = dict(zip(param_names, param_values))
                kwargs.update(self.fixed_parameters)
                model = self.model_cls(**kwargs)

                for _ in range(self.iterations):
                    self.run_model(model)
                    # Collect and store results:
```

I think moving the model initialisation to the inner loop would be a solution to this problem. See code below:

```
            for param_values in product(*param_ranges):
                kwargs = dict(zip(param_names, param_values))
                kwargs.update(self.fixed_parameters)

                for _ in range(self.iterations):
                    model = self.model_cls(**kwargs)
                    self.run_model(model)
                    # Collect and store results:
```

ericmjl:
I will concur on this; moving into inner loop solved a problem that I encountered with my batchruns giving identical results and running for order-of-magnitude shorter-than-expected amount of time.

ihopethiswillfi:
I just found out about this issue.
I used the same solution in this PR: https://github.com/projectmesa/mesa/pull/486

jackiekazil:
It looks like this is fixed. Please tell me otherwise if it is not. 
Thank you everyone for your work on this! YAY!

PR #456, opened 2018-01-19
[Feature Implemented]: Multiprocessing BatchRunner 

Maarten-vd-Sande:
See issue https://github.com/projectmesa/mesa/issues/455

This implementation of the BatchRunner Class is capable of using any number of processors when run_all() is called. Ǹow a process is spawned for each parameter combination and each iteration, and these processes are handled by a non-blocking and unordered worker pool. The max number of processors used can be specified as argument for run_all(). The default Python multiprocessing utilities could not be used, because lambda functions can not be pickled. Fortunately there is a fork of multiprocessing, called pathos, which is capable of doing this :). 

This implementation is also a fix for the bug discovered in issue 
https://github.com/projectmesa/mesa/issues/454 by @wesselklijnsma

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,36 @@
 Release History
 ---------------
 
+0.8.3 (2018-06-01) Hayden
++++++++++++++++++++++++++++++++++++++++++++
+
+**Improvements**
+
+* Datacollector fix #445
+* A first network grid model with visualization, using NetworkX and sigma.js #388
+* Cache pip packages for Travis setup #427
+* Remove localhost hardcoding + allow secure sockets #421
+* Update Chart.js to version 2.7.1 #401
+* Bank reserves example #432
+* Extended Grid to support hexagonal grids #409
+
+**Fixes**
+
+* Faster ContinuousSpace neighbor search #439
+* Updating license year to 2018 #450
+* Updating language on license in contributing file #446
+* Updating license year to 2018 #450
+* Removed mutable defaults from DataCollector constructor #434
+* [BUGFIX] Torus adjustment in Grid class #429
+* Batchrunfixedparameters #423
+* [BUGFIX] Fix sidebar visibility in Edge #436
+* Updating Travis svg to target #master, not branches. #343
+* Email list language updates and link updates #399
+* Fix math problems in flockers; use numpy in space #378
+* Only start tornado ioloop if necessary #339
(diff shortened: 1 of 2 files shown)


JamesArruda:
Cool!

What caused you to go with `pathos` over the built-in multiprocessing?

Maarten-vd-Sande:
The built-in multiprocessing depends on pickle. To my understanding, pickle currently does not support lambda functions, on which the BatchRunner depends. Pathos, however, depends on dill, which does support lambda functions.

Maarten-vd-Sande:
Storing the results in between runs results in the BatchRunner getting larger as the nr of runs grows. This resulted in unstable behaviour when the model_vars and agent_vars got too large. Fixed.

dmasad:
Maarten -- this looks really good! How would you feel about moving this feature into a new class (e.g. `BatchRunnerMP`) that inherits from `BatchRunner`? That would leave the code for the plain non-parallel `BatchRunner` cleaner and easier to read, and would allow Pathos to be an optional dependency rather than a required one.

Maarten-vd-Sande:
Seems like a reasonable thing to do, since the current implementation isn't the cleanest. I'll post an update on of these days.

Did (any of) you had a chance to look at https://github.com/projectmesa/mesa/issues/454.

Maarten-vd-Sande:
The current format is that everything related to one topic (space, time, batchrunner, etc.) is in one file. If I would add the BatchRunnerMP to the same batchrunner.py file, the problem is that pathos is not an optional dependency anymore. I can however add the import in the init() of the BatchRunnerMP, which would solve this problem. Would this be the desired format? Or would you prefer two different files?

Maarten-vd-Sande:
The current format is that everything related to one topic (space, time, batchrunner, etc.) is in one file. If I would add the BatchRunnerMP to the same batchrunner.py file, the problem is that pathos is not an optional dependency anymore. I can however add the import in the init() of the BatchRunnerMP, which would solve this problem. Would this be the desired format? Or would you prefer two different files?

dmasad:
Sorry for not responding to this earlier! I can think of two ways to do it: one is to make importing pathos optional using a try/catch import at the top of the file, and have BatchRunnerMP throw an exception in init if pathos isn't present; the other is to put the import in the BatchRunnerMP constructor itself. My sense is that the first way is more Pythonic.

Issue #458, opened 2018-01-24
[Bug report] iter_neighborhood() gives incorrect neighborhoods for both von Neumann and Moore


WouterVrielink:
Moore's neighborhood is defined as:
![Moore neighborhood](http://mathworld.wolfram.com/images/equations/MooreNeighborhood/NumberedEquation1.gif)
Or see [this link](http://mathworld.wolfram.com/MooreNeighborhood.html) for a more complete explanation.

In the code however, it is calculated by excluding everything that is outside the Euclidean distance:
```Py
if moore and (dy ** 2 + dx ** 2) ** .5 > radius > 1:
    continue
```
This can be corrected by just removing the condition, as the correct indices are already visited.

Von Neumann's (extended) neighborhood is defined as:
![von Neumann neighborhood](http://mathworld.wolfram.com/images/equations/vonNeumannNeighborhood/NumberedEquation1.gif)
Or see [this link](http://mathworld.wolfram.com/vonNeumannNeighborhood.html) for a more complete explanation.

In the code however, it is calculated by excluding any diagonals, which makes it more like a huge cross:
```Py
if not moore and dy != 0 and dx != 0:
    continue
```
This can be corrected as follows:
```Py
if not moore and abs(dx) + abs(dy) > radius:
    continue
```

PR #459, opened 2018-01-24
[Bugfix]  iter_neighborhood() now gives correct neighborhoods for both von Neumann and Moore

WouterVrielink:
See issue #458.


dmasad:
Looks good to me!

PR #464, opened 2018-01-31
Added Mesa Packages pages

tpike3:
I added: 

1. a mesa packages section in index.rst
2. a mesa packages main page
3. a package development checklist  page (tried to do the most basic I could think of, then will create an expanded version which gets them to PyPI standards)
4. a package template page
5. a user guide for packages page

I have an example package ready to add to the wiki page, when/if this is merged. Otherwise, I will regroup and try again.

### DIFF
--- a/docs/packages/package_main.rst
+++ b/docs/packages/package_main.rst
@@ -17,7 +17,7 @@ The Mesa core functionality is just a subset of what we believe researchers crea
 
 The purpose of this guide is to support new programmers by providing detailed guides on "How to" build and share a package. Let's get started!
 
-	You want add to create your package but have never done it before -- start here: :ref:`package-development`
+	You want create and share your package but have never done it before -- start here: :ref:`package-development`
 
 	You want to incorporate a package from a MESA user but don't know how--- try this: :ref:`user-guide`
 
--- a/docs/packages/package_main.rst
+++ b/docs/packages/package_main.rst
@@ -1,0 +1,65 @@
+.. _Mesa-Packages: 
+
+Mesa Packages 
+==============
+
+*Supporting MESA Agent Based Modeling with Crowd Sourced Packages*  
+------------------------------------------------------------------
+
+	*Bottom up models are virtual laboratories where controlled experiments distinguish noise from signal in the systems organization....
+	This approach may change our whole notion of scientific theory, which until now has been based on the theories of physics. Theories of
+	complex systems may never be reducible to simple analytical equations, but are more likely to be sets of conceptually simple mechanisms
+	(e.g. Darwinian natural selection) that produce different dynamics and outcomes in different contexts.*
+
+	Volker Grimm et al, "Pattern Oriented Modeling of Agent Based Complex Systems: Lessons from Ecology" Science, New Series, Vol. 310, No. 5750
+	(Nov. 11, 2005), pp. 987-991
+
+
+Mesa Packages Purpose: 
+----------------------
+
+
+**MESA Packages is a repository to share packages for use in Agent Based Models (ABMs) in pursuit of "unifying algorithmic theories of the relation between adaptive behavior and system complexity."**
+
+	**(1) Provide code which can be integrated into MESA ABMs to enable greater exploration of complex phenomenon.**
+	
+	**(2) Support new programmers by providing detailed guides on "How to" build and share a package.**
+
+
+Mesa Packages Overview: 
+-----------------------
+
+Mesa Packages does not have the capacity to verify the coding, the accuracy or the effectiveness of underlying theory and its instantiation in the supplied package. To ensure supplied packages are understandable and usable please adhere to the usability and veracity guidelines. 
+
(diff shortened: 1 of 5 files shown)


jackiekazil:
As discussed, let's get this merged into a branch for review on readthedocs... people submit a PR against the mesa_packages_pages branch.

docs/packages/package_main.rst
```diff
@@ -0,0 +1,65 @@
+.. _Mesa-Packages: 
+
+Mesa Packages 
+==============
+
+*Supporting MESA Agent Based Modeling with Crowd Sourced Packages*
```
jackiekazil: I think we should remove this quote. This is because it is more general and gets away from the content.

jackiekazil:
You can view the build here: https://github.com/projectmesa/mesa/pull/464/files#diff-f633d6492baeaaac3fdcffc2083a20c1

jackiekazil:
You can view doc build here: http://mesa.readthedocs.io/en/mesa_packages_pages/packages/package_main.html

docs/packages/package_main.rst
```diff
@@ -0,0 +1,65 @@
...
+
+	Volker Grimm et al, "Pattern Oriented Modeling of Agent Based Complex Systems: Lessons from Ecology" Science, New Series, Vol. 310, No. 5750
+	(Nov. 11, 2005), pp. 987-991
+
+
+Mesa Packages Purpose: 
+----------------------
+
+
+**MESA Packages is a repository to share packages for use in Agent Based Models (ABMs) in pursuit of "unifying algorithmic theories of the relation between adaptive behavior and system complexity."**
```
jackiekazil: Remove bold from this section to be consist with styling of other sections.

docs/packages/package_main.rst
```diff
@@ -0,0 +1,65 @@
...
+	*Bottom up models are virtual laboratories where controlled experiments distinguish noise from signal in the systems organization....
+	This approach may change our whole notion of scientific theory, which until now has been based on the theories of physics. Theories of
+	complex systems may never be reducible to simple analytical equations, but are more likely to be sets of conceptually simple mechanisms
+	(e.g. Darwinian natural selection) that produce different dynamics and outcomes in different contexts.*
+
+	Volker Grimm et al, "Pattern Oriented Modeling of Agent Based Complex Systems: Lessons from Ecology" Science, New Series, Vol. 310, No. 5750
+	(Nov. 11, 2005), pp. 987-991
+
+
+Mesa Packages Purpose:
```
jackiekazil: If you remove the quote above, I think you can remove this header.

docs/packages/package_main.rst
jackiekazil: Also... b/c people creating packages that are independent of a single repository maybe change the text to something like this.... (feel free to edit what I am writing if it needs to be smoothed out....)

Suggested text:
###
The core functionality that comes with Mesa is just a foundational subset of functionally that we believe researchers creating Agent Based Models (ABMs) user. We designed Mesa to be extensible, so that individuals from various domains can build and maintain their own Mesa packages in pursuit of "unifying algorithmic theories of the relation between adaptive behavior and system complexity."
###

jackiekazil: From the text to the numbered items below, there needs to be transition.... 
I am not sure how to tie the two together. For example.... 

###
The purpose of this guide is to.... 
We hope to ..... 
###

Reading the numbers I am thinking that num. 1 no longer stands, and num. 2 can be rolled into the paragraph above with something like... 

###
The purpose of this guide it to support new programmers by providing detailed guides on "How to" build and share a package. Let's get started!
###

^^ Thoughts?

docs/packages/package_main.rst
```diff
@@ -0,0 +1,65 @@
...
+
+
+**MESA Packages is a repository to share packages for use in Agent Based Models (ABMs) in pursuit of "unifying algorithmic theories of the relation between adaptive behavior and system complexity."**
+
+	**(1) Provide code which can be integrated into MESA ABMs to enable greater exploration of complex phenomenon.**
+	
+	**(2) Support new programmers by providing detailed guides on "How to" build and share a package.**
+
+
+Mesa Packages Overview:
```
jackiekazil: I feel like I need to stop the review here pending updates. This is because some of what is written is written for a collective repository of packages, which is not what we are encouraging. We are encouraging people to write their own packages under their own github usernames and then make the community award of them, which can be done via the list serv and maybe if they notify us, we can add it to our wiki page until one day when create a directory of packages that are floating out in the world. 

^^ Can you update the docs to send that message to folks?

jackiekazil:
Also, I made a small update in main file. I don't think it will come into conflict with yours, but you may want to [update your branch in your fork](https://stackoverflow.com/questions/3903817/pull-new-updates-from-original-github-repository-into-forked-github-repository).

jackiekazil:
I am going to kill this PR since we are working on this in a different way.

PR #465, opened 2018-02-05
Added mouse interactionHandler to close #457, fixed hexgrid drawLines

### LINKED ISSUE #457: [Feature Request]: Agent data on mouse over visualization
Is it possible to bundle agent data into the web browser side json so that when I mouse over an agent I can view that agent's data? Seems like it would mean adding some more fields to the portrayal method to include data that would only available on a mouse over. And of course javascript code to display that data on a mouse over. I don't know if the performance penalty would be too great? This would be particularly useful in early stages of model debugging and perhaps even later. If performance is an issue, it could be feature invoked only if needed.
Thanks

RMKD:
* enables viewing of portrayal values in a tooltip on mouseover
* ignores standard visual configuration features such as Shape, Color, x, y, etc
* supports multiple agents in the same coordinate space
* integrated into CanvasModule and CanvasHexModule but could be extended to other visualizations


mesa/visualization/templates/js/InteractionHandler.js
```diff
@@ -0,0 +1,165 @@
+
+/**
+Mesa Visualization InteractionHandler
+====================================================================
+
+This uses the context of an additional canvas laid overtop of another canvas
+visualization and maps mouse movements to agent position, displaying any agent
+attributes included in the portrayal that are not listed in the ignoredFeatures.
+
+The following portrayal will yield tooltips with wealth, id, and pos:
```
jackiekazil: What happens when there is a model that doesn't have agent wealth, but has some other attributes?... let's say ... level of trust, memory, and health?

RMKD:
This runs by default via the modifications to CanvasModule and CanvasHexModule. To test, add attributes to the portrayal an mouse over an agent:

For the bank_reserves example: 
```
portrayal['savings'] = agent.savings
portrayal['loans'] = agent.loans
```
mouseover

jackiekazil:
The rollover seems to work for me, but rendering isn't. :-( 
This is what is looks like in Chrome Version 66.0.3359.117. 

screen shot 2018-05-14 at 5 11 12 pm

SO CLOSE!!!

RMKD:
Looks like maybe layout conflicts with the visualizer from #468 - I'll look into this

jackiekazil:
@RMKD awesome! Thank you!

jackiekazil:
@RMKD can you provide an update on this? I am at Scipy Sprints and wondering if this is something that someone else can pick up OR if you made progress?

RMKD:
Sorry for the long delay. Minor tweaks to the css place the world grid under the monitor (let me know if that should be the other way around). Also pulled the css out to make updating a little cleaner.

jackiekazil:
LGTM! 🎉 🎊 🙌 

Thank you for all your hard work on this!!!

PR #468, opened 2018-02-10
[Feature]: Change size and tooltip text of nodes in D3 network visualization

cauemello:
Now we can change size and tooltip text of nodes in D3 network visualization. 
Examples have been updated to this new feature and optimized for the last NetworkX release.


dmasad:
This looks good to me! I just want to make sure there's no redundancy with #465.

cauemello:
This PR is much simpler and different from #465 because we use D3 to render the network.
But this solves #457 feature request for agents in network view.

PR #475, opened 2018-02-26
[Bugfix]: show grass color in wolf sheep example model

### LINKED ISSUE #474: [Bug Report] appearance of grass in wolf-sheep model
Small issue, but I've noticed, when running the wolf-sheep example with grass enabled, that the grass appears black and obscures the sheep. I believe I have fixed this in the corresponding server script.

I made the changes in my fork, but reading the contrib guidelines I've opened this issue first. I presume I should also create a new branch before making the pull request? I am sorry: I am still learning GitHub etiquette. Many thanks for your work.

janusd:
- Fix #474

| Before | After |
| ------ | ----------- |
| grafik   | grafik |

### DIFF
--- a/examples/wolf_sheep/wolf_sheep/server.py
+++ b/examples/wolf_sheep/wolf_sheep/server.py
@@ -28,9 +28,9 @@ def wolf_sheep_portrayal(agent):
 
     elif type(agent) is GrassPatch:
         if agent.fully_grown:
-            portrayal["Color"] = ["#00FF00", "#00CC00", "#009900"]
+            portrayal["Color"] = ["#00FF00"]
         else:
-            portrayal["Color"] = ["#84e184", "#adebad", "#d6f5d6"]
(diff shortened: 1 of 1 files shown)


jackiekazil:
So, this looks like a browser capability problem. This is what I see... 

screen shot 2018-03-01 at 9 00 33 pm

I believe it is a linear gradient that is causing the issue -- https://github.com/projectmesa/mesa/blob/master/mesa/visualization/templates/js/GridDraw.js#L151

So, the PR that you submitted is not quite the right fix. 

I am viewing in Version 64.0.3282.167 (Official Build) (64-bit) on Mac. 
Can you tell me which browser, version, and OS you are viewing the example on?
And have you tried another browser? Do you experience the same issue?

jackiekazil:
It looks like this is a version issue w/ the mesa that is available via pypi versus the examples and the code available in the repo. When we push the next version to pypi, this issue will be resolved. For more details see #474.

Issue #485, opened 2018-03-12
Batchrunner report creation error when values are one of dict/list/ndarray


ihopethiswillfi:
in batchrunner.py:
`ordered[param] = self.fixed_parameters[param]`

When values are a dict/list/np.ndarray, the above line throws the error:
ValueError: Length of values does not match length of index

This is because pandas interprets these datatypes as multiple values to fill a column with.

I used the following solution:
```
val = self.fixed_parameters[param]
if (type(val) == dict) or (type(val) == np.ndarray) or (type(val) == list):
    ordered[param] = str(val)
else:
    ordered[param] = val
```

It does the job for me, but it doesn't account for potential other datatypes that may have the same problem.  So not a very clean solution.  Any ideas how to handle this better?

ihopethiswillfi:
Fixed it and came up with a much better way in a single line of code :)

Will create a PR for it as soon as this other one is accepted:  https://github.com/projectmesa/mesa/pull/486

ihopethiswillfi:
Fix is part of https://github.com/projectmesa/mesa/pull/486
(one line of code, and better solution than I pointed out above as it catches all iterables)

PR #486, opened 2018-03-13
Batchrunner fixes: properly initialize models with correct parameters during subsequent runs.

ihopethiswillfi:
- Moved model initialization into the iterations loop to ensure that
models are also properly initialized when starting a new iteration.

- The deepcopy ensures that on model initialization the kwargs are always
the original ones specified by the user.
The current solution (before this code change) allows parameters to be
overwritten during a model run.  This causes subsequent model runs to
be initialized with the wrong parameters.  This happens for example when
one of the parameters is warehouse stock.  If the first model modifies the
amount of stock in the warehouse, then the second model run would be
initialized with that modified value.

### DIFF
--- a/mesa/batchrunner.py
+++ b/mesa/batchrunner.py
@@ -195,7 +195,12 @@ class BatchRunner:
         rest_cols = set(df.columns) - set(index_cols)
         ordered = df[index_cols + list(sorted(rest_cols))]
         ordered.sort_values(by='Run', inplace=True)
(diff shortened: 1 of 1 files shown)


ihopethiswillfi:
Included fix for https://github.com/projectmesa/mesa/issues/485

jackiekazil:
This all looks good to me. Ty for the contribution.

Issue #488, opened 2018-03-16
Update all requirements to most recent versions


jackiekazil:
Go this file and update the requirements one by one to the most recent versions and test them with Mesa. 

https://github.com/projectmesa/mesa/blob/master/requirements.txt 

- [ ] setuptools<34
- [ ] numpy
- [ ] pandas==0.17.1
- [ ] jupyter==1.0.0

- [ ] coverage==4.1
- [ ] nose==1.3.7
- [x] tornado
- [ ] flake8==2.5.2
- [ ] tqdm
- [x] networkx==2.1

jackiekazil:
Networkx was done here: https://github.com/projectmesa/mesa/commit/3380f6ced3386b6a93320a166e010dd4cbb45d5e

jackiekazil:
Tornado was/is addressed here: https://github.com/projectmesa/mesa/pull/497/commits/73f153018ded792fb1b5f0026f7bb8af734b599a

Corvince:
This is obsolete now!

PR #489, opened 2018-03-16
[BUGFIX] Updating version of tornado b/c old version breaks jupyter.

jackiekazil:
I ran into this issue https://github.com/jupyter/help/issues/272, because the version of tornado was old, so I updated it to the most recent version.

Also created this issue to put more attention into this: https://github.com/projectmesa/mesa/issues/488


dmasad:
Looks reasonable to me!

PR #492, opened 2018-03-28
Make batchrunner's "variable parameters" optional.

### LINKED ISSUE #466: BatchRunner's "variable parameters" is not strictly optional
I was trying to use `BatchRunner` to run multiple iterations of my model that had only fixed parameter sets. Initially, when reading the [docs](http://mesa.readthedocs.io/en/master/apis/batchrunner.html), I saw that the `variable_parameters` is specified as an optional keyword argument. However, if I do not pass in any variable parameters, then BatchRunner fails with the error:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-54-793336723b8d> in <module>()
      8                         iterations=10,
      9                         max_steps=12 * 20,
---> 10                         model_reporters={'Projects': stage_counts})
     11 batch_run.run_all()

~/anaconda/envs/prog_forecast/lib/python3.6/site-packages/mesa/batchrunner.py in __init__(self, model_cls, variable_parameters, fixed_parameters, iterations, max_steps, model_reporters, agent_reporters, display_progress)
     81         """
     82         self.model_cls = model_cls
---> 83         self.variable_parameters = self._process_parameters(variable_parameters)
     84         self.fixed_parameters = fixed_parameters or {}
     85         self._include_fixed = len(self.fixed_parameters.keys()) > 0

~/anaconda/envs/prog_forecast/lib/python3.6/site-packages/mesa/batchrunner.py in _process_parameters(self, params)
    101         params = copy.deepcopy(params)
    102         bad_names = []
--> 103         for name, values in params.items():
    104             if (isinstance(values, str) or
    105                     not hasattr(values, "__iter__")):

AttributeError: 'NoneType' object has no attribute 'items'
```

I felt confused when I first encountered this error, as I was expecting `BatchRunner` to gracefully handle the `variable_parameters=None` scenario. It thus appeared as a bug to me. 

If this is truly a bug, I think there's two different ways fixing this could go. If the intent is to make `variable_parameters` an optional keyword argument, then adding code to deal with its optionality would be good, e.g. some conditionals that deal with it. Otherwise, if the intent is to make `variable_parameters` a non-optional argument, it should not have a default value of `None`. 

May I ask which of these design choices is the intended one? If it's the latter, I'm happy to put in a PR to make it non-optional. If it's the former, then I'd like to ask for guidance on putting in a PR, as there's probably a few places I'll need to add in conditionals that I wouldn't be fully aware of.

ihopethiswillfi:
Fixes #466

### DIFF
--- a/mesa/batchrunner.py
+++ b/mesa/batchrunner.py
@@ -134,7 +134,6 @@ class BatchRunner:
                 for _ in range(self.iterations):
                     self.run_iteration(kwargs, param_values, run_count)
                     pbar.update()
-                   
 
     def run_iteration(self, kwargs, param_values, run_count):
         kwargscopy = copy.deepcopy(kwargs)
@@ -154,8 +153,6 @@ class BatchRunner:
             for agent_id, reports in agent_vars.items():
                 agent_key = model_key + (agent_id,)
                 self.agent_vars[agent_key] = reports
-
-
 
     def run_model(self, model):
         """ Run a model object to completion, or until reaching max steps.
--- a/mesa/batchrunner.py
+++ b/mesa/batchrunner.py
@@ -98,9 +98,6 @@ class BatchRunner:
         self.display_progress = display_progress
 
     def _process_parameters(self, params):
-        if params is None:
-             return None
-
         params = copy.deepcopy(params)
         bad_names = []
         for name, values in params.items():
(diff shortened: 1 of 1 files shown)


ihopethiswillfi:
Can anyone help write a test?  I looked at it for a while but wasn't sure how to best approach it.  I haven't written tests before :)

I was thinking to just copy the function at the bottom and set
`self.variable_params = {}`

Also I don't understand why the function `test_model_with_fixed_parameters_as_kwargs` doesn't seem to cover this.  It seems to launch a batchrunner WITH fixed_params and WITHOUT variable_params , which is what we need to test.

mesa/batchrunner.py
```diff
@@ -45,7 +45,7 @@ class BatchRunner:
     entire DataCollector object.
 
     """
-    def __init__(self, model_cls, variable_parameters=None,
+    def __init__(self, model_cls, variable_parameters={},
```
dmasad: It's generally not a good practice to use a [mutable value (e.g. `{}`) as a default argument](https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html). In this case I don't _think_ anything harmful would happen, but I'd suggest leaving `None` as the default value just in case, and for general Pythonic-ness.

dmasad:
It's been a while since I've looked at the test code, but it looks like in `test_model_with_fixed_parameters_as_kwargs`, the `variable_parameters` are still the default `self.variable_params` values set in the test default. To test not having any explicitly, you'd need to set `self.variable_params = None`.

jackiekazil:
@dmasad -- can you remind me / us where this left off? other than fixing conflicts? (I can't remember what as the explicit ask on this.)

dmasad:
Turns out this is slightly more complicated than expected because `BatchRunnerMP` also needs to be updated to work with this. I'm working on it.

dmasad:
Closing; superseded by #596

PR #500, opened 2018-04-02
[WIP] Fixes flockers b/c of ghost git files of diff casing was causing issues

jackiekazil:
Flockers was changed to flockers and that was causing issues. That needs to be fixed in the git tracking. The easiest way to do this was to delete it and then recreate flockers, which is what I did. You will see below how all the casing was mixed and git was confused. Hopefully, there should be no more issues with this merge. 

The first commit destroys flockers. The second commit rebuilds it.

screen shot 2018-04-01 at 10 33 55 pm


dmasad:
On my end, this just removes the flockers example entirely (with any capitalization)

jackiekazil:
I changed the status of this in WIP... I need to spend more time with this. 
I also removed the "easy_task" label, because it doesn't feel easy to me. 😡

djmitche:
I tried reproducing this on a case-insensitive-and-preserving filesystem (but one on linux, not HFS+) and git did OK with it.  I'm using Git 2.17.0.  This [StackOverflow post](https://stackoverflow.com/a/24979063) suggests that Git fixed this issue in 2.0.1.  @dmasad is it possible you're using an old version of Git?

At any rate, I don't think there's much to be done in the github repo to fix this, although it could be avoided:
 * rename Flockers to a different word (perhaps `boids-flocker`?)
 * put the example in a different directory

What @arowla has suggested is an option, but until everyone has updated past the problematic merge, and all PRs are rebased after it, repos will keep flipping between "flockers" and "Flockers" and need this fix applied every time.

It seems like the best option is to rename it.  I'll make a PR.

jackiekazil:
Closing this in lieu of other PRs
giphy

PR #506, opened 2018-04-24
[Feature Implemented]: Multiprocessing BatchRunner with pathos

Maarten-vd-Sande:
Continuation of https://github.com/projectmesa/mesa/issues/455 and https://github.com/projectmesa/mesa/pull/456. Decided to 'close and reopen' as to not clutter master with needless commits.

As suggested, it now tries to import pathos at the top, if importing failed, the init of BatchRunnerMP raises a custom MPSupport error.

Let me know if it does not work as expected/desired :smiley:

### DIFF
--- a/mesa/batchrunner.py
+++ b/mesa/batchrunner.py
@@ -11,6 +11,12 @@ import copy
 from itertools import product, count
 import pandas as pd
 from tqdm import tqdm
+try:
+    from pathos.multiprocessing import ProcessPool
+except ImportError:
+    pathos_support = False
+else:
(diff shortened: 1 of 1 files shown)


dmasad:
I'm having some trouble installing pathos on my machine to test it. I'll try to get that resolved over the weekend, but in the mean time, if anyone else wants to test this PR (or better yet, write a test for it) and verify that it works, go for it!

dmasad:
Looks good! I have tests that run locally, I'll add those in a subsequent PR.

PR #510, opened 2018-05-10
Schedule.agent.dict

dmasad:
Implement tracking the agents in the scheduler via `OrderedDict` as implemented in #386, add `agents` property to preserve backwards-compatible API, and update examples to conform.


dmasad:
Flockers are still (again?) missing from this. 
¯\_(ツ)_/¯

PR #511, opened 2018-05-12
Fix to collapse and decapitalize duplicate Flockers folders (re: #500)

dgomer:
It looks like some of the work on the Flockers folder was done on a case-insensitive file system (i.e. MacOS), while some other work was done on a case-sensitive file system (i.e. Linux). I believe this should fix the differing issues across file systems.


dgomer:
I am very new to this so if I've done anything incorrectly please let me know!

jackiekazil:
This deletes all the files... we will still want to keep the lowercase version. 
screen shot 2018-05-11 at 10 44 49 pm
screen shot 2018-05-11 at 10 45 21 pm

dgomer:
That's very interesting. I think this is definitely a case sensitivity issue though. 

This is a fresh clone of the repo on MacOS (case insensitive but preserving):
screen shot 2018-05-12 at 7 56 53 am

This is a fresh clone of the repo on Ubuntu 16.04 (case sensitive):
screen shot 2018-05-12 at 7 56 37 am

When I created the PR (on the Ubuntu system) and then merged it on my local MacOS copy, it kept all the files though. Do you mind sharing what OS your screen shots are from?

Issue #200, opened 2016-04-01
Some issues with the Tutorial: i.e. parameter to get a histogram of agent wealth


felixhelix:
Found some things in the tutorial:
When you introduced numpy to visualize the number of agents, the code reads
`cell_wealth = len(cell_content)`
but it isn't the wealth, but the number of agents that is visualized. This confused me a bit.

Under the heading "Collecting Data":
I guess instead of 
`end_wealth = agent_wealth.xs(19, level="Step")["Wealth"]`
it should be 
`end_wealth = agent_wealth.xs(99, level="Step")["Wealth"]`
since the model ran for 100 time steps?

Also in the code example for "Setting up the model" it says:
`self.running = True`
Which is later omitted, but not explained?

But most of all: Thanks for this great abm framework!
Your's Felix

dmasad:
Thanks! Fixed.

PR #518, opened 2018-05-14
rename 'flockers' to 'boid-flockers'

djmitche:
This avoids issues with case-preserving filesystems (Windows, MacOS)
where `flockers` and `Flockers` refer to the same directory, causing
issues when changing Git working copies.

This is a replacement for #200.


jackiekazil:
LGTM

Issue #520, opened 2018-05-14
Create mesa commands


jackiekazil:
Commands to create.... 
```
mesa startproject $projectname 
mesa run
```
^^ This command framework will lead to additional commands, but don't want to overload things at this moment. This is in the spirit of Django commands. 

Later you will be able to run examples by commands, which then leads to a fix for #134

jackiekazil:
Mesa run is complete.... but ```mesa startproject $projectname``` is still pending.

PR #522, opened 2018-05-14
Use click and add `mesa run`

djmitche:
This runs the project in the same way a user typically would, by
changing to the project directory (so that resources can be located
using relative paths).  However, rather than executing the `run.py`
directly, it assumes that importing `run` will produce a module with a
`server` attribute.  For the moment, it calls `server.launch()` but it
could be modified to call other methods or pass parameters to that
server, such as to run "headless"

This addresses #520, although it does not add the start-project option.


djmitche:
OK, if that goes green I think it's ready.

PR #525, opened 2018-05-14
Removing 'F'lockers.

jackiekazil:
This hopefully fixes the "F"lockers / "f"lockers issue. 
Related: #500, #525


djmitche:
Yep, somehow we ended up with both `flockers` and `Flockers`.  I renamed the first, but didn't notice the second.

Issue #523, opened 2018-05-14
align required dependencies between setup.py and requirements.txt


imbilltucker:
setup.py and requirements.txt don't install the same versions of things.  An error occurs when trying to run examples using the setup.py dependencies.

To reproduce:
clone mesa and switch to the directory
* install dependencies from setup.py:   `pip install -e .`
* save the list of install dependencies:  `pip freeze >> setup_req.txt`
* compare them to the ones specified:  `diff setup_req.txt requirements.txt`

to see an error, because of this discrepancy:
   * 'cd color_patches`
   * `python run.py`

You should see an IOLoop error because of the change from Tornado 4.5.x to Tornado 5:
```
/home/bill/.virtualenvs/mesa/bin/python3.6 /home/bill/programming/imbilltucker/mesa/examples/color_patches/run.py
Traceback (most recent call last):
  File "/home/bill/programming/imbilltucker/mesa/examples/color_patches/run.py", line 3, in <module>
    server.launch()
  File "/home/bill/programming/imbilltucker/mesa/mesa/visualization/ModularVisualization.py", line 317, in launch
    startLoop = not tornado.ioloop.IOLoop.initialized()
AttributeError: type object 'IOLoop' has no attribute 'initialized'```

imbilltucker:
Here's another issue I found:  
ImportError while importing test module '/home/bill/programming/imbilltucker/mesa/tests/test_space.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
test_space.py:3: in <module>
    import networkx as nx
E   ModuleNotFoundError: No module named 'networkx'

networkx needs to be referenced in setup.py

Corvince:
Solved with #535

PR #526, opened 2018-05-14
add version numbers for tornado in setup.py

imbilltucker:
this change to setup.py specifies the version of tornado should be less that 5. 

This keeps an IOLoop error from occurring in the color_patches examples with the tornado 5 changes.


jackiekazil:
fixes #523

PR #527, opened 2018-05-14
add versioning to tornado in setup.py

imbilltucker:
this sets the versioning for the tornado library in issue #523 to fix the `pip install -e` setup from an initial checkout.


jackiekazil:
LGTM

PR #528, opened 2018-05-14
Issue 523

imbilltucker:
add networkx to the dependency specified in setup.py for #523 to allow all the tests to run without error when installing with the `pip install -e .` command.


jackiekazil:
This is superseded by #535.

PR #529, opened 2018-05-14
Run examples as part of the tests

### LINKED ISSUE #509: Add examples to Travis
The formal tests test the basic functionality of each component individually, but not how they are used together in practice. The examples do use these components together, and additionally provide an example of the API. 

At the moment, verifying that changes to the core code don't break the examples is done by running the examples by hand. It would be helpful to automate this, and have Travis call a script that sets up and runs each example. Unlike the tests themselves, this wouldn't need to check the actual model output, only that the examples can be launched and run without throwing errors.

djmitche:
This fixes #509


djmitche:
As requested, this assumes each example has a specific form, with the model name, package name, and example name all the same.  And it makes the examples all match.

jackiekazil:
@dmasad can you give a status update on this? (not sure if you pushed forward OR if someone else can help push along.)

dmasad:
This looks good, but there are a few issues with how the models were modified. 

- in `boid-flockers`, `server.py` references `BoidServer` where it should be `BoidFlockers`; there's also an issue with the visualization that I'm chasing down.
- Similarly, in `epstein_civil_violence/server.py` there's a reference to the wrong model name.

dmasad:
Fixed minor issues and pushed in #564

PR #533, opened 2018-05-15
Fix deprecation warning in lifespan tests. Closes #531

### LINKED ISSUE #531: pandas FutureWarning
Pandas' ever changing api is throwing a future warning for a single test in test_lifespan.py:

```
mesa/tests/test_lifespan.py:85: FutureWarning: using a dict on a Series for aggregation
is deprecated and will be removed in a future version
  lambda x: len(x)})
```

Extremely minor issue and the fix is simple. But if we're running tests for the sprints, might be worth just clearing out?

Thanks!

pvomelveny:
Take 2

Minor change to the test to remove the deprecated functionality (using dicts in calling .agg). Should be stable regardless what pandas ends up doing with the .agg.

### DIFF
--- a/tests/test_lifespan.py
+++ b/tests/test_lifespan.py
@@ -81,9 +81,8 @@ class TestAgentLifespan(unittest.TestCase):
         assert self.df.steps.max() == 1
 
     def test_agent_lifetime(self):
-        lifetimes = self.df.groupby(["AgentID"]).Step.agg({"Step":
-                                                           lambda x: len(x)})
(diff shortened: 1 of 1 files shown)


jackiekazil:
I hate turning down a PR. :-(  But I am going to close this in leiu of #535, so it doesn't cause conflicts, because that PR has another element as well.

pvomelveny:
No problem at all! It was 30 seconds of work tops.

PR #544, opened 2018-05-24
Fixes #543

### LINKED ISSUE #543: User Settable Parameters fail to work for non-string datatype
As per title. Specifically, when using the `choice` option, passing in non-string choices and selecting these choices will result in a 
`Selected choice value not in available choices, selected first choice from 'choices' list` error, which will end up defaulting to the first value. (I was expecting it to throw an exception? Doesn't really seem right for us to allow choices that somehow aren't in the choices list...)
After poking around with the javascript, it seems like the error is caused by the jquery selector returning a string, as opposed to the original data type.

Shermjj:
Fixes #543 

When attaching the listeners, instead of using the jquery selector to obtain the value (which will necessarily return a string representation of the data), I propose obtaining the value directly from the obj variable. This will allow non-string datatypes to be returned as the value. 

I've mainly looked at using number data types (int and floats) as well as arrays. However, it seems like as long as there is a corresponding javascript data type, the fix should work. So dictionaries and boolean should work, but tuples will not.   

I'm not too sure how we would implement a test for this though. It would be good if someone could review this as well, I'm still new to js.

### DIFF
--- a/mesa/visualization/templates/js/runcontrol.js
+++ b/mesa/visualization/templates/js/runcontrol.js
@@ -156,9 +156,9 @@ var initGUI = function() {
 
         // Finally render the dropdown and activate choice listeners
         sidebar.append(template.join(''));
-        choiceIdentifiers.forEach(function (id) {
+        choiceIdentifiers.forEach(function (id,idx) {
             $('#' + id).on('click', function () {
-                var value = $(this).text();
+                var value = obj.choices[idx];
                 $('#' + domID).html(value + ' ' + span);
                 onSubmitCallback(param, value);
             });


dmasad:
This looks good to me, but I'd like someone who understands Javascript better (@TaylorMutch ?) to take a look and double check.

I'm also not sure what the best practice is for writing tests for Javascript interface components -- that's probably a good thing to have too.

stefanv:
This seems helpful (more explicit), but the same pattern occurs elsewhere in the file.  Want to fix it in those locations as well?

jackiekazil:
@Shermjj ... a few of us are having issues trying to see what this chances. 
Can you add an example that shows this broken in the current person and fixed in the PR? 
If you don't have an explicit example to add to the examples folder or a way that you can think to extend an example, I would also accept a toy example (in another repo) that demos the problem even if the toy data means nothing relevant.

jackiekazil:
@stefanv -- help me understand what you mean by other locations? Can you provide line numbers?

Shermjj:
@jackiekazil 
[Here's](https://gist.github.com/Shermjj/d3653dec3d0d97cdb31c1598df68a732) a gist with a test example for you to try out 

This is the output:
example
As you can see, using non-string datatypes results in the value being string coerced. This also results in the `Selected choice value not in available choices, selected first choice from 'choices' list` line being printed, because now the value doesn't match with the initial `choices` parameter provided. 
(Dictionaries are a little wonky, probably because JS doesn't allow numeric type for object keys?)

dmasad:
Merged, but at least on my end this doesn't actually seem to solve the numbers-coerced-to-strings behavior.

PR #560, opened 2018-07-14
Add a github issue template.

werwty:
When a new issue is opened, request that the reporter leave
a description, and a reproducer for the issue.


jackiekazil:
LGTM

PR #561, opened 2018-07-14
Changes nose to pytest

### LINKED ISSUE #534: Update tests to use pytest, not nose
Update tests to use pytest, not nose. nose is not maintained anymore.

jackiekazil:
Fixes #534. Assuming that Travis passes, I think this is all that needs to be done to change to pytest.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ script:
   # * E123 - indentation on data structures
   # * W504 - line break after binary operator
   - flake8 . --ignore=F403,E501,E123,E128,W504 --exclude=docs,build
-  - nosetests --with-coverage --cover-package=mesa
+  - py.test --cov=mesa tests/
   # - ./tests/test_end_to_end_viz.sh  # needs to be investigated on why this took forever
 after_success:
   - coveralls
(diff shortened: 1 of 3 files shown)


Corvince:
Seems quite fine, but I found some oddities:

1) When I run `py.test` locally it gives me this error
```
ImportError while importing test module 'C:\Users\x\mesa\examples\wolf_sheep\wolf_sheep\test_random_walk.py'.
Traceback:
examples\wolf_sheep\wolf_sheep\test_random_walk.py:13: in <module>
    from random_walk import RandomWalker
E   ModuleNotFoundError: No module named 'random_walk'
```
Don't know why it fails and the error vanishes if I test with `py.test --cov=mesa tests/` ??? It also works fine on Travis so I guess its a problem with my own set-up, but maybe someone has an idea about this.

2) In the Travis log it reports `Coverage.py warning: No data was collected. (no-data-collected)` It works locally on my machine.

3) Git(hub) questions: The testing branch is a bit outdated, but "Files changed" tab only compares to the testing branch and not the master branch? Can somebody explain? At this point I wouldn't even know what would happen if this were to be merged.

jackiekazil:
the most recent update is just to fix the branch with updates from master.

jackiekazil:
@dmasad this is done and ready for merge now.

PR #563, opened 2018-07-14
Update and clean up cookiecutter layout 

werwty:


mesa/visualization/modules/ChartVisualization.py
```diff
@@ -74,6 +74,9 @@ def render(self, model):
 
         for s in self.series:
             name = s["Label"]
-            val = data_collector.model_vars[name][-1]  # Latest value
+            try:
```
werwty: This was changed in https://github.com/projectmesa/mesa/commit/e52773b7231d66f6100d3700826344e40a1b071b#diff-c6a62ed767d052044af87fd60b6240acL77 

Unfortunately if the DataCollector doesn't see any data, this breaks all visualization. Instead of that happening, it's saner to catch and set to 0. (the right solution should probably be to update the front end with some message that data isn't being collected, but that is a lot more complex)

jackiekazil:
@strixcuriosus thank you for your original work on this. 
@werwty for bringing it home. 

This is really nice. <3

PR #565, opened 2018-07-15
Updating setup to move requirements to setup.py.

jackiekazil:
The purpose of this change is to get rid of the requirements file and have developers use the setup.py for local installation. By doing this they create an editable version as opposed to just adding the package to the python path, which means commands then work -- ie `mesa run`.


jackiekazil:
@werwty links are fixed. This is ready for a second pair of eyes.

Issue #134, opened 2015-07-03
Move Mesa examples to another repo with sub classing to this repo? 


jackiekazil:
Move Mesa examples to another repo with sub classing to this repo? 
Let's explore this more and figure out if this is something pattern that will work with subclassing git repo inside a git repo. 
#### Things to consider
- How pushing to pypi works
- Make sure if someone changes any of the core that the examples still run
  - The tests are based on some of our examples

dmasad:
I like the way [tornado does it](https://github.com/tornadoweb/tornado), where the examples are part of the repo but not the manifest that gets packaged up for pypi.

jackiekazil:
Here is what I was talking about: https://github.com/projectmesa/mesa/tree/134.subrepo
The tests will need to be updated a little bit, but before I do that, I want to know that we are interested in moving in this direction. 
@projectmesa/maintainers --- move forward or no?

jackiekazil:
When I was getting folks setup to run models in Mesa, I found that installing Mesa AND then running the examples was confusing, because the examples were inside Mesa. 

This repo exists and now it is up to date: https://github.com/projectmesa/mesa-examples

Question -- what is the purpose for us to have it in this repo, which couldn't be solved by having 2 repos on our machines / remote machine for testing? 
Why does examples NEED to be in Mesa? 
Discuss.
- Devil's advocate.

jackiekazil:
+ @projectmesa/maintainers

njvrzm:
The only strong argument I can think of for keeping the examples repo inside the main mesa repo is versioning: we're still in a phase where we're making breaking API changes pretty regularly, and if the examples repo is separate, even if we maintain parallel versioning, it's possible users will end up with mismatched versions of the two projects. If we do go with separate repos, we should handle this case gracefully, detect the mismatch and issue a clear error so people know what to do. This should be pretty straightforward.

This gets me thinking about having mesa be a command, not just a library. The [Click](https://github.com/pallets/click) library has some cool ideas around this. If we did that, after someone installed mesa via pip, they could do something like `mesa fetch_examples` to checkout the examples repo, then `mesa launch GameOfLife` to run that example. `mesa launch` by itself could look for a server.py in the current directory or its immediate subdirectories. We could integrate batch running and benchmarks this way too. We could also follow the example of virtualenv and include a `mesa init <project_name>` command that would set up a skeletal module in the standard format. This could make it easier for people to get started and to contribute new examples.

[...]

PR #568, opened 2018-07-31
Add mesa examples click command

Corvince:
The command `mesa examples` allows to either copy or run mesa example models.

`mesa examples` lists all available examples
`mesa examples --copy Schelling` copys the Schelling example to the current directory (copy is the default)
`mesa examples --run Schelling` runs the Schelling example directly in the mesa install dir (could this be a problem?) 
`mesa examples --copyall` copys the whole examples folder into the current directory

I think this makes the mesa examples MUCH more accesible for new users, as discussed in #134 

For the implemenation I used the importlib_resources library, also as suggested in #134. 
However, here is the downside: For importlib_resources to find the examples I had to move the examples to mesa/examples. I think this is less clean, but also necessary for the examples to actually get installed with mesa(?). Would it be possible to install a mesa_examples module with mesa without a different repo? I am not sure how to proceed with this. Any ideas about this? 
[This is also the reason Travis fails, as the nose tests don't find the examples folder]


jackiekazil:
This PR is blocked on us making a final decision on how we want to proceed with #134.  After a discussion at tonight's meeting, I am going to research patterns of other libraries and report back.

Corvince:
Outdated beyond repair

PR #572, opened 2018-08-11
Adding model random number generator with __new__

dmasad:
Gives each model its own random-number generator with a seed set at instantiation. This is done by implementing a custom `Model.__new__` method, which adds a seed and a `random.Random` object to the model object when it is created.


dmasad:
This is failing the tests due to `mock_shuffle.call_count` being off, but I'm not clear what's going on there, so I'm not sure how to rewrite the test to match the new implementation.

dmasad:
TODO: Once the tests are resolved, the examples should be rewritten to use `self.model.random` where appropriate prior to merging.

dmasad:
Tests fixed, and all models updated to use the model object `random` instead of the global random. This is ready to be reviewed and merged.

Corvince:
Great job so far.

Minor: The shape_example server.py file still uses random (can be removed completely)

Major: The cookiecutter model template still uses random

Challenge: The `space.py` module still uses random for the [`find_empty()` function](https://github.com/projectmesa/mesa/blob/3df184bbf0026d2441f1edbe550a216d716bfede/mesa/space.py#L345). Since the grid does not "know" the model I don't see a straight forward solution.

dmasad:
@Corvince Thanks for looking it over! I think I've fixed all those points. The way I dealt with the Space issues was to remove the random empty cell choice, and use the agent's own random property whenever placing an agent at random.

Corvince:
Great, I'll take a look deeper at it once again in the next few days. 

I think removing the find empty function should be unproblematic, because it's easy enough for modelers to implement. However, I am worried about breaking existing models. I think it's better to raise a depreciation warning first. It could include a notification that models will not be reproducible.

Corvince:
Looks good! I only added some small changes. 

But, as mentioned above, I am worried about simply removing "find_empty()". Since it kind of works as it should, I wouldn't risk breaking models. Just warn people that it is depreciated and not suited for reproducible models. 

Otherwise, I think its ready to merge!

dmasad:
Added `move_to_empty` back with a warning.

Issue #575, opened 2018-08-23
More efficient datacollection


Corvince:
As discussed on the mailing list the current implementation of DataCollector.collect can be improved in terms of speed and memory usage. I spend too much time in the last couple of days to profile the function and try to come up with a better one.  For model reporters performance is not an issue since they are only evaluated once per step and not for every agent. But agent reporters could be improved. Here is a comparison between no data collection, the current data collection and my proposed function:

![dc_compare](https://user-images.githubusercontent.com/13568919/44519634-dcefa700-a6cd-11e8-9923-c3fac6541fa7.png)

As you can see, the speed overhead is much smaller now, especially for few (<1000) agents. 

Here is the downside, tho: The performance gains are only substantial if the agent_reporters only consist of agent attributes given as string values (e.g. "x": "x"). In the case of custom functions and/or a mix, the performance (and code) will be almost the same. 

However, judging by the examples, custom functions are rarely used for agent reporters, so most models can benefit from the new code. 

This might also be related to #574, since (at least collecting) the attribute state now has minimal overhead.

Corvince:
New approach online: Instead of calculating reporter values directly we can use `map()` to "attach" functions to agents. This returns a generator that uses lazy evaluation to calculate its values. 

This means the downside for non string based agents mentioned above is circumvented and we have the result
![no_overhead](https://user-images.githubusercontent.com/13568919/44582374-c155d080-a7a1-11e8-8c8b-9695bea35062.png)
for ANY reporter function.

However, the work is only shifted to the `get_agent_vars_dataframe()` method. If we evaluate this function at the end of a model run, the total runtime will be the same as in the updated data collection above (again, with the potential fast path for attribute reporters). 

But if we are not interested in the collected data (for example for interactive runs) the collect method no longer slows down model runs. 

Two things to discuss:
    1) In the current implementation accessing the generator functions "empties" them, meaning we can only call `get_agent_vars_dataframe()` once. This is easy to prevent, but it could be a desirable side-effect to free up memory (e.g. dumping the data every step into a database). Don't know which version to prefer although I guess the emptying will be unexpected for most users (and so should be prevented)
    2) We could use the same approach for model reporters.

jackiekazil:
@Corvince Ty! This is great! <3 

Re: Discussion -- 
1. I am thinking hold it, but build emptying as an option? Because memory could be an issue.
2. Sounds good.... maybe separate PR? 

@dmasad do you have any feels about this?

PR #576, opened 2018-08-23
Faster agent attribute collection

Corvince:
Addresses #575 

This implementation adds a fast-track for collecting agent attributes if all reporters are string-based (no custom or lambda functions). If no string-based reporters are present or for a mixture of reporters it defaults to a only slightly improved version of the current implementation.

I also changed the structure of the class variable agent_vars, to store less data, it specifically stores the agent.unique_id only once and not for every reporter. This breaks code that relies on that variable, although none of the examples do. 

One idea to encourage the use of string-based reporters for agent reporters would be to include a warning if an agent reporter is not string-based and tell users that they could benefit from switching to string-based reporters (or even that support for non string-based reporters will be dropped in a much later version of mesa). 

Hope you like it

### DIFF
--- a/mesa/datacollection.py
+++ b/mesa/datacollection.py
@@ -136,6 +136,21 @@ class DataCollector:
         new_table = {column: [] for column in table_columns}
         self.tables[table_name] = new_table
 
+    def _record_agents(self, model):
+        """ Record agents data in a mapping of functions and agents. """
+        rep_funcs = self.agent_reporters.values()
+        if all([rep.__name__ == 'attr_collector' for rep in rep_funcs]):
+            prefix = ['model.schedule.steps', 'unique_id']
+            attributes = [func.attribute_name for func in rep_funcs]
+            get_reports = attrgetter(*prefix + attributes)
+        else:
+            def get_reports(agent):
+                prefix = (agent.model.schedule.steps, agent.unique_id)
+                reports = tuple(rep(agent) for rep in rep_funcs)
+                return prefix + reports
+        agent_records = map(get_reports, model.schedule.agents)
+        return agent_records
+
     def collect(self, model):
         """ Collect all the data for the given model object. """
         if self.model_reporters:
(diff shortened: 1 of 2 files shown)


Corvince:
Okay, forget this implementation! 

I found a much better way with virtually no overhead!! 

Instead of evaluating the reporter functions directly we can use the map function to create a generator and only calculate the values once we build up the dataframe. 

Something like:
`agent_records = map(reportgetter, model.schedule.agents)`

This will also work with custom reporter functions and not just string based reporters. I think I could still speed up the creation of the dataframe for pure string reporters, but I think this will be less critical.

Corvince:
Oh gosh I wish I could trust my eyes more... 

I somehow thought that map would auto-magically track the agents current state at the cost of increased memory usage. I quickly checked it and everything seemed fine, but I must have looked at the wrong data or something. 

Unfortunately (but logically), no auto-magic is happening. With the nature of lazy evaluation once you create the DataFrame, the values for all steps are actually only from the current state (the last step). 

Guess I was too excited 😢 I will look into this further and probably revert to the previous implementation (which speeds things up, but still with some overhead). 

Until then, please don't merge anything just yet.

Corvince:
Ok, new implementation online, again. And since I am spamming this thread already, here is some explanation how the current implementation is faster (for those interested).

In the original implementation we had to access each agent 2*reports times (because each reporter would also store the agents unique id). For 2000 agents and 5 reporters this would result in 20000 memory accesses and operations). 
In my implementation we only access each agent once and collect all its attributes **using only built-in functions** (i.e. only 2000 memory accesses). That is why the speed-up only works for attribute reporters. 
Note, however, that I know nothing about the internal process/optimization of python. The difference in memory accesses is probably smaller, but the main speed-up is from using a built-in function. 

Now in the current implementation I have "outsourced" the attribute collection into a function called `_record_agents` which maps the attribute collection to each agent. This mapping is very fast. The "slow" part of collecting data is evaluating the mapping and putting it in a list. This was missing in the previous (extremely fast) implementation and is now done in the collect function. The advantage of this approach is that we can also write other functions that, for example, write the mapping directly to a database or file without first constructing a list and then iterating over that list. 

NB: collect now stores agent records in a dictionary with model steps as the keys. That means it would also be trivial to add a "step" keyword to the `get_agent_vars_dataframe()` to return only the dataframe for a single step or a range of steps (the latter not so trivial, notation-wise)

PR #582, opened 2018-09-14
Use partial for attribute collection

### LINKED ISSUE #175: DataCollector and lambdas
Currently the default DataCollector API encourages the use of lambdas to define the model- and agent reporters. This is nice and readable, and doesn't require subclassing the DataCollector for each model. However, it turns out that lambdas are not pickle-able. This means that model objects using DataCollectors can't easily be serialized, which creates problems for checkpointing, model storage, and parallelization (since [ipyparallel](http://ipyparallel.readthedocs.org/en/latest/) uses pickle to send objects between engines.

One solution is just clarifying in the documentation that code needs to be written slightly differently for these cases. Another is figuring out another way to do reporters in DataCollector.

Corvince:
Small change that makes the DataCollector pickleable by using `functools.partial` instead of a local function to define the attribute collector.

This makes whole mesa models pickleable and fixes #175 (and it would greatly benefit mesa-geo if merged).

Oh and it now defaults to None if an object doesn't have the attribute, which makes multi-agent data-collection easier.

### DIFF
--- a/mesa/datacollection.py
+++ b/mesa/datacollection.py
@@ -36,8 +36,9 @@ The default DataCollector here makes several assumptions:
     * For collecting agent-level variables, agents must have a unique_id
 
 """
-from collections import defaultdict
 from functools import partial
+import itertools
+from operator import attrgetter
 import pandas as pd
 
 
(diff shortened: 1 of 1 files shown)


Corvince:
Think I messed up the rebase, better to merge #589

Issue #583, opened 2018-09-14
[Feature request] remove agents during Scheduler.step()


Maarten-vd-Sande:
**What's the problem this feature will solve?**
E.g. the step of the RandomActivation:
```
def step(self):
 """ Execute the step of all the agents, one at a time. """
    agent_keys = list(self._agents.keys())
    for agent_key in agent_keys:
        self._agents[agent_key].step()
    self.steps += 1
    self.time += 1
```
If lets say agent 1 'kills' agent 2, and this agent gets removed from _agents, we get the error that self._agents does not contain agent 2. This forces us to keep track of agents whether they are alive or not.

**Describe the solution you'd like**
Either use a generator to generate the keys, or do check in-place if the agent still exists. I would propose using a generator, something like this:
```
def mesa_keys(agents, shuffled=False):
    agent_keys = list(agents)
    if shuffled:
        random.shuffle(agent_keys)

    for key in agent_keys:
        if key in agents:
            yield key
```
Now you can simply replace the 
```
agent_keys = list(self._agents.keys())
for agent_key in agent_keys:
```
with
```
for agent_key in mesa_keys(agents, shuffled=False):
```

Would you guys be interested in something like this?

Corvince:
Good spot, that should definitely be possible! 

I think a PR would be very welcome, but a small observation: The `agents` argument actually requires the agent_keys, right? Or, I would write inside `mesa_keys`:

    agent_keys = list(agents.keys())

 And mesa_keys might not be the best name, but I couldn't come up with something better ;)

Oh and maybe you could also modify to yield the agents directly and not just the keys? I don't know why the current implementation doesn't do this already, since we just want to call the agents' `step` function, right?

Corvince:
Thanks!

Re 1) Ah, yes of course! I was confused because normally the 'agents' variable refers to a list of actual agents. Sorry, but for the sake of clarity please use `.keys()`

And I can't think of a logical assumption that agents should still be activated even though they are killed/removed from the scheduler. Of course, this shouldn't happen in the case of a simultaneous activation. 
In all other cases I would leave it to the modeler to create a list of agents that will be removed and only remove them after `schedule.step()` is called. That should be trivial to implement. 

Otherwise, I really wouldn't expect agents to do anything after they are killed/removed from the scheduler. But I agree that someone else should also have a look into this.

PR #584, opened 2018-09-21
agents can be removed and added during Scheduler.step()

Maarten-vd-Sande:
See [issue 584](https://github.com/projectmesa/mesa/issues/583) for previous discussion.

It does what the title says, however we were wondering if maybe some assumptions are now broken?


Maarten-vd-Sande:
Should I worry about the Travis CI build failing? I seems to be style-related?

Maarten-vd-Sande:
It now fails because the actual model.random is not (yet) implemented. I didn't see your comment about renaming it to iter_agents. I'll change it, but I think maybe buffer is actually more true.
I moved it to the BaseScheduler, because then I did not have to reference to the model and the agents, seemed better to me. Idk if you like it.

The function is a generator function, I made a small example of how it works, and that the current implementation works in our example:
```
def generator(agents):
    agent_keys = list(agents.keys())
    for key in agent_keys:
        if key in agents:
            yield agents[key]

agents = {1:'Maarten', 2:'Corvince', 3:'Dmasad'}

for agent in generator(agents):
    print(agent)
    if agent == 'Maarten':
        agents[4] = 'JackieKazil'
    if agent == 'Corvince':
        agents.pop(3, None)

>>> Maarten
>>> Corvince
```

Maarten-vd-Sande:
Or did you mean that the new agents are not stepped in the timestep they are added? That is true!

Corvince:
Hey there, thanks for the update. Don't worry about the name, that's fine. 

And yeah I meant the latter, but I was just misinterpreting your docstring. 
I read it like "...allows to remove agents (new functionality!) and add agents (new functionality!)". But it just states that we can remove and add agents and the latter isn't new, but of course correct. 

From a theoretical viewpoint it is quite interesting that we want to remove agents immediately, but add them only to the next time step. But yes, that is pretty much what you would expect. 

And with this implementation, if one wants to remove them in the next step you can use simultaneous activation and if you immediately want to step them you can just call their step() function outside the scheduler. 

And of course don't worry about the travis build failing, I will only merge this after #572 is merged.

PR #585, opened 2018-09-26
[WIP] Side by side visualization of multiple model runs

### LINKED ISSUE #573: Display different runs side by side
In Mesa (as in most ABM tools) the interactive visualization shows one model run at a time. Exploring different parameter values requires stopping the model, changing the value, and restarting. However, one way a model can illustrate the effect of a parameter is by showing multiple runs side by side, each with different values. Mesa could add a lot of value by making it easy to set up, run, and view side-by-side model instantiations (possibly with the same random seed) with different parameter values.

An example of what it could look like is here:
https://imgur.com/a/8M7q8

Corvince:
An approach to resolve #573 

Test it by running the Schelling example


Corvince:
@TaylorMutch Maybe you can help with this one? Especially with the user-settable parameters. So far I only managed to create a slider for each simulation, but completely with title, range, etc. I would like to only have N sliders, but with the same title, etc. (but different color for each simulation). `onSubmitCallback` should now return the simulation number for which the parameters should be changed (that already works).
If you (or any one else) run the Schelling example from this PR, hopefully you see what I mean.

jackiekazil:
Discussion during mesa dev meeting.... 
* Will the new javascript stuff being done effect this? 
* Does the batch running have a variant/parallel runner, which allows for the same seed to be set for all the models running, but changes the parameter?

Corvince:
> * Will the new javascript stuff being done effect this?
Hopefully only in a good way. Honestly this PR was one of the reasons why I felt the need to rewrite the JS core parts, because everything was getting too messy and I couldn't get the UserParams part to work as I want it to work for side-by-side visualizations. Ironically, the UserParams are the only part I didn't touch so still don't now how to do this ;) Although I would maybe just need another look since I have a better understanding of how JavaScript works now.

But I already thought about breaking this PR up and first only change the Python ModularServer to handle multiple model runs. And then the visualization part and then the individual viz elements. What do you think?

Corvince:
Stalled, work continued in #718

Issue #490, opened 2018-03-25
Add additional chart types -- bar & pie


jackiekazil:
We have line charts, but we should also have bar charts and pie charts
Add some pie and bar charts. Base them off of the way we did line charts OR update our line charts as well.

jackiekazil:
Related: https://github.com/projectmesa/mesa/issues/515

digitaldust:
is there currently any example of barchart in mesa?

jackiekazil:
No. That is why this ticket exists. Were you thinking of adding one? 😃

markblack96:
I'd be interested in helping. Does Mesa interface well with matplotlib and pyplot? Is this basically asking for wrappers for those implementations within Mesa?

PR #594, opened 2018-10-05
Add Pie Chart visualization and include in forest_fire example (Issue #490)

JamesHovet:
Created new pie chart visualization using d3.js modeled off the current implementation of line charts. 

The client-side interface for adding Pie charts is exactly the same as the that for line charts. 

The pie charts include automatically generated legends and hover-over tooltips. 

screen shot 2018-10-05 at 11 54 22 am


JamesHovet:
PS. I'm pretty bad at git, so feel free to tell me to clean up the commit history and remove that merge commit, but you'll just also have to tell me exactly how to do that.

Corvince:
RE: git

Two and a half reasons you shouldn't worry.

1. It really doesn't matter. There is nothing wrong with having lots of commits. At least you are missing the (for me obligatory) "fixing flake8/travis" commit. Thats already great ;)
2. GitHub has a "squash and merge" option for pull requests. Even if you have 327 commits in your branch, it can be only 1 commit in mesa.
3. I THINK that GitHub also automagically gets rid of these "merge remote-tracking" commits, even without "squash and merge". But thinking about it I am really unsure about this and not sure if it always would make sense. 

So although there are ways to get rid of these commits, its usually not worth the hassle.

JamesHovet:
Thanks for the advice, I appreciate it. 

Also, maintainers, I decided to also write code to add bar charts, so it would probably be wise to wait until I have finished that to do code review/accepting the pull request. I have the basics of the bar chart working now, but now I need to add the code to make the whole thing cleaner and more graceful.

JamesHovet:
image

So many chart possibilities!

JamesHovet:
In terms of testing, I would love to help build some up, but given that the only pre-existing chart type did not have tests associated with it, I am not really sure where to start. So I would appreciate some guidance on what those tests might look like, if applicable/useful.

Corvince:
Thank you so much for this. Unfortunately, I neither have the time nor the JS skill to really check your code. Can you give an estimate to how much reviewing this would need? 

Its really nice to have a consistent way of handling charts. That said I am not sure the current mesa viz way is necessarily the best. Is there something from your experience that could be improved or better generalized? 

Regarding testing: I think any ideas for how to do this are very much appreciated. As you discovered yourself: There is not basis for this yet.

examples/charts/Readme.md
```diff
@@ -0,0 +1,40 @@
+# Mesa Charts Example
+
+## Summary
+
+A modified version of the "bank_reserves" example made to provide examples of mesa's charting tools.
```
Corvince: Are there any modifications to the bank_reserves? If not, why not include the charts there? Also I agree that "charts example" would be more visible (maybe remove bank_reserves?)

mesa/visualization/modules/BarChartVisualization.py
```diff
@@ -0,0 +1,88 @@
...
+        data_collector_name: Name of the DataCollector object in the model to retrieve data from.
+
+    """
+
+    package_includes = ["d3.min.js", "BarChartModule.js"]
+
+    def __init__(self, fields, scope="model", sorting="none",
+                sort_by="none", canvas_height=400,
+                canvas_width=800, data_collector_name="datacollector"):
+        """
```
Corvince: This seems to be the same documentation as in the class description above, but with slightly different wording?

mesa/visualization/templates/js/PieChartModule.js
```diff
@@ -0,0 +1,82 @@
...
+    var svg_tag = "<svg width='" + canvas_width + "' height='" + canvas_height + "' ";
+    svg_tag += "style='border:1px dotted'></svg>";
+    // Append it to #elements
+    var svg_element = $(svg_tag)[0];
+    $("#elements").append(svg_element);
+
+    //create the legend
+    var legend_tag = "<div class='legend'></div>";
+    var legend_element = $(legend_tag)[0];
+    $("#elements").append(legend_element);
```
Corvince: I think it would be clearer to have a single "viz element" than the svg_element and the legend_element separately, but maybe there is a specific reason why you implemented it that way?

examples/charts/Readme.md
JamesHovet: My initial decision to keep the charts out of "bank_reserves" was just to avoid the bloating of bank_reserves. If we are assuming that the example is to show how mesa could be used "in the real world", it makes no sense to have four different charts, three of which tell you the exact same thing. I created the charts example to explicitly show the four kinds of charts that mesa has available: (copy pasted from the charts documentation):

- Line Charts for time-series data of multiple model parameters
- Pie Charts for model parameters
- Bar charts for both model and agent-level parameters

In my opinion, it is better to keep charts separate (as I just used "bank_reserves" because it already existed.) 

As to removing bank reserves, that would be up to y'all as project organizers, but if it served a purpose before it stands to reason that it would continue serving a purpose. (I certainly never felt there were too many examples when i was using NetLogo in my ecology class.)

mesa/visualization/modules/BarChartVisualization.py
JamesHovet: So this is one of those places where I followed the example from the pervious chart example, but where the previous example didn't really make sense to me. I agree that the repetition is odd, but I assumed it had something to do with the way the automatic documentation generation worked. 

I tried to somewhat differentiate what was in the class comment (things that you should know *about* the class) and the initializer comment (things you should know about *how* to use the class), but I would some guidance on if this was the right move or really just about how the documentation on something like this should go. (This is my first semi-big experience writing code/documentation for other people to read aside from my CS class TA)

Corvince:
Thanks for the quick reply. Your answer regarding the code review is already helpful. My question was motivated by a lack of (active) JS developers for mesa. I was just worried that your PR will hang around without any review until it is outdated. If you would have said you are an expert in D3, we could have merged this rather quickly. But it sounds like it might be a good idea to get at least some code review. Although code review sounds pretty formal, we basically want someone to have a second look at your code before merging it. But since this is "only" about visualization and not very likely to break mesa, I hope someone (possibly me) will find the time to look over it. It looks like a great addition.

JamesHovet:
@Corvince 

Sounds great. This is by no means my first D3 project, but no one would be happier than me to have a second set of eyes make sure that everything is in order. Let me know if I can be of any help in that process.

Corvince:
Sorry it took so long to update on this. 

To me this still seems pretty good and a useful addition to the current visualization stack. Since I don't know d3 I did not really check the code on that, but it seems to work pretty good! 

Since this has been waiting for so long, is there anything you want to add or change @JamesHovet ? If not, I will merge

JamesHovet:
Just made a quick bug fix because the charts example wasn't running. (Somehow a variable name mismatch happened?) But other than that, everything I tested worked as it should. I'm happy with it.

Corvince:
Oh, thanks for having another look! I first tested your code and then fixed an error on travis on the test_examples. By our naming conventions this test assumes that if the folder name is "charts" that it has a model named "Charts". So I changed BankReserves to Charts, but forgot to update the server file (and did not test again afterwards). 

Thanks again for the PR!

PR #600, opened 2018-11-06
F8 cleanup

jackiekazil:
Not sure if something snuck past, but there was a bunch of flake8 errors that were causing #561 to fail when I updated from master. 

Let me know if you have any specific questions about the changes that I made. Some of them do look a little odd. 

This should be a high priority, since it will mess up everything else down the line if not addressed.

### DIFF
--- a/mesa/batchrunner.py
+++ b/mesa/batchrunner.py
@@ -6,7 +6,6 @@ Batchrunner
 A single class to manage a batch run or parameter sweep of a given model.
 
 """
-import collections
 import copy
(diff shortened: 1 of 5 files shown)


mesa/batchrunner.py
```diff
@@ -19,18 +18,9 @@
     pathos_support = True
 
 
-def combinations(*items):
```
jackiekazil: This was not being used and the yield was causing an error, so I remove it.

PR #602, opened 2018-11-16
Updating history & version numbers for Kearny v0.8.5

jackiekazil:


jackiekazil:
@dmasad can you look this over and merge if you don't see any issues?

PR #614, opened 2019-01-03
install: Minimize install footprint

rht:
1. ~`click`: only used in tests, hence should be put in dev~
2. `jupyter`: only appears in examples
3. `networkx`: only appears in test_space.py and examples (technically it is implicitly required in `NetworkGrid`, but again, not every users have to install `networkx` if they don't use `NetworkGrid` at all)

This should minimize the install footprint of any libraries that depend on Mesa.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,6 @@ from codecs import open
 requires = [
     'click',
     'cookiecutter',
-    'jupyter',
     'networkx',
     'numpy',
     'pandas',


rht:
`pandas` is a heavy dependency. If only there is a way to construct a DF without having to pull in the full-blown library.

Corvince:
`click` is required for the "mesa" command, so shouldn't be removed
`jupyter` is a very good package to remove from the dependencies !!
`networkx` provides functionality that one would very much expect from any ABM package. I would strongly favor to keep it as a dependency (and its not that heavy either)
`pandas` is probably used by 95% of people using mesa, so I don't see the need to remove it as a dependency (and have people drip over the missing requirement)

It is true that we could remove networkx and pandas, but judging from the issues and the mailing list a lot of mesa users are python beginners. That's way I think of mesa as a "batteries included" package that should ship with everything needed. While it is really not that difficult to install missing requirements, I wouldn't want to risk losing people. Networkx is not that big and pandas is so widely used.

That said it would be nice to have the ability to only install "core" mesa, i.e. without `cookiecutter` , `click` and `pandas`. Could be useful for web services. But I don't know how we could provide this

rht:
> click is required for the "mesa" command, so shouldn't be removed

Ah, didn't see the `import click` at main.py. I have only grepped `from click import`.

My specific use case is for an ABM library that is built on top of Mesa. Having lots of those huge libraries slows down CI test considerably, especially the ones where they have to be cython-compiled even if the pip download has been cached. Also, indirect dependencies sometimes causes install gotchas.

General arguments on whether to include examples dependencies into library dependencies:
> It is true that we could remove networkx and pandas, but judging from the issues and the mailing list a lot of mesa users are python beginners. That's way I think of mesa as a "batteries included" package that should ship with everything needed. 

Every examples directory already has requirements.txt. Error messages caused by missing imports are explicit. The requirements of the examples (they will keep increasing) shouldn't be included in the library's requirements. What if there is an example that uses sklearn/tensorflow/pytorch/any-other-huge-library?
One middle ground is to add `extras.txt` (how networkx does it https://github.com/networkx/networkx/tree/master/requirements) and so people can install via `pip install mesa[extras]`.

Orthogonal to the rest:
> pandas is probably used by 95% of people using mesa, so I don't see the need to remove it as a dependency (and have people drip over the missing requirement)

That's an assumption along the line that 0.95 * everyone uses `matplotlib` and the 5% uses seaborn/bokeh/altair/...
(I use `pyarrow` myself because `pandas` is bloated)

rht:
Note: networkx import is 6x longer than scipy import, which means it is heavy.
```
python3 -X importtime -c 'import networkx' 2> out.log
tuna out.log
```
2019-01-08-233102_1870x845_scrot

Corvince:
My comment wasn't meant to be applicable to example requirements, but just `networkx`. Sorry if you got that impression. And so far `networkx `only gets imported if you use it, so I don't see how the import time is relevant here. 

But I am very positive that #617 will get merged and then networkx will be part of the main library anyway (since, as I said, networks are a key ABM feature). 

Regarding `pandas` I would like to get some more comments by the other developers, but I am not against removing that dependency. Sorry if my comment was misleading but I just don't think we *need* to remove it, but yes, it *could* be removed. 

For `click/cookiecutter` it would also be an idea to just have the mesa commands as an [cli] install (if that is possible?) or even a separate repo. I don't like the current way of simply adding a CLI without asking or even notifying the user.

rht:
> Why include networkx in the dev requirements? (Is that just to throw in the kitchen sink?) I am okay with there. I just wanted you to really think about it.

Because it is used in tests, `nx.complete_graph()`.

> Add it to the test build, because of #617 and then I think that would cover that.
> While the error would naturally show a failed import, not everyone building models is going to know what that means.

#617 makes the `networkx` dependency explicit.
Did you mean:
1. Do not include `networkx` in requirements, then
2. Do lazy-import of `networkx` at the initialization of `NetworkGrid`, if import fails, then print out a message to install networkx instead of printing out all the error stack
-- once again, this is more of to decide whether Mesa should be a kitchen sink library-suite or a base library for more specialized ABM libraries to build on top of.

rht:
Amended so that it removes `jupyter` only. All examples have `jupyter` in their requirements.txt files.

jackiekazil:
This kind of feels like the Django versus Flask approach. Both are valid. I am wondering if we can be both? 

Anyways... @rht for starting this & @Corvince for kicking off discussion.

Issue #612, opened 2018-12-27
Problem with grid and chart visualization


pshpkr:
Hello every one, I'm a student studying ABMS using mesa! Also, this is my first issue in the GitHub.  Please tell me if I made a mistake.

I am currently making a model using a multi-grid. However, when I increased the grid size, the grid module and the chad module overlap each other. I thought this was a problem for my code, but the same problem occurred in the money model example. I really want to know how can I fix it.
I need help. Thanks for reading!

**Describe the bug**
When I raised the grid size from 500 * 500 to 1000 * 1000 in the money model example, the following problems occurred.

**Expected behavior**
I expected the chart module to go under the grid. 

**To Reproduce**

![image](https://user-images.githubusercontent.com/5234756/50469226-d01def80-09ee-11e9-8d49-56e451e5f9ee.png)

**Additional context**
 I'm using Google Chrome and Spyder in Windows 10. Also, I'm using the Mesa 0.8.5

dmasad:
Good catch! The good news is that you didn't do anything wrong; I'm replicating the problem on my end. The bad news is that I'm not sure how to solve it yet. I'll look into it -- and if anyone else can see the issue, please jump in!

JamesHovet:
@pshpkr Just made a PR to hopefully solve this. Check out https://github.com/projectmesa/mesa/pull/615 if you want to use the fix for the project you are working on.

pshpkr:
@JamesHovet Thanks a lot for your help! It works very well!

Corvince:
Thanks a lot for reporting the issue!

PR #615, opened 2019-01-03
Fix issue #612

JamesHovet:
an issue with overlapping vizualization elements was caused due to the
way an interaction handler was implimented. In that implimentation, two
canvas elements were overlayed on top of eachother using absolute
positioning with subsequent elements being positioned with hardcoded
values.

This removed all those hardcoded values and keeps the interaction
handler functioning by wrapping both the canvas elements that the
interaction handler uses in a single div. This also helps maintain
flexability by keeping the visualization element contained in one html
element.

PS. this change was applied to both the standard grid and the hexagonal
grid.


dmasad:
This looks good! I'd like someone with better HTML knowledge to review the code as well, in case I'm missing something.
Should the change applied to the grids also be applied to other modules (especially charts and networks)?

JamesHovet:
@dmasad No need. The original issue arose because in the contributions that added the hover-over interaction (https://github.com/projectmesa/mesa/commit/aa21d223b6c80b3da0d416b66b622680f86a2706 and https://github.com/projectmesa/mesa/commit/105936910d45613b718f97ba55556a1c1ab414b1), values for positions of visualization elements were hard-coded in to allow for the interaction handler to work as that contributor intended. When the user attempted to make visualization elements at sizes other than the default, the hard-coded values reared their head. 

All this PR does is remove those hard-coded values and instead achieve the same goal (letting the interaction handler work) without hard-coded values.

PR #617, opened 2019-01-05
adds wrappers for networkx so you can modify the graph from within mesa

sophiawilliams:
Never done this before not sure what to write here. Goes with ticket I just opened.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -15,7 +15,7 @@ MultiGrid: extension to Grid where each cell is a set of objects.
 # pylint: disable=invalid-name
 
 import itertools
-
+import networkx
 import numpy as np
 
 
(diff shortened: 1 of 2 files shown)


dmasad:
This looks good! I'm adding a few comments on specific lines.

A more general implementation question -- would an alternative way to go about this be to modify the `NetworkGrid` class's `__getattr__` method, so that if the user types something like
`net_grid.add_nodes(node_list)`, the class checks to see whether `add_nodes` is a method of `self.G`, and if so uses that? On one hand, that would remove needing to write each method by hand. On the other hand, networkx has a weird mix of methods that belong to `Graph` vs functions that are under the package-wide `networkx` namespace and take a graph as their first argument, so that might not work in practice.

sophiawilliams:
My logic for not doing it that way was that I imagined if sometime in the future they wanted to support different graph packages or provide an option to use them then it would be easier if it was an interface than if it was all just inherited from NetworkX but I have no idea if that was a reasonable thought process. Maybe that would never be an issue?

rht:
In that case, for every method that wraps the graph operation, e.g. `_place_agent()` there needs to be an if-else code selection,
```python
if graph_is_networkx:
    self.G.node[node_id]['agent'].append(agent)
elif graph_is_graphtool:
    self.G.vertex(node_id)['agent'].append(agent)
elif graph_is_igraph:
    ...
```

Each method becomes more costly to run.

mesa/space.py
```diff
@@ -778,11 +778,120 @@ def out_of_bounds(self, pos):
 
 class NetworkGrid:
     """ Network Grid where each node contains zero or more agents. """
+    def __init__(self, G=None, generator="Graph", args=()):
+        """Default values create empty graph"""
+        if G is not None:
+            self.G = G
+        elif generator is not None:
+            if type(args) is tuple:
```
Corvince: I think the recommended way for type checks is
`isinstance(args, tuple)`
You might want to consider changing this here and below. 
Alternatively you might just as well skip the check and just use a `try...except` statement.
This comment applies multiple times below

sophiawilliams: I think try/except won't work (least not everywhere) because e.g. anything can be a node in networkx graphs.

But I take on board the isinstance comment, it makes sense that anything inheriting from those types could be valid too.

sophiawilliams: Ok I changed that in my branch, is that all I need to do?

Corvince: I can't see your changes, maybe you forgot to push your local changes to your forked repo?

sophiawilliams: herp derp no, forgot to add changed file!

dmasad:
I wrote the review when this was originally submitted, but apparently never actually pressed the button to have the review post -- sorry! 

re why not just inherit from networkx: I think that networkx is also sufficiently agnostic about what nodes are that having the added functionality that @sophiawilliams's wrappers add is necessary, even if there's never a different network library underlying it. (Also, if we were to add another library, I'd suggest just subclassing `NetworkGrid` and keeping the same API, instead of putting if-else in statements in each method).

mesa/space.py
```diff
@@ -15,7 +15,7 @@
 # pylint: disable=invalid-name
 
 import itertools
-
+import networkx
```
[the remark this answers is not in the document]
sophiawilliams: Nope no reason, happy to change it.

mesa/space.py
```diff
@@ -778,11 +778,120 @@ def out_of_bounds(self, pos):
 
 class NetworkGrid:
     """ Network Grid where each node contains zero or more agents. """
+    def __init__(self, G=None, generator="Graph", args=()):
```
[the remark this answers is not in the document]
sophiawilliams: Ok done. Not sure I worded it great though.

tests/test_space.py
```diff
@@ -2,6 +2,8 @@
 
 import networkx as nx
 import numpy as np
+import pytest
```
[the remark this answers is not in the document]
sophiawilliams: No, I just didn't know about self.assertRaises which I would have if I'd looked through the rest of the tests. Changed.

dmasad:
This has been sitting here (but looking in good shape) for the last six months or so. It looks good to me -- I'm bumping this to allow everyone to take another quick look, and unless anyone spots a new issue I'll approve the PR. Sorry for letting it drag on for so long, @6661696c!

pre.tests
```diff
@@ -0,0 +1,34 @@
+============================= test session starts ==============================
```
rht: This file is accidentally committed.

mesa/space.py
```diff
@@ -778,11 +778,120 @@ def out_of_bounds(self, pos):
...
+            else:
+                self.G = getattr(networkx, generator)(args)
 
-    def __init__(self, G):
-        self.G = G
         for node_id in self.G.nodes:
-            G.nodes[node_id]['agent'] = list()
+            self.G.nodes[node_id]['agent'] = list()
+
+    def add_nodes(self, node_ids):
```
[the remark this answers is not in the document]
rht: I still think the method to add a list of nodes, and a node has to be separate. In addition to the consistency with `networkx`'s API, doing this sort of multiple dispatch is not common in Python, and even with multiple dispatch, `List[Node]` and `Node` doesn't have a common superclass that makes sense.

mesa/space.py
```diff
@@ -778,11 +778,119 @@ def out_of_bounds(self, pos):
 
 class NetworkGrid:
     """ Network Grid where each node contains zero or more agents. """
+    """ kwargs = the arguments to the nx generator function
+        e.g for watts_strogatz_graph(n, k, p[, seed])"""
+    def __init__(self, G=None, generator="Graph", **kwargs):
```
rht: The generator should be an object that defaults to `nx.Graph`. This is in line with `networkx`'s `read_edgelist`'s `create_using` argument https://networkx.github.io/documentation/networkx-1.9/reference/generated/networkx.readwrite.edgelist.read_edgelist.html.

PR #621, opened 2019-01-11
[WIP] Visualization with Vega

Corvince:
Implement a VisualisationModule for [vega visualizations](https://vega.github.io/vega/) through [altair](https://altair-viz.github.io/).

As always for my work on the front-end an implementation example can be seen for the Schelling example. (The changes to the Schelling example seem bigger than they are because I have enabled auto-formatting with black which e.g. changed ''' to """ and realized too late)

Things to look out for:
 - Linked selections for mouseover of grid and neighbor-chart
 - Slider at bottom that lets you scroll back and forth through time (steps)

Looking for some feedback on the implementation, ideas for next steps.

### DIFF
--- a/examples/Schelling/model.py
+++ b/examples/Schelling/model.py
@@ -56,7 +56,7 @@ class Schelling(Model):
         self.datacollector = DataCollector(
             {"happy": "happy"},  # Model-level count of happy agents
             # For testing purposes, agent's individual x and y
-            {"x": lambda a: a.pos[0], "y": lambda a: a.pos[1]})
+            {"x": lambda a: a.pos[0], "y": lambda a: a.pos[1], "atype": "type"})
 
         # Set up agents
         # We use a grid iterator that returns
--- a/examples/Schelling/model.py
+++ b/examples/Schelling/model.py
@@ -5,18 +5,19 @@ from mesa.datacollection import DataCollector
 
 
 class SchellingAgent(Agent):
-    '''
+    """
     Schelling segregation agent
-    '''
+    """
+
     def __init__(self, pos, model, agent_type):
-        '''
+        """
          Create a new Schelling agent.
 
          Args:
             unique_id: Unique identifier for the agent.
             x, y: Agent initial location.
             agent_type: Indicator for the agent's type (minority=1, majority=0)
-        '''
+        """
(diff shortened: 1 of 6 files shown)


jdfoote:
This looks super, super cool. I don't know Vega or Mesa very well (yet) but it seems like a nice next step would be to add support for streaming data rather than just loading the data at the end? (e.g., https://vega.github.io/vega-lite/tutorials/streaming.html)

Corvince:
Thanks for your comment @jdfoote ! 

It actually is already streaming in the data! The specification just creates a named data table, which is then populated during model run time. Currently it just adds all the data, which lets you scroll through time, but increases memory consumption over time. The number of steps to be saved could be limited, but its hard to tell where the limit should be.

But in any case if you want to use the same specification for already generated model data, you would have to only change two lines, the "agents" and "model" data, e.g. to .csv files. I think thats one of the great advantages of Vega. 

If you have any ideas or proposals, please let me know

jdfoote:
Aha - wonderful!

I cloned your repository but I can't get the Schelling example to work - I'm getting:
`ImportError:` cannot import name 'VegaVisualization'`
and I can't figure out why

Corvince:
I guess you try running my example, but having installed the "normal" mesa package.

In the cloned repo folder try running `pip install -e .` (note the dot) to (re-)install mesa with the changes. (the -e switch keeps it updated, even when you change files or switch branches)

jackiekazil:
Discussion during Mesa-Dev meeting... 
* Is this still a work in progress? 
* Should we have Schelling model showing both ways? (with and without vega)

Corvince:
> * Is this still a work in progress?
Yes this is still a work in progress, but will take some more time to finish. It's already here to get some early feedback
> * Should we have Schelling model showing both ways? (with and without vega)
Currently the Schelling model is just my "test bed". I don't have a strong opinion on where a vega visualization could/should be showcased. 
> * Can tests be added for the python code added? Coveralls is failing
Yes and it will make sense to test the Python parts. It's on the todo for this PR. 
However, I don't know how to best test the JS code (currently the entire mesa JS code is untested)

Corvince:
Superseded by #664 and #718

Issue #619, opened 2019-01-11
travis pytest version conflict


kumom:
**Describe the bug**
Travis build failed with the hint 
`pkg_resources.VersionConflict: (pytest 3.3.0 (/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages), Requirement.parse('pytest>=3.6'))`.

**To Reproduce**
I used my branch position_agent and the original setup.py to build.

PR #622, opened 2019-01-11
set pytest version >=3.6

kumom:
As I stated in #619, the reason the build failed seems to be Travis is using an old version of pytest, which is not supported for the tested version of python (3.4, 3.5, 3.6) any more. I modified pytest to pytest >= 3.6 and now passes all travis build jobs.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ extras_require = {
     'dev': [
         'coverage',
         'flake8',
-        'pytest',
+        'pytest >= 3.6',
         'pytest-cov',
         'sphinx',
     ],


kumom:
Just a side note, this PR is actually the same as #620...because it is pretty much the first time I opened a pull request and I am not good at git as well. I want to separate different branches so that I can open other PRs later that are specific to one particular problem.

Corvince:
Thanks for spotting and solving this!

jackiekazil:
Yes! Thank you! <3 (I am playing catch up. This was a really good catch!)

Issue #623, opened 2019-01-11
Hierarchy of classes Grid, SingleGrid, and MultiGrid


kumom:
**Describe the bug**

Code snippet from class `Grid`
```python
    def remove_agent(self, agent):
        """ Remove the agent from the grid and set its pos variable to None. """
        pos = agent.pos
        self._remove_agent(pos, agent)
        agent.pos = None

    def _remove_agent(self, pos, agent):
        """ Remove the agent from the given location. """
        x, y = pos
        self.grid[x][y] = None
        self.empties.append(pos)
```

As you see, `Grid._remove_agent` is clearly assuming a SingleGrid behavior. This is fine. However, it is also assuming that `None` is defined as an empty cell, which is not what `Grid.is_cell_empty` does:

```python
    def is_cell_empty(self, pos):
        """ Returns a bool of the contents of a cell. """
        x, y = pos
        return True if self.grid[x][y] == self.default_val() else False
```

Since `Grid.default_value == None`, it seems ok so far. But I guess we need to make it consistent.

So I was thinking `SingleGrid` and `Grid` are really the same. By further looking at the class `SingleGrid`, I found these two methods in it: 

```python
    def position_agent(self, agent, x="random", y="random"):
        """ Position an agent on the grid.
        This is used when first placing agents! Use 'move_to_empty()'
        when you want agents to jump to an empty cell.
        Use 'swap_pos()' to swap agents positions.
        If x or y are positive, they are used, but if "random",
        we get a random position.
        Ensure this random position is not occupied (in Grid).

        """
        if x == "random" or y == "random":
            if len(self.empties) == 0:
                raise Exception("ERROR: Grid full")
            coords = agent.random.choice(self.empties)
        else:
            coords = (x, y)
        agent.pos = coords
        self._place_agent(coords, agent)

    def _place_agent(self, pos, agent):
        if self.is_cell_empty(pos):
            super()._place_agent(pos, agent)
        else:
            raise Exception("Cell not empty")
```

Where in class `Grid` they are
```python
    def place_agent(self, agent, pos):
        """ Position an agent on the grid, and set its pos variable. """
        self._place_agent(pos, agent)
        agent.pos = pos

    def _place_agent(self, pos, agent):
        """ Place the agent at the correct location. """
        x, y = pos
        self.grid[x][y] = agent
        if pos in self.empties:
            self.empties.remove(pos)
```

[...]

PR #625, opened 2019-01-16
Merge Grid, SingleGrid, and MultiGrid into one class

kumom:
This is some follow-up work for reasons explained in #623. I basically merge three classes (`Grid`, `SingleGrid`, `MultiGrid`) into one, and use `multigrid` as a boolean attribute to specify the behavior of the class. I haven't modified the test files, so it will for sure fail since I changed some method signatures. In addition, I notice `HexGrid` is a subtype of `Grid`, and this modification will cause some of its behaviors to fail potentially.

What I intend to do next, if the current hierarchy looks good to you, is to change `multigrid` into `grid_type`. Allowed values of `grid_type` would be `single`, `multi`, or `hex`, so that `HexGrid` would be "merged" to `Grid` class as well. Just a side node, I adopt this style of code because I am somewhat influenced by the [style of sklearn](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html).

I also had this [question](https://stackoverflow.com/questions/54212223/one-public-and-one-private-method-for-same-functionality) as I read the source code. The only reason I can think of is probably to avoid concurrency issues/code optimization done by compiler. Is there any good reason to write the code this way?

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -76,7 +76,7 @@ class Grid:
             *(range(self.width), range(self.height))))
 
     def __getitem__(self, index):
-        return self.grid[index]  # returns column or row?
+        return self.grid[index]
 
     def __iter__(self):
         # create an iterator that chains the
@@ -89,6 +89,15 @@ class Grid:
             return set()
         else:
             return None
+
+    def is_cell_empty(self, pos):
+        """ Returns a bool of the contents of a cell. """
+        x, y = pos
+        return self.grid[x][y] == self.empty_value
+
+    def exists_empty_cells(self):
+        """ Return True if any cells empty else False. """
+        return len(self.empties) > 0
 
     def coord_iter(self):
         """ Returns an iterator of tuples (agent, x, y) over the whole square grid. """
@@ -132,7 +141,7 @@ class Grid:
 
         return neighbors
 
-    def at_row(self, row, include_agents=False):
+    def row_iter(self, row, include_agents=False):
         """ Return an iterator over a specific row
 
         Args:
@@ -145,7 +154,7 @@ class Grid:
             else:
                 yield (x, y)
 
-    def at_col(self, col, include_agents=False):
+    def col_iter(self, col, include_agents=False):
         """ Return an iterator over a specific column """
         x, _ = self._torus_adj((col, 0))
         for y in range(self.height):
@@ -155,10 +164,6 @@ class Grid:
                 yield (x, y)
 
     def _moore(self, pos, radius, get_agents, include_empty):
-        """
-        http://www.conwaylife.com/wiki/Moore_neighbourhood
-        radius: range
-        """
         neighbors = set()
 
         for dx in range(-radius, radius + 1):
@@ -174,10 +179,6 @@ class Grid:
         return neighbors
 
     def _von_neumann(self, pos, radius, get_agents, include_empty):
-        """
-        http://www.conwaylife.com/wiki/Von_Neumann_neighborhood
-        radius: range
-        """
         neighbors = set()
 
         for dx in range(-radius, radius + 1):
@@ -230,7 +231,7 @@ class Grid:
             replace: if True, replace the possibly existed agent at `pos` with `agent`
         """
         if (pos[0] == "random") or (pos[1] == "random"):
-            pos = self._pick_random_position(agent, pos)
+            pos = self.pick_random_position(agent, pos)
         self._place_agent(pos, agent, replace)
         agent.pos = pos
 
@@ -253,8 +254,8 @@ class Grid:
                     raise Exception(
                         "Cell already occupied by agent {}".format(old_agent.unique_id))
 
-    def _pick_random_position(self, agent, pos):
-        if pos == ("random", "random") and self.exists_empty_cells():
+    def pick_random_position(self, agent, pos):
+        if (pos[0], pos[1]) == ("random", "random") and self.exists_empty_cells():
             return agent.random.choice(tuple(self.empties))
         else:
             empties = set()
@@ -291,11 +292,6 @@ class Grid:
         if self.is_cell_empty(pos):
             self.empties.add(pos)
 
-    def is_cell_empty(self, pos):
-        """ Returns a bool of the contents of a cell. """
-        x, y = pos
-        return self.grid[x][y] == self.empty_value
-
     def move_to_empty(self, agent):
         """ Moves agent to a random empty cell, vacating agent's old cell. """
         pos = agent.pos
@@ -306,6 +302,20 @@ class Grid:
         agent.pos = new_pos
         self._remove_agent(pos, agent)
 
+    def agents_on_coords(self, cells):
+        """ Given a list of cell coordinates (x, y), return a list of respective agents
+
+        Args:
+            cells: Array-like of (x, y) tuples, or single tuple.
+
+        Returns:
+            A list of agents corresponding to the given positions.
+        """
+        if not isinstance(cells, list):
+            cells = [cells]
+        return [self.grid[x][y] for x, y in cells]
+
+    # Deprecated methods below
     def find_empty(self):
         """ Pick a random empty cell. """
         from warnings import warn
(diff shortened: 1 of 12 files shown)


mesa/space.py
```diff
@@ -15,28 +15,11 @@
...
-            return wrapped_function(args[0], [args[1]])
-        else:
-            return wrapped_function(*args)
-
-    return wrapper
-
-
 class Grid:
-    """ Base class for a square grid.
+    """ Class for a square grid.
```
Corvince: Shouldn't it be a rectangle grid?

Also, the doc-string for the methods below is not updated

mesa/space.py
```diff
@@ -81,237 +62,207 @@ def __init__(self, width, height, torus):
...
-        """ Return an iterator over cell coordinates that are in the
-        neighborhood of a certain point.
+    def coord_iter(self):
+        """ Returns an iterator of tuples (agent, x, y) over the whole square grid. """
+        for x in range(self.width):
+            for y in range(self.height):
+                yield self.grid[x][y], x, y
 
+    def neighbors(self, pos, moore=True, radius=1, get_agents=False, include_empty=False):
+        """
```
Corvince: Summary line and explanation of the method is missing

mesa/space.py
```diff
@@ -81,237 +62,207 @@ def __init__(self, width, height, torus):
...
-            include_center: If True, return the (x, y) cell as well.
-                            Otherwise, return surrounding cells only.
-            radius: radius, in cells, of neighborhood to get.
+            A set of adjacent cells of a single cell at `pos`. 
 
-        Returns:
-            A list of coordinate tuples representing the neighborhood;
-            With radius 1, at most 9 if Moore, 5 if Von Neumann (8 and 4
-            if not including the center).
+            The number of cells
```
Corvince: Sentence repeated twice

mesa/space.py
```diff
@@ -81,237 +62,207 @@ def __init__(self, width, height, torus):
...
-        """
-        neighborhood = self.iter_neighborhood(
-            pos, moore, include_center, radius)
-        return self.iter_cell_list_contents(neighborhood)
+        return neighbors
 
-    def get_neighbors(self, pos, moore,
-                      include_center=False, radius=1):
-        """ Return a list of neighbors to a certain point.
+    def at_row(self, row, include_agents=False):
```
Corvince: Not sure about the function name. What does the "at" refer to? (for me it sound like it would return a boolean)
Maybe name it row_iter (as coord_iter function, but am not happy with that name either)

mesa/space.py
```diff
@@ -81,237 +62,207 @@ def __init__(self, width, height, torus):
...
-        self.grid[x][y] = agent
-        if pos in self.empties:
+        old_agent = self.grid[x][y]
+
+        if old_agent == self.empty_value:
+            self.grid[x][y] = set(agent) if self.multigrid else agent
             self.empties.remove(pos)
+        else:
+            if replace:
+                old_agent.pos = None
```
Corvince: I think this will fail for multigrid? As old_agent would be a set. It is also not clear how "replace" should work for a multigrid. Should it replace a single agent? All agents? Or can't it be used with MultiGrid?

mesa/space.py
```diff
@@ -81,237 +62,207 @@ def __init__(self, width, height, torus):
...
+                self.grid[x][y] = agent
+            else:
+                if self.multigrid:
+                    self.grid[x][y].add(agent)
+                else:
+                    raise Exception(
+                        "Cell already occupied by agent {}".format(old_agent.unique_id))
+
+    def _pick_random_position(self, agent, pos):
```
Corvince: Any reason why you made this a private function? Could be handy for model makers

mesa/space.py
```diff
@@ -81,237 +62,207 @@ def __init__(self, width, height, torus):
...
+        if pos == ("random", "random") and self.exists_empty_cells():
```
Corvince: I think the function should also work with other sequences, e.g. ["random", "random"]

mesa/space.py
```diff
@@ -81,237 +62,207 @@ def __init__(self, width, height, torus):
...
+            return agent.random.choice(self.empties)
+        else:
+            empties = set()
+            if pos[0] == "random":
+                # We pick a random cell in a specified row
```
Corvince: Comment and code got me confused. You don't pick a random cell here, you just make a set (why not list?) of empty cells and then randomly select one further down

mesa/space.py
```diff
@@ -322,21 +273,26 @@ def remove_agent(self, agent):
     def _remove_agent(self, pos, agent):
         """ Remove the agent from the given location. """
```
Corvince: The agents pos attribute should also be set to None

kumom:
Thanks a lot for the comments! I was rushing a bit to show the hierarchy. I was also a bit conservative due to backward compatibility problem like you said. I haven't really taken a good look at any other classes yet, so I am hesitant about merging `HexGrid` as well. I am implementing a model involving network and I will try to use `NetworkGrid` to test out the functions there. I will look into the code more carefully next week and push new code to the branch.

kumom:
btw, is there any motivation to name a network `NetworkGrid`? I was a bit confused by the name, but since I did not really look at the code and for backward compatibility I left this class intact.

Corvince:
Really quick, so you get this info if you want to work on this on the weekend:

I tested your branch and there are several severe bugs. I suggest adding 

```python

class SingleGrid(Grid):
    """Depreciated class."""
    def __init__(self, width, height, torus):
        super().__init__(width, height, torus, multigrid=False)


class MultiGrid(Grid):
    """Depreciated class."""
    def __init__(self, width, height, torus):
        super().__init__(width, height, torus, multigrid=True)
```

to your file and then run the tests to see a lot of errors (not just related to missing functions). I already made some changes.  I wanted to PR my changes into your repo, but since you did not fork from projectmesa, I wasn't able to tell GitHub where to PR. So [here is a gist](https://gist.github.com/Corvince/c7667c7681ee0dbd7acec3feffff3e39) of my file based on your changes.

Corvince:
As noted already there are several bugs in this implementation. See this PR to your repo for which ones and how I solved them. 

I would suggest the following approach: Make sure the current tests work (by adding in deprecated functions and/or resolve bugs). Then add new tests based on the current implementation. 

Again, I really like your changes/new structure but its not ready for a merge yet. I would also like to see a return of the get_neighborhood function (including center).

mesa/space.py
kumom: The height and width of the tiles are not really specified in this class, and these values are only given for visualization...so I am not sure they should be called square grid or rectangular grid. Should we introduce these new properties?

kumom:
I don't understand why I pass the tests locally but fail in Travis...I really don't want to push a lot to annoy you, but I am not sure why this happens.
```
tests/test_batchrunner.py ....                                           [  4%]
tests/test_datacollector.py ....                                         [  9%]
tests/test_examples.py .                                                 [ 10%]
tests/test_grid.py ...............................                       [ 45%]
tests/test_lifespan.py ..                                                [ 47%]
tests/test_main.py .                                                     [ 48%]
tests/test_scaffold.py F                                                 [ 50%]
tests/test_space.py .............................                        [ 82%]
tests/test_time.py ........                                              [ 92%]
tests/test_usersettableparam.py ....                                     [ 96%]
tests/test_visualization.py ...                                          [100%]Coverage.py warning: No data was collected. (no-data-collected)
````

In addition, the failure is 
```
=================================== FAILURES ===================================
________________ ScaffoldTest.test_scaffold_creates_project_dir ________________

self = <test_scaffold.ScaffoldTest testMethod=test_scaffold_creates_project_dir>

    def test_scaffold_creates_project_dir(self):
        with self.runner.isolated_filesystem():
            assert not os.path.isdir("example_project")
            self.runner.invoke(cli, ['startproject', '--no-input'])
>           assert os.path.isdir("example_project")
E           AssertionError: assert False
E            +  where False = <function isdir at 0x103c989d8>('example_project')
E            +    where <function isdir at 0x103c989d8> = <module 'posixpath' from '/Users/kumom/anaconda3/lib/python3.6/posixpath.py'>.isdir
E            +      where <module 'posixpath' from '/Users/kumom/anaconda3/lib/python3.6/posixpath.py'> = os.path

tests/test_scaffold.py:21: AssertionError
----------------------------- Captured stdout call -----------------------------
A valid repository for "/Users/kumom/anaconda3/lib/python3.6/site-packages/mesa/cookiecutter-mesa" could not be found in the following locations:
/Users/kumom/anaconda3/lib/python3.6/site-packages/mesa/cookiecutter-mesa
/Users/kumom/anaconda3/lib/python3.6/site-packages/mesa/cookiecutter-mesa
```

However, I have passed the local tests when I did the PR #622, and I haven't updated/downgraded the package in any way. Why would this error suddenly pops up?

**Comment by Corvince** [comment]
Exciting new commits! I will review your changes and give you more feedback in the next days.

RE tests: Maybe try to clean your local `.pytest_cache` folders or something like this? Also not sure which tests you pass locally

**Comment by Corvince** [comment]
> After cleaning the cache, I passes all the tests (with the buggy code)...

Thats strange. When I run your changes locally the same tests fail as on travis

**Comment by dmasad** [comment]
@kumom Thanks for working on this -- my memory is that the split into single/multi-grid was a path-dependent artifact and not a well-thought-out design choice, so this is probably long overdue.

I agree with @Corvince that it makes sense to have separate methods for getting `neighbors` (agents) and `neighborhood` (coordinates). Even if one is just a thin wrapper around the other, it's one less thing for a modeler to need to implement.

I'm also seeing the same test failures on my end as Travis is showing.

**Comment by kumom** [comment]
I am still debugging the issue with pytest, and I think I have figured out why all tests are passed - because mesa installed via pip was imported. However, after I ran `pip3 uninstall mesa` and did `pytest --cov=mesa tests/` again, I received the following error

```
__________________ ERROR collecting tests/test_batchrunner.py __________________
ImportError while importing test module '/Users/kumom/Documents/Projects/mesa/tests/test_batchrunner.py'.
Traceback:
tests/test_batchrunner.py:8: in <module>
    from mesa import Agent, Model
E   ImportError: cannot import name 'Agent'
```

I ran this under a new folder cloned from projectmesa/mesa. Why would this happen? Am I doing something wrong with pytest?

Corvince:
Pytest ist fine, but you still have to install Mesa. Instead of pip install Mesa, which pulls it from pypi, try to install it in the root dir (projectmesa/Mesa) with `pip install -e .[dev]`

kumom:
> Pytest ist fine, but you still have to install Mesa. Instead of pip install Mesa, which pulls it from pypi, try to install it in the root dir (projectmesa/Mesa) with `pip install -e .[dev]`

Thanks a lot! Problem solved already

Issue #631, opened 2019-01-31
Re-think Mesa Front-End


Corvince:
Let's be honest: The (codebase of) the current front-end is one big hack.

The good:
 - It works quite well 
 - only uses the Javascript scripts needed for the visualization and lets users provide own (local) JS files.

The bad: 
 - The main front-end logic is all written in a single `runcontrol.js` file
 - This file has grown over time to be a complete mess where function logics are scattered around. 
 - Every script litters the global namespace and depends on other scripts doing the same
 - We mix custom-made visualizations with several chart libraries/frameworks which is very hard to maintain

The ugly:
 - We don't have any style guide for JS files and styles are completely incoherent
 - Everything is written in "old-style" and not using newer JS features
 - Mesa depends on JQuery and bootstrap, adding complexity

In the past, I started several branches to improve the mesa front-end, but every time needed a long time to rethink and remember how everything works and am still struggling with some parts. I think this is a serious stumbling block to further improvements to the mesa visualizations. 

I am actively working on some of those issues (PR coming), but I would like to use this issue to gather further ideas or short-comings of the current state.

Corvince:
I want to re-use this issue to clarify the intention for some of the work I have been doing in the last couple of months.

Here are the goals I have for the next generation of the mesa front-end:

1) Scrub forward and backward in time without recalculating the model state (backend and frontend) #662 
2) Simulate multiple simulations side-by-side #700 #585 
3) Descriptive chart generation: Just tell mesa what you want to display, not how #664 #621 
4) Make charts interactive - Change model parameters from within the visualization, e.g. add agents #664 #660 

Re 1): 
Not on a PR yet, but on the backend we can simply pickle the model state and unpickle if we want to go back to that state. Seems to work rather nicely. On the front-end this is where #662 comes in. With its help we could simply throw the whole state of model attributes and properties to the front end. Visualization modules could then simply pick the variables they need and display them. No more fiddling on both the python and the javascript side, only javascript.

Re 2)
At least for me personally this would be much more straight-forward with #700. Server-side changes are already implemented in #585

Re 3)
Again with #662 we no longer have to fiddle around to get the data, but can concentrate on how to handle it. Also Vega #664 provides a very clean interface for declarative charts once you get used to it.

Re 4)
POC already available in #660 based on Vega. Not sure how an API could look like for that.

PR #632, opened 2019-01-31
Simplify Front-End JS code

Corvince:
In relation to #631 this is work in progress to improve the codebase of the front-end. To the largest part this is a refactoring of the current `runcontrol.js` with the following steps:

1. Update `runcontrol.js` to a JavaScript module `main.js`. This allows to import and export other JS files and split the logic into several files
2. `websocket.js`: A no-nonsense file that establishes the WebSocket and reacts to the messages. The Python now sends the GUI elements and the visualization elements by default upon opening the connection (no need to ask for them from JS). "Message reactions" are simplified/refactored into a appropriate location
3. `controls.js`: Implements a ModelController class that implements the start/stop/step/reset/done methods in a clear way and at a single location. Also connects the functions with the buttons
4. `gui.js`: Handles the creation of the the input GUI (sliders and such). Currently mostly a copy of the current `initGUI` function. Code for the individual elements is imported from:
5. `InputElements.js`: Code that creates the individual input elements. Uses the current bootstrap ~~Currently only exports a SliderInput that is based on a html <template>.~~    
6. A unified code style! I am actually not sure which one, but I use VS code with prettier and it default settings. Should be googleable. 

I also wrote a new ~~simplified~~ `index.html` file. 

I also removed any JQuery and Bootstrap dependencies. The former is perfectly replaceable by standard JS features, while the latter is very much arguable. I did the current work mostly as a JS-learning experience and therefore wanted to start from scratch. This is also the reason why this implementation currently lacks and style and looks completely bare. While the top bar is probably reasonable easy to style with css, the input elements require at least some work. But I honestly never completely understood the bootstrap implementation and had some problems modifying them (e.g. to make them compatibly with side-by-side visualizations). So how to proceed with this is a big discussion point, before I want to do any more work on this part. But, to start small, we could simply copy-paste the current Bootstrap-based style and come to that later. 

Another current issue are the inclusion of the custom VisualizationModules. Their JS_code is now send over the WebSocket and evaluated directly (and not part of the HTML template). This way they still have access to the "elements" variable, but their required scripts are executed in their own namespace. This means for example the `ChartsModule.js` refers to the `control.tick` counter for the x-axis and throws a ReferenceError. It should be sufficient to update all the `*Module.js` files to "real" modules and import the needed variables. But since this also puts the scripts into "strict mode" and throws some errors and just requires a bit of work. Also not on my todo-list if there is no interest for this PR ;)

I probably missed some caveats, but am happy for any comments. I especially would like to know if the code is readable now, especially from Python (or any other non-JS) developers (this refers mostly to `websocket.js` and `controls.js`).


Corvince:
> I also removed any JQuery and Bootstrap dependencies. The former is perfectly replaceable by standard JS features, while the latter is very much arguable. I did the current work mostly as a JS-learning experience and therefore wanted to start from scratch. This is also the reason why this implementation currently lacks and style and looks completely bare. While the top bar is probably reasonable easy to style with css, the input elements require at least some work. But I honestly never completely understood the bootstrap implementation and had some problems modifying them (e.g. to make them compatibly with side-by-side visualizations). So how to proceed with this is a big discussion point, before I want to do any more work on this part. But, to start small, we could simply copy-paste the current Bootstrap-based style and come to that later.

I thought about this a bit more and I think at this point it doesn't make sense to remove bootstrap, at least not without any alternatives. And there are no alternatives in sight (or maybe needed). As I said the changes were mostly a coding exercise, but I really think the simplified code helps mesa front-end development. But there is really no need to throw out the baby with the bathtub. Thanks to @TaylorMutch  the Bootstrap-code was already nice and modular seperated, I now (mostly) just copy-pasted it into its own JS module. The code could be further improved, but that can be done on a seperate PR.

So now the basics should look and work as always! 

Now there is only one thing on the todo-list:

- [ ] Update all the `*Module.js` scripts into real JS modules. Also a great opportunity to update the code style.

This shouldn't be to hard, but requires some time. So we should discuss the necessity for this whole PR first.

TaylorMutch:
Wow, this is really great stuff. Glad to see this happening!

Corvince:
Just realized that the only problem was with the ChartModule and fixed this one now. This PR should be mergeable now without any user-facing changes! 

Looking forward for some review!

Corvince:
superseded by #661

PR #633, opened 2019-02-01
Correct Model Naming

Naldershof:
the renaming of the flockers model appears to have broken the notebook.


jackiekazil:
@Naldershof Thank you for that catch! Much appreciated it! 
The travis integration failed, but it doesn't look like it was anything in your submission, so I am going to pass it through.

PR #634, opened 2019-02-01
Updating the docs for the new version of Chart.js

jdfoote:
The visualization module uses a version of Chart.js that is incompatible with the docs. This updates the docs so that the histogram tutorial works.


dmasad:
This looks good, thanks! It looks like the Travis failure is caused by an issue originating in the `master` branch, not this PR, so I'm going to approve it as-is.

PR #635, opened 2019-02-02
Example updates

jackiekazil:
This is a simple update that does ~~three~~ two things. 
1. Adds matplotlib to requirements files in examples where it was missing. 
2. Add jupyter to requirements files in examples where it was missing. 
3. ~~Renames boid-flockers to boid_flockers to match our styles.~~


Corvince:
Good catch to update these after #614 . Didn't think of that when I merged it

Issue #639, opened 2019-02-12
Viewing multiple agents within the same cell in MultiGrid


mmerrittsmith:
**What's the problem this feature will solve?**
I would like to be able to view all of the agents within a state at the same time when using MultiGrid. Currently the only way to do this visually is using the mouseover tooltip. You can, of course, output the data as to where each agent is, but I believe a more visual approach might prove beneficial.

**Describe the solution you'd like**
I would like each cell to have a sort of invisible grid within it, where each of the individual agents may be placed so all of them can be viewed at the same time.

**Additional context**
I've written this functionality already, and will be creating a pull request referencing this shortly.

quaquel:
solved via #2430 and even before

PR #640, opened 2019-02-12
Creating multi-agent viewing in MultiGrid referenced in #639

mmerrittsmith:
Addressing the feature request I brought up in issue #639 , I wrote these changes to GridDraw, Canvas GridVisualization, CanvasModule, and test_visualization. There may be other tests that I'm missing. 

I also tried to standardize the formatting in the Javascript I encountered, setting it to be four spaces. 

The feature works through the new 'Multiple' boolean variable that can be passed to agentPortrayal.


mmerrittsmith:
A significant downside to this change is that you would be required to pass the number of agents in your model when instantiating your CanvasGrid. So instead of looking like 

grid = CanvasGrid(agentPortrayal, 2, 2)

it would have to be changed to:

grid = CanvasGrid(agentPortrayal, 2, 2, modelName.numAgents).

This introduces backwards compatibility issues of course. 

Is there a more elegant way to do this that I'm not aware of?

Corvince:
Apologies for not responding in any way! On a first glance this looks quite fine! @mmerrittsmith  are you still using this yourself? Have you made any changes since to this since this PR? 

> A significant downside to this change is that you would be required to pass the number of agents in your model when instantiating your CanvasGrid. 

Maybe just give numAgents a default value of 1?

mmerrittsmith:
Hi Corvince, I haven't made any changes since this PR. 

Yeah, that'd be a pretty easy way to do it. I moved on a different method of visualization for the project I was working on, but I'd still be happy to make that slight edit if the team would want this feature added.

Corvince:
Shame we never implemented this PR, but I think it is obsolete by now (would need to be reopened at https://github.com/projectmesa/mesa-viz-tornado )

PR #649, opened 2019-02-21
Change empties from list to set

Corvince:
Currently we save empty cells in a list. This is problematic for large grids with long lists of empty cells.

Running the Boltzmann Wealth example with the following settings

```python
m = BoltzmannWealthModel(2500, 100, 100)
for _ in range(100):
    m.step()
```

on my machine took 68 seconds, of wich 60 were spent on looking up if a position is empty and then removing that position (i.e. look it up again). 

Changing the empties to a Set brings down the total runtime to 8 seconds.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -91,7 +91,7 @@ class Grid:
             self.grid.append(col)
 
         # Add all cells to the empties list.
-        self.empties = list(itertools.product(
+        self.empties = set(itertools.product(
             *(range(self.width), range(self.height))))
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


mesa/space.py
```diff
@@ -323,19 +322,19 @@ def _remove_agent(self, pos, agent):
...
-        return True if self.grid[x][y] == self.default_val() else False
+        return self.grid[x][y] == self.default_val()
 
     def move_to_empty(self, agent):
         """ Moves agent to a random empty cell, vacating agent's old cell. """
         pos = agent.pos
         if len(self.empties) == 0:
             raise Exception("ERROR: No empty cells")
-        new_pos = agent.random.choice(self.empties)
+        new_pos = agent.random.choice(sorted(self.empties))
```
Corvince: Looks clunky, but items in a set have no predictable order in Python and simply getting the first element or changing it to an unordered list would not yield reproducible results)

dmasad:
This looks good to me!

Issue #643, opened 2019-02-14
Proposal: Use a type annotated AbstractSpace to define a clear space API


Corvince:
Based on the discussion of #625 and [on the mailing list](https://groups.google.com/forum/#!topic/projectmesa-dev/H7IUeol8iRM), I started to implement my own version of a revised mesa space module 🙈 . And of course I went down the same pitfalls as the current implementation and @kumom ideas. Thus I took a step back and thought about what space features are really essential.

Based on my considerations, I want to propose the basics of a typed abstract class with an example implementation for a basic grid.

The proposal uses three custom types: Position (Any), Content (Any) and the Agent class
For a (multi-)grid the first two could resolve to GridCoordinates = Tuple[int, int] and CellContent = Set[Agent]

This would mean for every space implementation we have a clear definition of what a position is and how the content of a position looks like.

The beginning of the AbstractSpace could look something like this:

```python
from abc import ABC
from mesa.agent import Agent
from typing import Any, Tuple

Position = Any
Content = Any

class _AbstractSpace(ABC):
    @abstractmethod
    def __init__(self) -> None:
        super().__init__()

    @abstractmethod
    def __getitem__(self, pos: Position) -> Content:
        """Return the content of self at a position.
        Called by `_AbstractSpace()[pos]`.
        """

    @abstractmethod
    def __setitem__(self, pos: Position, content: Content) -> None:
        """Add content to self at position.
        Called by `_AbstractSpace()[pos] = content`.
        """

    @abstractmethod
    def __delitem__(self, content: Tuple[Position, Content]) -> None:
        """Delete content from the position in self.
        Called by `del _AbstractSpace()[pos, content]`.
        """

    def place_agent(self, agent: Agent, pos: Position) -> None:
        """Place an agent at a specific position."""

        self[pos] = agent
        setattr(agent, "pos", pos)

    def remove_agent(self, agent: Agent) -> None:
        """Remove an agent from the space."""

        old_pos = getattr(agent, "pos")
        del self[old_pos, agent]
        setattr(agent, "pos", None)

    def move_agent(self, agent: Agent, pos: Position) -> None:
        """Move an agent from its current to a new position."""

        old_pos = getattr(agent, "pos")
        del self[old_pos, agent]
        self[pos] = agent
        setattr(agent, "pos", pos)
```

The `@abstractmethod` decorator means that we can only instantiate a subclass of `_AbstractSpace` if all those methods are implemented.

[...]

PR #650, opened 2019-02-22
Optimize grid performance

Corvince:
Based on the discussions in #643 this is a small PR that tries to expand the tightest performance bottlenecks of the current grid implementation.

(At least on my machine) a typo (?) made it practically infeasible to create grids a magnitude larger then 100x100 cells. 

The changes include the changes from #649 and make it about twice as fast

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -18,6 +18,10 @@ import itertools
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


Corvince:
(Found some quirks with this and this also is not fully optimized, closing)

PR #654, opened 2019-03-14
removed  extra" .random" on line 178.

deborahduong:
self.model.random.random.shuffle(agent_keys) should be self.model.random.shuffle(agent_keys) .  gives Attribute error otherwise


jackiekazil:
@deborahduong ty for the submission. you have some way to test this or demo the change? If not a test, maybe add something to one of the examples in the example folder? (It makes sense, I just want  to double check.)

jackiekazil:
I did a deeper review and it looks like you are right. Ty for the catch.

PR #655, opened 2019-04-02
Fix VisualizationServer to be compatible with recent versions of Tornado

Corvince:
Since `IOLoop.initialized()` has been depreciated as of tornado 5.0, I changed `IOLoop.instance()` to `IOLoop.current()`- The latter already does check if an IOLoop is already running and returns it or otherwise create a new instance (which was the reason `IOLoop.initialized()` was used before)

Compatible with all recent versions of tornado


jackiekazil:
LGTM

PR #661, opened 2019-04-19
Restructured runcontrol.js

Corvince:
Since my previous PR #632  for restructuring the front-end was hanging around for quite some time without a review, here is yet another attempt to clear things up. 

While the previous PR has split up `runcontrol.js` into several modules, I here simply restructured the file. I think this way it is even easier to understand what is going on and overall a much better solution (because the previous split wasn't very clean).

This PR really shouldn't change anything about what is done in `runcontrol.js`, but rather how it is done. A code review should therefore primarily test if the visualizations for the examples remain unchanged and do not throw any errors. The only change is the addition of a step/tick display below the FPS slider. 

The core aspects:
 * The ModelController is now the primary actor for start(), step() and reset(). Any time-related user interaction goes through this. 
* The WebSocket `onmessage` function now just delegates the task to the controller (or `initGUI` for the initial set-up)
* No changes were made to the `initGUI` function

### DIFF
--- a/mesa/visualization/ModularVisualization.py
+++ b/mesa/visualization/ModularVisualization.py
@@ -173,6 +173,10 @@ class SocketHandler(tornado.websocket.WebSocketHandler):
     def open(self):
         if self.application.verbose:
             print("Socket opened!")
+        self.write_message({
+            "type": "model_params",
+            "params": self.application.user_params
+        })
 
     def check_origin(self, origin):
         return True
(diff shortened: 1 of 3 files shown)


Corvince:
Pycon update: 

This PR should make the main front-end code much more understandable. The model flow should be understandable from the `ModelController`. `ws` deals with the incoming messages from the Tornado Server. `initGUI` creates the user-settable parameter options.

dmasad:
This looks great!

PR #662, opened 2019-04-20
Add as_json function to model and agents

Corvince:
Yet another pull request from me ;)

This PR adds an `as_json` function to the model and agent classes. This function converts all model/agent attributes and properties into a JSON string. Typically, only built-ins are convertible into JSON. Other objects are converted to their string representations or filtered out (the default).

I see the main use case for this approach for visualizations purposes to simply throw over the whole model state, but it can also be used to quickly grab all attributes in a standardized format. 

Example output for an agent:
`agent.as_json(filter=False)`
```json
{
  "unique_id": [0, 0],
  "model": "<model.Schelling object at 0x000001B504590CF8>",
  "pos": [2, 7],
  "type": 1
}
```
With filter=True (default)
`agent.as_json(filter=True)`
```json
{
  "unique_id": [0, 0],
  "pos": [2, 7],
  "type": 1
}
```

For a model 
`model.as_json(filter=True, include_agents=False)`
```json
{
  "height": 10,
  "width": 10,
  "density": 0.8,
  "minority_pc": 0.2,
  "homophily": 3,
  "happy": 0,
  "running": true
}
```

With agents and filter (default):
`model.as_json()`
```
{
  "model": {
    "height": 10,
    "width": 10,
    "density": 0.8,
    "minority_pc": 0.2,
    "homophily": 3,
    "happy": 0,
    "running": true
  },
  "agents": [
    {"unique_id": [0, 0], "pos": [2, 7], "type": 1},
    {"unique_id": [0, 1], "pos": [8, 3], "type": 0},
    ... (more agents)
  ]
}
```

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -5,10 +5,12 @@ The agent class for Mesa framework.
 Core Objects: Agent
 
 """
+import json
 
 
 class Agent:
     """ Base class for a model agent. """
+
     def __init__(self, unique_id, model):
         """ Create a new agent. """
         self.unique_id = unique_id
@@ -21,3 +23,37 @@ class Agent:
     @property
     def random(self):
         return self.model.random
+
+    def as_json(self, filter: bool = True) -> str:
+        """Convert Agent attributes to JSON string.
+
+        Args:
+            filter: Whether to filter out unserializable objects and private attributes
+
+        Notes:
+            If an attribute is not JSON-serializable, it is replaced by its
+            string representation unless `filter` is set to True.
+        """
+
+        attributes = json.loads(
+            json.dumps(
+                self.__dict__, default=lambda a: "__REMOVE_ATR" if filter else str(a)
+            )
+        )
+
+        if filter:
+            attributes = {
+                key: value
+                for key, value in attributes.items()
+                if value != "__REMOVE_ATR" and not key.startswith("_")
+            }
+
+        properties = {
+            key: getattr(self, key)
+            for key, value in type(self).__dict__.items()
+            if type(value) == property
+        }
+
+        agent_json = json.dumps({**attributes, **properties})
+
+        return agent_json
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -30,9 +30,15 @@ class Agent:
         Args:
             filter: Whether to filter out unserializable objects and private attributes
 
+        Returns:
+            string representation of attributes and properties
+
         Notes:
             If an attribute is not JSON-serializable, it is replaced by its
             string representation unless `filter` is set to True.
+
+            The JSON representation also includes attributes of base classes, but
+            properties of base classes are currently not supported.
         """
 
         attributes = json.loads(
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -27,19 +27,27 @@ class Agent:
     def as_json(self, filter: bool = True) -> str:
         """Convert Agent attributes to JSON string.
 
-        Args:
-            filter: Whether to filter out unserializable objects and private attributes
-
         Returns:
             string representation of attributes and properties
 
         Notes:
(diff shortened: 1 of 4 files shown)


mesa/agent.py
```diff
@@ -21,3 +23,37 @@ def step(self):
...
+            filter: Whether to filter out unserializable objects and private attributes
+
+        Notes:
+            If an attribute is not JSON-serializable, it is replaced by its
+            string representation unless `filter` is set to True.
+        """
+
+        attributes = json.loads(
+            json.dumps(
+                self.__dict__, default=lambda a: "__REMOVE_ATR" if filter else str(a)
```
Corvince: Dumping and loading might not seem very efficient, but it is probably the most reliable way to find out which attributes can be serialized. (the attributes are later merged with the properties, hence the loading)

mesa/agent.py
```diff
@@ -21,3 +23,37 @@ def step(self):
...
+            attributes = {
+                key: value
+                for key, value in attributes.items()
+                if value != "__REMOVE_ATR" and not key.startswith("_")
+            }
+
+        properties = {
+            key: getattr(self, key)
+            for key, value in type(self).__dict__.items()
+            if type(value) == property
```
Corvince: Who would have guessed it is so difficult to get object properties?

Corvince:
Revised the model_json with option include_agents=True to be more consistent with the option being turned off. Expected output now:
```
{
  "height": 10,
  "width": 10,
  "density": 0.8,
  "minority_pc": 0.2,
  "homophily": 3,
  "happy": 0,
  "running": true,
  "agents": [
    {"unique_id": [0, 0], "pos": [2, 7], "type": 1},
    {"unique_id": [0, 1], "pos": [8, 3], "type": 0},
    ... (more agents)
  ]
}

Corvince:
pycon update:

This PR still needs some tests to make sure it creates the correct JSON output

rht:
I found a general library that "pickles" objects to json strings: https://jsonpickle.github.io/.

> jsonpickle can take almost any Python object and turn the object into JSON. Additionally, it can reconstitute the object back into Python.

Corvince:
jsonpickle looks promising at first, but I don't really see its usecase. The default json object will be (as promised) very pickle-like and not really usable for interoperability with other JSON decoders, because of the almost proprietary format. For example tuples are saved as `{"py/tuple": [9, 2]}`. jsonpickle also has the option `unpickleable=False`, which gives much simpler and standard output. But it first crashes my Python (maybe because of the circular relation between agents and models) and does not support object properties. 

Thanks for the tip though

Corvince:
I added some tests for this commit, so it is almost ready to merge. However, one of the test fails and I am not sure how to proceed. Recap: This PR outputs agent and model attributes and properties in a JSON format. If the attributes are not JSON-serializable, they are by default excluded or optionally replaced by their `str`-representation. 

For properties, the `__dict__` attribute of the object type is accessed and filtered for properties. However, this does not seem to work for the properties of the base class. However, this currently only affects the `random` property of agents, if users really want to have something like `"random": "<random.Random object at 0x000001B65CE2CF88>"` inside their agent-json (it is still visible in the model-json). 

So I think this is a no-issue for now. The only thing that could happen is that in the future the base classes of Agent and Model could include more properties which we actually would want to parse as JSON.

However, I would propose (and will do now) to add a note to the function and leave it with that for now. The only solutions I could come up with would be a bit overkill for something probably no one would use and slow things down for the things people might use.

Corvince:
Revised a bit, because it could fail for some properties. Implementation is now actually simpler, but I excluded the "filter" option.

Corvince:
closed due to inactivity (there was an issue I don't even remember which caused it)

jackiekazil:
Ty for the clean up. I think we need to do this all around. I hate passing
on things, but it will help us better manage.

PR #678, opened 2019-05-07
Add pipenv support for mesa

djfurman:
- Adds support for `pipenv` deterministic dependencies for developers and data

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -1,0 +1,14 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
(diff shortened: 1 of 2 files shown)


jackiekazil:
This is great addition. Thank you!

Corvince:
I know I am a bit late to the game (and still catching up on pipenv), but @djfurman can you explain what is the benefit of having a Pipfile for mesa as a library? More specifically, why should a user who wants deterministic dependencies for her project prefer the Pipfile inside mesa instead of simply `pipenv install mesa`?

PR #684, opened 2019-05-07
Bulk merge of docs from 'docs' to 'master'

jackiekazil:
This is a bunch of doc updates that are coming into master. These have already been reviewed, so I will do a bulk merge after the last one that I am expecting comes in:  #668, #669, #671, #672, #674, #675, #677, #682, #683, ...

PR #651, opened 2019-02-25
Batch runner params

dcunning11235:
PR that I think satisfies #607 

Added two new utility classes (ParameterProduct, ParameterSampler) for creating parameter combinations, and a base class FixedBatchRunner, which is now the super for BatchRunner, that takes a list(-like) of parameter combinations to run.

PR #691, opened 2019-05-18
Resources #651 in docs branch

jackiekazil:
RE: 684#issuecomment-492195704

This should restore #651. With this ticket, #684 should be ready for a merge into master.


jackiekazil:
@Corvince, since you caught, can you give this a review? & Then give #684 a final look over?

PR #694, opened 2019-05-23
Address some of the concerns in #623

rht:
Consider this to be the lightweight version of #625. The remaining problem is the `place_agent()`, `position_agent()` API.

### DIFF
--- a/examples/forest_fire/forest_fire/model.py
+++ b/examples/forest_fire/forest_fire/model.py
@@ -39,7 +39,7 @@ class ForestFire(Model):
                 # Set all trees in the first column on fire.
                 if x == 0:
                     new_tree.condition = "On Fire"
-                self.grid._place_agent((x, y), new_tree)
+                self.grid._place_agent(new_tree, (x, y))
                 self.schedule.add(new_tree)
 
         self.running = True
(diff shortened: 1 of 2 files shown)


examples/forest_fire/forest_fire/model.py
```diff
@@ -40,7 +40,7 @@ def __init__(self, height=100, width=100, density=0.65):
                 # Set all trees in the first column on fire.
                 if x == 0:
                     new_tree.condition = "On Fire"
-                self.grid._place_agent((x, y), new_tree)
+                self.grid._place_agent(new_tree, (x, y))
```
rht: This example is problematic because it is using an internal method.

Corvince:
Thanks for the PR. I would suggest also changing the problematic example

> The remaining problem is the `place_agent()`, `position_agent()` API.

I also recently stumpled over `iter_neighbors()` and `neighbor_iter()` which seem to do the same thing, except the latter is an older, limited version of the former.

rht:
Rebased.
> I would suggest also changing the problematic example

I will do this in a separate PR.

rht:
Superseded by #1240.

PR #699, opened 2019-06-07
Travis: Explicitly set the dist to be Xenial

rht:
Trusty has reached EOL. While in https://changelog.travis-ci.com/ there is an indication to set Xenial as the default, it still hasn't happened completely yet.

> Right now, 5% of the repositories building that do not explicitly specify an Operating System, os: key, or an Ubuntu Linux distribution, via the dist: key are now being routed to Ubuntu Xenial 16.04 instead of Ubuntu Trusty 14.04.

So better set it explicitly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: xenial
 language: python
 sudo: false
 os:
@@ -5,7 +6,6 @@ os:
 cache:
   pip: true
 python:
-  - "3.4"
   - "3.5"
   - "3.6"
   # - "nightly"
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,8 @@ cache:
 python:
   - "3.5"
   - "3.6"
-  # - "nightly"
+  - "3.7"
+  - "nightly"
 install:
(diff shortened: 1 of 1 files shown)


rht:
Note: 3.4 is removed because the oldest version in Xenial is 3.5.

Corvince:
A travis update is definitely necessary! 

> Note: 3.4 is removed because the oldest version in Xenial is 3.5.

On the other side of the build spectrum, I think it makes sense to include "3.7" and "nightly" (recent development version). I added a commit for those. 

Will merge if there is no opposition to include "nightly" ?

rht:
On nightly, there is an error when building wheel for pandas: https://travis-ci.org/projectmesa/mesa/jobs/542686219#L350. Most of the dependencies are probably not tested on nightly, so we are constrained by this.

Corvince:
I don't know if that is an opposition or not. However, given that 3.8 is already in beta status, I think we should test it. Might have to switch off nightly for the first 3.9 builds again, but by then we should enable 3.8 anyways so I keep it as it is

PR #700, opened 2019-06-13
New Front End

Corvince:
Okay, this is something I have been working on for far too long, but the process has been very insightful.

My journey started with #585 (side-by-side visualizations), where I just couldn't work out how to create individual sliders for each model (and style them as I want). The mixture of plain javascript, jquery and bootstrap seemed just too confusing to me. And I wanted to learn more JavaScript, so.... I recreated the mesa front end without any dependencies in plain JavaScript. It took quite long, but I eventually created something that kind-of-worked, looked kind-of-ugly and broke way too fast. Turns out there is a reason why people use bootstrap and create re-usable components. So my next adventure got me to the fancy frameworks (react, etc.), but this seems to be way to overkill. I finally learned about web components and the great [weightless](https://www.weightless.dev) component library. 

As an example compare the current bootstrap-modal
```html
    <div id="about" class="modal fade" tabindex="-1" role="dialog">
        <div class="modal-dialog modal-lg">
            <div class="modal-content">
                <div class="modal-header">
                    <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
                    <h4 class="modal-title">About {{ model_name }}</h4>
                </div>
                <div class="modal-body">
                    <div>{{ description }}</div>
                    <div>&nbsp;</div>
                    <div style="clear: both;"></div>
                </div>
            </div>
        </div>
```
With the weightless dialog equivalent
```html
        <wl-button onclick="this.nextElementSibling.show()">
          About
        </wl-button>
        <wl-dialog fixed backdrop>
          <h3 slot="header">Model Description</h3>
          <wl-text slot="content">{{ description }}</wl-text>
        </wl-dialog>
```
Under the hood the custom elements just shove a similar amount of nodes into the shadow dom, but it lets you write clean HTML. 
However, we need a couple of own custom elements for the mesa visualization. I finally opted for the relatively new [hybrids](https://hybrids.js.org) library, which lets you define custom elements as plain objects and  with pure functions. Unfortunately, it is not the most intuitive library, but once you get used to it, it is a joyful experience. 

Here is a reduced example of the FPS-Control
```javascript
const FPSControl = {
  store: parent(AppStore),
  fps: ({ store }) => store.fps,
  render: ({ fps }) => html`
    <wl-slider min="0"  max="25" step="1" value=${fps} 
      oninput=${updateFPS} label="FPS Control" thumbLabel>
    </wl-slider>
  `,
}
```

On my journey I learned that state management is really quite important. But I didn't want to use yet another library for that, so my implementation uses an <app-store> custom element created with hybrids as a top-level element that holds the complete state of the app: step counter, running state, finished, FPS and the list of user-settable parameters. Everything just automatically updates once the state variables change. And I took great care to clearly separate out any side effects: Namely communicating with the tornado server and updating the visualization state. Those happen as side effects of state changes or introduce changes of the state "from the outside". 

This should, in theory, allow to easily introduce new features without interfering with the rest of the front-end. For example it doesn't matter where, when and how the current step changes, everything will just act accordingly. 

Right now this is basically a clone of the current front-end without much new functionality and probably some quirks here and there. But I (personally) am now in complete control/knowledge of how the front-end works. This should allow me to quickly (once I find the time...) add new features from my WIP list, most notably side-by-side visualizations and scrubbing back- and forward through time.

For the interested: I'll add a screencast to this conversation to show how it looks now, or you can just any example from within this branch.


Corvince:
Okay, so here is how it looks like: 

chrome-capture (1)


And a static image showcasing some of the feature changes. 

NewViz

And for comparison how it currently looks

chrome-capture (2)


I am very happy to discuss further ideas or questions about my new implementation

dmasad:
This looks like a great start! A few quick questions / concerns off the top of my head:
- When I run it on my end, the collapsible sliders are showing up on top instead of on the right, each taking up the entire width, which takes up too much space.
- An advantage of using bootstrap is that my understanding is that it's pretty ubiquitous, so more contributors are likely to have experience with it. I worry that using newer / less-common libraries will create an additional barrier. 
- Is it worth considering at least initially having both versions existing side by side, with a parameter selecting which template to use?

Corvince:
Thanks for taking a look!
> * When I run it on my end, the collapsible sliders are showing up on top instead of on the right, each taking up the entire width, which takes up too much space.

Hm, I can take a look but maybe it is a caching issue of the CSS. Did you try to do a hard reload of the web page? (shift-f5)

> * An advantage of using bootstrap is that my understanding is that it's pretty ubiquitous, so more contributors are likely to have experience with it. I worry that using newer / less-common libraries will create an additional barrier.

I think bootstrap is fine, but we currently also use bootstrap-slider and bootstrap-, since these are not part of bootstrap itself, but those packages are not as well maintained. This leaves us with an amount of bootstrap that is way too overkill for what we actually use. But again, bootstrap is fine and I mainly wanted to try something new which I more thoroughly understand. But you are completely right that using "hybridsjs" is a very niche solution and probably should be reconsidered. 

> * Is it worth considering at least initially having both versions existing side by side, with a parameter selecting which template to use?

This. I thought the same and actually started to clone and modify the ModularServer into a new VegaServer that uses the new template and works slightly different and only with vega visualizations. While it will be incompatible with the current server, this approach allows me iterate much faster on my different projects (side-by-side viz, interactive viz and "rewind") without breaking the current state. And if things work nicely and prove to be better we can port features from VegaServer to ModularServer

Corvince:
Superseded by #718

Issue #558, opened 2018-07-14
Create a docker container with mesa to deploy easily


jackiekazil:
A few people have requested functionality to deploy Mesa more easily -- specifically docker has been requested. Please post if you have questions or if you want to "own" this.

alee:
Would be happy to contribute to this but I'm not too familiar with Mesa yet. What should be bundled in a Mesa docker image besides a basic Python 3 container that has mesa pip installed already? Jupyter notebooks, GIS tools?

dmasad:
Thanks for offering! 
1. The only GUI Mesa uses is via the browser. I'm not familiar with containerization at all; but if it can run a local server that the user can access in the browser, it should work fine.

2. At the moment, I don't _think_ anything is necessary besides Python + Mesa's dependencies. However, if anyone with specific use-cases for containers need something else, hopefully they'll say so here.

jackiekazil:
Sorry for the delay in our response. This look great. Hoping to have a chance to test soon. 
Is this something you would interested in submitting to the main repo as a PR?

alee:
Sure, I'd be happy to though should the Dockerfile that lives in this main repo be tailored for Mesa developers vs. Mesa model developers?

wincdev:
Hi,
I would like to work on this task. Can you assignees this task to me?

rht:
I'd say if it is possible to run the GUI in a Jupyter notebook, then it can be run+edited without any setup via Colaboratory / mybinder.org / Azure notebook.

MichaelJCole:
Hi, was there a reason not to use the official python image as a base?

https://hub.docker.com/_/python

alee:
The official Python image would have been fine, though I've been happy with the phusion base image for reasons mostly outlined here:

https://github.com/phusion/baseimage-docker

That said, it's probably unnecessary for the phusion image's additional services just to run / develop a mesa model. 

It's straightforward to create additional tagged images that mirror the latest Debian-based Python images.. we've got a dockerhub automated build set up at https://github.com/comses/docker-mesa so if you'd like to submit a PR for that please feel free!

MichaelJCole:
@alee There is an attack pattern that works like this:

1) Create useful public library/package/docker image
2) People use the package
3) Update package dependencies to include malicious code

Here's two recent examples: 

https://snyk.io/blog/malicious-code-found-in-npm-package-event-stream/
[...]

PR #702, opened 2019-06-19
Dockerizing Mesa - example Dockerfile and docker-compose.yml

MichaelJCole:
Hello, this commit on dockerizing Mesa was requested by @jackiekazil in https://github.com/projectmesa/mesa/issues/558

Because Mesa is packaged into a Pip library, there isn't anything magical the Dockerfile has to do other than:
- setup python3 
- setup dependency managers.  

Luckily [Docker already packages Python for us](https://hub.docker.com/_/python).

This `Dockerfile` shows how to package one of the examples - wolf_sheep.
The `docker-compose.yml` shows how to mount a custom model (vs packaging it).

See comments in `Dockerfile`, then `docker-compose.yml`

Michael

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,9 +14,9 @@
 # $ docker exec -it mymesa_instance /bin/bash 
 
 # FOR SECURITY:
-# - Don't use random Docker images or libraries.
-# - Use Docker images and libraries packaged by the same people.  
-# - Otherwise, who knows what's in that image?!
+# - Don't use random Docker images, libraries, or napkins in the bin
+# - Use official Docker images, or images packaged by the same people that made the library.
+# - Otherwise, who knows what code is in that image?!
 
 # This example Dockerfile packages the wolf_sheep example model.
 
@@ -42,9 +42,6 @@
 
 # Use official python image from Docker team: https://hub.docker.com/_/python
 FROM python:stretch
-
-# Our "home" directory in the Docker image
-WORKDIR /opt/mesa
 
 # Expose our port:
 EXPOSE 8521/tcp
@@ -73,10 +70,15 @@ COPY examples /opt/mesa/examples
 # Run wolf_sheep so it does something by default
 COPY examples/wolf_sheep /opt/mesa/mymodel
 
+# Our "home" directory in the Docker image
+WORKDIR /opt/mesa/mymodel
+
 # NOTE: your model's dependencies are installed here
-RUN (cd mymodel && pipenv install)
+RUN pip install -r requirements.txt
+# pipenv LIES and does sufficiently process requirements.txt
+# and then takes forever to do it.
+# RUN pipenv install
 
 # Run python
 ENTRYPOINT ["/usr/local/bin/python3"]
-# NOTE: replace with your own parameter
-CMD ["/opt/mesa/mymodel/run.py"]
+CMD ["run.py"]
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,22 +1,89 @@
...
+RUN apt-get update && apt-get upgrade -y 
+
+# Don't buffer output: https://docs.python.org/3.7/using/cmdline.html?highlight=pythonunbuffered#envvar-PYTHONUNBUFFERED
+ENV PYTHONUNBUFFERED=1 
+# Cruft? Set locale to use UTF-8: http://www.iac.es/sieinvens/siepedia/pmwiki.php?n=Tutorials.LinuxLocale
+ENV LANG=C.UTF-8
+ENV LC_ALL=C.UTF-8
+
+# Install pipenv
+RUN pip install --upgrade pipenv
```
rht: `pipenv` is not yet used as the default installation method, is it?

Dockerfile
```diff
@@ -1,22 +1,89 @@
...
+
+# Add examples for fun
+COPY examples /opt/mesa/examples
+
+# Our "home" directory in the Docker image
+WORKDIR /opt/mesa/mymodel
+
+# Copy requirements.txt w/o code.  This layer is cached so we don't have to push every build
+# NOTE: copy your model requirements.txt here
+COPY examples/wolf_sheep/requirements.txt .
```
rht: For convenience, maybe the `examples/wolf_sheep` can be abstracted to `$MODELDIR` ENV/ARG, so that only 1 variable needs to be specified.

Dockerfile
MichaelJCole: This file is an example.  Models use Mesa as a library.  So they may use pipenv for their dependencies.  This pipenv is for the person using the example, not Mesa.

Dockerfile
MichaelJCole: This file is an example.  In my experience, unnecessary abstraction makes examples more confusing.

Dockerfile
rht: The file is a generic Dockerfile that can run any simulations that may have various install methods, be it in requirements.txt, setup.py, Pipfile, or pyproject.toml. In some cases where requirements.txt is not sufficient (e.g. `cythonize`), only `setup.py install` can be used. Since the `pipenv install` line is commented out, this (optional) line should, too.

Dockerfile
rht: This is just my opinion: I'd say `$MODELDIR` naming is a sufficiently descriptive alias. `$MODELDIR/requirements.txt` is clearer than `path/to/model/requirements.txt` with `path/to/model` being repeated elsewhere in the file.

rht: Note: not a huge blocker though, the comments elsewhere are highly instructive additions.

Corvince:
@MichaelJCole you still available for this? The comments mention that there is no official mesa docker image and this is primarly an example for one. What would it take in your opinion to create an official image?

jackiekazil:
@Corvince would love to understand that as well! 
I feel like we should move forward with merging this. 

Are there any objections?

rht:
I made a newer version of this in #1193, which has a simpler UX for user to try out various models.

rht:
OK, #1193 has been merged. It doesn't have a docker-compose.yml yet, but it should be a quick add. Regarding with creating an official Docker image for Mesa, keep in mind that it needs to be updated regularly on Dockerhub, which is a maintenance overhead. But a better argument for not doing it is that Numpy doesn't do it either (but they do have 2 images, as you can see in https://hub.docker.com/u/numpy). I think people would usually create their own layers on top of the official Python images.

rht:
The Docker Compose version in #1197, and so, I will close this PR.

PR #705, opened 2019-06-29
intro tutorial: Clarify a function that is not defined in the class

### LINKED ISSUE #704: Add clarification to intro tutorial on compute_gini
There is some confusion on the compute_gini function in the intro tutorial for mesa. Review the docs and add clarification, so that another user doesn't run into the issue presented in #697.

rht:
This fixes #704.
The only case when the mistake happens is when someone is extracting only the class definition from the tutorial. Otherwise, this would be fine elsewhere in the .ipynb, examples/ files, `compute_gini()` when no partial copying is needed.

This is an in-place comment that would become visible in case the NameError (function not defined error) happens.

### DIFF
--- a/docs/tutorials/intro_tutorial.rst
+++ b/docs/tutorials/intro_tutorial.rst
@@ -672,7 +672,7 @@ measure of wealth inequality.
                 self.grid.place_agent(a, (x, y))
 
             self.datacollector = DataCollector(
-                model_reporters={"Gini": compute_gini},
+                model_reporters={"Gini": compute_gini},  # `compute_gini` defined above
                 agent_reporters={"Wealth": "wealth"})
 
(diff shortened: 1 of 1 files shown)


rht:
The comment breaks the 100 char width limit though.

docs/tutorials/intro_tutorial.rst
```diff
@@ -672,7 +672,7 @@ measure of wealth inequality.
                 self.grid.place_agent(a, (x, y))
 
             self.datacollector = DataCollector(
-                model_reporters={"Gini": compute_gini},
+                model_reporters={"Gini": compute_gini},  # `compute_gini` defined above
```
rht: `# fn is defined above`

PR #711, opened 2019-07-11
Typo in ModularServer's last parameter

yoon-gu:
It should have been placed in the form of `dictionary`.

```python
class ModularServer(tornado.web.Application):
    ...
    def __init__(self, model_cls, visualization_elements, name="Mesa Model",
                 model_params={}):
```


Corvince:
Thanks for the PR! Looks good and fixes #717

PR #718, opened 2019-07-31
Additional Visualization Server: Vega server

Corvince:
Okay so based on the discussion in #700 I put together a new, separate visualization server. It works slightly different from the current ModularServer. Instead of being based on different Visualization Modules, it solely supports Vega(-Lite) charts. This significantly reduces the boilerplate code to set up visualization modules both in Python and JavaScript. Instead, visualizations are generated from a list of Vega(-Lite) specifications. 

Otherwise, user-setup is exactly the same, but the following enhancements are made:

- Supports running multiple models side-by-side by simply setting `n_simulations`
- Allows to scrub backward and forward through visualization states
- Independently set `UserSettableParameters` for parallel models
- Allows to make charts interactive by defining an `on_click` function at the model level
- Re-Run models from previous states after making interactive changes

Since this is quite a large project I only submit a draft pull request for now. I am eagerly looking forward to some usability feedback. The PR includes two (updated) examples: 

* Conways game of life, where you can change the grid size and interactively set cells as alive by clicking on dead cells
* Schelling side-by-side, where you can compare 3 models by varying their density and/or switch agent type by clicking on cells

I think these changes allow for quite some interesting experiments.


Corvince:
Front-end preview available in https://github.com/projectmesa/mesa/pull/700#issuecomment-502024564

Corvince:
Thanks to the inspiration of @dmasad you can now quickly preview this version on Glitch! 🐟 

Just point your browsers to https://four-e.glitch.me/ and try the following:

1. Change the size of the grid to something relative small (otherwise it runs a bit slow on glitch) and hit reset
2. Click on the blue grid to set some cells alive.
3. Let the model run for some steps
4. Scrub backward and forward in time using the slider on the lower right

examples/conways_game_of_life/conways_game_of_life/server.py
```diff
@@ -1,12 +1,32 @@
...
-from .portrayal import portrayCell
-from .model import ConwaysGameOfLife
 
+from .model import ConwaysGameOfLife
 
-# Make a world that is 50x50, on a 250x250 display.
-canvas_element = CanvasGrid(portrayCell, 50, 50, 250, 250)
+grid_spec = """
+{
+    "$schema": "https://vega.github.io/schema/vega-lite/v3.json",
```
jackiekazil: Is `$schema` a standard naming convention? (I am not a fan of the `$`.)

examples/conways_game_of_life/conways_game_of_life/server.py
```diff
@@ -1,12 +1,32 @@
...
+      "x": {"type": "nominal", "field": "x"},
+      "y": {"type": "nominal", "field": "y"},
+      "color": {"type": "nominal", "field": "isAlive"}
+    }
+}
+"""
 
-server = ModularServer(ConwaysGameOfLife, [canvas_element], "Game of Life",
-                       {"height": 50, "width": 50})
+server = VegaServer(
```
jackiekazil: Because it says - `VegaServer`, are you suggesting that we maintain both or throw away the ModularServer?

examples/schelling_side_by_side/server.py
```diff
@@ -0,0 +1,49 @@
...
+    "density": UserSettableParameter("slider", "Agent density", 0.8, 0.1, 1.0, 0.1),
+    "schedule": UserSettableParameter(
+        "choice",
+        "Activation",
+        value="RandomActivation",
+        choices=["RandomActivation", "BaseScheduler"],
+    ),
+}
+
+server = VegaServer(Schelling, [grid_spec, line_spec], "Schelling", model_params, 3)
```
jackiekazil: It is not clear to me how the 3 instances are created -- is that by the "3" in this line?

jackiekazil:
This ticket brings me to another discussion with Mesa -- speed profiling. We have not consistently done this, but I feel like we should have some sort of check with changes to understand what the impact is from a speed perspective. (I haven't thought about this much past what I just said.) This feels like something that should be built into our CI/CD process. 

Thoughts? (The question is, do we try to get this type of feature integrated before or after we merge a change like this in.)

examples/conways_game_of_life/conways_game_of_life/server.py
Corvince: There was some discussion on this somewhere else and the idea is to maintain both for now. While VegaServer adds several new functionalities, it can't handle custom visualisation elements and would break every existing model

examples/schelling_side_by_side/server.py
Corvince: Ah yes, the 3 is for a keyword currently called "n_simulations", which could be named better and called explicitly

examples/conways_game_of_life/conways_game_of_life/server.py
Corvince: Yes, this is part of the JSON schema specifications. It's a special keyword and for example picked up by editors/IDEs to provide autocompletion

Corvince:
work continues at https://github.com/Corvince/mesa-viz

jackiekazil:
@Corvince I love that this work lived on as a separate repo. I added it to the wiki page for now --- https://github.com/projectmesa/mesa/wiki/Mesa-Packages

PR #732, opened 2019-09-25
Add pre-commit to automatically maintain a standard formatting with black

SebastianoF:
Hello!

Thanks for maintaining mesa!

Small PR to embed an automatic formatter (current version of the code is not compliant with pep8 convention). To achieve this I suggest pre-commit and black, as they provides strict formatting with no room for configurations.

https://pre-commit.com/
https://github.com/psf/black


Corvince:
Thank you, adding this is a wonderful idea and something I have been doing for a while now and had always wanted to add a similar PR.

But I think this needs a bit of clarifying for those unaware of these tools, especially since changing 88 files and ~3000 lines of code is *no* small PR ;-) But I see what you meant by that. 

[black](https://github.com/psf/black) is "the uncompromising code formatter." Since I have been using black I really stopped worrying or even thinking about code formating in Python. I just write whatever I want in whatever style I want (not thinking about indentation below parenthesis for example). I set up by editor to then run black automatically on save. And it will always look the same, for every project that uses black since it has (almost) not configuration possibilities. It is just great and a

SebastianoF:
It is no small PR, but not man made too. I just run black, so it is only text reformatting and there are no worries about possible mistakes.

If you want I can add some documentation in the readme or a link pointing at the wiki page of the github webpage, with even more info.

Corvince:
Yes I think you should add the black badge to the readme and write a short sentence in the contributions file that Mesa uses the black code style and some short instruction on how to set up and use pre-commit.

Maybe we should even use `black --check` on Travis

SebastianoF:
Thanks for the moral support. It seems black is not supported for python 3.5 (although it formats down to python 2.7, https://github.com/psf/black/issues/519). This is a problem for continuous integration. Any idea about how to exclude black when running Travis on python 3.5?

Corvince:
Thanks for fixing my half-baked solution to run black only on 3.6. 

I made some small edits to readme and contributing. I am pro-merge now, but as I said want to wait for some second opinion by @jackiekazil or @dmasad (or someone else)

SebastianoF:
All good for me, my seat belt is fasten and ready to land!

rht:
I'm in favor of using Black ASAP. Though it is kind of tedious that this PR has to be redone for every merge conflict with master.

SebastianoF:
It is enough to run pre-commit on this branch before merging it. It is up to you, as I do not know when/if you plan to land it!

rht:
@SebastianoF because of the conflicts, can you redo the applying the Black to the repo from master and the force-push to this PR?

SebastianoF:
Done! Let's see if I had messed up anything :smile:

SebastianoF:
... it seems not!

rht:
@SebastianoF you didn't redo the commits. You added commits which fix the early conflicts, but the conflicts are still there in the early commits. It is up to @jackiekazil / @dmasad to merge though.

SebastianoF:
Six months of maintenance while deciding if merging or not is quite a long time.
Closing this PR now, but please feel free to reuse `.pre-commit-config.yaml` and `.travis.yml` if intend to use black in the future.

rht:
Why not keep it open, to maintain visibility?

Corvince:
Since #775 was now merged I don't see any reason not to merge this one again, so I re-opened it. 

@dmasad since you merged #775, do you see any reason not to merge?

Corvince:
@SebastianoF I hope you don't mind me re-opening your PR. I'll take over any responsibility for your changes/feedback and will continue to update this PR if it gets outdated

SebastianoF:
> @SebastianoF I hope you don't mind me re-opening your PR. I'll take over any responsibility for your changes/feedback and will continue to update this PR if it gets outdated

Please do! Would be great to continue this work.

.travis.yml
```diff
@@ -13,13 +13,17 @@ python:
...
   - pip install codecov
   - pip install .[dev]
+  - if [[ $TRAVIS_PYTHON_VERSION == 3.6 ]]; then pip install -U black; fi
 script:
   - export PYTHONPATH=$PYTHONPATH:$(pwd)
   # * E501 - line length limit
   # * E123 - indentation on data structures
   # * W504 - line break after binary operator
-  - flake8 . --ignore=F403,E501,E123,E128,W504 --exclude=docs,build
+  - flake8 . --ignore=F403,E501,E123,E128,W504,W503 --exclude=docs,build
```
jackiekazil, line 22: Why was this added in particular?

.travis.yml
```diff
@@ -13,13 +13,17 @@ python:
 install:
   - pip install codecov
   - pip install .[dev]
+  - if [[ $TRAVIS_PYTHON_VERSION == 3.6 ]]; then pip install -U black; fi
```
jackiekazil, line 16: Is there a reason that this is only explicitly for Python 3.6 and not >= 3.6?

jackiekazil:
Sorry for the delay on this. Sometimes it is hard to keep up. @SebastianoF ty for submitting. I was going to put this together, but you beat me to it. @Corvince thank you readopting. @rht thank you raising to keep it open.

.travis.yml
Corvince: The answer is more difficult than I thought, but the short version is: W503 is actually incompatible with pep8. Flake8 actually ignores W503 by default, but when you customize the ignore list it is re-enabled. By the way, with flake8's default settings the black formatted code would only trigger E501. But since we "need" to ignore that one we also have to ignore W503. That would actually already be enough at least for now, since black makes sure F403, E123, E128 and W504 are not triggered.
Or we could go with blacks own flake8 ignore list:
`flake8 . --ignore=E203,E266,E501,W503`

What do you think?

.travis.yml
Corvince: I think the idea was that black only needs to run once (since the formatted code is independent of the python version). But just to make sure nothing breaks once we deprecate 3.6 it is probably safer to change it to `>=3.6`. I will make that change once we have settled the above comment

.travis.yml
jackiekazil: The only one that I feel truly passionate about is E501, which is in both lists. 
I could go either way on this. 

If we stay with the current list then we have to add that and continue to support it. It is also not as clean cut as just using black. 

If we move to black's list, then we might have to do some clean up. I think all of it would be magic with Black and easy clean up except for F403. 

I am leaning towards black's list, but could be talked into the other. Someone would need to make sure that the F403 error was cleaned up. 

The other option is to manually force blacks list with the addition of F403 & then put F403 as a TODO item that is in a separate PR/Ticket. 

So, I punt back to you. How do you want to move forward?

jackiekazil: Also, let me save you the trouble... https://www.flake8rules.com/

Corvince:
@jackiekazil 
Re Flake8: I opted for blacks exclude list. Your concern with F403 is actually a non-issue, since F403 is never triggered. I now also created a seperate .flake8 config file, since I think this is a much cleaner approach. Previously I always had do look up the ignore list from travis, now everyone can just run `flake8` and be happy (and linters should also pick this up).

Re travis: Travis does not allow `>=3.6` syntax, so I changed that to `!=3.5`. I also removed the `sudo: false` statement, since travis was nagging me that `sudo` is deprecated.

jackiekazil:
@Corvince LGTM. I am onboard with merging this. Did you want me to or did you want to? (I _think_ it is ready, but didn't want to assume.

Corvince:
Ah, forget to add my opinion on that: given that this is such a large change I thought maybe it is better to first release a new Mesa version before merging this. I mean the changes should only be cosmetic and all checks are passing, but... I guess you never know. 

But feel free to merge anytime you feel it is right. You are more experienced with releases.

jackiekazil:
@Corvince that is a good point. Working on #819 tonight to push this forward. Going to mark this as "blocked" until #819 is pushed thru.

PR #736, opened 2019-09-26
Fixup citation links in docs tutorial section

mwiens91:
Commit citations in the docs weren't using reST's citation support (https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#citations). Using these makes the docs a bit more user friendly.


Corvince:
Thanks!

Issue #749, opened 2019-10-24
Virus on Network example incompatible with networkx >= 2.4


SebastianoF:
Hello!

After raising the PR #748 ,
https://travis-ci.org/projectmesa/mesa?utm_medium=notification&utm_source=github_status
Travis fails with the error in object, although only the documentation have changed from the previous version. 

Tests are passed in local.

It seems the current master would not pass the tests on travis (?).

Possibly also PR #747 affected by the same issue.

---

Edit: I was wrong **it fails in local** too from a fresh virtualenv. The problem seems to be located in `setup.py`.

SebastianoF:
Problem addressed in PR above. :bear: :bear: :bear: the master version is currently broken due to networkx upadte to version 2.4.

To avoid having the master again broken on master after third parties update, pinning dependencies or reconsidering the use of Pipenv should be the next task.

Corvince:
This is currently fixed with setting the requirement for networkx to <2.4

However, I encourage anyone to rather update the example and unpin the requirement. 

Since mesa is a modeling library I think it is a *bad* idea to pin dependencies. A minimum version would be okay, but we shouldn't prevent users from updating there packages and try our best to always work with the latest versions of our dependencies.

In this particular case we should have seen the depreciation warnings long ago on travis. However, with the way the examples are run the warnings don't seem to passed on to py.test. To make things clear: Mesa **does** work with `networkx 2.4`, it is just the example that is failing. I think this is another reason to split out the examples from the main repository.

rht:
> To make things clear: Mesa does work with networkx 2.4, it is just the example that is failing. I think this is another reason to split out the examples from the main repository.

The error happens at mesa/space.py,

>  File "/home/rht/code/venv/lib/python3.7/site-packages/mesa/space.py", line 810, in _place_agent
>    self.G.node[node_id]['agent'].append(agent)
> AttributeError: 'Graph' object has no attribute 'node'

So it is not just the virus example that is failing, it is an error in the Mesa library.

PR #750, opened 2019-10-24
Issue #749 networkx incompatibility

SebastianoF:
The 17th of October release of networkx broke the tests on master:
https://networkx.github.io/documentation/stable/index.html

This is a quick fix, pinning the library to a lower version.

It seems that despite Pipenv libraries must be pinned anyway. Possibly the import of all the libraries must be reviewed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ __pycache__/
 # Distribution / packaging
 .Python
 env/
+venv/
 build/
 develop-eggs/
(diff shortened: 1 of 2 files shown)


Corvince:
Thanks for the quick fix and sorry for the slow response.

rht:
What is the breaking change caused by the newer version of networkx? Maybe we should fix that.

Issue #756, opened 2019-11-26
Error in runing shape_example 


YasinEnigma:
**Describe the bug**

When I run shape_example encountered with this error.
from shape_example.server import launch_shape_model
ImportError: cannot import name 'launch_shape_model'

**Expected behavior**

**To Reproduce**

**Additional context**

tpike3:
At the end of the the server.py file (~ line 44) add `server.launch()`

You may also need to change line 4 of the server.py file to : 

`from shape_example.model import Walker, ShapeExample` if you get `No module named '__main__.shape'; '__main__' is not a package` error

Thanks for letting us know and remember the example really just shows what arrowheads look like with a visualization. 

Let me know if this doesn't work

YasinEnigma:
Thanks, problem solved by adding `server.launch` to server.py file and create an __init__.py to shape_example directory (to create a python package and use it in run.py).

> At the end of the the server.py file (~ line 44) add `server.launch()`
> 
> You may also need to change line 4 of the server.py file to :
> 
> `from shape_example.model import Walker, ShapeExample` if you get `No module named '__main__.shape'; '__main__' is not a package` error
> 
> Thanks for letting us know and remember the example really just shows what arrowheads look like with a visualization.
> 
> Let me know if this doesn't work

Thanks, problem solved by adding `server.launch()` to server.py file and create an __init__.py file in shape_example directory (to create a python package and use it in run.py).

PR #757, opened 2019-11-27
Update server.py

YasinEnigma:
Add server.launch() to solve #756 issue.

### DIFF
--- a/examples/shape_example/shape_example/server.py
+++ b/examples/shape_example/shape_example/server.py
@@ -41,3 +41,4 @@ server = ModularServer(
 )
 server.max_steps = 0
 server.port = 8521
+server.launch()


rht:
LGTM

Note: It's probably worth to scan through the rest of the examples for a missing `server.launch()` as well.

Corvince:
I think all other examples put the `server.launch()` inside the `run.py` file and this is also how `test_examples.py` "expects" it to be. Putting `server.launch()` inside the server file breaks those tests.

YasinEnigma:
> I think all other examples put the `server.launch()` inside the `run.py` file and this is also how `test_examples.py` "expects" it to be. Putting `server.launch()` inside the server file breaks those tests.

if change run.py file to: 
`from shape_example.server import server
server.launch()
` 
the problem have work. Do I need to make a new Pull Reques?

PR #758, opened 2019-11-27
Create __init__.py

YasinEnigma:
create __init__.py to solve #756 issue.


rht:
LGTM

Also look out if other examples have missing `__init__.py`, to solve the issue once for all.

rht:
Actually, a lot of the examples don't have `__init__.py`, so I think another solution other than `__init__.py` should be used.

rht:
@YasinEnigma I was able to run `shape_example` after adding `server.launch()`. Is the `__init__.py` fix really necessary?

YasinEnigma:
> @YasinEnigma I was able to run `shape_example`'s `run.py` after adding `server.launch()` fix. Is the `__init__.py` fix really necessary?



No, it's not. I tried again, you are right. only adding `server.launch()` solve the problem.

rht:
@YasinEnigma you closed the wrong PR.

PR #762, opened 2019-12-04
Update run.py

YasinEnigma:
Update run.py to solve #756 issue.

### DIFF
--- a/examples/shape_example/run.py
+++ b/examples/shape_example/run.py
@@ -1,3 +1,3 @@
-from shape_example.server import launch_shape_model
+from shape_example.server import server 
 
-launch_shape_model()
+server.launch()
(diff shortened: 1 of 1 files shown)


rht:
There is Travis error due to linting: trailing whitespace: https://travis-ci.org/projectmesa/mesa/jobs/620792741#L312.

YasinEnigma:
> There is Travis error due to linting: trailing whitespace: https://travis-ci.org/projectmesa/mesa/jobs/620792741#L312.

I don't know what the problem is that doesn't confirm :(

examples/shape_example/run.py
```diff
@@ -1,3 +1,3 @@
-from shape_example.server import launch_shape_model
+from shape_example.server import server
```
rht: There is a trailing whitespace here.

Corvince:
Thanks!

examples/shape_example/run.py
YasinEnigma: Oh, Thanks.

PR #763, opened 2019-12-04
NetworkGrid: Update to networkx 2.4 API

### LINKED ISSUE #749: Virus on Network example incompatible with networkx >= 2.4

rht:
This fixes #749 
This is based on networkx 2.4 announcement changelog:
> G.node –> use G.nodes

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ from codecs import open
 requires = [
     'click',
     'cookiecutter',
-    'networkx < 2.4',
+    'networkx',
     'numpy',
     'pandas',
     'tornado',
--- a/examples/virus_on_network/virus_on_network/server.py
+++ b/examples/virus_on_network/virus_on_network/server.py
@@ -28,7 +28,7 @@ def network_portrayal(G):
         return 2
 
     def get_agents(source, target):
-        return G.node[source]['agent'][0], G.node[target]['agent'][0]
+        return G.nodes[source]['agent'][0], G.nodes[target]['agent'][0]
 
     portrayal = dict()
     portrayal['nodes'] = [{'size': 6,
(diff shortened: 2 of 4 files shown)


rht:
The changelog itself: https://networkx.github.io/documentation/stable/release/release_2.4.html

setup.py
```diff
@@ -8,7 +8,7 @@
 requires = [
     'click',
     'cookiecutter',
-    'networkx < 2.4',
+    'networkx >= 2.4',
```
rht: `G.nodes` has existed on the previous version, so this constraint is actually not necessary. Let me know if the version constraint needs to be removed.

Corvince: I vaguely remember that the API change occured with 2.0 and 'G.nodes' exists since then. Could you check that and then change the required version accordingly?

rht: `G.nodes` has existed since 1.0: https://networkx.github.io/documentation/networkx-1.8/reference/api_1.0.html#nodes-and-nodes-iter.

rht: *at least since 1.0, and could be earlier.

rht: I removed the version constraint because 1.0 is a very old release, and the majority of anyone using mesa would use a version that is much later.

rht: @Corvince any update?

Corvince: Yes, that sounds very good! `G.nodes` changed its return type and `G.node` was depreciated since 2.0, that is properly what I remembered. But I totally agree that we should just leave the version constraint.

PR #764, opened 2019-12-05
networkx node list comes from 'G.nodes' rather than 'G.node'

jdfoote:
This fixes what appears to be a small typo bug in the virus_on_network example, where the node list was being called incorrectly.


rht:
The change in this PR is a subset of the change in #763.

Corvince:
Thank you for the effort, but as @rht said, changed are already applied with #763

jdfoote:
Oops - should have checked! :)

Issue #719, opened 2019-08-07
Mesa Graph Visualisation on server add edges


WesselDeZeeuw:
**What's the problem this feature will solve?**
Hi all, I am quite new to GitHub, and quite new to Mesa so I am not sure if this is the correct way of asking you guys a question. If it is not, I apologize. However, that still doesn't change the fact that I am dying to use Mesa for a project where we are trying to model network structures. We want to visualize these using the Mesa Server. So as a start I have used 'Virus on a link' to start with. Now, a significant change in my model is that the connections / edges between agents can change each time step. This is also the part where my visualization messes up. For some reason, each time a add an edge, my Mesa server draws the _entire_ graph again, instead of only the new edge. This makes my graph unusable. I am wondering how to fix this, but I couldn't find any information online. I am thinking that it's something in my portrayal code but I don't have enough experience with it just yet.

 Could you guys help me out?

**Describe the solution you'd like**
I would like a solution to visualize changing graphs without having them duplicated in my Server window.

**Additional context**
I have added two visualizations of the start of the model, and after 3 iterations so that you can see what's happening. If needed, I am very willing to share the code!

Thanks in advance! :-) 

<img width="446" alt="mesastart" src="https://user-images.githubusercontent.com/42937749/62601188-132f3200-b8f1-11e9-911b-a9af004caa85.png">

<img width="449" alt="mesaafter3it" src="https://user-images.githubusercontent.com/42937749/62601187-132f3200-b8f1-11e9-8039-cc649f509696.png">

Corvince:
Hi there and welcome to GitHub ;)

The main place to ask questions about mesa and help is actually the mailing list. That said, your problem might be a bug, in which case GitHub would indeed be the right place. 
However, it is quite difficult to say without any code. This might be a bug in the network visualization or a bug in your model code, where you create multiple graphs rather then modifying your graph, in which case the visualization would be rather useless, but correct. 
I would start debugging by changing the visualization backend from `d3` to `sigma`. For the virus-on-network example this would be [this line in `server.py`](https://github.com/projectmesa/mesa/blob/0e73e395bd4bd8ff9d0e1383bdf8ef9b82478f34/examples/virus_on_network/virus_on_network/server.py#L50)
```python
network = NetworkModule(network_portrayal, 500, 500, library='sigma')
```
If this doesn't help please try to come up with a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) and we can have a look at whats going on.

[...]

Issue #765, opened 2019-12-05
D3 Network Visualization overwrites rather than updating


jdfoote:
**Describe the bug**
At every tick the Network Visualization receives a new set of nodes and edges but doesn't remove the previous nodes and edges.

**Expected behavior**
The previous network should be updated rather than appended to.

**To Reproduce**
The virus_on_network example shows this behavior. If you inspect the page you will see a proliferation of <g> elements.

**Additional context**

I believe that the problem is in NetworkModule_d3.js and is likely very simple to fix. I just don't know enough JS to be the one to fix it.

tpike3:
Fix merged #767

PR #767, opened 2019-12-10
Fix issue 765

JamesHovet:
see #765 

The D3 network visualization was creating a whole new set of objects
each render and stacking it on top of the previous. This was wasteful,
unperformant, and poor D3.js style.

This change restructures that D3 code to take advantage of D3's update
model. Instead of creating new objects every render, the code now edits
existing objects, adding and removing as needed. (This is the intended
way to write D3.js code and the library makes this pattern very easy.)

This change also fixes a possible future issue where adding multiple svg
elements to a viz would break this one.


JamesHovet:
Looking back through the list of github issues, I believe this will also fix issue #719

Corvince:
I only know the very basics of d3 so I can not really judge your changes, but it seems to be fine. Just a question: Between lines 49 and 69 you repeatedly call 
```javascript
        links
            .selectAll("line")
            .data(graph.edges)```

Is there are reason not just to do everything after the first selection? (similarly with nodes) Just trying to understand d3

Issue #769, opened 2019-12-15
Parameter order when calling SingleGrid in examples/shelling/model.py is the wrong way round


Mat-pa:
Instead of SingleGrid(height, width, torus=True), it should be SingleGrid(width, height, torus=True)

rht:
@dmasad @Corvince this issue has been solved.

PR #770, opened 2019-12-15
 #769: Fix parameter order in initializing SingleGrid object

Mat-pa:
Parameter order should be width, height, torus according to https://mesa.readthedocs.io/en/master/apis/space.html

### DIFF
--- a/examples/pd_grid/pd_grid/model.py
+++ b/examples/pd_grid/pd_grid/model.py
@@ -31,7 +31,7 @@ class PdGrid(Model):
                            Determines the agent activation regime.
             payoffs: (optional) Dictionary of (move, neighbor_move) payoffs.
         '''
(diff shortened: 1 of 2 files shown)


rht:
Nice catch. I grepped for the same pattern and found the mistake in one more file:
```
$ rg 'singlegrid\(height'
examples/schelling/model.py
53:        self.grid = SingleGrid(height, width, torus=True)

examples/pd_grid/pd_grid/model.py
34:        self.grid = SingleGrid(height, width, torus=True)
```

@Mat-pa can you update the PR to include the fix for pd_grid as well?

Mat-pa:
@rht Sure, done :)

rht:
LGTM. cc: @Corvince

dmasad:
Looks good, thanks!

Issue #771, opened 2019-12-22
Broken link to pipenv docs


josiahvinson:
**Describe the bug**
The link to `pipenv` docs in the intro tutorial is broken.

**Expected behavior**
I expect clicking the link to `pipenv` to reference a valid url.

**To Reproduce**
Click on the `pipenv` link on https://mesa.readthedocs.io/en/master/tutorials/intro_tutorial.html.

**Additional context**

Corvince:
Thanks for the report and fix!

PR #772, opened 2019-12-22
Update pipenv link

josiahvinson:
Fixed broken link to `pipenv` docs, as mentioned in #771


josiahvinson:
Not sure how changing one line of ignored code would drop test coverage.

PR #775, opened 2020-01-08
Apply black to mesa/ directory

rht:
I separate this from #732. So that this PR can be redone whenever there is merge conflict.

### DIFF
--- a/mesa/batchrunner.py
+++ b/mesa/batchrunner.py
@@ -22,8 +22,10 @@ else:
 
 
 class ParameterError(TypeError):
-    MESSAGE = ('parameters must map a name to a value. '
-               'These names did not match paramerets: {}')
+    MESSAGE = (
+        "parameters must map a name to a value. "
+        "These names did not match paramerets: {}"
+    )
 
     def __init__(self, bad_names):
         self.bad_names = bad_names
@@ -33,8 +35,10 @@ class ParameterError(TypeError):
 
 
 class VariableParameterError(ParameterError):
-    MESSAGE = ('variable_parameters must map a name to a sequence of values. '
-               'These parameters were given with non-sequence values: {}')
+    MESSAGE = (
+        "variable_parameters must map a name to a sequence of values. "
+        "These parameters were given with non-sequence values: {}"
+    )
 
     def __init__(self, bad_names):
         super().__init__(bad_names)
@@ -51,10 +55,18 @@ class FixedBatchRunner:
     run. To get step by step data, simply have a reporter store the model's
     entire DataCollector object.
     """
-    def __init__(self, model_cls, parameters_list=None,
-                 fixed_parameters=None, iterations=1, max_steps=1000,
-                 model_reporters=None, agent_reporters=None,
-                 display_progress=True):
+
+    def __init__(
+        self,
+        model_cls,
+        parameters_list=None,
+        fixed_parameters=None,
+        iterations=1,
+        max_steps=1000,
+        model_reporters=None,
+        agent_reporters=None,
+        display_progress=True,
+    ):
         """ Create a new BatchRunner for a given model with the given
         parameters.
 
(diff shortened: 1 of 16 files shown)


rht:
Installation failed in the nightly build. Since the numpy dependency is not tested on nightly anyway, I think the nightly build should just be removed.

jackiekazil:
I totally agree that black needs to be run. Maybe even added to the ci/cd (don't think it is there). However, when you run black on a whole repo, that could end up making all other PRs come in conflict. 

Thinking out loud... 
- Option 1: Do this now and fix conflicts later. 
- Option 2: One file or two at a time as we work on things. 
- Option 3: 3 month sprint to clean out existing PRs --  ambitious, but that would minimize impact. Then do a black cleanup. ... who need everyone familiar with the project on board to get this done. 
- Other option? 

I am truly on the fence about this. 

@rht - any other approaches that you can think of?
@dmasad? @Corvince? How do you all feel?

mesa/batchrunner.py
```diff
@@ -22,8 +22,10 @@
 
 
 class ParameterError(TypeError):
-    MESSAGE = ('parameters must map a name to a value. '
-               'These names did not match paramerets: {}')
+    MESSAGE = (
+        "parameters must map a name to a value. "
+        "These names did not match paramerets: {}"
```
dmasad: Minor typo here

dmasad:
I'd lean towards either 2 or 3 -- basically, not to leave conflicts hanging around. 

In terms of conflicts involving incoming PRs -- should running Black on those tend to resolve most of the issues?

jackiekazil:
RE: conflicts on incoming PRs -- Theoretically.

rht:
IMO since most of the changes by Black are trivial, like changing quotation, the merge conflicts will be easy to resolve by hand, or perhaps nonexistent. In case a contributor is MIA when pinged for update, I can make a substitute PR containing the commit from the said contributor.

Corvince:
> I totally agree that black needs to be run. Maybe even added to the ci/cd (don't think it is there). 

First note that the sister PR #732 includes the addition of pre-commit to the local build line and a travis check for code style. The former would make sure that at least any new commits/PRs adhere to the black code style. Problem is we can't enforce people to use pre-commit. And the latter always checks the whole repo, so it is only possible if we blacken the whole repo in the first place. 

> However, when you run black on a whole repo, that could end up making all other PRs come in conflict.

This is certainly true and I did not fully think of that. However, the whole point of black is an unambiguous code style, so I would guess (most?) merge conflicts would be resolved by running black on the PRs or rebasing them. At least, that would be my impression.

In conclusion I am in favor of merging #732 after solving its merge conflicts (which I would do if thats the way forward)

rht:
What I have in mind in this PR is to always redo the Black whenever needed, and so #732 is left for adding Black to CI and pre-commit. I am used to redoing PR and force-pushing so that the commit history remains clean. But if @SebastianoF can force-push #732 to keep the commit history clean, then we can merge #732 in its entirety. Either way works.

SebastianoF:
Please see #732 where conflicts had been solved and history is linear.

rht:
Commit history in #732 contains past commits where the merge conflict still exists, and so is not clean.

Corvince:
I would prefer this issue not to depart into a discussion about the best strategy to deal with PRs (merging, squash-and-merge or prefer force-pushes).

Just to keep everyone on board: there is nothing wrong with either @SebastianoF branch nor this one. 

(Just commenting because I thought @rht comment might sound a bit confusing for people not familiar with git, as it sounds like there would be something wrong with the commit history)

mesa/batchrunner.py
rht: #786 can't fix right away because the port for git is blocked in my current internet.

rht:
I have remade the PR by doing it on master post-Mypy annotation of space.py.

rht:
PR remade once again.

PR #777, opened 2020-01-13
Updated travis to 3.8 and updated gitignore

Corvince:
Travis seems to be failing again (see #775), so i switched from nightly to 3.8.

Also a small update to gitignore and deletion of some cache files that somehow slipped onto GitHub.


rht:
LGTM

PR #779, opened 2020-01-13
space: Add type annotation to Grid class

rht:
This is an update to #394. This is as precise as it can get.
The thing up for debate is whether the content of a cell should be `Any` or `object`.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -13,6 +13,7 @@ class Agent:
         """ Create a new agent. """
         self.unique_id = unique_id
         self.model = model
+        self.pos = None
 
     def step(self):
         """ A single step of the agent. """
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -18,9 +18,10 @@ import itertools
 
 import numpy as np
 
-from typing import Any, Iterable, Iterator, List, Set, Tuple, Union
+from typing import Iterable, Iterator, List, Optional, Set, Tuple, Union
 from .agent import Agent
 Coordinate = Tuple[int, int]
+GridContent = Union[Optional[Agent], Set[Agent]]
 
 
 def accept_tuple_argument(wrapped_function):
@@ -86,10 +87,10 @@ class Grid:
         self.width = width
         self.torus = torus
 
-        self.grid = []  # type: List[List[Any]]
+        self.grid = []  # type: List[List[GridContent]]
 
         for x in range(self.width):
-            col = []  # type: List[Any]
+            col = []  # type: List[GridContent]
             for y in range(self.height):
                 col.append(self.default_val())
             self.grid.append(col)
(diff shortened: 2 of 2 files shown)


rht:
Wait, since the default value of a cell in `Grid` is always `None`, I can just call the content of a cell to be `Optional[Agent]`.

rht:
I have removed the usage of `Any`.

rht:
On second thought, I am reverting back the usage of `GridContent` because the child class `MultiGrid` has different type of `GridContent` that is not `Optional[Agent]`, but rather, `Optional[Set[Agent]]`.

Corvince:
Great work! I'll do a code review soonish (on mobile right now), but it looks very good!

Regarding the GridContent: I would prefer an explicit type in the form of 
`GridContent = Union[Optional[Agent], Optional[Set[Agent]]`

I don't think this is not optimal, but shows a good advantage of using these type hints: Multigrid actually messes up the type of it's parent class which is not a good thing and potentially confusing. You can't easily switch between Grid and MultiGrid. This should eventually be sorted out, but for this PR we should just document the current state. And by using a GridContent type alias we can later just change the allowed content. But currently it should be just Agents/Set* of Agents and nothing else.

* It was mentioned somewhere else that set is actually a poor choice, since its order is non deterministic and doesn't allow for reproducible model runs

rht:
Added `GridContent` in the second commit.

> * It was mentioned somewhere else that set is actually a poor choice, since its order is non deterministic and doesn't allow for reproducible model runs

Should `set` be replaced with `OrderedDict`?

rht:
> Should set be replaced with OrderedDict?

Actually since Python 3.6, using `dict` is sufficient (see https://stackoverflow.com/questions/14956313/why-is-dictionary-ordering-non-deterministic).

rht:
Raised #782.

mesa/space.py
```diff
@@ -82,10 +87,10 @@ def __init__(self, width, height, torus):
         self.width = width
         self.torus = torus
 
-        self.grid = []
+        self.grid = []  # type: List[List[GridContent]]
```
Corvince: What do you think of dropping support for Python 3.5 and using the newer type style like
`self.grid: List[List[GridContent]] = []`
Dropping 3.5 would also ease the use of black (well, at least for the build on travis)

mesa/space.py
```diff
@@ -243,7 +250,7 @@ def get_neighbors(self, pos, moore,
         return list(self.iter_neighbors(
             pos, moore, include_center, radius))
 
-    def torus_adj(self, pos):
+    def torus_adj(self, pos: Coordinate) -> Tuple[int, int]:
```
Corvince: Return type should be Coordinate as well

mesa/space.py
```diff
@@ -339,7 +346,7 @@ def move_to_empty(self, agent):
         agent.pos = new_pos
         self._remove_agent(pos, agent)
 
-    def find_empty(self):
+    def find_empty(self) -> Union[Coordinate, None]:
```
Corvince: Could be `Optional[Coordinate]`

mesa/agent.py
```diff
@@ -13,6 +13,7 @@ def __init__(self, unique_id, model):
         """ Create a new agent. """
         self.unique_id = unique_id
         self.model = model
+        self.pos = None
```
Corvince: Is this required to make mypy happy? Otherwise I am not sure how I feel about this regarding models without any spatial dimension

mesa/space.py
rht: Fixed.

mesa/space.py
rht: Fixed.

mesa/agent.py
rht: Yes, otherwise mypy complains that an `Agent` doesn't have a `pos` attribute.

mesa/space.py
rht: Python 3.5 is still used in Xenial though, and might break existing code. I don't have strong opinion on whether to drop 3.5 or not.

mesa/agent.py
rht: I think it is harmless to have a `pos` attribute in the Agent specification, given that it is optional (i.e. has a default value of `None`). Otherwise using `pos` without specifying it would be too ad-hoc.

Corvince:
I am in favor of merging this, but given the previous discussions on adding type annotations I'll wait for a second opinion (preferably by @dmasad or @jackiekazil) to chime in before merging this

dmasad:
This looks like a reasonable, and fairly compact, starting point. I don't have a great sense of how it's going to interact with the Black auto-formatting, but I think it's worth integrating this and figuring out how to make it work.

PR #781, opened 2020-01-14
Use dict to store MultiGrid cell content

rht:
This will at least make MultiGrid deterministic for Python >=3.6.


rht:
Nvm, this doesn't work because `Grid`'s __getitem__ and __iter__ expect a list-like object, which set does.

PR #783, opened 2020-01-16
MultiGrid: Set to using list for cell content

### LINKED ISSUE #782: The order of the cell content of MultiGrid is non-deterministic
This is because the cell content is a Python `set`, and `set` doesn't preserve insertion order.
It is possible to use `dict` for Python >=3.6 instead, but this requires a lot of redefinition of the methods `__iter__` and `__getitem__`.

rht:
Fixes #782.


Corvince:
Thanks for the PR. Sorry it hadn't got merged earlier.

PR #792, opened 2020-02-02
Add Mypy annotation to time, agent, and model

rht:


mesa/agent.py
```diff
@@ -5,20 +5,27 @@
...
+    def step(self) -> None:
         """ A single step of the agent. """
         pass
 
+    def advance(self) -> None:
+        pass
+
     @property
-    def random(self):
+    def random(self) -> Any:
```
rht: I tried setting this method to return `random.Random`, but it said `random.Random` is undefined, so I set this to `Any` instead.

mesa/agent.py
```diff
@@ -5,20 +5,27 @@
...
         self.unique_id = unique_id
         self.model = model
         self.pos = None
 
-    def step(self):
+    def step(self) -> None:
         """ A single step of the agent. """
         pass
 
+    def advance(self) -> None:
```
rht: This needs a docstring, but I can't figure what to write here.

Corvince:
Hey there,

was looking and working on your branch. Some suggestions can be seen on [this PR](https://github.com/rht/mesa/pull/3) to your PR. 

Also to keep the discussion on this thread:

Question: You had some code blocks like this

if False:
    from .model import Model
I know this is for circular references and I remember having those as well when working with mypy earlier, but I just changed them to regular imports and it seems to work. Can you maybe clarify under what circumstances this might break?

Also note that it is now recommended to use

from typing import TYPE_CHECKING
if TYPE_CHECKING: 
    ...

rht:
> Can you maybe clarify under what circumstances this might break?

I just wanted the runtime to be not polluted with cross-imports of mostly abstract base classes / "interfaces".
But if it is written explicitly that the reason why agent.py imports `Model` is due to typing, and so the intent is isolated, I think it should be fine.

> from typing import TYPE_CHECKING
> if TYPE_CHECKING:

Yeah, right I was aware of this. I used "if False" out of muscle memory. But we are not going to use this anyway since we are doing explicit imports of `Model`.

Your suggestion LGTM, I will incorporate into this PR

PR #802, opened 2020-03-11
Fixed pip install from github by specifying egg

napsternxg:
Added `#egg=mesa`


rht:
Surprised that this documentation problem has been lurking for so long. LGTM.

PR #804, opened 2020-03-21
Fixes #803 Advanced Tutorial #elements

### LINKED ISSUE #803: Advanced Tutorial #elements
**Describe the bug**
In the `HistogramModule` JS, the Advanced Tutorial currently appends the `canvas` to `body`, instead of the `#elements` div. This means that the histogram element does not behave like the other visualization elements: it is always at the bottom of the page (i.e., it does not respect the order of the elements passed into `ModularServer`) and fills us the width of the page instead of just the div.

**Expected behavior**
The histogram should be located above the chart and have the same width.

**To Reproduce**
Follow the steps in the current version of the [Advanced Tutorial](https://mesa.readthedocs.io/en/master/tutorials/adv_tutorial.html#client-side-code), specifically, this JavaScript:
```
// Append it to body:
var canvas = $(canvas_tag)[0];
$("body").append(canvas);
```

gzgracez:
Fixes #803 by changing the docs to reflect that `canvas` should be added to the `#elements` div.

### DIFF
--- a/docs/tutorials/adv_tutorial.ipynb
+++ b/docs/tutorials/adv_tutorial.ipynb
@@ -257,9 +257,9 @@
     "    // Create the tag:\n",
     "    var canvas_tag = \"<canvas width='\" + canvas_width + \"' height='\" + canvas_height + \"' \";\n",
     "    canvas_tag += \"style='border:1px dotted'></canvas>\";\n",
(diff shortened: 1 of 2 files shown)


dmasad:
Thanks, that looks great!

PR #810, opened 2020-03-29
tutorial fix recommendations; Thanks Baumgartner

tpike3:
Fixes as follows based on recommendation: 

How do users get the requirements.txt? There's instructions on how to access it, but should they clone the repository, download the file from the example, or copy paste it into their own local file?

FIX: Change to pip install from .txt file url; Pip command then installs required dependencies directly from .txt at git url location  

MoneyModel.py - is the casing correct following coding standards for python files? The closest I could find is this: https://www.python.org/dev/peps/pep-0008/#package-and-module-names - which indicates it should be something like money_model.py.

FIX : Changed to money_model.py

Citation link is broken: Comer2014. URL: https://mesa.readthedocs.io/en/master/tutorials/intro_tutorial.html#comer2014 - the link goes to this, which is not found: http://gradworks.umi.com/36/23/3623940.html

FIX: Linked to George Mason MARs repository and directly to Comer. pdf link

"Every agent is expected to have a step method, which takes a model object as its only argument". The way this is worded is unclear. The step method doesn't have to take the model as an argument, it takes the agent it's defined on as self. I also think this part could use some better sequencing to build on a users knowledge. For example, the user defines a step function that doesn't do anything, then the model is executed. How is the user able to know whether the step function is working properly? Rather than build a step function that's just a pass, could they build one that prints the model id? Then for the exercise add a line to the step function that prints the wealth (which should be 1 at this point for all agents)? I think it's important to give the user feedback that the step function is working, but a step function that does nothing when the model is run doesn't help here.

FIX: Initial step method now prints agent id and exercise is to change print to agent wealth

In "Running your first model" at the phrase: "You’ll probably see something like the distribution shown below." the plot is actually above this part in the narrative.

FIX: Moved to above distribution  

The moore argument to get_neighborhood might be unclear what this does for new users. Should this be changed to diagonal? It requires them to understand the difference between Moore and Von Neuman, but a new user is more likely to understand "can they move diagonal or not?"
defining give_money - the other agent is now called other, but is called other_agent in the step method. It might help new users if other agents are defined consistently.
In "Adding Space", after the line "Now, putting that all together should look like this:" - the defintion order of MoneyModel and MoneyAgent is switched. Previously the agent was defined first, then the model. In this snippet, the Model is defined first, then the Agent. The remaining code snippets define Agent first.

FIX: Switch to MoneyAgent, MoneyModel sequencing. 

Under "Batch Run" - from mesa.batchrunner import BatchRunner gets imported, but not used in the next code block. The snippets make it unclear which file should be importing BatchRunner - the snippet right after the import is the model definition file, which is not where this should be imported. For clarity, this snippet with the import could wait until the block of code that executes the model.

FIX: Moved import cell to above batchrunner cell

In the batch runner example, we do not import compute_gini. This is related to the above point - compute_gini is defined in the model definition program, but not in the model execution program. At this point in the tutorial, it's not clear whether new code goes in the model definition script or the model execution script. When we use the batch runner and define the reporter to capture the Gini Coefficient, the execution script has not imported compute_gini from the model definition script.

FIX: Placed compute Gini in this section of the MoneyModel as well

Let me know if you have any issues.

### DIFF
--- a/docs/tutorials/intro_tutorial.ipynb
+++ b/docs/tutorials/intro_tutorial.ipynb
@@ -53,11 +53,17 @@
     "    $ pip install mesa\n",
     "```\n",
     "\n",
-    "When you do that, it will install Mesa itself, as well as any dependencies that aren't in your setup yet. Additional dependencies required by this tutorial can be found in the **examples/boltzmann_wealth_model/requirements.txt** file, which can be installed by running:\n",
+    "When you do that, it will install Mesa itself, as well as any dependencies that aren't in your setup yet. Additional dependencies required by this tutorial can be found in the **examples/boltzmann_wealth_model/requirements.txt** file, which can be installed directly form the github repository by running:\n",
     "\n",
     "```bash\n",
-    "    $ pip install -r requirements.txt\n",
-    "```"
+    "    $ pip install -r https://raw.githubusercontent.com/projectmesa/mesa/master/examples/boltzmann_wealth_model/requirements.txt  \n",
+    "```\n",
+    "\n",
+    "This will install the dependencies listed in the requirements.txt file which are:  \n",
+    " * jupyter (Ipython interactive notebook)  \n",
+    " * matplotlib (Python's visualization library)  \n",
+    " * mesa (this ABM library -- if not installed)  \n",
+    " * numpy (Python's numerical python library)  "
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


tpike3:
The sphinx build overwrote the .rst changes with the old build. I need to figure out why. Effectively the changes are only in the intro_tutorial.ipynb file.

PR #814, opened 2020-04-05
Travis: Add black checking

rht:


rht:
Because I have only applied `black` to `mesa/` directory, this check only applies to `mesa/`.

rht:
It appears that black doesn't support python 3.5: https://github.com/psf/black/blob/master/setup.py#L89.

Corvince:
Yes, See the workaround in #732

rht:
I changed the PR to be according to what is written in the diff of .travis.yml of #732.

Corvince:
I just reopened #732 since it 
  * includes the travis check (like this PR)
  * checks all mesa files including examples
  * adds pre-commit hooks (optional)
  * updates readme and contributing files to reflect the use of black 

Just want to check back with you @rht : Was there any specific reason to apply black only to the mesa folder? Am I missing some reason not the blacken all files here?

rht:
@Corvince I explained here: https://github.com/projectmesa/mesa/pull/814#issuecomment-609228488.

Corvince:
@rht I meant why you only applied black to the Mesa directory, not why the check only checks it

rht:
Because I didn't want to introduce too big of a change in a single PR, which if so, would have created too many moving parts.

PR #820, opened 2020-04-25
Advanced indexing of grid

Corvince:
Currently mesa only allows direct grid access via `grid[x][y]`

This PR advances the possibilities to the following forms:

1. `grid[x][y]`
1. `grid[x, y]`
2. `grid[:, :]`
4. `grid[(x1, y1), (x2, y2), ...]`

where (1) is the current way, (2) is a tuple of coordinates (the usual `pos` attribute of agents) and (3) is a sliced version of (2): It allows to quickly select a single column (`grid[1, :]`), row (`grid[:, 2]`) or any other rectangular shape/neighborhood (`grid[1:3, 2:7:2]`). 

(4) accepts an sequence of positions, so for example something like that would work:
```python
neighborhood = grid.get_neighborhood(pos)
content = grid[neighborhood]
```
The implementation differs from `get_cell_list_contents` in that it will always return a list of the same length as the input, whereas `get_cell_list_contents` only returns the content of non-empty cells. 

This PR is not associated with any issue, but aims to ease grid access.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -105,8 +105,48 @@ class Grid:
         """ Default value for new cell elements. """
         return None
 
-    def __getitem__(self, index: int) -> List[GridContent]:
-        return self.grid[index]
+    def __getitem__(
+        self,
+        index: Union[int, Tuple[int, int], Tuple[slice, slice], Tuple[Coordinate]],
+    ) -> Union[List[GridContent], GridContent]:
+
+        if isinstance(index, int):
+            # grid[x]
+            return self.grid[index]
+
+        if isinstance(index[0], tuple):
+            # grid[(x1, y1), (x2, y2)]
+            cells = []
+            for pos in index:
+                x, y = self.torus_adj(pos)
+                cells.append(self.grid[x][y])
+            return cells
+
+        x, y = index
+
+        if isinstance(x, int) and isinstance(y, int):
+            # grid[x, y]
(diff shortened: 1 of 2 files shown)


Corvince:
Travis currently fails, because of the @overload decorator used that makes flake8 cry. However it actually cries because of pyflakes, which is pinned to a version `<2.2.0` via flake8. This should be updated in the next version of flake8 (the [PR for that issue](https://gitlab.com/pycqa/flake8/-/merge_requests/417) is already merged)

dmasad:
Looks good to me - this is a great feature addition!

EwoutH:
It looks like this PR [broke the linting run](https://github.com/projectmesa/mesa/runs/2101962055?check_suite_focus=true#step:7:38):
```
Successfully built black
Installing collected packages: typing-extensions, typed-ast, regex, pathspec, mypy-extensions, appdirs, black
Successfully installed appdirs-1.4.4 black-20.8b1 mypy-extensions-0.4.3 pathspec-0.8.1 regex-2021.3.17 typed-ast-1.4.2 typing-extensions-3.7.4.3
would reformat /home/runner/work/mesa/mesa/mesa/space.py
would reformat /home/runner/work/mesa/mesa/tests/test_grid.py
Oh no! 💥 💔 💥
2 files would be reformatted, 114 files would be left unchanged.
```

PR #823, opened 2020-04-26
[PERF] Add neighborhood cache to grids and improve iter_cell_list_contents

Corvince:
This is a purely performance related issue. For a little more insights check #815, the take-away is that this potentially increases speed by a factor of at least 2. 

Two things are changed:
1. `get_neighborhood` now uses a cache to lookup already calculated neighborhoods, since they never change over the lifetime of a grid. Also the algorithm was slightly tweaked (with caching this doesn't matter but look out for it if you do a code review)

2. `iter_cell_list_contents` was modified. Currently it access the content of every cell twice: Once to check if the cell is empty and once to return the content. Content is now retrieved only once. 

`

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -247,7 +247,7 @@ class Grid:
             An iterator of non-None objects in the given neighborhood;
             at most 9 if Moore, 5 if Von-Neumann
             (8 and 4 if not including the center).
-
+g
         """
         neighborhood = self.get_neighborhood(pos, moore, include_center, radius)
         return self.iter_cell_list_contents(neighborhood)
@@ -258,8 +258,8 @@ class Grid:
         moore: bool,
         include_center: bool = False,
         radius: int = 1,
-    ) -> List[Coordinate]:
-        """Return a list of neighbors to a certain point.
+    ) -> List[GridContent]:
+        """ Return a list of neighbors to a certain point.
 
         Args:
             pos: Coordinate tuple for the neighborhood to get.
@@ -287,8 +287,7 @@ class Grid:
         elif not self.torus:
             raise Exception("Point out of bounds, and space non-toroidal.")
         else:
-            x, y = pos[0] % self.width, pos[1] % self.height
-        return x, y
+            return pos[0] % self.width, pos[1] % self.height
 
     def out_of_bounds(self, pos: Coordinate) -> bool:
         """
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -310,7 +310,10 @@ class Grid:
             An iterator of the contents of the cells identified in cell_list
 
         """
-        return (self[x][y] for x, y in cell_list if not self.is_cell_empty((x, y)))
+        for (x, y) in cell_list:
+            content = self.grid[x][y]
+            if not content == self.default_val():
+                yield content
 
     @accept_tuple_argument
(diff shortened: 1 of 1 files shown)


mesa/space.py
```diff
@@ -311,10 +311,7 @@ def iter_cell_list_contents(
             An iterator of the contents of the cells identified in cell_list
 
         """
-        for (x, y) in cell_list:
-            content = self.grid[x][y]
-            if not content == self.default_val():
-                yield content
+        return filter(None, (self.grid[x][y] for x, y in cell_list))
```
rht, line 314: `default_val()` is not always `None`, why the hardcode to `None`?

Corvince: This is not a hard-code to None. The filter function in Python is a bit unusual in the way that instead of a filter function you can use None to filter anything that bools to True (equivalent to `item for item in iterable if item`)

In this context `None, set() and list()` are all False.

Check the docs at [https://docs.python.org/3/library/functions.html#filter]()

rht:
I compared the Travis duration between this PR and #828. It appears that on 3.6 and 3.7, the test is 10s faster. Is it supposed to be noticeable in tests? If so, why is there no improvement on 3.8?

Corvince:
I don't think travis runtimes can be reliably used as a performance indicator. The performance gains are also insignificant if you run those functions only a couple of times (as done in the tests). The only difference could be observed for running the examples, but the test_examples actually does this: 
```python
(model.step() for _ in range(100))
```
which just creates a generator expression, but never runs the models. Just discovered this and will submit a PR in a couple minutes. When I change this line my local runtime jumps from 10 seconds to 90 seconds on master. On this branch it goes to 45 seconds. Let's see what travis thinks about this.

mesa/space.py
```diff
@@ -101,6 +101,9 @@ def __init__(self, width: int, height: int, torus: bool) -> None:
         # Add all cells to the empties list.
         self.empties = set(itertools.product(*(range(self.width), range(self.height))))
 
+        # Neighborhood Cache
+        self._neighborhood_cache = dict()  # type: Dict[Any, List[Coordinate]]
```
rht, line 105: Since Python 3.5 has been removed from support, we no longer have to use type comments.

rht: #984

mesa/space.py
```diff
@@ -252,7 +259,7 @@ def get_neighbors(
         moore: bool,
         include_center: bool = False,
         radius: int = 1,
-    ) -> List[Coordinate]:
+    ) -> List[GridContent]:
```
rht, line 262: I have also noticed this mis-annotation. Again, "neighborhood"/"neighbor" to denote either a neighbo coordinate or a content of a neighbor coordinate is confusing. This fix can be put in a separate PR so that it can be merged soon.

mesa/space.py
rht: I don't think `filter` will offer any speedup over list comprehension. Last time I tested, list comprehension was a bit faster even.

Corvince: This is not about speedup from filter vs list comprehension. The equivalent generator expression would be:
```python
(self[x][y] for x, y in cell_list if self[x][y])
```
However this would call/access `self[x][y]` twice, which makes this slower.

PR #829, opened 2020-05-09
Update test_examples.py

Corvince:
changed unused generator expression to actually run the models.

Originally 100 steps were proposed, but this increased runtime dramatically from abound 10 seconds to 90 seconds on my local machine. I think this would be quite an anoying wait time to test since the intend was only to see if the models actually run. So I changed it to 10 steps (resulting in 20 seconds test time locally).


rht:
LGTM

PR #835, opened 2020-05-11
space: more type annotations

Corvince:
Even more type annotations for space.py (after #779 and #796)

This PR makes mypy happy for grid, multigrid and hexgrid, except for some cases where the subclasses override the base class types (which is not solvable with type annotations, but would require changes to the logic itself).  

@rht since you did the previous type annotations, could you do a code review?

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -5,16 +5,18 @@ Core Objects: Agent
 
 """
 # mypy
-from typing import Optional, Union
-from mesa.model import Model
-from mesa.space import Coordinate, FloatCoordinate
+from typing import Optional, Union, TYPE_CHECKING
 from random import Random
+
+if TYPE_CHECKING:
+    from mesa.model import Model
+    from mesa.space import Coordinate, FloatCoordinate
 
 
 class Agent:
     """ Base class for a model agent. """
 
-    def __init__(self, unique_id: int, model: Model) -> None:
+    def __init__(self, unique_id: int, model: "Model") -> None:
         """ Create a new agent. """
         self.unique_id = unique_id
         self.model = model
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -5,7 +5,9 @@ Core Objects: Agent
 
 """
 # mypy
+from typing import Optional, Union
 from mesa.model import Model
+from mesa.space import Coordinate, FloatCoordinate
 from random import Random
 
 
@@ -16,7 +18,7 @@ class Agent:
         """ Create a new agent. """
         self.unique_id = unique_id
         self.model = model
-        self.pos = None
+        self.pos: Optional[Union[Coordinate, FloatCoordinate, int]] = None
 
     def step(self) -> None:
         """ A single step of the agent. """
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -5,12 +5,12 @@ Core Objects: Agent
 
 """
 # mypy
-from typing import Optional, Union, TYPE_CHECKING
+from typing import Optional, TYPE_CHECKING
 from random import Random
 
 if TYPE_CHECKING:
     from mesa.model import Model
-    from mesa.space import Coordinate, FloatCoordinate
+    from mesa.space import Position
 
 
 class Agent:
@@ -20,7 +20,7 @@ class Agent:
         """ Create a new agent. """
         self.unique_id = unique_id
         self.model = model
-        self.pos: Optional[Union[Coordinate, FloatCoordinate, int]] = None
+        self.pos = None  # type: Optional[Position]
 
     def step(self) -> None:
         """ A single step of the agent. """
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -21,6 +21,7 @@ from typing import (
     Any,
     Callable,
     Dict,
+    Generic,
     Iterable,
     Iterator,
     List,
@@ -34,12 +35,16 @@ from typing import (
 from mesa.agent import Agent
 
 Coordinate = Tuple[int, int]
+FloatCoordinate = Union[Tuple[float, float], np.ndarray]
+NetworkCoordinate = int
+
+Position = Union[Coordinate, FloatCoordinate, NetworkCoordinate]
+
 GridContent = Optional[Agent]
 MultiGridContent = Set[Agent]
-# used in ContinuousSpace
-FloatCoordinate = Union[Tuple[float, float], np.ndarray]
 
 F = TypeVar("F", bound=Callable[..., Any])
+Content = TypeVar("Content")
 
 
 def accept_tuple_argument(wrapped_function: F) -> F:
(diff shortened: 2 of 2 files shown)


rht:
Travis error: https://travis-ci.org/github/projectmesa/mesa/jobs/685832864#L363.

Also, it appears that mypy check hasn't been added to CI yet, might as well do it in this PR.

mesa/agent.py
```diff
@@ -17,7 +19,7 @@ def __init__(self, unique_id: int, model: Model) -> None:
         """ Create a new agent. """
         self.unique_id = unique_id
         self.model = model
-        self.pos = None
+        self.pos: Optional[Union[Coordinate, FloatCoordinate, int]] = None
```
rht, line 22: I think it would be cleaner to abstract out `Union[Coordinate, FloatCoordinate, int]`, so that when new coordinate type is added, agent.py will stay the same.

mesa/space.py
```diff
@@ -349,6 +366,7 @@ def _place_agent(self, pos: Coordinate, agent: Agent) -> None:
     def remove_agent(self, agent: Agent) -> None:
         """ Remove the agent from the grid and set its pos variable to None. """
         pos = agent.pos
+        assert isinstance(pos, Tuple[int, int])
```
rht, line 369: Why is it hardcoded to be `Tuple[int, int]`, not `Coordinate`?

mesa/space.py
```diff
@@ -856,7 +881,7 @@ def get_neighbors(self, node_id: int, include_center: bool = False) -> List[int]
 
     def move_agent(self, agent: Agent, node_id: int) -> None:
         """ Move an agent from its current node to a new node. """
-
+        assert isinstance(agent.pos, int)
```
rht, line 884: This assertion seems redundant if the user were to use mypy to make sure that `node_id` is `int`. If this assertion is added, then there are many places where similar assertions need to be added, and it might pollute the code.

mesa/agent.py
Corvince: Do you mean putting something like `Position = Union[Coordinate, FloatCoordinate, int]` into `space.py` and importing this in agent.py`?

mesa/space.py
Corvince: Because otherwise mypy throws an error of "The type alias to Tuple is invalid in runtime context". However, at runtime `Tuple[int, int]` throws an "TypeError: Subscripted generics cannot be used with class and instance checks", so neither actually works. I will see what to do about this.

mesa/space.py
Corvince: The assert is here to specifically make mypy happy. `agent.pos` might be of any type specified in the `agent.pos` type annotation and anything but int will throw an error in the `_remove_agent()` function (or, more specifically the call signature will not match)

Corvince:
Thanks for the quick look through @rht! I see there are some issues to solve, I will ping you again once an update version is online. 

> Travis error: https://travis-ci.org/github/projectmesa/mesa/jobs/685832864#L363.
> 
> Also, it appears that mypy check hasn't been added to CI yet, might as well do it in this PR.

Maybe I understand something wrong about this, but unless we first resolve all mypy errors won't this just cause travis builds to always fail?

mesa/agent.py
rht: Yes.

rht:
> Maybe I understand something wrong about this, but unless we first resolve all mypy errors won't this just cause travis builds to always fail?

Maybe without the `--strict` for now? Though I don't know how big the PR will be if all the `--strict` errors are addressed.

Corvince:
@rht
I have a new version online. I primarily added a generic BaseGrid with Content as a type variable and changed the assertions to casts. Can you take another look?

Regarding travis CI: Somewhat related to this PR are still some type errors present for HexGrid, which redefines the call signature of several methods, which violates the Liskov Substitution Principle. I am not sure how to proceed with this. Aditionally there are some type errors for the ContinuousSpace, which I am currently not planning to address in this PR. Thus, even without `--strict` mypy will cry.

mesa/space.py
```diff
@@ -118,29 +122,26 @@ def __init__(self, width: int, height: int, torus: bool) -> None:
...
+    def coord_iter(self) -> Iterator[Tuple[Content, int, int]]:
         """ An iterator that returns coordinates as well as cell contents. """
         for row in range(self.width):
             for col in range(self.height):
                 yield self.grid[row][col], row, col  # agent, x, y
 
-    def neighbor_iter(
-        self, pos: Coordinate, moore: bool = True
-    ) -> Iterator[GridContent]:
+    def neighbor_iter(self, pos: Coordinate, moore: bool = True) -> Iterator[Agent]:
```
rht, line 144: Did I make a mistake when I annotated the return type? Seems like this change can be put in a separate commit for clarity.

Corvince: Yes, the return type annotations were wrong. I also just cached this. But despite its name `iter_cell_list_contents` does not return `Iterator[Content]`, but rather `Iterator[Agent]` (since empty cells are not iterated). And this trickles down to all neighbor functions. I'll see what I can do about separating the commits.

Corvince: Good idea to separate the commit. Found several more places were the return type was wrong.

rht:
LGTM

mesa/space.py
```diff
@@ -415,6 +412,25 @@ def exists_empty_cells(self) -> bool:
...
+    def _place_agent(self, pos: Coordinate, agent: Agent) -> None:
+        """ Place the agent at the correct location. """
+        x, y = pos
+        self.grid[x][y] = agent
+        self.empties.discard(pos)
+
+    def _remove_agent(self, pos: Coordinate, agent: Agent) -> None:
+        """ Remove the agent from the given location. """
+        x, y = pos
+        self.grid[x][y] = None
```
rht, line 430: Wait, why is this not `self.default_val()`?

Corvince: Hmm it probably should be, but I just copied the function here, so this seems to be an already present bug (and unrelated to this PR)

mesa/agent.py
rht: Want to know what Agents.jl does for this? They define separate agent struct for each space type, e.g. `ContinuousAgent` https://github.com/JuliaDynamics/Agents.jl/blob/0c660872f983b5a981671bc8f1b5c901c8799a93/src/core/agents.jl#L100

rht: This has never been a problem in Python because of duck typing and that methods can have any type, but once we go to explicit typing, we get mess like in Agents.jl.

Corvince:
I have rebased and updated this PR and it is again good to go!

rht:
Closed in favor of #1219. Most notable is that I choose not to implement the generic content type, because the type of the grid content is not interchangeable, unlike `List[...]`, which can be `List[str]`, `List[int]`, and so on.

PR #853, opened 2020-06-10
This should fix cookie cutter in the install. Fixes #850

### LINKED ISSUE #850: Cookiecutter doesn't work on 0.8.7 release
**Describe the bug**
`mesa startproject` fails after `pipenv install mesa`
```
A valid repository for "/home/neil/.local/share/virtualenvs/baseline-economy-6fg_iky1/lib/python3.8/site-packages/mesa/cookiecutter-mesa" could not be found in the following locations:
...
```

**Expected behavior**
Generate the project layout

**To Reproduce**
- pipenv install mesa
- mesa startproject

**Additional context**
The cookiecutter directory from the repo is missing from the installation.
Additionally there is no help message for `startproject` when you run `mesa --help`

jackiekazil:
This fixes #850. The cookiecutter directory doesn't seem to be have been included in the manifest and no one noticed before... I guess.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,6 +3,7 @@ include LICENSE
 include HISTORY.rst
 include README.rst
 include setup.py
+include mesa/cookiecutter-mesa/*
 include mesa/visualization/templates/*.html
 include mesa/visualization/templates/css/*
 include mesa/visualization/templates/fonts/*


rht:
Should there be a step to validate `mesa startproject` on Travis? Or it is not necessary since the PR fixes the problem once for all?

PR #863, opened 2020-06-30
remove relative imports -- fix #855

### LINKED ISSUE #855: API Documentation for Space module not available on readthedocs.io
**Describe the bug**
The API documentation section in mesa.readthedocs.io is missing the Space module.
**Expected behavior**
Space module API should be accessible through readthedocs.io
**To Reproduce**
https://mesa.readthedocs.io/en/master/apis/api_main.html

Corvince:
#855 fails due to sphinx having problem with relative imports, depending on the folder structure. With our current folder structure we can not use relative imports. This PR changes the relative import to absolute ones. 

Further details about the underlying issue can be found at [](https://stackoverflow.com/questions/20251007/sphinx-and-relative-imports-in-python-3)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,6 @@ target/
 .mypy_cache/
 .dmypy.json
 dmypy.json
+
+# pythonenv
+pythonenv*/
--- a/mesa/__init__.py
+++ b/mesa/__init__.py
@@ -7,8 +7,8 @@ Core Objects: Model, and Agent.
 """
 import datetime
 
-from .model import Model
-from .agent import Agent
+from mesa.model import Model
+from mesa.agent import Agent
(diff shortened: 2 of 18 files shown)


jackiekazil:
What about time? https://github.com/projectmesa/mesa/blob/Corvince-fix-issue-855/mesa/time.py

Corvince:
I removed the missing relative imports (time and agent).

Also removed the `# -*- coding: utf-8 -*-` line from all files. It is not needed for python 3. Sorry for mixing up this PR in this respect, but I feel like this is a change that could happen anywhere (and only realized in hindsight that it affected so many files)

jackiekazil (approved):

PR #864, opened 2020-06-30
Update pytest requirement to fix issues on travis

Corvince:
Travis builds are currently failing, this is an attempt to fix that


Corvince:
@jackiekazil @dmasad @tpike3 
Pinging you all because the patch is rather trivial, but important to fix errors in travis, so I am hoping for a quick merge.

PR #868, opened 2020-07-22
BatchRunner Final Draft

tpike3:
I still need to update the documentation but wanted let you start looking at the substantive changes n the branch. 

Summary of Changes:

BatchRunnerMP
----
- removed pathos (unnecessary as unorderedmap in pathos is same as multiprocessing)
- refactored so compatible with Windows, Mac and Linux OS

Batchrunner
----
- add two functions and updated run functions so users can call model datacollectors
	- get_collector_model: returns dictionary of dataframes for datacollector model reporters
	- get collector_agents: returns dictionary of dataframes for datacollector agent reporters
	(key for dictionaries is unique parameter variation with iteration number)

DataCollector
----
- expanded to allow more collection types to be passed in model reporter (agents are handled internally so I did not think it was needed). This was necessary because lambda cannot be used at the model level for multiprocessing
key_name: <class attribute>
key_name: function (e.g schedule.get_agent_count)
key_name: lambda function
key_name: [function, [<arguments for function>]

I had some non-related test failures which primarily consisted of path references on test examples (worried I inadvertently broke something), but the tests pass on my computer. One exception chart tests agent_bar_chart was causing and error I still need to fix.


tpike3:
Updated documentation, will just do one pull request

PR #869, opened 2020-07-24
Updates Documentation and Intro tutorial for BatchRunner-DataCollection Updates

tpike3:
Updates Intro Tutorial to show BatchRunner passing data collection from model into dictionary of dataframes consisting of each run


tpike3:
Dammit...sorry should be to BatchRunner branch

Issue #882, opened 2020-08-16
Make mesa faster


ghost:
**What's the problem this feature will solve?**
Sadly, Python belongs to the slowest programming languages, but it's friendly code makes it popular and extremely usable.
Agent-based simulations can involve very large number of agents (e.g. millions), where concurrency control, parallelism and  run-time speed of simulations could be crucial.

**Describe the solution you'd like**
Using:
- PyPy
- Cython (like numpy, abcEconomics: an agent-based stuff in economics)
- numba 

could help a lot in simulation speed ups.

rht:
This is the first time PyPy is mentioned in this repo.
Disclaimer: I am also one of the maintainers of abcEconomics.

It should be easy to test if Mesa runs on PyPy with Travis (see e.g. https://github.com/AB-CE/abce/blob/master/.travis.yml#L24).

rht:
In abcEconomics, there is an option to run the execution of each agents in a separate CPU (https://github.com/AB-CE/abce/blob/master/abcEconomics/scheduler/multiprocess.py). If there are 20 agents and 4 CPUs, each CPU could execute 5 agents in parallel. See any `pool.map` in the file I linked before. Care is needed to ensure that the objects are able to be pickled so they can run in each CPUs.

Corvince:
There is also a discussion about mesas performance in #798. 

To recap my 2 cents on this issue: Mesa is mostly a framework and by itself it doesn't "do" much that could be improved by your suggestions. It really depends on the models it is used for (and if the actions are computationally expensive numba might be a very good and easy to use recommendation). 

Re parallelism: Depending on the model running agents concurrently is either trivial or (almost) impossible. But usually you will run multiple simulations any way and running those in parallel is far easier, either manually or by using BatchrunnerMP (although it doesn't work at the moment, but it is being worked on).

rht:
It's worth trying [Cinder](https://github.com/facebookincubator/cinder). Currently discussed in https://news.ycombinator.com/item?id=36612835.

EwoutH:
I'm going to close this issue as completed, now that we have mesa-frames officially under the Mesa umbrella:
- https://github.com/projectmesa/mesa-frames

If there are any specific issues or ideas for Mesa performance, feel free to open a new issue or [discussion](https://github.com/projectmesa/mesa/discussions)!

PR #883, opened 2020-08-18
Travis: Add pypy3

rht:
This addresses #882.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ python:
   - "3.6"
   - "3.7"
   - "3.8"
+  - "pypy3"
 install:
   - pip install codecov
   - pip install .[dev]


rht:
I'm not sure why there is no Travis running here ...

Corvince:
> I'm not sure why there is no Travis running here ...

No idea why it doesn't show up. It did run though, but failed at installing pandas (https://travis-ci.org/github/projectmesa/mesa/jobs/718907972)

rht:
It looks like it fails at pandas' dependency, ujson. I will keep this PR open and retry on future version of PyPy. Though PyPy developers claimed that pandas has been supported since pypy5.9 (https://doc.pypy.org/en/latest/release-v5.9.0.html#pypy2-7-and-pypy3-5-v5-9-dual-release). The current Travis PyPy is 7.1.1.

Corvince:
Your link only mentions pypy2.7, but from what I could infer from googling around, yes, it should work with pandas. Maybe just an issue with travis? @rht could you (or someone else) try it locally?

rht:
Tested on my laptop. 
pypy3-7.3.1:
```
97 passed, 1 warning in 11.26s ================================================
```

With normal python 3.8
```
97 passed, 2 warnings in 11.63s ===============================================
```

rht:
I probably need a simulation with a huge grid size in order to highly PyPy's speedup.

rht:
I ran the comparison on the `bank_reserves` example, , because it doesn't have the web GUI.
The speedup is significant!
PyPy3-7.3.1
```
python batch_run.py  89.72s user 2.78s system 100% cpu 1:31.72 total
```
CPython 3.8
```
python batch_run.py  173.73s user 2.19s system 100% cpu 2:55.10 total
```
A 1.9x speedup!

rht:
Here I removed 3.5 to make room for PyPy3. This is also justified by the fact that 3.5's EOL 2020-09-13.

PR #902, opened 2020-08-29
BatchRunner and DataCollector Updates

tpike3:


tpike3:
Fix codecov/patch failure

PR #904, opened 2020-08-29
fix flake8 failures

tpike3:


tpike3:
Should go to branch_redux

PR #906, opened 2020-08-29
Fix flake8 failure take 2

tpike3:


tpike3:
Should have been branched to BatchRunner Redux

PR #907, opened 2020-08-29
fix flake8 failures take 2

tpike3:


tpike3:
There is no longer a comma in line 56 to produce the error. Must be a Travis issue build

PR #908, opened 2020-08-29
flake8 take 3

tpike3:
Was looking at the wrong comma error fixed


tpike3:
Fix codecov/patch failure

PR #909, opened 2020-08-29
increase codecov patch

tpike3:
Also checked flake8 in program to reduce nonsense builds


tpike3:
Codecov still not high enough

PR #911, opened 2020-08-30
Increase test in datacollector and batchrunner

tpike3:
increased number of test to reach codecov patch target


tpike3:
Fix test_examples failure

PR #912, opened 2020-08-30
fix test_examples failure

tpike3:
changes in datacollector caused test_example failures this resolves and keeps added tests


tpike3:
fix test examples failures

PR #914, opened 2020-08-30
fix flake8 failures

tpike3:
ugggh forgot again


tpike3:
random git error try again

PR #915, opened 2020-08-30
attempt 2 -- random git error

tpike3:
pruned git tree attempting again


tpike3:
still need to increase codecov patch

Issue #919, opened 2020-09-07
Duplicate unique_id handling in BaseScheduler.add


im-sophie:
**What's the problem this feature will solve?**

Mesa's great! One thing though - I spent a couple hours yesterday trying to figure out why one of my agent's `step()` method wasn't being called, and then I realized that I was using the same `unique_id` value for both agents. Here's a simplified example:

```python
from mesa import Agent, Model
from mesa.time import RandomActivation

class TestAgent(Agent):
    def __init__(self, unique_id, model):
        super().__init__(unique_id, model)

    def step(self):
        print("agent with unique_id {0} gets stepped".format(self.unique_id))

class TestModel(Model):
    def __init__(self):
        super().__init__()

        self.scheduler = RandomActivation(self)

        # note there's two agents added here, but both with the same unique_id
        self.scheduler.add(TestAgent(0, self))
        self.scheduler.add(TestAgent(0, self))

    def step(self):
        self.scheduler.step()

m = TestModel()
m.step()
```

This outputs the following, showing only one agent got added. I get why, since `BaseScheduler._agents` is a dictionary.

```
agent with unique_id 0 gets stepped
```

**Describe the solution you'd like**

I understand if you don't want this to be implemented because it's a change to a commonly-used method in the API and the behavior of `add` methods is kind of subjective, but what do you think about adding an exception if the same `unique_id` is added multiple times? I'd be happy to make the change, small as it is.

**Additional context**

tpike3:
Absolutely, I think that would be great! Would be happy to incorporate an exception if a user attempts to reuse an ID.

Corvince:
Yes, as the name suggests "unique_id" should be unique and I don't think this would break any existing models and we should add an exception if agents with the same id are added multiple times (otherwise we should at least emit a deprecation warning and add an exception in a later version of mesa)

sophiekitkatz:
Great! If it's OK with you, I'll write up a proposal as I'm able.

PR #920, opened 2020-09-12
Fix for #919, adding exception for duplicate ids

sophiekitkatz:


tpike3:
Awesome @sophiekitkatz , thanks! 

Would you be interested in adding the test to this line so it passess the codecov check? (if not we can do it) 

Essentially what you would do is add in test_time.py a function to re add an agent that exists something like 
     `self.assertRaises("String of exception raised", model.schedule.add, MockAgent("A"))`

Let me know

sophiekitkatz:
Yep! I'll add it when I'm able.

Corvince:
Thanks again @sophiekitkatz

PR #924, opened 2020-10-03
Re-Implementation of BatchRunner 

Corvince:
This is a draft PR for a reimplemented BatchRunner following discussions on the dev mailing list. The goal is to simply the code and allow easier addition of new features. Currently models are run but no data is collected. 

Unfortunately the file diff looks a bit strange.


Corvince:
This is now at a point where it will succesfully run multiple runs with or without multiprocessing and return the data from the specified reporters.

here is how it works currently:

```python
model_params = {"param_A": range(3), "param_B": 5}
batch_run(MyModel, model_params, iterations: 2, model_reporters: {"value_1": some_func})
```
Output:
```python
[{'param_A': 0, 'param_B': 5, 'value_1': 5, 'iteration': 1, 'RunId': 0},
 {'param_A': 1, 'param_B': 5, 'value_1': 6, 'iteration': 1, 'RunId': 1},
 {'param_A': 2, 'param_B': 5, 'value_1': 7, 'iteration': 1, 'RunId': 2},
 {'param_A': 2, 'param_B': 5, 'value_1': 7, 'iteration': 2, 'RunId': 3},
 {'param_A': 1, 'param_B': 5, 'value_1': 6, 'iteration': 2, 'RunId': 4},
 {'param_A': 0, 'param_B': 5, 'value_1': 5, 'iteration': 2, 'RunId': 5}]
```
So for each run it outputs a dictionary of input parameters used, values reported and informations on the runID and iteration count. The same is true for agent reporters:

```python
batch_run(MockModel, model_params, model_reporters={"value_1": "C"}, agent_reporters={"agent_value": "my_attribute"})
```

```python
[{'param_A': 0, 'param_B': 5, 'value_1': 5, 'AgentID': 0, 'agent_value': 'A', 'iteration': 1, 'RunId': 0}, 
 {'param_A': 0, 'param_B': 5, 'value_1': 5, 'AgentID': 1, 'agent_value': 'B', 'iteration': 1, 'RunId': 0},
 {'param_A': 2, 'param_B': 5, 'value_1': 7, 'AgentID': 0, 'agent_value': 'A', 'iteration': 1, 'RunId': 1},
 {'param_A': 2, 'param_B': 5, 'value_1': 7, 'AgentID': 1, 'agent_value': 'B', 'iteration': 1, 'RunId': 1},
 {'param_A': 1, 'param_B': 5, 'value_1': 6, 'AgentID': 0, 'agent_value': 'A', 'iteration': 1, 'RunId': 2},
 {'param_A': 1, 'param_B': 5, 'value_1': 6, 'AgentID': 1, 'agent_value': 'B', 'iteration': 1, 'RunId': 2}]
```

Here, the dictionary is somewhat repeated for each agent. This allows for an easy creation of a Pandas DataFrame in "long" or "tidy" form. Just do `df = pd.DataFrame(data)` and continue with your analysis.

The current state could be improved by also collecting the data every n-th step (new dict-key/column "step") or by adding the model seed for reproducible runs. 

However, first I would like to discuss on the agent and model reporters. These are the same as data collectors found usually in model runs and I wonder if we really need to repeat that for batch runs. Or should we use the existing data collectors and just use the latest value or every n-th step? I think this would further simplify the API and would prevent accidentally collecting data twice and would resolve issues about how to retrieve step data, if needed.

Corvince:
This is now ready to review @jackiekazil !!

I further simplified the API. You can now run multiprocessing batch runs on all platforms by simply calling:

```python
from model import MyModel
results = batch_run(MyModel, {"dictionary_of": "model_kwargs", "with_possible": ["i", "t", "e", "r", "a", "b", "l", "e", "s"]}
```

The result will look like in my previous comment, with better ordering. 

`batch_run` now reuses the existing DataCollectors and by default collects only data from the last step. However, with the `ì_steps` parameter you can also enable collecting data every, well, i steps. The code that allows this could be refactored into the datacollector module, but this is a task for a future PR.

I hope this really simplifies the usage of the BatchRunner.

jackiekazil:
Thank you for this. I have it on my todo list... currently catching on tickets.

tpike3:
@Corvince  Sorry for the lapse on my part I will review this weekend

tests/test_batch_run.py
```diff
@@ -0,0 +1,125 @@
+from mesa.agent import Agent
```
tpike3, line 1: Why aren't unittests being used like in other test files?

tests/test_batch_run.py
```diff
@@ -0,0 +1,125 @@
...
+        MockModel,
+        {
+            "variable_model_params": range(5),
+            "variable_agent_params": ["H", "E", "L", "L", "O"],
+        },
+    )
+
+
+def test_batch_run_single_core():
+    batch_run(MockModel, {}, nr_processes=1, iterations=10)
```
tpike3, line 125: To run Pools with Windows you have to use freeze_support otherwise it fails while also working all other OSs.
I thought we added Mac OS and Windows OS to our testing if not then this is just a good to know.  

```
if __name__ == "__main__":
    freeze_support()
```

tpike3 (commented):
@Corvince This is great! so compact and elegant. To backbrief this adds the function batch_run while keeping the old classes so users now have an alternative. Should we add a deprecation warning to other classes? 

The only item of note was in tesst it wasn't using unittests like the other test modules, so I had to add some lines to run locally

One potential item was requires use of freeze_support for Windows OS

tests/test_batch_run.py
Corvince: With pytest there should be no need to import unittest. Was there anything in particular not working for you?

tests/test_batch_run.py
Corvince: Hmm I was developing this on a windows machine and never had any issues. But from what I heard it should not hurt to add freeze_support!? This is only for running the tests, right?

Corvince:
Thanks for the review @tpike3 Not sure which issues you were running into, I developed this on Windows. Can you give me more details on what to improve or even amend this PR with a commit of your own?

tests/test_batch_run.py
tpike3: Hmm it wasn't letting me run the tests, maybe I forgot a step setting up my ide since I just switched computers,  I will relook

tests/test_batch_run.py
tpike3: That is weird until I added freeze_support I was getting errors and saying use freeze_support and I have never had a multiprocessing batchrunner work unless I have used freeze_support. I am intrigued, let me dig a little deeper and see if I can find a reason.

tpike3: @Corvince as usual I learn a ton from reading your code. So aplogies it took so long, but thank you!

Apparently it all does boil down to pytest and how it is handling it tests although I could not find an explicit discussion of pytest and multiprocessing.

Regardless, the code is all good and just for reference I found this description of what was happening when I was using unittest and getting a runtime error: [differences between multiprocessing windows and linux](https://www.pythonforthelab.com/blog/differences-between-multiprocessing-windows-and-linux/l)

tpike3 (approved):
Phenomenal!

adrien-perello:
Maybe it would be worth updating the `_prepare_report_table` method too? (see [#1077](https://github.com/projectmesa/mesa/issues/1077))

toohuman:
@Corvince Thank you for your work on this new version of BatchRunner. One issue I have found is that some of the parameters being passed to the model may be strings, and in that case your version interprets a string parameter as being a character array to create batches from. Should this be default behaviour? I can enclose string arguments inside a list but this caught me off-guard.

examples/bank_reserves/batch_run.py
```diff
@@ -181,25 +181,36 @@ def run_model(self):
...
+    data = batch_run(
+        BankReservesModel,
+        br_params,
+        model_reporters={"Rich": get_num_rich_agents},
+        agent_reporters={"Wealth": "wealth"},
+    )
+    br_df = pd.DataFrame(data)
+    br_df.to_csv("BankReservesModel_Data.csv")
+    # br.run_all()
+    """
```
rht, line 208: The docstring-ed code below should probably be removed?

tpike3:
I am really sorry I am not sure what my IDE did here. I did not intend to do any changes, I was updating my local and then pushing to my fork but is somehow went to upstream. I am working to delete these changes.

examples/bank_reserves/batch_run.py
jackiekazil: @rht I want to push this PR thru. Can you either fix this in another PR or create a ticket?

EwoutH:
Congratulations to everyone involved for getting this in!

This PR has breaking changes right? If so, how well are they currently documented? Should a migration guide be available?

jackiekazil:
@EwoutH moving discussion to address in #1108. Ty for bringing it up.

examples/bank_reserves/batch_run.py
rht: I made #1110.

Issue #856, opened 2020-06-17
Modular Server doesn't handle multiple connections properly


NeilW:
**Describe the bug**
If you open multiple tabs in your browser to the Mesa server you still only get one model running. Altering the paramenters on one tab affects the model output on all the others.  

**Expected behavior**
Each tab should run its own instance of the model. Messages from the client should be applied to
the model associated with the web socket. 

**To Reproduce**
Open multiple tabs and point them at the mesa server. Adjusting the parameters on one tab will alter the output on the others. 

**Additional context**
This fault stops the mesa server from being used as a public facing application server. It's fairly easy to fix. The model just has to be lifted from the application to the socket handler and the socket can act as the session layer. I've made the necessary alterations here: https://github.com/newwayland/baseline-economy/blob/master/BaselineEconomy/ModularVisualization.py

Corvince:
Thank you for the bug report! Could you submit your patch as a Pull Request??

Corvince:
@NeilW 
I saw that you have a fix for this issue on your fork. Can you submit a PR with that? Don't worry about failing tests or something, I can add/modify them if you don't want to do that. I could also copy paste your fix, but I would like to give the credit to you

NeilW:
I've added a PR. Apologies for not getting onto the tests. They looked like they may need a bit of refactoring and I haven't had the bandwidth to tackle them

rht:
I tested running multiple tabs for the Solara viz, and one of the tabs crashed. It looks like the only way for the Solara viz to be accessed in parallel in public is via Jupyter.

Corvince:
Interesting, what was the error message?

rht:
I just tested with 2 tabs (one in incognito) via Jupyter. The 2 simulations in the 2 tabs run in parallel, independently. This means that we can offload #481 to point to how to expose a Jupyter server to be web public.

rht:
One complication is that, some users probably don't want to expose their full source code. I will ask at Solara Discord about this topic.

rht:
It seems to be a fixable problem. From Solara Discord:
> Solara should work in multiple tabs. I can be that somewhere, your app has global shared state that's accessed from multiple threads (each tab runs in it's own thread)

rht:
Fixed in #1759 for Solara.

PR #934, opened 2020-10-25
Allow multiple connections

NeilW:
This is the code fix.

Sorting the tests appears to be a bit of work - which I haven't got around to unfortunately.


mesa/visualization/ModularVisualization.py
```diff
@@ -106,6 +107,11 @@
 
 from mesa.visualization.UserParam import UserSettableParameter
 
+template_path = os.path.join(
```
Corvince, line 110: Just for clarification: Why did you add this?

mesa/visualization/ModularVisualization.py
```diff
@@ -236,25 +287,38 @@ class ModularServer(tornado.web.Application):
     # Handlers and other globals:
     page_handler = (r"/", PageHandler)
     socket_handler = (r"/ws", SocketHandler)
+    health_handler = (
```
Corvince, line 290: When/how is this being used?

NeilW: It's the check path on Kubernetes clusters so you can tell when the container is ready for service.

Corvince:
Thanks for the PR! As I said I can take care of the tests. If you get around to answering two small questions regarding your code (in the code comments above), that would be great!

mesa/visualization/ModularVisualization.py
NeilW: As per commit message

"Obtain the template directory from the module installation path
not the specific file - allowing you to have variations of the file
in your model if required"

mesa/visualization/ModularVisualization.py
NeilW: Again as per commit message

Corvince: Sorry, didn't think of reading the commit message :see_no_evil:
In the line below is "healthz" intentional (the z) or by accident

Corvince: sorry, still hadnt read teh commit message, seems intentional

NeilW: 'healthz' is the default path for Kubernetes health checks. It annoys me as well.

Corvince:
Testing was indeed more challenging than I initially thought. There is no (simple?) way to access the SocketHandler instance and googling around I decided to extract the methods further more. They are now functions outside of any handlers. This probably makes more sense and is easier to test.

jackiekazil:
I don't have anything to add at the moment other than this looks cool.

jackiekazil:
Resolves: #856

Corvince:
After agreeing to merge this one in todays dev meeting, I tested it again and it seems not to work completely as expected. 
While you can run multiple models in multiple tabs and they don't interrupt each other, as soon as you modify a parameter, that parameter is changed across all models, once they are reset.
To easily reproduce:
 1. Run Schelling example in Tab1
 2.  Open in second Tab (Tab2)
 3.  Decrease agent density in Tab 2
 4. Reset model in Tab 1, agent density will be lower

Can you confirm this @NeilW or am I doing something wrong?

Corvince:
Sorry I accidently killed this PR, I will reopen a truly working solution

Issue #854, opened 2020-06-14
Server Tornado NotImplementedError Python 3.8


aidiss:
**Describe the bug**

NotImplementedError when launching server in Python 3.8

**Expected behavior**

No error should happen

**To Reproduce**
1. Install Python3.8
2. Create a server with ModularVisualization

**Additional context**

```
Interface starting at http://127.0.0.1:8521
Traceback (most recent call last):
  File "run.py", line 3, in <module>
    server.launch(open_browser=True)
  File "C:\Users\admin\AppData\Local\pypoetry\Cache\virtualenvs\conference-market-abm-paper-t0bN9evo-py3.8\lib\site-packages\mesa\visualization\ModularVisualization.py", line 335, in launch
    self.listen(self.port)
  File "C:\Users\admin\AppData\Local\pypoetry\Cache\virtualenvs\conference-market-abm-paper-t0bN9evo-py3.8\lib\site-packages\tornado\web.py", line 2116, in listen
    server.listen(port, address)
  File "C:\Users\admin\AppData\Local\pypoetry\Cache\virtualenvs\conference-market-abm-paper-t0bN9evo-py3.8\lib\site-packages\tornado\tcpserver.py", line 152, in listen
    self.add_sockets(sockets)
  File "C:\Users\admin\AppData\Local\pypoetry\Cache\virtualenvs\conference-market-abm-paper-t0bN9evo-py3.8\lib\site-packages\tornado\tcpserver.py", line 165, in add_sockets
    self._handlers[sock.fileno()] = add_accept_handler(
  File "C:\Users\admin\AppData\Local\pypoetry\Cache\virtualenvs\conference-market-abm-paper-t0bN9evo-py3.8\lib\site-packages\tornado\netutil.py", line 279, in add_accept_handler
    io_loop.add_handler(sock, accept_handler, IOLoop.READ)
  File "C:\Users\admin\AppData\Local\pypoetry\Cache\virtualenvs\conference-market-abm-paper-t0bN9evo-py3.8\lib\site-packages\tornado\platform\asyncio.py", line 100, in add_handler
    self.asyncio_loop.add_reader(fd, self._handle_events, fd, IOLoop.READ)
  File "C:\Users\admin\AppData\Local\Programs\Python\Python38\lib\asyncio\events.py", line 501, in add_reader
    raise NotImplementedError
NotImplementedError
```

Solution

The issue is known and documented here
https://www.tornadoweb.org/en/stable/index.html#installation
[...]

PR #936, opened 2020-10-28
ModularServer updates: Fix EventLoop and changes to default port

Corvince:
First, this is a fix to an issue on windows, where the default event loop is incompatible with tornado (#854)

Second, I changed the default port to use an environmental variable `$PORT`, if it exists. This is for example used by Heroku and will make it a bit easier to deploy models there.

### DIFF
--- a/mesa/visualization/ModularVisualization.py
+++ b/mesa/visualization/ModularVisualization.py
@@ -236,7 +236,7 @@ class ModularServer(tornado.web.Application):
 
     verbose = True
 
-    port = 8521  # Default port to listen on
+    port = int(os.getenv("PORT", 8521))  # Default port to listen on
     max_steps = 100000
 
     # Handlers and other globals:
--- a/mesa/visualization/ModularVisualization.py
+++ b/mesa/visualization/ModularVisualization.py
@@ -95,7 +95,9 @@ Client -> Server:
     }
 
 """
+import asyncio
 import os
+import sys
 import tornado.autoreload
 import tornado.ioloop
 import tornado.web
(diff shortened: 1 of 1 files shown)


jackiekazil:
Before I set a VM for windows, @tpike3 or @dmasad have the ability to test this on windows?

Corvince:
Copy and paste from the email list:

Hmm, maybe some more background on the change, as I think it is actually pretty important to merge it as soon as possible:

Python 3.8 (released Oct 14th 2019) changed the default event loop from SelectorEventLoop to ProactorEventLoop. This broke Tornado (see this issue). Initially tornado didn't want to resolve this issue and put a note that any packages dependent on tornado (e.g. jupyter) should set the event loop policy explicitly. So the code change merely does nothing on older versions of Python and changes it to the old default (and only working) event loop for Python >= 3.8. This is now somewhat outdated with the release of tornado 6.1 (released Oct 30th 2020, less then a week ago), which falls back to running selector in a second thread.

This means that mesa on windows was broken (at least the visualization) for more than a year already, if someone uses the latest version of python. And since I currently use windows I can confirm this. And until last week the only way to resolve this would be via the code change in the PR. Now we also have the option to change the requirements for python to be at least version 6.1, but since this was released just a week ago I am not sure this is appropriate. Although in the long term this is probably the right way to do it.

Corvince:
Forgot the conclusion re testing:
Currently the visualizations of mesa are broken on windows. So the worst case of merging this is that they remain broken, but since this is a recommended code change it should work (and does on my laptop)

dmasad:
Tested this on Windows, and it seems to work for me.

jackiekazil:
With #936 merged... i am putting this release together.

PR #937, opened 2020-10-28
Update travis python versions

Corvince:
Update to include Python 3.9 testing on travis.

Since we already removed support for Python 3.5 earlier, we don't need to juggle around with python versions for black and I now added black as a developer tool/install in `setup.py`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,17 +8,15 @@ python:
   - "3.6"
   - "3.7"
   - "3.8"
+  - "3.9"
 install:
   - pip install codecov
   - pip install .[dev]
-  - if [[ $TRAVIS_PYTHON_VERSION != 3.5 ]]; then pip install -U black; fi
 script:
   - export PYTHONPATH=$PYTHONPATH:$(pwd)
   - flake8 .
   - py.test --cov=mesa tests/ --cov-report=xml
-  - if [ $PYTHON != "3.5" ];
-    then black --check --exclude=mesa/cookiecutter-mesa/* .;
-    fi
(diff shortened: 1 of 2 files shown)


jackiekazil:
@Corvince can you resolve the conflict on this?

Corvince:
@jackiekazil done

rht:
Then there is no slot left for PyPy?

rht:
Also, there is this [HN comment](https://news.ycombinator.com/item?id=24837142) on the problem of using 3.9 for scipy stack. Not sure if it has been fixed.

Corvince:
@rht 
Mesa uses travis mainly for testing incoming changes. The question being asked: "Does this code change still pass our tests for the supported python versions and is it formatted correctly?" 

We can't answer that question for PyPy, because the installation of PyPy itself fails on travis. Once this is resolved I am very happy to include it on travis. May I ask why you think this PR leaves no slot for PyPy?? 

But we can answer that question for Python 3.9. Does this mean Python 3.9 is recommended to use right now? No, but we know that it doesn't fail. (but thanks for the HN comment, I wasn't aware of this)

rht:
> May I ask why you think this PR leaves no slot for PyPy??

I don't have a source to back this, but IIRC Travis only has 4 slots of free instances for FOSS projects.

> Does this mean Python 3.9 is recommended to use right now? No, but we know that it doesn't fail. (but thanks for the HN comment, I wasn't aware of this)

We should wait until 3.9.1 in December.

rht:
This issue can be closed.

Issue #943, opened 2020-10-30
ContinuousSpace.get_heading returns wrong result sometimes in toroidal space mode


senden9:
**Describe the bug**
`ContinuousSpace.get_heading` returns wrong result when going over the middle line and toroidal space is enabled.

**Expected behavior**
Heading from `(51, 51)` to `(49, 49)` should return `(-2, -2)`.

**To Reproduce**
```python
import numpy as np
from mesa import space

cspace = space.ContinuousSpace(100, 100, True)
print(cspace.get_heading((51, 51), (49, 49)))
```
Prints `(98.0, 98.0)`. So it goes 180 Degree in the wrong direction.

If I shift the points down I get the right results:
```python
import numpy as np
from mesa import space

cspace = space.ContinuousSpace(100, 100, True)
print(cspace.get_heading((11, 11), (9, 9)))
```
Prints `(-2.0, -2.0)`. So the correct result.

catherinedevlin:
Confirmed that the bug is still present in v1.2.1.

rht:
Fixed by #1686.

PR #944, opened 2020-10-30
Fixes problem with ContinuousSpace.get_heading in toroidal space mode

senden9:
See issue projectmesa/mesa#943

It is for sure not a beautiful fix but it works for my use-case.


senden9:
Closed in the interest of #1664.

Issue #787, opened 2020-01-25
BatchRunnerMP raises pickle typeError


nathanvdlei:
**Describe the bug**
Using the BatchRunnerMP.run_all() function results in a pickle TypeError:

> can't pickle dict_values objects

I suspect the bug is in the multiprocess module. Updating pickle, mesa, multiprocess dill etc did not solve the problem.

**Additional context**
Python 3.7.3
Working in jupyter notebook
OS: Windows 10
Intel Core i7-8565U Processor

Full stacktrace:
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-2e9800f54d7c> in <module>
     15                       max_steps=500,
     16                       model_reporters=data_collector)
---> 17 batch.run_all()
     18 
     19 data = batch.get_model_vars_dataframe()

c:\users\natha\agent based modelling\notebooks\1. mesa introduction\src\mesa\mesa\batchrunner.py in run_all(self)
    378             results = []
    379             for task in job_queue:
--> 380                 for model_vars, agent_vars in list(task):
    381                     results.append((model_vars, agent_vars))
    382                 pbar.update()

~\Anaconda3\lib\site-packages\multiprocess\pool.py in next(self, timeout)
    746         if success:
    747             return value
--> 748         raise value
    749 
    750     __next__ = next                    # XXX

~\Anaconda3\lib\site-packages\multiprocess\pool.py in _handle_tasks(taskqueue, put, outqueue, pool, cache)
    429                         break
    430                     try:
--> 431                         put(task)
    432                     except Exception as e:
    433                         job, idx = task[:2]

~\Anaconda3\lib\site-packages\multiprocess\connection.py in send(self, obj)
    207         self._check_closed()
    208         self._check_writable()
--> 209         self._send_bytes(_ForkingPickler.dumps(obj))
    210 
    211     def recv_bytes(self, maxlength=None):

~\Anaconda3\lib\site-packages\multiprocess\reduction.py in dumps(cls, obj, protocol, *args, **kwds)
     52     def dumps(cls, obj, protocol=None, *args, **kwds):
     53         buf = io.BytesIO()
---> 54         cls(buf, protocol, *args, **kwds).dump(obj)
     55         return buf.getbuffer()
     56 

~\Anaconda3\lib\site-packages\dill\_dill.py in dump(self, obj)
    443             raise PicklingError(msg)
    444         else:
--> 445             StockPickler.dump(self, obj)
[...]

PR #947, opened 2020-11-03
convert dict_values to list in BatchRunnerMP

Boukefalos:
Simple fix for issue #787


Corvince:
Thanks for the effort, seriously appreciate it. However, #787 has been resolved recently and your code change converts something to a list that now already is a list.

Issue #954, opened 2020-11-12
Option title for ChartModule


anjalinair012:
**What's the problem this feature will solve?**
Adding a title to Charts in order to make their purpose clear

**Describe the solution you'd like**
A way to pass the title text along with the chart data

jackiekazil:
This is a fair request. Right now the only way to make a title is just add free text in between, but a title should be a param that you pass on chart.

PR #955, opened 2020-11-12
Added optional title feature to ChartModule

anjalinair012:
A suggested solution to issue #954

### DIFF
--- a/docs/issue_954.rst
+++ b/docs/issue_954.rst
@@ -1,0 +1,31 @@
+issue: Optional title in ChartModule #954
+
+mesa.visualization.modules package
+==================================
+
+Submodules
+----------
+
+mesa.visualization.modules.ChartVisualization module
+----------------------------------------------------
+Extra parameter "chart_title" added to ChartModule __init__() . 
+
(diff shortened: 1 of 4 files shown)


jackiekazil:
This looks good. Can you add an example implementation into one of the examples?

anjalinair012:
Example \mesa\examples\charts\charts\server.py has been modified for testing. Please refer to mesa\docs\issue_954.rst .

JaslinRikkaz123:
I have tried this but it doesn't show the title in the chart module output. What i have to do to get the chart title?

catherinedevlin:
I tried producing an updated PR with this code - https://github.com/projectmesa/mesa/pull/1666 - but like @JaslinRikkaz123, did not see the title actually showing up on the screen.  I also think that, if it is made to work, it should work with all the visualizations (bar, pie, ...), not just line.

Corvince:
Thank you for the PR and sorry for never putting it forward 

If this is still actively followed please re-open the PR at https://github.com/projectmesa/mesa-viz-tornado
But note that we are moving towards a new pure python frontend

Issue #938, opened 2020-10-29
Travis-ci.org shutting down Dec 31st 2020


Corvince:
I somehow missed that until today, but www.travis-ci.org will shut down at the end of the year. Not a very much big deal, since it will continue to operate under www.travis-ci.com, but we would need to create an account there and migrate the integration. This *should* be fairly easy ([here is the migration guide](https://docs.travis-ci.com/user/migrate/open-source-repository-migration)). However, we could also consider switching to [GitHub Actions](https://github.com/features/actions) for our CI/CD pipeline. 

What are some thoughts on this?

jackiekazil:
Good catch! I pinged a few friends who are more into this than me. I will update when they get back to me. 
Another option is CircleCi.

rht:
One vote for CircleCI -- it is used by Zulip (https://github.com/zulip/zulip/blob/master/.circleci/config.yml).

jackiekazil:
I heard a lot of hype around Circle. So I did a bunch of research by polling some folks I respect and reading up.  It seems 2/1 that people lean on github actions versus other options (eg Travis or Circle) . Github actions seemed to have more weight than I expected. People leaning towards github actions if there is a change on the project for the following reason -- It is built in and if someone forks it the actions are also forked and CI/CD is run on the fork. 

Assuming that is true. I think that is a good reason as well as not having to deal with another interface. 

My vote is for taking this opportunity to move to github actions. (I am willing to take this on unless someone else is inspired.)

(Honestly I would love to move to gitlab, but I feel we are too established on Github... that is a conversation for another day.)

jackiekazil:
@rht said Circle. 
I say github actions (or fix travis, but that makes me feel - meh.) 

@Corvince what are your final thoughts? 
@tpike3, @dmasad do either of you have thoughts on this?

Corvince:
I would vote for GitHub Actions too. I have used it very rudimentary for mesa-viz and it was quite straight forward to set up (also my setup was very simplified, so I can't really judge that). CircleCI also seems a bit more restricted with only build environments for windows and unix, where GA also has macOS. And the "runs on forks" argument is really compelling - and it makes sense, since the configuration becomes part of the repo.

That said I don't have a qualified opinion on CircleCI. @rht do you know of some unique advantages?

tpike3:
[...]

PR #966, opened 2020-12-15
Add GitHub Action for continuous integration testing

Corvince:
Given the issue of travis.org shutting down on Dec. 31st 2020, which is just 2 weeks away (tracked in #938), this PR would add the github actions required to continue our testing as usual with GA. 

I know #938 is not yet resolved and I don't take any damage if we decide not to use GA, but CircleCI or something else (but the current vote favors GA so I went ahead).


jackiekazil:
Thank you @Corvince! I was just digging through issues trying to catch up. 
What do you think is the best way to "test" this? Other than.... 

1. Is it working with GA? 
2. Does it do at least what the old stuff did?

rht:
Is GA not yet enabled? I can't find the GA logs. The coverage was probably generated by Travis. Maybe also delete .travis.yml in this PR?

Corvince:
Good idea to delete the travis configuration.

The GA are working the logs are here (for my fork)
https://github.com/Corvince/mesa/actions/runs/454126773

Currently it does what we did on travis: runs flake8, black and pytest and submits the code coverage to coveralls.

Runs on 3.6--3.9 on ubuntu

rht:
Tests look good, e.g. https://github.com/Corvince/mesa/runs/1629438627?check_suite_focus=true#step:7:39.

.github/workflows/python.yml
```diff
@@ -0,0 +1,38 @@
...
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install codecov black
+        pip install .[dev]
+        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
```
rht, line 24: Why is this `if` clause required?

Corvince: Ah, it's an artifact of the GA template. Don't know why I kept it, mesa doesn't have any requirements file. Which also explains why there is an if statement, but not why we need it. Should I delete the line?

rht: IMO just delete it. It will become confusing in the future without proper comment explaining why.

Corvince: You are probably right, I deleted it

jackiekazil:
This looks good to me. 
@Corvince, i am going to merge to push forward. Can you submit another PR with the line deletion that @rht caught?

PR #970, opened 2021-01-01
GA: Run on Windows and macOS

rht:


rht:
The build fails because there is a problem with asyncio package on Windows on 3.6. Not sure how to specifically remove 3.6 from the CI matrix.

rht:
OK, I managed to exclude 3.6 on Windows.

rht:
Caveat: Windows tests are slow because it takes ages to install the dependencies.

Corvince:
Thanks for PR. I would even go further and add macos builds and potentially pypy3 builds (if they work, would be nice if you could test that. IIRC it was working on your local machine, but somehow not on travis).

> 
> 
> OK, I managed to exclude 3.6 on Windows.

Err, I don't think this is the right approach. This error shows the value of testing mesa on windows, so we should not exclude a failing build (if we want to support Python 3.6), but rather fix it. I can submit a PR for this specific error (I am somewhat responsible, since I introduced the failing part to fix Python 3.8+ builds on windows).

rht:
That 3.6 on Windows bug should probably not block the other Windows tests from being enabled, hence this PR.

rht:
Added macos-latest.

rht:
I find the number of parallel builds excessive, even though it is MSFT's money that is being burnt. Maybe only do one version of Python for each Windows and macOS?

Corvince:
> I find the number of parallel builds excessive, even though it is MSFT's money that is being burnt. Maybe only do one version of Python for each Windows and macOS?

I agree. Not sure if we should only test the oldest or latest version of python. I lean towards latest version, but don't have a strong opinion yet.

jackiekazil:
What about the two latest? (I feel like that is reasonable. Just in case we end up in a situation where the latest is still working out bugs.)

rht:
I have updated the PR to only do 3.9 for macos-latest and windows-latest.

Corvince:
> What about the two latest? (I feel like that is reasonable. Just in case we end up in a situation where the latest is still working out bugs.)

I don't care if we use 1 or 2 latest versions, but we could just delay updating the "latest" version if there are bugs being worked out. Just an idea to put this PR forward (since @rht seems to prefer just the latest version)

rht:
I misinterpreted "two latest" as two oses' latest version. :/

rht:
The list will get even more crowded with pypy3 in the mix. Are you sure with two latest? Or maybe 3.8 only for windows-latest, macos-latest; leave pypy3 for ubuntu-latest? 3.8 and pypy3 for windows-latest and macos-latest?

rht:
Numba doesn't seem to work on 3.9 yet, but I don't know of any projects using Mesa that also use Numba.

rht:
> Numba doesn't seem to work on 3.9 yet, but I don't know of any projects using Mesa that also use Numba.

Nvm, they just fixed it: https://github.com/numba/numba/issues/6345 ~5 days ago, so not yet in release.

Corvince:
I am all for merging this as-is

Corvince (approved):

PR #972, opened 2021-01-02
GA: Add pypy3

rht:


rht:
Failed to install `black` on `pypy3`.

rht:
https://github.com/psf/black/issues/727

Corvince:
*sigh* seems like it will take a bit until black is supported. Not sure if its worth the hassle to set up an action that just runs pylint without black on pypy, but if you @rht or someone else gives it a shot I think it would be worth to add.

rht:
Black is not the only problem. I used the older version of Black as per https://github.com/psf/black/issues/727#issuecomment-547338207. Then the CI seemed to be stuck at installing Pandas.

rht:
Wow, it succeeded. It was slow to install the dependencies (~12min) but not more than that. If we enable caching I think adding pypy3 to CI will be feasible.

jackiekazil:
@rht is there another PR for enabling caching? (Would like to merge this.)

rht:
There is some complication with caching. Since there is no requirements.txt, then setup.py has to be hashed. As long as setup.py isn't changed, the cache persists. This means that we are no longer using the latest versions of the dependencies anymore. Most libraries pin their requirements to specific versions, and update them periodically, but it's not the case with Mesa.

(For reference, this is an example of caching: https://github.com/django-ftl/fluent-compiler/blob/a6048b856a3db0761ae05a9d3049489f6bb11cbe/.github/workflows/pythonpackage.yml#L29)

Corvince:
Mesa has a pipenv file we can use: https://github.com/actions/cache/blob/main/examples.md#python---pipenv

The pipfile.lock could use some love though... But we could set up dependabot to automatically create merge requests on new versions

Corvince:
Just as an idea we could run this in another github action, then we also don't need to use an old version of black.

rht:
I added cache to this PR as a proof of concept. It worked. Dependencies install down to 43s from 12min.

> Just as an idea we could run this in another github action, then we also don't need to use an old version of black.

Or put black installation in a separate step, and pin the version if matrix.python-version is pypy3.

Corvince:
> Or put black installation in a separate step, and pin the version if matrix.python-version is pypy3.

Better yet: Run black only on a single python version, since the output for different python versions would be the same anyway. I can make that change once this is merged.

While this POC works, I think it will fail as soon as we add windows and macOS, since the pip cache directory would change. Maybe also run pypy on ubuntu, anyways? (You can resolve this either in this PR or the windws/macos one). As soon as you give feedback on this and delete the "testing whitespace" from the readme, I think this is good to merge.

rht:
> Maybe also run pypy on ubuntu, anyways?

Did you mean "maybe only run pypy on ubuntu"?

Corvince:
Yes, sorry

Corvince:
@rht Do you want to update this PR since #970 is now merged or should someone else take over?

rht:
Updating it now.

rht:
Updated. The work to do Black only on a specific Python version can be done in a separate PR.

PR #974, opened 2021-01-03
GA: Only run CI when pushed to master

rht:
See #973 for example of wasteful CI running because it is pushed to a non-master branch.


jackiekazil:
LGTM

PR #996, opened 2021-02-19
GitHub Actions: run black only on ubuntu 3.8

Corvince:
Follow-up from #972 

Hopefully this will resolve any GA actions for the foreseeable future.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -37,18 +37,16 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        # We pin black's version until either PyPy3 becomes Python 3.8 or black
-        # removes its typed-ast dependency. See
-        # https://github.com/psf/black/issues/727
-        pip install codecov black==19.3b0
         pip install .[dev]
     - name: Lint with flake8
       run: |
         # Use settings from mesas .flake8 file
         flake8 . --count --show-source --statistics
     - name: Lint with black
+      if: matrix.python-version == '3.8'
       run: |
-        black --check --exclude=mesa/cookiecutter-mesa/*
+        pip install black
+        black --check --exclude=mesa/cookiecutter-mesa/* .
     - name: Test with pytest
       run: |
         pytest --cov=mesa tests/ --cov-report=xml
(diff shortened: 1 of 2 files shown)


Corvince:
Interestingly black wasn't configured correctly until now on GitHub actions (a `.` was missing). This caused a formatting error to slip through. Added another commit to solve.

.github/workflows/python.yml
```diff
@@ -37,18 +37,16 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        # We pin black's version until either PyPy3 becomes Python 3.8 or black
-        # removes its typed-ast dependency. See
-        # https://github.com/psf/black/issues/727
-        pip install codecov black==19.3b0
```
rht, line 43: codecov appears to be not installed in the new changes.

rht: Looks like `pip install .[dev]` does it?
```
    "dev": ["coverage", "flake8", "pytest >= 4.6", "pytest-cov", "sphinx"],
```
coverage? pytest-cov?

Corvince: > Looks like `pip install .[dev]` does it?

Yes, coverage is already included in [dev], so it was redundant to install it separately.

> coverage? pytest-cov?

Not sure what you want to comment here. Please try to formulate sentences, it makes communication much clearer and easier for everyone.

rht: > coverage? pytest-cov?

I just wasn't sure which of the packages in "dev" does the codecov install, since coverage/pytest-cov is not codecov, but probably has codecov as a dependency.

rht: It was already clear from the context what I was referring to when I said the module names with question marks.

rht: My point is: don't unnecessarily accuse people.

jackiekazil: It is hard to determine tone with just text. I don't think @Corvince was being accusatory. I didn't understand what you were referring to either. This was a misunderstanding.

jackiekazil:
@Corvince I _believe_ this is ready for a merge. Is that correct?

.github/workflows/python.yml
rht: Since more than one person were not able to understand what I said, then it is my mistake. I am sorry.

Corvince:
> @Corvince I _believe_ this is ready for a merge. Is that correct? 

Yes, should be

Issue #958, opened 2020-11-19
Using numpy.random instead of the build-in random module


Kubasinska:
**What's the problem this feature will solve?**
In my projects I have often implemented [numpy.Random](https://numpy.org/doc/stable/reference/random/generator.html) instead of using the random generator already implemented in the base class model.Model. The numpy.Random module offers much more functionality, an example: extracting a number from a poisson distribution. 

**Describe the solution you'd like**
What do you think? I can work on the implementation.

jackiekazil:
This is an interesting question. If memory serves me right, numpy.random.seed() isn't thread safe. I don't believe we are using multi-threads at this time, but I could see a world where we would.... so I am on the fence on this. 

But you bring up some good points. 

This feels a little over the top, but I am wondering if we would want to support both and maybe make it a settings function that triggers which one is used. Thoughts? 

(I would also want one of the other maintainers to jump in.)

Kubasinska:
I think a "switch" is the best option in case we want to implement this feature. As I have already said, I can implement it without any problem, therefore I am waiting for further updates.

Corvince:
I am not really convinced by this. Sure, I think it often makes sense to use numpy.random, but I don't see the general benefit of adding such complexity and dependency (we only depend on numpy for the continous grid atm).

Isn't this readily solved by a single one-liner in ones `model.__init__` function a lá?

```python
self.random = np.random.default_rng(seed)
```

I think it would probably make sense to put this in the useful snippets section of the docs, but not more. (just my opinion)

tcdejong:
I'm not concerned about thread safety. Given Python's GIL, concurrency primarily speeds up programs if it lets you avoid waiting for IO operations and the likes, if sharing state is required. That's not usually the bottleneck for ABMs. In other words, In other words, parallelizing the execution of _one_ model is hard. 

A much more realistic approach to concurrency is executing multiple configurations in parallel, which each execution being fully single threaded. This eliminates the difficulties of sharing state with multiprocessing, the GIL with multithreading, and provides more flexibility with proposals like this one. 

TLDR: I don't think thread-safety is a valid argument against this proposal.

jackiekazil:
@tcdejong Agreed. But adding more complexity that is not clear can be more difficult for anyone trying to test out parallelizing the execution. 

I am currently leaning with @Corvince on the complexity argument, but could be talked out of it. 

@dmasad @tpike3 @rht -- any of you have strong feels one way or another?

[...]

PR #998, opened 2021-02-25
[docs improvement] Add snippet about using `numpy`'s random

majdal:
As described in #958


Corvince:
Thanks for the PR @majdal! But be careful, you are "just" setting an additional random module for the agent, so you will have `agent.random` and `model.random` and if you want reproducible runs you have to set both seeds. I think a saner approach would be to override `model.random` with with the numpy module.

jackiekazil:
@majdal also, can you submit the PR to the docs branch, so we can do a test build there? 
_Sorry, we should probably document that in our contributor guide._

majdal:
@Corvince Actually, I was thinking about this when I used numpy's random in my own model. I was checking [this line](https://github.com/projectmesa/mesa/blob/master/mesa/agent.py#L29), and it seems like `agent.random` is just `agent.model.random`, and since we're setting the `model.random`, we wouldn't need to modify the model class directly, no?

@jackiekazil I'll do that now! And yes, including this in the contributor guide would be super useful.

Corvince:
> @Corvince Actually, I was thinking about this when I used numpy's random in my own model. I was checking [this line](https://github.com/projectmesa/mesa/blob/master/mesa/agent.py#L29), and it seems like `agent.random` is just `agent.model.random`, and since we're setting the `model.random`, we wouldn't need to modify the model class directly, no?
> 

Yes you are technically correct! And while this is the best kind of correct, I think it is also super confusing and counterintuitive. I still would prefer to modify `model.random` directly :)

rht:
I cherry-picked the last commit of this PR in #1204. I will do the further improvement in that PR.

PR #1001, opened 2021-03-04
Public remove_agent function for NetworkGrid

### LINKED ISSUE #1000: No public version of NetworkGrid._remove_agent, though code exists
Currently it is not possible to remove an Agent from a NetworkGrid using a **public** function. Currently, you can do the following:

```python
...
model.grid = NetworkGrid(G)
...
a = Agent(...)
model.grid._remove_agent(a, a.pos)
...
```
But this function is a private function, therefore it isn't shown in documentation or in your editor. I'd like to be able to remove agents like with the other space modules.

I'd like a similar solution as used for the other space modules:

Solution in space.Grid:
```python
def remove_agent(self, agent: Agent) -> None:  # solution in space.Grid
        """ Remove the agent from the grid and set its pos variable to None. """
        pos = agent.pos
        self._remove_agent(pos, agent)
        agent.pos = None

```
Solution for NetworkGrid:
```python
def remove_agent(self, agent: Agent) -> None:  # New solution
        """ Remove the agent from the network and set its pos variable to None. """
        pos = agent.pos
        self._remove_agent(agent, pos)
        agent.pos = None

```

**Additional context**
The _remove_agent function in NetworkGrid
```python
def _remove_agent(self, agent: Agent, node_id: int) -> None:
        """ Remove an agent from a node. """

        self.G.nodes[node_id]["agent"].remove(agent)
```

Casper-Smet:
Closes #1000.

This pull request implements a new feature: a public function for removing an agent from a NetworkGrid. For more, see issue #1000. 

The pull request only adds a view lines of code, most of which was just a slightly altered version of what already existed. In particular, this code exposes the `_remove_agent` in the same way that the `space.Grid` superclass does.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -871,6 +871,12 @@ class NetworkGrid:
 
         self.G.nodes[node_id]["agent"].remove(agent)
 
+    def remove_agent(self, agent: Agent) -> None:
+        """ Remove the agent from the network and set its pos variable to None. """
+        pos = agent.pos
+        self._remove_agent(agent, pos)
+        agent.pos = None
+
     def is_cell_empty(self, node_id: int) -> bool:
         """ Returns a bool of the contents of a cell. """
         return not self.G.nodes[node_id]["agent"]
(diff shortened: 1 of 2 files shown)


Corvince:
Thanks for adding this @Casper-Smet 
Quite interesting that this went unnoticed for so long!

Issue #969, opened 2021-01-01
Rename 'master' branch to 'main'


rht:
https://github.com/github/renaming

Corvince:
TLDR: We should wait a little bit longer until there is a renaming tool available (scheduled for January 2021). Once it is available we should consider the renaming

rht:
The wait is over: https://github.com/github/renaming/commit/040636ea50e945875c70cc964d586f645e9bd810.

Corvince:
Great!

jackiekazil:
I am onboard. I will bring it up during our dev meeting, which got bumped to this week to see if there are any concerns.

jackiekazil:
This was discussed during the meeting and everyone is onboard. I am ready to change, but was wondering if we need to consider timing or announce it first or just the "main" people doing dev is good enough. 

PRs are updated automatically, but local branches will need to be updated. 

```
git branch -m master main
git fetch origin
git branch -u origin/main main
```

dmasad:
Worth putting a note on top of the readme for a little while?

jackiekazil:
master is now main. 
https://github.com/projectmesa/mesa/branches

PR #1011, opened 2021-03-22
CI: Run on pushed to main and release branches

EwoutH:
Now that the default branch is changed to `main` in #969, the GitHub Actions CI should be updated as well to run on pushes to that branch. It now also runs on pushes to release branches (starting with `release`).


dmasad:
Good catch, thanks!

PR #707, opened 2019-07-07
examples/Readme.md: Update description to be consistent with the folder names

rht:

PR #1012, opened 2021-03-25
Docs/examples: Update links to use main instead of master as branch

EwoutH:
I noticed that a few of the links and buttons were broken while reading the documentation, which was caused by renaming the default branch from `master` to `main` (in #969). This PR aims to fix all those broken links. I think I got them all, but please check carefully.

One link we have to watch carefully is http://mesa.readthedocs.org/en/master/. /main/ doesn't work yet, but it could break on the next release or docs update.

**Edit:** In commit e4855d3 from #707 the `examples/Schelling` directory was renamed to `examples/schelling`, decapitalizing it. This PR now also fixed a few of these cases.


EwoutH:
@dmasad Could you (assign a) review(er)?

rht:
LGTM

The changes in this PR only modify either the documentation, or the docstrings in a code, and so won't change the behavior of the code and tests.

Note: the CI test failed because of an unrelated problem: space.py and test_grid.py haven't been reformatted with Black yet.

PR #1018, opened 2021-04-05
CI: Add workflow to update Pipfile.lock every month

EwoutH:
Adds a workflow that generates a up to date lockfile based on the current Pipfile. The workflow runs every 1st day of each month, and if either this configuration or the Pipfile itself is modified. A run will open a PR with the updated pipfile.lock, and if a PR is already open it will update that PR.

By running and opening/updating a PR each month it's ensured the pipfile.lock stays up to date and updated frequently at convenient times, while reproducible builds are still ensured.

The schedule can be adjusted to fit development, of course.

### DIFF
--- a/.github/workflows/update-pipfile.yml
+++ b/.github/workflows/update-pipfile.yml
@@ -13,6 +13,8 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - uses: actions/setup-python@v2
+      with:
+        python-version: 3.7
     - run: pip install wheel
     - run: pip install pipenv
     - run: pipenv lock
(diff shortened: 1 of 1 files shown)


EwoutH:
An example of how such a monthly PR will look can be found here: https://github.com/EwoutH/mesa/pull/3

This can also replace all the separate PRs by Dependabot.

jackiekazil:
Where can I find more info on `peter-evans/create-pull-request@v3`

EwoutH:
Hi @jackiekazil! It's quite a frequently GitHub Action to open pull request from GitHub actions jobs. You can find more information about it here: https://github.com/peter-evans/create-pull-request

jackiekazil:
LGTM

PR #1023, opened 2021-04-17
Black clean up.

jackiekazil:


rht:
Black seems to require `typed_ast`, which breaks the PyPy setup. Can't use Black in PyPy until PyPy supports 3.8.

PR #1072, opened 2021-11-03
Tweaking and improving the documentation

dmasad:
Tweaking autodocs configuration to make the API documentation on ReadTheDocs better. 

You can verify how it looks here:
https://mesa.readthedocs.io/en/oct_2021_doc_update/apis/api_main.html


jackiekazil:
As discussed -- the issue is with black. I will merge and add another PR to fix black.

PR #1074, opened 2021-11-04
CI: Maintenance: Update to Python 3.10, split of lint jobs

EwoutH:
- Update main build jobs to Python 3.10
- Split of lint steps in separate jobs, so that the other jobs passes if the lint jobs fails
- Install black[jupyter] to also lint Jupyter notebooks with black

The new build matrix now looks like this:
Screenshot_671

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -9,19 +9,22 @@ on:
 
 jobs:
   build:
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}-latest
     strategy:
+      fail-fast: False
       matrix:
-        os: [windows-latest, ubuntu-latest, macos-latest]
-        python-version: [3.9]
+        os: [windows, ubuntu, macos]
+        python-version: ["3.10"]
         include:
-          - os: ubuntu-latest
-            python-version: 3.6
-          - os: ubuntu-latest
-            python-version: 3.7
-          - os: ubuntu-latest
-            python-version: 3.8
-          - os: ubuntu-latest
+          - os: ubuntu
+            python-version: "3.9"
+          - os: ubuntu
+            python-version: "3.8"
+          - os: ubuntu
+            python-version: "3.7"
+          - os: ubuntu
+            python-version: "3.6"
+          - os: ubuntu
             python-version: pypy3
 
     steps:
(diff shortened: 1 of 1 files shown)


EwoutH:
@jackiekazil Hi! Could you review this PR?

.github/workflows/python.yml
```diff
@@ -9,47 +9,66 @@ on:
...
   build:
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}-latest
     strategy:
+      fail-fast: False
       matrix:
-        os: [windows-latest, ubuntu-latest, macos-latest]
-        python-version: [3.9]
+        os: [windows, ubuntu, macos]
+        py: ["3.10"]
```
rht, line 17: Sticking to `python-version` is probably clearer.

EwoutH: Done!

EwoutH:
@jackiekazil I rebased this PR, and with #1078 merged the CI now passes! With this PR Jupyter notebooks are now also linted, so no formatting errors should get through.

.github/workflows/python.yml
```diff
@@ -36,20 +39,36 @@ jobs:
...
+  lint-black:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.10
+      uses: actions/setup-python@v2
+      with:
+        python-version: "3.10"
+    - name: Install dependencies
+      run: pip install .[dev]
```
rht, line 71: Is this line necessary for linting? The following line already does the Black installation.

.github/workflows/python.yml
```diff
@@ -36,20 +39,36 @@ jobs:
...
+  lint-flake:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.10
+      uses: actions/setup-python@v2
+      with:
+        python-version: "3.10"
+    - name: Install dependencies
+      run: pip install .[dev]
```
rht, line 57: Same question about the necessity as below.

.github/workflows/python.yml
jackiekazil: Good catch. Is there a reason why we wouldn't just add the black[juypter] to dev setup? And get rid of the second line?

.github/workflows/python.yml
jackiekazil: This line installs it. The one below runs it. So I think this is necessary.

.github/workflows/python.yml
rht: It may take longer time to install all the packages in `.[dev]`. It is better to just install the required dependencies to save time.

rht: Especially since the dependencies are not cached in this job.

jackiekazil: Fair point. 
@EwoutH can you adjust?

EwoutH: Thanks, good catch!

EwoutH:
@rht Thanks for the critical review, appreciate it! Now only the necessary dependencies are installed, so `flake8` and `black[jupyter]` for both lint jobs respectively. Saves about 25 seconds of runtime on both jobs.

EwoutH:
@jackiekazil This PR is ready for review and merge!

jackiekazil:
LGTM

EwoutH:
Thanks for merging!

PR #1083, opened 2021-11-29
CI: Only run Codecov on Ubuntu jobs and update to v2

EwoutH:
It isn't necessary to run Codecov on all jobs, so only run them on the Ubuntu jobs.

Also updates the [codecov/codecov-action](https://github.com/codecov/codecov-action) to v2. v1 will be [depreciated](https://github.com/codecov/codecov-action/tree/v2.1.0#%EF%B8%8F--deprecration-of-v1)  on February 1, 2022.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -42,8 +42,9 @@ jobs:
       run: pip install .[dev]
     - name: Test with pytest
       run: pytest --cov=mesa tests/ --cov-report=xml
-    - name: Codecov
-      uses: codecov/codecov-action@v1.0.15
+    - if: matrix.os == 'ubuntu'
+      name: Codecov
+      uses: codecov/codecov-action@v2
 
   lint-flake:
     runs-on: ubuntu-latest


.github/workflows/python.yml
```diff
@@ -42,8 +42,9 @@ jobs:
       run: pip install .[dev]
     - name: Test with pytest
       run: pytest --cov=mesa tests/ --cov-report=xml
-    - name: Codecov
-      uses: codecov/codecov-action@v1.0.15
+    - if: matrix.os == 'ubuntu'
```
rht, line 45: Should we limit it to 1 Python version on Ubuntu, instead of multiple Python version on Ubuntu? Probably the oldest supported version is prioritized.

EwoutH: It seems to be recommended to upload them al, since they can differ between versions. And this PR already reduces the number of reports uploaded.

https://community.codecov.com/t/upload-of-coverage-of-several-python-versions/2349

tpike3:
LGTM

PR #1090, opened 2021-12-01
Docs: space.py: Fix single case of neighbor spelled as neighbour

EwoutH:
Fix single case of neighbor spelled as neighbour in the space.py docs.


tpike3:
LGTM

PR #1091, opened 2021-12-01
Docs: Add back some comments in space.py

EwoutH:
Improve the space.py docs by adding back some comments (after some were removed in #1072).

Apparently, starting directly with `Args:` results in the function not being included in the docs. This fixes it by always starting with a short description (at least in space.py).

Also make the whitelines at the begin and end of comments consistent.


tpike3:
Looks Good To Me

PR #1099, opened 2021-12-07
Convert all text strings to f-strings

EwoutH:
Reformats all the text from the old "%-formatted" and .format(...) format to the newer f-string format, as defined in [PEP 498](https://www.python.org/dev/peps/pep-0498/). This Requires Python 3.6+.

[Flynt](https://github.com/ikamensh/flynt) 0.69 was used to reformat the strings. 16 f-strings were created in 10 files.

F-strings are in general more readable, concise and performant. See also: [PEP 498#Rationale](https://www.python.org/dev/peps/pep-0498/#Rationale).

```
Running flynt v.0.69
Can't parse mesa\mesa\cookiecutter-mesa\{{cookiecutter.snake}}\run.py as a python file.
Can't parse mesa\mesa\cookiecutter-mesa\{{cookiecutter.snake}}\{{cookiecutter.snake}}\model.py as a python file.

Flynt run has finished. Stats:

Execution time:                            1.377s
Files modified:                            10
Character count reduction:                 206 (0.06%)

Per expression type:
Old style (`%`) expressions attempted:     2/2 (100.0%)
`.format(...)` calls attempted:            19/19 (100.0%)
No concatenations attempted.
F-string expressions created:              16
```

### DIFF
--- a/examples/boltzmann_wealth_model_network/boltzmann_wealth_model_network/server.py
+++ b/examples/boltzmann_wealth_model_network/boltzmann_wealth_model_network/server.py
@@ -16,7 +16,7 @@ def network_portrayal(G):
             "color": "#CC0000" if not agents or agents[0].wealth == 0 else "#007959",
             "label": None
             if not agents
-            else "Agent:{} Wealth:{}".format(agents[0].unique_id, agents[0].wealth),
+            else f"Agent:{agents[0].unique_id} Wealth:{agents[0].wealth}",
         }
         for (node_id, agents) in G.nodes.data("agent")
     ]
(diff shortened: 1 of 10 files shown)


EwoutH:
> The diff size for each file is way too big.

Yeah I agree, but I have no idea why.

> Is there a way to ensure only the lines with the string formattings are changed?

Actually, if you look at the split diffs, that is the case. For some reason when more than one line is changed, it marks the whole file as different.

EwoutH:
I fixed it! I was working on a Windows system, and apparently that sometimes [handles line endings](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings#global-settings-for-line-endings) in an other way. I ran `git config --global core.autocrlf true`, ran git reset and recommitted, and now the diff is correct!

rht (approved):

tpike3:
Very cool thanks! Good note about the windows and there issues with line endings too.

PR #1105, opened 2021-12-14
Pyupgrade 3.6: Update syntax with Python 3.6+ features

EwoutH:
Updated Python code with Python 3.6+ features:
- utf-8 encoding is now the default ([PEP 3120](https://www.python.org/dev/peps/pep-3120/), Python 3.0+)
- Replace list comprehensions by Generator Expressions ([PEP 289](https://www.python.org/dev/peps/pep-0289/), Python 2.4+)
- Replace yield loop by yield from ([PEP 380](https://www.python.org/dev/peps/pep-0380/), Python 3.3+)

Used [pyupgrade](https://github.com/asottile/pyupgrade) 2.29.1 with `--py36-plus`.

### DIFF
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -62,7 +62,7 @@ class BaseScheduler:
 
         if agent.unique_id in self._agents:
             raise Exception(
-                "Agent with unique id {0} already added to scheduler".format(
+                "Agent with unique id {} already added to scheduler".format(
                     repr(agent.unique_id)
                 )
             )
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -62,9 +62,7 @@ class BaseScheduler:
 
         if agent.unique_id in self._agents:
             raise Exception(
-                "Agent with unique id {} already added to scheduler".format(
-                    repr(agent.unique_id)
-                )
+                f"Agent with unique id {repr(agent.unique_id)} already added to scheduler"
             )
 
         self._agents[agent.unique_id] = agent
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-# -*- coding: utf-8 -*-
 #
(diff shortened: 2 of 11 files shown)


mesa/time.py
```diff
@@ -62,7 +62,7 @@ def add(self, agent: Agent) -> None:
 
         if agent.unique_id in self._agents:
             raise Exception(
-                "Agent with unique id {0} already added to scheduler".format(
+                "Agent with unique id {} already added to scheduler".format(
```
rht, line 65: Why is this not in f-string?

EwoutH: For some reason it wasn't picked up by the automatic conversion. Maybe because it the value uses `repr()` string conversion of the object. I formatted it as a f-string!

rht:
LGTM

mesa/time.py
rht: Note that there are still other places where `.format()` is used:
```
batchrunner.py
27:        return self.MESSAGE.format(self.bad_names)

visualization/UserParam.py
71:        msg = self._ERROR_MESSAGE.format(self.param_type, name)

visualization/modules/PieChartVisualization.py
59:        new_element = new_element.format(fields_json, canvas_width, canvas_height)

visualization/modules/HexGridVisualization.py
71:        new_element = "new CanvasHexModule({}, {}, {}, {})".format(

visualization/modules/BarChartVisualization.py
67:        new_element = new_element.format(

visualization/modules/ChartVisualization.py
72:        new_element = new_element.format(series_json, canvas_width, canvas_height)

visualization/modules/CanvasGridVisualization.py
87:        new_element = "new CanvasModule({}, {}, {}, {})".format(
```

rht: The template string ones are not changed for a good reason, but the `new Canvas...` ones should be changed to f-strings.

rht: This is out of scope for this PR though.

PR #1110, opened 2021-12-20
bank_reserves: Say that the commented out legacy code is for comparison

rht:


rht:
Bump. @tpike3 @jackiekazil

tpike3:
Sorry for the delay..Thanks for this!

PR #1113, opened 2021-12-21
Fixing batch_run bug

### LINKED ISSUE #1112: batch_run only data collection only works if there are agent_reporters
**Describe the bug**

The new `batch_run()` function does not return any results if there are only `model_reporters` in the model's `DataCollector`.

**Expected behavior**

It should return the data collected from the `model_reporters`.

The problem appears to be in `_model_run_func()`. I will work on a PR.

jdfoote:
There may be a more elegant fix but this just checks to make sure there are agent variables being collected. If not, it just stores the model data.

Fixes #1112 

Apologies for not knowing how to just include the correct commit. :)


mesa/batchrunner.py
```diff
@@ -172,10 +172,15 @@ def _model_run_func(
...
-        stepdata = [
-            {**{"Step": step}, **kwargs, **model_data, **agent_data}
-            for agent_data in all_agents_data
-        ]
+        if all_agents_data:
+            stepdata = [
+                {**{"Step": step}, **kwargs, **model_data, **agent_data}
+                for agent_data in all_agents_data
+            ]
+        else:
```
rht, line 180: Needs a comment describing what this else step does.

rht:
You need to squash all the 4 commits into 1.

rht:
The commit message needs to be more descriptive.

jdfoote:
I created a new PR at #1115 and I'll close this one.

PR #1115, opened 2021-12-22
 Adding logic to check whether there is agent data

jdfoote:
There was a bug where the model output was only produced
 if there were agent_reporters.

 This is because the output data was created in a list
 comprehension that ran once for every agent in
 agent_reporters.

 The new logic checks if there are agent reporters;
 if not, it just outputs the model_reporters.

 Also adds a simple unit test to make sure that output
 is produced when there aren't agent_reporters.

Fixes #1113

### DIFF
--- a/mesa/batchrunner.py
+++ b/mesa/batchrunner.py
@@ -172,10 +172,15 @@ def _model_run_func(
     for step in steps:
         model_data, all_agents_data = _collect_data(model, step)
 
-        stepdata = [
-            {**{"Step": step}, **kwargs, **model_data, **agent_data}
-            for agent_data in all_agents_data
-        ]
+        # If there are agent_reporters, then create an entry for each agent
+        if all_agents_data:
+            stepdata = [
+                {**{"Step": step}, **kwargs, **model_data, **agent_data}
+                for agent_data in all_agents_data
+            ]
+        # If there is only model data, then create a single entry for the step
+        else:
+            stepdata = [{**{"Step": step}, **kwargs, **model_data}]
         data.extend(stepdata)
 
     return tuple(kwargs.values()), data
--- a/tests/test_batch_run.py
+++ b/tests/test_batch_run.py
@@ -47,6 +47,7 @@ class MockModel(Model):
         variable_agent_param=None,
         fixed_model_param=None,
         schedule=None,
+        enable_agent_reporters=True,
         **kwargs
     ):
         super().__init__()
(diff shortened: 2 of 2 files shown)


mesa/batchrunner.py
```diff
@@ -172,10 +172,15 @@ def _model_run_func(
...
-            {**{"Step": step}, **kwargs, **model_data, **agent_data}
-            for agent_data in all_agents_data
-        ]
+        # If there are agent_reporters, then create an entry for each agent
+        if all_agents_data:
+            stepdata = [
+                {**{"Step": step}, **kwargs, **model_data, **agent_data}
+                for agent_data in all_agents_data
+            ]
+        # Otherwise, just create a single entry for the step
```
rht, line 181: 1. You didn't say that the single entry is from the model data.
2. Actually, it is confusing that the model data is put in as if it were another agent data

rht: OK, nvm, upon reading the code more, I saw that the model data is put in as input even in the case when there are agent_data.

rht:
Remaining TODOs:
- You need to squash the 8 commits into 1.
- If you want to update your git branch, do it with `git pull --rebase`. This is cleaner and without the messy merge commits.

mesa/batchrunner.py
rht: Although, I would additionally say in the comment that the entry contains only model data, without agents data.

rht:
This is how I usually squash my commits: https://stackoverflow.com/a/5189600.

jdfoote:
Hahaha - I worked so hard to try to squash these and didn't even notice that I got it wrong! I'll try again. :)

rht:
Looks like you managed to successfully squash.

tests/test_batch_run.py
```diff
@@ -47,6 +47,7 @@ def __init__(
         variable_agent_param=None,
         fixed_model_param=None,
         schedule=None,
+        agent_reporters=True,
```
rht, line 50: This variable is a boolean, and then the type changes into a dictionary, which is actually used for a different purpose. This is confusing. It's best to rename this boolean variable to `enable_agent_reporters` or something.

rht:
The coverage report is outdated. I tested your change locally, and that `else` branch is actually covered by the newly added test.
Remaining TODOs:
- Rename `agent_reporters=True` to `enable_agent_reporters=True`
- Use imperative mood for the commit messages. You can reword the commit message by doing the same `git rebase -i HEAD~n`, and then change `pick` into `reword` for the commits you want to reword
 
See https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project
In that article, you can search for the paragraph that contains this sentence for more detail:
>  Write your commit message in the imperative: "Fix bug" and not "Fixed bug" or "Fixes bug."

jdfoote:
@rht Thanks so much for patiently helping me to figure this process out.

I changed the commit message for the primary commit. Should I squash again so that all of the changes are in just one commit?

rht:
Yeah, you should squash them into 1 commit.

tpike3:
@jdfoote Thanks for the update! and @rht Thanks for the reviews and support!

LGTM

PR #1116, opened 2021-12-23
[PERF] Improve move_to_empty performance

### LINKED ISSUE #1052: SingleGrid position_agent very slow for large grids
It's fine for square widths of around 20 cells, but when you 200 it slows right down.

It appears to be caused by the line `agent.random.choice(sorted(self.empties))` where it sorts the empties first. I guess it does this so you can replicate the results using random seed?

If you don't care about replication however, it's unreasonably slow and can be fixed by using `agent.random.choice(list(self.empties))` instead (speeding it enormously).

I'm not sure of the design decisions here, but maybe it's worth having two functions to make it clear or passing some sort of bool?

https://mesa.readthedocs.io/en/stable/mesa.html#mesa.space.SingleGrid.position_agent

rht:
This fixes #1052. This is based on https://github.com/JuliaDynamics/Agents.jl/pull/541 (cc @Libbum).
Note:
- The first commit is #1110. I will rebase this PR once that one is merged
- I need to measure the speedup vs without `sorted(self.empties)`, since the `sorted()` operation is the slowest part as reported in #1052
- I haven't yet measured the cutoff value that result in a break-even with the old method. It seems to be smaller than Agents.jl's 0.998

I benchmarked this on the Schelling model.

benchmark
The black line is a horizontal line at y = 1.
Haven't had the time to repeat the run multiple times to get a plot with standard deviation, but at least here you can see how much the speedup increases as we increase the grid size.

```python
densities = np.linspace(0.01, 0.9, 20)
for L in [5, 10, 15, 20]:
    speedups = []
    for density in densities:
        model_params = {
            "height": L,
            "width": L,
            # Agent density, from 0.8 to 1.0
            "density": density,
            # Fraction minority, from 0.2 to 1.0
            "minority_pc": 0.2,
            # Homophily, from 3 to 8
            "homophily": 3,
        }
        elapseds = []
        for cutoff in [0.0, 0.998]:
            model = Schelling(move_to_empty_cutoff=cutoff, **model_params)
            tic = time.time()
            for i in range(100):
                model.step()
            elapsed = time.time() - tic
            elapseds.append(elapsed)
        speedup = elapseds[0] / elapseds[1]
        speedups.append(speedup)
    plt.plot(densities, speedups, label=f"L={L}")
plt.axhline(1, color="black", label="1")
plt.legend()
plt.xlabel("Agent density")
plt.ylabel("Speedup (current main branch / this PR)")
plt.savefig("benchmark.png")
```

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -39,6 +39,12 @@ GridContent = Union[Optional[Agent], Set[Agent]]
 FloatCoordinate = Union[Tuple[float, float], np.ndarray]
 
 
+def clamp(x: float, lowest: float, highest: float) -> float:
+    # This should be faster than np.clip for a scalar x.
+    # TODO: measure how much faster this function is.
+    return max(lowest, min(x, highest))
+
+
 def accept_tuple_argument(wrapped_function):
     """Decorator to allow grid methods that take a list of (x, y) coord tuples
     to also handle a single position, by automatically wrapping tuple in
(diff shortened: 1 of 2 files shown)


rht:
The next question is to whether remove the grid's `self.empties` attribute altogether. Currently, the remaining use of this attribute is that to check whether the grid is full in the function `position_agent`. Agents.jl doesn't do this. And I can see that instead of checking that the grid is full every time, we can do during the initializatio to check that the ratio between the number of agents and the grid area must not exceed 1. Thoughts?

This decision needs to be resolved before I can fix the tests.

tpike3:
Sorry, it took me awhile to really think through this. I don't think we should get rid of self.empties as my sense is it will break a lot of models. 

On question in comments you say Agents.jl uses 0.998 as the cutoff score but say the number is different here. Not being familiar with how Julia stores numbers are you saying that although the numbers are the *same* the way they are stored is actually comparing a different value?

rht:
> I don't think we should get rid of self.empties as my sense is it will break a lot of models.

I have checked `self.empties` usage. It is only used in `move_to_empty` and `position_agent` (when x and y are 'random'). It shouldn't be a public attribute that Mesa users should use either.

rht:
> On question in comments you say Agents.jl uses 0.998 as the cutoff score but say the number is different here. Not being familiar with how Julia stores numbers are you saying that although the numbers are the same the way they are stored is actually comparing a different value?

My code comment is certainly misleading. I currently hardcode the cutoff to be 0.998 (same as Agents.jl) because I haven't done the search for the optimum number for Mesa. I will probably have to use the exact same code @Libbum use (translated to Python from Julia) to get the number.

Just wanting to emphasize that the speedup is much, much higher when the grid size is 200 x 200 (i.e. the problem raised in #1052), it's just that it is taking too long on my laptop to measure. I'd consider 200x200 to be a common laptop-scale problem.

rht:
@tpike3 I have made sure that the logical diff of this PR is as small as possible, and so this PR is ready for review. This means that `self.empties` is still present. But this PR still improves the speed of `move_to_empty`.
See
benchmark

rht:
I had to modify the test because the `move_to_empty` requires the total number of agents in order to compute the threshold. This requires access to the `model` object. Some of the the mock agents don't have `model` object attached to them.

I checked Agents.jl's implementation, and it operates on the model directly (https://github.com/JuliaDynamics/Agents.jl/pull/541/files#diff-007fa315900dda19e79dff8b7fe546cd50232c4faf79c5551ca37fa9fc9d1f3bR85), and so a grid is never independent of a model.

rht:
If we choose to always require a grid object to be tied to a model, this PR will become simpler.
The only drawback is that the grid & agents are no longer standalone, and so we sacrifice modularity.

I'm not aware of projects that use Mesa's grid & its `move_to_empty` without using Mesa's model and scheduler.
Thoughts?

tpike3:
Boy, I had to really chew on this one. 

I agree I have never seen a model that uses Mesa's grid & its move_to_empty without using Mesa's model and scheduler. So I am confident with backwards compatibility we are fine there.

I am hesitant to reduce the modularity by always having a grid, the seminal zero intelligence traders does not need one and I have done SIR models just on a network. In general, I think being able to use a network abstraction without a physical location is a useful option. So, I am in favor of keeping it modular. 

So @jackiekazil thoughts?  This LGTM if you agree and want to merge.

rht:
There are 2 directions of dependency here:
- a grid that depends on a model.
- a model that depends on a grid. While I read that Agents.jl has the `random_empty` requiring a model (with a discrete space) as an argument, they can totally have a model without physical location, too.

jackiekazil:
I did a review of this... yeah. I think we can merge and move the remaining to future discussions and PRs.

PR #1117, opened 2021-12-28
Fix numbering typos in docs/README.md

jonahaag:


jonahaag:
On a side note: is there a specific reason why this is the only markdown file of the project?

jackiekazil:
Ty for the fix. 
Also, I can't tell you why. Could be an artifact that was never changed?

rht:
The series of `1.`'s is actually not a typo. Markdown will render them as incrementally increasing numbers. On the contrary, it is better with `1.`'s because there is a potential human error with manually incrementing them.

jonahaag:
Thanks for pointing that out, I didn't know that.

However, I just double checked and the rendered file on GitHub did only have the `1.`'s for the most part since the bullet points in between were not indented properly (two tabs necessary here I think). As a result only `1's` that are not "separated" by bullet points were recognized as the same numbered list and thus incremented; all the others were treated as new numbered lists starting with `1.`. 

Should I open a new PR to go back to the old style and add the tabs or just leave as it is?

rht:
Ah, that's right. The numbers we see in https://github.com/projectmesa/mesa/blob/main/docs/README.md are probably not necessarily interpreted as numbered list by Markdown. Yeah, I think using `1.`'s and checking that they render to incremental numbers give a stronger guarantee that they are rendered as Markdown numbered list.

rht:
Ah, that's right. The numbers we see in https://github.com/projectmesa/mesa/blob/main/docs/README.md are probably not necessarily interpreted as numbered list by Markdown. Yeah, I think using `1.`'s and checking that they render to incremental numbers give a stronger guarantee that they are rendered as Markdown numbered list.

rht:
Also yes for fixing the whitespaces for nested lists.

PR #1118, opened 2021-12-29
Update docs/README.md

jonahaag:
As discussed in #1117, I went back to all `1.`'s in the numbered lists and instead let markdown increment automatically. For this to work properly, whitespaces in front of nested bullet points were replaced by two tabs.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -11,35 +11,35 @@ Updating docs can be confusing. Here are the basic setups.
 
 #### Create the rST files from ipynb files
 1. Change to the appropriate directory (usually docs/tutorials)
-  * ```cd tutorials```
-2. Create rST files using nbconvert
-  * ```jupyter nbconvert --to rST *.ipynb```
-  * **Requires**
-    * jupyter: `pip install jupyter`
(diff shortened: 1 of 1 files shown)


rht:
In order for your branch commit history to be clean (without merge commits), you should add this in your ~/.gitconfig:
```
[pull]
    rebase = true
```
This way, your branch's commits are re-applied on top of the main branch you pull from, instead of it doing messy merge operations.

rht:
Other than the extra merge commit issue (you can look up from the internet on how to delete the last commit, and then you do `git pull` again), the PR content LGTM.

jonahaag:
True, I am sorry for that. I forgot to do the rebase before and then solved the resulting conflict via merge commit using the GitHub web IDE, which was just me being lazy

jonahaag:
I hope this is fine now, took me a while to figure out how to do this in a clean way. 

Changes remain the same for the most part, I also removed some unnecessary code marks to keep everything consistent

rht:
LGTM. As can be seen in https://github.com/jonahaag/mesa/tree/update-docs-readme/docs.

PR #1141, opened 2022-01-22
update number_processes and associated docs

tpike3:
This updates `nr_processes` to `number_processes` in batch_run and the associated tutorial.

I looked through Keras, Pytorch, multiprocessing to follow the *principle of least surprise*. Both the deep learning libraries use threads which did not seem appropriate. and so trying to mirror multiprocessing conventions I used `number_processes`. 

other changes: 
 - black updated the .pre-commit-config file
 - jupyterlab updated some hidden fields

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/ambv/black
-    rev: stable
+    rev: 21.12b0
     hooks:
     - id: black
       exclude: mesa/cookiecutter-mesa/*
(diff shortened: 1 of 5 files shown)


rht:
PyTorch does have multiprocessing module: https://pytorch.org/docs/stable/_modules/torch/multiprocessing/spawn.html#spawn, and they use `nrprocs`.

But they also say:
> The API is 100% compatible with the original module - it’s enough to change import multiprocessing to import torch.multiprocessing to have all the tensors sent through the queues or shared via other mechanisms, moved to shared memory.

So they are not being consistent.

tpike3:
@rht I completely missed that, thanks. I think with a lack of consistency and to your original point of making it more obvious, I am good with `number_processes`. But happy to defer to the group consensus.

jackiekazil:
LGTM.

PR #1142, opened 2022-01-22
PoC: Multiple-agent types scheduling and datacollection

EwoutH:
_Proof of concept. This code is not meant for production._

This commit adds support for multiple agent-types, including scheduling and data collection.

The scheduler is based on the `RandomActivationByBreed` scheduler from the `wolf_sheep` example. It supports RandomActivation by agent type, in the order that they are initialized. See [examples/wolf_sheep/wolf_sheep/schedule.py](https://github.com/projectmesa/mesa/blob/246c69d592a82e89c75d555c79736c3f619d434a/examples/wolf_sheep/wolf_sheep/schedule.py).

The DataCollection module is modified to support multiple agent types. It uses a dictionary structure for all agent reporters and records, with the agent class as the key.

The ChartModule has been extended to enable data collection both Agent and Model variables, and agents of different types. Contains currently a breaking change by requiring `"Type"` key in the ChartModule input dictionary, which can either be `"Model"` or `"Agent"`. If `"Agent"`, a key `"Agent_type"` is also required, which should be the exact name of the class as a string (`__class__.__name__`).

A Mesa Model class with a datacollector can now look like this:
```Python
class ExampleModel(Model):
    def __init__(self, n_dogs, n_cats):
        super().__init__()
        self.schedule = RandomActivationByType(self)

        self.population = n_dogs + n_cats

        for _ in range(self.n_cats):
            cat = Cat(self.next_id(), self)
            self.schedule.add(cat)

        for _ in range(self.n_dogs):
            dog = Dog(self.next_id(), self, population_dist)
            self.schedule.add(dog)

        model_reporters = {"Population": "population",
        agent_reporters = {Cat: {"Weight": "weight", "Size": "size"},
                           Dog: {"Weight": "weight"}}

            model_reporters=model_reporters,
            agent_reporters=agent_reporters)
```
Charts for a Mesa Server can now look like this:
```Python
chart_pop = ChartModule(
    [{"Type": "Model", "Label": "Population", "Color": "Black"}], data_collector_name="datacollector"
)
chart_weight = ChartModule(
    [{"Type": "Agent", "Agent_type": "Cat", "Label": "Weight", "Color": "Blue"},
     {"Type": "Agent", "Agent_type": "Dog", "Label": "Weight", "Color": "Red"}], data_collector_name="datacollector"
)
chart_size = ChartModule(
    [{"Type": "Agent", "Agent_type": "Cat", "Label": "Size", "Color": "Blue"}],
    data_collector_name="datacollector"
)
server = ModularServer(
    ExampleModel, [chart_pop, chart_weight, chart_size], "Cats and Dogs", {"n_dogs": 10, "n_cats": 10}
)
```

**To-do**
 - [ ] Scheduler: Allow custom order of agent types in staging (and not just the order in which their created).
 - [ ] DataCollector: Ensure it still works with only a single/default agent type
 - [ ] ChartVisualisation:
   - [ ] 1) Try to reduce breaking changes
   - [ ] 2) ensure it still works with a single/default agent-type
   - [x] 3) Allow different kinds of statistical processing then mean (median, sum, min, max, standard deviation, etc.) (maybe implement this in DataCollector)
 - [ ] Cleanup
 - [ ] Testing
 - [ ] Documentation (including examples)

### DIFF
--- a/mesa/datacollection.py
+++ b/mesa/datacollection.py
@@ -95,20 +95,31 @@ class DataCollector:
             {"Model_Function":[function, [param_1, param_2]]}
 
         """
-        self.model_reporters = {}
-        self.agent_reporters = {}
+        self.agent_types = agent_reporters.keys()
 
+        self.model_reporters = {}
         self.model_vars = {}
-        self._agent_records = {}
-        self.tables = {}
 
         if model_reporters is not None:
             for name, reporter in model_reporters.items():
                 self._new_model_reporter(name, reporter)
 
-        if agent_reporters is not None:
-            for name, reporter in agent_reporters.items():
-                self._new_agent_reporter(name, reporter)
+        self.agent_reporters = {}
+        self._agent_records = {}
+        self.agent_attr_index = {}
+        self.agent_name_index = {}
+
+        for agent_type in self.agent_types:
+            self.agent_reporters[agent_type] = {}
+            self._agent_records[agent_type] = {}
+            self.agent_name_index[agent_type] = {}
+        self.tables = {}
+
+        for agent_type in self.agent_types:
+            if agent_reporters[agent_type] is not None:
+                for name, reporter in agent_reporters[agent_type].items():
+                    self.agent_name_index[agent_type][name] = reporter
+                    self._new_agent_reporter(name, reporter, agent_type)
 
         if tables is not None:
             for name, columns in tables.items():
@@ -127,7 +138,7 @@ class DataCollector:
         self.model_reporters[name] = reporter
         self.model_vars[name] = []
 
-    def _new_agent_reporter(self, name, reporter):
+    def _new_agent_reporter(self, name, reporter, agent_type):
         """Add a new agent-level reporter to collect.
 
         Args:
@@ -140,7 +151,7 @@ class DataCollector:
             attribute_name = reporter
             reporter = partial(self._getattr, reporter)
             reporter.attribute_name = attribute_name
-        self.agent_reporters[name] = reporter
+        self.agent_reporters[agent_type][name] = reporter
 
     def _new_table(self, table_name, table_columns):
         """Add a new table that objects can write to.
@@ -153,13 +164,16 @@ class DataCollector:
         new_table = {column: [] for column in table_columns}
         self.tables[table_name] = new_table
 
-    def _record_agents(self, model):
+    def _record_agents(self, model, agent_type):
         """Record agents data in a mapping of functions and agents."""
-        rep_funcs = self.agent_reporters.values()
+        rep_funcs = self.agent_reporters[agent_type].values()
         if all([hasattr(rep, "attribute_name") for rep in rep_funcs]):
             prefix = ["model.schedule.steps", "unique_id"]
             attributes = [func.attribute_name for func in rep_funcs]
+            self.agent_attr_index[agent_type] = {k: v for v, k in enumerate(prefix + attributes)}
+            print(self.agent_attr_index[agent_type])
             get_reports = attrgetter(*prefix + attributes)
+
         else:
 
             def get_reports(agent):
@@ -167,7 +181,7 @@ class DataCollector:
                 reports = tuple(rep(agent) for rep in rep_funcs)
                 return _prefix + reports
 
-        agent_records = map(get_reports, model.schedule.agents)
+        agent_records = map(get_reports, model.schedule.agents_by_type[agent_type].values())
         return agent_records
 
     def _reporter_decorator(self, reporter):
@@ -190,9 +204,10 @@ class DataCollector:
                 else:
                     self.model_vars[var].append(self._reporter_decorator(reporter))
 
-        if self.agent_reporters:
-            agent_records = self._record_agents(model)
-            self._agent_records[model.schedule.steps] = list(agent_records)
+        for agent_type in self.agent_types:
+            if self.agent_reporters[agent_type]:
+                agent_records = self._record_agents(model, agent_type)
+                self._agent_records[agent_type][model.schedule.steps] = list(agent_records)
 
     def add_table_row(self, table_name, row, ignore_missing=False):
         """Add a row dictionary to a specific table.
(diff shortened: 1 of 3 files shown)


rht:
The Agents.jl implementation is much simpler and modular, because they can just use multiple dispatch. :/

EwoutH:
I read a bit up about multiple dispatch, and it indeed seems very useful in this case to be able to handle both the model and different agent types.

A few Python packages implement this. [multipledispatch](https://github.com/mrocklin/multipledispatch) seems unmaintained, while [plum](https://github.com/wesselb/plum) is newer and active ly maintained but less used.

Another option that could simplify things is using [Structural Pattern Matching](https://www.python.org/dev/peps/pep-0622/). This does require Python 3.10 however, and dropping support for all previous versions might be a bit too early.

rht:
Actually, upon rereading Agents.jl's [implementation](https://github.com/JuliaDynamics/Agents.jl/blob/903b3e987c8fa03d492c54ed6161d677159201e3/src/submodules/schedulers.jl#L100-L116), they use a flat list of agents and reconstruct the set of agents grouped by type on-the-fly.

And so the main design decision is whether we should use a flat dict like before, or a dict of dict.
@EwoutH is it correct that all the scheduler implementations need to have `agents_by_type` in this PR?

EwoutH:
I moved the metric aggegration code to the `datacollector`, in a new function `get_agent_metric()`. It returns a single value from the agent variables, using either a function in the statistics module or one of the built-in functions `"min"`, `"max"`, `"sum"` or `"len"`.

_Example:_

```Python
ExampleModel.datacollector.get_agent_metric("Weight", Cat, "min")
```
In the `ChartModule` an optional `"Metric"` can now be called, which defaults to `"mean"`.

_Example:_
```Python
chart_size = ChartModule(
    [{"Type": "Agent", "Agent_type": "Cat", "Label": "Size", "Metric": "max", "Color": "Blue"}],
    data_collector_name="datacollector"
)
```
Again, both architecture and implementation notes are much appreciated! 

(also CC @tpike3)

EwoutH:
> And so the main design decision is whether we should use a flat dict like before (and generates the `agents_by_type` on the fly like Agents.jl does), or a dict of dict.

Yes, this is a big design decision indeed. Love to hear your and everyone elses opinion about it!

One idea I had, it to detect in the `datacollector` if the first layer of keys of the `agent_reporters` is a `Class`, and if so, use the `agents_by_type` construct in a nested dict. If not, just use the regular implementation.

Another idea is to let the `agents_by_type` just default to the only agent class in the model, if there is only one. I think this may be the cleanest and simplest solution.

I think both options would be able to maintain backwards compatibility.

> @EwoutH is it correct that all the scheduler implementations need to have `agents_by_type` in this PR?

In this current draft, yes. One of the above solutions should be able to mitigate this as well.

tpike3:
@EwoutH This is awesome! Glad to see so much progress, particularly as I have a very bad habit of ignoring the visualization piece. 

To make sure I am understanding correctly, right now the current approach has a flat dict `self._agents` and a dict of dict `self.agents_by_type`. The question is do we want to calculate the `agents_by_type` on the fly? Is that correct or am I missing something. 

My initial response would be whatever is faster. (I will try and run some comparisons but due to my current job probably can't get to it until next weekend...sorry I am the slow one in this group)

My more thoughtful response is on the fly might allow for hypertypes. For the seminal wolf-sheep model this isn't a consideration but thinking of a social model where agents may change their type based on their situation (e.g. civil violence, politics, alliance) this could add some more capability where you can run the standard static type (e.g. wolf-sheep) but you can also allow for type adaption.     


> One idea I had, it to detect in the `datacollector` if the first layer of keys of the `agent_reporters` is a `Class`, and if so, use the `agents_by_type` construct in a nested dict. If not, just use the regular implementation.

This thought would mess up the class approach though because the type could also be an attribute and not just a class.

So I am not sure I helped or muddled the discussion, but I think grouping agents on the fly may offer some additional capabilities for more diverse model approaches.

rht:
Things to consider:
- Generating the by-type grouping on the fly is definitely slower when the agent types are stable, but this should rarely be the limiting factor of a simulation. If someone needs the pre-grouped dict, they can implement this scheduler in their code. And if there is a lot of demand on this pre-grouped scheduler, then it can be added in the library.
- If we are to detect whether the model uses random activation by type or not, in order to do the code selection, it'd be more semantic if the `DataCollector` checks the type of the scheduler object itself, i.e. check for `isinstance(self.model.scheduler, RandomActivationByType)`. But this means that there is an order dependence: the scheduler object must be initiialized before the data collection object.

EwoutH:
Thank you both for you responses according the architecture!

In the next week I will focus on finishing up my courses for this quarter (I'm still a student), and the week after that I'm a week on vacation and offline. So somewhere after February 7th, I will continue this effort.

If anyone in the mean time wants to continue or try some things, feel free to checkout my branch (https://github.com/EwoutH/mesa/tree/multiple-agent-types) or install it (`pip install -e git+https://github.com/EwoutH/mesa@multiple-agent-types#egg=mesa`).

A particularly piece of this PR wasn't about multi-agent support, but just about reading and plotting agent variables. Thus, I've created a separate PR, that implements this:

- #1145

If we get that in first, we can built multiple-agent types on top of it. Maybe #1145 can even make the 0.9.0 release!

Also Tom, don't even sweat it for a second, your pretty involved as a maintainer. Thank you!

tpike3:
@EwoutH  sorry it didn't back the 0.9.0 release but these are pretty big challenges

EwoutH:
No worries, this isn't even close to ready yet. Purely a proof of concept.

I hope to spend some more time on it in the coming weeks.

tpike3:
@EwoutH Based on #1162, we can close this correct?

EwoutH:
#1162 is only the scheduler, there is some DataCollector and Visualisation stuff in here to.

I did like to leave it open for a while, until I can properly implement it or someone else picks it up.

tpike3:
Superseded by #1287 and #1162

PR #1149, opened 2022-01-28
width and height were changed up

LizaRoelofsen:
This meant you could not change up the canvas space to make it rectangle, with these changes you can.


tpike3:
@LizaRoelofsen Thanks! I want to merge this but when I run the workflows it seems lint-black is failing and I think this is why the other checks are just running until they get cancelled.  

Did you install pre-commit? The [documentation](https://github.com/projectmesa/mesa/blob/main/CONTRIBUTING.rst) will walk you through setting up your environment so these are automatically fixed.

PR #1162, opened 2022-02-20
feat: Implement random activation by type

rht:
This is the least controversial way of putting in random activation by
type into the library, by simply using RandomActivationByBreed found in
sugarscape_cg and wolf_sheep (I have checked that the schedule.py in
both examples are identical).

Some differences:
- replace the term "breed" by a more general "type"
- add Mypy typings

If people want to get logging specific to one agent type, they can
either
- loop through all agents, and filter by their type
- loop through model.scheduler.agents_by_type[type_class]

We don't support agents that change their type, because in Python, you
can't have an object that changes its intrinsic type over its lifetime.

There could perhaps be a `RandomActivationByAttribute`, that groups
agents by their attributes, which may change throughout the course of a
simulation.


rht:
@EwoutH check this PR out. It's much less controversial than #1142, and so require fewer points to be discussed for a consensus.

EwoutH:
From an architecture perspective I agree this is the right course of action and a nice first step.

Haven't looked closely at the exact implementation, I can do that tomorrow if you want (however I think I'm not going to find much).

mesa/time.py
```diff
@@ -198,3 +198,78 @@ def step(self) -> None:
...
+        """
+        Remove all instances of a given agent from the schedule.
+        """
+
+        del self._agents[agent.unique_id]
+
+        agent_class: Type[Agent] = type(agent)
+        del self.agents_by_type[agent_class][agent.unique_id]
+
+    def step(self, by_type: bool = True) -> None:
```
EwoutH, line 243: Can we add something like an optional `order` input that allows to input the order in which the agent types are executed? Maybe that input could be either the string `"reversed"` to loop through `reversed(self.agents_by_type)` or a custom list with agent types.

I suggested a code snipped below.

mesa/time.py
```diff
@@ -198,3 +198,78 @@ def step(self) -> None:
...
+            by_type: If True, run all agents of a single type before running
+                      the next one.
+        """
+        if by_type:
+            for agent_class in self.agents_by_type:
+                self.step_type(agent_class)
+            self.steps += 1
+            self.time += 1
+        else:
+            super().step()
```
EwoutH, line 257:
```suggestion
        if by_type:
	    if order == "regular"
            	for agent_class in self.agents_by_type:
                    self.step_type(agent_class)
	    if order == "reverse"
		for agent_class in reversed(self.agents_by_type):
                    self.step_type(agent_class)
	    else:
		for agent_class in order:
                    self.step_type(agent_class)
            self.steps += 1
            self.time += 1
        else:
            super().step()
```

mesa/time.py
```diff
@@ -198,3 +198,78 @@ def step(self) -> None:
...
+
+    def step_type(self, type_class: Type[Agent]) -> None:
+        """
+        Shuffle order and run all agents of a given type.
```
EwoutH, line 261: This is equivalent to "ask breed" in [NetLogo](http://ccl.northwestern.edu/netlogo/docs/dict/breed.html) right? In that case, maybe explicitly document that.

EwoutH (commented):
Thanks for the PR, I added two comments, one about custom step orders and a documentation nitpick.

mesa/time.py
rht: It's already stated in the main docstring: https://github.com/projectmesa/mesa/pull/1162/files#diff-2ae534e35a540eca09bc3f43b6e3ffd91695f0b178f4824869c8975a0c0423a5R208.

mesa/time.py
rht: That's a good idea. Also good to compare with the API of NetLogo, MASON, and Agents.jl.

rht: Agents.jl has 2 params: `shuffle_types`, `shuffle_agents`: https://github.com/JuliaDynamics/Agents.jl/blob/5d1182dc7dd80001492c27333898c50462bdb20e/src/submodules/schedulers.jl#L92-L99.

mesa/time.py
EwoutH: I think this is a better place to (also) mention it since it's a function not available in other schedulers.

mesa/time.py
rht: NetLogo' leaves the order up to the user's implementation with `ask [breed]`: https://github.com/NetLogo/models/blob/2bfb84e92b367f312e397338e27a4c85889f13a0/Sample%20Models/Biology/Wolf%20Sheep%20Predation.nlogo#L57-L75.

EwoutH: Maybe we can allow three strings for `order`: `regular` (default), `reverse` and `shuffle`, and a list of agent types.

rht: Can't find a wolf-sheep model for MASON. It is also N/A in https://juliadynamics.github.io/Agents.jl/stable/comparison/.

rht:
2 points that need consensus:

First, the `by_type` boolean is not necessary, and complicates the API without much benefit. If the user wants the regular `RandomActivation`, they can simply use `RandomActivation`.

Second, the step order of each type. I can't think of a use case when a modeler wants to do the reverse of the original insertion order of the agent types during the model initialization. I think a more general solution is to have an argument called  `agent_type_sequence`, with a type of `Optional[Sequence[Type[Agent]]]`. The modeler can specify any order they want.

EwoutH:
> First, the `by_type` boolean is not necessary, and complicates the API without much benefit. If the user wants the regular `RandomActivation`, they can simply use `RandomActivation`.

Agreed, lets drop the `by_type` boolean.

> Second, the step order of each type. I can't think of a use case when a modeler wants to do the reverse of the original insertion order of the agent types during the model initialization. I think a more general solution is to have an argument called `agent_type_sequence`, with a type of `Optional[Sequence[Type[Agent]]]`. The modeler can specify any order they want.

A custom order like `Optional[Sequence[Type[Agent]]]` is indeed the most important. You can of course implement reversing or shuffling the order yourself, but it might be nice to just very quickly be able experiment and iterate with it.

mesa/time.py
rht: OK, I additionally mentioned here, so it is mentioned twice, just to be sure.

rht:
Shuffling may be in demand, but I'm not sure about the reverse order.

rht:
I forgot to say that I removed the unnecessary `by_type` arg, and added `shuffle_types` and `shuffle_agents`. At least, featurewise, it is the same as Agents.jl. I'm tempted to say that adding custom types order argument can be deferred to a separate PR. We can always iterate on this activation before the next release.

rht:
Bump. @tpike3 thoughts so far? I think this is ready to merge. Once you ack, I will squash and merge. After that, I will port my [implementation](https://github.com/AB-CE/examples/tree/master/examples/sugarscape) of Sugarscape with trading ({G1}, {M, T}) and make a PR for it.

tpike3:
Dang, I am way behind... @rht and @EwoutH this is awesome.  This is the implementation I am most familiar with so am ready to merge and I agree the terms (e.g type vs breed and shuffle) are the preferred. 

> We can always iterate on this activation before the next release.

To @rht's point thinking future customizations in new PRs before the next release ... I will concur and support two additional features in later PRs
1. I concur on the ability for a customized type order. To me the ability to customize is critical as *growing* simulation in different ways is often more insightful than the actual results 
2. If I am reading it right since I did not get a chance to play around with the code but don't want to hold up progress, we should have the option to run all agents in a random order regardless of type.. in [some models it will matter](https://dblp.org/rec/phd/basesearch/Comer14) and is supporting evidence to Lorenz's seminal work on [Deterministic Nonperiodic Flow.](https://www.astro.puc.cl/~rparra/tools/PAPERS/lorenz1962.pdf)

rht:
>  we should have the option to run all agents in a random order regardless of type

You can do this with:
```
if fully_random:
else:
    self.scheduler = RandomActivationByType(self)
```

And so we can keep the API minimal.

tpike3:
> > we should have the option to run all agents in a random order regardless of type
> 
> You can do this with:
> 
> ```python
> if fully_random:
>     self.schedule = RandomActivation(self)
> else:
>     self.scheduler = RandomActivationByType(self)
> ```
> 
> And so we can keep the API minimal.

@rht That is perfect! I didn't even think of that...so obvious I looked right over it.  Squash and merge at your discretion

rht:
Yay! Can't wait for the sugarscape variations being in the repo (at least the ones that doesn't require Multi-level Mesa).

PR #1170, opened 2022-02-22
CI: Split off codespell job, don't run build on doc changes

EwoutH:
Split the `python.yml` job into a `build.yml` with the build and lint jobs and `codespell.yml` with the codespell jobs.

The build and lint jobs now don't run when only changes to Markdown (`.md`) or reStructuredText (`.rst`) files have been made.

### DIFF
--- a/.github/workflows/build_lint.yml
+++ b/.github/workflows/build_lint.yml
@@ -5,7 +5,13 @@ on:
     branches:
     - main
     - release**
+    paths-ignore:
+      - '**.md'
+      - '**.rst'
   pull_request:
+    paths-ignore:
+      - '**.md'
+      - '**.rst'
 
 jobs:
   build:
@@ -72,12 +78,3 @@ jobs:
     - run: pip install black[jupyter]
     - name: Lint with black
       run: black --check --exclude=mesa/cookiecutter-mesa/* .
-
-  codespell:
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v2
-    - uses: codespell-project/actions-codespell@master
-      with:
-        ignore_words_file: .codespellignore
-        skip: .*bootstrap.*,*.js,.*bootstrap-theme.css.map
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,18 @@
+name: build
+
+on:
+  push:
+    branches:
+    - main
+    - release**
+  pull_request:
+
+jobs:
+  codespell:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: codespell-project/actions-codespell@master
+      with:
+        ignore_words_file: .codespellignore
+        skip: .*bootstrap.*,*.js,.*bootstrap-theme.css.map


.github/workflows/codespell.yml
```diff
@@ -0,0 +1,15 @@
+name: build
+
+on:
+  push:
```
rht, line 4: Need to restrict push to main branch. Otherwise, this will cause unnecessary CI runs for the forks.

rht: `main` and `release**`.

.github/workflows/build_lint.yml
```diff
@@ -5,7 +5,13 @@ on:
     branches:
     - main
     - release**
+    paths-ignore:
```
rht, line 8: Just wanting to understand this behavior:
- If the PR contains a mixture of .md and .py changes, this job will run
- If only .md, this job won't run
- If only .py, this job will run
Is this correct?

rht: OK, the doc spelled it out clearly:
> When all the path names match patterns in paths-ignore, the workflow will not run. If any path names do not match patterns in paths-ignore, even if some path names match the patterns, the workflow will run.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpaths

rht:
This means that if only .gitignore, CITATION.bib, LICENSE, *.ipynb, etc. are modified, the tests are still triggered. We may want to instead include *.py, *.js, *.yml, *.yaml, *.lock? Both `paths` and `paths-ignore` require about the same effort to make it comprehensive, so I don't see which way is easier.

.github/workflows/codespell.yml
EwoutH: This job is insanely quick so I think it’s fine and even beneficial to run it on every job. That way spelling errors can be catches before PRs are even openend.

rht:
Most people (including me) don't usually check the tick mark of the CI in the commits in their own fork. I had done so, but rarely. Currently, the CI is run twice for each PR. Once, because it's a PR, twice, because it is pushed to the fork branch.

.github/workflows/codespell.yml
rht: Most people (including me) don't usually check the tick mark of the CI in the commits in their own fork. I had done so, but rarely. Currently, the CI is run twice for each PR. Once, because it's a PR, twice, because it is pushed to the fork branch.

EwoutH: I have no idea how other people develop, but I like to see as many parts of the CI pass as feasible before I open an PR. And running the code spell job has a very low cost, so I argue for keeping it this way.

EwoutH:
> This means that if only .gitignore, CITATION.bib, LICENSE, *.ipynb, etc. are modified, the tests are still triggered. We may want to instead include *.py, *.js, *.yml, *.yaml, *.lock? Both `paths` and `paths-ignore` require about the same effort to make it comprehensive, so I don't see which way is easier.

I think the vast majority of the cases in which the CI is unnecessary triggered is when modifying Markdown or reStructuredText docs. Excluding those is enough in my opinion, we're not that busy that the very occasional .gitignore, CITATION.bib, LICENSE or *.ipynb change warrants all kind of special behavior. Let's keep it simple and maintainable.

> Most people (including me) don't usually check the tick mark of the CI in the commits in their own fork. I had done so, but rarely. Currently, the CI is run twice for each PR. Once, because it's a PR, twice, because it is pushed to the fork branch.

The penalty of running this few-second single-instance job is so low that I prefer to keep the CI as simple as possible to reduce the maintenance effort.

rht:
.gitignore, CITATION.bib, LICENSE changes are rare, but *.ipynb changes are common. Also, updating the tutorial .png files are also common.

> The penalty of running this few-second single-instance job is so low that I prefer to keep the CI as simple as possible to reduce the maintenance effort.

I fail to see how adding 2 lines of .yml (which pattern is already observed in the test .yml) makes it harder to maintain.

rht:
I'm delegating the final decision to @tpike3.

tpike3:
@rht and @EwoutH  First, this is very exciting to see some hard conversations about moving Mesa forward. I cannot thank you two enough, particularly as I struggle to keep up. 

Regarding my tie breaker vote as I have been burned by ```.ipynb``` more times then I can remember  my vote is to add the lines. Ideally, we will grow this project to the points that it does matter. :)

tpike3:
@EwoutH  Are you willing to restrict this to the main branch so we can close this PR?

.github/workflows/codespell.yml
EwoutH: Updated the job so it only runs on pushes to main and release branches

rht:
@EwoutH you need to rebase on main, and squash the 2 commits.

tpike3:
Thanks @Ewout!!! (Ha I was just writing you!)

EwoutH:
Quick response times here! :)

I rebased, reset and recommitted. Should be ready for merge!

rht:
LGTM. Further iteration on whether to fine-tune the ignore can be discussed in an issue.

Issue #1148, opened 2022-01-25
The tutorial.rst doesn't mention that the Pandas DataFrame output can be in CSV


rht:
See the thread in #1143.

Issue #1107, opened 2021-12-17
batchrunnerMP example?


swirya:
Adapting the bank_reserves example to use batchrunnerMP results in the following error:
MaybeEncodingError: Error sending result: '((25, 5, 0, 0), <__main__.BankReservesModel object at 0x7f5d26e1a700>)'. Reason: 'AttributeError("Can't pickle local object 'BankReservesModel.__init__.<locals>.<lambda>'")'

Here's the command: 
batch = BatchRunnerMP(
            BankReservesModel,
            nr_processes=None,
            variable_parameters=br_params,
            iterations=1,
            max_steps=1000,
            model_reporters={"Data Collector": lambda m: m.datacollector},
)
batch.run_all()

What am I missing? Thanks in advance.

rht:
This seems to be a known problem. There are 2 solutions:
1. Use https://github.com/uqfoundation/multiprocess instead of Python stdlib's `multiprocessing`'. Basically, do `pip install multiprocess`, and then replace https://github.com/projectmesa/mesa/blob/d2243971e6101e812422c0853eb7ae1a1671ffbc/mesa/batchrunner.py#L11 with `from multiprocess import Pool, cpu_count`
2. Use `mp.Process` instead, which doesn't require pickling

The first solution is a one-line fix, but requires installing an external library. The second one uses stdlib, but requires more changes in the code. Also, the code won't look clean (note: this messiness is invisible to the user; it's messy only for Mesa developers).

@swirya  if you want a quick fix, you can do the first solution.

But a permanent solution requires more discussion.

tpike3:
@swirya  I am going to close as this resolved based on @rht's comments. 

Also just for your awareness, we just did a release with ```batch_run``` and this uses multiprocessing by default. Overview is in the new [docs](https://mesa.readthedocs.io/en/main/)

rht:
I think this is still an open issue. `batch_run` uses the same multiprocessing mechanism as the old `BatchrunnerMP`, i.e. `imap_unordered`, and so there will still be a pickling problem. My second solution, which is to use `mp.Process`, should work, as I had encountered this pickling problem in the past, and solved it with `mp.Process`.

rht:
I tried adding a lambda function in the model reporter of examples/bank_reserves/batch_run.py. This breaks `BatchRunnerMP` with the pickling problem, but it somehow works in `batch_run`. Both use `imap_unordered`, but maybe because `batch_run` is a simple function, it works.

[...]

PR #1182, opened 2022-03-06
update intro tutorial to include pandas and csv

tpike3:
Fix for #1148


docs/tutorials/intro_tutorial.rst
```diff
@@ -713,10 +713,11 @@ At every step of the model, the datacollector will collect and store the
 model-level current Gini coefficient, as well as each agent’s wealth,
 associating each with the current step.
 
-We run the model just as we did above. Now is when an interactive
-session, especially via a Notebook, comes in handy: the DataCollector
-can export the data it’s collected as a pandas DataFrame, for easy
-interactive analysis.
+We run the model just as we did above. Now is when an interactive session, especially via a Notebook, comes in handy: the DataCollector can export the data its collected as a pandas\* DataFrame, for easy interactive analysis.
+
+\*If you are new to Python, please be aware that pandas is already installed as a dependency of Mesa and that `pandas`_ is a "fast, powerful, flexible and easy to use open source data analysis and manipulation tool". pandas is great resource to help analyze the data collected in your models
```
rht, line 718: *Pandas.

docs/tutorials/intro_tutorial.rst
```diff
@@ -852,6 +853,19 @@ Or to plot the wealth of a given agent (in this example, agent 14):
 
 .. image:: intro_tutorial_files/intro_tutorial_45_1.png
 
+You can also use pandas to export the data to a csv (comma separated value),
```
rht, line 856: *CSV

rht:
Should we remove the cell output in the .ipynb? The diff is unnecessarily huge whenever there is a change. Imagine changing this 20x, then we basically have 20 copies of the .ipynb file in the repo.

docs/tutorials/intro_tutorial.rst
tpike3: @rht I went back and forth on that in the end I went with the fact that they did not capitalize it on their [docs](https://pandas.pydata.org/) but I am good either way.... capitalize?

tpike3:
> Should we remove the cell output in the .ipynb? The diff is unnecessarily huge whenever there is a change. Imagine changing this 20x, then we basically have 20 copies of the .ipynb file in the repo.

Yes... I knew it was something simple so will do

docs/tutorials/intro_tutorial.rst
rht: Ugh, I see. It's so hard to keep track of the software that insist that their proper name is lower cased! Even Wikipedia does it in lowercase: https://en.wikipedia.org/wiki/Pandas_(software). I guess we have to respect the lower casing, just like in macOS, nginx, and so on.

tpike3:
- I made updates to CSV

- I added note about windows OS and batch_run(). 

- It also looks like the cells form intro_tutorial.ipynb were not cleared last time. So my diff is still very large (not as large). I don't think there is anything I can do about this, except make it standard practice for the future. (but happy to execute if there is a better idea)

docs/tutorials/intro_tutorial.rst
rht, line 718: *pandas is a great resource ...

docs/tutorials/intro_tutorial.rst
```diff
@@ -990,7 +1004,32 @@ iteration).
 
     245it [00:25,  9.75it/s]
 
+Note: If you are running a Windows OS you may have an issue with `batch_run`. One way to overcome this is to take your
```
rht, line 1007: You should say what the error message on Windows is, so that people know it is the same problem.

rht: Also, I don't think it needs to be outside Jupyter. The `freeze_support()` can be run inside Jupyter.

rht: Can't test because I don't have access to a Windows machine.

tpike3: I added a note before the cell. If  you change number of processes to 1 it works on WIndows OS. Even adding freeze_support etc you get the same error. Based on the traceback Jupyter is blocking multiprocessing's ability to get the task.

docs/tutorials/intro_tutorial.rst
rht: @tpike3 you forgot "a" here.

docs/tutorials/intro_tutorial.rst
```diff
@@ -970,6 +984,8 @@ iteration).
 .. code:: python
 
     from mesa.batchrunner import batch_run
+**Note for Windows OS users:** We recommend changing the the parameter `number_processes = None` to `number_processes = 1.`
```
rht, line 987: This reads like a general recommendation that Windows users should better use `number_processes = 1`, which is not the case. The problem is specific to running in Jupyter on Windows.

rht: Probably should also explain that the `number_processes` is for setting for multiple CPUs.

docs/tutorials/intro_tutorial.rst
```diff
@@ -990,7 +1006,35 @@ iteration).
 
     245it [00:25,  9.75it/s]
 
+**Note for Windows OS users:** You may have an issue with `batch_run` and `number_processes = None`. Your cell will
```
rht, line 1009: I think we need a stronger statement, "you will" instead of "you may"? Isn't it happening 100% of the time?

rht:
@tpike3 have you experimented with running `freeze_support()` inside `batch_run` when on Windows, so that people don't have to manually call it?

There is another possible solution: using pathos's multiprocess (https://github.com/projectmesa/mesa/issues/1107#issuecomment-996526753). This has the added bonus of being able to pickle more object types than the builtin pickling mechanism.

rht:
What I want to know is that whether `pathos`'s multiprocess already does the `freeze_support` under the hood.

docs/tutorials/intro_tutorial.rst
tpike3: dang..I got distracted with the batch_run will fix

tpike3:
> @tpike3 have you experimented with running `freeze_support()` inside `batch_run` when on Windows, so that people don't have to manually call it?
> 
> There is another possible solution: using pathos's multiprocess ([#1107 (comment)](https://github.com/projectmesa/mesa/issues/1107#issuecomment-996526753)). This has the added bonus of being able to pickle more object types than the builtin pickling mechanism.

@rht No I have not yet. The challenge is you will still need ```if __name__ == '__main__':``` which from my understanding has to initiate. But much like freeze_support and Pathos they have clearly found a way around it.   

BatchRunnnerMP was originally done through Pathos, but if I remember correctly around Python 3.7, an update in Python broke BatchRunnerMP with Pathos and so we transitioned to multiprocessing to prevent such issues in the future.

rht:
According to the answer in https://stackoverflow.com/questions/50921350/pathos-pool-statement-gets-hanged, it looks like `freeze_support` is needed even with `pathos`.

docs/tutorials/intro_tutorial.rst
```diff
@@ -970,6 +984,9 @@ iteration).
 .. code:: python
 
     from mesa.batchrunner import batch_run
+**Note for Windows OS users:** If you are running this tutorial in Jupyter, we recommend changing the the parameter
+that controls the number of processors you are using from `number_processes = None` (this uses all available processors)
+to number_processes = 1. More information on leveraging batch_run's multiprocessing capability is below.
```
rht, line 989: You forgot a pair of backticks.

rht:
Other than the backticks, LGTM. Feel free to merge anytime.

docs/tutorials/intro_tutorial.rst
tpike3: Ohhh nice catch....fixed

rht:
@tpike3  Black failed.

tpike3:
@rht Thanks for the CSV fix. I also noticed those multiprocessing code did not appear because sphinx needs an extra space. That is now fixed. #1201

Issue #980, opened 2021-01-28
space: Confusing API with methods with similar names


rht:
The class `Grid` has:
- `neighbor_iter`:  "Iterate over position neighbors"
- `iter_neighborhood`: "Return an iterator over cell coordinates that are in the neighborhood of a certain point"
- `iter_neighbors`: "Return an iterator over neighbors to a certain point"

`neighbor_iter` and `iter_neighbors` seem to be a duplicate of each other.
`iter_neighborhood` should probably be called `iter_neighbor_coordinates` for the type of its return value.

Corvince:
> `neighbor_iter` and `iter_neighbors` seem to be a duplicate of each other.

Yes, except `neighbor_iter` has a restricted API, but HexGrid only implements `neighbor_iter` and not `iter_neighbors` 🤦 

`iter_neighbors` should be the only function IMO.

> `iter_neighborhood` should probably be called `iter_neighbor_coordinates` due to the type of its return value.

Disagree. For me it was always clear that neighbors => Agents, neighborhood => Coordinates

rht:
> Disagree. For me it was always clear that neighbors => Agents, neighborhood => Coordinates

Is it just me or that the dictionary definition of "neighborhood" refers to people, not geographical coordinates?

From https://en.wikipedia.org/wiki/Neighbourhood:
```
A neighbourhood (British English, Australian English and Canadian English) or neighborhood (American English; see spelling differences—u is omitted in American English) is a geographically localised community within a larger city, town, suburb or rural area. Neighbourhoods are often social communities with considerable face-to-face interaction among members.
```

Corvince:
> Is it just me or that the dictionary definition of "neighborhood" refers to people, not geographical coordinates?

Well maybe look it up in a dictionary!?

From https://www.dictionary.com/browse/neighborhood?s=t

`the area or region around or near some place or thing; vicinity`

If you would have read the wikipedia just a little bit further, it would also say this:

`Neighbourhood is generally defined spatially as a specific geographic area and functionally as a set of social networks.`

So both definition exists and I definitely see your point, but actually both definitions are irrelevant in this context. 

The origin of neighborhoods in the context of ABMs relies on the mathematical definition of neighborhoods, which is just an open set of points that contains a specific point. No people involved

rht:
If `iter_neighborhood` is find, then the problematic wording is the `neighbor_iter` and `iter_neighbors`. They should be called `iter_neighborhood_content` for clarity.

[...]

PR #1184, opened 2022-03-08
Deprecate neighbor_iter in favor of iter_neighbors

rht:
See https://github.com/projectmesa/mesa/issues/980.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -197,8 +197,13 @@ class Grid:
             moore: Boolean for whether to use Moore neighborhood (including
                    diagonals) or Von Neumann (only up/down/left/right).
         """
-        neighborhood = self.get_neighborhood(pos, moore=moore)
-        return self.iter_cell_list_contents(neighborhood)
+        from warnings import warn
+
+        warn(
+            "`neighbor_iter` is deprecated in favor of `iter_neighbors` "
+            "and will be removed in the subsequent version."
+        )
+        return self.iter_neighbors(pos, moore)
 
     def iter_neighborhood(
         self,
@@ -606,7 +611,7 @@ class HexGrid(Grid):
     Methods:
         get_neighbors: Returns the objects surrounding a given cell.
         get_neighborhood: Returns the cells surrounding a given cell.
-        neighbor_iter: Iterates over position neighbors.
+        iter_neighbors: Iterates over position neighbors.
         iter_neighborhood: Returns an iterator over cell coordinates that are
             in the neighborhood of a certain point.
     """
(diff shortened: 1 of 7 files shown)


mesa/space.py
```diff
@@ -197,8 +197,12 @@ def neighbor_iter(
             moore: Boolean for whether to use Moore neighborhood (including
                    diagonals) or Von Neumann (only up/down/left/right).
         """
-        neighborhood = self.get_neighborhood(pos, moore=moore)
-        return self.iter_cell_list_contents(neighborhood)
+        from warnings import warn
+        warn(
+            "`neighbor_iter` is deprecated in favor of `iter_neighbors` "
+            "and will be removed in the subsequent version."
+        )
```
rht, line 204: Notice I intentionally didn't use `DeprecationWarning`, because when I tested this warn, it didn't show up when I used `DeprecationWarning`. It will show up as `UserWarning`.

tpike3:
I concur this makes the methods consistent. Action (i.e. get or iter) and then target (i.e. neighbors or neighborhood). 

```
        get_neighbors: Returns the objects surrounding a given cell.
        get_neighborhood: Returns the cells surrounding a given cell.
        iter_neighbors: Iterates over position neighbors.
        iter_neighborhood: Returns an iterator over cell coordinates that are
            in the neighborhood of a certain point.
```

Based on the diversity of opinions in  #980 and as we are only a week away from the dev session, to ensure inclusivity we can bring up for concurrence. 

Could you please squash your commits?

rht:
Squashed.

tpike3:
@jackiekazil Can you due to the debate on this subject can you give this pull request an additional review. I concur with @rht approach.

rht:
I think this PR is not as controversial as #980. The `iter_*` vs `get_*` is not included in this PR. This PR only removes the `neighbor_iter`, which I haven't seen anyone arguing the case for it.

rht:
I'm planning to renew #835, but this work depends on getting this PR merged, because I can't cause a merge conflict by unnecessarily annotating `neighbor_iter`.

jackiekazil:
LGTM

jackiekazil:
Thank you for your work on this @rht

EwoutH:
Could it be that `.iter_neighbors()` now returns a `Coordinate` (`Tuple(Int, Int)`) instead of an `Agent`? If so, is that intended behaviour with the switch from `.neighbor_iter()` to `.iter_neighbors()` in Mesa 2.0?

rht:
It still return agents: https://github.com/projectmesa/mesa/blob/7f5a32e63db93b9be2b00a705d0c28a66bc5a9dd/mesa/space.py#L722.

EwoutH:
Ah I accidentally replaced `.neighbor_iter()` with `.iter_neighborhood()` instead of `.iter_neighbors()`.

I also see where I went wrong, it's stated that way in the 2.0.0 release notes.

https://github.com/projectmesa/mesa/blob/7f5a32e63db93b9be2b00a705d0c28a66bc5a9dd/HISTORY.rst?plain=1#L50C13-L50C13

I opened a PR to fix this: #1807.

PR #1193, opened 2022-03-19
Improve Dockerfile

rht:
This PR:
- uses the barebone Python 3.10 from official Python Docker image.
- parameterizes the model directory, so that user can use the same Docker image to try out various models easily with e.g. `docker run --name mymesa_instance -p 8521:8521 -e MODEL_DIR=examples/sugarscape_cg -it mymesa_image`.
- supersedes #702.

The Python 3.10 base slim image is 125 MB. The Mesa Docker in this PR is 359 MB (including Mesa Git repo, which is >50 MB).


tpike3 (approved):
LGTM

Issue #1166, opened 2022-02-20
CI: test flake: batch runner sometimes takes 6 hours then gets killed by GitHub Actions


rht:
See e.g. https://github.com/projectmesa/mesa/runs/5259005495?check_suite_focus=true.

We might want to add timeout to batch runner tests, with https://pypi.org/project/pytest-timeout/.

rht:
More findings at https://github.com/projectmesa/mesa/pull/1194#issuecomment-1073221814.

rht:
I tried 2 methods:
- using spawn #1199, failed because the simulation became much slower
- using explicit import of multiprocessing #1198, failed after the 3rd attempt

rht:
I have narrowed down the cause: test_batchrunnerMP.py (see #1255).

PR #1194, opened 2022-03-20
ci: Add 6 min timeout for the test jobs

rht:


tpike3:
@rht this always fails with Ubuntu 3.8 and test batch_run/batch_runner. I imagine this is another multiprocessing issue. 

Doing some research it seems like Ubuntu 3.8 is also resulting in a deadlock. I found this [potential solution](https://pythonspeed.com/articles/python-multiprocessing/) but haven't tried it yet. Optimistically, I am hoping this will fix Ubuntu 3.8 and the windows issue if we put it in `batch_run`. But, it seems like users will still need `if __name__ == '__main__'`

rht:
See also https://github.com/scipy/scipy/issues/11835.

It seems to not be specific to 3.8, e.g. see the CI result of #1181.

But anyway, as you can see, the test hang is easily detectable with this PR, so I think we should further discuss in #1166 instead of here.

tpike3:
Makes sense to me.

PR #1196, opened 2022-03-20
update intro tutorial for pandas and CSV and batch_run and windows

tpike3:
Updated tutorial. 

The .ipynb is still large mainly due to the cells being cleared but following this practice in the future should prevent this from happening again.

### DIFF
--- a/docs/tutorials/intro_tutorial.rst
+++ b/docs/tutorials/intro_tutorial.rst
@@ -713,10 +713,11 @@ At every step of the model, the datacollector will collect and store the
 model-level current Gini coefficient, as well as each agent’s wealth,
 associating each with the current step.
 
-We run the model just as we did above. Now is when an interactive
-session, especially via a Notebook, comes in handy: the DataCollector
-can export the data it’s collected as a pandas DataFrame, for easy
-interactive analysis.
+We run the model just as we did above. Now is when an interactive session, especially via a Notebook, comes in handy: the DataCollector can export the data its collected as a pandas\* DataFrame, for easy interactive analysis.
+
+\*If you are new to Python, please be aware that pandas is already installed as a dependency of Mesa and that `pandas`_ is a "fast, powerful, flexible and easy to use open source data analysis and manipulation tool". pandas is a great resource to help analyze the data collected in your models
+
+.. _pandas: https://pandas.pydata.org/docs/
 
 .. code:: python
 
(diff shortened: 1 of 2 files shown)


rht:
You need to rebase your branch on top of the upstream main branch. Somehow the .pre-commit-config.yaml change is included in this PR.

tpike3:
Fixed, I think I squashed to many commits.

EwoutH:
Thanks for figuring this out, documenting it and adding it to the tutorial! This was a big issue on Windows.

There is no way we can fix this in the class/method source code itself, by the way?

docs/tutorials/intro_tutorial.rst
```diff
@@ -852,6 +853,19 @@ Or to plot the wealth of a given agent (in this example, agent 14):
...
 
+You can also use pandas to export the data to a CSV (comma separated value),
+which can be opened by any common spreadsheet application or opened by pandas.
+
+If you do not specify a file path, the file will be saved in the local directory.
+After you run the code below you will see two files appear (*model_data.csv* and *agent_data.csv*)
+
+.. code:: python
+
+    #save the model data (stored in the pandas gini object) to csv
```
rht, line 864: Need a space after the "#", as per PEP 8.

tpike3: fixed

tpike3:
> Thanks for figuring this out, documenting it and adding it to the tutorial! This was a big issue on Windows.
> 
> There is no way we can fix this in the class/method source code itself, by the way?

@EwoutH This is a non-trivial problem see #1194

rht:
@EwoutH 's question seems to be specifically on whether the `freeze_support()` can be auto-called when the `batch_run()` is called. Reading from internet, people always suggest putting it right after `if __name__ == "__main__":`. It could work inside `batch_run()` as well, but I can't test it since I don't have access to a Windows machine.

docs/tutorials/intro_tutorial.rst
```diff
@@ -852,6 +853,19 @@ Or to plot the wealth of a given agent (in this example, agent 14):
...
 
+You can also use pandas to export the data to a CSV (comma separated value),
+which can be opened by any common spreadsheet application or opened by pandas.
+
+If you do not specify a file path, the file will be saved in the local directory.
+After you run the code below you will see two files appear (*model_data.csv* and *agent_data.csv*)
+
+.. code:: python
+
+    # save the model data (stored in the pandas gini object) to csv
```
rht, line 864: Was about to say `CSV`, but we have gone through too many back and forth. Later I will fix it across the repo once for all.

PR #1198, opened 2022-03-21
test: Explicitly import multiprocessing to prevent batch_run hang

rht:
See https://github.com/scipy/scipy/issues/11835


rht:
I tried rerunning the CI several times. On the 3rd attempt, the hang happened. Closing now since this is not the solution to the problem.

PR #1199, opened 2022-03-21
test: Use spawn for batch_run multiprocessing to prevent hang

rht:
See https://pythonspeed.com/articles/python-multiprocessing/


rht:
Using `spawn` makes the test run much slower, and so it is not a viable solution either.

PR #1204, opened 2022-03-22
Add snippet about using numpy's random

### LINKED ISSUE #958: Using numpy.random instead of the build-in random module

rht:
Fixes #958.

Supersedes #998.

### DIFF
--- a/docs/useful-snippets/snippets.rst
+++ b/docs/useful-snippets/snippets.rst
@@ -9,5 +9,20 @@ If you have `Multiple` type agents and one of them has time attribute you can st
 
 .. code:: python
 
-if self.model.schedule.time in self.discrete_time:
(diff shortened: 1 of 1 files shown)


rht:
Regarding with @jackiekazil 's comment (https://github.com/projectmesa/mesa/pull/998#issuecomment-787400585), I don't think I have to make the PR to the docs branch. You can check that it renders correctly in https://github.com/rht/mesa/blob/majdal_random/docs/useful-snippets/snippets.rst.

tpike3 (approved):

PR #1212, opened 2022-03-26
mypy: Improve annotations

rht:
This is a subset of #835 to make it easier to review and merge.

### DIFF
--- a/.github/workflows/build_lint.yml
+++ b/.github/workflows/build_lint.yml
@@ -32,8 +32,6 @@ jobs:
           - os: ubuntu
             python-version: "3.7"
           - os: ubuntu
-            python-version: "3.6"
-          - os: ubuntu
             python-version: pypy3
 
     steps:
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -5,14 +5,20 @@ Core Objects: Agent
 
 """
 # mypy
-from mesa.model import Model
+from typing import Optional, TYPE_CHECKING
 from random import Random
+
+if TYPE_CHECKING:
+    # We ensure that these are not imported during runtime to prevent cyclic
+    # dependency.
+    from mesa.model import Model
+    from mesa.space import Position
 
 
 class Agent:
     """Base class for a model agent."""
 
-    def __init__(self, unique_id: int, model: Model) -> None:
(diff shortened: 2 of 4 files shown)


rht:
I'm deleting Python 3.6 from CI because its numpy version is too old, which doesn't have the Mypy stubs.

rht:
For [perf: ContinuousSpace: Use math instead of np for get_distance](https://github.com/projectmesa/mesa/pull/1212/commits/847693fcc3fea60601d41234329614c1c37a8f37), it's a well known issue that `numpy` functions are much slower than `math` functions for scalar arguments.

tpike3 (changes requested):
- Well it took me awhile, but I am learning a lot researching your improvements
- Can you squash your commits, although in this pull request it seems three things are really happening (remove, 3.6, improve annotations, improve dist calc). 
- Other than that LGTM

rht:
If I squash the commits, it will be hard to tell what that 1 commit actually does. Too many things happening in 1 commit.

rht:
I'm fine with multiple commits in 1 PR as long as each commit does different things logically, and the commits are not messy.
See also https://news.ycombinator.com/item?id=11407485.

rht:
The reason why I did the `np` -> `math` in this PR is that it is a side effect of not relying on `numpy` for typing. But later on I figured out that my `mypy` setup uses an outdated `numpy` where it didn't have stubs yet.

tpike3:
> If I squash the commits, it will be hard to tell what that 1 commit actually does. Too many things happening in 1 commit.

I agree with this and am okay with multiple commits. However, to make sure we are synced as I look at this as cementing our processes. When we would say to do multiple pull requests vs multiple commits? Per the ycombinator discussion there is no easy answer but how would you describe the "rule of thumb" we should use?

rht:
When this PR is merged, each of the commits will contain a link to this PR on GitHub. I'd say it is fine to have multiple commits in 1 PR as long as they are within the same theme.

tpike3:
> When this PR is merged, each of the commits will contain a link to this PR on GitHub. I'd say it is fine to have multiple commits in 1 PR as long as they are within the same theme.

Works for me.

PR #1035, opened 2021-05-11
ModularVisualization: Perform a check on whether the user has added datacollector.collect in model __init__

rht:
Fix #1030.

PR #1213, opened 2022-03-26
ModularVisualization: Perform a check on whether the user has added datacollector.collect in model __init__

rht:
Fix https://github.com/projectmesa/mesa/issues/1030.

The error message looks like this:
```
mesa.visualization.ModularVisualization.NoCollectInInitError:
You initialized a DataCollector object in your model but haven't added a `self.datacollector.collect(self)` in your model __init__.
```

I reopened #1035 as this PR. Waiting for inputs from other maintainers on whether this PR is fine from readability standpoint.

### DIFF
--- a/mesa/visualization/ModularVisualization.py
+++ b/mesa/visualization/ModularVisualization.py
@@ -117,6 +117,10 @@ if platform.system() == "Windows" and platform.python_version_tuple() >= ("3", "
     asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
 
 
+class NoCollectInInitError(Exception):
+    pass
+
+
 class VisualizationElement:
     """
     Defines an element of the visualization.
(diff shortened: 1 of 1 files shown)


tpike3 (changes requested):
This should be a great help to many, thanks!

Can you add a test, please?

rht:
You are fine with the addition in this PR that makes the code more complicated?

tpike3:
> You are fine with the addition in this PR that makes the code more complicated?

I think this goes to the strategy discussion we need to have. Right now we have a demand signal from Universities to get new students into ABMs. My understanding is we have seen users have these issues a few times.  So it seems prudent to add the error message. Do you disagree? I don't see the associated cost of adding this error message as too high, am I missing something?

rht:
One problem is that with this, `VisualizationElement` assumes/demands too much of what a model should do, making the components less orthogonal.

rht:
I have another solution: instead of doing this error checking after the fact, what about we define a method in model:
```python
# sugarscape cg
self.initialize_data_collector({"SsAgent": lambda m: m.schedule.get_type_count(SsAgent)})
```

Under the hood, this method is defined as follows:
```python
def initialize_data_collector(self, *args):
    self.datacollector = DataCollector(*args)
```

tpike3:
> I have another solution: instead of doing this error checking after the fact, what about we define a method in model:
> 
> ```python
> # sugarscape cg
> self.initialize_data_collector({"SsAgent": lambda m: m.schedule.get_type_count(SsAgent)})
> ```
> 
> Under the hood, this method is defined as follows:
> 
> ```python
> def initialize_data_collector(self, *args):
>     self.datacollector = DataCollector(*args)
>     self.datacollector.collect(self)
> ```

Just make the data_collector initialize be default. I certainly like that idea in principle its safe to assume one wants to collect data with a model and it should make it more accessible. Let me play around with it as well to try an internalize it.

rht:
Ah, right, here is another alternative:
```python
# in model __init__
self.datacollector = DataCollector(self, ...)
# in DataCollector __init__
def __init__(self, model, ...):
    ....
    self.collect(model)
```

rht:
One argument against `self.initialize_data_collector`: with this approach, people can't define and use their own `DataCollector` class easily. I think `self.datacollector = DataCollector(self, ...)` has the best trade-off.

rht:
From the April 2022 meeting: we can still use a custom data collector class with the `initialize_data_collector(self,...)` solution. We just need to define an optional argument called `datacollector_class` that is used as the class instead of the built-in `DataCollector`. And so, it's no longer a problem.

rht:
Let's discuss the alternative implementations (they are all better than this PR's approach, IMO) in an issue. Closing now.

PR #1215, opened 2022-03-28
setup.py: Update setup classifiers and add python_requires for Python>=3.7

EwoutH:
Update `setup.py` with the Programming Language [classifiers](https://pypi.org/classifiers/) and [python_requires](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#python-requires) argument to reflect Mesa supporting Python 3.7 and higher.


rht:
LGTM

PR #1217, opened 2022-03-29
CI: Update checkout, setup-python and cache actions to v3

EwoutH:
Updates the [actions/checkout](https://github.com/actions/checkout), [actions/setup-python](https://github.com/actions/setup-python) and [actions/cache](https://github.com/actions/cache) all to v3.

Note that the pypy minor version now needs to be specified for actions/setup-python@v3.

### DIFF
--- a/.github/workflows/build_lint.yml
+++ b/.github/workflows/build_lint.yml
@@ -32,15 +32,15 @@ jobs:
           - os: ubuntu
             python-version: "3.7"
           - os: ubuntu
-            python-version: pypy3
+            python-version: 'pypy-3.8'
 
     steps:
-    - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


rht:
LGTM

.github/workflows/build_lint.yml
```diff
@@ -32,15 +32,15 @@ jobs:
           - os: ubuntu
             python-version: "3.7"
           - os: ubuntu
-            python-version: pypy3
+            python-version: 'pypy-3.8'
```
rht, line 35: Was about to mention this minor nit: double quote so that it is consistent with the other quotes usage, but nvm.

EwoutH: Good catch, will keep it in mind next time!

PR #1219, opened 2022-03-29
mypy: Improve space.py annotation, part 2

rht:
The remaining missing parts from #835 not yet implemented:
- type casting: https://github.com/projectmesa/mesa/pull/835/commits/9f6fb57121db63314fe1127934756c86f650b852#diff-0b13b631d2293f06d685774185de133aaacba213d467419dbfe3061ff130151bR346. This can be implemented at any time just by following Mypy error messages, and so there is no need to consult this commit
- generic content type: https://github.com/projectmesa/mesa/pull/835/commits/56a63b83a169b3214e991fbbc3994427effffac4. I don't think it's worth the complexity of using generic content type, and so I didn't implement it

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -569,6 +569,8 @@ class MultiGrid(Grid):
         get_neighbors: Returns the objects surrounding a given cell.
     """
 
+    grid: List[List[MultiGridContent]]
+
     @staticmethod
     def default_val() -> MultiGridContent:
         """Default value for new cell elements."""
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -430,7 +430,7 @@ class Grid:
     def _remove_agent(self, pos: Coordinate, agent: Agent) -> None:
         """Remove the agent from the given location."""
         x, y = pos
-        self.grid[x][y] = None
+        self.grid[x][y] = self.default_val()
         self.empties.add(pos)
 
     def is_cell_empty(self, pos: Coordinate) -> bool:
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -197,9 +197,7 @@ class Grid:
             for col in range(self.height):
                 yield self.grid[row][col], row, col  # agent, x, y
 
-    def neighbor_iter(
-        self, pos: Coordinate, moore: bool = True
-    ) -> Iterator[GridContent]:
+    def neighbor_iter(self, pos: Coordinate, moore: bool = True) -> Iterator[Agent]:
         """Iterate over position neighbors.
 
         Args:
@@ -304,7 +302,7 @@ class Grid:
         moore: bool,
         include_center: bool = False,
         radius: int = 1,
-    ) -> Iterator[GridContent]:
+    ) -> Iterator[Agent]:
         """Return an iterator over neighbors to a certain point.
 
         Args:
@@ -332,7 +330,7 @@ class Grid:
         moore: bool,
         include_center: bool = False,
         radius: int = 1,
-    ) -> List[GridContent]:
+    ) -> List[Agent]:
         """Return a list of neighbors to a certain point.
 
         Args:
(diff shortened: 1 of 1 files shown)


rht:
Another main reason for not using generic content type: the type of the content is not interchangeable, unlike `List[...]`, which can be `List[str]`, `List[int]`, and so on.

mesa/space.py
```diff
@@ -197,9 +197,7 @@ def coord_iter(self) -> Iterator[Tuple[GridContent, int, int]]:
             for col in range(self.height):
                 yield self.grid[row][col], row, col  # agent, x, y
 
-    def neighbor_iter(
-        self, pos: Coordinate, moore: bool = True
-    ) -> Iterator[GridContent]:
+    def neighbor_iter(self, pos: Coordinate, moore: bool = True) -> Iterator[Agent]:
```
tpike3, line 200: Appreciating right now GridContent is just an agent object, but thinking about @Holzhauer 's (see Matrix) need for a wall. Do all items in the list need to be agent objects? This is non-trivial and off the top of my head I can think of reasons for or against so wanted to bring it up.

rht: A wall is an agent that does nothing in its `step()`. If a model doesn't use a `MultiGrid`, it is effectively preventing an agent to get into a cell. And so, I don't think there needs to be another abstract class for it.

Holzhauer: Thanks, @tpike3 for bringing it up. There are reasons to distinguish between moving agents and solid walls (e.g. when walls should not be redrawn), but from a modeller's perspective a wall is no agent (no [re-]action, no interaction, no objective). Therefore, it makes sense to introduce a distinction (i.e. abstract class) above Agent in the hierarchy rather than below.

rht: Precedent in MASON: a `Wall` class (https://github.com/eclab/mason/blob/2f75282b044a9ce5fea978ed220b6498a5f19745/contrib/physics2d/sim/app/robots/Wall.java) that extends [`StationaryObject2D`](https://github.com/eclab/mason/blob/2f75282b044a9ce5fea978ed220b6498a5f19745/contrib/physics2d/sim/physics2D/physicalObject/StationaryObject2D.java) that extends [`PhysicalObject2D`](https://github.com/eclab/mason/blob/2f75282b044a9ce5fea978ed220b6498a5f19745/contrib/physics2d/sim/physics2D/physicalObject/PhysicalObject2D.java) that extends [`SimplePortrayal2D`](https://github.com/eclab/mason/blob/2f75282b044a9ce5fea978ed220b6498a5f19745/mason/src/main/java/sim/portrayal/SimplePortrayal2D.java)

rht: This is how the `Wall` object is used: https://github.com/eclab/mason/blob/2f75282b044a9ce5fea978ed220b6498a5f19745/contrib/physics2d/sim/app/collisions/MobilePoly.java#L27. It is essentially a custom procedure inside a `step()` of an agent (you can see the definition of `checkWallCollision` in the same file).

tpike3: Ohhhh now we are getting to some thorny issues, very exciting. 

Just to be explicit and try and reduce this down to its fundamental levels: 
If I were to build a model now with a "passive" agent (e.g. a wall) I would build an agent object, place it in the multigrid, but not add it to the schedule. This would prevent the agent from slowing down the model as it would never be called in step but would serve its purpose of influence/shaping/limiting agent behavior.

In this context this debate really becomes a user abstraction issue, a "feature" or "fixture" class would be more intuitively obvious but add lines of code.  Per the [matrix conversation](https://matrix.to/#/!CAAdpmuJhdNJkxurrp:matrix.org/$Uo9uVvHzI0WWXdZu1QEyZHeWDAfsCLssl6uvUoFe008?via=matrix.org) I would propose (and please feel free to disagree): 

- we merge this commit
- add to the issues and work to build an "seaborn version" of mesa that is less abstract

rht: Yeah, let's open a new GH issue for this.

PR #1100, opened 2021-12-07
[PoC] Visualization: Update to Bootstrap 4.6.1

EwoutH:
Proof of Concept: Update from Bootstrap [v3.3.7](https://github.com/twbs/bootstrap/releases/tag/v3.3.7) to [v4.6.1](https://github.com/twbs/bootstrap/releases/tag/v4.6.1).

It also updates [bootstrap-slider](https://github.com/seiyria/bootstrap-slider) to 11.0.2 and bootstrap-switch its latest commit, https://github.com/Bttstrp/bootstrap-switch/commit/2b316be8001254ef9f6e229144b625c4674560cb.

**To-do (any help is appreciated!)**
- [ ] Check the [Migrating to v4](https://getbootstrap.com/docs/4.6/migration/#summary) guide and implement necessary changes
- [ ] Check if all files are necessary.
  - [ ] [mesa/visualization/templates/fonts](https://github.com/projectmesa/mesa/tree/main/mesa/visualization/templates/fonts) can probably be removed since built-in fonts are now used.
  - [ ] In [mesa/visualization/templates/css/](https://github.com/projectmesa/mesa/tree/main/mesa/visualization/templates/css) the bootstrap-them files need to be updated of removed.
- [ ] Test as much models as possible.

My working branch is [bootstrap-4](https://github.com/EwoutH/mesa/tree/bootstrap-4). Feel free to work from there!

The branch can be installed with `pip install git+https://github.com/EwoutH/mesa.git@bootstrap-4` for testing.

Screenshot_686Screenshot_687

PR #1220, opened 2022-03-30
js: Download Bootstrap on-the-fly during install instead

rht:
This PR:
- removes the need to attribute Bootstrap in a THIRDPARTY file
- reduces future diffs due to dependencies upgrade because they are no longer vendored entirely as files


rht:
This should also make upgrading Bootstrap easier. @EwoutH if you notice, you miss some of the files in the first commit of #1100, e.g. the fonts and the npm.js.

rht:
> @EwoutH if you notice, you miss some of the files in the first commit of https://github.com/projectmesa/mesa/pull/1100, e.g. the fonts and the npm.js.

I wasn't aware that you mentioned them as candidates for removal. Anyway, they will be auto-removed anyway since they are not included in the 4.6.1 zip file.

EwoutH:
Maybe make the version a variable and use f-strings to incorporate the version in the paths, so that you can upgrade the version by changing a single variable.

rht:
Yeah, I had in mind of putting explicit versions in the src path in mesa/visualization/templates/modular_template.html. This way, there is no collision between versions.

rht:
I have isolated Bootstrap-specific files in the 2nd commit. Now, there is no problem with upgrading Bootstrap. But also notice the complication with having to update MANIFEST.in manually when upgrading Bootstrap.

This PR is ready for review.

setup.py
```diff
@@ -20,6 +24,27 @@
 with open("README.rst", "rb", encoding="utf-8") as f:
     readme = f.read()
 
+# Ensure Bootstrap
+# Important: when you update the Bootstrap version, make sure to also update the
+# harcoded included files and versions in:
```
tpike3, line 29: hardcoded is missing a "d" --- ha I kept subconsciously inserting the "d" and couldn't figure out the issue

tpike3:
I feel  like in this comment and #1219 we are changing the squashed commits approach. I am fine with this as I think it will provided a cleaner historical record, but just want to call it out explicitly. Effectively, I would describe our commit approach as each commit should encapsulate a distinct update. Make sense?

rht:
Yes, multiple commits are fine as long as they do orthogonal, atomic changes.

setup.py
rht: It's not caught by Codespell, ugh.

tpike3 (approved):
LGTM

Issue #667, opened 2019-04-29
Add an example that uses directed graphs in Mesa


jackiekazil:
RE: #653 

Seems like we need to add an example of directed graphs to Mesa. This might be adjusting how the network module works as well. 

You could do this one of two ways. 

1. You could create a directed version of the existing virus on a network, which would mean that the virus could only move in certain directions. 

2. You could recreate this model: http://ccl.northwestern.edu/netlogo/models/DiffusiononaDirectedNetwork 

It seems like #1 would be an easier lift, but I will let you decided what you prefer.

kennethells:
I'll work on this.

rht:
Diffusion on a directed network's code: https://github.com/NetLogo/models/blob/master/Sample%20Models/Networks/Diffusion%20on%20a%20Directed%20Network.nlogo
(it's kind of hard to parse because everything is in 1 file without boundaries)

I think it would cover more systems by porting 2) instead.

rozenlicht:
I've chipped in by creating a PR that adds arrowheads to the D3.js visualization when `"directed": True` is added to edge in the `portrayal` dict inserted in `NetworkModule`. If this is the desired behavior, I could also code it for sigma. 
As the initial issue suggested implementing a "directed virus network" example, I've included this as well.

rht:
This should be very easy to do with the new pure Python/Solara frontend, since it is essentially doing a NetworkX draw.

rht:
@ankitk50 this is another low hanging fruit for warming up. The hardest part would be to find a model that behaves differently with a directed network instead of undirected network. This doesn't have to be from that NetLogo model I linked above.

ankitk50:
this sounds interesting, but also complex, I can add this to my list.

rht:
I think one of the version proposed by @jackiekazil should do:

> You could create a directed version of the existing [virus on a network example](https://github.com/projectmesa/mesa/tree/master/examples/virus_on_network), which would mean that the virus could only move in certain directions.

So it is not going to be a drastic modification over virus on a network.

tpike3:
Closing to make issues more impactful and relevant to current Mesa issues (anything before Mesa 3.0 release Nov 9, 2024) per May 26, 2026 dev meeting.

PR #846, opened 2020-05-17
Added directed edges functionality to D3 Visualization

rozenlicht:
Added DirectedGraph example (VirusOnNetworkDirected). 
Appended arrowhead SVG to D3.js module.
(partially) fixes issue #667

PR #1223, opened 2022-03-30
 Added directed edges functionality to D3 Visualization

AdamCordner641:
Added DirectedGraph example (VirusOnNetworkDirected).
Appended arrowhead SVG to D3.js module.
Fixes issue https://github.com/projectmesa/mesa/issues/667 (partially).
Based off pull request https://github.com/projectmesa/mesa/pull/846.


mesa/visualization/templates/js/NetworkModule_d3.js
```diff
@@ -24,6 +24,45 @@ var NetworkModule = function(svg_width, svg_height) {
             g.attr("transform", d3.event.transform);
         }));
 
+    svg.append("svg:defs").selectAll("marker")
```
rht, line 27: The 2 `svg.append...`'s here are very similar, and so they could be deduplicated to make it clear.

rht: There are 3 of them actually.

examples/virus_on_network_directed/README.md
```diff
@@ -0,0 +1,46 @@
+# Virus on a Network
+
+## Summary
+
+This model is based on the NetLogo model "Virus on Network".
```
rht, line 5: Needs to say that this model is almost the same as examples/virus_on_network except that the graph is directed. I haven't fully read the code to confirm that this is the only difference.

rht:
In the .git/config of your Mesa git repo, make sure to add
```
[pull]
    rebase = true
```

This will prevent the noisy merge commits when you pull from Mesa main to update your branch.

tpike3:
@AdamCordner641 My impression is you are new to open source development so let us know if you have any questions. Everyday I am learning something new.

mesa/visualization/templates/js/NetworkModule_d3.js
AdamCordner641: I've deduplicated the shared code so the changes are as short as possible. Also, the network won't be drawn as directed unless "directed": True in portrayal["edges"]. If "directed": False, or not specified, the network will be drawn as undirected.

examples/virus_on_network_directed/README.md
AdamCordner641: I've added some additional text to clarify that the only difference is that the graph is directed.

rht:
I diff-ed the model.py of this PR with the undirected version:
```
18c18
<     return sum(1 for a in model.grid.get_all_cell_contents() if a.state is state)
---
>     return sum([1 for a in model.grid.get_all_cell_contents() if a.state is state])
33c33
< class VirusOnNetwork(Model):
---
> class VirusOnNetworkDirected(Model):
```

1. ` return sum(1 for a in model.grid.get_all_cell_contents() if a.state is state)` is preferred
2. this means that the model graph is actually undirected, because there is no difference. `erdos_renyi_graph` has `directed` as an argument, which is not used

rht:
Here is a diff of server.py:
```
8c8
< from .model import VirusOnNetwork, State, number_infected
---
> from .model import VirusOnNetworkDirected, State, number_infected
37c37,39
<             "tooltip": f"id: {agents[0].unique_id}<br>state: {agents[0].state.name}",
---
>             "tooltip": "id: {}<br>state: {}".format(
>                 agents[0].unique_id, agents[0].state.name
>             ),
47a50,51
>             "directed": True,
>             "marker_size": "large"
68c72
<         ratio_text = "&infin;" if ratio is math.inf else f"{ratio:.2f}"
---
>         ratio_text = "&infin;" if ratio is math.inf else "{0:.2f}".format(ratio)
138c142
<     VirusOnNetwork, [network, MyTextElement(), chart], "Virus Model", model_params
---
>     VirusOnNetworkDirected, [network, MyTextElement(), chart], "Directed Virus Model", model_params
140c144
< server.port = 8521
---
> server.port = 8521
```

The `"...".format(...)` needs to be replaced with the more modern f-strings.

rht:
@AdamCordner641  I'm not sure how it happened, but you have accidentally included a commit in the main branch in this PR. Have you made sure to add https://github.com/projectmesa/mesa/pull/1223#issuecomment-1084010129 ? And then you do `git pull upstream main`, assuming `projectmesa/mesa` remote is called upstream in your .git/config.

mesa/visualization/templates/js/NetworkModule_d3.js
```diff
@@ -24,6 +24,24 @@ var NetworkModule = function(svg_width, svg_height) {
...
+        .attr("viewBox", "0 -5 10 10")
+        .attr("refX", 15)
+        .attr("refY", 0.5)
+        .attr("orient", "auto")
+        .append("svg:path")
+        .attr("d", "M0,-5L10,0L0,5");
+    }
+
+    var marker_sizes = ["end", "endlarge", "endxl"];
+    for (let index = 0; index < marker_sizes.length; index++) {
```
rht, line 40: Use `for ... of`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

mesa/visualization/templates/js/NetworkModule_d3.js
```diff
@@ -24,6 +24,24 @@ var NetworkModule = function(svg_width, svg_height) {
...
+        addMarkers(svg.append("svg:defs").selectAll("marker").data([marker_sizes[index]]));
+        d3.select(`#${marker_sizes[index]}`).attr("markerWidth", 4 + index).attr("markerHeight", 4 + index);
```
rht, line 42: ? Why is the index necessary here? Also, the 2 lines above seem to be under indented. Use https://github.com/prettier/prettier to auto-format the js file.

rht: The whole file seems to not have been formatted properly. Wait until #1225 before you pull and apply prettier to this PR.

tpike3: #1225 Merged

rht: #1225 has been merged, you need to pull from latest main, and then force push to your branch.

rht:
```
# In your fork's main branch
git pull upstream main
git push -f
```
Assuming that this is specified in your .git/config
```
[remote "upstream"]
	url = git@github.com:projectmesa/mesa.git
	fetch = +refs/heads/*:refs/remotes/upstream/*
```

mesa/visualization/templates/js/NetworkModule_d3.js
AdamCordner641: Updated to use for ... of

mesa/visualization/templates/js/NetworkModule_d3.js
AdamCordner641: Removed use of index with for...of. And applied prettier to PR files.

AdamCordner641:
Pushed and tested fixes which should resolve the issues, but let me know if I missed anything :)

examples/virus_on_network_directed/virus_on_network_directed/model.py
```diff
@@ -0,0 +1,163 @@
...
+
+def number_susceptible(model):
+    return number_state(model, State.SUSCEPTIBLE)
+
+
+def number_resistant(model):
+    return number_state(model, State.RESISTANT)
+
+
+class VirusOnNetwork(Model):
```
rht, line 33: Needs to be `VirusOnNetworkDirected`.

rht: Reminder to rename all occurrences of `VirusOnNetwork` inside examples/virus_on_network_directed/.

mesa/visualization/templates/js/NetworkModule_d3.js
```diff
@@ -80,6 +109,13 @@ var NetworkModule = function (svg_width, svg_height) {
       })
       .attr("stroke", function (d) {
         return d.color;
+      })
+      .attr("marker-end", function (d) {
+        return d.directed
+          ? "url(#end" +
+              (d.marker_size !== undefined ? d.marker_size : "") +
```
rht, line 116: You can do this instead
`(d.marker_size ?? "")`

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

rht:
I still think that the example in this PR doesn't showcase directed networks. It's behaviorally identical to the undirected version.
To make it different, I think we need to implement `get_neighbors` in a new class called `DirectedNetworkSpace` [1] in such a way that it returns either of:
- all in and out neighbors
- all in neighbors only
- all out neighbors only

[1] let's call it "space" instead of "grid"

rht:
I think we should defer this PR, and do a separate PR that implements `DirectedNetworkSpace` with tests. Once this is done, we can revisit this PR.

tpike3:
> I think we should defer this PR, and do a separate PR that implements `DirectedNetworkSpace` with tests. Once this is done, we can revisit this PR.

I concur. @AdamCordner641 Thank you very much for all the hard work you have put into this. Your work definitely revealed some work we need to do. I hope you can use the code you developed in your project and would be happy for any assistance you could provide on building out a `DirectedNetworkSpace.`

examples/virus_on_network_directed/virus_on_network_directed/model.py
AdamCordner641: All occurrences should now be renamed

AdamCordner641:
> > I think we should defer this PR, and do a separate PR that implements `DirectedNetworkSpace` with tests. Once this is done, we can revisit this PR.
> 
> I concur. @AdamCordner641 Thank you very much for all the hard work you have put into this. Your work definitely revealed some work we need to do. I hope you can use the code you developed in your project and would be happy for any assistance you could provide on building out a `DirectedNetworkSpace.`

No problem! Thank you both for helping with PR feedback. I can definitely use these changes I have made in my project. I would also like to be able to use the `get_neighbors` functionality rht described in my project, so I'll try to make time to have a go at implementing a `DirectedNetworkSpace` class.

Corvince:
Thank you for the PR and sorry for never putting it forward

If this is still actively followed please re-open the PR at https://github.com/projectmesa/mesa-viz-tornado
But note that we are moving towards a new pure python frontend

PR #1236, opened 2022-04-03
js deps: Move Bootstrap to be inside external folder

rht:
This is so that the directory content of mesa/visualization/templates
remains neat. And this helps clarify that the content inside external/
is downloaded manually, not part of the repo.

@EwoutH FYI, need to wait for this PR before doing anything in #1100.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,3 +81,6 @@ dmypy.json
 
 # pythonenv
 pythonenv*/
+
+# JS dependencies
+mesa/visualization/templates/external/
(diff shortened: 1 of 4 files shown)


rht:
Also important: while this works on Linux, the `/` pathing might not work on Windows.If someone can test this on Windows, that would be great.

tpike3:
> Also important: while this works on Linux, the `/` pathing might not work on Windows.If someone can test this on Windows, that would be great.

I should be able to test by wednesday and if no issues will merge

tpike3 (approved):
LGTM - install worked on windows

PR #1240, opened 2022-04-04
refactor: Make _place_agent args consistent with place_agent

rht:
Supersedes #694

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -407,15 +407,15 @@ class Grid:
         """
         pos = self.torus_adj(pos)
         self._remove_agent(agent.pos, agent)
-        self._place_agent(pos, agent)
+        self._place_agent(agent, pos)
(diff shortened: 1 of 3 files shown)


mesa/space.py
```diff
@@ -407,15 +407,15 @@ def move_agent(self, agent: Agent, pos: Coordinate) -> None:
         """
         pos = self.torus_adj(pos)
         self._remove_agent(agent.pos, agent)
```
tpike3, line 409: Does `_remove_agent` need to be updated to it always follows (agent, position)?

tpike3 (approved):
LGTM

One question is should the `remove_agent` also be (agent, position) but that can be another commit/PR

rht:
Yes, the making `_remove_agent` consistent will be in a separate PR.

PR #1242, opened 2022-04-04
space: Clarify the return object of get_cell_list_contents

rht:


mesa/space.py
```diff
@@ -15,6 +15,7 @@
 
 import itertools
 import math
+from warnings import warn
```
rht, line 18: This is unrelated refactor, accidentally comitted in. Oh well.

tpike3 (approved):
This is good I am not sure why we are import warn three different times in the same file.

Issue #1231, opened 2022-04-03
UX: Simplify schedulers class names


rht:
What if, instead of
```python
from mesa.time import RandomActivation
self.schedule = RandomActivation(...)
```
we do
```python
import mesa.time
self.schedule = mesa.time.Random(...)
# or, as per https://github.com/projectmesa/mesa/issues/1229
self.schedule = ma.time.Random(...)
```
As in, the context that it is a scheduler, is reflected by the namespace, and so we don't need to redundantly say that it is an "activation":
- `StagedActivation` -> `Staged`
- `RandomActivationByType` -> `RandomByType`
- and so on.

This is inspired by the way Agents.jl name their schedulers.

rht:
We may need to rename `mesa.time` to `mesa.scheduler` for clarity.

quaquel:
all schedulers are deprecated with mesa 3.0

PR #1244, opened 2022-04-04
feat: Implement flat Mesa namespace alternative

### LINKED ISSUE #1229: UX: flat Mesa namespace alternative
Consider that instead of
```python
from mesa import Model, Agent
from mesa.space import SingleGrid
```

We do a single import (like `import matplotlib.pyplot as plt`)
```python
import mesa.all as ma

self.space = ma.SingleGrid
...
self.datacollector = ma.DataCollector(...)
```
This way, we can have modular namespaces that are not polluted, while at the same time have a flat namespace for ease of use.

rht:
Close #1229.
Examples:
```ipython
In [1]: import mesa.all as ma

In [2]: ma.Agent
Out[2]: mesa.agent.Agent

In [3]: ma.Model
Out[3]: mesa.model.Model

In [4]: ma.RandomActivation
Out[4]: mesa.time.RandomActivation

In [5]: ma.SingleGrid
Out[5]: mesa.space.SingleGrid

In [6]: ma.ModularServer
Out[6]: mesa.visualization.ModularVisualization.ModularServer

In [7]: ma.UserSettableParameter
Out[7]: mesa.visualization.UserParam.UserSettableParameter

In [8]: ma.DataCollector
Out[8]: mesa.datacollection.DataCollector
```

I'm thinking of whether it should be called `import mesa.flat as mf` instead of `import mesa.all as ma`.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -5,3 +5,4 @@ branch = True
 [report]
 omit =
     tests/*
+    mesa/all/*
--- a/mesa/all/__init__.py
+++ b/mesa/all/__init__.py
@@ -1,0 +1,5 @@
+from mesa import *  # noqa
+from mesa.time import *  # noqa
+from mesa.space import *  # noqa
+from mesa.datacollection import *  # noqa
+from .visualization import *  # noqa
--- a/mesa/all/visualization.py
+++ b/mesa/all/visualization.py
@@ -1,0 +1,3 @@
+from mesa.visualization.ModularVisualization import *  # noqa
+from mesa.visualization.modules import *  # noqa
+from mesa.visualization.UserParam import *  # noqa


rht:
My contenders so far:
- `import mesa.all as ma`
- `import mesa.flat as mf`
- `import mesa.simple as ms`
- `import mesa.lite as ml`

Let's bring this up in the next dev meeting.

tpike3:
@rht This is awesome, so simple yet very powerful. My gut is says `mesa.all`. @jackiekazil , @EwoutH thoughts?

rht:
Note: I intentionally put the time in `ma.time` because of the concern in #1231.

jackiekazil:
I see the value. Quick question -- what if there are name space collisions? (I guess we handle when they occur?)

jackiekazil:
Also do libraries normally maintain both?

rht:
Yeah, collisions can be handled by hand. One example from other library: `import matplotlib.pyplot as plt`.

Another alternative is to import into `mesa` namespace directly, so we have:
```
import mesa
mesa.DataCollector  # in addition to mesa.datacollection.DataCollector
mesa.UserSettableParameter
...
```

clarkpetri:
Excited to follow this. Loving Mesa so far for my school work.

rht:
Reimplemented in #1294

Issue #1232, opened 2022-04-03
JS refactor: We should replace all usage of `var x = ...` with `const` / `let` in the JS code


rht:

SangameshItagi:
Hi @rht , I can help here

rht:
@SangameshItagi sounds good. Make sure to not do one big PR while doing it. It's better to do several small-medium sized PRs so that it is easier to review and merge.

trickroll:
Tried to tackle a few of the files instead of a big PR!

rht:
To avoid duplication of work:

In #1248, @SangameshItagi is working on
- mesa/visualization/templates/js/CanvasHexModule.js
- mesa/visualization/templates/js/CanvasModule.js
- mesa/visualization/templates/js/GridDraw.js
- mesa/visualization/templates/js/HexDraw.js

In #1246, @trickroll is working on
- mesa/visualization/templates/js/NetworkModule_sigma.js
- mesa/visualization/templates/js/PieChartModule.js
- mesa/visualization/templates/js/TextModule.js

rht:
Remaining JS files up for grabs:
- [ ] BarChartModule.js
- [ ] ChartModule.js
- [ ] NetworkModule_d3.js
- [ ] runcontrol.js

rht:
Closed via #1265.

rht:
I missed one more JS file: there is still InteractionHandler.js. However, we should wait until #1266 is merged.

rht:
Closed via #1273.

PR #1246, opened 2022-04-04
[Issue 1232] Refactor NetworkModuleSigma PieChartModule TextModule JS

trickroll:
Addresses some, but not all, of the files referenced in issue #1232 

Refactored "var" into const/let for:
- NetworkModule_sigma.js
- PieChartModule.js
- TextModule.js

Please tell me if there is anything I may need to fix or change! Thank you and welcome the feedback.

### DIFF
--- a/mesa/visualization/templates/js/NetworkModule_sigma.js
+++ b/mesa/visualization/templates/js/NetworkModule_sigma.js
@@ -1,13 +1,13 @@
-var NetworkModule = function(canvas_width, canvas_height) {
+const NetworkModule = function(canvas_width, canvas_height) {
 
-    var div_tag = "<div id='graph-container' style='width: " +
+    const div_tag = "<div id='graph-container' style='width: " +
         canvas_width + "px; height: " + canvas_height + "px;'></div>";
 
     // Append it to #elements:
-    var div = $(div_tag)[0];
+    const div = $(div_tag)[0];
     $("#elements").append(div);
 
-    var s = {
+    let s = {
         container: 'graph-container',
         settings: {
             defaultNodeColor: 'black',
@@ -17,7 +17,7 @@ var NetworkModule = function(canvas_width, canvas_height) {
     };
 
     this.render = function(data) {
-        var graph = JSON.parse(JSON.stringify(data));
+        const graph = JSON.parse(JSON.stringify(data));
 
         // Update the instance's graph:
         if (s instanceof sigma) {
--- a/mesa/visualization/templates/js/PieChartModule.js
+++ b/mesa/visualization/templates/js/PieChartModule.js
@@ -2,25 +2,24 @@
 //Note: This pie chart is based off the example found here:
 //https://bl.ocks.org/mbostock/3887235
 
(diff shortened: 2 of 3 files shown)


Corvince:
Hi @trickroll and thanks for the contribution! I am not sure which rule you applied for deciding between let and const. Normally you should use const whenever possibly and only use let if it is not possible otherwise.

trickroll:
@Corvince Thank you for feedback! Let me revisit the const vs let. Apologies about the thrash and appreciate the patience. Understand that there are benefits to making variables immutable

mesa/visualization/templates/js/NetworkModule_sigma.js
```diff
@@ -1,13 +1,13 @@
...
+    const div_tag = "<div id='graph-container' style='width: " +
         canvas_width + "px; height: " + canvas_height + "px;'></div>";
 
     // Append it to #elements:
-    var div = $(div_tag)[0];
+    const div = $(div_tag)[0];
     $("#elements").append(div);
 
-    var s = {
+    let s = {
```
trickroll, line 10: Changed to let since s can be reassigned in line 30

mesa/visualization/templates/js/PieChartModule.js
```diff
@@ -2,25 +2,25 @@
...
     // Create the overall chart div
-    var chart_div_tag = "<div class='pie chart' width='" + canvas_width + "'></div>";
-    var chart_div = $(chart_div_tag)[0];
+    const chart_div_tag = "<div class='pie chart' width='" + canvas_width + "'></div>";
+    const chart_div = $(chart_div_tag)[0];
     $("#elements").append(chart_div);
 
     // Create the tag:
-    var svg_tag = "<svg width='" + canvas_width + "' height='" + canvas_height + "' ";
+    const svg_tag = "<svg width='" + canvas_width + "' height='" + canvas_height + "' ";
```
rht, line 12: This will cause error because of `svg_tag += ...` right after. The solution is to stick to `const`, and combine L12 and L13.

rht:
Everything else looks good to me.
Make sure to:
- Squash the commits into 1 commit
- Use imperative voice when writing the commit message, i.e. "refactor: Replace var with const/let in some JS files". You can edit a commit's message by doing `git commit --amend`. https://github.com/projectmesa/mesa/pull/1247#issuecomment-1088126125= has a guide on this

rht:
If you want to programmatically catch these reassignment errors, you can use [eslint](https://eslint.org/).
After installing, you need to create .eslintrc.js in the root of your Git repo, containing this:
```
module.exports = {
    "env": {
        "browser": true,
    },
    "extends": "eslint:recommended",
    "parserOptions": {
        "ecmaVersion": 12
    },
    "rules": {
    }
};
```

It should tell you this error if you specify `svg_tag` as `const` while doing the `+=`:
```
15:5   error  'svg_tag' is constant                                no-const-assign
```
Pinging @SangameshItagi because this info is useful for him as well.

rht:
LGTM

rht:
Merged, thank you!

PR #1250, opened 2022-04-05
CI test: Increase timeout to 10 minutes

rht:
The PyPy CI consistently errors out, because it keeps failing to install the new dependencies, because the cache has been invalidated. I'm increasing the timeout here.

### DIFF
--- a/.github/workflows/build_lint.yml
+++ b/.github/workflows/build_lint.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ${{ matrix.os }}-latest
     # We need an explicit timeout because sometimes the batch_runner test never
     # completes.
-    timeout-minutes: 6
+    timeout-minutes: 10
     strategy:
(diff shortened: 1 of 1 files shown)


tpike3 (approved):

rht:
@tpike3 the PyPY build still failed. 10 min is not long enough. Either:
- we drop PyPy support in dev. We run PyPy tests only for the release build. The build failure emails are too annoying. 
- we increase the timeout until PyPy build successfully, which could be 30 min or worse

tpike3:
> @tpike3 the PyPY build still failed. 10 min is not long enough. Either:
> 
> * we drop PyPy support in dev. We run PyPy tests only for the release build. The build failure emails are too annoying.
> * we increase the timeout until PyPy build successfully, which could be 30 min or worse

I vote for option 1 - drop PyPy support in dev, run PyPy tests only for the release build

PR #1251, opened 2022-04-05
batch_run: Print the correct iteration number in output

EwoutH:
Currently the `batch_run()` function in `batchrunner.py` print the iteration number from -1 to N-2. This commits changes that to 1 to N.

For example: `batch_run(iterations=5)` now returns values from 1 to 5 instead of -1 to 3.

### DIFF
--- a/mesa/batchrunner.py
+++ b/mesa/batchrunner.py
@@ -87,7 +87,7 @@ def batch_run(
                     run_id = next(run_counter)
                     data = []
                     for run_data in rawdata:
-                        out = {"RunId": run_id, "iteration": iteration - 1}
+                        out = {"RunId": run_id, "iteration": iteration + 1}
                         out.update(run_data)
                         data.append(out)
                     results.extend(data)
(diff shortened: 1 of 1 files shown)


mesa/batchrunner.py
```diff
@@ -102,7 +102,7 @@ def batch_run(
                     run_id = next(run_counter)
                     data = []
                     for run_data in rawdata:
-                        out = {"RunId": run_id, "iteration": iteration - 1}
+                        out = {"RunId": run_id, "iteration": iteration + 1}
```
rht, line 105: Does this `iteration` for the mp version really start from 0?

EwoutH: Yes, it's defined as a range: https://github.com/projectmesa/mesa/blob/f390714fa4eae5c6c8956a00b377db6a376ce88d/mesa/batchrunner.py#L84

rht: That's true only when the `number_processes` is 1. Otherwise, it is an element from a `Counter` object.

EwoutH: That's implemented a bit weird

rht:
Bump. You just need to fix is on the single-process code branch of `batch_run`. The multi-process branch is fine.

rht:
Superseded by #1310.

PR #1254, opened 2022-04-06
ci: Disable PyPy for now

rht:


tpike3:
This should be fixed by #1256 correct?

rht:
This is a separate error, caused by PyPy taking too long to install dependencies.

tpike3 (approved):

PR #1255, opened 2022-04-06
test: Remove batchrunnerMP

rht:
I suspect this is the cause of the hang. I will try running the CI 10 times and see the result.


rht:
10 out of 10 the CI passed when test_batchrunnerMP.py is removed. I have found the cause to the CI hang!

PR #1257, opened 2022-04-07
js: Download bootstrap-slider during install

rht:


rht:
Once this is merged, the procedure for doing the same for the rest of the JS dependencies are straightforward, and I will that in a separate single PR>

tpike3 (approved):
LGTM

Issue #1258, opened 2022-04-07
We should apply Prettier to the JS files


rht:
i.e. https://prettier.io/
An example PR can be found at #1225. Make sure to not apply Prettier to the external dependencies.

akrabdev:
can assign this to me

rht:
Assigned.

rht:
Closed via #1266.

PR #1265, opened 2022-04-07
refactor: Change var to const/let for remaining js files

trickroll:
Update remaining files for #1232:
-  BarChartModule.js
-  ChartModule.js
-  NetworkModule_d3.js
-  runcontrol.js

PR #1266, opened 2022-04-07
style: prettify js files

akrabdev:
This  is addressing #1258. By running Prettier formatter on js files in `mesa/visualization/templates/..` I ignored `*.min.js` files as well.

I used the default prettier configuration which was also used in #1225 

I recommend adding prettier config file `.prettierrc` and `.prettierignore`  in the future.

### DIFF
--- a/mesa/visualization/templates/js/BarChartModule.js
+++ b/mesa/visualization/templates/js/BarChartModule.js
@@ -1,153 +1,186 @@
-'use strict';
+"use strict";
 // Note: This grouped bar chart is based off the example found here:
 // https://bl.ocks.org/mbostock/3887051
-const BarChartModule = function(fields, canvas_width, canvas_height, sorting, sortingKey) {
-    // Create the overall chart div
-    const chart_div_tag = "<div class='bar chart' width='" + canvas_width + "'></div>";
-    const chart_div = $(chart_div_tag)[0];
-    $("#elements").append(chart_div);
+const BarChartModule = function (
+  fields,
+  canvas_width,
+  canvas_height,
+  sorting,
+  sortingKey
+) {
+  // Create the overall chart div
+  const chart_div_tag =
+    "<div class='bar chart' width='" + canvas_width + "'></div>";
+  const chart_div = $(chart_div_tag)[0];
+  $("#elements").append(chart_div);
 
-    // Create the tag:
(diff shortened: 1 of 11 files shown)


rht:
Thank you for using conventional commit message (https://www.conventionalcommits.org/en/v1.0.0/). It's not enforced in this repo, but always good. Sorry for the merge conflict, this is because of #1265, which is done manually.

Note that bootstrap-switch is an external dependency.

We are doing standard Prettier, so .prettierrc is not needed for now. But adding .prettierignore (in a separate PR) is a good idea. The external dependencies will soon be moved to a separate folder that can then be ignored.

akrabdev:
I reverted changes to boostrap-switch.js and resolved the conflicts

rht:
@akrabdev it would be great to squash the commits into 1 commit, because logically they are within the same theme. It is fine to have multiple commits in a PR as long as each commit does logically different change. This is the convention used in projectmesa/mesa.

rht:
First, you should rebase your branch on top of projectmesa/mesa's main: `git pull --rebase upstream main`, assuming you have already defined upstream as a remote in your .git/config.

akrabdev:
Commits squashed!, thank you for the heads up

rht:
@akrabdev you haven't removed the bootstrap-switch.js in your latest commit. Also, you haven't added NetworkModule_d3.js

akrabdev:
@rht cleaned up the mess. `NetwrokModule_d3.js` already formatted as in #1225

rht:
I forgot that I did the NetworkModule_d3.js one.

rht:
LGTM

akrabdev:
Thank you too for the detailed instructions! @rht 

when the state of js files  is ready to add `.prettierignore` , I can work on it.
 
Also we can look into [pre-commit hook](https://prettier.io/docs/en/precommit.html) tool to ensure committed .js files are  formatted.

rht:
Sounds good.

> Also we can look into [pre-commit hook](https://prettier.io/docs/en/precommit.html) tool to ensure committed .js files are formatted.

Currently, there is no `npm` config being set up in the development environment of Mesa. We haven't decided yet on whether we should do this, because it increases the effort for the setup. For now, the `prettier` can be run in the CI instead.

Issue #1235, opened 2022-04-03
UX: Split UserSettableParameter into various classes


rht:
What if, instead of
```python
number_option = UserSettableParameter('number', 'My Number', value=123)
boolean_option = UserSettableParameter('checkbox', 'My Boolean', value=True)
slider_option = UserSettableParameter('slider', 'My Slider', value=123, min_value=10, max_value=200, step=0.1)
```

we do
```python
number_option = Number("My Number", 123)
boolean_option = Checkbox("My boolean", True)
slider_option = Slider("My slider", 123, min=10, max=200, step=0.1)
```

In addition to looking simpler, more importantly, this allows users to define their own UserParam class.

Corvince:
I think this is an excellent idea.

PR #1272, opened 2022-04-10
feat: Implement Slider UserParam class

rht:
Partially addresses #1235.

### DIFF
--- a/examples/wolf_sheep/wolf_sheep/server.py
+++ b/examples/wolf_sheep/wolf_sheep/server.py
@@ -51,20 +51,15 @@ model_params = {
     "grass": mesa.visualization.UserSettableParameter(
         "checkbox", "Grass Enabled", True
     ),
-    "grass_regrowth_time": mesa.visualization.UserSettableParameter(
-        "slider", "Grass Regrowth Time", 20, 1, 50
+    "grass_regrowth_time": mesa.visualization.Slider("Grass Regrowth Time", 20, 1, 50),
+    "initial_sheep": mesa.visualization.Slider(
+        "Initial Sheep Population", 100, 10, 300
     ),
-    "initial_sheep": mesa.visualization.UserSettableParameter(
-        "slider", "Initial Sheep Population", 100, 10, 300
+    "sheep_reproduce": mesa.visualization.Slider(
+        "Sheep Reproduction Rate", 0.04, 0.01, 1.0, 0.01
     ),
-    "sheep_reproduce": mesa.visualization.UserSettableParameter(
-        "slider", "Sheep Reproduction Rate", 0.04, 0.01, 1.0, 0.01
-    ),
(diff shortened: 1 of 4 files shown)


rht:
Given that changing from `UserSettableParameter` to `Slider` in the wolf-sheep example won't work on 0.9.0, I need to remove the change to the wolf-sheep in this PR. This means that the examples can only be changed once there is a 0.10.0 or 1.0.0 release with the `Slider`.

Corvince:
I think the change itself looks good, but I would find it a bit weird to have only Slider separated into its own Class. I know this is temporary, but maybe you could target `projectmesa/userparam` instead of `projectmesa/main` and we collect the individual UserParameter PRs there and once everything is ready we merge into `main`. This way the PRs remain small until everything is ready and merging `userparam` into `main` should be a fast review, since everything has been reviewed before.

rht:
That will do, only if this PR contains breaking change. But the `Slider` class is an addition to the existing `UserSettableParameters`, and so it can be in `main` at any time without breaking compatibility.

One concern with accumulating the changes into a feature branch is that the regular rebases on top of main may require manual intervention, because of the huge diff.

Corvince:
I know the changes are not breaking, but I just find it confusing to have only a set of X UserParameters outside of UserSettableParameters, where X changes over time and is not guaranteed to eventually include all Parameters. 

And what do you mean with regular rebases? I am not experienced with rebases vs merges, is this necessary for rebases? For merging this isn't needed until a merge conflict arises and in that case you have to resolve that either way.

rht:
I don't find it a problem if the main branch doesn't contain all the features yet. It is only problematic if the next release doesn't have all the new user params class. This PR is quick to implement anyway.

Rebase is necessary if the branch becomes far behind the main branch. It will also have conflicts if the diff can't be resolved automatically.

> Given that changing from UserSettableParameter to Slider in the wolf-sheep example won't work on 0.9.0, I need to remove the change to the wolf-sheep in this PR. This means that the examples can only be changed once there is a 0.10.0 or 1.0.0 release with the Slider.

For this, I think it is fine for this PR to modify the examples, as long as people reading the RTD are aware that they are on the development branch, not release.

Corvince:
> I don't find it a problem if the main branch doesn't contain all the features yet. It is only problematic if the next release doesn't have all the new user params class. This PR is quick to implement anyway.

Yes, but until the other user param classes are merged this either becomes a blocker for a new release or we ship with an inconsistent state. Just imagine someone finding a critical bug tomorrow and we need to release new version asap. 

But anyway since this isn't such an important feature at all I think it is better to move forward - so I am pro merging if at least one other person from the core developer teams agrees that this is a good feature (I already said so myself in #1235 , but no one else joined the discussion so far)

rht:
OK I will implement the rest.

Issue #1274, opened 2022-04-10
docs: Improve examples install instructions


rht:
If you look at https://github.com/projectmesa/mesa/graphs/traffic, #315 is the 3rd most visited page. This means that the readme of the examples are not clear enough to indicate that they must be in the directory containing the requirements.txt.

rht:
#1275 is a good example on how to clarify the readme. Help wanted for the rest of the examples!

Corvince:
I don't think #1275 is a good way to solve this problem. I'll explain myself later (probably tomorrow) but please don't change the other examples before having a discussion

tpike3:
@Corvince intriguing, I look forward to seeing what you are thinking.

Corvince:
Yes, sorry for not stating my concerns directly, but I was on the move and just saw that recently there have been a lot of contributions to mesa, which is great, but I feared someone would start working on this directly. And I also wanted to commit myself ;)

Anyways, I have 2 main concerns with instructing people to use git to clone the repository: 
  1. `git` is not always available. If people trip over `pip install -r requirements` (which I think I also did using mesa for the first time), they are probably rather inexperienced. People without a git installation (i.e. primarly windows users) will probably trip over a failing `git clone` as well.
  2. If people manage to clone the repository, they now have a local, but uninstalled copy of mesa. Chances are they will try to modify mesa code and wonder why it doesn't work, not realizing this is not the mesa code they use when doing `import mesa`.

I am not sure what the best intermediant solution is, but the cleanest solution I always had on top of my mind was to create a `mesa download_examples` command that downloads and extracts a zip of the matching examples. But we would need to provide that versioned zip. We could automate this with each release and for example use Github Pages, so we have a link like `https://projectmesa.github.io/examples/v0.9.0/examples.zip`. Versioning is also important because currently the examples might be out of sync with the latest or certainly older mesa releases and even if people manage to clone them might not be able to run them.

rht:
#1228 is another option.

Corvince:
> #1275 may not be a comprehensive solution to the problem, but at least it makes the steps explicit, instead of assuming that people have already cloned the Git repo. People will see the `git clone` command and look up on the internet on how to set up Git. They no longer assume: "oh, `pip install mesa` is sufficient!".

Hmm I think we are getting somewhere, where we make too many assumptions on what people might or might not think. But I don't follow the argument why people would think `pip install mesa` is sufficient when it fails to install the requirements vs when git is not installed. 

[...]

PR #1275, opened 2022-04-10
examples: Clarify install instruction in wolf_sheep

rht:
Partially addresses #1274.


tpike3 (approved):
Thanks! I had not realized this was such a big issue for some many.

PR #1278, opened 2022-04-12
Visualization changes

GaelLucero:
Implemented additional core agent functionalities including adding a heading for rotational movements (forward, backwards, right, left, setxy). As well as, removal from scheduler and grid (die)

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -28,6 +28,7 @@ class Agent:
         self.unique_id = unique_id
         self.model = model
         self.pos: Optional[Position] = None
+        self.heading = 90
 
     def step(self) -> None:
         """A single step of the agent."""
@@ -35,6 +36,10 @@ class Agent:
 
     def advance(self) -> None:
         pass
+
+    def setxy(self, x,y):
+        """Sets the current position to the specified x,y parameters"""
+        self.pos = (x,y)
 
     @property
     def random(self) -> Random:
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -42,7 +42,10 @@ class Agent:
         new_x = float(self.pos[0]) + sign * math.cos(self.heading * math.pi / 180) * amount
         new_y = float(self.pos[1]) + sign * math.sin(self.heading * math.pi / -180) * amount
         next_move = (new_x, new_y)
-        self.model.space.move_agent(self, next_move)
+        try:
+            self.model.space.move_agent(self, next_move)
+        except:
+            print("agent.py (forward_backwards): could not move agent within self.model.space")
 
     def forward(self, amount):
         """Moves the agent forward by the amount given"""
@@ -74,8 +77,14 @@ class Agent:
 
     def die(self):
         """Removes the agent from the schedule and the grid """
-        self.model.schedule.remove(self)
-        self.model.space.remove_agent(self)
+        try:
+            self.model.schedule.remove(self)
+        except:
+            print("agent.py (die): could not remove agent from self.model.schedule")
+        try:
+            self.model.space.remove_agent(self)
+        except:
+            print("agent.py (die): could not remove agent from self.model.space")
 
     @property
     def random(self) -> Random:
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -73,7 +73,7 @@ class Agent:
         return self.distancexy(another_agent.pos[0], another_agent.pos[1])
 
     def die(self):
-        """Removes the agent from the schedule and the grid"""
+        """Removes the agent from the schedule and the grid """
         self.model.schedule.remove(self)
         self.model.space.remove_agent(self)
 
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -72,6 +72,11 @@ class Agent:
         """Gives you the distance between the agent and another agent"""
         return self.distancexy(another_agent.pos[0], another_agent.pos[1])
 
+    def die(self):
+        """Removes the agent from the schedule and the grid"""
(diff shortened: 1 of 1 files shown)


rht:
Thank you for the PR! I have seen @aiwhoo's presentation on your behalf from last month, and has been waiting for this.

mesa/agent.py
```diff
@@ -28,6 +28,7 @@ def __init__(self, unique_id: int, model: "Model") -> None:
         self.unique_id = unique_id
         self.model = model
         self.pos: Optional[Position] = None
+        self.heading = 90
```
rht, line 31: I think we should define a new class for the new attribute and methods, `AgentWithHeading` / `AgentWithDirection` / if you have a better name for it. Because not all models need the feature.

mesa/agent.py
```diff
@@ -36,6 +37,81 @@ def step(self) -> None:
     def advance(self) -> None:
         pass
 
+    def forward_backward(self, amount, sign):
```
rht, line 40: I think it is clearer to specify the names with verbs: `move_forward_or_backward`, `move_forward`, `move_backward`. This conveys that they are actions.

mesa/agent.py
```diff
@@ -36,6 +37,81 @@ def step(self) -> None:
...
+
+    def forward(self, amount):
+        """Moves the agent forward by the amount given"""
+        self.forward_backward(amount, 1)
+
+    def back(self, amount):
+        """Moves the agent backwards from where its facing by the given amount"""
+        self.forward_backward(amount, -1)
+
+    def right(self, degree):
```
rht, line 58: Similarly, `turn_right`, and so on, is clearer.

mesa/agent.py
```diff
@@ -36,6 +37,81 @@ def step(self) -> None:
...
+        """Turns the agent right by the given degree"""
+        self.heading = (self.heading - degree) % 360
+
+    def left(self, degree):
+        """Turns the agent left by the given degree"""
+        self.heading = (self.heading + degree) % 360
+
+    def setxy(self, x, y):
```
rht, line 66: What about `set_position(self, position)`? This way, the API remains the same in the implementation with higher number of dimensions, instead of e.g. `set_xyz` for 3D.

mesa/agent.py
```diff
@@ -36,6 +37,81 @@ def step(self) -> None:
     def advance(self) -> None:
         pass
 
+    def forward_backward(self, amount, sign):
+        """Does the calculation to find  the agent's next move and is used within the forward and backward functions"""
+        new_x = float(self.pos[0]) + sign * math.cos(self.heading * math.pi / 180) * amount
+        new_y = float(self.pos[1]) + sign * math.sin(self.heading * math.pi / -180) * amount
+        next_move = (new_x, new_y)
```
rht, line 44: Should be `new_pos`. `next_move` implies the variable is an action, like "move up".

mesa/visualization/modules/simple_continuous_canvas.js
```diff
@@ -0,0 +1,332 @@
+var ContinuousVisualization = function(height, width, context, background_src=null) {
```
rht, line 1: Need to use `const`/`let` instead of `var` in all of the `var` usage in this file. You case see how to do it in #1246.

mesa/agent.py
rht: `setxy` needs to be removed now that we have `set_pos`.

mesa/visualization/modules/simple_continuous_canvas.js
rht: The entire file needs to be formatted with [`prettier`](https://prettier.io/).

rht:
I think this PR is too big, and will take a while to be merged. It is better to make a separate PR that does only the Python part.

Corvince:
+1 for giving this a new name. From skimming the code this is primarily supposed to be for the continuous space, right? I think the name should be somewhat representative for that (don't know if it also works for grid, but I guess at least not for network)

aiwhoo:
We're working on breaking this PR into 2 pull requests: one for the ABM language (e.g. fd, bk, rotate, etc) in agents.py, the other for the visualization improvements (canvas,  background images, shapes).

The ABM language would be for continous space, while the visualization improvements would be for all spaces.

tpike3:
> We're working on breaking this PR into 2 pull requests: one for the ABM language (e.g. fd, bk, rotate, etc) in agents.py, the other for the visualization improvements (canvas, background images, shapes).
> 
> The ABM language would be for continous space, while the visualization improvements would be for all spaces.

@aiwhoo Thanks for the update!

rht:
Check out the reserved fields of hash.ai's hengine for inspiration for the names: https://hash.ai/docs/simulation/creating-simulations/anatomy-of-an-agent/state#Reserved-Fields.
```javascript
  // Can be used for custom movement logic. Will affect the agent's visualization
  // Suggested values: 0 to 1 for each axis
  "direction": [x, y, z],

  // Can be used for custom movement logic. Will affect the agent's visualization
  // Suggested values: 0 to 1 for each axis
  "velocity": [x, y, z],
```

They most likely also have methods associated with those attributes. It's good to check their naming choice.

rht:
For context, hash.ai's hEngine is a performant ABM library written in Rust, with interfaces to Javascript, Python, and other languages.

jackiekazil:
Should we close this in lieu of two PRs or will this be one of the PRs?

quaquel:
closing this as stale, but linked to it from #2967

PR #1281, opened 2022-04-16
Improve MANIFEST.in

rht:
Aside, if you look at  https://github.com/holoviz/holoviews/blob/master/MANIFEST.in, they seem to include the examples inside the installed `pip` package. This makes it easy to initialize the examples after a `pip install`.

```
conda install -c pyviz holoviews bokeh
holoviews --install-examples
cd holoviews-examples
```

(see https://github.com/projectmesa/mesa/issues/246#issuecomment-511218671=)


jackiekazil:
LGTM

Issue #1267, opened 2022-04-07
CSS support for custom visualization element


wang-boyu:
**What's the problem this feature will solve?**

Currently custom visualization elements can only be added as JS scripts through `local_includes`.

**Describe the solution you'd like**

Have a similar function as `local_includes` for CSS files.

**Additional context**

mesa-geo uses Leaflet for map visualization. Both `leaflet.js` and `leaflet.css` are required.

wang-boyu:
Addressed in https://github.com/projectmesa/mesa/pull/1283.

Issue #1210, opened 2022-03-25
batch_run should default to running on 1 CPU


rht:
Currently, it uses up all the CPUs available in a system. This causes complication with the tutorial for Windows, where a nasty workaround is needed for `multiprocessing` (i.e. the `freeze_support`).

Then, if `number_processes` is > 1 and on Windows, we show a warning that tells user to add `freeze_support()` (we pre-emptively tell the user of this issue instead of waiting for it to happen). This warning can be disabled via a flag to the function:
```python
batch_run(
    ...
   hide_windows_freeze_support_warning=True,
   ...
)
```

Originally posted in https://github.com/projectmesa/mesa/discussions/1202.

PR #1283, opened 2022-04-16
support package and local css includes

wang-boyu:
This PR tries to address https://github.com/projectmesa/mesa/issues/1267. The user interface remains the same, e.g.:

```python
class MyCustomModule(VisualizationElement):
    package_includes = ["PackageJavaScript.js", "PackageStylesheet.css"]
    local_includes = ["MyCustomJavaScript.js", "MyCustomStylesheet.css"]
```

Then these included files get saved into `package_js_includes`, `package_css_includes`, `local_js_includes` and `local_css_includes` attributes. Whether it's a stylesheet or not depends on whether the filename ends with `.css`.

**An alternative approach** that is not implemented with this PR would be to let users decide by themselves on demand. For example like this:

```python
class MyCustomModule(VisualizationElement):
    package_js_includes = ["PackageJavaScript.js"]
    package_css_includes = ["PackageStylesheet.css"]  # this is what Mesa-Geo needs
    local_js_includes = ["MyCustomJavaScript.js"]
    local_css_includes = ["MyCustomStylesheet.css"]
```

In this way we don't need to check whether files are stylesheets or scripts. But this breaks API and introduces more complexity in developing customized elements.

Please feel free to suggest other possible approaches.

### DIFF
--- a/mesa/visualization/ModularVisualization.py
+++ b/mesa/visualization/ModularVisualization.py
@@ -125,10 +125,10 @@ class VisualizationElement:
     Defines an element of the visualization.
 
     Attributes:
-        package_includes: A list of external JavaScript files to include that
-                          are part of the Mesa packages.
-        local_includes: A list of JavaScript files that are local to the
-                        directory that the server is being run in.
+        package_includes: A list of external JavaScript and CSS files to
+                          include that are part of the Mesa packages.
+        local_includes: A list of JavaScript and CSS files that are local to
+                        the directory that the server is being run in.
         js_code: A JavaScript code string to instantiate the element.
 
     Methods:
@@ -174,8 +174,10 @@ class PageHandler(tornado.web.RequestHandler):
             port=self.application.port,
             model_name=self.application.model_name,
             description=self.application.description,
-            package_includes=self.application.package_includes,
-            local_includes=self.application.local_includes,
+            package_js_includes=self.application.package_js_includes,
+            package_css_includes=self.application.package_css_includes,
+            local_js_includes=self.application.local_js_includes,
+            local_css_includes=self.application.local_css_includes,
             scripts=self.application.js_code,
         )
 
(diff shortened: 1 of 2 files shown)


rht:
I think the JS ones should be `package_includes` and `local_includes` still, so that this PR does not introduce a breaking change. We haven't decided whether the next release will be 0.10.0 or 1.0.0 yet.

wang-boyu:
Oh sorry, perhaps I miscommunicated...

This PR (the current implementation) does not break API - both JS and CSS files are specified using `package_includes` and `local_includes` by users. We decide internally whether they are scripts or stylesheets.

Are you suggesting we go for the alternative approach, but without breaking the API? For example:

```python
class MyCustomModule(VisualizationElement):
    package_includes = ["PackageJavaScript.js"]       # same as before
    package_css_includes = ["PackageStylesheet.css"]  # new attribute
    local_includes = ["MyCustomJavaScript.js"]        # same as before
    local_css_includes = ["MyCustomStylesheet.css"]   # new attribute
```

rht:
```python
class MyCustomModule(VisualizationElement):
    package_includes = ["PackageJavaScript.js"]       # same as before
    package_css_includes = ["PackageStylesheet.css"]  # new attribute
    local_includes = ["MyCustomJavaScript.js"]        # same as before
    local_css_includes = ["MyCustomStylesheet.css"]   # new attribute
```

Yes, this one.

mesa/visualization/ModularVisualization.py
```diff
@@ -174,8 +174,10 @@ def get(self):
             port=self.application.port,
             model_name=self.application.model_name,
             description=self.application.description,
-            package_includes=self.application.package_includes,
-            local_includes=self.application.local_includes,
+            package_js_includes=self.application.package_js_includes,
```
rht, line 177: The `self.application.package_js_includes` breaks e.g. https://github.com/projectmesa/mesa/blob/eb73aac0cf7bf129a4ac93187ff76a4fd9e3ae39/mesa/visualization/modules/ChartVisualization.py#L45.

rht:
At the same time, implementing #1210 will necessarily introduce breaking change (this change is necessary to simplify the tutorial for Windows and macOS). Maybe we should go to 1.0.0 instead of 0.10.0 after all (@tpike3 @jackiekazil)? Regardless, the rename to `package_js_includes` and `local_js_includes` can be done in a separate PR.

mesa/visualization/ModularVisualization.py
wang-boyu: `ChartModule` inherits from `VisualizationElement` (which still has `package_includes` and `local_includes` unchanged):

The `package_js_includes`, `package_css_includes`, `local_js_includes` and `local_css_includes` are changed inside `ModularServer`. This breaks API unless someone subclass it (actually perhaps I could do it in Mesa-Geo).

It might be better to explain the current implementation with a diagram:

```mermaid
classDiagram
class VisualizationElement
VisualizationElement: +package_includes
VisualizationElement: +local_includes
class ChartModule
ChartModule: +package_includes
ChartModule: +local_includes
VisualizationElement <|-- ChartModule
class ModularServer
ModularServer: +package_js_includes
ModularServer: +package_css_includes
ModularServer: +local_js_includes
ModularServer: +local_css_includes
```

rht: Ah, I didn't see that you didn't rename the `element.package_includes` (I thought you renamed all of them). There is no API breaking then. In that case, you need to update the docstring of `VisualizationElement` to say that `package_includes` and `local_includes` now support CSS files.

wang-boyu: Yup updated the doctring.

rht:
LGTM

Issue #1280, opened 2022-04-14
Batchrunner does not handle exotic input parameters?


jwalzberg:
My model class takes inputs such as below:

class ExampleModel(Model):
    def __init__(self,
                 seed: int = 0,
                 input_data: Type[pd.DataFrame] = pd.read_csv('InputData.csv'),
                 scenario: str = 'baseline',
                 social_network_parameters: dict = {
                     "average_degree": 10, "rewiring_probability": 0.1},
                 year_period: int = 30,
                 etc. ) -> None:

It would be great if we could have a way to iterate through a list of strings for instance. In my case, the scenario parameter will select part of the Data Frame input based on the scenario string that is given. Ideally, I would like to give batch_run a list such as: ['baseline', 'scenario1', 'scenario2'].

It seems that because the batch_run function wants to launch batches based on iterables it takes the name of my first Data Frame column as an input for the input_data parameter or the first letter of my string (i.e., 'b' in 'baseline' for the example above) for the scenario parameter.

Although it was already not possible to do too many exotic things with input parameters to iterate through in the old BatchRunner classes, at least we could have strings, dictionaries, and so on as input parameters because of the "fixed_parameters" input of the BatchRunner classes. Now it seems batch_run will base its batch of simulations based on any iterable. I am not sure how that could be fixed but I think it could improve Mesa's functionalities. 

PS: I do love the simplicity of the new batch_run function, so I am not advocating going back to the old BatchRunner classes but just improving batch_run.

rht:
I think being able to pass in a DataFrame as params to `batch_run` is a useful feature.

Axeln78:
Hi @jwalzberg. 
I have the same type of issue, a str passed as a parameter to batch_run is then interpreted character by character, not ideal! 
Some parameters might not be able to be put as default args in the constructor of the model class. Then what is best practice?

rht:
> I have the same type of issue, a str passed as a parameter to batch_run is then interpreted character by character, not ideal!

I actually don't fully understand this problem. Shouldn't it work if you wrap the string inside a list?
Here is an example
```python
from mesa.batchrunner import _make_model_kwargs
print(_make_model_kwargs({"key": "value"}))
print(_make_model_kwargs({"key": ["value"]}))
```
which outputs
```
[{'key': 'v'}, {'key': 'a'}, {'key': 'l'}, {'key': 'u'}, {'key': 'e'}]
[{'key': 'value'}]
```

rht:
```
[...]

PR #1289, opened 2022-04-22
batch_run: Do not iterate values when it is a single string

rht:
This addresses one of the concern in #1280.


Corvince:
I originally did not include this check, because I wanted to preserve the use case of actually using strings as iterables. But this is still possible as `['s', 't', 'r', 'i', 'n', 'g']` and is probably seldom needed, whereas the current required workaround of `['string']` for strings as constants leads to more confusion.

so :+1: for this PR

tpike3 (approved):
LGTM

PR #1290, opened 2022-04-23
examples: Initialize Sugarscape ({G1}, {M, T})

rht:


.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/ambv/black
-    rev: 22.1.0
+    rev: 22.6.0
```
rht, line 3: This is tangent, hence needs to be in its own commit to the main branch.

examples/sugarscape_g1mt/sugarscape_g1mt/model.py
```diff
@@ -49,21 +48,21 @@ def __init__(self, width=50, height=50, initial_population=100, seed=42):
...
+                        [a.prices for a in m.schedule.agents_by_type[Trader].values()]
                     )
                 ),
             },
         )
 
         # Create sugar
         sugar_distribution = np.genfromtxt("sugarscape_g1mt/sugar-map.txt")
-        spice_distribution = sugar_distribution.T
+        spice_distribution = np.flip(sugar_distribution, 1)
```
rht, line 65: What is the reasoning for this?

examples/sugarscape_g1mt/sugarscape_g1mt/model.py
```diff
@@ -110,27 +109,27 @@ def step(self):
             sugar.step()
         for spice in self.schedule.agents_by_type[Spice].values():
             spice.step()
-        ssAgents = self.schedule.agents_by_type[SsAgent].values()
-        for agent in ssAgents:
+        Traders = self.schedule.agents_by_type[Trader].values()
```
rht, line 112: Needs to be `traders`, not `Traders`.

examples/sugarscape_g1mt/sugarscape_g1mt/server.py
```diff
@@ -28,15 +29,25 @@ def SsAgent_portrayal(agent):
         portrayal["w"] = 1
         portrayal["h"] = 1
 
+    elif type(agent) is Spice:
+        if agent.amount != 0:
+            portrayal["Color"] = spice_dic[agent.amount]
+        else:
+            portrayal["Color"] = "#D6F5D6"
```
rht, line 36: What color is this?

.pre-commit-config.yaml
tpike3: sorry about that, I keep getting a pre-commit error and was trying some fixes, forgot to undo that one.

examples/sugarscape_g1mt/sugarscape_g1mt/model.py
rht: Ah, I see. It's a symmetric matrix, and so transposing won't change it.

examples/sugarscape_g1mt/sugarscape_g1mt/server.py
rht: I just checked; it's a nice green color. However, both sugar and spice seem to have the same color.

tpike3: I made spice a "wheat" color just for test purposes. Are you saying you want them to have the same color or they do have the same color? It should look like....

image

My intent is to keep them different colors until I code the larger color (i.e. more sugar or more spice) taking precedence. .

rht: Oh, I wasn't aware that the color is different. Yes, definitely different color.

.pre-commit-config.yaml
tpike3: Fixed via #1382

examples/sugarscape_g1mt/sugarscape_g1mt/agents.py
```diff
@@ -266,8 +266,8 @@ def trade(self, other):
 
         # Our extra sanity check. The agent must still have enough resource
         # before the trade.
-        assert self.sugar > 0
-        assert self.spice > 0
+        # assert self.sugar > 0
```
rht, line 269: This constraints are important. Otherwise they can become negative.

examples/sugarscape_g1mt/sugarscape_g1mt/agents.py
```diff
@@ -340,3 +340,9 @@ def calculate_MRS(self, sugar=None, spice=None):
             spice = self.spice
         # See GAS page 102 equation 3.
         return (spice / self.metabolism_spice) / (sugar / self.metabolism_sugar)
+
+    def step(self):
```
rht, line 344: This changes the simulation implementation complete. Each agent now does all the following steps before the next again. This may cause first-move-advantage problem.

examples/sugarscape_g1mt/sugarscape_g1mt/agents.py
tpike3: I am still working on this one it only seems to be throwing an error with the server but very irregularly.

examples/sugarscape_g1mt/sugarscape_g1mt/agents.py
tpike3: Yes, I plan to get back to each part of the process fix but first had to figure out why the time wasn't advancing, it has to call through `schedule.step` if they want the`mesa.time` code to be implemented.

tpike3: A challenge is the model is really set up as stagedactivationbytype...which we do not have.

rht: We can just manually shuffle the agents, and advance the time inside the model step. Simplest solution for now.

examples/sugarscape_g1mt/sugarscape_g1mt/agents.py
```diff
@@ -53,16 +55,17 @@ def step(self):
 class Trader(mesa.Agent):
     def __init__(
         self,
-        pos,
+        id,
```
rht, line 58: Needs to be `unique_id` for consistency.

examples/sugarscape_g1mt/sugarscape_g1mt/agents.py
```diff
@@ -266,18 +266,20 @@ def trade(self, other):
...
+        """
         if other.sugar <= 0 or other.spice <= 0:
             return
-
+        """
         # Agent and neighbor compute their MRSs; if these are equal then end,
         # else continue
         mrs_self = self.calculate_MRS()
         welfare_self = self.calculate_welfare()
+        welfare_self = self.calculate_welfare()
```
rht, line 282: Duplicate line.

examples/sugarscape_g1mt/sugarscape_g1mt/agents.py
```diff
@@ -303,13 +305,15 @@ def trade(self, other):
             sold = self.maybe_sell_spice(other, price, welfare_self, welfare_other)
             if not sold:
                 return
+            self.prices.append(price)
```
rht, line 308: Make sense.

examples/sugarscape_g1mt/sugarscape_g1mt/agents.py
```diff
@@ -303,13 +305,15 @@ def trade(self, other):
...
         else:
             # self is a spice buyer
             sold = other.maybe_sell_spice(self, price, welfare_other, welfare_self)
             if not sold:
                 return
+            self.prices.append(price)
         # continue trading
-        self.trade(other)
+        # self.trade(other)
```
rht, line 316: Why?

examples/sugarscape_g1mt/sugarscape_g1mt/agents.py
```diff
@@ -303,13 +305,15 @@ def trade(self, other):
             sold = self.maybe_sell_spice(other, price, welfare_self, welfare_other)
             if not sold:
                 return
+            self.prices.append(price)
         else:
             # self is a spice buyer
             sold = other.maybe_sell_spice(self, price, welfare_other, welfare_self)
             if not sold:
                 return
+            self.prices.append(price)
```
rht, line 314: You should just de-dent this line, and remove the previous `self.prices.append`, because this append happens in all cases regardless of the if-else situation.

examples/sugarscape_g1mt/sugarscape_g1mt/agents.py
tpike3: I don't know why I commented this out, I have since uncommented.

examples/sugarscape_g1mt/sugarscape_g1mt/agents.py
tpike3: Will do

examples/sugarscape_g1mt/sugarscape_g1mt/agents.py
```diff
@@ -144,9 +145,7 @@ def move(self):
         ]
         max_welfare = max(welfares)
         candidate_indices = [
-            i
-            for i in range(len(welfares))
-            if math.isclose(welfares[i], max_welfare, rel_tol=1e-02)
+            i for i in range(len(welfares)) if math.isclose(welfares[i], max_welfare)
```
rht, line 148: Why are the `rel_tol`'s removed?

examples/sugarscape_g1mt/sugarscape_g1mt/agents.py
```diff
@@ -168,13 +167,13 @@ def eat(self):
         # Sometimes the position has absolutely no sugar content, and so we
         # check for its existence.
         if sugar_patch:
-            self.sugar = self.sugar - self.metabolism_sugar + sugar_patch.amount
+            self.sugar = (self.sugar + sugar_patch.amount) - self.metabolism_sugar
```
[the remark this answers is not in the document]
rht: My hunch is that the removal of the `rel_tol` is the cause. While this is a red herring.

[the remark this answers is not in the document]
tpike3: I would think so to, I would think they will produce the same results,  but I just ran them both in that function and printed the results. 
removed line: 
sugar: 52
spice: 41

changed line: 
sugar:  49 
spice: 43

At the macro level the spread is different changed line has better convergence to 1 while the old line has greater spread.

examples/sugarscape_g1mt/sugarscape_g1mt/agents.py
tpike3: 1e-09 is the default and changing it had minimal impact if any, so it just makes it shorter.

examples/sugarscape_g1mt/sugarscape_g1mt/agents.py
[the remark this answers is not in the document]
rht: I tried running the latest version of this PR 3 times. Each time I get different plots, and different final number of agents. Is it non-deterministic only on my machine?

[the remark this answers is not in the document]
rht: I managed to make my simulation deterministic by doing this in run.py
```python
model = SugarscapeG1mt(seed=96)
model.run_model()
```
and I found that both the removed and changed line have identical results.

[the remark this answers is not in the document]
tpike3: Thanks goodness, it was a red herring! I was baffled by that and made it worse with a bad test code. (These two versions colab tutorial and the example are killing me) 

Regardless, reverted back to the original code and am getting the right results.  I also synced the colab code, the main body of the code should be the same now except for the comments.

examples/sugarscape_g1mt/sugarscape_g1mt/server.py
jackiekazil: @tpike3 that looks good to me.

examples/sugarscape_g1mt/sugarscape_g1mt/model.py
```diff
@@ -0,0 +1,224 @@
...
+                self.schedule.add(spice)
+                agent_id += 1
+
+        # Create agent:
+        for i in range(self.initial_population):
+            x = self.random.randrange(self.width)
+            y = self.random.randrange(self.height)
+            # See GAS page 108 for parameters initialization.
+            # Each agent is endowed by a random amount of sugar and spice
+            sugar = int(self.random.uniform(self.endowment_min, self.endowment_max + 1))
```
rht, line 127: Reminder to remove the `+ 1`.

rht: I just saw in the video that the `int()` excludes the endpoint because it's always rounded down. Maybe it should be `randrange` instead?

rht: It works either way, but with `randrange`, we don't have to explain why we need to add `+ ` even though `random.uniform` includes the endpoint.

examples/sugarscape_g1mt/sugarscape_g1mt/agents.py
```diff
@@ -0,0 +1,334 @@
...
+            return spice_patch.amount
+        return 0
+
+    def get_trader(self, pos):
+        this_cell = self.model.grid.get_cell_list_contents(pos)
+        for agent in this_cell:
+            if isinstance(agent, Trader):
+                return agent
+
+    def is_occupied(self, pos):
```
rht, line 113: @tpike3 what about this explanatory comment? With this, we can clearly separate the condition. What do you think of renaming `is_occupied` to `is_occupied_by_other`?
```python
if pos == self.pos:
    # We consider the agent's own position to be "unoccupied" in the sense that it is one of the legal positions the agent can occupy.
    return False
this_cell = self.model.grid.get_cell_list_contents(pos)
for a in this_cell:
    if isinstance(a, Trader):
        # Occupied by other agent.
        return True
return False
```

examples/sugarscape_g1mt/sugarscape_g1mt/model.py
tpike3: Unless you object I am ok with living it in because I think it is a good teaching point on the importance of documentation and understanding how functions work. 

The more I do this, the more I viewing it as an advanced Python tutorial as well, that addresses thing I learned over years.

Let me know what you think.

rht: I pick `randrange`, because in addition to having to explain the endpoint, you also have to explain why `sugar` and `spice` are initialized differently from `x`, and `y`, even though both are `int` -- based on Occam's razor. The teaching point about documentation can be done in another section of the code. This way, we have shorter video / have extra time for other stuff.

examples/sugarscape_g1mt/sugarscape_g1mt/agents.py
tpike3: That works

examples/sugarscape_g1mt/sugarscape_g1mt/agents.py
```diff
@@ -0,0 +1,334 @@
...
+        candidates = [neighbors[i] for i in candidate_indices]
+
+        # 3. Find the nearest patch among the candidate.
+        min_dist = min([get_distance(self.pos, pos) for pos in candidates])
+        final_candidates = [
+            pos
+            for pos in candidates
+            if math.isclose(get_distance(self.pos, pos), min_dist)
+        ]
+        self.random.shuffle(final_candidates)
```
rht, line 159: Reminder: should be clarified as
```python
final_candidate = self.random.choice(final_candidates)
self.model.grid.move_agent(self, final_candidate)
```

examples/sugarscape_g1mt/sugarscape_g1mt/model.py
```diff
@@ -0,0 +1,224 @@
...
+            self.grid.place_agent(trader, (x, y))
+            self.schedule.add(trader)
+            agent_id += 1
+        self.running = True
+
+    def randomize_traders(self):
+        """
+        Helper function for step to change agent order to prevent mover positional advantage
+        """
+        Traders = self.schedule.agents_by_type[Trader].values()
```
rht, line 157: Variable names should be in lowercase, `traders`, whenever possible.

rht: PEP8: https://peps.python.org/pep-0008/#function-and-variable-names

examples/sugarscape_g1mt/sugarscape_g1mt/model.py
```diff
@@ -0,0 +1,224 @@
...
+        return sum(x) / len(x)
+    return 0.0
+
+
+def flatten(list_of_lists):
+    return [item for sublist in list_of_lists for item in sublist]
+
+
+def get_trade(agent):
+    if type(agent) == Trader:
```
rht, line 26: Needs to be `isinstance(agent, Trader)`.

examples/sugarscape_g1mt/sugarscape_g1mt/model.py
```diff
@@ -0,0 +1,224 @@
...
+    if len(x) > 0:
+        return sum(x) / len(x)
+    return 0.0
+
+
+def flatten(list_of_lists):
+    return [item for sublist in list_of_lists for item in sublist]
+
+
+def get_trade(agent):
```
rht, line 25: `get_trade_partners` is clearer.

rht:
Superseded by https://github.com/projectmesa/mesa-examples/pull/18.

PR #1293, opened 2022-04-27
fix: Make bank_reserves batch_run example work

rht:
The example of both batch_run and legacy BatchRunner is not working. This PR is the fix.

### DIFF
--- a/examples/bank_reserves/batch_run.py
+++ b/examples/bank_reserves/batch_run.py
@@ -190,8 +190,6 @@ if __name__ == "__main__":
     data = batch_run(
         BankReservesModel,
         br_params,
-        model_reporters={"Rich": get_num_rich_agents},
-        agent_reporters={"Wealth": "wealth"},
     )
     br_df = pd.DataFrame(data)
     br_df.to_csv("BankReservesModel_Data.csv")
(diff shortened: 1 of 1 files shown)


examples/bank_reserves/batch_run.py
```diff
@@ -190,23 +190,27 @@ def run_model(self):
...
     # you want to migrate your code to use `batch_run()` from `BatchRunner`.
+    # Things to note:
+    # - You have to set "reserve_percent" in br_params to `[5]`, because the
+    #   legacy BatchRunner doesn't auto-detect that it is single-valued.
+    # - The model reporters need to be explicitly specified in the legacy
+    #   BatchRunner
     """
-    br = BatchRunner(
+    from mesa.batchrunner import BatchRunnerMP
+    br = BatchRunnerMP(
```
tpike3, line 207: Should this be converted to batch_run since we are deprecating BatchRunnerMP?

rht: The code here is intentionally left behind, so that people can compare how it is done via `batch_run` and `BatchRunnerMP`.

tpike3: Ahhh right... that is what I get for rushing and also trying to work on the script.

tpike3 (approved):
LGTM

PR #1294, opened 2022-04-28
Implement simpler Mesa namespace

### LINKED ISSUE #1229: UX: flat Mesa namespace alternative

rht:
This fixes #1229. This is a different solution from #1244. To see how it works, you can read the test_import_namespace.py in this PR.

### DIFF
--- a/mesa/__init__.py
+++ b/mesa/__init__.py
@@ -9,8 +9,12 @@ import datetime
 from mesa.model import Model
 from mesa.agent import Agent
 
+import mesa.time as time
+import mesa.space as space
+import mesa.flat.visualization as visualization
+from mesa.datacollection import DataCollector
 
-__all__ = ["Model", "Agent"]
+__all__ = ["Model", "Agent", "time", "space", "visualization", "DataCollector"]
 
 __title__ = "mesa"
 __version__ = "0.9.0"
--- a/.coveragerc
+++ b/.coveragerc
@@ -5,3 +5,4 @@ branch = True
 [report]
 omit =
     tests/*
+    mesa/flat/*
--- a/examples/wolf_sheep/wolf_sheep/agents.py
+++ b/examples/wolf_sheep/wolf_sheep/agents.py
@@ -1,4 +1,4 @@
-from mesa import Agent
+import mesa
 from wolf_sheep.random_walk import RandomWalker
 
 
@@ -92,7 +92,7 @@ class Wolf(RandomWalker):
                 self.model.schedule.add(cub)
 
 
-class GrassPatch(Agent):
+class GrassPatch(mesa.Agent):
     """
     A patch of grass that grows at a fixed rate and it is eaten by sheep
     """
--- a/examples/wolf_sheep/wolf_sheep/model.py
+++ b/examples/wolf_sheep/wolf_sheep/model.py
@@ -9,15 +9,13 @@ Replication of the model found in NetLogo:
     Northwestern University, Evanston, IL.
 """
 
-from mesa import Model
-from mesa.space import MultiGrid
-from mesa.datacollection import DataCollector
+import mesa
 
 from wolf_sheep.scheduler import RandomActivationByTypeFiltered
 from wolf_sheep.agents import Sheep, Wolf, GrassPatch
 
 
-class WolfSheep(Model):
+class WolfSheep(mesa.Model):
     """
     Wolf-Sheep Predation Model
     """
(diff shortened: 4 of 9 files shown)


mesa/__init__.py
```diff
@@ -9,8 +9,12 @@
 from mesa.model import Model
 from mesa.agent import Agent
 
+import mesa.time as time
+import mesa.space as space
+import mesa.visualization_flat as visualization
+from mesa.datacollection import DataCollector
```
rht, line 15: This is the line in this PR that needs discussion. I decided to explicitly import the single class from `mesa.datacollection` because a wildcard import will confuse `flake8`. Should we define a new class in `mesa.datacollection` in the future, we should be able to quickly detect that we need to explicitly include it in this line.

mesa/__init__.py
```diff
@@ -9,8 +9,12 @@
 from mesa.model import Model
 from mesa.agent import Agent
 
+import mesa.time as time
```
rht, line 12: Actually, here is another point that requires a consensus.

Should we do
```
import mesa
...
mesa.time.RandomActivation
# or
mesa.RandomActivation
```
?

rht: I chose `mesa.time.RandomActivation`, because I want it to be consistent that other groupings, such as `mesa.visualization.Slider`, require a namespace prefix. To the reader of the code, `mesa.visualization.Slider` gives better context of what it is, more than `mesa.Slider`.

rht: The trade-off here is that `mesa.Slider` is comparatively easier to type for quick prototyping in a Jupyter notebook.

And so, I think we can have the namespace hierarchy in this PR as the default, for the general use case, and still have #1244 with flat namespace (no `time`, `space`, `visualization` grouping; everything is flat) for the quick prototyping use case.

tpike3:
This is so simple, yet a very perceptual change and I keep going back and forth. This week I will write up a quick overview of the options you outlined here and post as the #1 agenda item for the next dev session. 

Then I will work to get feedback via two channels: 
1- Contributors can vote even if they can't make the next dev session
2-Send to user base (e,g, TU Delft, U. Mary Washington) to see what they think (particularly as they have brought this up) then show with the group

Critically I just want to make sure the contributors has  some time and data to think about it and don't have to make a rash vote in the dev sessions. 

Does this plan make sense?

tpike3:
@rht can you give me a common sense check on what I am sending out to key user groups......

*We are planning on flattening Mesa's name space to make it easier. We have narrowed it down to two alternatives. Could you please let us know what you and your students would prefer and why. The debate is whether our time.py file and visualization.py file need context. This is backwards compatible so the current approach will continue to work.* 

```
#current approach 4 examples
mesa.datacollection.DataCollector
mesa.visualization.ModularVisualization.ModularServer
mesa.visualization.UserParam.Slider
mesa.time.RandomActivation

#Option 1: As flat as possible  
mesa.Datacollector
mesa.ModularServer
mesa.Randomactivation
mesa.Slider  

#Option 2: With context for time.py and visualization
mesa.DataCollector
mesa.visualization.ModularServer
mesa.time.RandomActivation
mesa.visualization.Slider

```

rht:
> The debate is whether our time.py file and visualization.py file need context.

And also space.py

The options are:
1. non-breaking: have 2 options simultaneously: 1. with context, e.g. `mesa.time.RandomActivation`, AND 2. flat, but under `mesa.flat` namespace, e.g. `mf.RandomActivation`
2. non-breaking: implements only with context, i.e. `mesa.time.RandomActivation`
3. non-breaking: implements only the `mesa.flat`
4, non-breaking: flatten everything to be directly in `mesa`

Note that the 4 options still let you import using current namespace hierarchies, and hence why they are non-breaking.
(I think it is a matter of choosing option 1 vs 4, maybe we should just like 1 and 4?)

If the `mesa.flat` is pursued, we need to vote on its names:
```
import mesa.all as ma
import mesa.flat as mf
import mesa.simple as ms
import mesa.lite as ml
```

Corvince:
For me Option 1) and 4) sound like good alternatives, options 2) and 3) seem a bit half-baked.

I personally prefer Option 4), but another name suggestion for 1) could be ´mesa.interactive as mia´ to highlight its suitability for interactive prototyping.

tpike3:
We got our first feedback from TU Delft: 

They like mesa (so no mesa.flat etc)  with context. Their reasoning....

*It both delineates and simplifies. It makes it explicit that you are picking a space, or time or interface, but does not overburden the user with in between layers and extra abstractions.*

U. of Mary Washington will give us feedback Friday.

rht:
This PR is ready. I implemented both the simplified with-context and mesa.flat.

tpike3 (approved):
Thanks rht!

LGTM

PR #1300, opened 2022-05-01
Default batch_run to 1 CPU

### LINKED ISSUE #1210: batch_run should default to running on 1 CPU

rht:
This fixes #1210.
This is a BREAKING CHANGE.
This simplifies the introductory tutorial.

I didn't implement the warning on Windows when using `multiprocessing`, because I want to keep this PR small, and I think Python already does the warning by default?

### DIFF
--- a/docs/tutorials/intro_tutorial.ipynb
+++ b/docs/tutorials/intro_tutorial.ipynb
@@ -840,7 +840,7 @@
     "\n",
     "* `number_processes`\n",
     "<br/><br/>\n",
-    "Number of processors used to run the sweep in parallel. Optional. If not specified, defaults to use all the available processors.\n",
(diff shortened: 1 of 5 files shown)


tpike3 (approved):
Very cool you did a snippet...

PR #1308, opened 2022-05-12
visualization: Specify tooltip without jQuery

rht:
There might be a race condition where the tooltip DOM object doesn't
exist yet when the $(tooltipID).tooltip() is called.


tpike3 (approved):
LGTM

Issue #1309, opened 2022-05-12
batch_run module double multiplies iteration when running


clumdee:
**Describe the bug**
Module `batch_run` module encounters `iterations` twice.

**Expected behavior**
If `iterations=1` gets `2it`, then `iterations=2` gets `4it`, and `iterations=3` gets `6it`. 

**To Reproduce**
Simplified version of [tutorial](https://mesa.readthedocs.io/en/latest/tutorials/intro_tutorial.html#batch-run).
Set `iterations=1` gets `2it`, set `iterations=2` gets `8it`, set `iterations=3` gets `18it`. 

```python
params = {"width": 10, "height": 10, "N": [1, 2]}

results = batch_run(
    MoneyModel,
    parameters=params,
    iterations=1,
    max_steps=1,
    number_processes=1,
    data_collection_period=-1,
    display_progress=True,
)
```

**Additional context**
`iterations` here should be removed
https://github.com/projectmesa/mesa/blob/006359b2a7e9a3b550b790c1b16c5a089521bb27/mesa/batchrunner.py#L70

rht:
Did you meant this line? https://github.com/projectmesa/mesa/blob/006359b2a7e9a3b550b790c1b16c5a089521bb27/mesa/batchrunner.py#L78=

rht:
Yeah, appreciate a PR.
It needs to be `iterations` removed in L78 instead, because the multi-process version of the loop in L98-110 uses `kwargs_list` without the nested loop of `for iteration in range(iteration)` and `for kwargs in kwargs_list`.

clumdee:
I see. Then we need to find a way to properly loop in L85-95. Because leaving L70 as is will keep the single-process loop to continue double counting `iterations`.

I can't even get the multi-process to run on my machine. Not sure if it is related to this issue.

rht:
You can leave the multiprocess version intact, and simply remove the `for iteration in range(iterations)` in the single process version, because the `kwargs_list` variable has already been multiplied by `iterations`. If you are on Windows, you need to do this extra step: https://github.com/projectmesa/mesa/blob/main/docs/useful-snippets/snippets.rst#using-multi-process-batch_run-on-windows= (this is also mentioned in the tutorial in case you missed it)

clumdee:
Thanks for the suggestion. 
However removing `for iteration in range(iterations)` will affect L91. 

Hence, we need to find another way to track iteration logging. 
Maybe change L86 to enum?

[...]

PR #1310, opened 2022-05-12
fix double multiply of iterations in singleprocess

clumdee:
Fixing this issue https://github.com/projectmesa/mesa/issues/1309

### DIFF
--- a/mesa/batchrunner.py
+++ b/mesa/batchrunner.py
@@ -75,39 +75,34 @@ def batch_run(
         data_collection_period=data_collection_period,
     )
 
-    total_iterations = len(kwargs_list) * iterations
+    total_iterations = len(kwargs_list)
     run_counter = count()
 
     results: List[Dict[str, Any]] = []
 
     with tqdm(total_iterations, disable=not display_progress) as pbar:
-        if number_processes == 1:
-            for iteration in range(iterations):
-                for kwargs in kwargs_list:
-                    _, rawdata = process_func(kwargs)
-                    run_id = next(run_counter)
-                    data = []
-                    for run_data in rawdata:
-                        out = {"RunId": run_id, "iteration": iteration - 1}
-                        out.update(run_data)
-                        data.append(out)
-                    results.extend(data)
-                    pbar.update()
+        iteration_counter: Counter[Tuple[Any, ...]] = Counter()
 
+        def _fn(paramValues, rawdata):
+            iteration_counter[paramValues] += 1
+            iteration = iteration_counter[paramValues]
+            run_id = next(run_counter)
+            data = []
+            for run_data in rawdata:
+                out = {"RunId": run_id, "iteration": iteration - 1}
+                out.update(run_data)
+                data.append(out)
+            results.extend(data)
+            pbar.update()
+
+        if number_processes == 1:
+            for kwargs in kwargs_list:
+                paramValues, rawdata = process_func(kwargs)
+                _fn(paramValues, rawdata)
         else:
-            iteration_counter: Counter[Tuple[Any, ...]] = Counter()
             with Pool(number_processes) as p:
                 for paramValues, rawdata in p.imap_unordered(process_func, kwargs_list):
-                    iteration_counter[paramValues] += 1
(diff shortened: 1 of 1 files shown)


mesa/batchrunner.py
```diff
@@ -75,24 +75,23 @@ def batch_run(
...
-                        out.update(run_data)
-                        data.append(out)
-                    results.extend(data)
-                    pbar.update()
+            for iteration, kwargs in enumerate(kwargs_list):
+                _, rawdata = process_func(kwargs)
+                run_id = next(run_counter)
+                data = []
+                for run_data in rawdata:
+                    out = {"RunId": run_id, "iteration": iteration - 1}
```
rht, line 90: Btw, this is another bug that already exists in the original code: since `iteration` starts from 0, `iteration - 1` starts with -1 instead, which is wrong. We can fix it in another PR.

mesa/batchrunner.py
```diff
@@ -75,24 +75,23 @@ def batch_run(
...
-                    _, rawdata = process_func(kwargs)
-                    run_id = next(run_counter)
-                    data = []
-                    for run_data in rawdata:
-                        out = {"RunId": run_id, "iteration": iteration - 1}
-                        out.update(run_data)
-                        data.append(out)
-                    results.extend(data)
-                    pbar.update()
+            for iteration, kwargs in enumerate(kwargs_list):
```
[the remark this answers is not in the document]
rht: Hmm, wait, this is strange. If the length of `_make_model_kwargs(parameters)` is 2, and `iterations` is 2, this `iteration` goes to 4, which is wrong. It should have only gone up to 2. Essentially it is the same as `run_id`. If you can confirm this?

rht:
Your latest version of the PR seems to work, but this also works, while at the same time it deduplicates the single-process and multiprocess code:
```python
    with tqdm(total_iterations, disable=not display_progress) as pbar:
        iteration_counter: Counter[Tuple[Any, ...]] = Counter()

        def _fn(paramValues, rawdata):
            iteration_counter[paramValues] += 1
            iteration = iteration_counter[paramValues]
            run_id = next(run_counter)
            data = []
            for run_data in rawdata:
                out = {"RunId": run_id, "iteration": iteration - 1}
                out.update(run_data)
                data.append(out)
            results.extend(data)
            pbar.update()

        if number_processes == 1:
            for kwargs in kwargs_list:
                paramValues, rawdata = process_func(kwargs)
                _fn(paramValues, rawdata)
        else:
            with Pool(number_processes) as p:
                for paramValues, rawdata in p.imap_unordered(process_func, kwargs_list):
                    _fn(paramValues, rawdata)

```

clumdee:
Your suggestion is much nicer. I made change accordingly.

mesa/batchrunner.py
```diff
@@ -75,39 +75,34 @@ def batch_run(
...
-                    pbar.update()
+        iteration_counter: Counter[Tuple[Any, ...]] = Counter()
+
+        def _fn(paramValues, rawdata):
+            iteration_counter[paramValues] += 1
+            iteration = iteration_counter[paramValues]
+            run_id = next(run_counter)
+            data = []
+            for run_data in rawdata:
+                out = {"RunId": run_id, "iteration": iteration - 1}
```
rht, line 92: This needs to be without the `- 1`.

rht:
2 last points:
- just need to make sure that you observe the speedup in the multi-process version?
- need to squash the commits to 1

clumdee:
> 2 last points:
> 
> * just need to make sure that you observe the speedup in the multi-process version?
> * need to squash the commits to 1

- Comparing to with single-process, yes.
Trying the below code gives single-process ~15it/s and multi-process ~60it/s.

- I made a squashed commit.
```python
params = {"width": 10, "height": 10, "N": [100, 200, 300]}
iterations = 100
max_steps = 100


results = batch_run(
    MoneyModel,
    parameters=params,
    iterations=iterations,
    max_steps=max_steps,
    number_processes=1,
    #number_processes=None,
    data_collection_period=-1,
    display_progress=True,
)
```

rht:
LGTM

PR #1312, opened 2022-05-13
ci: Ensure wheel is install before `pip install` step

rht:
Without this PR's fix, you see this message:
https://github.com/projectmesa/mesa/runs/6422630638?check_suite_focus=true#step%3A5%3A161=.


jackiekazil:
LGTM

PR #1313, opened 2022-05-14
fix: sugarscape_cg: Use better way to check if a cell is occupied by SsAgent

rht:
The original method relies on an assumption that a cell always contains
a sugar patch.
We shouldn't assume. Also, we are probably going to optimize the code by
removing the sugar patches that have values that are always 0.


Corvince:
Good change, but I find your code a bit difficult to parse. Would `return any(isinstance(agent, SsAgent) for agent in this_cell)` also work?

rht:
That's clearer. I have changed it.

rht:
@Corvince I just realized that you did a merge operation instead of rebase and merge. The latter has a cleaner Git log, and is preferred.

PR #1319, opened 2022-05-15
ChartModule.js: Use more semantic HTML element creation method

rht:
In this PR, I removed jQuery usage, and I use a more semantic way to create the HTML elements instead of strings.


tpike3 (approved):
LGTM

PR #1323, opened 2022-05-17
examples: Use nicer color for chart

rht:
Before 
2022-05-17-052427_877x566_scrot
After
2022-05-17-052534_888x568_scrot
The latter is taken from Matplotlib tab10 color palette.

### DIFF
--- a/examples/charts/charts/server.py
+++ b/examples/charts/charts/server.py
@@ -17,12 +17,13 @@ Accessed on: November 2, 2017
 Author of original code: Taylor Mutch
 """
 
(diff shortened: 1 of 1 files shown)


rht:
There are no lines in the "after" screenshot because I forgot to run the example, but at least you can see the new coloring.

rht:
Here is the screenshot with the lines, just in case:
2022-05-17-053339_874x564_scrot

jackiekazil:
LGTM.

Corvince:
Thank you for this, those colors were really kinda ugly 😅

PR #1325, opened 2022-05-18
frontend: Update Bootstrap 4.6.1 -> 5.1.3

rht:
And also:
- remove bootstrap-switch dependency because Bootstrap already has
  switch
- Bootstrap 5 no longer uses jQuery under the hood

Testing checklist:
- Check that the start/step/reset works
- Check that on mobile view, the start/step/reset is hidden by the hamburger button
- Check the frame per second input works
- Check the modal (the about button) fades in/out correctly
- Check the user input title shows up correctly
- Check the user input description hover text appears correctly
- Check all the user inputs:
  * slider (boltzmann_wealth_model)
  * choice (pd_grid)
  * boolean switch (wolf_sheep)
  * raw text (virus_on_network)
  * number (no need to check this, because it's not in the examples, and I already check this myself)

### DIFF
--- a/mesa/visualization/templates/js/runcontrol.js
+++ b/mesa/visualization/templates/js/runcontrol.js
@@ -188,25 +188,30 @@ const initGUI = function (model_params) {
 
   const addBooleanInput = function (param, obj) {
     const domID = param + "_id";
-    sidebar.append(
-      [
-        "<div>",
-        "<p><label for='" +
-          domID +
-          "' class='badge badge-primary'>" +
-          obj.name +
-          "</label></p>",
-        "<input class='model-parameter' id='" + domID + "' type='checkbox'/>",
-        "</div>",
-      ].join("")
-    );
-    $("#" + domID).bootstrapSwitch({
-      state: obj.value,
-      size: "small",
-      onSwitchChange: function (e, state) {
-        onSubmitCallback(param, state);
-      },
+    const _switch = document.createElement("div");
+    _switch.className = "form-check form-switch";
+
+    const label = `
+      <label class="form-check-label badge bg-primary" for="${domID}">
+        ${obj.name}
+      </label>
+    `;
+    _switch.innerHTML += label;
+
+    const input = document.createElement("input");
+    Object.assign(input, {
+      className: "form-check-input model-parameter",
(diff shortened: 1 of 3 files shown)


rht:
I have checked everything in the checklist, but definitely need additional checking by someone else, because this is a huge change.

mesa/visualization/templates/js/runcontrol.js
```diff
@@ -188,25 +188,31 @@ const initGUI = function (model_params) {
...
-    $("#" + domID).bootstrapSwitch({
-      state: obj.value,
-      size: "small",
-      onSwitchChange: function (e, state) {
-        onSubmitCallback(param, state);
-      },
+    const _switch = document.createElement("div");
+    _switch.className = "form-check form-switch";
+
+    const label = `
```
Corvince, line 194: Can you indent this like this? 
```js
    const label = `
      <label class="form-check-label badge bg-primary" for="${domID}">
        ${obj.name}
      </label>
    `;
´´´

File: mesa/visualization/templates/js/runcontrol.js [hunk 2]
```diff
@@ -188,25 +188,31 @@ const initGUI = function (model_params) {
...
+
+    const input = document.createElement("input");
+    Object.assign(input, {
+      className: "form-check-input model-parameter",
+      type: "checkbox",
+      id: domID,
+      checked: obj.value,
     });
+    input.setAttribute("role", "switch");
+    input.addEventListener(
```
Corvince, line 209: This is formatted differently with prettier. Can you run prettier on this file?

Corvince (approved):
Looks very good. I tested the whole list and everything seems to still work as intended.

Just two comments regarding code style, but if you fix those please feel free to merge.

mesa/visualization/templates/js/runcontrol.js
rht: Fixed.

mesa/visualization/templates/js/runcontrol.js
rht: Fixed.

rht:
Thank you for the review! I will merge once the tests pass.

PR #1330, opened 2022-05-18
examples: Convert nodes to list when drawing random sample

### LINKED ISSUE #1329: random.sample with a set is removed in Python 3.11
Using [`random.sample`](https://docs.python.org/3/library/random.html#random.sample) with a set has been depreciated in Python 3.9 and was removed in Python 3.11.

> Changed in version 3.11: The population must be a sequence. Automatic conversion of sets to lists is no longer supported.

Two cases of this function being used with a set have been detected by the CI in Mesa, both in example models:

1. https://github.com/projectmesa/mesa/blob/5135be8759d9fdac551c5c6ed13e9ae5205bce3e/examples/boltzmann_wealth_model_network/boltzmann_wealth_model_network/model.py#L28
2. https://github.com/projectmesa/mesa/blob/5135be8759d9fdac551c5c6ed13e9ae5205bce3e/examples/virus_on_network/virus_on_network/model.py#L84

What would be the most efficient way to get `self.G.nodes()` in a sequence form (so probably list or tuple)? I feel like just putting it inside a `list()` isn't the proper way to do it.

EwoutH:
Using [`random.sample`](https://docs.python.org/3/library/random.html#random.sample) with a set has been depreciated in Python 3.9 and was removed in Python 3.11.

> Changed in version 3.11: The population must be a sequence. Automatic conversion of sets to lists is no longer supported.

This commit fixes two cases in which this happens, by converting the networkx.Graph.nodes set to a list before drawing a random sample from it.

See [networkx.Graph.nodes](https://networkx.org/documentation/stable/reference/classes/generated/networkx.Graph.nodes.html) for the recommended conversion from nodes to a list.

Closes #1329.

### DIFF
--- a/examples/boltzmann_wealth_model_network/boltzmann_wealth_model_network/model.py
+++ b/examples/boltzmann_wealth_model_network/boltzmann_wealth_model_network/model.py
@@ -25,7 +25,7 @@ class BoltzmannWealthModelNetwork(mesa.Model):
             agent_reporters={"Wealth": lambda _: _.wealth},
         )
 
-        list_of_random_nodes = self.random.sample(self.G.nodes(), self.num_agents)
+        list_of_random_nodes = self.random.sample(list(self.G), self.num_agents)
 
         # Create agents
         for i in range(self.num_agents):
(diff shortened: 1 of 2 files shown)


EwoutH:
CI passes, and the depreciation warning is gone, as expected!

rht:
LGTM

rht:
Yeah no worries.

PR #1333, opened 2022-05-19
Convert hex_snowflake, pd_grid, schelling to simple namespace; [BREAKING] Remove class name collision

rht:
- Convert hex_snowflake, pd_grid, schelling to simple namespace
- fix: Import TextVisualization to flat namespace (I somehow missed to import this file)
- BREAKING: Rename mesa.visualizations.modules.TextElement to TextModule

The last one is a breaking change, and it matters because: it collides with `mesa.visualizations.TextVisualization.TextElement`.
I chose to change it to TextModule *now*, because it will get harder to change the name to make it consistent with how the rest of the viz modules are named in the future.

This, together with `batch_run` defaulting to 1 CPU, need to be emphasized in the release note.

### DIFF
--- a/examples/hex_snowflake/hex_snowflake/cell.py
+++ b/examples/hex_snowflake/hex_snowflake/cell.py
@@ -1,7 +1,7 @@
-from mesa import Agent
+import mesa
 
 
-class Cell(Agent):
+class Cell(mesa.Agent):
     """Represents a single ALIVE or DEAD cell in the simulation."""
 
     DEAD = 0
--- a/examples/hex_snowflake/hex_snowflake/model.py
+++ b/examples/hex_snowflake/hex_snowflake/model.py
@@ -1,11 +1,9 @@
-from mesa import Model
-from mesa.time import SimultaneousActivation
-from mesa.space import HexGrid
+import mesa
 
 from hex_snowflake.cell import Cell
 
 
-class HexSnowflake(Model):
+class HexSnowflake(mesa.Model):
     """
     Represents the hex grid of cells. The grid is represented by a 2-dimensional array of cells with adjacency rules specific to hexagons.
     """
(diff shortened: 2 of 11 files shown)


rht:
Expect bug reports from people, with content like:
```
AttributeError: 'HappyElement' object has no attribute 'package_includes'
```

rht:
I don't have strong opinion with the breaking change. I am also fine with making it non-breaking, and keep the name `TextElement` for the module, but rename the less commonly used text visualization class. But I also think that the latter should have been called `TerminalVisualization` or `ASCIIVisualization`, for clarity.

rht:
What about renaming `mesa.visualizations.TextVisualization.TextElement` to `ASCIIElement`, and keep the other as is? It's just that the naming `mesa.visualizations.modules.TextElement` is inconsistent with the rest (e.g. `ChartModule`) -- this is the only downside.

tpike3:
> 

This took me awhile to unpack, so to be clear and make sure I am not misinterpreting this, the structure is as follows....

1. In `mesa.visualization.modules` there is `TextVisualization.py` which calls the js file for text visualization
2. In `mesa.visualization` there is also a `TextVisualization.py` but this is pure python and renders ascii codes over the grid. 

Based on this, I think is should be `ASCIIElement` and honestly I think the whole file ( bullet 2) should be changed from `TextVisualization.py` to `ASCIIVisualization.py`.  The approach is of a different breed than modules that call the js. However, I don't think this is necessary now as this gets to the idea of making python more module and can be more effectively addressed when the visualization repo is created.

rht:
OK, let's rename the ASCII one to `ASCIIVisualization`, and I will undo the breaking change.

rht:
@tpike3 this PR is ready. Note that we still to say there is a breaking change. While `mesa.visualization.TextVisualization.ASCIIElement` is a public class, it is probably not used by people, and so it is a much safer breaking change.

rht:
I have only changed the 2nd last commit (the one with the breaking change) since you last reviewed.

tpike3 (approved):
LGTM... Thanks for doing all those updates and finding that collision. Definitely becomes a case study for a more modular Mesa that you and @Corvince have been discussing.

Issue #1263, opened 2022-04-07
Jupyter interface for the GUI


rht:
It would be great if we can use Jupyter UI widgets and components (instead of Mesa-specific UserSettableParams, ChartModule, etc), where we can have interactive simulation running inside a Jupyter session. With this, we can have easy-to-run executable examples using Binder/Colab. Currently, this is not doable? See also https://github.com/projectmesa/mesa/discussions/1259.

jackiekazil:
100% agree. There were some attempts for Jupyter integration in the past during PyCon US sprints, but they were attempts, but not completely finished & they are old -- so not sure how helpful they are now. However I wanted to dust off these PRs to provide thoughts in how others were thinking about it... but a lot of with Jupyter/Collab has changed since these were created. 

https://github.com/projectmesa/mesa/pull/247
https://github.com/projectmesa/mesa/pull/469

rht:
Progress update: I have simplified the non-straightforward code in #469. My prototype simply needs to be cleaned up.
![2022-05-18-065916_1195x549_scrot](https://user-images.githubusercontent.com/395821/169042043-b4799061-5701-4056-b069-195e1145cdaa.png)

PR #1334, opened 2022-05-19
Add React via esbuild

Corvince:
# What is this?

This is the start of a journey to modernize and enhance the mesa frontend. The basic premise is that I added a lot of cool functionality with my mesa-viz package, but I ended up with rewriting large parts of the server and writing very unexperienced frontend react code. While I guess the package still works I think it is outdated beyond repair. But to salvage some of the functionality (like time traveling model steps and parallel simulation runs) I want to pursue a more gentle approach for updating the stack. 

Over the last couple days I have been experimenting on how to best add react to our frontend stack. Most frameworks rely on hosting there own dev server, but this makes it quite hard to fit to our current ´modular_tempate.html´. That was also the main reason why mesa-viz required a new server and did not make use of tornados templating. 

In this approach I only use esbuild to create a single `main.ts` file that is script-included in the template. Currently that means only a single invisible div is added at the front-end, but this allows us to gradually adopt react code. 

The immediate goals are to move the existing html code and our `runcontrol.js` logic into react. The other modules can stay and should work independently like before. 

The long term goals would be to move everything over to react so that we will only have a single ´js´ file that bundles all our dependencies and js modules together. That means we can have a small package footprint, while at the same time improve the development experience of the front end stack (which really is the point of all this).

While the actual need for react is discussable, it just enables to write more functional, descriptive code. Just as an example instead of writing 

```js
const canvas = document.createElement("canvas");
  Object.assign(canvas, {
    width: canvas_width,
    height: canvas_height,
    style: "border:1px dotted",
  });
```

in react you only write (via JSX)

```js
<canvas width={canvas_width} height={canvas_height} style={{border: "1px dotted"}}></canvas>
```

Things like a ChartJS chart can become concise as `<Chart type='line' data={chartData} />`

Any thoughts and opinions are very welcomed! Especially @rht since you are currently doing a lot of frontend development


mesa/visualization/frontend/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "main": "main.ts",
+  "scripts": {
+    "build": "esbuild src/main.tsx --bundle --outfile=../templates/js/main.js --minify",
+    "dev": "esbuild src/main.tsx --bundle --outfile=../templates/js/main.js --watch"
+  },
+  "author": "Project Mesa Team",
+  "repository": {
+    "url": "https://github.com/projectmesa/mesa"
+  },
+  "license": "MIT",
```
rht, line 14: Needs to be Apache-2.0 for consistency.

rht:
```jsx
<canvas width={canvas_width} height={canvas_height} style={{border: "1px dotted"}}></canvas>
```

Yes, this is more readable. We could approximate it using vanilla JS's backticks, but it's not semantic.

Regarding with using React, I assume that you want to use the SPA functionality of it as much as possible? My only concern with this is that then it becomes harder to run viz on a Jupyter notebook (#1263). We could have 2 separate ("parallel universes") viz libraries for the Tornado-based (or nodejs) server, and the Jupyter. But this may increase the maintenance cost.

Corvince:
> Regarding with using React, I assume that you want to use the SPA functionality of it as much as possible? My only concern with this is that then it becomes harder to run viz on a Jupyter notebook (#1263). We could have 2 separate ("parallel universes") viz libraries for the Tornado-based (~or nodejs~) server, and the Jupyter. But this may increase the maintenance cost.

Interesting point. I think with Jupyter we have have two options:
1) Display the server inline, which I never managed to do reliable. It should be possible, but it always feels a bit alien and doesn't allow close coupling with parameter changes
2) Custom visualization logic like in #1263. In that case we always will have some parallel logic e.g. for the UserSettableParameters. I don't think it matters that much how the "normal" looks like in that case, no?

But if we architecture this with this issue in mind the components based approach of react might actually help us. If we have a visualisation component (i.e. without controls and navbar, etc.), we might be able to display just that component inside a jupyter widget. I haven't evaluated how realistic that is, just putting out ideas here.

rht:
> Display the server inline, which I never managed to do reliable.

This problem is documented in https://github.com/projectmesa/mesa/discussions/1259.

rht:
Authoring a Jupyter widget with React: https://blog.jupyter.org/build-a-jupyter-widget-with-react-and-typescript-d83e07340fa3.

mesa/visualization/frontend/src/initGUI.js
```diff
@@ -5,185 +5,192 @@ export const initGUI = function (model_params, send) {
...
-
-        // Finally render the dropdown and activate choice listeners
-        sidebar.append(template.join(""));
-        choiceIdentifiers.forEach(function (id, idx) {
-            $("#" + id).on("click", function () {
-                const value = obj.choices[idx];
-                $("#" + domID).html(value + " " + span);
-                onSubmitCallback(param, value);
-            });
+      const domID = param + "_id";
```
rht, line 98: The `Choice` implementation has been made much simpler in https://github.com/projectmesa/mesa/commit/7b3b072ed1167666110339f1ec39daeb8a67547e.

Corvince: Huh, thanks. I was actually waiting for this commit to be merged, but I must have made a mistake while updating my files with your changes.
Anyhow I will evolve this pr more to be more idiomatic react code and not the current mixture of declarative and Imperative code. So this is all temporary

rht: It's more of with the change from HTML dropdown to HTML select. The latter is much simpler. Dropdown should be used only when each dropdown items has different event listener.

rht:
Is this PR obsolete, given that Solara is based on React principles?

Corvince:
Not so much because of its react principles, but yes, solara suits mesa much better. This was meant to replace ModularServer, but a solara based approach makes more sense.

PR #1343, opened 2022-05-26
feat: Implement NumberInput UserParam class

### LINKED ISSUE #1235: UX: Split UserSettableParameter into various classes

rht:
Fixes #1235.

### DIFF
--- a/mesa/visualization/UserParam.py
+++ b/mesa/visualization/UserParam.py
@@ -1,4 +1,6 @@
 from numbers import Number
+from warnings import warn
+
 
 NUMBER = "number"
 CHECKBOX = "checkbox"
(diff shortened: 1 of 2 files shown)


rht:
I decided to call it `NumberInput` instead of `Number`, to anticipate for a future name conflict inside the `mesa.visualization` namespace.

tpike3 (approved):
LGTM I think we could debate ad nauseum whether everything should have `Input` for consistency but I think `Number` is a special case per your collision statement and agree this is optimal.

PR #1345, opened 2022-05-26
mypy: Use "|" operator instead of Union/Optional

### LINKED ISSUE #1341: type annotations: Replace Union[a, b] with a | b, Optional[a] with a | None
Looks like it is possible to do it even on Python older than 3.10. See https://github.com/projectmesa/mesa-geo/pull/67#discussion_r881122416=.

Update: from implementing #1345, looks like older Pythons don't support type aliases with `|` yet. For this issue to be completely resolved, we need to wait until 3.10 is the oldest supported version.

rht:
Partially addresses #1341. To completely remove `Union`, sadly, we need to wait until 3.10 is the oldest supported version.

See https://peps.python.org/pep-0604/.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -4,8 +4,12 @@ The agent class for Mesa framework.
 Core Objects: Agent
 
 """
+# Mypy; for the `|` operator purpose
+# Remove this __future__ import once the oldest supported Python is 3.10
+from __future__ import annotations
+
 # mypy
-from typing import Optional, TYPE_CHECKING
+from typing import TYPE_CHECKING
 from random import Random
 
 if TYPE_CHECKING:
@@ -27,7 +31,7 @@ class Agent:
         """
         self.unique_id = unique_id
         self.model = model
-        self.pos: Optional[Position] = None
+        self.pos: Position | None = None
 
     def step(self) -> None:
         """A single step of the agent."""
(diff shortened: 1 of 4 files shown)


rht:
@wang-boyu CI failed here. How did you make it work?

wang-boyu:
I'm guessing it's due to the type aliases `TimeT = float | int`, `GridContent = Agent | None` etc.

From [here](https://mypy.readthedocs.io/en/stable/runtime_troubles.html#future-annotations) it mentions

> Even with the __future__ import, there are some scenarios that could still require string literals or result in errors, typically involving use of forward references or generics in:
   type aliases;
   ...

A similar issue was reported: https://github.com/python/mypy/issues/9610

Corvince:
`TimeT: TypeAlias = float | int` potentially fixes this, from what I read

rht:
> `TimeT: TypeAlias = float | int` potentially fixes this, from what I read

I tried this in the 2nd commit. For Python < 3.10, failed with the following error message:

```
E   ImportError: cannot import name 'TypeAlias' from 'typing' 
```

rht:
> `TimeT: TypeAlias = float | int` potentially fixes this, from what I read

I tried this in the 2nd commit. For Python < 3.10, failed with the following error message:

```
E   ImportError: cannot import name 'TypeAlias' from 'typing' 
```

rht:
I think I will use Union for the type aliases for now. Only 4 lines of code need them. The rest can be safely set to use `|`.

rht:
PR is ready for review.

tpike3 (approved):
LGTM

PR #1346, opened 2022-05-26
feat: Implement auto-conversion of function to TextElement

### LINKED ISSUE #1233: UX: Simplify TextElement usage
What if, in the Schelling example, instead of
```python
class HappyElement(TextElement):
    """
    Display a text count of how many happy agents there are.
    """

    def __init__(self):
        pass

    def render(self, model):
        return "Happy agents: " + str(model.happy)
happy_element = HappyElement()
```

we instead do
```python
def happy_element(model):
    return "Happy agents: " + str(model.happy)
```

In the `ModularServer` class, we detect if a visualization element is a function, if so, we then convert it to a `TextElement` automatically.

rht:
Fixes #1233.
Note: the `TextElement` class definition is moved to be inside ModularVisualization.py, because otherwise, I get circular import.


tpike3 (approved):
LGTM

PR #1349, opened 2022-05-31
Add alignment options to agent portrayals in CanvasGridVisualization

VenomPaco:
Adds `xAlign` and `yAlign` properties to agent portrayals for a `CanvasGrid`. Both default to 0.5 when not specified (in line with the existing implementation). The main use case for this is displaying multiple agents that are in the same grid cell without having to resort to different layers or sizes.

image

Currently, it is only implemented for `circle` and `rect`, and values are clamped between 0.0 and 1.0, meaning that in some cases shapes can overflow past the cell boundaries. If needed, I can make some changes to ensure that this will never happen using the circle radius or the rect dimensions.

### DIFF
--- a/mesa/visualization/modules/CanvasGridVisualization.py
+++ b/mesa/visualization/modules/CanvasGridVisualization.py
@@ -25,9 +25,13 @@ class CanvasGrid(VisualizationElement):
             For Circles:
                 "r": The radius, defined as a fraction of cell size. r=1 will
                      fill the entire cell.
+                "xAlign", "yAlign": Alignment of the circle within the cell.
+                                    Defaults to 0.5 (center).
             For Rectangles:
                 "w", "h": The width and height of the rectangle, which are in
                           fractions of cell width and height.
+                "xAlign", "yAlign": Alignment of the rectangle within the
+                                    cell. Defaults to 0.5 (center).
             For arrowHead:
                 "scale": Proportion scaling as a fraction of cell size.
                 "heading_x": represents x direction unit vector.
--- a/mesa/visualization/templates/js/GridDraw.js
+++ b/mesa/visualization/templates/js/GridDraw.js
@@ -79,10 +79,19 @@ const GridVisualization = function (
       // If the stroke color is not defined, then the first color in the colors array is the stroke color.
       if (!p.stroke_color) p.stroke_color = p.Color[0];
 
+      // Default alignments to 0.5 (center of a cell)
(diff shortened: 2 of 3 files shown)


mesa/visualization/modules/CanvasGridVisualization.py
```diff
@@ -25,9 +25,13 @@ class CanvasGrid(VisualizationElement):
             For Circles:
                 "r": The radius, defined as a fraction of cell size. r=1 will
                      fill the entire cell.
+                "x_align", "y_align": Alignment of the circle within the cell.
+                                      Defaults to 0.5 (center).
             For Rectangles:
                 "w", "h": The width and height of the rectangle, which are in
                           fractions of cell width and height.
+                "x_align", "y_align": Alignment of the rectangle within the
```
rht, line 33: Need to be camel case -- `xAlign`, `yAlign`. I know that some of the JS code use underscores, but we should make it consistent for newly added code.

VenomPaco: Done. Would it also be a good idea to implement the constraints I mentioned to prevent the shapes from going outside the grid cells?

rht: You meant, the clamp shouldn't between 0.0 and 1.0, but instead `r` and `1.0 - r`? Yeah sounds good.

VenomPaco: Done. Tested with both circles and rectangles and it is working as intended.

mesa/visualization/templates/js/GridDraw.js
```diff
@@ -140,15 +149,23 @@ const GridVisualization = function (
...
     radius,
     colors,
     stroke_color,
     fill,
     text,
     text_color
   ) {
-    const cx = (x + 0.5) * cellWidth;
-    const cy = (y + 0.5) * cellHeight;
+    // Prevent circle from being drawn outside cell bounds.
```
rht, line 161: I haven't checked this, but you are probably right. Because of the naming, that makes radius / 2 to be very confusing lol. We can clarify the code in a separate PR.

rht:
This LGTM. Can you squash the first 6 commits, but leaving "Prevent shapes from being drawn outside cell bounds" on its own? And so, 2 commits in total.

rht:
Merged, thank you @VenomPaco !

tpike3:
Awesome thanks @VenomPaco and @rht

Issue #1318, opened 2022-05-15
refactor: Remove dependency on jQuery


rht:
We should replace the `$(...)` with vanilla JS.

rht:
See #1319.

rht:
See https://www.infoq.com/news/2020/08/bootsrap-5-drops-jquery/.

> Whats the motivation behind this?

jQuery is an old tech.

Corvince:
Oh, didn't see #1325, great work! 

Old tech doesn't mean bad, but I think we definitely don't have good use cases for keeping it in and removing it should be straight forward

PR #1351, opened 2022-06-01
refactor: frontend: Remove remaining usage of jQuery

rht:
Partially addresses #1318. The remaining work is to remove the jQuery dependency from setup.py and test that everything is working. But I want to keep this PR small.


tpike3 (approved):
LGTM

PR #1352, opened 2022-06-01
fix: Remove mesa.visualization.Number

rht:
This was an instance of the builtin numbers.Number. If we keep it as is,
people will accidentally use it instead of
mesa.visualization.NumberInput.


tpike3 (approved):
LGTM

PR #1355, opened 2022-06-07
Remove jQuery dependency completely

### LINKED ISSUE #1318: refactor: Remove dependency on jQuery

rht:
Fixes #1318.

I tested on the same checklist in #1325:
- Check that the start/step/reset works
- Check that on mobile view, the start/step/reset is hidden by the hamburger button
- Check the frame per second input works
- Check the modal (the about button) fades in/out correctly
- Check the user input title shows up correctly
- Check the user input description hover text appears correctly
- Check all the user inputs:
  * slider (boltzmann_wealth_model)
  * choice (pd_grid)
  * boolean switch (wolf_sheep)
  * raw text (virus_on_network)
  * number (no need to check this, because it's not in the examples, and I already check this myself)

Additionally, I tested all the modules:
- canvas grid, bar chart, pie chart, chart (using the charts example
- network (using the virus_on_network example)
- hex grid (using the hex_snowflake example


tpike3 (approved):
LGTM

PR #1356, opened 2022-06-07
examples: Remove all usage of jQuery

rht:
Turns out that boid_flockers custom visualization JS still uses jQuery. But I have swept through all the JS in examples. There should be no more jQuery left.


tpike3 (approved):
LGTM

PR #1358, opened 2022-06-07
test2

rht:
Ignore this PR. I am experimenting with speeding up the CI pip install caching.


rht:
The experiment was a great success! I was able to reduce the `pip install` step from 24 s to 4 s (+3 s for the virtualenv initialization, which could be made optional).

jackiekazil:
@rht can we close is this?

Issue #1315, opened 2022-05-15
Docs: Update tutorial


tpike3:
**EDIT**

Documentation does not match new namespace ~~or batch_run default `number_processors` to 1~~ or new datacollector

TODO:
- update documentation to reflect namespace.
- update documentation to reflect new datacollector initialization

rht:
The `batch_run` is already described to default to 1 in the tutorial.

tpike3:
> The `batch_run` is already described to default to 1 in the tutorial.

Yep... just realized that as I was looking to link the snippet. So all that needs to be done is the namespace.

PR #1361, opened 2022-06-11
docs: simplified namespace tutorial update

tpike3:
There are some pretty significant diff files, based on further refining the process of converting the `.ipynb` to `.rst` to make the build files for the html. Most of it is cosmetic, and should be significantly less in the future.  Following this process is necessary to prevent the `.ipynb` from being different than the `.rst`, which had happened with the advanced tutorial.


docs/tutorials/adv_tutorial.rst
```diff
@@ -1,22 +1,27 @@
...
+of the model. In this section we’ll walk through creating a
 visualization using built-in components, and (for advanced users) how to
 create a new visualization element.
 
-First, a quick explanation of how Mesa's interactive visualization
+**Note for Jupyter users: Due to conflicts with the tornado server Mesa
+uses and Jupyter, the interactive browser of your model will load but
+likely not work. This will require you to use run the code from .py
+files. The Mesa development team is working to develop a Jupyter
+compatible interface.**
```
rht, line 22: Need to add a tracking issue for this.

rht: This one #1263.

rht:
Regarding with the png files, I think we should seed the model so that the output is always the same, and so we don't keep adding new png files to the repo for every tutorial update.

rht:
Orthogonal to this PR: I was wondering if we could just drop the .rst tutorial files, and point people to the .ipynb links, because GH renders them just fine. https://github.com/projectmesa/mesa/blob/main/docs/tutorials/adv_tutorial.ipynb doesn't show the figures, but this is probably fixable.

tpike3:
> Regarding with the png files, I think we should seed the model so that the output is always the same, and so we don't keep adding new png files to the repo for every tutorial update.

Usually you don't have too but in this case two of the number changed so I just did them all. I also have to update the .rst so it looks in the file. In retrospect I could have just kept the pictures the same since we have to fix the links anyway.

tpike3:
> Orthogonal to this PR: I was wondering if we could just drop the .rst tutorial files, and point people to the .ipynb links, because GH renders them just fine. https://github.com/projectmesa/mesa/blob/main/docs/tutorials/adv_tutorial.ipynb doesn't show the figures, but this is probably fixable.

I have worked on the assumption that the .rst is necessary for sphinx to render the html pages for the readthedocs. But I have not verified that assumption specifically, however the readthedocs and the .rst had the right code for the visualizations while the jupyter did not, so I suppose that is evidence for the assumption.

Corvince:
> > Orthogonal to this PR: I was wondering if we could just drop the .rst tutorial files, and point people to the .ipynb links, because GH renders them just fine. https://github.com/projectmesa/mesa/blob/main/docs/tutorials/adv_tutorial.ipynb doesn't show the figures, but this is probably fixable.
> 
> I have worked on the assumption that the .rst is necessary for sphinx to render the html pages for the readthedocs. But I have not verified that assumption specifically, however the readthedocs and the .rst had the right code for the visualizations while the jupyter did not, so I suppose that is evidence for the assumption.  

You are basically correct, but there exists a sphinx plugin that renders jupyter Notebooks!
I think I even did a test at some point specifically for mesa and as far as I remember it did work rather easily. Don't remember why I never opened a PR, but it should be doable

tpike3:
> 

Sure enough here is [jupyter and sphinx](https://docs.readthedocs.io/en/stable/guides/jupyter.html). If we get the NUMFOCUS grant it it seems we can get to the point where docs are an interactive model.

rht:
I'm not sure if RTD will let you run a tornado instance. It's most likely that the plugin is rendering the .ipynb statically.

That said, being able to replay simulation result client-side on the browser would be cool.

It's not that we can't run ABM simulations purely client-side. It is definitely doable to build an on-browser JS ABM library and run a model fully client-side. hash.ai likely can let you do this.

tpike3:
Forgot to link to #1315

docs/tutorials/adv_tutorial.rst
tpike3: Done

docs/tutorials/adv_tutorial.ipynb
```diff
@@ -15,6 +15,8 @@
     "\n",
     "So far, we've built a model, run it, and analyzed some output afterwards. However, one of the advantages of agent-based models is that we can often watch them run step by step, potentially spotting unexpected patterns, behaviors or bugs, or developing new intuitions, hypotheses, or insights. Other times, watching a model run can explain it to an unfamiliar audience better than static explanations. Like many ABM frameworks, Mesa allows you to create an interactive visualization of the model. In this section we'll walk through creating a visualization using built-in components, and (for advanced users) how to create a new visualization element.\n",
     "\n",
+    "**Note for Jupyter users: Due to conflicts with the tornado server Mesa uses and Jupyter, the interactive browser of your model will load but likely not work. This will require you to use run the code from .py files. The Mesa development team is working to develop a** [Jupyter compatible interface](https://github.com/projectmesa/mesa/issues/1363>).**\n",
```
rht, line 18: Does the link work with the `>` at the end?

rht: I tried. It didn't work.

rht:
> > Regarding with the png files, I think we should seed the model so that the output is always the same, and so we don't keep adding new png files to the repo for every tutorial update.
> 
> Usually you don't have too but in this case two of the number changed so I just did them all. I also have to update the .rst so it looks in the file. In retrospect I could have just kept the pictures the same since we have to fix the links anyway.

Which 2 numbers changed? Why is it deterministic without specifying the seed?

docs/tutorials/adv_tutorial.ipynb
tpike3: Oh geez, thanks for the catch... that epitomizes the pain of having an  `.rst` and `.ipynb`. Fixed

tpike3:
> > > Regarding with the png files, I think we should seed the model so that the output is always the same, and so we don't keep adding new png files to the repo for every tutorial update.
> > 
> > 
> > Usually you don't have too but in this case two of the number changed so I just did them all. I also have to update the .rst so it looks in the file. In retrospect I could have just kept the pictures the same since we have to fix the links anyway.
> 
> Which 2 numbers changed? Why is it deterministic without specifying the seed?

The output file number changed as we had added some cells (e.g. saving the agents' data to `.csv`). It is just how `.ipynb` creates the `.rst` file

rht:
LGTM

PR #1369, opened 2022-06-22
frontend: ModularServer: Allow user to specify custom package includes

rht:
See the discussion in https://github.com/projectmesa/mesa/discussions/1364#discussioncomment-2993622.
I haven't tested this PR. Figured that @wang-boyu can quickly test this instead of me.

### DIFF
--- a/mesa/visualization/ModularVisualization.py
+++ b/mesa/visualization/ModularVisualization.py
@@ -276,7 +276,12 @@ class ModularServer(tornado.web.Application):
     EXCLUDE_LIST = ("width", "height")
 
     def __init__(
-        self, model_cls, visualization_elements, name="Mesa Model", model_params=None
+        self,
+        model_cls,
+        visualization_elements,
+        name="Mesa Model",
+        model_params=None,
+        package_includes=None,
     ):
         """Create a new visualization server with the given elements."""
         if model_params is None:
@@ -290,12 +295,13 @@ class ModularServer(tornado.web.Application):
         self.local_js_includes = set()
         self.local_css_includes = set()
         self.js_code = []
+        if package_includes is not None:
+            # This is custom JS/CSS files specified by the user.
+            for include_file in package_includes:
+                self._add_include(include_file)
         for element in self.visualization_elements:
             for include_file in element.package_includes:
-                if self._is_stylesheet(include_file):
-                    self.package_css_includes.add(include_file)
-                else:
-                    self.package_js_includes.add(include_file)
+                self._add_include(include_file)
             for include_file in element.local_includes:
                 if self._is_stylesheet(include_file):
                     self.local_css_includes.add(include_file)
@@ -368,6 +374,12 @@ class ModularServer(tornado.web.Application):
     @staticmethod
     def _is_stylesheet(filename):
         return filename.lower().endswith(".css")
+
+    def _add_include(self, include_file):
+        if self._is_stylesheet(include_file):
+            self.package_css_includes.add(include_file)
+        else:
+            self.package_js_includes.add(include_file)
 
     def _auto_convert_fn_to_TextElement(self, x):
         """


wang-boyu:
I just realized that the js/css files are already defined in Mesa-Geo's `MapModule`:

```python
class MapModule(VisualizationElement):
    """A MapModule for Leaflet maps."""

    package_includes = ["external/leaflet.js", "MapModule.js"]
    local_includes = []
```

Thus it might be redundant to have the users specify them again in `ModularServer`:

```python
map_element = MapModule(schelling_draw, [52, 12], 4, 500, 500)

server = ModularServer(
    GeoSchelling,
    [map_element, happy_element, happy_chart],
    "Schelling",
    model_params,
    package_includes=[PATH_TO_LEAFLET_JS/CSS]  # <-- already included in map_element
)
```

Shall we move these "extension/plugin includes" into `VisualizationElement`? For instance:

```python
class VisualizationElement:
    package_includes = []    # relative path to JS/CSS files within Mesa
    local_includes = []      # relative path to users' own JS/CSS files
    extension_includes = []  # absolute path to extension/plugin's JS/CSS files
    js_code = ""
    render_args = {}
```

Then in Mesa-Geo I'll update `MapModule` accordingly.

rht:
If it is already in `MapModule`, you just need to change the `package_includes` there to use absolute path.

rht:
```python
class MapModule(VisualizationElement):
    """A MapModule for Leaflet maps."""

    package_includes = [pathlib.Path(os.path.abspath(os.path.dirname(os.path.dirname(__file__)))) / "templates" / "js / "external/leaflet.js", "MapModule.js"]
```

wang-boyu:
I'm getting `WARNING:tornado.access:404 GET` because `modular_template.html` is looking for `package_includes` under `/static/js/` folder: 

```
    <!-- Script includes go here -->
	{% for file_name in package_js_includes %}
		<script src="/static/js/{{ file_name }}" type="text/javascript"></script>
	{% end %}
	{% for file_name in local_js_includes %}
		<script src="/local/{{ file_name }}" type="text/javascript"></script>
	{% end %}
```

rht:
I see. I wouldn't work with `<script src="/Users/boyu/...">` either because the handler is pegged to either current directory (local includes), or Mesa's visualization folder (package includes). Let's figure out an alternative.

rht:
I made https://github.com/rht/mesa/tree/modularserver_ext. This is guaranteed to eventually work, but requires me to add `extra_js_includes` and `extra_css_includes`, which I haven't implemented. IMO, the API is too overwhelming with 6 include args in total.

wang-boyu:
How about we put `extra_includes` or `extension/plugin_indluces` inside the `VisualizationElement` class too, in the same way as `package_includes` and `local_includes`?

So instead of 

```python
class MapModule(VisualizationElement):
    package_includes = ["external/leaflet.js", "MapModule.js"]
    local_includes = []
```

we can have

```python
class MapModule(VisualizationElement):
    package_includes = []
    local_includes = []
    extra/extension/plug_includes = [ABSOLUTE_PATH_TO_external/leaflet.js,
                                     ABSOLUTE_PATH_TO_MapModule.js]
```

Then it's more of an API for extension developers, and the end users simply uses `MapModule`.

The question is how to make those absolute paths work, or is there a better way?

rht:
I have a cleaner solution. Check out https://github.com/rht/mesa/tree/modularserver_ext again. I basically made it so that the local handler's path can be specified by the custom `VisualizationElement`. local_includes and extra_includes are basically overlapping anyway.

rht:
It's still not fully robust, because if there are multiple local dirs, they are mapped to `/local/(.*)` path. Not sure how tornado does this, but it should be a quick fix. I want you to test that branch first.

rht:
This also solves the problem where a Mesa simulation needs to be run in the correct directory. It can then be run from anywhere.

rht:
> I have a cleaner solution. Check out https://github.com/rht/mesa/tree/modularserver_ext again. I basically made it so that the local handler's path can be specified by the custom `VisualizationElement`. local_includes and extra_includes are basically overlapping anyway.

Bump. @wang-boyu ?

wang-boyu:
Sorry but I haven't got the chance to look into this. Have created an issue just now to remind myself.

Will update here once I try it out.

wang-boyu:
I made some changes but the files were not picked up as expected. In `mesa_geo/visualization/modules/MapVisualization.py`, I changed from

```python
package_includes = ["external/leaflet.js", "external/leaflet.css", "MapModule.js"]
local_includes = []
```

to

```python
local_includes = []
local_dir = [
        pathlib.Path(
            f"{os.path.abspath(os.path.dirname(os.path.dirname(__file__)))}/../templates/js/MapModule.js"
        ),
        pathlib.Path(
            f"{os.path.abspath(os.path.dirname(os.path.dirname(__file__)))}/../templates/js/external/leaflet.js"
        ),
        pathlib.Path(
            f"{os.path.abspath(os.path.dirname(os.path.dirname(__file__)))}/../templates/css/external/leaflet.css"
        ),
    ]
```

Is this how `local_dir` supposed to be defined? Did I just mess it up?

rht:
I revisited my branch again. I'm not sure what I'm doing is allowed by Tornado. I'm mapping multiple local directories to 1 URL pattern `r"/local/(.*)"`. There should be only 1 local handler.

As per
```python
        static_handler = (
            r"/static/(.*)",
            tornado.web.StaticFileHandler,
            {"path": os.path.dirname(__file__) + "/templates"},
        )
```
you don't need to specify the files, you just need to specify the absolute path of the custom templates folder.

wang-boyu:
I changed the following and now it's working!

In file `mesa_geo/visualization/modules/MapVisualization.py`:

from

```python
package_includes = ["external/leaflet.js", "external/leaflet.css", "MapModule.js"]
local_includes = []
```

to

```python
local_includes = [
    "js/MapModule.js",
    "css/external/leaflet.css",
    "js/external/leaflet.js",
]
local_dir = os.path.dirname(__file__) + "/../templates"
```

rht:
Let me turn it into a PR then.

PR #1376, opened 2022-07-06
Perf: ContinuousSpace: speed-up add/remove agents

### LINKED ISSUE #1375: Adding many agents to a ContinuousSpace is slow
Dear *,

I am working on speeding up a [program that uses Mesa](https://github.com/ODISSEI-School-Choice/school-choice/).
One of the bottlenecks i found, was that adding lots of agents (60K+) to a ContinuousSpace is very slow.

https://github.com/projectmesa/mesa/blob/60bc768f5eede2dc4806684ae4b0b3046eb07d27/mesa/space.py#L796

Mesa is keeping track of agents on the grid using a numpy.array() ; however, when adding an agent numpy.array.append() has to reallocate a new array, and copy over the old one (see the note [here](https://numpy.org/doc/stable/reference/generated/numpy.append.html#numpy.append) ). This can lead to serious slow downs.

Using regular python arrays would already be orders of magnitude faster. See for instance [this stack overflow question](https://stackoverflow.com/questions/7133885/fastest-way-to-grow-a-numpy-numeric-array) where growing a numpy.array takes ~30 seconds, and growing a python list take 0.3 seconds.

If a lookup of agents is a performance bottleneck, an option would be to convert the python list to a numpy array when all the agents are added; or when the looking up an agent.

A quick fix to replace the numpy array with a list is [here](https://github.com/ODISSEI-School-Choice/school-choice/tree/jisk-v2/mesa-fixes) (based on mesa 0.9.0 from pip)

jiskattema:
Adding / removing many (100+K) Agents to a ContinuousSpace is slow.
This is due to ```numpy.array.append()``` not being optimized for appending.

This pull request changes the internal bookkeeping of the ```_agents_points``` numpy array, and the ```_agent_to_index``` and ```_index_to_agent``` dictionaries.
* re-use, or calculate the ```_agents_points``` array and the 2 dictionaries when needed (neighbor lookups)
* update the ```_agent_points``` array when moving an agent.
* drop the ```_agent_points``` array and clear the ```_index_to_agent``` dict when adding/removing agents
* The ```_agent_to_index``` dictionary is used to keep track of the agents placed in the space, also when the other two  (```_index_to_agent``` and ```_agent_points```) are cleared.

I've also added a test that just adds many agents (200K). Before this pull request, running the full test suite took 43 seconds; after this change the full test suite takes 6.5 seconds.

Fixes #1375

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -747,8 +747,11 @@ class ContinuousSpace:
     """Continuous space where each agent can have an arbitrary position.
 
     Assumes that all agents are point objects, and have a pos property storing
-    their position as an (x, y) tuple. This class uses a numpy array internally
-    to store agent objects, to speed up neighborhood lookups.
+    their position as an (x, y) tuple.
+
+    This class uses a numpy array internally to store agent objects, to speed
+    up neighborhood lookups. This array is calculated on the first neighborhood
+    lookup, and is reused (and updated) until agents are added or removed.
     """
 
     _grid = None
@@ -782,7 +785,23 @@ class ContinuousSpace:
 
         self._agent_points: npt.NDArray[FloatCoordinate] | None = None
         self._index_to_agent: Dict[int, Agent] = {}
-        self._agent_to_index: Dict[Agent, int] = {}
+        self._agent_to_index: Dict[Agent, int | None] = {}
+
+    def _build_agent_cache(self):
+        """Cache Agent positions to speed up neighbors calculations."""
+        self._index_to_agent = {}
+        agents = self._agent_to_index.keys()
+        for idx, agent in enumerate(agents):
+            self._agent_to_index[agent] = idx
+            self._index_to_agent[idx] = agent
+        self._agent_points = np.array(
+            [self._index_to_agent[idx].pos for idx in range(len(agents))]
+        )
+
+    def _invalidate_agent_cache(self):
+        """Clear cached data of Agents and positions in the space."""
+        self._agent_points = None
+        self._index_to_agent = {}
 
     def place_agent(self, agent: Agent, pos: FloatCoordinate) -> None:
         """Place a new agent in the space.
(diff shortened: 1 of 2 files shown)


mesa/space.py
```diff
@@ -808,11 +823,15 @@ def move_agent(self, agent: Agent, pos: FloatCoordinate) -> None:
             pos: Coordinate tuple to move the agent to.
         """
         pos = self.torus_adj(pos)
-        idx = self._agent_to_index[agent]
-        self._agent_points[idx, 0] = pos[0]
-        self._agent_points[idx, 1] = pos[1]
         agent.pos = pos
 
+        if self._agent_points is not None:
```
rht, line 828: Why is there no else: self._build_agent_cache() after this if?

rht: I see. Because the cache is only needed whenever `get_neighbors` is called.

jiskattema: The cache only needs to be built directly before using it.
If you would add ```self._build_agent_cache()``` here, then moving an agent becomes the bottleneck (as each move includes recalculating the _agent_points``` etc.).

rht:
@jiskattema the CI fails at codespell and the lints. So you might want to fix them. Also, you need to edit your first message in this PR to contain "Fixes #1375" because your 2nd comment is not considered by GH.

jiskattema:
A bulk function place_agents() would be a cleaner approach, indeed.
However, i wanted to make as few changes as possible, and not introduce new API.

Corvince:
> I was about to say that the code would have been more readable if instead you define the bulk function `place_agents`, but this would still result in slow `place_agent`, so it's not ideal. I agree with the approach of this PR.

That was my first thought as well. Why do you think this wouldn't work?

rht:
@jiskattema just need you to squash the commits to 1, disable the test (because a perf test will slow down the CI), and then I will merge.

rht:
Also, change the commit message to be descriptive of the content instead of saying "Fix issue x".

Corvince:
Wait, there is a neighborhood cache that is only (Re-)calculated when agents are added or removed? What happens if an agent moves? Won't this lead to a wrong neighborhood?

Also regarding place agent vs place agents. I think a function for bulk inserting could be even faster so we are leaving some performance gains on the table if we only call place agent sequentially (which I think is sufficiently fast for adding single agents, as intended)

rht:
Both add and remove will invalidate the cache. Moving agent will update cache, but only if it exists.

With the current implementation in this PR, `add_agent` is as fast as any potential new `add_agents`.

Corvince:
Thanks for the clarification, yes after looking closely the removable overhead of place agent in a potential place agents function seems negligible. LGTM

rht:
Also, make sure it is not too long:
> Capitalized, short (50 chars or less) summary.

You can just prefix your message with "perf:" to indicate a performance improvement.

rht:
You have changed the PR title, but not the commit message.

The imperative commit message thing is pretty standard in most projects, but I will update the CONTRIBUTING.rst anyway. Thanks for letting me know.

rht:
Merged. Thank you @jiskattema for the perf improvement!

PR #1378, opened 2022-07-11
ci: Cancel previous obsolete runs

rht:
This should reduce the number of unnecessary CI runs.


tpike3 (approved):
I had to do some [research](https://docs.github.com/en/actions/using-jobs/using-concurrency) to understand what this does. It looks like a good process. 

LGTM

PR #1379, opened 2022-07-11
docs: Add Tim Pope's guideline for proper Git commit msg

rht:


tpike3:
LGTM as the standard we adopt.... I will give 1 more day for any alternative approaches and then merge.

PR #1380, opened 2022-07-11
Fix stray " in modular_template.html

joedight:
It confuses my text editor.


rht:
Merged. Thank you for the catch.

PR #1381, opened 2022-07-12
Fix zoom on network visualisation

joedight:
d3.event was removed in https://github.com/d3/d3/releases/tag/v6.0.0, which breaks zooming / translating.

Also, centre the graph with the zoom object rather than beforehand, to prevent everything from jumping when you first zoom.

### DIFF
--- a/mesa/visualization/templates/js/NetworkModule_d3.js
+++ b/mesa/visualization/templates/js/NetworkModule_d3.js
@@ -23,11 +23,12 @@ const NetworkModule = function (svg_width, svg_height) {
     .attr("class", "d3tooltip")
     .style("opacity", 0);
 
-  svg.call(
-    d3.zoom().on("zoom", () => {
-      g.attr("transform", d3.event.transform);
-    })
-  );
+  const zoom = d3.zoom()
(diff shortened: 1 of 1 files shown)


rht:
> Also, centre the graph with the zoom object rather than beforehand, to prevent everything from jumping when you first zoom.

Which line does this? It would be great if you can do this in a separate 2nd commit, so that it is clearer. As it is right now, I can't tell.

joedight:
Sorry, hopefully that's clearer.

rht:
I suppose `d3.zoomIdentity` is the key for the centering?

joedight:
I think that zoom keeps an internal transform, which is just the identity by default (i.e. it ignores the old attr(transform) bit). 
So when we zoom, it will apply the zoom to the identity then apply that transform to our graph, and we end up zooming at (0, 0).
If we apply the transform to the zoom thing, its internal transform updated to the right position, and further zooming, etc is properly relative to that.

But I know nothing about d3, so could very well be wrong.

rht:
OK, it works anyway, so, merging.

rht:
Thanks for the fix!

Issue #1385, opened 2022-07-17
bug: Duplicate IDs not idenitfied in RandomActivationByType


tpike3:
**Describe the bug**
Duplicate ids are not identified when using `RandomActivitionByType`

**Expected behavior**
If the code tries to add an agent with the same id as a previous agent then an error message is raised.  

**To Reproduce**
Build a `RandomActiviationByType` model and duplicate agent ids. No error message will be raised

**Additional context**
I believe the issue is the add function in the `BaseScheduler`  [Line 67](https://github.com/projectmesa/mesa/blob/a2afbdb40a6f39e2781b30e0de6968ea3090fe38/mesa/time.py#L67)  should raise the exception but the add function for `RandomActivationByType`  [Line 231](https://github.com/projectmesa/mesa/blob/a2afbdb40a6f39e2781b30e0de6968ea3090fe38/mesa/time.py#L231) supersedes Line 67 and does not have the message.

PR #1386, opened 2022-07-17
fix: Ensure agent id is unique in RandomActivationByType.add

rht:


examples/sugarscape_cg/sugarscape_cg/model.py
```diff
@@ -44,9 +44,11 @@ def __init__(self, width=50, height=50, initial_population=100):
         import numpy as np
 
         sugar_distribution = np.genfromtxt("sugarscape_cg/sugar-map.txt")
+        agent_id = 0
```
tpike3, line 47: I just updated all the sugarscape g1mt to be unique_id so it is consistent. I am good with this being left as agent_id but want to make sure we sync.

rht: I think this needs to have `agent_` prefix, because in the class definitions, we know that it is in the context of the agent, but here we need to make it explicit that it is not used for something else.

rht: You can use it for something else like getting the number of agents, but it is mainly used tforgiving identifiers to the agents.

tpike3 (approved):
LGTM

rht:
You forgot to open an issue documenting the gotcha so that it can be found from search engine.

tpike3:
> You forgot to open an issue documenting the gotcha so that it can be found from search engine.

Issue #1385, now linked and closed

PR #1387, opened 2022-07-17
Fix broken monospace links

fkohrt:
Nested inline markup – here: a combination of [inline literals] and [embedded URIs] – is currently [not supported] by reStructuredText. Now using a substitution reference defined with the "replace" directive.

[inline literals]: https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#inline-literals
[embedded URIs]: https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#embedded-uris-and-aliases
[not supported]: https://docutils.sourceforge.io/FAQ.html#is-nested-inline-markup-possible

### DIFF
--- a/docs/tutorials/intro_tutorial.rst
+++ b/docs/tutorials/intro_tutorial.rst
@@ -652,13 +652,15 @@ applies each agent-level collection function to each agent currently in
 the schedule, associating the resulting value with the step of the
 model, and the agent’s ``unique_id``.
 
-Let’s add a DataCollector to the model with
-```mesa.DataCollector`` <https://github.com/projectmesa/mesa/blob/main/mesa/datacollection.py>`__,
+Let’s add a DataCollector to the model with |mesa.DataCollector|__,
 and collect two variables. At the agent level, we want to collect every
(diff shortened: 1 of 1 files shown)


rht:
I tested by looking at https://github.com/fkohrt/mesa/blob/patch-1/docs/tutorials/intro_tutorial.rst. It doesn't seem to be working.

fkohrt:
> It doesn't seem to be working.

At least not in GitHub. But using a public [reST](https://livesphinx.herokuapp.com/) [viewer](http://rst.aaroniles.net/) I get the indented result for the changed source.

rht:
OK, I tested at https://livesphinx.herokuapp.com/, and it seems to be working. Merging now. And will check the RTD version as a follow up. Thank you for the fix!

rht:
It works! But the `mesa.DataCollector` is not in blue, and so people are actually less likely to know that it is clickable. I think we should just remove the monospace code quote and just do mesa.Datacollector.

rht:
@fkohrt do you want to do a PR making them blue (remove the code quote)?

fkohrt:
Yes, see #1388.

PR #1391, opened 2022-07-21
fix: time: Recompute agent_keys between stages

rht:


rht:
I should have added a test case where agents are removed within the stage, where I define a `MockModel` that can remove itself. But this PR has been confirmed to work by @clarkpetri https://github.com/projectmesa/mesa/discussions/1390#discussioncomment-3194264.

tpike3 (approved):
LGTM

tpike3:
> I should have added a test case where agents are removed within the stage, where I define a `MockModel` that can remove itself. But this PR has been confirmed to work by @clarkpetri [#1390 (reply in thread)](https://github.com/projectmesa/mesa/discussions/1390#discussioncomment-3194264).

With the Sugarscape g1mt, I am starting to think there is more complications for these approaches but since we have an identified bug I think it is best to approve and then we can add testing and potential other fixes/enhancements later as this is non-trivial.

rht:
I'm worried about the g1mt reappearing. As such, I have made #1392. The test should apply to all schedulers type, not just `RandomActivationByType`. But it's a start.

PR #1395, opened 2022-07-26
Fix batchrunner progress bar

joedight:
It currently just shows the number of iterations done, not the progress, because total is passed in wrong (see [https://tqdm.github.io/docs/tqdm/#\_\_init\_\_](https://tqdm.github.io/docs/tqdm/#__init__))


rht:
LGTM! Thanks for the fix.

PR #1401, opened 2022-08-09
mypy: space: Remove unnecessary type imports

rht:
I just discovered in the [current doc](https://mypy.readthedocs.io/en/stable/kinds_of_types.html) that importing these are no longer necessary. @wang-boyu check this out for Mesa-Geo, in particular, in https://github.com/projectmesa/mesa-geo/blob/main/mesa_geo/raster_layers.py.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -26,13 +26,9 @@ import numpy as np
 from typing import (
     Any,
     Callable,
-    Dict,
     Iterable,
     Iterator,
-    List,
-    Set,
     Sequence,
-    Tuple,
     TypeVar,
     Union,
     cast,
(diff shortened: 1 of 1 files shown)


rht:
From the CI result, it looks like it is for Python >= 3.9 only. But maybe I can do `from __future__ import ...` stuff.

rht:
According to this [SO comment](https://stackoverflow.com/questions/63460126/typeerror-type-object-is-not-subscriptable-in-a-function-signature#comment118727625_63460126), `from __future__ import annotations` should have fixed it. But it doesn't seem to be the case.

wang-boyu:
In this case shall I still do it for mesa-geo? Perhaps not?

rht:
Let's revisit this refactor in the future.

PR #1409, opened 2022-09-07
Fix stray " in visualisation dropdown labels

joedight:
It appears in the label, just a trivial annoyance.


rht:
Good catch. The CI failed due to a spelling issue, but has nothing to do with this PR.

PR #1412, opened 2022-09-19
Add "falsy" to .codespellignore

rht:
This fixes this Codespell error: https://github.com/projectmesa/mesa/actions/runs/3010330720/jobs/4836305531#step:4:18.
This is due to newer Codespell considers this a typo.


tpike3:
Thanks!

PR #1413, opened 2022-09-19
Allow batch_run to take arbitrary parameters

### LINKED ISSUE #1280: Batchrunner does not handle exotic input parameters?

joedight:
Currently parameter values must be hashable to index the iteration counter (and dictionaries, etc. aren't).

The most sensible way of fixing this seems to be removing the counters
and deciding run_id and iteration number ahead of time (which additionally makes
it deterministic which is nice).

Fixes #1280

### DIFF
--- a/mesa/batchrunner.py
+++ b/mesa/batchrunner.py
@@ -15,7 +15,6 @@ from multiprocessing import Pool, cpu_count
 from warnings import warn
 from typing import (
     Any,
-    Counter,
     Dict,
     Iterable,
     List,
@@ -67,7 +66,13 @@ def batch_run(
         [description]
     """
 
-    kwargs_list = _make_model_kwargs(parameters) * iterations
+    runs_list = []
+    run_id = 0
+    for iteration in range(iterations):
+        for kwargs in _make_model_kwargs(parameters):
+            runs_list.append((run_id, iteration, kwargs))
+            run_id += 1
+
     process_func = partial(
         _model_run_func,
         model_cls,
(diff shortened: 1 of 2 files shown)


rht:
The implementation logic is simpler too.

> (which additionally makes it deterministic which is nice)

Though, I think the current implementation is already deterministic.

My only concern before merging this is that since each param is repeated `iterations` time first, before going to the next param. The ordering is different, and hence the random numbers are used differently. This may cause a problem with people who seed their simulation, only to find their simulation giving a different result mysteriously when they update Mesa.

Maybe we can get away with issuing a warning about this in the release note.

joedight:
> My only concern before merging this is that since each param is repeated `iterations` time first, before going to the next param. The ordering is different, and hence the random numbers are used differently. This may cause a problem with people who seed their simulation, only to find their simulation giving a different result mysteriously when they update Mesa.

I've changed that now, the behaviour should match what it was before.

> Though, I think the current implementation is already deterministic.

Sorry, I meant determinism in assigning iteration and run numbers given imap_unordered can return out-of-order (I think).

rht:
That was a simple fix. Merged, thank you!

PR #1416, opened 2022-09-22
readme: Improve the pip install for Git repo instruction

rht:
This fixes #1415.

I'm not sure if the `-U` is necessary. It's a bit unclear even after reading https://stackoverflow.com/questions/17710947/pip-pulling-updates-from-remote-git-repository. I no longer can test since #1415 is solved. But I don't think adding `-U` will cause any problem.


jackiekazil:
LGTM

PR #1422, opened 2022-09-26
Upgrade pre-commit CI (with pyupgrade and syntax checks)

EwoutH:
This commit upgrades the [pre-commit](https://pre-commit.com/) configuration in the following ways:
 - It fixes and updates the [Black](https://github.com/psf/black) configuration (black-jupyter is inclusive of black)
 - It adds [pyupgrade](https://github.com/asottile/pyupgrade)
 - It check for trailing whitespaces and if toml and yaml syntax is correct
 - Updates itself monthly (using [pre-commit.ci](https://pre-commit.ci/))

pre-commit currently isn't enabled in Mesa's CI. I would recommend enabling it on https://pre-commit.ci/. Otherwise a custom GitHub Actions workflow could be written (but that's slower).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,21 @@
+ci:
+    autoupdate_schedule: 'monthly'
+    autofix_prs: false
+
 repos:
--   repo: https://github.com/ambv/black
-    rev: 22.6.0
+-   repo: https://github.com/psf/black
+    rev: 22.8.0
     hooks:
-      - id: black
-        exclude: ^mesa/cookiecutter-mesa/
       - id: black-jupyter
+        exclude: ^mesa/cookiecutter-mesa/
+-   repo: https://github.com/asottile/pyupgrade
+    rev: v2.38.2
+    hooks:
+    -   id: pyupgrade
+        args: [--py37-plus]
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.3.0  # Use the ref you want to point at
+    hooks:
+    -   id: trailing-whitespace
+    -   id: check-toml
+    -   id: check-yaml


jackiekazil:
> pre-commit currently isn't enabled in Mesa's CI. I would recommend enabling it on https://pre-commit.ci/. Otherwise a custom GitHub Actions workflow could be written (but that's slower).

^^ Slower in speed and by how much? I am inclined to keep everything in github actions & github when possible for simplicity. But if there is a huge speed difference or something else, I could be talked into it.

rht:
About 3x faster, based on the graph in https://pre-commit.ci/. GH Actions takes 45 s, while pre-commit.ci takes ~15 s.

> auto fixing pull requests: if tools make changes to files during a pull request, pre-commit.ci will automatically fix the pull request.

> pre-commit.ci will run on all pull request commits, but will not push for commits made by bots.

This will likely complicate the PR branch that people have to check for the auto-added commits and have to `git pull`, otherwise, they will observe that their `git push` is rejected. Not a problem for people who are used to Git and GH interface, but you can't assume this knowledge by default.

EwoutH:
Their own benchmarks talk about around 4x. The largest difference is that pre-commit.ci caches it's configs by default (with some effort this can probably be replicated). 

I experience around a 2 to 4 times speed difference for most runs on other repos.

But I think speed is not the big differentiator, having pre-commit.ci enabled means we don't have to maintain another CI configuration. GitHub Actions just requires a bunch of extra steps to setup the environment which pre-commit.ci does by itself.

There was an GitHub Action [available](https://github.com/pre-commit/action) previously, but it's depreciated in favor of pre-commit.ci.

Having looked further into it, `pre-commit-update.yml` workflow isn't even needed with pre-commit.ci, because it will auto update weekly by itself.

> This will likely complicate the PR branch that people have to check for the auto-added commits and have to `git pull`

This is a good point. How many of the developers would not be familiar enough with Git to be able to work with this, do you think? Would some documentation help?

rht:
- I'm in favor of pre-commit.ci since it is officially endorsed by the pre-commit GH Actions repo
- Once a week is too frequent. I'd say once a month, just like with the Pipfile.lock upgrade
- I can't estimate the contributors demographic, but disabling auto-fix and instead, just do the checks sounds good to me

jackiekazil:
I agree with @rht

EwoutH:
Agreed, I made the changes. Now only a maintainer needs to enable it on [pre-commit.ci](https://pre-commit.ci/).

rht:
I tried enabling it. But it requires this permission:
> Act on your behalf

Which I am not comfortable of granting it. It shouldn't require this much permission to make periodic upgrade and PR checks.

rht:
Raised https://github.com/pre-commit-ci/issues/issues/146. Let's wait for their response.

rht:
I have enabled it. Merging now.

PR #1428, opened 2022-09-30
feat: Allow user-specified local dir to be served by Tornado

rht:


mesa/visualization/ModularVisualization.py
```diff
@@ -342,6 +344,15 @@ def __init__(
         self.model_kwargs = model_params
         self.reset_model()
 
+        local_dirs.append(local_dir)  # Set this to be the last one.
+        for _local_dir in local_dirs:
+            local_handler = (
+                r"/local/(.*)",
```
rht, line 350: Need to check with Tornado doc on whether this reusing of the pattern is even allowed.

mesa/visualization/ModularVisualization.py
```diff
@@ -328,6 +327,9 @@ def __init__(
                     self.local_css_includes.add(include_file)
                 else:
                     self.local_js_includes.add(include_file)
+            if len(element.local_includes) > 0:
+                if element.local_dir != "":
+                    local_dirs.append(element.local_dir)
```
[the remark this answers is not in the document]
wang-boyu: I'm also wondering what'll happen if two origin folders with the same structure get mapped to the same destination folder.

Mesa-Geo has a `js` folder and a `css` folder under `templates` folder that maps to `/local`. Perhaps Mesa-Geo users also define their own frontend elements in the same way. Then we have another set of `js` and `css` folders mapped to `/local`. Not sure whether this will work.

mesa/visualization/ModularVisualization.py
wang-boyu: I'm also wondering what'll happen if two origin folders with the same structure get mapped to the same destination folder.

Mesa-Geo has a `js` folder and a `css` folder under `templates` folder that maps to `/local`. Perhaps Mesa-Geo users also define their own frontend elements in the same way. Then we have another set of `js` and `css` folders mapped to `/local`. Not sure whether this will work.

rht: I found https://groups.google.com/g/python-tornado/c/xYp6DJTlCsw. One suggestion was to define a multi file handler custom class.

Aside, I suppose the only way to know that it is allowed is to test it. Since it works on Mesa-Geo, does that mean it is fine?

wang-boyu: It is fine most likely due to there is only one defined `local_dir` from Mesa-Geo. I did some tests with my https://github.com/wang-boyu/agents-and-networks-in-python example. Initially it also works fine, when I use your branch and mine.

There is an issue when I define a custom frontend module in the example, using the same folder structure. Something like

```python
class CustomFrontendModule(VisualizationElement):
    local_includes = [
        "js/CustomFrontendModule.js",
        "css/external/test.css",
        "js/external/test.js",
    ]
    local_dir = os.path.dirname(__file__) + "/../templates"

    def __init__(self):
        super().__init__()
```

Then I use the new element together with that from Mesa-Geo. If I put it after Mesa-Geo's `MapModule` like this:

```python
    map_element = MapModule(
        agent_draw, **map_params[args.campus], map_height=600, map_width=600
    )
    custom_element = CustomFrontendModule()
    server = ModularServer(
        AgentsAndNetworks,
        [map_element, custom_element, clock_element, status_chart, friendship_chart],
        "Agents and Networks",
        model_params=model_params,
    )
```

then the map gets loaded properly and I get the following messages from the command line:

```bash
404 GET /local/css/external/test.css (127.0.0.1) 2.49ms
404 GET /local/js/CustomFrontendModule.js (127.0.0.1) 3.43ms
404 GET /local/js/external/test.js (127.0.0.1) 2.42ms
```

If I swap them, i.e.,

```diff
server = ModularServer(
        AgentsAndNetworks,
-        [map_element, custom_element, clock_element, status_chart, friendship_chart],
+        [custom_element, map_element, clock_element, status_chart, friendship_chart],
        "Agents and Networks",
        model_params=model_params,
    )
```

then the map is gone and I get the following:

```bash
404 GET /local/css/external/leaflet.css (127.0.0.1) 1.70ms
404 GET /local/js/MapModule.js (127.0.0.1) 2.65ms
404 GET /local/js/external/leaflet.js (127.0.0.1) 1.27ms
```

So it seems that `/local` gets mapped on a first-come-first-served basis.

Perhaps the `MultiStaticFileHandler` worth a try? But if there are multiple files from different `local_dir` with the same name, I guess the first one gets found and loaded.

Another more complex implementation is we somehow keep track of which package each `local_dir` comes from, and create dedicated folder accordingly. For instance we can have something like:

```
local
├── mesa_geo
│   ├── js
│   │   └── index.js
│   └── css
└── agents-and-networks
    ├── js
    │   └── index.js
    └── css
```

We also need to be able to load files from their folders only, so that the two `index.js` get loaded from their own packages correctly.

rht: Another alternative (hopefully simpler) is to have a static file handler that serves from a Python string instead.
```python
local_includes = ["local/my_custom_path/script.js"]
local_handler = (
    r"/local/my_custom_path/script.js",
    tornado.web.StaticFileHandler,
    {"file_content": the_file_content_str},
)
```

rht: I can't test this idea easily, but maybe you can. Here is a sketch:
1. ask custom viz author to specify a dictionary of path -> string_content
2.
```python
custom_handlers = []
for path, content in custom_contents.items():
    class MyHandler(tornado.web.RequestHandler):
        def get(self):
            return content
    custom_handlers.append(("/" + path, MyHandler))
self.handlers += custom_handlers
```

wang-boyu: Hmm interesting. Let me see what I can do.

rht: Yeah, that looks like another clean solution. I think you can make a PR to projectmesa/mesa straight away, to make the process simpler.

wang-boyu: Created https://github.com/projectmesa/mesa/pull/1435.

rht:
Superseded by #1435

PR #1429, opened 2022-09-30
refactor: Apply pyupgrade --py37-plus

rht:
This should make the pre-commit.ci check to finally pass.


jackiekazil:
LGTM - also -- I didn't realize these were in the code.

PR #1432, opened 2022-10-01
ModularServer: Put new optional arg port last

### LINKED ISSUE #1431: `OSError: Int or String expected` when running boid_flockers example
**Describe the bug**

Running the `boid_flockers` example results in `OSError: Int or String expected`.

**Expected behavior**

Examples should be able to run without errors.

**Additional context**

This is likely due to a breaking change introduced through https://github.com/projectmesa/mesa/pull/1403: a new parameter `port` is added before `model_params` in `ModularServer.__init__()`, i.e.,

```diff
    def __init__(
-         self, model_cls, visualization_elements, name="Mesa Model", model_params=None
+         self,
+         model_cls,
+         visualization_elements,
+         name="Mesa Model",
+         port=None,
+         model_params=None,
    ):
```

As a result, in the `boid_flockers` example, `model_params` gets passed into `__init__()` as `port`:

```python
server = mesa.visualization.ModularServer(
    BoidFlockers, [boid_canvas], "Boids", model_params
)
```

Examples such as `bank_reserves` are not affected:

```python
server = mesa.visualization.ModularServer(
    BankReserves,
    [canvas_element, chart_element],
    "Bank Reserves Model",
    model_params=model_params,
)
```

rht:
This is a newly added argument, and so we should preserve the existing arguments order instead.

Fixes #1431.

### DIFF
--- a/mesa/visualization/ModularVisualization.py
+++ b/mesa/visualization/ModularVisualization.py
@@ -257,8 +257,8 @@ class ModularServer(tornado.web.Application):
         model_cls,
         visualization_elements,
         name="Mesa Model",
-        port=None,
         model_params=None,
+        port=None,
     ):
         """
         Args:


jackiekazil:
LGTM

PR #1399, opened 2022-08-03
ModularServer: Always set model.running = True on reset

rht:
Addresses https://github.com/projectmesa/mesa/discussions/1397.
- [x] tested.
~Do not merge yet~ can merge now.

PR #1435, opened 2022-10-03
feat: Allow user-specified local dir to be served by Tornado

wang-boyu:
This is an alternative implementation of https://github.com/projectmesa/mesa/pull/1428 and https://github.com/projectmesa/mesa/pull/1369, and helps implement a feature in Mesa-Geo: https://github.com/projectmesa/mesa-geo/issues/76.

Some related discussions: https://github.com/projectmesa/mesa/discussions/1364#discussioncomment-2993622 and https://github.com/projectmesa/mesa/pull/1428#discussion_r985328707.

### DIFF
--- a/mesa/visualization/ModularVisualization.py
+++ b/mesa/visualization/ModularVisualization.py
@@ -136,6 +136,11 @@ class VisualizationElement:
         local_includes: A list of JavaScript and CSS files that are local to
                         the directory that the server is being run in.
         js_code: A JavaScript code string to instantiate the element.
+        local_dir: A full path to the directory containing the local includes.
+                   If a relative path is given, it is relative to the working
+                   directory where the server is being run. If an absolute path
+                   is given, it is used as-is. Default is the current working
+                   directory.
 
     Methods:
         render: Takes a model object, and produces JSON data which can be sent
@@ -147,6 +152,7 @@ class VisualizationElement:
     local_includes = []
     js_code = ""
     render_args = {}
+    local_dir = ""
 
     def __init__(self):
         pass
@@ -291,13 +297,7 @@ class ModularServer(tornado.web.Application):
             tornado.web.StaticFileHandler,
             {"path": os.path.dirname(__file__) + "/templates"},
         )
-        local_handler = (
-            r"/local/(.*)",
-            tornado.web.StaticFileHandler,
-            {"path": ""},
-        )
-
-        self.handlers = [page_handler, socket_handler, static_handler, local_handler]
+        self.handlers = [page_handler, socket_handler, static_handler]
 
         self.settings = {
             "debug": True,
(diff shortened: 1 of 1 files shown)


mesa/visualization/ModularVisualization.py
```diff
@@ -147,6 +148,7 @@ class VisualizationElement:
     local_includes = []
     js_code = ""
     render_args = {}
+    local_dir = ""
```
rht, line 151: Probably need a comment explaining what this default behavior is. Is it relative to the directory of user launching the Python code or something else? What if `local_dir` is precomputed on the fly to be the `__file__` of the class definition?

wang-boyu: Good point! Based on Tornado's source code: [[1]](https://github.com/tornadoweb/tornado/blob/7689e3f01e6ba82555f2cfac204f6f2cb7cae82b/tornado/web.py#L2610) and [[2]](https://github.com/tornadoweb/tornado/blob/7689e3f01e6ba82555f2cfac204f6f2cb7cae82b/tornado/web.py#L2751-L2766), if `local_dir` is absolute path such as `os.path.dirname(__file__) + "/../templates"`, then this path will be used.

If it's relative path such as the default `""` or `../src/templates`, I guess the current working directory is used: see [os.path.abspath](https://docs.python.org/3/library/os.path.html#os.path.abspath).

wang-boyu: Updated docstring.

rht:
LGTM

rht:
We need this feature in Mesa-Geo. And we have had several features merged since 1.0.0. I think we should release 1.1.0 soon? @jackiekazil @tpike3?

jackiekazil:
@rht sounds like a plan. I don't mile doing a deploy, but could use help this week putting it together. Can you start up a ticket for the release & help tag anything with the Stafford milestone https://github.com/projectmesa/mesa/milestone/31  that should be included in the history file?

rht:
IMO, the highlighted changes:
- #1376 > 6x perf speedup for add/remove agent in `ContinuousSpace`
- #1391 correctness fix for `SimultaneousActivation` and `StagedActivation`
- #1399 make `self.running = True` optional. We need to tell existing users that initializing this is no longer necessary, and so, reducing the boilerplate code
- #1435 (this PR)

rht:
And #1413, which is a huge feature. I somehow missed this one when going through the commits.

rht:
I have made #1436, but is there a way to automatically add the issues that have been closed since 1.0.0 to the history file? Adding them to the milestone seems like duplication of work.

rht:
I'm able to get all the newly closed issues with this search query: https://github.com/projectmesa/mesa/issues?q=is%3Aissue+is%3Aclosed+updated%3A%3E%3D2022-07-07. But these are clearly not representative of the commits that were merged without closing a ticket.

rht:
I used this search query `is:closed updated:>=2022-07-07` and had put all the recently closed issues and PRs to the milestone.

edgeofchaos42:
Yes 100% we need a release, particularly with the tutorial I need to eliminate the requirement for `self.running=True`. 

Thanks @rht for getting the notes together!!!

PR #1439, opened 2022-10-05
Prevent seed and random from being shared between instances

### LINKED ISSUE #1438: Seed and random should not be class attributes
**Describe the bug**

Every time a model instance is created, it updates the seed and random attribute of every other instance of the same class.

**Expected behavior**

I would expect each instance to have their own attribute. This way, the behavior of each instance is independent and predictable.

**To Reproduce**

```pycon
>>> from mesa import Model
>>> class Ex(Model):
...   def __init__(self, seed=2):
...     pass
...
>>> a = Ex(seed=1)
>>> print(a._seed, a.random)
1 <random.Random object at 0x282a8d0>
>>> b = Ex(seed=2)
>>> print(a._seed, a.random)
2 <random.Random object at 0x282b2d0>
```

balkian:
This commit aims to change two related issues:

- New instances of a model should not be able to overwrite the value of the seed of previous (and possibly running) models.
- The `random` attribute should be instance-level, not class-level.

Prior to this commit, both `_seed` and the `random` were class attributes that were set/changed every time a new instance was created.

This fixes #1438

### DIFF
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -21,9 +21,10 @@ class Model:
 
     def __new__(cls, *args: Any, **kwargs: Any) -> Any:
         """Create a new model object and instantiate its RNG automatically."""
-        cls._seed = kwargs.get("seed", None)
-        cls.random = random.Random(cls._seed)
-        return object.__new__(cls)
+        obj = object.__new__(cls)
+        obj._seed = kwargs.get("seed", None)
+        obj.random = random.Random(obj._seed)
+        return obj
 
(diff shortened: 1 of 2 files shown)


rht:
The solution looks good. The remaining blocker to this PR being merged is that you need to format your changes with `black`.

rht:
Sorry, one more request, do you mind squashing your commits to 1?

balkian:
Sure. I hope I've done it right :)

rht:
LGTM

rht:
Merged. Thank you @balkian!

PR #1440, opened 2022-10-07
Improved clamp function and measured performance

Tortar:
Hi to all!

I've measured the performance of the possible alternatives for the clamp function, this is the code:

```
import timeit

a = timeit.timeit(stmt="clamp(0.3, 0.0 ,1.0)", setup='import numpy as np\ndef clamp(x: float, lowest: float, highest: float) -> float: return np.clip(x, lowest, highest)')

b = timeit.timeit(stmt="clamp(0.3, 0.0 ,1.0)", setup='def clamp(x: float, lowest: float, highest: float) -> float: return max(lowest, min(x, highest))')

c = timeit.timeit(stmt="clamp(0.3, 0.0 ,1.0)", setup='def clamp(x: float, lowest: float, highest: float) -> float: return lowest if x < lowest else highest if x > highest else x')

print(a, b, c)

# result on test run: 12.690470200002892 0.3120805000071414 0.16836619999958202
```

The performance comparison doesn't change for numbers outside the range. Then, the fastest alternative is option c, which is about 75 times faster than the clip function on the test run.  Finally, I replaced the function with that alternative, hope it helps!

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -56,9 +56,8 @@ F = TypeVar("F", bound=Callable[..., Any])
 
 
 def clamp(x: float, lowest: float, highest: float) -> float:
-    # This should be faster than np.clip for a scalar x.
-    # TODO: measure how much faster this function is.
-    return max(lowest, min(x, highest))
+    # much faster than np.clip for a scalar x.
+    return lowest if x <= lowest else (highest if x >= highest else x)
 
 
 def accept_tuple_argument(wrapped_function: F) -> F:


rht:
That's great. I don't remember the operator precedence for the inequalities and the if-else, and so it would be clearer if you add parentheses. Also, it needs to be `>=`, `<=`, not `>`, `<`.

Tortar:
I added your suggestions @rht

mesa/space.py
```diff
@@ -56,9 +56,8 @@
 
 
 def clamp(x: float, lowest: float, highest: float) -> float:
-    # This should be faster than np.clip for a scalar x.
-    # TODO: measure how much faster this function is.
-    return max(lowest, min(x, highest))
+    # much faster than np.clip for a scalar x.
+    return lowest if x <= lowest else (highest if x >= highest else x)
```
rht, line 60: Are you sure about the `<=` and `>=` sign? Looks wrong to me.

rht:
I can do the rebase from Git only, not from the GH interface.

Have you read https://github.com/projectmesa/mesa/blob/main/CONTRIBUTING.rst about how to update your PR properly?

mesa/space.py
Tortar: I'm pretty sure about it, when you have `x `lower than `lowest` it returns  `lowest` as in the original function and so on...

rht: Oh nvm, looks correct. However, it probably needs to be like this?
```python
x = highest if x >= highest else x
# with the new x
return lowest if x <= lowest else x
```

Tortar: mmmh...I think that the order of the operations doesn't matter if `highest >= lowest`. But in some way the other case should be treated in a special way it could be done your way, what do you think about this?

Tortar: anyway timing your alternative it goes on about the same speed `c = timeit.timeit(stmt="clamp(0.3, 0.0 ,1.0)", setup='def clamp(x: float, lowest: float, highest: float) -> float: \n\tx = highest if x >= highest else x\n\t# with the new x\n\treturn lowest if x <= lowest else x')`

rht: Your implementation is better. If you test on arbitrary 3 numbers, sometimes it is faster, because when `x <= lowest` is true, the second if-else block doesn't even get evaluated, which saves time.

rht: Now you just need to:
1. rebase properly
2. squash your commits
3. properly write your commit message according to Tim Pope's guideline (see CONTRIBUTING.rst)

Anyway, this is all already specified in CONTRIBUTING.rst.

Tortar: thanks, I'm working on that!

Tortar:
I finally made it @rht !

rht:
Sorry for another round of request, but the commit message header needs to be:
> Write your commit message in the imperative: "Fix bug" and not "Fixed bug"
or "Fixes bug."  This convention matches up with commit messages generated
by commands like git merge and git revert.

as per https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html.

Tortar:
my fault @rht , now it's formatted

rht:
Merged, thank you @Tortar!

PR #1446, opened 2022-10-11
Memoize positions in HexGrid.find_neighbors

Tortar:
Hi to all !

I found that the HexGrid.find_neighbors didn't memoize positions, making it really slow when the radius of the neighborhood increases, with this PR it becomes much faster, here it is a local comparison I made in relation to the size of the radius:

Figure_2

As you can see, now it's much faster...it becomes 2500 times faster when the radius is 9! A small drop of 4% and 2% in performance instead when the radius is 1 and 2 respectively.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -633,9 +633,15 @@ class HexGrid(Grid):
         def torus_adj_2d(pos: Coordinate) -> Coordinate:
             return (pos[0] % self.width, pos[1] % self.height)
 
+        explored = dict()
         coordinates = set()
 
         def find_neighbors(pos: Coordinate, radius: int) -> None:
+
+            if pos in explored and explored[pos] >= radius:
+                return
+            explored[pos] = radius
+
             x, y = pos
 
             """
--- a/tests/test_grid.py
+++ b/tests/test_grid.py
@@ -391,6 +391,9 @@ class TestHexGrid(unittest.TestCase):
         neighborhood = self.grid.get_neighborhood((1, 1), include_center=True)
         assert len(neighborhood) == 7
 
+        neighborhood = self.grid.get_neighborhood((1, 1), radius=3)
+        assert len(neighborhood) == 12
+
 
 class TestHexGridTorus(TestBaseGrid):
     """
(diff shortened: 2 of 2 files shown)


mesa/space.py
```diff
@@ -666,10 +666,13 @@ def find_neighbors(pos: Coordinate, radius: int) -> None:
             if radius > 1:
                 [find_neighbors(coords, radius - 1) for coords in adjacent]
 
+        explored, coordinates = dict(), set()
```
rht, line 669: They should have been defined in L636. And should have been split in 2 lines.

mesa/space.py
```diff
@@ -666,10 +666,13 @@ def find_neighbors(pos: Coordinate, radius: int) -> None:
...
         find_neighbors(pos, radius)
 
-        if not include_center and pos in coordinates:
-            coordinates.remove(pos)
+        if include_center:
```
rht, line 672: Just commenting for other reviewers: this code chunk is a refactor, and has nothing to do with the memoization.

mesa/space.py
```diff
@@ -646,9 +649,6 @@ def find_neighbors(pos: Coordinate, radius: int) -> None:
             """
             adjacent = [(x, y - 1), (x, y + 1)]
 
-            if include_center:
```
rht, line 649: I'm not sure about removing these lines. This might change the behavior. And it seems you intend this to be a refactor unrelated to the memoization.

rht:
The optimization is great. However, I suggest undoing the refactor part, and do it in a separate PR, so that this PR can be merged sooner. Also, you can squash the commits to avoid additional review roundtrip, since the LOC change in this PR is small.

mesa/space.py
Tortar: yes, you are right the two parts you pointed out are not related to the memoization, but I think they are nonetheless an improvement because you don't need to test the include_center condition inside the recursive function but just at the end, but I think that since this is a different change from the memoization part it's better to address it in another pr

rht: The problem with your refactor is that there are holes in the `adjacent` regardless of `include_center` is true or not, because you removed L649-650. If you can't decisively prove that this doesn't change current behavior, you should remove the refactor.

Tortar: we discuss it in another pr, I followed your suggestions to remove refactoring from this one :-)

mesa/space.py
Tortar: done!

rht:
LGTM

PR #1447, opened 2022-10-12
Drop Python 3.7 support and require Python 3.8

EwoutH:
This PR drops Python 3.7 support and starts requiring Python 3.8. This PR includes dropping the Python 3.7 CI job and running pyupgrade (in pre-commit) with `--py38-plus`.

Dropping Python 3.8 is in line with [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html), the depreciation policy many scientific Python projects follow. Currently users of Python 3.7 can still use the brand new [1.1.0 release](https://github.com/projectmesa/mesa/releases/tag/v1.1.0) of mesa. With this release done, dropping 3.7 allows to start using 3.8 features in current development.

Python 3.8 provides many new features, including Assignment Expressions (the walrus operator `:=`), Positional-Only Parameters and three Typing-related PEPs.

> **Major new features of the 3.8 series, compared to 3.7**
> 
> - [PEP 572](https://www.python.org/dev/peps/pep-0572/), Assignment expressions
> - [PEP 570](https://www.python.org/dev/peps/pep-0570/), Positional-only arguments
> - [PEP 587](https://www.python.org/dev/peps/pep-0587/), Python Initialization Configuration (improved embedding)
> - [PEP 590](https://www.python.org/dev/peps/pep-0590/), Vectorcall: a fast calling protocol for CPython
> - [PEP 578](https://www.python.org/dev/peps/pep-0578), Runtime audit hooks
> - [PEP 574](https://www.python.org/dev/peps/pep-0574), Pickle protocol 5 with out-of-band data
> - Typing-related: [PEP 591](https://www.python.org/dev/peps/pep-0591) (Final qualifier), [PEP 586](https://www.python.org/dev/peps/pep-0586) (Literal types), and [PEP 589](https://www.python.org/dev/peps/pep-0589) (TypedDict)
> - Parallel filesystem cache for compiled bytecode
> - Debug builds share ABI as release builds
> - f-strings support a handy = specifier for debugging
> - continue is now legal in finally: blocks
> - on Windows, the default asyncio event loop is now ProactorEventLoop
> - on macOS, the spawn start method is now used by default in multiprocessing
> - multiprocessing can now use shared memory segments to avoid pickling costs between processes
> - typed_ast is merged back to CPython
> - LOAD_GLOBAL is now 40% faster
> - pickle now uses Protocol 4 by default, improving performance
> 
> There are many other interesting changes, please consult the ["What's New"](https://docs.python.org/3/whatsnew/3.8.html) page in the documentation for a full list.


rht:
Nit: the commit message needs to say >=3.8 instead.

EwoutH:
Thanks for merging!

I know you've been working on typing, there are a [few new typing features](https://docs.python.org/3/whatsnew/3.8.html) that now can be used!

PR #1448, opened 2022-10-12
Refactor HexGrid.iter_neighborhood

Tortar:
Small refactor of the include_center condition inside HexGrid. 

This works because 

- if `include_center == True` then we add the center after the execution of the `find_neighbors` function. Before this PR, the function appended `pos`  at each call, but that was not necessary because after the first execution which appends the center, it appends positions already present in the `coordinates` set because of previous executions of the function. Doing it this way makes the function a bit faster ;
- if `include_center == False`  we need to discard the center after the calls to `find_neighbors` because it could be present anyway, for example with the possible sequence `(x, y) -> (x + 1, y - 1) -> (x + 1, y - 1)`.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -646,9 +646,6 @@ class HexGrid(Grid):
             """
             adjacent = [(x, y - 1), (x, y + 1)]
 
-            if include_center:
-                adjacent.append(pos)
-
             if x % 2 == 0:
                 adjacent += [(x - 1, y + 1), (x - 1, y), (x + 1, y + 1), (x + 1, y)]
             else:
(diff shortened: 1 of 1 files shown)


rht:
>  but that was not necessary because after the first execution which appends the center, it appends positions already present in the coordinates set because of previous executions of the function.

The centers are not being appended in all the `find_neighbors`'s because you removed `adjacent.append(pos)`. Only the original center (the input argument of `iter_neighborhood`) is appended at the end, not all of the "off" centers.

Tortar:
Try this:

```
from mesa.space import HexGrid

width, height = 100, 100
neighborhood = HexGrid(width, height, torus=False).get_neighborhood((10, 10), include_center=True, radius=4)
```

rewriting the original condition as:

```
 if include_center:
     if pos not in coordinates: print(pos)
     adjacent.append(pos)
```

so that it prints when the condition is satisfied, you'll see that the only case when that happens it with the original center

rht:
OK that's a decisive argument. To summarize, the "off" centers are already being added to `coordinates` even before them become the center in the recursive calls with `radius - 1`.

rht:
Keep them (the perf optimizations) coming!

PR #1449, opened 2022-10-13
Updates doc build to 3.9, 3.10 not supported

jackiekazil:
Previous PR caused a break in our doc build - https://github.com/projectmesa/mesa/pull/1447 

3.10 is causing build errors - but it looks like 3.9 is supported - https://github.com/readthedocs/readthedocs.org/issues/7554
This small PR should fix the issue. (Unless it turns out 3.9 isn't supported as well.)


rht:
LGTM

PR #1450, opened 2022-10-13
Update to python 3.8, 3.9 breaks the build

jackiekazil:
RE: Readthedocs -- `HEAD is now at c76546b Updates doc build to 3.9, 3.10 not supported`

Need to downgrade further to fix the build.

Related: #1449 #1447


rht:
The commit message needs to be imperative, and it needs to say that it is about the doc build.

rht:
LGTM

PR #1452, opened 2022-10-15
Defer memoization in HexGrid.find_neighbors

Tortar:
This PR defers the memoization of the positions. In this way less calls to the function HexGrid.find_neighbors are made and also the calculation of the neighbors with radius 1 doesn't memoize positions which indeed was not necessary. Besides, the recursion is a bit more explicit since now it's no more inside a list comprehension.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -638,10 +638,6 @@ class HexGrid(Grid):
 
         def find_neighbors(pos: Coordinate, radius: int) -> None:
 
-            if pos in explored and explored[pos] >= radius:
-                return
-            explored[pos] = radius
-
             x, y = pos
 
             """
@@ -667,7 +663,13 @@ class HexGrid(Grid):
             coordinates.update(adjacent)
 
             if radius > 1:
-                [find_neighbors(coords, radius - 1) for coords in adjacent]
+                decr_radius = radius - 1
+                for coord in adjacent:
+                    if coord in explored and explored[coord] >= decr_radius:
+                        continue
+                    explored[coord] = decr_radius
+                    # recursive call
+                    find_neighbors(coord, decr_radius)
 
         find_neighbors(pos, radius)
 


rht:
Can you quantify the percentage reduction of the function calls? How much faster does this PR get than current main?

mesa/space.py
```diff
@@ -667,7 +663,13 @@ def find_neighbors(pos: Coordinate, radius: int) -> None:
             coordinates.update(adjacent)
 
             if radius > 1:
-                [find_neighbors(coords, radius - 1) for coords in adjacent]
+
```
rht, line 666: There shouldn't be an empty line here.

Tortar:
I tested it with 

```
import timeit, mesa.space, mesa.space_2

res = []
for radius in range(1,18):
    setup_1 = "from mesa.space import HexGrid\nwidth, height = 100, 100\ngrid = HexGrid(width, height, torus=False)"
    setup_2 = "from mesa.space_2 import HexGrid\nwidth, height = 100, 100\ngrid = HexGrid(width, height, torus=False)"

    stmt = f"grid.get_neighborhood((1, 10), include_center=True, radius={radius})"

    a = min(timeit.repeat(stmt=stmt, setup=setup_1, repeat=100, number=100))
    b = min(timeit.repeat(stmt=stmt, setup=setup_2, repeat=100, number=100))

    print([a, b, a/b])
    res.append([a, b, a/b])
print(res)
```

where `mesa.space_2` has the modified version, these are the results:

graph

So it's from 5 to 20% faster in this range, not so much indeed.

I also checked putting a global counter inside the two functions the number of calls, and the new version has more or less 1/5 the number of calls of the other other one independently of the radius.

rht:
20% is definitely significant.

The decrease in the number of function calls is definitely not a good proxy for the perf improvement. Most of the original function calls that are gone are essentially the branches where the function exits early anyway. The cause of the speedups seems to be that we now have only 1 `coord in explored and explored[coord] >= decr_radius` replacing multiple of them inside the child `find_neighbors` calls.

Tortar:
I have a better version! So I think this one is no more necessary

PR #1453, opened 2022-10-16
Delete recursion from HexGrid.iter_neighborhood

Tortar:
I deleted the recursive function find_neighbors inside HexGrid.iter_neighborhood and made the function iterative, I also tweaked other parts of the code to speed it up a bit the function (but I'm not satisfied with the black formatting, what do you think about it?). I tested it with 

```
import timeit, mesa.space, mesa.space_2

res = []
for radius in range(1,18):
    
    setup_1 = "from mesa.space import HexGrid\nwidth, height = 100, 100\ngrid = HexGrid(width, height, torus=True)"
    setup_2 = "from mesa.space_2 import HexGrid\nwidth, height = 100, 100\ngrid = HexGrid(width, height, torus=True)"

    stmt = f"grid.get_neighborhood((1, 10), include_center=True, radius={radius})"

    a = min(timeit.repeat(stmt=stmt, setup=setup_1, repeat=10, number=1000))
    b = min(timeit.repeat(stmt=stmt, setup=setup_2, repeat=10, number=1000))

    print([a, b, a/b])
    res.append([a, b, a/b])
print(res)
```

where `mesa.space_2` contains the modified code. The results when `self.torus == False` is that it is always faster, here it is the case when `self.torus == True`, when the radius is equal to 1 there is a drop of 4%, in all other cases is faster as can be seen in the following graph:

graph

From the algorithmic point of view we passed from recursive dfs to iterative bfs which in a language like Python, where recursion is not so efficiently managed, makes the code run  faster

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -18,6 +18,7 @@ MultiGrid: extension to Grid where each cell is a set of objects.
 from __future__ import annotations
 
 import itertools
+import collections
 import math
 from warnings import warn
 
@@ -604,11 +605,20 @@ class HexGrid(Grid):
             in the neighborhood of a certain point.
     """
 
+    def torus_adj_2d(self, pos: Coordinate) -> Coordinate:
+        return pos[0] % self.width, pos[1] % self.height
+
     def iter_neighborhood(
         self, pos: Coordinate, include_center: bool = False, radius: int = 1
     ) -> Iterator[Coordinate]:
         """Return an iterator over cell coordinates that are in the
-        neighborhood of a certain point.
+        neighborhood of a certain point. To calculate the neighborhood
+        for a HexGrid the parity of the x coordinate of the point is
+        important, the neighborhood can be sketched as:
+
+            Always: (0,-), (0,+)
+            When x is even: (-,+), (-,0), (+,+), (+,0)
+            When x is odd:  (-,0), (-,-), (+,0), (+,-)
 
         Args:
             pos: Coordinate tuple for the neighborhood to get.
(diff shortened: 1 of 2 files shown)


Tortar:
what do you think about this PR @rht ?

mesa/space.py
```diff
@@ -630,46 +640,55 @@ def iter_neighborhood(
...
-            else:
-                adjacent = [torus_adj_2d(coord) for coord in adjacent]
-
-            coordinates.update(adjacent)
-
-            if radius > 1:
-                [find_neighbors(coords, radius - 1) for coords in adjacent]
-
-        find_neighbors(pos, radius)
+        while radius != 0:
```
rht, line 647: Should be `> 0` just in case.

rht:
> what do you think about this PR @rht ?

It's not clear why it is faster. Also, in this case, it probably needs more test cases to ensure the code has identical output as the previous version.

rht:
The algorithm looks fine to me. The only bottleneck to this PR being merged is to test it with more cases, taking into account of corner cases.

rht:
> From the algorithmic point of view we passed from recursive dfs to iterative bfs which in a language like Python, where recursion is not so efficiently managed, makes the code run faster

Oh, I didn't see this until just now. Did you edit your first post? If so, it would have been noticeable if you made a new comment instead.

rht:
By corner cases, I meant unexpected rare gotchas that may happen for some of the inputs.

Tortar:
Added some tests with also some corner cases

rht:
I think the one with `self.torus = True` is already covered by `TestHexGridTorus`. Both `TestHexGrid` and `TestHexGridTorus` seem to cover the 2 cases of `include_center`. The only test that needs to be added is probably for when the x, y is at the corner of the grid for both `TestHexGrid` and `TestHexGridTorus`, where we expect both cases to differ, and this PR has the same output as the current main.

Tortar:
clear, I missed TestHexGridTorus :-)

Tortar:
@rht I added (maybe) the right tests

rht:
That looks correct. I ran the tests in this PR with the current main, and they passed. I found it strange that the (0, 0) corner has only 1 extra neighbor when torus is true, but this concern is out of the scope of this PR; the current main has the same behavior anyway. Merging now.

rht:
I suppose this means we have ~10k speedup when the radius is 9 over the original recursive code without the memoization? @Tortar it would be great if you could measure it just for radius = 9, so that we can highlight it in the next release note. This is exciting!

Issue #1243, opened 2022-04-04
refactor: space.py's _place_agent method should be absorbed into place_agent


rht:
We don't need 2 separate methods. Only `place_agent` is needed, so that the library is simpler.
To work on this, wait until #1240 is merged.

Tortar:
This should be solved now

rht:
Closed via #1454

Tortar:
this is still not closed though :-)

PR #1454, opened 2022-10-17
Absorb _place_agent method in place_agent

Tortar:
This solves [#1243](https://github.com/projectmesa/mesa/issues/1243) absorbing _place_agent method in place_agent


rht:
LGTM

PR #1455, opened 2022-10-18
Change OrderedDict in dict

Tortar:
Given that Mesa 0.8.8 is the last version supporting Python 3.5 then we can safely move from OrderedDict  to dict I think


Tortar:
It should be good to go now!

rht:
LGTM

PR #1456, opened 2022-10-18
Remove unnecessary list operations

Tortar:
The transformation in list is not necessary to iterate over the keys

### DIFF
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -89,7 +89,7 @@ class BaseScheduler:
 
     def get_agent_count(self) -> int:
         """Returns the current number of agents in the queue."""
-        return len(self._agents.keys())
+        return len(self._agents)
 
     @property
     def agents(self) -> list[Agent]:
@@ -142,14 +142,12 @@ class SimultaneousActivation(BaseScheduler):
 
     def step(self) -> None:
         """Step all agents, then advance them."""
-        agent_keys = list(self._agents.keys())
-        for agent_key in agent_keys:
-            self._agents[agent_key].step()
-        # We recompute the keys because some agents might have been removed in
-        # the previous loop.
-        agent_keys = list(self._agents.keys())
-        for agent_key in agent_keys:
-            self._agents[agent_key].advance()
+        for agent in self._agents.values():
+            agent.step()
+        # the previous steps might remove some agents, but
+        # this loop will go over the remaining existing agents
+        for agent in self._agents.values():
+            agent.advance()
         self.steps += 1
         self.time += 1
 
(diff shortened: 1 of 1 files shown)


mesa/time.py
```diff
@@ -142,12 +142,12 @@ class SimultaneousActivation(BaseScheduler):
 
     def step(self) -> None:
         """Step all agents, then advance them."""
-        agent_keys = list(self._agents.keys())
+        agent_keys = self._agents.keys()
```
rht, line 145: I think you can combine with the next line to do:
```python
for agent in self._agents.values():
    agent.step()
```

mesa/time.py
```diff
@@ -142,12 +142,12 @@ class SimultaneousActivation(BaseScheduler):
...
         for agent_key in agent_keys:
             self._agents[agent_key].step()
         # We recompute the keys because some agents might have been removed in
         # the previous loop.
-        agent_keys = list(self._agents.keys())
+        agent_keys = self._agents.keys()
```
rht, line 150: The comment above will become obsolete. We should change it to say that the previous step might remove some agents, but this loop will go over the remaining existing agents, and so, there is no problem.

mesa/time.py
Tortar: right, it seems unnecessary to step the keys and not the values

Tortar: wait but maybe we can even not calculate everytime the keys!

Tortar:
Made all necessary changes I think @rht

mesa/time.py
```diff
@@ -142,14 +142,13 @@ class SimultaneousActivation(BaseScheduler):
...
-        # the previous loop.
-        agent_keys = list(self._agents.keys())
-        for agent_key in agent_keys:
-            self._agents[agent_key].advance()
+        agent_values = self._agents.values()
+        for agent in agent_values:
+            agent.step()
+        # the previous steps might remove some agents, but
+        # this loop will go over the remaining existing agents
+        for agent in agent_values:
```
rht, line 150: Why not just `for agent in self._agents.values()` for both? With the current version, it hasn't been proven that the generator is auto-updated when some agents are removed in `step()`.

Tortar: mmhh I'm not following you, I did it with `agent_values = self._agents.values()` so that not to call again the `dict.values()` again, it can have some side effects in your opinion? I think that it's okay because `agent_values` is a reference to a view object so it updates automatically

rht: `dict.values()` returns a generator-like object in Python 3, so it's a low-cost operation.

> I think that it's okay because agent_values is a reference to a view object so it updates automatically

Need a proof just in case it doesn't introduce regression.

Tortar: tested the cost of `dict.values()` -> really low cost, so I put the way you thought

rht:
LGTM

PR #1459, opened 2022-10-19
Increase maximum line length to 100 characters

EwoutH:
This PR adds a `pyproject.toml` file that configures [Black](https://github.com/psf/black) with a maximum line length of 100 characters, up from the default of 88.

In many cases, this makes code more readable by allowing assignments and expressions to stay on a single line. It also encourages the use of longer, more descriptive variable names, which can make code easier to understand.

Please scroll through the diff and see if you agree this makes the code more readable. I would suggest viewing both the _Split_ and _Unified_ diff.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -212,9 +212,7 @@ latex_elements = {
 # Grouping the document tree into LaTeX files. List of tuples
 # (source start file, target name, title,
 #  author, documentclass [howto, manual, or own class]).
-latex_documents = [
-    ("index", "Mesa.tex", "Mesa Documentation", "Project Mesa Team", "manual")
-]
+latex_documents = [("index", "Mesa.tex", "Mesa Documentation", "Project Mesa Team", "manual")]
 
 # The name of an image file (relative to this directory) to place at the top of
 # the title page.
--- a/docs/tutorials/intro_tutorial.ipynb
+++ b/docs/tutorials/intro_tutorial.ipynb
@@ -953,11 +953,7 @@
     "# First, we filter the results\n",
     "one_episode_wealth = results_df[(results_df.N == 10) & (results_df.iteration == 2)]\n",
     "# Then, print the columns of interest of the filtered data frame\n",
-    "print(\n",
-    "    one_episode_wealth.to_string(\n",
-    "        index=False, columns=[\"Step\", \"AgentID\", \"Wealth\"], max_rows=25\n",
-    "    )\n",
-    ")\n",
+    "print(one_episode_wealth.to_string(index=False, columns=[\"Step\", \"AgentID\", \"Wealth\"], max_rows=25))\n",
     "# For a prettier display we can also convert the data frame to html, uncomment to test in a Jupyter Notebook\n",
     "# from IPython.display import display, HTML\n",
     "# display(HTML(one_episode_wealth.to_html(index=False, columns=['Step', 'AgentID', 'Wealth'], max_rows=25)))"
--- a/examples/bank_reserves/bank_reserves/model.py
+++ b/examples/bank_reserves/bank_reserves/model.py
@@ -40,9 +40,7 @@ def get_num_mid_agents(model):
     """return number of middle class agents"""
 
     mid_agents = [
-        a
-        for a in model.schedule.agents
-        if a.loans < 10 and a.savings < model.rich_threshold
+        a for a in model.schedule.agents if a.loans < 10 and a.savings < model.rich_threshold
     ]
     return len(mid_agents)
 
--- a/examples/bank_reserves/batch_run.py
+++ b/examples/bank_reserves/batch_run.py
@@ -54,9 +54,7 @@ def get_num_mid_agents(model):
     """list of middle class agents"""
 
     mid_agents = [
-        a
-        for a in model.schedule.agents
-        if a.loans < 10 and a.savings < model.rich_threshold
+        a for a in model.schedule.agents if a.loans < 10 and a.savings < model.rich_threshold
     ]
     # return number of middle class agents
     return len(mid_agents)
--- a/examples/boid_flockers/boid_flockers/model.py
+++ b/examples/boid_flockers/boid_flockers/model.py
@@ -60,14 +60,7 @@ class BoidFlockers(mesa.Model):
             pos = np.array((x, y))
             velocity = np.random.random(2) * 2 - 1
             boid = Boid(
-                i,
-                self,
-                pos,
-                self.speed,
-                velocity,
-                self.vision,
-                self.separation,
-                **self.factors
+                i, self, pos, self.speed, velocity, self.vision, self.separation, **self.factors
             )
             self.space.place_agent(boid, pos)
             self.schedule.add(boid)
--- a/examples/boid_flockers/boid_flockers/server.py
+++ b/examples/boid_flockers/boid_flockers/server.py
@@ -18,6 +18,4 @@ model_params = {
     "separation": 2,
 }
 
-server = mesa.visualization.ModularServer(
-    BoidFlockers, [boid_canvas], "Boids", model_params
-)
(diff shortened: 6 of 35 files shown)


jackiekazil:
I am kind of in favor of 120. @rht do you have any strong feels?

EwoutH:
Didn't expect the discussion to be about 100 vs 120 instead of 88 vs 100! :)

For writing code, I like going to 120. For refactoring or reviewing I'm not so sure, since with side-by-side viewing sometimes 2x 120 gets quite wide.

I'm curious what others think!

rht:
If you look at the list of projects featured in https://github.com/psf/black#used-by, [Django](https://github.com/django/django/commit/7119f40c9881666b6f9b5cf7df09ee1d21cc8344), [pandas](https://github.com/pandas-dev/pandas/blob/d2acdee72b33c8c52bb5f8b902190bbfcf27ca47/pyproject.toml#L59), [pytest](https://github.com/pytest-dev/pytest/blob/baada535a3e31ceb541a3e81319309a35bec934c/pyproject.toml#L115) stick to 88. [Zulip](https://github.com/zulip/zulip/blob/f2f4bed85c7259b113941958327800165f53b8a1/pyproject.toml#L2) chooses 100.

I think going from 88 to 100 is less abrupt. Another use case is that in my case, I often use split view for my text editor to juxtapose 2 different lines, and I only have a small laptop screen while doing this.

rht:
cc: @wang-boyu as the decision will affect Mesa-Geo.

wang-boyu:
Thank you @rht 

To make things more complicated if it isn't already, there's another option of '90-ish' from [Black documentation](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length):

> You probably noticed the peculiar default line length. Black defaults to 88 characters per line, which happens to be 10% over 80. This number was found to produce significantly shorter files than sticking with 80 (the most popular), or even 79 (used by the standard library). In general, [90-ish seems like the wise choice](https://youtu.be/wf-BqAjZb8M?t=260).

rht:
I watched the rest of that PyCon video; I highly recommend watching it. The speaker talked about wrapping an API library that is written in idiomatic Java, to idiomatic Python.

I think the number of bytes saved by going to higher line length approximately plateaus after 88.

EwoutH:
On thing to consider is the increased use of [typing](https://docs.python.org/3/library/typing.html), which adds a lot of characters, but often are nice to read from left to right.

Another thing I really like to keep on one line are list, dict and generator comprehensions.

jackiekazil:
One line comprehensions, while they speed up aren’t necessarily always
better because they can be harder to read.

jackiekazil:
I think we have a few different options here but agree we need to extend…

1. Leave at standard
2. 88
3. 90
4. 100
5. 120

How do people feel at the moment?
Quick vote.

jackiekazil:
My vote is for #2 or #3 — 88 or 90… I lean towards 88 since that is black
standard.

Tortar:
Don't know if I can partecipate but I'd vote for 100 for the same reasons of @EwoutH  :-)

EwoutH:
We currently use the Black standard of 88. So 1 and 2 are the same, except if we want to move the opposite way to shorter lines.

Option 3, moving to 90 is so close that's insignificant, it's basically the same 1 and 2.

So I think it's between staying at 88, or going up to 100 or 120.

jackiekazil:
> So I think it's between staying at 88, or going up to 100 or 120.

Ty for the clarification. Yes, you are right.

jackiekazil:
There are items that are more readable, but there are also items that are list comprehensions that are probably too long to begin with. I will +1 @Corvince's comment as well with respect to landscape on a phone. It doesn't happen often for me but enough to not want to impact that.

wang-boyu:
+1 for the default 88.

The total number of lines that gets reduced is around 150 for the whole repository which doesn't look significant enough for a change. This however increases the risks of having complicated logic in one single line, with added difficulties of side-by-side views.

Agreed with @EwoutH that we either go directly to 100 or 120, or stay at 88.

rht:
> Don't know if I can partecipate but I'd vote for 100 for the same reasons of @EwoutH :-)

@Tortar yes you definitely can.

rht:
More datapoints from the "used by" list in the Black repo readme:
- [Pillow[(https://github.com/python-pillow/Pillow) 88
- [Poetry](https://github.com/python-poetry/poetry/blob/16046d9ac9b72a49e1bc4618fb686695cc64821c/pyproject.toml#L129-L134) 88
- [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy/blob/bbf68345f4993245689b96cc6c6a50013afa3caa/pyproject.toml#L8-L11) 79
- [virtualenv](https://github.com/pypa/virtualenv/blob/cdf4225c689150010af8da1cc913e8f32bee06d2/pyproject.toml#L9) 120

I'm currently waiting for an answer from chat.zulip.org (Zulip dev chat) on why Zulip uses 100.
Relevant Black issue: https://github.com/psf/black/issues/290 (closed).

jackiekazil:
Thank you for pulling together all these data points. Looking forward to
hear why Zulip uses 100.

rht:
The reason seems to be for readability of the import statements: https://chat.zulip.org/#narrow/stream/19-documentation/topic/code.20line.20length/near/1454606. The rest of https://github.com/zulip/zulip code is mostly unaffected when switched from 100 to 88.

rht:
I misread about the import statements; it looks like Tim misconfigured and changed the `isort` config instead of `black`:
> Oops, yeah, I was only changing the isort configuration, not the black configuration.

> But the short answer is probably that while most Python code will be the same when changing a length limit of 100 characters to 88 characters (for example, the only changes in zerver/actions/*.py would be import statement changes), the code that is incrementally wrapped generally felt less readable to me.

EwoutH:
When using smaller screens, having a slightly longer line wrapped to two lines is sometimes better readable then having it spread over 3 lines, and still wrapped.

Line wrapping in editors is quite good.

Screenshot_36

rht:
The tally so far:
- 88: 3 (@Corvince, @jackiekazil, @wang-boyu)
- 100: 2 (@EwoutH, @Tortar)

The majority of the projects featured by Black picks 88. pandas is most relevant to Mesa since it is for data analysis, which likely contains inline computation on arrays. NumPy and SciPy haven't adopted Black.
SciPy requires the line length to be 79 in its [pycodestyle config](https://github.com/scipy/scipy/blob/5350d42d2bafbf4d500fe874ba9af603fff012a8/tools/lint_diff.ini#L2).

Regarding with the argument for Mypy making the function signature longer, it seems that Poetry still picks 88 regardless (e.g. https://github.com/python-poetry/poetry/blob/master/src/poetry/json/__init__.py).

Considering all of the above, my vote is 88.

EwoutH:
So with your vote it's 4-2, I think it's settled. Shall I close this PR?

On a side note, we might have to be a bit careful about voting. It doesn't always encourage constructive discussion. Sometimes it can be a useful tool, but let's be careful to (quickly) default to it.

rht:
@tpike3 anything to add?

I think it's fine to have early voting, which is usually accompanied by reasons for the choice. See how the [video storage voting](https://github.com/projectmesa/mesa/discussions/1411) has fleshed out arguments from both sides.

Corvince:
> 
> On a side note, we might have to be a bit careful about voting. It doesn't always encourage constructive discussion. Sometimes it can be a useful tool, but let's be careful to (quickly) default to it.

I totally agree on that one

tpike3:
@rht Thanks for the bump.  On this one I don't really have a preference. I am content with whatever decision is made.

jackiekazil:
@rht thank you for pushing to close the loop on this one!

rht:
@EwoutH I'm closing the PR. At least we learn something new, even if this doesn't get merged.

PR #1464, opened 2022-10-21
Make Grid.get_neighborhood more readable

Tortar:
All the code was indented but it was possible not to do it, besides the `include_center` condition is now outside of the loop


rht:
Needs a rebase on current main.

rht:
Oh, this is `Grid`, not `HexGrid`.

rht:
It definitely is clearer; LGTM.

rht:
If you are looking into optimizing `Grid.get_neighborhood`, I think using Cython/Numba is the next step. Numba requires less setup.

PR #1465, opened 2022-10-21
Cache results for Hexgrid.iter_neighborhood

Tortar:
Caching strategy added for HexGrid as was added in #823 .

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -632,6 +632,12 @@ class HexGrid(Grid):
             equals at most 9 (8) if Moore, 5 (4) if Von Neumann (if not
             including the center).
         """
+        cache_key = (pos, include_center, radius)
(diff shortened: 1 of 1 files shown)


mesa/space.py
```diff
@@ -632,6 +632,11 @@ def iter_neighborhood(
             equals at most 9 (8) if Moore, 5 (4) if Von Neumann (if not
             including the center).
         """
+        cache_key = (pos, include_center, radius)
+        neighborhood = self._neighborhood_cache.get(cache_key, None)
+
+        if neighborhood is not None:
+            return neighborhood
```
rht, line 639: Shouldn't this be `yield from neighborhood`?

Tortar: ops, sure

rht:
LGTM

PR #1466, opened 2022-10-21
Initialize Grid with list comprehension

Tortar:
Initializing the Grid in list comprehension makes the total time of initialization of a grid instance ~15-20% faster


rht:
LGTM

PR #1470, opened 2022-10-22
docs: overview.rst: Change to new Mesa namespace style

rht:
There is still the `BatchRunner` usage left, but it outputs a DataFrame, which `batch_run` doesn't do out of the box. Should be a quick fix. I can leave that for a separate PR later.

### DIFF
--- a/docs/overview.rst
+++ b/docs/overview.rst
@@ -29,11 +29,9 @@ The skeleton of a model might look like this:
 
 .. code:: python
 
-    from mesa import Agent, Model
-    from mesa.time import RandomActivation
-    from mesa.space import MultiGrid
+    import mesa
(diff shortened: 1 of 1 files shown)


Corvince:
Was there a discussion about the new namespace being the preferred way of importing mesa?

I really like the option of using the new namespace, but personally not yet convinced that it should be the default

rht:
> Was there a discussion about the new namespace being the preferred way of importing mesa?

Yes, and @tpike3 did a survey on the users that they preferred the new namespace.

The rest of the tutorial has used the new namespace, so I'm just making it more consistent here so that it is easier.

jackiekazil:
LGTM

PR #1471, opened 2022-10-23
Refactor Grid.__getitem__

Tortar:
With this refactor the __getitem__ function use list comprehensions when useful and specializes a bit more the subcases of the function so that they each has a different return statement.


rht:
LGTM

PR #1472, opened 2022-10-23
Exchange iter_neighborhood with get_neighborhood

Tortar:
the two methods must be switched to be as in the Grid class


rht:
Ah, I forgot to say that you should prefix your commit message with "HexGrid:", otherwise it is going to be confusing for the changelog.

Tortar:
Amended, it should be good

PR #1476, opened 2022-10-26
Make Grid.get_neighborhood faster

Tortar:
<s>I created two different algorithms to speed up neighborhood calculations, one for Von Neumann neighborhood and one for Moore neighborhood</s>, these are the results of the new get_neighborhood vs the old one. There is now one unique algorithm for all cases. It takes the right neighborhood from the start, so that you don't need to check or transform anything later except for the von Neumann checking

These are the results: you can see how many times faster the new implementation is in respect to the old one in each case for the moore, torus and include center conditions:

Figure_12
[Updated at the latest commit](no more, see the graph with cache enabled below in the comments)

Tested with:

```
import timeit, itertools, mesa.space, mesa.space_2

res = dict()
combs = itertools.product([True, False], repeat=3)
for val_torus, val_include, val_moore in combs:
    
    comb = f"torus={val_torus}, center={val_include}, moore={val_moore}"
    print("\n"+comb+"\n")
    res[comb] = []
    for radius in range(1,11):
        
        setup_1 = f"from mesa.space import Grid\nwidth, height = 100, 100\ngrid = Grid(width, height, torus={val_torus})"
        setup_2 = f"from mesa.space_2 import Grid\nwidth, height = 100, 100\ngrid = Grid(width, height, torus={val_torus})"

        stmt = f"grid.get_neighborhood((50, 50), include_center={val_include}, radius={radius}, moore={val_moore})"

        a = min(timeit.repeat(stmt=stmt, setup=setup_1, repeat=20, number=2500))
        b = min(timeit.repeat(stmt=stmt, setup=setup_2, repeat=20, number=2500))

        print([radius, b, a, b/a])
        res[comb].append([radius, b, a, b/a])

for comb in res: plt.plot([x[0] for x in res[comb]], [x[-1] for x in res[comb]], "-o", label=comb)

plt.xlabel("radius", fontsize=13); plt.ylabel("times faster", fontsize=13)
plt.yticks([1+0.5*x for x in range(11)]); plt.xticks([x for x in range(1, 11)])
plt.hlines([1], xmin=1, xmax=10, linestyles='dashed'); plt.margins(0,0.01)
plt.legend()
plt.show()
```

To achieve the same results the ` _neighborhood_cache` needs to be disabled (otherwise after the first iteration the position is cached and the two algorithms are not compared) and create a copy called `space_2.py` with the update.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -256,29 +256,37 @@ class Grid:
         if neighborhood is not None:
             return neighborhood
 
-        coordinates: set[Coordinate] = set()
+        neighborhood = []
 
         x, y = pos
-        for dy in range(-radius, radius + 1):
-            for dx in range(-radius, radius + 1):
-                # Skip coordinates that are outside manhattan distance
-                if not moore and abs(dx) + abs(dy) > radius:
-                    continue
+        if self.torus:
+            x_radius = min(radius, self.width // 2)
+            y_radius = min(radius, self.height // 2)
 
-                coord = (x + dx, y + dy)
+            for dx in range(-x_radius, x_radius + 1):
+                for dy in range(-y_radius, y_radius + 1):
 
-                if self.out_of_bounds(coord):
-                    # Skip if not a torus and new coords out of bounds.
-                    if not self.torus:
+                    if not moore and abs(dx) + abs(dy) > radius:
                         continue
-                    coord = self.torus_adj(coord)
 
-                coordinates.add(coord)
+                    nx, ny = (x + dx) % self.width, (y + dy) % self.height
+                    neighborhood.append((nx, ny))
 
-        if not include_center:
-            coordinates.discard(pos)
+        else:
+            x_range = range(max(0, x - radius), min(self.width, x + radius + 1))
+            y_range = range(max(0, y - radius), min(self.height, y + radius + 1))
 
-        neighborhood = sorted(coordinates)
+            for nx in x_range:
+                for ny in y_range:
+
+                    if not moore and abs(nx - x) + abs(ny - y) > radius:
+                        continue
+
+                    neighborhood.append((nx, ny))
+
+        if not include_center and neighborhood:
+            neighborhood.remove(pos)
+
         self._neighborhood_cache[cache_key] = neighborhood
 
         return neighborhood


Tortar:
I tried to run the schelling model with this change and It runs in nearly half the time

Tortar:
sorry, last comment was not relevant (the schelling model was already faster than before due to some recent changes, but the grid was in fact too small so that the calculation of the neighborhood was relevant)

mesa/space.py
```diff
@@ -256,29 +259,88 @@ def get_neighborhood(
...
 
-                coord = (x + dx, y + dy)
+        if moore:
+
+            x_from, x_to, x_max = x - radius, x + radius + 1, self.width
+            y_from, y_to, y_max = y - radius, y + radius + 1, self.height
+
+            if self.torus:
+
+                if radius >= x_max // 2:
```
rht, line 271: You should pack this slice calculation in a method/function like how it is done with `torus_adj`. And so it can be generalized to give the slice for both x and y.

It seems to me that the source of the speedup is due to the bound checking is only done once for all.

Tortar: in fact this way it's clearer

Tortar: now it should be all done

Tortar: yes anyway that is one of the reasons  plus that in the torus case it doesn't need to update the position and also that no sum between the deltas and the position is done inside the for loops

Corvince:
Wait, I don't fully get this. Why did you disable the cache? 

This doesn't sound like a realistic scenario

Tortar:
Try this

```
import timeit, itertools, mesa.space, mesa.space_2

res = dict()
combs = itertools.product([True, False], repeat=3)
for val_torus, val_include, val_moore in combs:
    
    comb = f"torus={val_torus}, center={val_include}, moore={val_moore}"
    print("\n"+comb+"\n")
    res[comb] = []
    for radius in range(1,11):
        
        setup_1 = f"from mesa.space import Grid\nfrom random import randint\nwidth, height = 250, 250\ngrid = Grid(width, height, torus={val_torus})"
        setup_2 = f"from mesa.space_2 import Grid\nfrom random import randint\nwidth, height = 250, 250\ngrid = Grid(width, height, torus={val_torus})"

        stmt = f"grid.get_neighborhood((randint(0, width), randint(0, height)), include_center={val_include}, radius={radius}, moore={val_moore})"

        a = min(timeit.repeat(stmt=stmt, setup=setup_1, repeat=10, number=1000))
        b = min(timeit.repeat(stmt=stmt, setup=setup_2, repeat=10, number=1000))

        print([radius, b, a, b/a])
        res[comb].append([radius, b, a, b/a])

for comb in res: plt.plot([x[0] for x in res[comb]], [x[-1] for x in res[comb]], "-o", label=comb)

plt.xlabel("radius", fontsize=13); plt.ylabel("times faster", fontsize=13)
plt.yticks([1+0.5*x for x in range(11)]); plt.xticks([x for x in range(1, 11)])
plt.hlines([1], xmin=1, xmax=10, linestyles='dashed'); plt.margins(0,0.01)
plt.legend()
plt.show()
```

where now cache can be active because the point is random but the difference is nearly the same (1.7 instead of 2 for the moore case with radius 1) and I suspect the difference with the previous one can be explained also by the fact that calling the randint function can reduce the comparison accuracy

mesa/space.py
```diff
@@ -114,6 +114,9 @@ def __init__(self, width: int, height: int, torus: bool) -> None:
 
         # Neighborhood Cache
         self._neighborhood_cache: dict[Any, list[Coordinate]] = dict()
+        # Increments Cache used when calculating Von Neumann neighborhood
```
rht, line 117: Should be von Neumann.

Tortar:
Finally found the right algorithm! Need to format properly and adjusting some minor thing but it should be good!

Tortar:
Ok done everything. There is even the added bonus that running it with pypy gives 

[1, 0.010747599999999968, 0.0017191999999999902, 6.251512331316909]
[2, 0.04063670000000008, 0.002143000000000006, 18.962529164722337]
[3, 0.10254690000000011, 0.0027807999999998056, 36.87676208285647]
[4, 0.1971425, 0.0035098000000006735, 56.1691549375925]
[5, 0.32203939999999953, 0.004266399999998782, 75.48270204389918]
[6, 0.492471799999997, 0.005235700000000065, 94.06035487136216]
[7, 0.7025907999999994, 0.0065768000000012705, 106.82867047802331]
[8, 0.9140565000000009, 0.008295699999997908, 110.18437262681046]
[9, 1.2228171000000003, 0.010296799999998996, 118.75700217544475]
[10, 1.5172522000000015, 0.012148699999997348, 124.89008700522136]

torus=True, center=True, moore=False

[1, 0.005645100000009506, 0.001963899999992691, 2.874433525144109]
[2, 0.019730299999991985, 0.0032858999999945127, 6.004534526316971]
[3, 0.04333880000000079, 0.0053157000000112475, 8.152980792728913]
[4, 0.0853341000000114, 0.007909900000001358, 10.788265338373021]
[5, 0.14030119999999613, 0.011408900000006383, 12.2975221099245]
[6, 0.21180929999999876, 0.015416399999992336, 13.739219272988768]
[7, 0.30273950000000127, 0.019935300000014422, 15.1861020400888]
[8, 0.39695050000000265, 0.025007899999991423, 15.87300413070025]
[9, 0.5405505000000232, 0.031201600000002827, 17.324448105224548]
[10, 0.6930121000000042, 0.0377973999999881, 18.33491457084938]
...
...

So pypy likes even more the change because now all is done with just lists.

Corvince:
> I think it's still fine to have the algo optimized as long as it still is readable and intuitive. 

Totally agree with that statement. It should be a tradeoff between readability and performance. 

I don't want to sound discouraging @Tortar but I just don't see it here with this PR. I think the code is much harder to understand and the benefits seem to be marginal at best for realistic cases, that is with caching enabled. 

As you already pointed out only models with very large grids might benefit from this (at which point the choice of a grid becomes questionable). But even then any serious model probably only spends a fraction of it's time on figuring out neighborhoods. I don't think it is worth optimizing that (with harder to read code).

Again, I don't want to discourage you and am happy that you seem rather actively contributing to mesa.

Tortar:
thank you for the feedback @Corvince but I disagree with you on the fact that now the code is much harder to understand, it just precalculates the ranges where the points are located, the rest is almost untouched

Corvince:
Okay, I have to be more precise: it is much harder **for me** to understand 🤣 

That might be linked with me having no idea what the `ls_` prefix means, what is the difference between `nx` and `dx` and what the `c_` prefix means. You don't have to explain that here, I am sure they make perfect sense, it just isn't obvious for me who hasn't written the code.

And I would like to put more importance of my critique on the performance side. If you can come up with speedups with caching enabled I might be willing to reconsider :)

Tortar:
ok I came up with something to show the algorithmic optimization in case of active caching :-)

```
import timeit, itertools, mesa.space, mesa.space_2
res = dict()
worst = (True, False, False)
best = (False, True, True)
cases = [worst, best]
for val_torus, val_include, val_moore in cases:
    
    comb = f"torus={val_torus}, center={val_include}, moore={val_moore}"
    print("\n"+comb+"\n")
    res[comb] = []
    for radius in range(1,11):
        
        setup_1 = f"from mesa.space import Grid\nwidth, height = 30, 30\ngrid = Grid(width, height, torus={val_torus})"
        setup_2 = f"from mesa.space_2 import Grid\nwidth, height = 30, 30\ngrid = Grid(width, height, torus={val_torus})"

        stmt = f"for x in range(30):\n\tfor y in range(30):\n\t\tk=(x,y)\n\t\tgrid.get_neighborhood(k, include_center={val_include}, radius={radius}, moore={val_moore})"

        a = min(timeit.repeat(stmt=stmt, setup=setup_1, repeat=20, number=100))
        b = min(timeit.repeat(stmt=stmt, setup=setup_2, repeat=20, number=100))
        c = min(timeit.repeat("for x in range(30):\n\tfor y in range(30):\n\t\tk=(x,y)", repeat=20, number=100))
        
        print([radius, a-c, b-c, (a-c)/(b-c)])
        res[comb].append([radius, a-c, b-c, (a-c)/(b-c)])
```

this is equivalent at running 100 ticks and scanning all the grid 30 x 30 at each tick in the worst case and best case scenarios regarding the choice of the parameters value of the function, the results for cpython and pypy are these ones:

Figure_1

mesa/space.py
```diff
@@ -256,29 +254,37 @@ def get_neighborhood(
...
-
         x, y = pos
-        for dy in range(-radius, radius + 1):
-            for dx in range(-radius, radius + 1):
-                # Skip coordinates that are outside manhattan distance
-                if not moore and abs(dx) + abs(dy) > radius:
-                    continue
 
-                coord = (x + dx, y + dy)
+        ls_ranges_x = self.inbound_ranges(x, self.width, radius)
```
rht, line 259: Yeah, expanding the term "ls" would improve the readability. Also, the method name `inbound_ranges` should be changed to be a verb, maybe `calculate_inbound_ranges`?

mesa/space.py
```diff
@@ -330,14 +333,32 @@ def get_neighbors(
...
             radius: radius, in cells, of neighborhood to get.
-
         Returns:
             A list of non-None objects in the given neighborhood;
             at most 9 if Moore, 5 if Von-Neumann
             (8 and 4 if not including the center).
         """
         return list(self.iter_neighbors(pos, moore, include_center, radius))
 
+    def inbound_ranges(self, c: int, c_max: int, radius: int) -> tuple[range]:
```
rht, line 343: Maybe we should make it an internal method, prefixed with `_`, and also say in the docstring that it is currently only used in `get_neighborhood`.

mesa/space.py
```diff
@@ -256,29 +254,37 @@ def get_neighborhood(
...
+                        continue
+                    if not moore:
+                        dx, dy = abs(nx - x), abs(ny - y)
+                        if self.torus:
+                            # In the torus case the deviation from the given point
+                            # can derive from traversing the grid from the outside
+                            dx_out = self.width - dx
+                            dy_out = self.height - dy
+                            if dx_out < dx:
+                                dx = dx_out
```
rht, line 278: You can do the whole thing in 1 line with `dx = min(dx, self.width - dx)`

mesa/space.py
Tortar: Deleted with the simpler `x_ranges` and `y_ranges`

mesa/space.py
Tortar: mmh It's much slower this way unfortunately

mesa/space.py
```diff
@@ -256,29 +254,37 @@ def get_neighborhood(
...
 
-                if self.out_of_bounds(coord):
-                    # Skip if not a torus and new coords out of bounds.
-                    if not self.torus:
-                        continue
-                    coord = self.torus_adj(coord)
+        neighborhood = []
 
-                coordinates.add(coord)
+        for x_range, y_range in itertools.product(x_ranges, y_ranges):
```
rht, line 264: Why do we now require 3 levels of loop, where previously it was just 2?

Tortar: because there is the possibility that there two different ranges for x and/or y in the torus case, so you need to make all the combinations of ranges

Tortar: for example with a grid 10x10 radius 1 from (0,0) gives you range(0,1) and range(9,10)

Tortar: or we could chain together the ranges to solve this

Tortar: I realized there is no way to change this because itertools.chain generates an iterator which is exausted after the first iteration but it can happen to iterate more than one time

mesa/space.py
Tortar: do you think going that way must be dome the same?

rht: Yes, `:=` sounds good, though I'm rather surprised that `min(...)` is significantly slower. By how much? It's also used in https://github.com/projectmesa/mesa/blob/952857c7accdd696cd791474e9f44e4f2a90988a/mesa/space.py#L936. If we implement our own `min`, which does the same as using the `if` statement in this line, would it be faster?

rht:
```python
def get_min(a, b):
  return a if a < b else b
```

rht:
```ipython
In [13]: np.mean(
    ...:     timeit.repeat(
    ...:         "f(4,5)", setup="f = lambda a, b: a if a < b else b", repeat=20, number=100
    ...:     )
    ...: )
Out[13]: 4.7881399950711057e-05

In [14]: np.mean(timeit.repeat("min(4,5)", repeat=20, number=100))
Out[14]: 7.098469868651591e-05
```

mesa/space.py
rht: There might be a stdlib function that can concatenate the range objects.

Tortar: Searched it and didn't find anything unfortunately

mesa/space.py
Tortar: Yet three 2.5 times slower than an inline expression:
min(timeit.repeat("a if a < b else b", setup="a=4; b=5", repeat=20, number=100000)) -> 0.0037641999078
min(timeit.repeat("f(4,5)", setup="f = lambda a, b: a if a < b else b", repeat=20, number=100000)) -> 0.0103960999986

Tortar: Applied the `:=` operator

Tortar:
@rht @Corvince what do you think about it as it is now?

Corvince:
I think the code is much better now and thanks for the clear performance test for caching enabled. 

But I am still not understanding everything about the code immediately without probably taking some time to understand it. 

And I still think the performance improvements are simply not worth it. In the best case scenario we save 1 order of magnitude, which I think is the baseline of accepting performance improvements. 
But on what scale? How often do we have need to call get_neighborhood to save even 1s of runtime? About a million times? (Just guessing) then how long does a mesa model that needs to call that function a million times typically run? 

Looking at it from a software development perspective the maintainability of your PR is much lower and I don't think it is worth it (again, from a pure performance perspective it would be)

Corvince:
By the way I am really enjoying all your performance improvements, because you seem to go down many of the same rabbit holes I went to many moons ago. 

Something I was always interested in but never followed up on is the grid instance itself. Currently we use a list of lists, but some models implement a grid as a single list (think concatenated rows). I always wondered how this would impact performance (as always there are some tradeoffs). Maybe you would be interested to investigate on this?

Tortar:
In reality we save 2 order of magnitudes I squased on purpose the results running 100 ticks, but in reality to save 1 second of runtime for example with radius equal ten we just need to run it 1000 times i.e. the first tick is all what matters because after that there is caching

Tortar:
Yes, I can try to investigate on that, but I intuitively think that it can't impact so much having just a list because calculations become more complex

rht:
I think there are 2 remaining issues to get the code in this PR to be readable+intuitive:
- The 3-level loops are definitely more complex
- `_compute_inbound_ranges` just needs to have more detailed documentation. `torus_adj` is easier to explain, because it is a local operation. But the case is not so different

@Corvince if the 2 issues are addressed, would you consider this PR to be merge-able?

Tortar:
I have the solution for the first requirement!!

Tortar:
The graph represents the perf situation now, it improved even a bit!

Corvince:
> 
> @Corvince if the 2 issues are addressed, would you consider this PR to be merge-able?

Not sure. Honestly I would like to explore the performance improvements myself before I might be convinced, but I am not sure when I will be able to do this. I think we need a tie breaker here. 

That being said I think the code could be made easier to read by avoiding the many nested if clauses. 

Especially the inbound function could benefit from early returns. For example this code

```python
            elif c_to > c_max: 
                 rg_in, rg_out = range(c_from, c_max), range(0, c_to - c_max) 
                 return list(itertools.chain(rg_in, rg_out))
```

Could probably be rewritten as
```
            if c_to > c_max:                  return itertools.chain(range(c_from, c_max), range(0, c_to - c_max))
```

So for example no elif necessary and is there are reason why the ranges were converted to lists before? It be even faster without the conversion?

Corvince:
Also if you check for if not torus you can return early and we can save one level of nesting. So actually the whole function can be implemented without any nesting

Tortar:
mmh don't understand at all your if elif consideration, clearly [using elif is more efficient]( https://stackoverflow.com/questions/22782785/when-to-use-if-vs-elif-in-python) and better sintax

Tortar:
using the first if is good because when you have a radius greater than c_max//2 it would be useless to have the chained ranges in the second elif (it would result in an error if not used because the  neighborhood list would have repetitions)

Tortar:
For the unnesting part, yes it can be done that way but I think it's less readable because the two cases are not clearly separated but I think it's a minimal detail

Corvince:
> Also consider that using the first if is good because when you have a radius greater than c_max//2 it would be useless to have the chained ranges in the second elif (also it would result in an error if not used because the neighborhood list would have duplicates values)

Not sure what do you meany maybe I wasn't clear about which if statement I was referring to? I was talking about switching this line 
```
If self.torus
```

With 
```
If not self.torus
```

Tortar:
Putting aside the perf considerations which can be negligible the fact is that the cases are mutually exclusive using elif states that more clearly than using just ifs, this is clearly stated in many answers in the link

Tortar:
Putting aside the perf considerations which can be negligible the fact is that the cases are mutually exclusive using elif states that more clearly than using just ifs, this is clearly stated in many answers in the link

Tortar:
Got it about the `if not self.torus` condition , but I think it's less readable unnesting that way

Tortar:
I want also to stress that this implementation has another advantage besides the performance benefit: we can create a real iterator if we want, the iterator right now yield from an already built list (so it's a bit useless), now we can yield the tuples if we want because we don't have duplicates anymore

rht:
> I want also to stress that this implementation has another advantage besides the performance benefit: we can create a real iterator if we want, the iterator right now yield from an already built list (so it's a bit useless), now we can yield the tuples if we want because we don't have duplicates anymore

I was also wondering about the iterator from a full list. But what if the user doesn't iterate all the way to the end? Then the caching mechanism, which is triggered only when the entire iterator has been exhausted, doesn't get triggered.

Tortar:
yes, indeed I realized that after writing the comment and tried to delete it, think this in the end is another good reason to deprecate `iter_neighborhood` and keep only `get_neighborhood`. 
Also, even if the user iterates till the end anyway the cache is made of lists so the memory usage doesn't decrease defeating any useful property of iterators

mesa/space.py
```diff
@@ -256,29 +256,33 @@ def get_neighborhood(
...
-                    # Skip if not a torus and new coords out of bounds.
-                    if not self.torus:
-                        continue
-                    coord = self.torus_adj(coord)
+        neighborhood = []
 
-                coordinates.add(coord)
+        for nx in range_x:
+            for ny in range_y:
+                if nx == x and ny == y and not include_center:
```
rht, line 268: Why is it faster that this happens for every loop, vs just discarding it at the end, like before?

mesa/space.py
```diff
@@ -338,6 +342,28 @@ def get_neighbors(
         """
         return list(self.iter_neighbors(pos, moore, include_center, radius))
 
+    def _compute_inbound_ranges(self, c: int, c_max: int, radius: int) -> list[int]:
+        """Compute the 1d-ranges of values where the points in the neighborhood
+        within a certain radius reside. As of now this method is used only in
+        Grid.get_neighborhood."""
+        c_from, c_to = c - radius, c + radius + 1
+        if self.torus:
```
rht, line 350: This is probably what @Corvince meant to reduce the nesting
```python
if not self.torus:
    rg_in = range(c_from if c_from >= 0 else 0, c_to if c_to < c_max else c_max)
    return list(rg_in)

# The rest of the torus case proceeds here without being inside an `else`.
```

Corvince:
Yes, the iter function should be deprecated.

 Historically we had a real iterator and generated the list from it, so it was the other way around. Once we implemented caching it was faster with how it is done currently, but the iter function was kept for backward compatibility

mesa/space.py
Tortar: because a deletion from a list it's O(n) while for a set it is O(1) so I suppose it's (much) slower but I will give it a try

Tortar: but considering that the condition is evaluated n times maybe it could be the same

mesa/space.py
```diff
@@ -338,6 +342,29 @@ def get_neighbors(
...
         return list(self.iter_neighbors(pos, moore, include_center, radius))
 
+    def _compute_inbound_ranges(self, c: int, c_max: int, radius: int) -> list[int]:
+        """Compute the 1d-ranges of values where the points in the neighborhood
+        within a certain radius reside. As of now this method is used only in
+        Grid.get_neighborhood."""
+        c_from, c_to = c - radius, c + radius + 1
+        if not self.torus:
+            rg_in = range(c_from if c_from >= 0 else 0, c_to if c_to < c_max else c_max)
+            return list(rg_in)
```
Corvince, line 352: I think you missed this in my previous comment, but do we need to convert to a list here?

mesa/space.py
Tortar: now I thought that we can remove x and y from the ranges before the loop so that we have better perf

mesa/space.py
Tortar: actually no but the difference I think is very low and for consistency we can just return lists, also the removal can be done the way I mentioned in the other comment just if we convert the range in list

Tortar: in theory where we don't chain we can return a range, will measure the difference between returning ranges sometimes + removal at the end and returning always lists + removal at the start

Tortar: thank you for the good idea anyway! On what you bet? :D

mesa/space.py
Tortar: wrong sorry otherwise we loose other points also

mesa/space.py
Tortar: good idea, my previous considerations were wrong, I will make the proposed changes

mesa/space.py
rht: The output of `list(a_set)` is deterministic, at least since 3.8 (see this [comment](https://github.com/projectmesa/mesa/issues/1052#issuecomment-888468826)). I haven't tested, but `list(a_set)` is likely faster than `sorted(a_set)`.

rht: OK, but at least for a `set[tuple[int, int]]` it is?

Tortar: I read that ints are deterministic, but don't know about tuples unfortunately, at the same type it seems like there is no guarantee on the stability of sets between different versions, so I think it can be risky to rely on that. Also, given the problem at end I don't think using sets would result in better performance actually (this discussion is anyway useful for example for move_to_empty behaviour anyway)  becuase iterating on sets is slower, creating sets is slower, we just have a better removal but I think that the total time would be slightly slower anyway

Tortar: the answer is...no:

try

```
from random import randint
a = set((randint(1,30), randint(1,30)) for x in range(100000))
b = set((randint(1,30), randint(1,30)) for x in range(100000))
# check if the len is 900 before equality between lists
print(len(a) == len(b) == 900)
print(list(a) == list(b))
```

Tortar: or yes maybe, this works for example

```
from random import randint
k = [(randint(1,30), randint(1,30)) for x in range(100000)]
a = set(k)
b = set(k)
print(list(a) == list(b))
```

mesa/space.py
Corvince: Nice, and shouldn't this also work with the chained iterators?

mesa/space.py
Tortar: included the discarding at the end, slightly faster

Tortar: Tried to use a set for `neighborhood` just for seeing if it is faster, it is actually slightly slower (also for the user iterating over it will become slower)

mesa/space.py
Tortar: nope, because the iterators are exausted after the first iteration, while range objects do not, we iterate more than once because we have a double for loop to create neighborhood

Tortar:
One test failed because the neighborhood happened to be empty, I fixed it checking if the  list neighborhood was  non empty after removal, but I realized there is a discussion to make about the current behaviour vs. the old one:

the new behaviour assumes that the user passes a point inside the grid, while the previous one returned only the point inside the grid which were in the neighborhood of the point passed, even if the point was outside, this was an undocumented feature, which I think could be considered a bug because unexpected, so there are two ways to go I think:

- If considered a bug, raising an error if the user passes a point outside the grid;

- If not, adjust the current behaviour to support the old one.

I consider it a bug because I wouldn't expect this behaviour

mesa/space.py
Corvince: Interesting, I did not think about this. But this has me wonder if instead of the nested for loops we could use itertools.product or itertoola.combinations!?

Tortar: like it, will do that way!

Tortar: only a small concern before proceeding, it is ~5-7% slower than the alternative with for loops because iterating on a product iterator is a bit slower, probably an irrelevant difference

Tortar: measure also for pypy it is ~10% slower, with radius 10 and 100 ticks, 11.7 vs. 13 times faster

Tortar: Clearly as you suggested removing all list operations, what do you think?

mesa/space.py
```diff
@@ -256,29 +256,34 @@ def get_neighborhood(
...
                         continue
-                    coord = self.torus_adj(coord)
 
-                coordinates.add(coord)
+                coord = (nx, ny)
+                neighborhood.append(coord)
 
-        if not include_center:
-            coordinates.discard(pos)
+        if not include_center and neighborhood:
```
rht, line 284: I think another constraint to consider when designing the code is that whether it would be easier to translate to Numba/Cython. It would be harder with `set`.

mesa/space.py
Corvince: Thanks for doing this, I did not expect product to be slower! 
What about something like 
`for (x,y) in ((x,y) for x in range_x ( for y in range_y).`

Corvince: Thanks for doing this, I did not expect product to be slower! 
What about something like 
`for (x,y) in ((x,y) for x in range_x ( for y in range_y).`

Corvince: Well I guess that would save one level of nesting, but not sure if this is more or less readable 😅

Tortar: Besides a bit unreadable :D It is actually 3 times slower
```
timeit.timeit("for (x,y) in ((x,y) for x in range(5) for y in range(5)): pass") -> 4.5 secs
timeit.timeit("for x in range(5):\n\tfor y in range(5): pass") -> 1.5 secs
```

mesa/space.py
Tortar: yes clearly true, lists are easier to pass to those libraries and also the speed up boost is much more relevant

Tortar: what do you think about my last comment on points outside the grid?

rht:
> I consider it a bug because I wouldn't expect this behaviour

I agree it has to be a bug with an exception raised.

rht:
The other behavior of returning `[]`, is implicit and has to be checked anyway. It is rather erroneous for e.g. a data collector to collect the number of neighbors as 0 when the length of the return is 0.

Tortar:
Then will raise an exception, for what regards your second comment about the case where list is empty, this will be no longer possible when the fix about points outside the grid will be made, each point inside the grid if the radius passed is >= 1 has at least a neighbor point

Corvince:
> Then will raise an exception, for what regards your second comment about the case where list is empty, a empty list returned will be no longer possible when the fix about points outside the grid will be made, each point inside the grid if the radius passed is >= 1 has at least a neighbor point

Yes i think an exception makes the most sense. I was arguing in my head that maybe on a toroidal space there is no outside and one could argue that the points should be automatically wrapped around, but that can also be handled on the user side and raising an exception will be most consistent

mesa/space.py
Corvince: Oopsie! Interesting, thanks for trying that out. So in general I would prefer itertools.product since the performance difference is negligible, but in this case I would say we stay with the nested for loops (for the best performance)

rht:
Content LGTM. Waiting for an approval from @Corvince before merging.

Though there is a decision on whether to include the raising an exception for out of bound position as a part of this PR, due to it being a breaking change. I'd say we should keep this PR as 100% refactor, leave the exception-raising part out, so that at least the refactor part gets merged sooner.

Tortar:
Rebased the PR not including the exception handling, will do that in a separate PR as you suggested

Corvince:
Yesterday I was finally back at my laptop and could do some testing with this. But I still haven't made my final conclusion. 

One thing is still don't fully understand is where the speed gains really come from. What seems a bit expensive with the current implementation is that we do some work multiple a second time as soon as we call torus_adj, i.e. checking out of bounds and checking for torus. But I feel like this could be prevented with a simpler code then this PR? 

@Tortar what are your impressions as to where the performance boost comes from?

Tortar:
Take the best case, when torus, moore and include_center are all true. The advantages of the new algorithm are:

- Never call self.out_of_bounds inside the loop, while before that was done for each coordinate;
- Never call self.torus_adj, while before that was done for each out of bounds coordinate;
- Don't sum the differences dx and dy to the initial point, because we find the right range of points from the start;
- Don't use a set but use a list, because there is no more the possibility that there are repetitions in neighborhood because the ranges calculated with _compute_inbound_ranges accounts for any overlapping of the points.

Tortar:
Don't think there is a way actually to simplify the code further keeping all these advantages, because to have all of them (or even a subset I think) we need to compute the ranges before looping

Tortar:
Just to address your point about the double cecking of out_of_bounds I changes torus_adj in 

`def torus_adj(self, pos: Coordinate) -> Coordinate:  return pos[0] % self.width, pos[1] % self.height`

and the 10 radius case passes from x4.75 to x4.15, so it doesn't impact a lot

Tortar:
Had some buggy code in local in my previous perf test. To address your point about the double cecking of `self.out_of_bounds` I changed `self.torus_adj` in 

`def torus_adj(self, pos: Coordinate) -> Coordinate:  return pos[0] % self.width, pos[1] % self.height`

and the results are for radius 10:

best from x4.37 to x4.25 and worst from x1.23 to x1.15, so it's something it doesn't impact actually

Corvince:
> Take the best case, when torus, moore and include_center are all true. The advantages of the new algorithm are:
> 
> * Never call `self.out_of_bounds` inside the loop, while before that was done for each coordinate;
> * Never call `self.torus_adj`, while before that was done for each out of bounds coordinate;
> * Don't sum the differences `dx` and `dy` to the initial point, because we find the right range of points from the start;
> * Don't use a set but use a list for `neighborhood`, because there is no more the possibility that there are repetitions in `neighborhood` since the ranges calculated with `self._compute_inbound_ranges` account for any overlapping of the points.

I had a lot of fun trying to come up with my own solution that checks all those boxes. I came up with a very clean and fast solution. However, it didn't pass all the test because of all the corner cases ;)

Nevertheless I found a solution that **I** think is better (i.e. more readable), but my whole critique revolves around the point that your code only appears to be easily readable because you wrote it yourself, so the same might have happened to me.

The solution is actually a tiny bit faster for most cases, although marginally slower for your "best case". But I don't have pypy installed, so it would be great if you could run it through pypy. 

Here is the relevant part: 


```python
        neighborhood = dict()

        if self.torus:
            for (dx, dy) in itertools.product(range(-radius, radius + 1), repeat=2):

                if not moore and abs(dx) + abs(dy) > radius:
                    continue

                neighborhood[((x + dx) % self.width, (y + dy) % self.height)] = None

        else:
            x_range = range(max(0, x - radius), min(self.width, x + radius + 1))
            y_range = range(max(0, y - radius), min(self.height, y + radius + 1))

            for (nx, ny) in itertools.product(x_range, y_range):

                if not moore and abs(nx - x) + abs(ny - y) > radius:
                    continue

                neighborhood[(nx, ny)] = None

        if not include_center and neighborhood:
            del neighborhood[pos]

        neighborhood = list(neighborhood)
```

As you can see the main if/else swith is based on torus. If we have a torus we can work with the relative coordinates and just wrap them around width/height. If we dont have a torus we clamp the x-  and y-range before looping over the coordinates directly.

You can also see that I used a dict to store the coordinates. This is faster than a set, but slower than your list (but prevents double entries).

What do you think @Tortar, @rht

Tortar:
Very Beautiful! Like it a lot, I have a way to make your algorithm even faster than mine in most cases and use a list instead of a dict, but we need nested for loops :-)

```
        neighborhood = []

        x, y = pos
        
        if self.torus:

            radius_x = min(radius, self.width // 2)
            radius_y = min(radius, self.height // 2)

            for dx in range(-radius_x, radius_x + 1):
                for dy in range(-radius_y, radius_y + 1):

                    if not moore and abs(dx) + abs(dy) > radius:
                        continue

                    nx, ny = (x + dx) % self.width, (y + dy) % self.height
                    neighborhood.append((nx, ny))

        else:
            x_range = range(max(0, x - radius), min(self.width, x + radius + 1))
            y_range = range(max(0, y - radius), min(self.height, y + radius + 1))

            for nx in x_range:
                for ny in y_range:

                    if not moore and abs(nx - x) + abs(ny - y) > radius:
                        continue

                    neighborhood.append((nx, ny))

        if not include_center and neighborhood:
            neighborhood.remove(pos)
```

Tortar:
Surely at this point (after almost 100 comments) I need to put you @Corvince  and @rht as co-authors of this pull!

mesa/space.py
```diff
@@ -256,29 +256,37 @@ def get_neighborhood(
...
 
-                if self.out_of_bounds(coord):
-                    # Skip if not a torus and new coords out of bounds.
-                    if not self.torus:
+                    if not moore and abs(dx) + abs(dy) > radius:
                         continue
-                    coord = self.torus_adj(coord)
 
-                coordinates.add(coord)
+                    nx, ny = (x + dx) % self.width, (y + dy) % self.height
```
rht, line 272: I suppose `_compute_inbound_ranges` has complex conditionals due to the torus case, which wrapping is already handled by the `% self.width`. Doing the wrapping handling twice was redundant.

Corvince: Exactly, that was one of the things that I noticed with the previous approach.

Corvince (approved):

PR #1478, opened 2022-10-27
Use list transformation only when shuffled is True

Tortar:
The list transformations are needed just when shuffling

### DIFF
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -28,7 +28,7 @@ from __future__ import annotations
 from collections import defaultdict
 
 # mypy
-from typing import Iterator, Union
+from typing import Iterator, Union, Iterable
 from mesa.agent import Agent
 from mesa.model import Model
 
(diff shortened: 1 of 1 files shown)


mesa/time.py
```diff
@@ -264,8 +267,9 @@ def step(self, shuffle_types: bool = True, shuffle_agents: bool = True) -> None:
             shuffle_agents: If True, the order of execution of each agents in a
                             type group is shuffled.
         """
-        type_keys: list[type[Agent]] = list(self.agents_by_type.keys())
+        type_keys: KeysView[type[Agent]] = self.agents_by_type.keys()
         if shuffle_types:
+            type_keys = list(type_keys)
```
rht, line 272: The type doesn't seem to be consistent here, because it was a `KeysView`, then here it becomes a `list`.

Tortar: Made it Iterable

rht: We should enable Mypy in CI so that I don't have to catch this manually.

rht:
LGTM

PR #1481, opened 2022-10-29
Fix #1419. DataCollector accepts an arbitrary schedule at creation (d…

### LINKED ISSUE #1419: Feature Request: Agent DataCollection Can't Handle Different Attributes in ActivationByType
**Describe the Feature**
Receive attribute error when running an agent reporter through datacollector when using RandomActivationByType 

**Desired behavior**
I can collect against different attributes based on agent type

**To Reproduce**
Create a model with different agent types and attributes and try to collect against that agent attribute

**Additional context**
Feature would need to update at line 170 of datacollection.py `agent_records = map(get_reports, model.schedule.agents)`

woolgathering:
…efaults to model.schedule otherwise) and will return None if an attribute is not found instead of throwing an AttributeError.

Fix #1419. `DataCollector` now accepts an optional fourth argument `schedule` which is an arbitrary schedule in the model. If none is provided, the `DataCollector` defaults to previous behavior, assuming that there exists an attribute to each agent called `model.scheudle`.

This also does not result in an `AttributeError` when an attribute does not exist for an agent. It instead returns `None`.


woolgathering:
I'm unclear why the black hook is failing. Running `flake8 . --ignore=F403,E501,E123,E128,W504,W503 --exclude=docs,build` on my machine returns no errors. Please advise.

Tortar:
You need to apply black formatter to your changed file after pip installing it

woolgathering:
Whoop, yes done.

mesa/datacollection.py
```diff
@@ -151,28 +155,31 @@ def _new_table(self, table_name, table_columns):
...
-                return _prefix + reports
-
-        agent_records = map(get_reports, model.schedule.agents)
+        agent_records = map(
+            partial(self._get_reports, self, schedule.steps), schedule.agents
+        )
         return agent_records
 
+    @staticmethod
+    def _get_reports(collector, steps, agent):
```
rht, line 166: This method name is too generic. It could have meant getting the model report.

rht: It's not used elsewhere in the code. Why not put the function back in `_record_agents`?

rht:
The only reason we need to pass in `model` in `collect` is to access its scheduler. I wonder if we should make `model` an optional argument in `collect`.

mesa/datacollection.py
woolgathering: I wrongly assumed that by avoiding redefining the same function every call we'd get a performance improvement. This was not the case:

I used the introductory tutorial at ["Collecting Data"](https://mesa.readthedocs.io/en/latest/tutorials/intro_tutorial.html#collecting-data) and removed the `model_reporters`, then timed this:
```python
from timeit import default_timer as timer # not called while timing, just here to show what package I was using
start = timer()
model.datacollector.collect(model)
end = timer()
print(end - start) # time in seconds
```
For each variation:
```python
6.0040998505428433e-05 # original
0.00016220801626332104 # this commit
0.00014566699974238873 # leaving the function in _record_agents
4.683298175223172e-05 # leaving the function in _record_agents and removing the model argument in _record_agents that is no longer used
```

So actually I was mistaken but we can get it to be pretty fast by leaving the func inside and removing the unused arg. The functions in question for the last timed test are:
```python
def _record_agents(self, schedule):
    """Record agents data in a mapping of functions and agents."""

    def get_reports(agent):
        _prefix = (schedule.steps, agent.unique_id)
        reports = tuple(rep(agent) for rep in rep_funcs)
        return _prefix + reports

    agent_records = map(get_reports, schedule.agents)
    return agent_records

def collect(self, model):
    """Collect all the data for the given model object."""
    if self.schedule is None:
        schedule = model.schedule
    else:
        schedule = self.schedule

    if self.model_reporters:

        for var, reporter in self.model_reporters.items():
            # Check if Lambda operator
            # Check if model attribute
            elif isinstance(reporter, partial):
                self.model_vars[var].append(reporter[0](*reporter[1]))
            else:
                self.model_vars[var].append(self._reporter_decorator(reporter))

    if self.agent_reporters:
        agent_records = self._record_agents(schedule)
        self._agent_records[schedule.steps] = list(agent_records)
```

Thoughts?

woolgathering:
> The only reason we need to pass in `model` in `collect` is to access its scheduler. I wonder if we should make `model` an optional argument in `collect`.

I thought about this but I decided against it because that would negate what I was trying to do: namely, to have an arbitrary schedule be data-collected where a model can have several schedules.

mesa/datacollection.py
rht: I think having the function defined inside `_record_agents` is clearer. This way, you can have the name remain short, `get_reports`.

rht:
But you can still have a data collector `collect` receiving arbitrary schedule without `model`. Making it optional doesn't negate this capability.

woolgathering:
To make sure I'm understanding, we're suggesting that `collect` have two arguments: `model` (optional) and `schedule` (optional), correct?

I noticed that we pass `model` to the `reporter()` when we loop over the reporters. If `schedule` is defined in `collect`, we would just do `reporter(schedule.model)`. But if `model` is defined in `collect`, we then do `reporter(model)`, yes? I just want to make sure I understand before I start making commits.

I'm also unclear on why we would want to pass in a schedule to a datacollector at the `collect` stage. This to me would imply that someone would want to change the collected schedule while the model is running which seems weird. I would think that a datacollector is "tied" to a specific schedule with specific agents and that we wouldn't want to change it while the model is running, hence having the datacollector have a `schedule` attribute.

woolgathering:
> We could also just pass in a list of agents to the collect method and then collect over these agents. I am not sure anymore about the advantage of baking them in. 

This is nice: to be able to collect on a subset of agents (a representative sample, say). My question is then: what about model level reporters? Would `collect` accept a list of agents and a `model` on which to collect model level info?

> Expanding on this idea we could gain maximal flexibility by also providing the reporter functions per collect call.

Is there a reason one would want to change the reporter functions during the execution of a model? I also don't know enough about CS to know whether or not this would create performance issues. What're your thoughts?

Corvince:
> > We could also just pass in a list of agents to the collect method and then collect over these agents. I am not sure anymore about the advantage of baking them in. 
> 
> This is nice: to be able to collect on a subset of agents (a representative sample, say). My question is then: what about model level reporters? Would `collect` accept a list of agents and a `model` on which to collect model level info? I also can't think of an instance in which I'd want to have more than one model running in a simulation, though it's entirely possible that I just haven't encountered the use-case yet.
> 

Yeah, I didn't really think about model level parameters. I don't have an idea on the top of my head, but I don't think we need to consider multi model simulations. I don't think this is a common use case (although maybe one should think about it, maybe it is just something that was feasible until now).



> > Expanding on this idea we could gain maximal flexibility by also providing the reporter functions per collect call.
> 
> Is there a reason one would want to change the reporter functions during the execution of a model? I also don't know enough about CS to know whether or not this would create performance issues. What're your thoughts?

Na I don't think reporters would change, I was thinking in the direction that you could use different reporters for different sets of agents. ( So some reporters a, b for type A, some other reporters c, d for type B)

Corvince:
> > We could also just pass in a list of agents to the collect method and then collect over these agents. I am not sure anymore about the advantage of baking them in. 
> 
> This is nice: to be able to collect on a subset of agents (a representative sample, say). My question is then: what about model level reporters? Would `collect` accept a list of agents and a `model` on which to collect model level info? I also can't think of an instance in which I'd want to have more than one model running in a simulation, though it's entirely possible that I just haven't encountered the use-case yet.
> 

Yeah, I didn't really think about model level parameters. I don't have an idea on the top of my head, but I don't think we need to consider multi model simulations. I don't think this is a common use case (although maybe one should think about it, maybe it is just something that was feasible until now).



> > Expanding on this idea we could gain maximal flexibility by also providing the reporter functions per collect call.
> 
> Is there a reason one would want to change the reporter functions during the execution of a model? I also don't know enough about CS to know whether or not this would create performance issues. What're your thoughts?

Na I don't think reporters would change, I was thinking in the direction that you could use different reporters for different sets of agents. ( So some reporters a, b for type A, some other reporters c, d for type B)

woolgathering:
> Yeah, I didn't really think about model level parameters. I don't have an idea on the top of my head, but I don't think we need to consider multi model simulations. I don't think this is a common use case (although maybe one should think about it, maybe it is just something that was feasible until now).

Agreed. I suppose one new thing that would be possible would be the possibility of the same agents acting with different models in the same simulation. But even then I'm unsure how this could not be accomplished by just using different groups of agents.

> Na I don't think reporters would change, I was thinking in the direction that you could use different reporters for different sets of agents. ( So some reporters a, b for type A, some other reporters c, d for type B)

Makes sense. My way of solving this was to just have different schedules (which are then passed to different datacollectors, potentially) and mixing them when stepping through the model. I was able to make a simple subclass of the `BaseScheduler` that took in multiple schedules and mixed the agents in them. So sort of like `RandomActivation` but it could mix agents from multiple schedules. This achieves the same functionality as what you're suggesting if all the agents are the same type, if I'm understanding correctly.

I suppose I'm also weary of breaking changes which might be the case if we made it possible to send in a list of agents to collect data from and completely separated the `DataCollector` out. I'd leave that up to the existing contributors to decide if/when that is appropriate.

rht:
I forgot that `model` is needed in `collect` for the model reporters. Let's keep the scope of this PR small for now, i.e. to allow passing an explicit `schedule` object in `collect`. The use case is specifically when you have multiple schedulers in one mode..

rht:
> to allow passing an explicit schedule object in collect. 

This is not accurate. It should be "to allow associating a `DataCollector` object with a custom scheduler, in case a model has multiple schedulers"

woolgathering:
@rht Great, shall I commit with the changes above; i.e. keeping `get_reports` defined inside `_record_agents` and passing in `model` to `collect()`? I thought it was pretty clean overall and it ran pretty quick.

rht:
Yes, sounds good.

EwoutH:
Thanks for this effort! When you've processed the review comments, could you add a (few) usage examples?

Maybe we also can mention it in one of the tutorials.

woolgathering:
> Thanks for this effort! When you've processed the review comments, could you add a (few) usage examples?

Yes, no problem. Should that go in examples/ or shall I update some of the docs?

rht:
I think for the scope of this PR, adding a test case for a custom schedule object would be sufficient.

As for the doc/example, we don't have a documentation .rst that describes the complete features of the `DataCollector` yet. Having to make an example simulation code for a particular feature is overkill, considering that an example usually consists of multiple files. Maybe a short doc in https://github.com/projectmesa/mesa/tree/main/docs/useful-snippets would do for now. And we can defer this documentation to a separate PR.

woolgathering:
Looking at `test_datacollector.py` as an example, it looks like I'd need to edit the `initialize_data_collector()` function in the `Model` class since it would assign the DataCollector to `self.datacollector` in every case (i.e. overwrite if I called it multiple times for multiple collectors and schedules).

To be completely honest, I've never written a unit test in my life so I suppose I'm looking for some guidance here. Should I just write a simple script, semi-based on `test_datacollector.py` with a bunch of assertions?

EwoutH:
> Ended up just adding some sample code to the useful-snippets file.

Content wise this is perfect and really useful, especially since Mesa is used a lot as a teaching tool. Good, accessible documentation and examples are essential for that, so thanks!

The document doesn't seem to render correctly however, I think you need to add a white line after the `.. code:: python` part.

woolgathering:
Ah, good catch; added the newline. Let me know if anything else is amiss! Very happy to be contributing to this project.

EwoutH:
> Ah, good catch; added the newline.

It works, but you also need another one on [line 68](https://github.com/projectmesa/mesa/pull/1481/files#diff-bc80902ed9508d44a350d6dec73b22af7f4a5839857af38dcf2bfea8e53a3d2eR68).

woolgathering:
Bah, sorry. Way too many balls in the air right now.

docs/useful-snippets/snippets.rst
```diff
@@ -58,3 +58,50 @@ code as follows.
...
+  # simple schedule mixer
+  class ScheduleMixerRandom:
+      def __init__(self, schedules: list):
+          self.schedules = schedules
+
+      def step(self):
+          agents = []
+          for s in self.schedules:
+              agents.extend(s.agents)
+          agents = shuffle(agents) # step through them randomly
```
rht, line 80: I think one of the main reason one wants to have multiple schedulers is so that agents in each of the schedulers are executed separately, in groups. It would have driven the point better with
```python
for s in self.schedules:
    s.step()
```

rht: Oh, nvm, the name of the class itself says that it is a mixer.

That makes me wonder. For the use case for when you just need a `RandomActivation` for the model step, but want to do something complex with data collection, it is rather overkill to have multiple schedulers just for slicing the agents list. While it's not used in model step at all. Maybe it is cleaner after all, with just passing an agents list to the `DataCollector` initialization, as @Corvince said in https://github.com/projectmesa/mesa/pull/1481#issuecomment-1297470299?

rht: In any case, the example needs to encapsulate properly the use case that is covered by this new feature. Reconstructing the agents list from scratch for every model step seem slow to me.

rht: For this snippet, I would define `ScheduleMixerRandom` as
```python
def __init__(self, schedules):
    self._agents = []
    for s in schedules:
        self._agents.extends(s.agents)

def step(self):
    # do stuff with self._agents
    ...
```

woolgathering: > Maybe it is cleaner after all, with just passing an agents list to the DataCollector initialization, as @Corvince said in https://github.com/projectmesa/mesa/pull/1481#issuecomment-1297470299?

My off-the-cuff concern here would be if one wants to add/remove agents during the execution of the model. It would seem that if we passed a list at creation, we'd have to update both the schedule and the list in the `DataCollector` which feels sort of redundant and cumbersome. Also if we remove agents arbitrarily, we'd have to search the list in the `DataCollector` and find the right one to remove.

> In any case, the example needs to encapsulate properly the use case that is covered by this new feature. Reconstructing the agents list from scratch for every model step seem slow to me.

Yes, I wouldn't re-create the list each time in a real model, I'd probably do some sort of master list indexing or force interaction with the sub-schedules via the schedule mixer to ensure things update when necessary. My thought with the snippet was to show how this functionality is used in the dumbest possible way so there is no fancy coding or confusion as to what it is or how it works. Any optimization (especially of a custom class) or more elaborate uses would be the responsibility of the user/implementer and would not necessarily appear in a snippet.

rht: Yeah, the add/remove agent would have to be done redundantly, or it depends on the model. Though, the list search problem can be fixed if we use a dict.

There is a case if the main scheduler used by the model step is `RandomActivationByType`, which is a dict of dict. If the `DataCollector` accepts a custom dict, then you could pass in `model.schedule[agent_type]`. In this case, add/remove needs to be done once per agent death/birth event, since there is only 1 scheduler object.

> I wrote it the way I did to ensure the capture of any changes to the agent list like additions or subtractions.

OK, I can see how the implementation seeks to minimize error for the users who use the snippet. Ensuring correctness is indeed safer than performance. But at the same time, we could add 1 documenting sentence in `ScheduleMixerRandom` saying that the user needs to be mindful when adding/remove an agent.

woolgathering: Yeah, it would be a bit ugly in any way.

Updated. Let me know if there are any additional concerns, thoughts, etc.

docs/useful-snippets/snippets.rst
```diff
@@ -58,3 +58,56 @@ code as follows.
...
+  # simple schedule mixer
+  class ScheduleMixerRandom:
+      """
+      Note that ScheduleMixerRandom as written here would not capture changes to
+      the agent lists in the sub-schedules. The user would need to be mindful if
+      this functionality were to be included.
+      """
+
+      def __init__(self, schedules):
+          self._agents = []
```
rht, line 80: `BaseScheduler._agents` is a `dict[int, Agent]`, and so, this should have been a dict.

rht: Ah, nvm, because `BaseScheduler.agents` returns a `list[Agent]`. But removing an agent would be costly with a list.

The only way to handle the add/remove agent seems to be to aggregate  using private method `BaseScheduler._agents`.

docs/useful-snippets/snippets.rst
```diff
@@ -58,3 +58,56 @@ code as follows.
...
+One may occasionally want multiple schedules with distinct sets of agents and collect data on each of those schedules independently while allowing the agents to interact in the same model. This would be particularly useful if the agents in your model do not all have the same attributes and you want to collect data on them in groups. You can create a simple scheduler class to mix the agents in each schedule and data collect on each schedule independently.
+
+.. code:: python
+
+  from random import shuffle
+
+  # simple schedule mixer
+  class ScheduleMixerRandom:
+      """
+      Note that ScheduleMixerRandom as written here would not capture changes to
```
rht, line 74: Replacing "note that" with "be careful:" would better alarm me as a reader, of the consequence of not updating `ScheduleMixerRandom._agents` together with the schedules.

docs/useful-snippets/snippets.rst
rht: What do you think of this?
```python
self._agents = {}
for s in schedules:
    for a in s.agents:
        self._agents[a.unique_id] = a
```

woolgathering: Ya, let's do that. As long as it's clean for users to see what's going on, I'm ok with whatever we do. My `ScheduleMixerRandom` is just a prototype.

docs/useful-snippets/snippets.rst
woolgathering: Adjusted both.

docs/useful-snippets/snippets.rst
```diff
@@ -58,3 +58,57 @@ code as follows.
...
+      Be careful: ScheduleMixerRandom as written here would not capture changes to
+      the agent lists in the sub-schedules. The user would need to be mindful if
+      this functionality were to be included.
+      """
+
+      def __init__(self, schedules):
+          self._agents = {}
+          for s in schedules:
+          for a in s.agents:
+              self._agents[a.unique_id] = a # each agent must have a unique id across models
```
rht, line 83: Needs 2 spaces before the `#`. The code snippet should be formatted with Black before put in the snippets.rst.

docs/useful-snippets/snippets.rst
```diff
@@ -58,3 +58,57 @@ code as follows.
...
+          self.schedule_two = BaseScheduler()
+
+          # each DataCollector instance is associated with a specific schedule (and therefore its agents)
+          self.datacollector_one = DataCollector(agent_reporters = {'value_one': 'value_one'}, schedule = self.schedule_one)
+          self.datacollector_two = DataCollector(agent_reporters = {'value_two': 'value_two'}, schedule = self.schedule_two)
+
+          self.schedule_mixer = ScheduleMixerRandom(schedules=[self.schedule_one, self.schedule_two])
+
+      def step():
+          self.schedule_mixer.step() # calls step() on each agent in both schedules, mixed together
```
rht, line 102: To save extra review roundtrip: the comments should be in their own lines, so that the line doesn't exceed 88 chars length limit imposed by Black.

tpike3:
How close are we to a merge on this? I can use this to show snippets as I rerecord session 6

rht:
It's almost there. Just need to tidy the code snippet.

docs/useful-snippets/snippets.rst
jacob-thrackle: I ran Black but it said it didn't format RST files. Should I run just the Python through the formatter after pasting it into another file?

docs/useful-snippets/snippets.rst
jacob-thrackle: No problem.

jacob-thrackle:
Will have it done by this afternoon PST.

docs/useful-snippets/snippets.rst
EwoutH: > Should I run just the Python through the formatter after pasting it into another file?

Yes that works!

You can also try [blacken-docs](https://github.com/adamchainz/blacken-docs), that should format code blocks in Markdown and ReStructured Text.

docs/useful-snippets/tmp.py
```diff
@@ -0,0 +1,49 @@
+from random import shuffle
```
rht, line 1: This file is accidentally committed.

rht:
OK, pending the discussion result of #1506.

EwoutH:
Hi @woolgathering! @rht, @tpike3, @jackiekazil and I were discussed multiple agent types for the DataCollector. We discussed multiple options, including adding a type attribute to the Agent class (consistent, but conditional filtering would have take a large performance impact) and layering the DataCollector with a dict key for each agent type like in #1142 (not very scalable).

The problem to we converged to was that Mesa supports a single scheduler and a single datacollector by default. This PR adds multiple schedulers with multiple datacollectors. However, there is also a use case for a single scheduler (for example `RandomActivationByType`) but with multiple datacollectors.

Do you think this PR supports (or can be adapted to) that case of single scheduler with multiple data collectors, and if so would you be willing to add an additional example for that?

rht:
> Do you think this PR supports (or can be adapted to) that case of single scheduler with multiple data collectors, and if so would you be willing to add an additional example for that?

`RandomActivationByType.agents_by_type` can be considered to be a collection of schedulers (which keys are static), so I think it would work out of the box with the solution in this PR.

Use cases to consider:
- Currently, `RandomActivationByType` groups agents by their agent class, obtained via `type(agent)`. But sometimes, one might want to group by various agent attributes. It might be possible to have 2 agent classes sharing the same attribute, hence, grouped in the same schedule. Or the other way around: 2 agents with the same agent class, but have different attributes 
- An agent attributes may change over time, and so, the agent might be moved to a different `step` stage as the simulation progresses.

EwoutH:
Very good cases!

Aside from examples, how complex would it be to add tests for these cases?

rht:
> Aside from examples, how complex would it be to add tests for these cases?

I think we can do this in a separate PR, because it requires a scheduler that groups not by `type(agent)`. And once the 1 scheduler with multiple data collectors has been documented, it would require a little tweak to accommodate for the scheduler with custom grouping.

EwoutH:
@woolgathering, what do you think?

woolgathering:
I'm generally of the opinion that PRs should be fairly limited in scope so adding the test cases would probably be best in another PR. I'm honestly not sure how common those cases will be.

I was thinking about the single scheduler, multiple collector this weekend. It seems that the cleanest way would be to just pass a list to the collector of the agents one wants to track. The problem there, of course, is that this seems it would require relatively significant reworking of the DataCollector class and would complicate adding/removing agents. Given the way this fix is currently written, I can't think of a simple way of doing it.

Suppose I have the problem here: that I have 1000 agents but I only want to track 100 specific agents. This would be a case where one could put all the agents into a single scheduler, then only capture in the data collector the 100 desired agents. This is my understanding of the problem. However... the same thing could be accomplished using multiple schedulers mixed together as in the example I wrote in the snippets. This could even be adapted to the case where one wants a specific order (in a single scheduler, all agents have a `unique_id` and the mix could be ordered according to that), or to activate them by type (no different than having multiple schedulers with different types executed in series).

So I think overall I agree with @rht in that those cases could work out of the box and get the same end result, even if on the execution layer it looks different.

EwoutH:
Thanks for your extensive reply and the time thinking about this!

While I feel using multiple schedulers is not the most intuitive for users, it’s an improvement over the current situation when that means the datacollector can be used for each scheduler. If you and @rht think another example with that use case is out of the scope of this PR, I would say go ahead and merge it.

> How close are we to a merge on this? I can use this to show snippets as I rerecord session 6

@tpike3 this opportunity window has probably passed right?

rht:
I sketched an example for 3 data collectors, 1 scheduler (`RandomActivationByType`) in https://github.com/rht/mesa/tree/test_dc_schedule. It didn't work, because the data collector strictly assumes that the object being passed has an attribute `agents` (see https://github.com/rht/mesa/blob/dbb29de3decb350fac406ba985fb846e14597960/mesa/datacollection.py#L167).

woolgathering:
Hmmm, I wonder if it would be best to be able to pass a list of agents to the DataCollector after all... then if it's `None` it would just default to `schedule.agents` (all the agents). It would then be on the user to make sure that if they're adding/removing agents to do it in both places.

That may not be as hard, though, especially if the list were passed on each `collect`. It just feels kind of cumbersome and bloated. Thoughts?

rht:
It could as well be a `dict` instead of a `list`, or maybe have `DataCollector.__getitem__`, not sure.

I think for now, to solve the Sugarscape G1MT for the complexity tutorial use case (1 scheduler, multiple data collectors), we should just implement a custom filter. The mental model with the custom filter is much simpler. Also, it's only 200 agents at most, and so should be fine. We can have a more performant option later.

rht:
> The mental model with the custom filter is much simpler.

I should add, that this is convenient for rapid prototyping a model, before scaling it up.

Simpdaddyjosh (approved):

tpike3:
Inspired other updates to datacollector

PR #1482, opened 2022-10-29
Refactor Grid.move_to_empty

Tortar:
The previous implementation was a bit strange, it used the number of agents parameter in the if branch, but I don't think it's the relevant parameter, the one to use should be the number of empty cells so that you can find out the density `num_empty_cell / number_of_points_of_the_grid` to choose one branch or the other. Don't know if I'm wrong on something. Maybe we should warn on the number of agents parameter being deprecated and leave it as a parameter to remove in the future or not?

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -56,11 +56,6 @@ MultiGridContent = List[Agent]
 F = TypeVar("F", bound=Callable[..., Any])
 
 
-def clamp(x: float, lowest: float, highest: float) -> float:
-    # much faster than np.clip for a scalar x.
-    return lowest if x <= lowest else (highest if x >= highest else x)
-
-
 def accept_tuple_argument(wrapped_function: F) -> F:
     """Decorator to allow grid methods that take a list of (x, y) coord tuples
     to also handle a single position, by automatically wrapping tuple in
@@ -427,7 +422,7 @@ class Grid:
             warn(
                 (
                     "`num_agents` is being deprecated since it's no longer used "
-                    "inside the function, it shouldn't be passed as a parameter."
+                    "inside `move_to_empty`. It shouldn't be passed as a parameter."
                 ),
                 DeprecationWarning,
             )
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -103,6 +103,7 @@ class Grid:
         self.height = height
         self.width = width
         self.torus = torus
+        self.num_cells = height * width
 
         self.grid: list[list[GridContent]]
         self.grid = [
@@ -422,32 +423,24 @@ class Grid:
         self, agent: Agent, cutoff: float = 0.998, num_agents: int | None = None
     ) -> None:
         """Moves agent to a random empty cell, vacating agent's old cell."""
-        if len(self.empties) == 0:
+        if num_agents is not None:
+            warn(
+                (
+                    "`num_agents` is being deprecated since it's no longer used "
+                    "inside the function, it shouldn't be passed as a parameter."
+                ),
+                DeprecationWarning,
+            )
+        num_empty_cells = len(self.empties)
+        if num_empty_cells == 0:
             raise Exception("ERROR: No empty cells")
-        if num_agents is None:
-            try:
-                num_agents = agent.model.schedule.get_agent_count()
-            except AttributeError:
-                raise Exception(
-                    "Your agent is not attached to a model, and so Mesa is unable\n"
-                    "to figure out the total number of agents you have created.\n"
-                    "This number is required in order to calculate the threshold\n"
-                    "for using a much faster algorithm to find an empty cell.\n"
-                    "In this case, you must specify `num_agents`."
-                )
-        new_pos = (0, 0)  # Initialize it with a starting value.
-        # This method is based on Agents.jl's random_empty() implementation.
-        # See https://github.com/JuliaDynamics/Agents.jl/pull/541.
-        # For the discussion, see
-        # https://github.com/projectmesa/mesa/issues/1052.
-        # This switch assumes the worst case (for this algorithm) of one
-        # agent per position, which is not true in general but is appropriate
-        # here.
-        if clamp(num_agents / (self.width * self.height), 0.0, 1.0) < cutoff:
-            # The default cutoff value provided is the break-even comparison
-            # with the time taken in the else branching point.
-            # The number is measured to be 0.998 in Agents.jl, but since Mesa
-            # run under different environment, the number is different here.
+
+        # This method is based on Agents.jl's random_empty() implementation. See
+        # https://github.com/JuliaDynamics/Agents.jl/pull/541. For the discussion, see
+        # https://github.com/projectmesa/mesa/issues/1052. The default cutoff value
+        # provided is the break-even comparison with the time taken in the else
+        # branching point.
+        if 1 - num_empty_cells / self.num_cells < cutoff:
             while True:
                 new_pos = (
                     agent.random.randrange(self.width),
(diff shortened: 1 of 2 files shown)


Tortar:
Right now I realized that `self.empties` is something we shouldn't want to mantain but I don't know about how much this could break because it is called each time an agent moves and it's a bit too much in my opinion

rht:
`self.empties` is going to be deprecated.

mesa/space.py
```diff
@@ -418,41 +419,23 @@ def is_cell_empty(self, pos: Coordinate) -> bool:
...
-            # The number is measured to be 0.998 in Agents.jl, but since Mesa
-            # run under different environment, the number is different here.
+
+        # This method is based on Agents.jl's random_empty() implementation. See
+        # https://github.com/JuliaDynamics/Agents.jl/pull/541. For the discussion, see
+        # https://github.com/projectmesa/mesa/issues/1052. The default cutoff value
+        # provided is the break-even comparison with the time taken in the else
+        # branching point.
+        cutoff: float = 0.998
+        if num_empty_cells / self.num_cells < cutoff:
```
rht, line 434: Shouldn't this be `1 - num_empty_cells / num_cells`? What about the `MultiGrid` case?

rht: The original Agents.jl implementation uses number of agents, but I agree that `num_empty_cells` is more accurate.

mesa/space.py
```diff
@@ -418,41 +419,23 @@ def is_cell_empty(self, pos: Coordinate) -> bool:
...
             while True:
-                new_pos = (
-                    agent.random.randrange(self.width),
-                    agent.random.randrange(self.height),
-                )
+                rand_x = agent.random.randrange(self.width)
```
rht, line 436: Why is this code refactored? I don't see how the newer code is clearer.

mesa/space.py
Tortar: Don't think the MultiGrid case it's a problem because we are looking for the number of empty cells, it isn't important the number of agent per cell

Tortar: > Shouldn't this be `1 - num_empty_cells / num_cells`? What about the `MultiGrid` case?
yep, right It was late yesterday night :-)

mesa/space.py
```diff
@@ -418,36 +419,19 @@ def is_cell_empty(self, pos: Coordinate) -> bool:
...
-            # with the time taken in the else branching point.
-            # The number is measured to be 0.998 in Agents.jl, but since Mesa
-            # run under different environment, the number is different here.
+
+        # This method is based on Agents.jl's random_empty() implementation. See
+        # https://github.com/JuliaDynamics/Agents.jl/pull/541. For the discussion, see
+        # https://github.com/projectmesa/mesa/issues/1052. The default cutoff value
+        # provided is the break-even comparison with the time taken in the else
+        # branching point.
+        cutoff: float = 0.998
```
rht: I think I haven't fully calibrated this number for Mesa/Python case. https://github.com/JuliaDynamics/Agents.jl/pull/541#discussion_r687563557

If you keep this as an optional argument, it will be easier to benchmark.

The break even for Mesa seems to depend on the performance of `sorted`. This could be hardware dependent?

rht: Just in case; the cutoff could be different depending on the environment, or whether the code is run on PyPy.

rht: @Tortar ^

Tortar: yes, we should measure in another PR, and the hardware dependency can be a problem, I'll try to investigate that. For now I apply the suggestion to leave it as a parameter

Tortar: Think it can be merged

mesa/space.py
```diff
@@ -418,36 +419,18 @@ def is_cell_empty(self, pos: Coordinate) -> bool:
         x, y = pos
         return self.grid[x][y] == self.default_val()
 
-    def move_to_empty(
-        self, agent: Agent, cutoff: float = 0.998, num_agents: int | None = None
-    ) -> None:
+    def move_to_empty(self, agent: Agent, cutoff: float = 0.002) -> None:
```
rht, line 422: While the new cutoff definition is easier to understand (with a number close to epsilon), this breaks backward compatibility, by changing the definition of the `cutoff` argument.

We should defer this breaking change in a separate PR. In that PR, it might be safer to rename the argument to `threshold` instead of cutoff.

Tortar: ok, will open soon

rht: Actually, the fact that `num_agents` is removed already makes it backward incompatible. This is a tricky decision to make. I think we should keep this PR as 100% refactor, and so, the `num_agents` argument should remain present, even though it is not going to be used.

rht: There should be a warning shown to the user when `num_agents` is not `None`.

mesa/space.py
```diff
@@ -419,8 +419,12 @@ def is_cell_empty(self, pos: Coordinate) -> bool:
         x, y = pos
         return self.grid[x][y] == self.default_val()
 
-    def move_to_empty(self, agent: Agent, cutoff: float = 0.998) -> None:
+    def move_to_empty(
+        self, agent: Agent, cutoff: float = 0.998, num_agents: int | None = None
+    ) -> None:
         """Moves agent to a random empty cell, vacating agent's old cell."""
+        if num_agents is not None:
+            warn("num_agents is being deprecated since it's no more necessary", DeprecationWarning)
```
rht: Should be "no longer necessary" instead.

Tortar: done

Corvince:
> Right now I realized that `self.empties` is something we shouldn't want to mantain (but I don't know about how much this could break things done in the past) because it is used each time an agent moves and it's a bit too much in my opinion for how much it is useful, we cah just have `num_empty_cell` in all the library I guess

This is a tricky one. The usefulness depends on the question of wether the model use it. Currently you pay a constant cost for each move. If you never use the empties set this is a unnecessary cost. However,without it, if you do use the set (by regularly moving to an empty cell) you would have to create the empties list on each function call. And the creation scales with grid size, i.e. with larger grids the cost scales quadratically.

So right now some models pay a linear cost, but the alternative would be for some models to pay a quadratic or even exponential cost.

Tortar:
agree but at the least we need to make it optional with an user settable bool attribute for the Grid instance as `track_empties`

Tortar:
This way we have just a minor extra cost for all the models without any loss of features

mesa/space.py
```diff
@@ -422,32 +423,24 @@ def move_to_empty(
         self, agent: Agent, cutoff: float = 0.998, num_agents: int | None = None
     ) -> None:
         """Moves agent to a random empty cell, vacating agent's old cell."""
-        if len(self.empties) == 0:
+        if num_agents is not None:
+            warn(
+                (
+                    "`num_agents` is being deprecated since it's no longer used "
+                    "inside the function, it shouldn't be passed as a parameter."
```
rht, line 430: One last comment: the comma here should be full stop or semicolon. Everything else LGTM.

rht: Also, you should say "`move_to_empty`" instead of "the function."

mesa/space.py
```diff
@@ -422,32 +423,24 @@ def move_to_empty(
...
-                )
-        new_pos = (0, 0)  # Initialize it with a starting value.
-        # This method is based on Agents.jl's random_empty() implementation.
-        # See https://github.com/JuliaDynamics/Agents.jl/pull/541.
-        # For the discussion, see
-        # https://github.com/projectmesa/mesa/issues/1052.
-        # This switch assumes the worst case (for this algorithm) of one
-        # agent per position, which is not true in general but is appropriate
-        # here.
-        if clamp(num_agents / (self.width * self.height), 0.0, 1.0) < cutoff:
```
rht, line 446: I just noticed that since `clamp` is no longer used, its function definition can be deleted.

rht:
Content LGTM, and is not controversial that I assume @Corvince is fine with.

PR #1489, opened 2022-10-30
Fix accident from https://github.com/projectmesa/mesa/pull/1488

rht:
#1488 was accidentally merged, because I thought I was making a PR to github.com/rht/mesa. I was able to change the settings, and hence made me think that it was rht/mesa.

The correct step should be to not force-push the main branch (rewrite history).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,5 +1,4 @@
 Mesa: Agent-based modeling in Python 3+
-squash
 =========================================
 
 .. image:: https://github.com/projectmesa/mesa/workflows/build/badge.svg


rht:
We should enable GH config where a maintainer can't merge their own PR without an approval from the other maintainer: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches#require-pull-request-reviews-before-merging.

wang-boyu:
Don't worry. Things like this always happen : )

As an alternative, shall we simply set the `main` branch as a protected branch, disallowing forced push? Not sure whether this was already done or not.

Ref: [About protected branches](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches)

rht:
@jackiekazil ^ for disallowing force push to `main` branch for both Mesa and Mesa-Geo.

tpike3:
@jackiekazil ^ for disallowing force push to main branch for both Mesa and Mesa-Geo.

Corvince:
I made `main` a protected branch on Mesa, but I don't have the same access rights on Mesa-geo, so...

@jackiekazil ^ for disallowing force push to main branch for both Mesa-Geo.

jackiekazil:
@Corvince i made mesa-geo to match. 
@tpike3 - can you add permissions discussion to next month's mesa discussion?  We need to add some logic as to who can do these things and where?

PR #1491, opened 2022-10-30
SingleGrid: Remove extraneous attribute declaration (empties)

Tortar:


rht:
Merged after rephrasing the commit message and PR title. Because technically, `SingleGrid.empties` is still being used.

PR #1492, opened 2022-10-31
Use built-in function for setting default_val

Tortar:
Setting default value with the type function makes faster start-up initialization for Grid and MultiGrid and everytime the attribute is called, for example timing it with

```
a = min(timeit.repeat("Grid(100,100,True)", "from mesa.space import Grid", repeat=20, number=20))
b = min(timeit.repeat("Grid(100,100,True)", "from mesa.space_2 import Grid", repeat=20, number=20))
print(a/b)
```

where `space_2` contains the modified code, gives me a ~30% less for the new way, also it is ~60% faster when calling it alone:

```
a = min(timeit.repeat("grid.default_val()", "from mesa.space import Grid\ngrid = Grid(5,5,True)", repeat=20, number=5000))
b = min(timeit.repeat("grid._default_val()", "from mesa.space_2 import Grid\ngrid = Grid(5,5,True)", repeat=20, number=5000))
print(a/b)
```

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -99,10 +99,11 @@ class Grid:
         self.width = width
         self.torus = torus
         self.num_cells = height * width
+        self._default_val = type(self.default_val())
 
         self.grid: list[list[GridContent]]
         self.grid = [
-            [self.default_val() for _ in range(self.height)] for _ in range(self.width)
+            [self._default_val() for _ in range(self.height)] for _ in range(self.width)
         ]
 
         # Add all cells to the empties list.
@@ -405,14 +406,14 @@ class Grid:
         if (pos := agent.pos) is None:
             return
         x, y = pos
-        self.grid[x][y] = self.default_val()
+        self.grid[x][y] = self._default_val()
         self.empties.add(pos)
         agent.pos = None
 
     def is_cell_empty(self, pos: Coordinate) -> bool:
         """Returns a bool of the contents of a cell."""
         x, y = pos
-        return self.grid[x][y] == self.default_val()
+        return self.grid[x][y] == self._default_val()
 
     def move_to_empty(
         self, agent: Agent, cutoff: float = 0.998, num_agents: int | None = None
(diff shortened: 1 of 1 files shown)


Tortar:
Mhh alternatively we could mantain the default_value function so that you can do even what you did before, and use the new way just internally

Tortar:
It doesn't work for non empty default val

rht:
I'd say using Numba is most likely to give a huge speedup for space.py, while at the same time keeping the code readable.

Tortar:
I tried to make some tests with Numba with rather disappoint results actually:

define two functions `a` and `b`, one with the njit decorator and the other a normal one which fill a set given as input and then compare times:

```
@njit
def a(set_to_fill):
    for x in range(10): set_to_fill.add(x)
    return set_to_fill

def b(set_to_fill):
    for x in range(10): set_to_fill.add(x)
    return set_to_fill

t_1 = timeit.timeit("a(set(1))", "from numba import njit\nfrom __main__ import a") # --> 4 sec
t_2 = timeit.timeit("b(set([1]))", "from __main__ import b")  # --> 1 sec
print(t_1, t_2)
```

so every operation with sets will result much slower

rht:
```python
import timeit
from numba import njit

@njit(cache=True)
def a(set_to_fill):
    for x in range(10): set_to_fill.add(x)
    return set_to_fill

def b(set_to_fill):
    for x in range(10): set_to_fill.add(x)
    return set_to_fill

t_1 = timeit.timeit("a(set(range(100)))", "from numba import njit\nfrom __main__ import a\na(set([1]))") # --> 4 sec
t_2 = timeit.timeit("b(set(range(100)))", "from __main__ import b")  # --> 1 sec
print(t_1, t_2)
```
> 17.71975239997846 12.584562428004574

Still slower. But Numba doc says its reflected set is going to be deprecated anyway: https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-reflection-for-list-and-set-types.

PR #1493, opened 2022-10-31
Remove extraneous spaces from docstrings in modules

Tortar:
Need to do for the other modules before merging


Tortar:
Not sure actually if the space is good or not

rht:
Yeah, it's fine to remove them.

PR #1497, opened 2022-11-01
Add default_value function to NetworkGrid

Tortar:
Consistent with the other Grids class

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -955,7 +955,12 @@ class NetworkGrid:
     def __init__(self, G: Any) -> None:
         self.G = G
         for node_id in self.G.nodes:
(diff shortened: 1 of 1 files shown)


mesa/space.py
```diff
@@ -955,7 +955,12 @@ class NetworkGrid:
     def __init__(self, G: Any) -> None:
         self.G = G
         for node_id in self.G.nodes:
-            G.nodes[node_id]["agent"] = list()
+            G.nodes[node_id]["agent"] = self.default_val()
+
+    @staticmethod
+    def default_val() -> list:
+        """Default value for new node."""
```
rht, line 962: Should be "for a new node"

rht:
LGTM

PR #1498, opened 2022-11-01
Refactor NetworkGrid docstrings and iter/get_cell_list_contents

Tortar:
I think the docstrings are better this way, any better wording is appreciated :-) also it's clear that there are right now two redundant methods `iter_cell_list_contents` and `get_cell_list_contents` so one should be made really an iterator, will do after this one is merged (hopefully)

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -986,8 +986,8 @@ class NetworkGrid:
         return not self.G.nodes[node_id]["agent"]
 
     def get_cell_list_contents(self, cell_list: list[int]) -> list[GridContent]:
-        """Returns a list of the agents in the nodes identified in `cell_list`;
-        nodes with empty content are excluded.
+        """Returns a list of the agents contained in the nodes identified
+        in `cell_list`; nodes with empty content are excluded.
         """
         list_of_lists = [
             self.G.nodes[node_id]["agent"]
@@ -1001,5 +1001,7 @@ class NetworkGrid:
         return self.get_cell_list_contents(self.G)
 
     def iter_cell_list_contents(self, cell_list: list[int]) -> Iterator[GridContent]:
-        """Returns an iterator of the agents in the nodes identified in `cell_list`."""
+        """Returns an iterator of the agents contained in the nodes identified
+        in `cell_list`.
+        """
         yield from self.get_cell_list_contents(cell_list)
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -1002,6 +1002,6 @@ class NetworkGrid:
 
     def iter_cell_list_contents(self, cell_list: list[int]) -> Iterator[GridContent]:
         """Returns an iterator of the agents contained in the nodes identified
-        in `cell_list`.
+        in `cell_list`; nodes with empty content are excluded.
         """
         yield from self.get_cell_list_contents(cell_list)
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -958,14 +958,12 @@ class NetworkGrid:
             G.nodes[node_id]["agent"] = list()
 
     def place_agent(self, agent: Agent, node_id: int) -> None:
-        """Place a agent in a node."""
-
+        """Place an agent in a node."""
         self.G.nodes[node_id]["agent"].append(agent)
         agent.pos = node_id
 
     def get_neighbors(self, node_id: int, include_center: bool = False) -> list[int]:
         """Get all adjacent nodes"""
-
         neighbors = list(self.G.neighbors(node_id))
         if include_center:
             neighbors.append(node_id)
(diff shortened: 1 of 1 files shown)


Tortar:
I think I will evolve this PR in a refactor of NetworkGrid taking into account the consideration made in the first comment

mesa/space.py
```diff
@@ -989,22 +986,20 @@ def is_cell_empty(self, node_id: int) -> bool:
         return not self.G.nodes[node_id]["agent"]
 
     def get_cell_list_contents(self, cell_list: list[int]) -> list[GridContent]:
-        """Returns the contents of a list of cells ((x,y) tuples)
-        Note: this method returns a list of `Agent`'s; `None` contents are excluded.
+        """Returns the agents in a list of nodes ((x,y) tuples)
+        Note: this method returns a list of `Agent`.
```
rht, line 990: I haven't checked, but isn't the old version more correct? IIRC excluding `None`'s is a general behavior of `get_cell_list_contents` in various grid implementations.

Tortar: the problem is that from what I understand it checks it checks if the list is empty, there is no None default value in NetworkGrid, this would be confusing, and it is clear I think that if a list is empty it doesn't have any agent so no need to specify that

rht: This docstring seems to be copy-pasted from `Grid.get_cell_list_contents`. It shouldn't have an `(x, y)` coordinate either, but a node ID instead. But at least, it should say nodes with empty content are excluded -- if this is excluded, it would be confusing.

rht: I meant, if the statement that "nodes with empty content are excluded" is not said in the docstring, it would be confusing.

Tortar: good catch! done

mesa/space.py
```diff
@@ -989,22 +986,20 @@ def is_cell_empty(self, node_id: int) -> bool:
...
             for node_id in cell_list
             if not self.is_cell_empty(node_id)
         ]
         return [item for sublist in list_of_lists for item in sublist]
+
+    def get_all_cell_contents(self) -> list[GridContent]:
+        """Returns a list of all the agents in the network."""
+        return self.get_cell_list_contents(self.G)
+
+    def iter_cell_list_contents(self, cell_list: list[int]) -> Iterator[GridContent]:
```
rht, line 1003: Given that falsy contents are not included, the annotation should have been `Iterator[Agent]`. And in other cases it should have been `Iterator[MultiGridContent]`, because technically `NetworkGrid` is a multi grid, due to its content being a list. But this should be fixed in a separate PR.

Tortar: I agree to address all of this in a separate PR,  good to have a reminder here, there is something else to do to merge this one instead? :-)

mesa/space.py
```diff
@@ -989,22 +986,20 @@ def is_cell_empty(self, node_id: int) -> bool:
         return not self.G.nodes[node_id]["agent"]
 
     def get_cell_list_contents(self, cell_list: list[int]) -> list[GridContent]:
-        """Returns the contents of a list of cells ((x,y) tuples)
-        Note: this method returns a list of `Agent`'s; `None` contents are excluded.
+        """Returns the agents in a list of nodes; nodes with empty content are excluded.
```
rht, line 989: "the agents in a list of nodes" sounds ambiguous to me. It could mean 1. agents that are taken from the list of nodes, 2. agents, in the form of a list of nodes. Additionally, the argument name is `cell_list`, but the docstring says a list of nodes. I think saying cell contents instead of agents is fine, given that the 2nd sentence further reinforces that it is a list of `Agent`'s.

rht: The 2nd sentence could be removed once the return type annotation has been changed to `list[Agent]`.

Tortar: What about this wording? It seems to me better than the previous version and I think it addresses all your points. It seems to me that also, the second sentence can be removed the same even with the wrong return type like this

Tortar: I will lint as soon as I can do it

mesa/space.py
```diff
@@ -989,22 +986,20 @@ def is_cell_empty(self, node_id: int) -> bool:
         return not self.G.nodes[node_id]["agent"]
 
     def get_cell_list_contents(self, cell_list: list[int]) -> list[GridContent]:
-        """Returns the contents of a list of cells ((x,y) tuples)
-        Note: this method returns a list of `Agent`'s; `None` contents are excluded.
+        """Returns a list of the agents in the nodes identified in `cell_list`;
```
rht, line 989: It's definitely better. Just to be sure, I'd say "returns a list of the agents contained in the nodes identified in `cell_list`".

Tortar: perfect, done

rht:
LGTM

PR #1504, opened 2022-11-03
Hexgrid: use get_neighborhood in iter_neighbors

Tortar:
As in the other Grids class Hexgrid.iter_neighbors should use get_neighborhood because more efficient


rht:
LGTM

PR #1507, opened 2022-11-05
Make swap_pos part of Grid instead of SingleGrid

Tortar:
Making `swap_pos`  part of Grid so that it can be used also by MultiGrid. Will add some tests for the MultiGrid case.

### DIFF
--- a/tests/test_grid.py
+++ b/tests/test_grid.py
@@ -169,6 +169,36 @@ class TestBaseGrid(unittest.TestCase):
         assert agent.pos is None
         assert self.grid.grid[x][y] is None
 
+    def test_swap_pos(self):
+
+        # Swap agents positions
+        agent_a, agent_b = random.sample(list(self.grid), k=2)
+        pos_a = agent_a.pos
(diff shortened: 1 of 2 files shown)


Tortar:
I actually just put the tests in TestBaseGrid, I think it's ok because many function are not tested in the subclasses, and because  `swap_pos`  uses function present both in SingleGrid and in MultiGrid all should safely work

rht:
cc: @jnsp

tests/test_grid.py
```diff
@@ -169,6 +169,36 @@ def test_agent_remove(self):
         assert agent.pos is None
         assert self.grid.grid[x][y] is None
 
+    def test_swap_pos(self):
+
+        # Swap agents positions
+        agent_a, agent_b = list(filter(None, self.grid))[:2]
```
rht, line 175: Note: the random sampling failed in the first because when the test is moved to `TestBaseGrid`, the `self.grid` in `TestBaseGrid` is not full.

rht:
LGTM

PR #1508, opened 2022-11-06
Make MultiGrid.place_agent faster

Tortar:
The condition tested now is a constant time operation while before it was a linear time operation in relation to the size of the list of agents

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -552,7 +552,7 @@ class MultiGrid(Grid):
         if agent.pos != pos:
             self.grid[x][y].append(agent)
             agent.pos = pos
-            self.empties.discard(pos)
+        self.empties.discard(pos)
 
     def remove_agent(self, agent: Agent) -> None:
         """Remove the agent from the given location and set its pos attribute to None."""
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -549,7 +549,7 @@ class MultiGrid(Grid):
     def place_agent(self, agent: Agent, pos: Coordinate) -> None:
         """Place the agent at the specified location, and set its pos variable."""
         x, y = pos
-        if agent.pos != pos:
+        if agent.pos != pos or agent.pos not in self.grid[x][y]:
             self.grid[x][y].append(agent)
             agent.pos = pos
             self.empties.discard(pos)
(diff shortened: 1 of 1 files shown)


Tortar:
Need to resort to a somewhat in between implementation :any time the agent.pos position is not preassigned to where you want to place the agent in the grid it takes a constant amount of time, otherwise it goes to scan the list

mesa/space.py
```diff
@@ -549,10 +549,10 @@ def default_val() -> MultiGridContent:
     def place_agent(self, agent: Agent, pos: Coordinate) -> None:
         """Place the agent at the specified location, and set its pos variable."""
         x, y = pos
-        if agent not in self.grid[x][y]:
+        if agent.pos != pos or agent not in self.grid[x][y]:
```
rht, line 552: This needs to be `agent.pos is None`. The case when `agent.pos` is something that is not `None`, but is not `pos` either, means something is wrong. The agent might exist in both `self.grid[x][y]` and somewhere else.

Corvince: Good catch! I was thinking something similar, where agent.pos is simply set to something different, but dismissed it as unlikely. But agents being on grid and another space (e.g. network) is something I see happening. And in that case the agent.pos might as well be set to None by the network. 

I think the current implementation is the most defensive mechanism, not a bad thing.

Tortar: very good points but what you@are saying @Corvince make me think about something to resolve ambiguities, in a sense as you point out it doesn't make much sense to have a pos variable shared by all spaces, we should create a pos variable for each agent related to each space to which the agent belongs!

rht: By somewhere else, I meant `self.grid[x_other][y_other]`. `agent.pos` is not `None` happens when the modeler calls consecutive `place_agent`'s without ensuring that an agent is removed from the grid before a `place_agent` happens.

Tortar: yes, indeed this means that the current implementation is prone to errors

EwoutH:
I'm curious, with Python, in a conditional `or` statement like:
```python
if agent.pos is None or agent not in self.grid[x][y]:
```If the first condition proves false, does Python still evaluate the second condition? Or does it continue directly without evaluating it?

Tortar:
@EwoutH i think you meant "if the first condition proves **true**" yes it doesn't evaluate the second one

EwoutH:
I meant that indeed. That also means that the order of the or statement matters. That's interesting, thanks!

rht:
LGTM

PR #1509, opened 2022-11-06
fix: position_agent: Enforce type of x and y

rht:


examples/schelling/model.py
```diff
@@ -70,7 +70,7 @@ def __init__(self, width=20, height=20, density=0.8, minority_pc=0.2, homophily=
                     agent_type = 0
 
                 agent = SchellingAgent((x, y), self, agent_type)
-                self.grid.position_agent(agent, (x, y))
+                self.grid.position_agent(agent, x, y)
```
rht, line 73: This is a bug where in the Schelling example, the number of agents in the scheduler does not match the number of agents in the space.

Tortar:
It seems to me that `position_agent` should be deprecater because everything can be done just by `place_agent` and `move_to_empty`, to me it doesn't make much sense to have also `position_agent`

rht:
> It seems to me that `position_agent` should be deprecated because everything can be done just by `place_agent` and `move_to_empty`, to me it doesn't make much sense to have also `position_agent`

Yes, I wanted to say this but forgot to send the message.

tpike3:
LGTM... glad we Kaunta identified, thanks for fixing @rht

PR #1512, opened 2022-11-06
Deprecate SingleGrid.position_agent

Tortar:
As noted in #1509 `SingleGrid.position_agent` should be deprecated in favor of `place_agent` and `move_to_empty`

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -504,6 +504,16 @@ class SingleGrid(Grid):
         If x or y are positive, they are used.
         Use 'swap_pos()' to swap agents positions.
         """
+        warn(
+            (
+                "`position_agent` is being deprecated; use instead "
+                "`place_agent` to place an agent at a specified "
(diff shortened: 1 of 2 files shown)


mesa/space.py
```diff
@@ -504,6 +504,14 @@ def position_agent(
         If x or y are positive, they are used.
         Use 'swap_pos()' to swap agents positions.
         """
+        warn(
+            (
+                "`position_agent` is being deprecated in favor of "
+                "`place_agent` and `move_to_empty`."
```
rht, line 510: I think the previous warning message describing when to use `place_agent` and `move_to_empty` is better.

Tortar: new way :-)

rht:
The `position_agent` in the Schelling example should be replaced with `place_agent`.

rht:
There is a merge conflict.

rht:
LGTM

PR #1514, opened 2022-11-07
Update NetworkGrid.__init__ docstring

Tortar:
It should be specified that NetworkGrid requires a Networkx  graph instance.


rht:
LGTM

PR #1515, opened 2022-11-07
Remove key check in BaseScheduler.agent_buffer

Tortar:
It is not necessary to check the presence of the key

### DIFF
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -102,8 +102,7 @@ class BaseScheduler:
             self.model.random.shuffle(agent_keys)
 
         for key in agent_keys:
-            if key in self._agents:
-                yield self._agents[key]
+            yield self._agents[key]
 
 
 class RandomActivation(BaseScheduler):


Corvince:
On mobile so I can't check if I'm thinking correctly, but I think we check again in case an agent was removed in another agents step (e.g. killing)

Tortar:
I thought this was the meaning bit it isn't possible to do so, it would result in an error because .keys cannot be iterated while removing or adding

Tortar:
mmh yes but the problem is that it doesn't change the fact that you can't add agents during stepping anyway

Tortar:
right I think it's my fault :-) seems like we should revert it

Tortar:
I think that apart this bug there are some other bugs related to this inside some more advanced schedulers not checking the condition

Tortar:
I think that a comment stating  why the list transformation is necessary should be added

PR #1516, opened 2022-11-07
Revert changes of #1478 and #1456

Tortar:
Reverts projectmesa/mesa#1478 and parts of projectmesa/mesa#1456 because they introduce the bug discussed in #1515

### DIFF
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -139,7 +139,7 @@ class SimultaneousActivation(BaseScheduler):
     def step(self) -> None:
         """Step all agents, then advance them."""
         # To be able to remove and/or add agents during stepping
-        # it's necessary to cast the keys view to a list
+        # it's necessary to cast the keys view to a list.
         agent_keys = list(self._agents.keys())
         for agent_key in agent_keys:
             self._agents[agent_key].step()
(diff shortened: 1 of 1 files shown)


Corvince:
Thanks for the work. Would be awesome if we would have a test case for this, esp. the removing-while-running part. Do you think you could add those as well? No worries if you don't have time for that, we don't have them now so this is not a requirement for this PR to be merged

Tortar:
Will do that for sure, but right now I can't unfortunately

mesa/time.py
```diff
@@ -185,18 +186,16 @@ def __init__(
...
-        agent_keys = self._agents.keys()
+        agent_keys = list(self._agents.keys())
         if self.shuffle:
-            agent_keys = list(agent_keys)
             self.model.random.shuffle(agent_keys)
         for stage in self.stage_list:
             for agent_key in agent_keys:
                 getattr(self._agents[agent_key], stage)()  # Run stage
             # We recompute the keys because some agents might have been removed
             # in the previous loop.
```
rht, line 196: Needs to document why the casting to `list` is necessary.

Tortar: done

rht:
LGTM

PR #1517, opened 2022-11-08
Fix bug in Grid.get_neighborhood

Tortar:
One (hopefully last) bug was left unnoticed before merging, if a dimension is even and the radius is the maximum one possible, then we introduce duplicates in the list, consider the case of the width dimension equal to 4, and the user passes a value of 2 of radius and a x pos equal to 2 then we have a range(0, 5) but this means that we have two times the 0 position (0 % 4 and 4 % 4) . 

The way it is written now is not so good, hope to find a better way (a function for all the initial range conditions in the torus case?).

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -265,7 +265,9 @@ class Grid:
 
             # For each dimension, in the edge case where the radius is as big as
             # possible and the dimension is even we need to shrink by one the range
-            # of values to avoid duplicates in neighborhood
+            # of values to avoid duplicates in neighborhood. For example, if the
+            # width = 4 and x = radius = 2, we loop through range(0, 5) in respect 
+            # to x but this means that the 0 position is repeated since 0 % 4 = 4 % 4.
             xdim_even, ydim_even = (self.width + 1) % 2, (self.height + 1) % 2
             kx = int(x_radius == x_max_radius and xdim_even)
             ky = int(y_radius == y_max_radius and ydim_even)
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -253,12 +253,12 @@ class Grid:
 
         # Using a list for neighhborhood in respect to any other built-in data
         # structure allows to optimize the code further (e.g. with Cython or Numba)
-        # more easily. Also, lists are more optimized in PyPy than in Cpython. Look at
+        # more easily. to better understand how the algorithm was conceived. Look at
         # https://github.com/projectmesa/mesa/pull/1476#issuecomment-1306220403 and in
-        # general to the PR#1476 to better understand how the algorithm was conceived.
+        # general to the PR#1476.
         neighborhood = []
-        x, y = pos
 
+        x, y = pos
         if self.torus:
             x_max_radius, y_max_radius = self.width // 2, self.height // 2
             x_radius, y_radius = min(radius, x_max_radius), min(radius, y_max_radius)
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -266,7 +266,7 @@ class Grid:
             # For each dimension, in the edge case where the radius is as big as
             # possible and the dimension is even we need to shrink by one the range
             # of values to avoid duplicates in neighborhood. For example, if the
-            # width = 4 and x = radius = 2, we loop through range(0, 5) in respect 
+            # width = 4 and x = radius = 2, we loop through range(0, 5) in respect
             # to x but this means that the 0 position is repeated since 0 % 4 = 4 % 4.
             xdim_even, ydim_even = (self.width + 1) % 2, (self.height + 1) % 2
             kx = int(x_radius == x_max_radius and xdim_even)
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -253,7 +253,7 @@ class Grid:
 
         # Using a list for neighhborhood in respect to any other built-in data
         # structure allows to optimize the code further (e.g. with Cython or Numba)
-        # more easily. to better understand how the algorithm was conceived. Look at
+        # more easily. To better understand how the algorithm was conceived, look at
         # https://github.com/projectmesa/mesa/pull/1476#issuecomment-1306220403 and in
         # general to the PR#1476.
         neighborhood = []
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -251,22 +251,33 @@ class Grid:
         if neighborhood is not None:
             return neighborhood
 
+        # Using a list for neighhborhood in respect to any other built-in data
+        # structure allows to optimize the code further (e.g. with Cython or Numba)
+        # more easily. Also, lists are more optimized in PyPy than in Cpython. Look at
+        # https://github.com/projectmesa/mesa/pull/1476#issuecomment-1306220403 and in
+        # general to the PR#1476 to better understand how the algorithm was conceived.
         neighborhood = []
-
         x, y = pos
-        if self.torus:
-            x_radius = min(radius, self.width // 2)
-            y_radius = min(radius, self.height // 2)
 
-            for dx in range(-x_radius, x_radius + 1):
-                for dy in range(-y_radius, y_radius + 1):
+        if self.torus:
+            x_max_radius, y_max_radius = self.width // 2, self.height // 2
+            x_radius, y_radius = min(radius, x_max_radius), min(radius, y_max_radius)
+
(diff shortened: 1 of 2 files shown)


Tortar:
this was left unnotices because the grid tested have only odd dimensions :-)

Tortar:
It is a minimum probability situation (why does it exist?! :D) but we need to account for it

Tortar:
@rht if the difference will be so minimal I will be with you, will do some testing, I feel like using lists is preferable also because of the advantages of conversion of the algorithm in other forms (Numba, Cython, (I think that lists are also more optimized in pypy)) as you already pointed out many times :-)

Tortar:
Results:

- no difference for cpython
- pypy likes more lists: from 2.5 to 10 times faster than the other with dictionary with radius 10

Tortar:
Added a comment explaining why it's needed to do that and modified and added tests so that also the edge case is covered

tests/test_grid.py
```diff
@@ -41,7 +41,7 @@ def setUp(self):
         Create a test non-toroidal grid and populate it with Mock Agents
         """
         width = 3  # width of grid
-        height = 5  # height of grid
+        height = 6  # height of grid
```
rht, line 44: Needs to document that the height has to be even here.

Tortar: done

rht:
> @rht if the difference will be so minimal I will be with you, will do some testing, I feel like using lists is preferable also because of the advantages of conversion of the algorithm in other forms (Numba, Cython, (I think that lists are also more optimized in pypy)) as you already pointed out many times :-)

That's true. I think we should document https://github.com/projectmesa/mesa/pull/1476#issuecomment-1304540328 in the code, and the Numba/Cython/PyPy optimization.

Tortar:
I added the story about Cython/Numba/Pypy (maybe it's a bit verbose) . The comment doesn't really apply with the code we have right now. But I could link to the PR discussion maybe?

Tortar:
I think that qouting https://github.com/projectmesa/mesa/pull/1476#issuecomment-1306220403 is better maybe

mesa/space.py
```diff
@@ -251,22 +251,33 @@ def get_neighborhood(
         if neighborhood is not None:
             return neighborhood
 
+        # Using a list for neighhborhood in respect to any other built-in data
+        # structure allows to optimize the code further (e.g. with Cython or Numba)
+        # more easily. Also, lists are more optimized in PyPy than in Cpython. Look at
+        # https://github.com/projectmesa/mesa/pull/1476#issuecomment-1306220403 and in
```
rht, line 257: Yeah, linking to this particular [comment](https://github.com/projectmesa/mesa/pull/1476#issuecomment-1306220403) is more informative.

Tortar: I think so too. Should something else be changed? :-)

Corvince:
Just out of interest, is the assertion that using a list will be faster rooted in some actual observations? I am really doubtful that it will be useful for mesa (the framework). 

Again, this is more for my curiosity, I think in this PR/setting it is the better choice anyways.

Corvince:
But I think it is super annoying that we need this bug fix for such a corner case ;)

Tortar:
My own non-authoritative observation is that arrays/vectors are easier data structures than hashmaps in C/C++, which I think has an impact on Cython and the others. For example, I compared the two algos with PyPy and it liked more the list alternative. It's difficult to find comparisons on the Internet for Cython or Numba because usually they are used for different purposes. From the point of view to the benefits to Mesa, I'm not sure actually too but anyway I like more this one than the alternative because it exploits better the topology.  But yes, surely rather annoying that we need to deal with such a special case :D

Tortar:
Actually also in the faster Cpython 3.11 which I just downloaded now it is slightly favoured the way with lists (a step in the right direction? :D)

Tortar:
Indeed adding 3.11 to the build Github actions I think could be good

rht:
> Indeed adding 3.11 to the build Github actions I think could be good

I made #1519, but we don't have performance testing framework yet.

Corvince:
Oops, sorry, I was confused by your answers and now realised that my previous comment was saying something completely different than what I had in mind.

I didn't want to discuss lists vs dicts. As you point out they serve different purposes (and in that sense the code comments notion that "lists are more optimized in PyPy than in Cpython" should maybe be removed). 

But the question I had in mind was about numba/cython providing speed-ups for mesa. I have my doubts that it does. I used numba years ago and for some calculations it provided extreme speedups, but I doubt we have many (or any) functions where it helps. But this is purely a gut feeling of mine, so I was interested if there are actual observations. I only found the small tests in https://github.com/projectmesa/mesa/pull/1492 but there numba seemed slower

Tortar:
From what I observed, Numba has more overhead for function calls, which means that with small inputs can become slower than pure python, while Cython suffers less by this so it can be the optimal solution if someone has the courage to go that way :D

Tortar:
I will try to come up with some small inputs benchmarks for Cython in the future

Tortar:
Can I ask you @Corvince an evaluation of https://github.com/projectmesa/mesa/discussions/1513#discussioncomment-4069436, because I'd like to implement it if others are ok with it

rht:
Numba/Cython can provide speedup to Mesa space module, if `Grid` internal is a blackbox without Python objects that is accessed by occasional API calls. `get_neighborhood` definitely will get a speedup with `nopython` and `nogil` (while `dict` requires GIL). `get_neighbors` should only return a list of agent IDs, because the `Grid.grid` has to be a typed array. An array of Python objects likely requires GIL. The part might be slower is if `__getitem__` is accessed multiple times from Python.

Corvince:
> Numba/Cython can provide speedup to Mesa space module, if `Grid` internal is a blackbox without Python objects that is accessed by occasional API calls. `get_neighborhood` definitely will get a speedup with `nopython` and `nogil` (while `dict` requires GIL). 

See, these are those assertions I am talking about. How do you know that? I am not saying this is wrong, but is there any proof or at least indication for that?

Corvince (approved):

rht:
If the computation happens without GIL, and with explicit types, of course it will be fast.

mesa/space.py
```diff
@@ -251,22 +251,33 @@ def get_neighborhood(
...
-            x_radius = min(radius, self.width // 2)
-            y_radius = min(radius, self.height // 2)
+            x_max_radius, y_max_radius = self.width // 2, self.height // 2
+            x_radius, y_radius = min(radius, x_max_radius), min(radius, y_max_radius)
 
-            for dx in range(-x_radius, x_radius + 1):
-                for dy in range(-y_radius, y_radius + 1):
+            # For each dimension, in the edge case where the radius is as big as
+            # possible and the dimension is even we need to shrink by one the range
+            # of values to avoid duplicates in neighborhood
```
rht, line 268: One last thing: needs to add the example `(0 % 4 and 4 % 4)`.

Tortar: Added

mesa/space.py
```diff
@@ -251,22 +251,35 @@ def get_neighborhood(
...
+            x_max_radius, y_max_radius = self.width // 2, self.height // 2
+            x_radius, y_radius = min(radius, x_max_radius), min(radius, y_max_radius)
 
-            for dx in range(-x_radius, x_radius + 1):
-                for dy in range(-y_radius, y_radius + 1):
+            # For each dimension, in the edge case where the radius is as big as
+            # possible and the dimension is even we need to shrink by one the range
+            # of values to avoid duplicates in neighborhood. For example, if the
+            # width = 4 and x = radius = 2, we loop through range(0, 5) in respect
+            # to x but this means that the 0 position is repeated since 0 % 4 = 4 % 4.
```
rht, line 270: It's unclear to have the `=` being used for both the assignment and the equality operator. Should be "0 % 4 and 4 % 4 are both 0".

rht: @Tortar I can fix this instead, and some docs stuff in the future, if you don't mind, so there are fewer review roundtrips.

Tortar: yes, it would be perfect if you can do it, I think you are more proficient than me in english after all :-)

Tortar: Anyway just a "==" instead of "=" I think would be sufficient

PR #1519, opened 2022-11-10
ci: Add testing on Python 3.11 [WIP]

rht:


rht:
Test failed on 3.11. I will fix this later.

.github/workflows/build_lint.yml
```diff
@@ -32,6 +32,8 @@ jobs:
         os: [windows, ubuntu, macos]
         python-version: ["3.10"]
         include:
+          - os: ubuntu
+            python-version: "3.11"
```
EwoutH, line 36: Thanks, that needed to be done indeed! I would say let's bump the version that runs on all OSes to 3.11, and add a 3.10 job for Ubuntu.
```suggestion
        python-version: ["3.11"]
        include:
          - os: ubuntu
            python-version: "3.10"
```
Do you have any idea what the error is about? It doesn't look that Python 3.11 specific.

rht: >  I would say let's bump the version that runs on all OSes to 3.11, and add a 3.10 job for Ubuntu.

Will do, though I'm still investigating. I have reproduced the bug in a Docker instance.

Tortar: I found the source of the bug, but I'm not sure why the other Python didn't identify it as a bug actually

Tortar: If you want to add the mod to this PR, I will close mine

rht: Let's solve it in #1521, and limit the scope of this PR to updating the CI.

jackiekazil:
@rht are we waiting for resolution on #1521 & merge both at once? (I could go either way... I am leaning towards merging this, but wanted to check in and see if you were in the same head space.)

rht:
I need to understand why the problem happens in 3.11, on whether it is specific to `unittest.mock.patch`, or it could happen again in other context, due to the conflicting import namespace.

Tortar:
Maybe trying it again with 3.11.1 bugfix release?

rht:
It works on 3.11.1! :tada:

EwoutH:
Awesome! One of the first times I have seen a bug actually from CPython itself.

.github/workflows/build_lint.yml
EwoutH: > Thanks, that needed to be done indeed! I would say let's bump the version that runs on all OSes to 3.11, and add a 3.10 job for Ubuntu.

@rht What do you think of this?

rht: Sounds good. Doing it now.

EwoutH (approved):
Awesome, LGTM!

PR #1521, opened 2022-11-10
Fix test failed for 3.11 build

Tortar:
This should fix test failed in #1519

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -20,7 +20,7 @@ class TestCli(unittest.TestCase):
         sys.path[:] = self.old_sys_path
 
     def test_run(self):
-        with patch("mesa.visualization.ModularServer") as ModularServer:
+        with patch("mesa.visualization.ModularVisualization.ModularServer") as ModularServer:
             example_dir = os.path.abspath(
                 os.path.join(os.path.dirname(__file__), "../examples/wolf_sheep")
             )
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -20,7 +20,7 @@ class TestCli(unittest.TestCase):
         sys.path[:] = self.old_sys_path
 
     def test_run(self):
-        with patch("mesa.visualization.ModularVisualization.ModularServer") as ModularServer:
+        with patch("mesa.flat.visualization.ModularServer") as ModularServer:
             example_dir = os.path.abspath(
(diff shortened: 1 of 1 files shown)


EwoutH:
Quite interesting that this fails on 3.11 but passes on all other Python versions

tests/test_main.py
```diff
@@ -20,7 +20,7 @@ def tearDown(self):
         sys.path[:] = self.old_sys_path
 
     def test_run(self):
-        with patch("mesa.visualization.ModularServer") as ModularServer:
+        with patch("mesa.flat.visualization.ModularServer") as ModularServer:
```
rht, line 23: This solves the problem, but I still need to know the underlying cause. It seems there is a collision between https://github.com/projectmesa/mesa/blob/77599fa67e74331ec958482ff4d053bb4da514b6/mesa/__init__.py#L13 and the actual `mesa/visualization` folder https://github.com/projectmesa/mesa/tree/main/mesa/visualization.

rht: When an import happens.

rht: I renamed `mesa/visualization` to `mesa/viz`, and
```python
with patch("mesa.visualization.ModularServer") as ModularServer:
    ....
```
works.

Tortar: What's the way to go? leave it as it is now or change the name of the folder? I think changing the name would be a breaking change (?)

rht: Yeah, changing the folder name is a breaking change for 1 of the 3 supported way to do imports. I will go through the Python 3.11 changelog by the end of the weekend.

EwoutH:
Did you have time to investigate the Python 3.11 failure @rth?

rht:
I did scan through https://docs.python.org/3/whatsnew/changelog.html and couldn't find a relevant change that caused it. Will go through it again when I have the time window to do it.

rht:
It is possible that the change happened in `unittest.mock.patch`. I also did compare the `patch` implementation between 3.10 and 3.11, but I forgot what I found out.

Tortar:
Anyway It is probably a bug of Python 3.11 not a feature given that it's its first release, if you find out the reason then it should be reported I think

EwoutH:
There are no deprecation warnings in 3.10 that give anything away right?

Tortar:
We can close this one given that the bugfix release 3.11.1 works

PR #1523, opened 2022-11-11
Enhance schedulers to support intra-step removal of agents

Tortar:
As discussed in #1515 the only scheduler which checks for the presence of the agents was the BaseScheduler (and the ones which inherit the method `agent_buffer`), I suppose that this should be done also for the other schedulers except to the SimultaneousActivation one

### DIFF
--- a/tests/test_time.py
+++ b/tests/test_time.py
@@ -29,7 +29,15 @@ class MockAgent(Agent):
         self.steps = 0
         self.advances = 0
 
+    def kill_other_agent(self):
+        for agent in self.model.schedule.agents:
+            if agent != self:
+                self.model.schedule.remove(agent)
+                break
+
     def stage_one(self):
+        if self.model.enable_kill_other_agent:
+            self.kill_other_agent()
         self.model.log.append(self.unique_id + "_1")
 
     def stage_two(self):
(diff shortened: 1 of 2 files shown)


rht:
L105-107 has existed in the code (was added via #584, which didn't add the tests). I could merge this PR and add the tests later, so that people can use the feature sooner, but at this point I think we should add the tests now.

Tortar:
Will put some tests

Tortar:
I added some tests :-)

Tortar:
I Is it better according to you to have a subclass of MockAgent with a different step instead of how it is done now?

tests/test_time.py
```diff
@@ -29,7 +29,14 @@ def __init__(self, unique_id, model):
         self.steps = 0
         self.advances = 0
 
+    def kill_agent(self):
```
rht, line 32: `kill_other_agent` sounds clearer. Also, I think removing just 1 agent is sufficient.

tests/test_time.py
```diff
@@ -29,7 +29,14 @@ def __init__(self, unique_id, model):
...
+        for agent in self.model.schedule.agents:
+            if agent != self:
+                self.model.schedule.remove(agent)
+
     def stage_one(self):
+        if self.model.calling_func == "test_intrastep_remove":
```
rht, line 38: Maybe define a boolean attribute `do_kill_other_agent` at agent init? `kill_other_agent` might be used in other tests.

rht:
> @rht Is it better according to you to have a subclass of MockAgent with a different step instead of how it is done now?

I think the current implementation is fine. The step logic is still clear even with the addition of the `kill_other_agent` conditional.

tests/test_time.py
rht: `enable_kill_other_agent` seems better. `do_kill_other_agent` could be erroneously interpreted as a function/action (although, `enable_kill_other_agent` also does).

Tortar: done!

rht:
LGTM

PR #1527, opened 2022-11-15
Refactor iter_cell_list_contents

Tortar:
This is easier and a bit faster 

import timeit
timeit.timeit("for y in (a[0][x] for x in range(s) if a[0][x]): pass", "a=[[None]*30+[1]*3+[None]*30+[1]*5]; s=len(a[0])") # -> 4 sec
timeit.timeit("for y in filter(None,(x for x in range(s))): pass", "a=[[None]*30+[1]*3+[None]*30+[1]*5]; s=len(a[0])") # -> 5 sec

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -388,10 +388,8 @@ class Grid:
         Returns:
             An iterator of the contents of the cells identified in cell_list
         """
-        # Note: filter(None, iterator) filters away an element of iterator that
-        # is falsy. Hence, iter_cell_list_contents returns only non-empty
-        # contents.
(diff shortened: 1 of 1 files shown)


Tortar:
Actually it's very good this simple change decreases the time of schelling by ~15%

rht:
> Actually it's very good this simple change decreases the time of schelling by ~10%

That's good to hear. I'm tweaking the commit message with `perf:` prefix so that we can keep track of the perf commits in the changelog.

rht:
LGTM

PR #1531, opened 2022-11-17
Simplify accept_tuple_argument decorator in space.py

Tortar:
This more readable decorator creates also less overhead

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -64,11 +64,11 @@ def accept_tuple_argument(wrapped_function: F) -> F:
     to also handle a single position, by automatically wrapping tuple in
     single-item list rather than forcing user to do it."""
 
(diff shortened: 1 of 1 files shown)


mesa/space.py
```diff
@@ -391,7 +391,6 @@ def iter_cell_list_contents(
         # iter_cell_list_contents returns only non-empty contents.
         return (self.grid[x][y] for x, y in cell_list if self.grid[x][y])
 
-    @accept_tuple_argument
```
rht, line 394: I can merge without this line change. I think we can discuss this later.

Tortar: not a problem

Tortar: reverted

rht:
LGTM

Issue #1528, opened 2022-11-16
Is the ContinuousSpace.get_heading function correct?


Tortar:
I looked at the implementation of the ContinuousSpace.get_heading function which should return the heading angle between two vectors, I searched online and nothing suggests that the way it's done right now makes sense, am I wrong? It seems to lack some passages to return the correct value, see https://stackoverflow.com/questions/19058764/calculate-heading-angle-from-x-and-y-information or https://stackoverflow.com/questions/4403515/heading-from-point-a-to-point-b-in-2d-space

Right now it is

```python
    def get_heading(
        self, pos_1: FloatCoordinate, pos_2: FloatCoordinate
    ) -> FloatCoordinate:
        """Get the heading angle between two points, accounting for toroidal space.
        Args:
            pos_1, pos_2: Coordinate tuples for both points.
        """
        one = np.array(pos_1)
        two = np.array(pos_2)
        if self.torus:
            one = (one - self.center) % self.size
            two = (two - self.center) % self.size
        heading = two - one
        if isinstance(pos_1, tuple):
            heading = tuple(heading)
        return heading
```

rht:
I agree that the docstring is wrong, even though the usage in the Boid flockers example is correct. [This](https://github.com/projectmesa/mesa/commit/10b262a9eaeeb5145a41814bca9faab18c6d218b) is the commit that added the docstring. [This](https://github.com/projectmesa/mesa/commit/600c62b35dbac6de9300da471377b0e200b60da8) is the commit that implemented `get_heading`. I think a seamless fix would be to update the docstring only.

Tortar:
Yes, I agree with you, didn't think there were two different commits

Tortar:
Solved by #1535

PR #1535, opened 2022-11-20
Improve docstrings of ContinuousSpace

Tortar:
This also should solve #1528

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -934,8 +934,8 @@ class ContinuousSpace:
     def get_heading(
         self, pos_1: FloatCoordinate, pos_2: FloatCoordinate
     ) -> FloatCoordinate:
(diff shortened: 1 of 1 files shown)


mesa/space.py
```diff
@@ -909,7 +907,7 @@ def remove_agent(self, agent: Agent) -> None:
     def get_neighbors(
         self, pos: FloatCoordinate, radius: float, include_center: bool = True
     ) -> list[Agent]:
-        """Get all objects within a certain radius.
+        """Get all agents within a certain radius.
```
rht, line 910: I suppose this is true in general -- a space contains only agents. It's always agents, even for fixtures.

rht: Even if there are such models, the current space implementation doesn't accommodate them yet (also evident from the Mypy typing). As such, I'm merging now.

PR #1536, opened 2022-11-22
Simplify code in ContinuousSpace

Tortar:


mesa/space.py
```diff
@@ -849,18 +849,15 @@ def __init__(
...
-        self._agent_points = np.array(
-            [self._index_to_agent[idx].pos for idx in range(len(agents))]
-        )
+        # Since dicts are ordered by insertion, we can iterate through agents keys
+        self._agent_points = np.array([agent.pos for agent in self._agent_to_index])
 
     def _invalidate_agent_cache(self):
         """Clear cached data of agents and positions in the space."""
         self._agent_points = None
-        self._index_to_agent = {}
```
rht, line 863: This attribute is used in `get_neighbors`, which hasn't been refactored away.

rht: The cache clearing functionality should not depend on how the cache is checked/rebuilt. `self._index_to_agent` is a cache that should be cleared in this method regardless of how it is used elsewhere.

Tortar: Understood 👍

rht:
LGTM

Issue #1537, opened 2022-04-01
How to adapt tick labels for horizontal axis in ChartModule?


stefaneidelloth:

wang-boyu:
Issue transferred to mesa for advice

rht:
I made #1540.

PR #1540, opened 2022-11-30
feat: viz: Allow initial tick label to be settable in ChartModule

rht:
This attempts to address #1537 as a quick fix. There could be a better solution. I'm making this approach to start a discussion.

### DIFF
--- a/mesa/visualization/modules/ChartVisualization.py
+++ b/mesa/visualization/modules/ChartVisualization.py
@@ -48,6 +48,7 @@ class ChartModule(VisualizationElement):
         canvas_height=200,
         canvas_width=500,
         data_collector_name="datacollector",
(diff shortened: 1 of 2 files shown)


wang-boyu:
I guess he/she wanted something slightly different?

Instead of `model.schedule.steps` or `model.schedule.time`, could we have the x axis ticks to be something customizable such as `model.year`?

rht:
It's harder to generalize the implementation to support `model.year` or any other custom attribute. With this PR, you simply need to specify the initial point. And can be extended to support custom step multiplier. Though, this PR doesn't support `datetime` x axis.

rht:
Closing because we haven't come to a conclusion on how to solve this.

PR #1542, opened 2022-12-02
breaking: NetworkGrid: modify get_neighbors and create get_neighborhood

Tortar:
NetworkGrid didn't have a get_neighbors function which returns the adjacent agents before this PR, but I think it's an important part of the API, unfortunately it has already a get_neighbors function which returns ids, but clearly changing it would be a breaking change. 

I think if we want to break it in 2.0 we could rename get_neighbors to get_neighborhood and rename this function to get_neighbors. Is it better to include this PR now as it is and make a separate PR with the breaking change or modify this one with the proposed changes?

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -1069,22 +1069,27 @@ class NetworkGrid:
         self.G.nodes[node_id]["agent"].append(agent)
         agent.pos = node_id
 
-    def get_neighbors(
+    def get_neighborhood(
         self, node_id: int, include_center: bool = False, radius: int = 1
     ) -> list[int]:
         """Get all adjacent nodes within a certain radius"""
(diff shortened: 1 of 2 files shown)


rht:
It would be less confusing to do this all at once for 2.0, instead of defining a new temporary API `get_agent_neighbors`, only for it to be renamed. I have updated #1511 to include this PR.

Tortar:
ok @rht, not it should be ready to be merged in 2.0

rht:
I have rebased the PR, resolving the merge conflicts, and create https://github.com/projectmesa/mesa-examples/pull/40 for the examples update. Merging; thank you @Tortar.

rht (approved):

PR #1543, opened 2022-12-02
Make self.empties optional using track_empties

Tortar:
This is a draft for making self.empties optional so that is not tracked unless the user is interested, as it is a breaking change for 2.0 it is used a decorator @property so that the user is warned about the change if he uses self.empties.


Tortar:
What do you think @rht of this PR? Is it worth it?

rht:
There are 2 paths forward:
- This PR
- Remove `empties` altogether, simplifying the code. Users who need it can implement it themselves. I can't make any statement on whether most models have dense agent density in the grid or not. It is also possible for the agents to begin as sparse, then grow to be dense.

This PR doesn't complicate the code that much, and so, I'm fine with it. Note that the fallback Agents.jl code for `move_to_empty` is to compute all the empties from scratch every time. This is reasonable because Julia is fast, but not so for Python.

Tortar:
I agree with you @rht, and I prefer to keep the functionality. The PR should be ready for review, some parts are not covered by tests, but it those parts are safe I think (only the move_agent part in SingleGrid seems important to cover, but somehow, though I added a test for that part, Codecov still complains about it).

mesa/space.py
```diff
@@ -102,18 +104,37 @@ def __init__(self, width: int, height: int, torus: bool) -> None:
...
+            self.num_empties = self.num_cells
 
         # Neighborhood Cache
         self._neighborhood_cache: dict[Any, list[Coordinate]] = dict()
 
+    @property
+    def empties(self):
+        # this method can be removed in some further release, when
+        # users will be warned about the change in the API.
+        if self.track_empties is False:
```
rht, line 129: `if not self.track_empties`

mesa/space.py
```diff
@@ -414,24 +435,36 @@ def move_agent(self, agent: Agent, pos: Coordinate) -> None:
             pos: Tuple of new position to move the agent to.
         """
         pos = self.torus_adj(pos)
-        self.remove_agent(agent)
-        self.place_agent(agent, pos)
+        if self.track_empties:
+            self._empties.discard(pos)
+            self._empties.add(agent.pos)
+        x, y = agent.pos
```
rht, line 441: This is reinventing the wheel. Why replace `remove_agent` and `place_agent`?

mesa/space.py
```diff
@@ -551,16 +588,24 @@ def position_agent(
             )
 
         if x == "random" or y == "random":
-            if len(self.empties) == 0:
+            if len(self._empties) == 0:
```
rht, line 591: `self._empties` isn't always defined.

mesa/space.py
```diff
@@ -551,16 +588,24 @@ def position_agent(
...
+        if self.is_cell_empty(pos):
+            Grid.move_agent(self, agent, pos)
+        else:
+            raise Exception("Cell not empty")
+
     def place_agent(self, agent: Agent, pos: Coordinate) -> None:
+        """Place the agent at the specified location, and set its pos variable."""
         if self.is_cell_empty(pos):
-            super().place_agent(agent, pos)
+            Grid.place_agent(self, agent, pos)
```
rht, line 608: Why replace the `super()`?

mesa/space.py
```diff
@@ -596,19 +641,31 @@ def place_agent(self, agent: Agent, pos: Coordinate) -> None:
         """Place the agent at the specified location, and set its pos variable."""
         x, y = pos
         if agent.pos is None or agent not in self.grid[x][y]:
+            if self.track_empties:
```
rht, line 644: This 4-line pattern is repeated several times. Maybe we should define a method for it.

mesa/space.py
Tortar: I think that `super()` is less explicit and should be reserved for more complex cases, it is also 2-2.5 times slower than calling the method of the superclass

mesa/space.py
Tortar: right, actually that piece of code was redundant because `move_to_empty` then check if the grid is full too, so I removed it

mesa/space.py
Tortar: done

mesa/space.py
Tortar: because it has 2 advantages:

- when `track_empties` is False we don't do anything on `num_empties` because it remains unchanged
- `agent.pos` is not reset twice
- we spare two functions calls

but let me know if you think those are negligible

Tortar:
I'm thinking of another way actually right now, why not to adopt another strategy where `empties` is built and updated only if it happens that the user calls `empties` or  `move_to_empty` ? I think it could be a better option, I'm willing to reimplement it, it wouldn't be a breaking change this way

Tortar:
the code would be also much less verbose than this one

mesa/space.py
rht: The priority is readability > performance, unless the performance benefit is very significant. There is also the burden of having to explain why it is done this way instead of the other, for future readers of the code.

rht:
> I'm thinking of another way actually right now, why not to adopt another strategy where _empties is built and updated only if it happens that the user calls empties or move_to_empty ? I think it could be a better option, I'm willing to reimplement it, it wouldn't be a breaking change this way
> the code would be also much less verbose than this one

Yeah, SGTM.

mesa/space.py
rht: https://docs.python.org/3/library/functions.html#super
There is the other use case:
> There are two typical use cases for super. In a class hierarchy with single inheritance, super can be used to refer to parent classes without naming them explicitly, thus making the code more maintainable. This use closely parallels the use of super in other programming languages.

Tortar: I don't buy that much this explaination :D, it seems to me that the mantainability is more or less the same, also in our case I think our inheritance tree is rather static so...I have seen many programmers actually against this use case for super :D

Tortar:
Close this one in favor of #1546

PR #1546, opened 2022-12-05
Evaluate empties set more lazily

Tortar:
This implementation is better than the one in #1543 , it builds the empties set only if it is required to do so with no breaking change.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -108,8 +108,10 @@ class Grid:
             [self.default_val() for _ in range(self.height)] for _ in range(self.width)
         ]
 
-        # Add all cells to the empties list.
-        self.empties = set(itertools.product(range(self.width), range(self.height)))
+        # Flag to check if the empties set has been created. Better than initializing
+        # _empties as set() because in this case it would become impossible to discern
+        # if the set hasn't still being built or if it has become empty after creation.
+        self.empties_built = False
 
         # Neighborhood Cache
         self._neighborhood_cache: dict[Any, list[Coordinate]] = dict()
@@ -118,6 +120,21 @@ class Grid:
     def default_val() -> None:
         """Default value for new cell elements."""
         return None
+
+    @property
+    def empties(self) -> set:
+        if not self.empties_built:
+            self.build_empties()
+        return self._empties
+
+    def build_empties(self) -> None:
+        self._empties = set(
+            filter(
+                self.is_cell_empty,
+                itertools.product(range(self.width), range(self.height)),
+            )
+        )
+        self.empties_built = True
 
     @overload
     def __getitem__(self, index: int) -> list[GridContent]:
@@ -421,7 +438,8 @@ class Grid:
         """Place the agent at the specified location, and set its pos variable."""
         x, y = pos
         self.grid[x][y] = agent
-        self.empties.discard(pos)
+        if self.empties_built:
+            self._empties.discard(pos)
         agent.pos = pos
 
     def remove_agent(self, agent: Agent) -> None:
@@ -430,7 +448,8 @@ class Grid:
             return
         x, y = pos
         self.grid[x][y] = self.default_val()
-        self.empties.add(pos)
+        if self.empties_built:
+            self._empties.add(pos)
         agent.pos = None
 
     def swap_pos(self, agent_a: Agent, agent_b: Agent) -> None:
@@ -551,8 +570,6 @@ class SingleGrid(Grid):
             )
 
         if x == "random" or y == "random":
-            if len(self.empties) == 0:
-                raise Exception("ERROR: Grid full")
             self.move_to_empty(agent)
         else:
             coords = (x, y)
@@ -598,15 +615,16 @@ class MultiGrid(Grid):
         if agent.pos is None or agent not in self.grid[x][y]:
             self.grid[x][y].append(agent)
             agent.pos = pos
-            self.empties.discard(pos)
+            if self.empties_built:
+                self._empties.discard(pos)
 
     def remove_agent(self, agent: Agent) -> None:
         """Remove the agent from the given location and set its pos attribute to None."""
         pos = agent.pos
         x, y = pos
         self.grid[x][y].remove(agent)
-        if self.is_cell_empty(pos):
-            self.empties.add(pos)
+        if self.empties_built and self.is_cell_empty(pos):
+            self._empties.add(pos)
         agent.pos = None
 
     @accept_tuple_argument


Tortar:
Tests are failing due to the move of examples folder to another repo, is there a way to use them though they are in another repo?

mesa/space.py
```diff
@@ -470,7 +486,11 @@ def move_to_empty(
                 ),
                 DeprecationWarning,
             )
-        num_empty_cells = len(self.empties)
+        if not self._empties:
+            self.build_empties()
+
+        num_empty_cells = len(self._empties)
```
rht, line 492: Why not do this in 1 line, `num_empty_cells = len(self.empties)`?

Tortar: sounds good, done

mesa/space.py
```diff
@@ -488,7 +504,7 @@ def move_to_empty(
                 if self.is_cell_empty(new_pos):
                     break
         else:
-            new_pos = agent.random.choice(sorted(self.empties))
+            new_pos = agent.random.choice(sorted(self._empties))
```
rht, line 507: This should be `self.empties` whenever possible.

Tortar: yes, in this case it doesn't impact so we can do it

Tortar: but why in your opinion? it's clear that calling self.empties is more expensive  to call self._empties so  I don't think it's really necessary

rht: `self.empties` is for public consumption. The inner workings might change, so using `self.empties` is more robust. Again, if the performance benefit is marginal, it's not worth it.

Tortar: perfect, thank you for the reminder not to look just at the perf side :D, anyway I changed it

Tortar:
we need actually to create a flag instead of using empties itself, because
otherwise exists_empties could become very very slow when empties was built
but became empty

Il Mar 6 Dic 2022, 10:13 rht ***@***.***> ha scritto:

Tortar:
ok, everything should be hopefully done

Tortar:
no it's not true, say the user built `self.empties`  in some way, and then populate the list till its maximum capacity, so that `self.empties` becomes empty, then say the user checks (say at every step) that the maximum capacity has been reached  calling `exists_empty_cells`  then build.empties would be executed again and again because the set will remain empty each time even after building.

Tortar:
Maybe it should be documented that uses a flag has this benefit.

Tortar:
Maybe it should be documented that using a flag has this benefit.

Tortar:
ohh right, my bad using self._empties and not self.empties solves the problem.

Tortar:
wait but at the same time, it solves anyway the problem if the user instead of using self,exists_empty_cells  uses len(self.empties) > 0 which can happen

Tortar:
wait but at the same time, it solves anyway the problemthat can happen if the user instead of using self.exists_empty_cells uses len(self.empties) > 0, which can happen. Or more generally, whenever a read operation is done on `self.empties` for some reason if it is empty then you have this problem. Maybe it should be documented that using a flag has this benefit.

Tortar:
Also using self._empties and not self.empties in exists_empty_cells would result in a bug if exists_empty_cells was called before the creation of self._empties through self.empties, because it would as empty set

rht:
I see, then it sounds good to me. Need to be documented, though. One option is to initialize `self._empties = None` and check for `is not None`, but this is less explicit.

Tortar:
Added an explaination, feel free to modify/extend it if you think it can be improved :-)

Tortar:
on my pc anyway the grid creation now takes 1/4 of the time!

rht:
I saw from my inbox that you measured a 25% speedup in grid construction. Can't find it on GH. Did you remove it?

Tortar:
yes, because I realized that it's a 1 time operation per model in the end (but anyway it was not a 25% speedup but a 4 times speed up for a 1000 x 1000 grid). I think the perf benefits are 1) Less memory usage if empties is not built and 2) Something like (just guessing) ~15-20% speedup for place_agent, remove_agent, move_agent for Grid, SingleGrid and MultiGrid if empties is not built.

rht:
These are significant improvements, and need to be communicated properly to the users. I think we should measure just one of the examples to see how much holistic speedup we get to running a simulation (without GUI), for a concrete illustration.

rht:
What about sugarscape_cg? See if it is also ~10%.

Tortar:
No it's not, it seems much smaller. There are many more operations in this model so it makes sense.

Tortar:
yes, I agree that saying ~15-25% speedup for place_agent, remove_agent, move_agent is more informative 👍

rht:
Though we need actual hard numbers instead of speculation of the speedup amount.

Tortar:
i  made a little script to check all cases:

```
import timeit

mock_agent = """
class a:
    def __init__(self,pos):
        self.pos=pos
agent= a(None); pos_1=(3,5); pos_2=(3,4)"""

grid_setup = """from mesa.space import Grid; grid = Grid(10,10,True);\n""" + mock_agent
grid_setup_2 = """from mesa.space_2 import Grid; grid = Grid(10,10,True);\n""" + mock_agent
multigrid_setup = """from mesa.space import MultiGrid; grid = MultiGrid(10,10,True);\n""" + mock_agent
multigrid_setup_2 = """from mesa.space_2 import MultiGrid; grid = MultiGrid(10,10,True);\n""" + mock_agent
singlegrid_setup = """from mesa.space import SingleGrid; grid = SingleGrid(10,10,True);\n""" + mock_agent
singlegrid_setup_2 = """from mesa.space_2 import SingleGrid; grid = SingleGrid(10,10,True);\n""" + mock_agent

ggrid_stmt_place = """grid.place_agent(agent, pos_1)"""
ggrid_stmt_remove = """grid.place_agent(agent, pos_1); grid.remove_agent(agent)"""
ggrid_stmt_move = """grid.place_agent(agent, pos_1); grid.move_agent(agent, pos_2)"""


setups = {"grid":[grid_setup, grid_setup_2],
          "singlegrid": [singlegrid_setup, singlegrid_setup_2],
          "multigrid": [multigrid_setup, multigrid_setup_2]}

for key,val in setups.items():

    x, y = val
    a = sum(timeit.repeat(ggrid_stmt_place ,x, number=1, repeat=10000))
    b = sum(timeit.repeat(ggrid_stmt_remove ,x, number=1, repeat=10000))
    c = sum(timeit.repeat(ggrid_stmt_move ,x, number=1, repeat=10000))
    d = sum(timeit.repeat(ggrid_stmt_place ,y, number=1, repeat=10000))
    e = sum(timeit.repeat(ggrid_stmt_remove ,y, number=1, repeat=10000))
    f = sum(timeit.repeat(ggrid_stmt_move ,y, number=1, repeat=10000))

    print(key)
    print("place_agent: ", a, d, d/a)
    print("remove_agent: ", b-a, e-d, (e-d)/(b-a))
    print("move_agent: ", c-a, f-d, (f-d)/(c-a))
    print()
```
This gives

```
grid
place_agent:  0.004205699979138444 0.0056156999789891415 1.335259292589754
remove_agent:  0.0038545999996131286 0.004931200063765573 1.2793026680486947
move_agent:  0.011213199984013045 0.012697600046521984 1.1323797011223635

singlegrid
place_agent:  0.010900000004767207 0.012187399977847235 1.1181100892217406
remove_agent:  0.003727199994500552 0.004700500034232391 1.2611343746426096
move_agent:  0.01603950003391219 0.018579900006443495 1.1583839874784223

multigrid
place_agent:  0.005325500073013245 0.006638100047894113 1.2464745013397738
remove_agent:  0.003202099952432036 0.00773579991528095 2.415852106492026
move_agent:  0.01120519990035973 0.0170470000084606 1.5213472459258246
```

So we arrive at 2.5x in multigrid case for removal! (Because the checking condition is simpler). Another thing to notice is that place_agent in singlegrid is impacted less than the one in grid because of the super() call, I'm for the removal of it, I don't think it's more mantainable in this simple case.

Corvince:
I didn't follow the complete conversation and hope there is nothing blocking, but I think the general approach seems quite clever @Tortar !

Tortar:
@rht We have those hard numbers now on the speed-up👍

PR #1549, opened 2022-12-06
Add some missing const declarations

Tortar:


Tortar:
anyway I'm a total beginner in javascript so probably I'm wrong :-) indeed I found it looking for Mesa on https://lgtm.com/projects/g/projectmesa/mesa/ .

Don't know if someone else looked at this before, but all other alerts seems false positive to me.

rht:
That looks correct.

PR #1558, opened 2022-12-09
Remove auto-update GH Actions for Pipfile.lock

rht:
With this, the Pipfile.lock is going to be frozen. We should decide whether to remove it for 1.2 or 2.0. I think 1.2 is fine, but this decision is not blocking this PR.


EwoutH:
I would say as long as we have a pipfile, let's keep it up to date. Otherwise we end up with an unmaintained environment file.

So my suggestion would be first make a decision about removal, and then then remove them together if that's decided.

I'm personally not using the Pipfile, so I would be fine by it.

rht:
The lockfile already hasn't been updated for at least a month. It will still work even if it hasn't been updated for several more months. 2.0 is coming in a few months.

> So my suggestion would be first make a decision about removal, and then then remove them together if that's decided.

I thought the gist of #1495 comments is that we are switching to supporting Poetry?

Corvince:
What's the advantage of removing the pipfile? I don't see how we would know how many people might use it. Even if the number is relatively low (which I guess it is), I see no reason to abandon those. Given we have this GH Action the maintaineance effort is basically 0, except when tests start to fail, but then that is good to know as well.

rht:
The problem is that having the Pipfile.lock in the repo means installation via this method is still endorsed by the project, which shouldn't be the case.

Corvince:
I get that, but I was asking why is that a problem?

rht:
If you are not endorsing an installation method, why would you support it?

Corvince:
I don't know how else to frame my question. I am really just interested in your opinion on why we should remove the pipfile.

tpike3:
Per the dev session 4 Feb.. merging for 1.2 release. 

@jackiekazil 1.2 release ready at your convenience. 

@Corvince If I am capturing the discussion correctly, this keeps us consistent with [practices of python development](https://packaging.python.org/en/latest/tutorials/packaging-projects/) while removing older practices and reducing code.

PR #1563, opened 2022-12-14
Construct default for comparisons in Grid and NetworkGrid

Tortar:
Instead of building a new default value at each comparison, we can build it just once

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -112,7 +112,7 @@ class Grid:
         # _empties as set() because in this case it would become impossible to discern
         # if the set hasn't still being built or if it has become empty after creation.
         self.empties_built = False
-        
+
         # Just for comparisons
         self.default = self.default_val()
 
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -113,8 +113,8 @@ class Grid:
         # if the set hasn't still being built or if it has become empty after creation.
         self.empties_built = False
 
-        # Just for comparisons
-        self.default = self.default_val()
+        # Useful to check if a cell is empty
+        self.empty_value = self.default_val()
 
         # Neighborhood Cache
         self._neighborhood_cache: dict[Any, list[Coordinate]] = dict()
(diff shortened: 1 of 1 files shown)


rht:
Not sure about this one being significantly faster to warrant the added attribute. It seems changing `default_val` to a property would have worked fine, but this becomes a breaking change.

Tortar:
I calculated how much it is faster, something like 50 % faster than the
previous one for `is_cell_empty`, I think it is not that significant after
all, but we can always make the attribute private if we want. Regarding the
property function, it wouldn't solve the fact that for comparisons (not for
instantiating cells though because we could have problems given that we
reference the same object) it's not necessary to build each time a new
object, we just need one of them.

Il Gio 15 Dic 2022, 10:13 rht ***@***.***> ha scritto:

rht:
Can we instead initialize `self.empty_value` at `__init__`, and have `self.default_value()` access this instead? No API change is needed this way.

Tortar:
Can we instead initialize self.empty_value at __init__, and have
self.default_value() access this instead? No API change is needed this way.

If I understand you well, this would mean to create specialized `__init__`
methods for each subclass and handling the creation also in the superclass.
Also the problem is that we should call it in `default_val` with something
like `type(self.empty_value)()` which is what I remember I did in a closed
PR sometimes ago. This would be more complicated, but I have no preference
though. What do you think?

Il Ven 16 Dic 2022, 07:00 rht ***@***.***> ha scritto:

Tortar:
Anyway no API change is needed also for this one

Tortar:
Added a better explaination of the use case of `default_value`, @rht do you think it is better to make it private or not?

rht:
You meant #1492? What I meant:
```python
def __init__(self, ...):
    ...
    self._empty_value = None

def default_value(self):
    return self._empty_value
```

rht:
My solution doesn't work for `MultiGrid` for construction, because `[]` is mutable and changing 1 will change the rest.
It looks like your solution is only safe for comparison. I think the readability/comprehensibility concern outweighs the performance benefit, that it's not worth it. And any Python-specific optimizations might become obsolete in Numba/Cython.

rht:
Closing because of https://github.com/projectmesa/mesa/pull/1563#issuecomment-1364473151.

PR #1564, opened 2022-12-17
Implement radius for NetworkGrid.get_neighbors

Tortar:
Implement a way to retrieve the neighbors of a certain node even when the radius is greater than 1. It should be the intended way with the networkx library. Still need to add some tests. There are no problem for reproducibility because everything inside `single_source_shortest_path_length` is done with dicts and not with sets.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -26,6 +26,7 @@ import math
 from warnings import warn
 
 import numpy as np
+import networkx as nx
 
 from typing import (
     Any,
@@ -1033,11 +1034,21 @@ class NetworkGrid:
         self.G.nodes[node_id]["agent"].append(agent)
         agent.pos = node_id
 
-    def get_neighbors(self, node_id: int, include_center: bool = False) -> list[int]:
-        """Get all adjacent nodes"""
-        neighbors = list(self.G.neighbors(node_id))
-        if include_center:
-            neighbors.append(node_id)
+    def get_neighbors(
+        self, node_id: int, include_center: bool = False, radius: int = 1
+    ) -> list[int]:
+        """Get all adjacent nodes within a certain radius"""
+        if radius == 1:
+            neighbors = list(self.G.neighbors(node_id))
+            if include_center:
+                neighbors.append(node_id)
+        else:
+            neighbors_with_distance = nx.single_source_shortest_path_length(
+                self.G, node_id, radius
+            )
+            if not include_center:
+                del neighbors_with_distance[node_id]
+            neighbors = list(neighbors_with_distance.keys())
         return neighbors
 
     def move_agent(self, agent: Agent, node_id: int) -> None:
--- a/tests/test_space.py
+++ b/tests/test_space.py
@@ -340,7 +340,7 @@ class TestSingleNetworkGrid(unittest.TestCase):
         """
         Create a test network grid and populate with Mock Agents.
         """
-        G = nx.complete_graph(TestSingleNetworkGrid.GRAPH_SIZE)
+        G = nx.cycle_graph(TestSingleNetworkGrid.GRAPH_SIZE)
         self.space = NetworkGrid(G)
         self.agents = []
         for i, pos in enumerate(TEST_AGENTS_NETWORK_SINGLE):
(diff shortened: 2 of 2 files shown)


Tortar:
all should be ok now!

Tortar:
@rht do you have any concerns about this implementation?

mesa/space.py
```diff
@@ -1033,11 +1034,19 @@ def place_agent(self, agent: Agent, node_id: int) -> None:
...
-        """Get all adjacent nodes"""
-        neighbors = list(self.G.neighbors(node_id))
-        if include_center:
-            neighbors.append(node_id)
+    def get_neighbors(
+        self, node_id: int, include_center: bool = False, radius: int = 1
+    ) -> list[int]:
+        """Get all adjacent nodes within a certain radius"""
+        if radius == 1:
+            neighbors = list(self.G.adj[node_id])
```
rht, line 1042: `G.neighbors` is more readable.

mesa/space.py
```diff
@@ -1033,11 +1034,19 @@ def place_agent(self, agent: Agent, node_id: int) -> None:
...
+            if include_center:
+                neighbors.append(node_id)
+        else:
+            neighbors = nx.single_source_shortest_path_length(self.G, node_id, radius)
+            if not include_center:
+                del neighbors[node_id]
+            neighbors = list(neighbors)
```
rht, line 1049: `neighbors` was a dict, then you only keep the keys here? The Mypy annotation is going to be messy. Should be call the `dict` `neighbors` a different name?

mesa/space.py
Tortar: yet it's just a wrapper of `G.adj` which is anyway a fundamental function of a nx graph so I'd keep this one

mesa/space.py
Tortar: changed the name in `neighbors_with_distance` for the dict

mesa/space.py
rht: They give the same output. For people reading the code, it is easier to understand `G.neighbors` without having to consult NetworkX doc.

Tortar: ok, not a big deal using one or the other, done :-)

mesa/space.py
```diff
@@ -26,6 +26,7 @@
 from warnings import warn
 
 import numpy as np
+import networkx as nx
```
rht, line 29: This makes we wonder if we should set `networkx` as an optional dependency. Not everyone uses the `NetworkGrid`. This concern is not a blocker to this PR, however.

rht:
LGTM

mesa/space.py
Tortar: yes, this can be good in my opinion, but I'm fine also with the current way

Tortar:
just noticed it, I was wrong when I said that reproducibility is established because the function used dicts, I rechecked the code and this is the internal part of the function I used


```
def _single_shortest_path_length_2(adj, firstlevel, cutoff):
    """Yields (node, level) in a breadth first search
    Shortest Path Length helper function
    Parameters
    ----------
        adj : dict
            Adjacency dict or view
        firstlevel : dict
            starting nodes, e.g. {source: 1} or {target: 1}
        cutoff : int or float
            level at which we stop the process
    """
    seen = {}  # level (number of hops) when seen in BFS
    level = 0  # the current level
    nextlevel = set(firstlevel)  # set of nodes to check at next level
    n = len(adj)
    while nextlevel and cutoff >= level:
        thislevel = nextlevel  # advance to next level
        nextlevel = set()  # and start a new set (fringe)
        found = []
        for v in thislevel:
            if v not in seen:
                seen[v] = level  # set the level of vertex v
                found.append(v)
                yield (v, level)
        if len(seen) == n:
            return
        for v in found:
            nextlevel.update(adj[v])
        level += 1
    del seen
```

As you see, it uses sets internally in BFS, so it is not safe, but thinking about it, it is because we assume that `node_id` is an `int` so sets are deterministic in this case because the hash of a integer is the integer itself. Alternatively, we can put a `sorted` instead of `list` in the code.

rht:
Using `sorted` SGTM.

PR #1566, opened 2022-12-21
Make Grid.coord_iter return content, pos

Tortar:
All the examples which use `coord_iter` transform the output so that to have a `pos` tuple, this should be avoided, right now I changed coord_iter,  so this is a breaking change, but we can even (probably better) deprecate this method and create a new one with a new name, what do you suggest as new name @rht, @Corvince? `pos_iter`?


Tortar:
also we can do inside something like `return zip(iter(self), itertools.product(range(self.width), range(self.height)))` which should be faster

Tortar:
ok done this way, because seems much better, still need to add some tests and maybe the deprecation and changing the name if you have any suggestion

rht:
We probably should just do a breaking change to `coord_iter` itself in Mesa 2.0. Adding methods is costly and people can get confused by the plethora of methods.

Tortar:
Agree, all should be ok for merging in 2.0. :-)

rht:
Redone in #1723, because handling the merge conflict takes longer.

PR #1568, opened 2022-12-24
Make the internal grid and empties_built in Grid class private

Tortar:
The internal grid in the Grid class should be made private as was done in #815, the internal mechanic of the class should not be used by the user (if he doesn't really want to do so) and we should be free to change it as we want if needed.

Something should be changed in the tests folder also, but I'd like to finish the PR when it gets some approval :-)

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -89,7 +89,6 @@ class Grid:
     Properties:
         width, height: The grid's width and height.
         torus: Boolean which determines whether to treat the grid as a torus.
-        grid: Internal list-of-lists which holds the grid cells themselves.
     """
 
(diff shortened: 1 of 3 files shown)


rht:
Yeah, I agree with this proposal.

And in general, it is indeed more effective if you sketch a draft PR and wait for agreement, so that your effort is not wasted.

Tortar:
will do that more often!

Tortar:
Everything should be done now, I realized that also `empties_built` shouldn't be public so I put all together in this PR, hope this is welcomed!

rht:
`Grid.grid` has been a public attribute, but it has never been endorsed in the docs, so I think it's fine to make it private without having to wait to do the change in Mesa 2.0.

PR #1570, opened 2022-12-29
Refactor iter/get_cell_list_contents methods

Tortar:
Improved the methods mentioned: for MultiGrid it is more or less a <del>3x</del>1.8x improvement, for NetworkGrid it is more or less a <del>2x</del>1.3x improvement. For Grid there will be a drop in performance of 2x due to make it consistent with the other classes.

Tested with:

```python
from mesa.mesa.space import Grid, MultiGrid, NetworkGrid
from time import perf_counter
import networkx as nx

class MockAgent():
    def __init__(self): self.pos = None

def timing(l_pos, grids):
    res = []
    for i in range(3):
        pos, grid = l_pos[i], grids[i]
        for x in range(width*height):
            grid.place_agent(MockAgent(), pos[x])
            q = 0
            for _ in range(1000):
                t_0 = perf_counter()
                grid.get_cell_list_contents(pos)
                t_1 = perf_counter()
                q += t_1-t_0
            res.append(q)
    return res

width, height = 10, 10
l_pos = [[(x, y) for x in range(width) for y in range(height)],[(x, y) for x in range(width) for y in range(height)],
       [x for x in range(100)]]
G = nx.cycle_graph(100)
grids = [Grid(width, height, True), MultiGrid(width, height, True), NetworkGrid(G)]
res_1 = timing(l_pos, grids)
from mesa.mesa.space_2 import Grid, MultiGrid, NetworkGrid
grids = [Grid(width, height, True), MultiGrid(width, height, True), NetworkGrid(G)]
res_2 = timing(l_pos, grids)
f = [x/y for x,y in zip(res_1, res_2)]
print(min(f[:100]), max(f[:100]), sum(f[:100])/100)
print(min(f[100:200]), max(f[100:200]), sum(f[100:200])/100)
print(min(f[200:]), max(f[200:]), sum(f[200:])/100)
```

where `space_2` contains the updated methods.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -397,29 +397,31 @@ class Grid:
     def iter_cell_list_contents(
         self, cell_list: Iterable[Coordinate]
     ) -> Iterator[Agent]:
-        """Returns an iterator of the contents of the cells
-        identified in cell_list.
+        """Returns an iterator of the agents contained in the cells identified
+        in `cell_list`; cells with empty content are excluded.
 
         Args:
             cell_list: Array-like of (x, y) tuples, or single tuple.
 
         Returns:
-            An iterator of the contents of the cells identified in cell_list
+            An iterator of the agents contained in the cells identified in `cell_list`.
         """
         # iter_cell_list_contents returns only non-empty contents.
-        return (self._grid[x][y] for x, y in cell_list if self._grid[x][y])
+        return (
+            self._grid[x][y]
+            for x, y in itertools.filterfalse(self.is_cell_empty, cell_list)
+        )
 
     @accept_tuple_argument
     def get_cell_list_contents(self, cell_list: Iterable[Coordinate]) -> list[Agent]:
-        """Returns a list of the contents of the cells
-        identified in cell_list.
-        Note: this method returns a list of `Agent`'s; `None` contents are excluded.
+        """Returns an iterator of the agents contained in the cells identified
+        in `cell_list`; cells with empty content are excluded.
 
         Args:
             cell_list: Array-like of (x, y) tuples, or single tuple.
 
         Returns:
-            A list of the contents of the cells identified in cell_list
+            A list of the agents contained in the cells identified in `cell_list`.
         """
         return list(self.iter_cell_list_contents(cell_list))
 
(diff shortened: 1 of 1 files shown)


mesa/space.py
```diff
@@ -639,7 +639,7 @@ def iter_cell_list_contents(
             A iterator of the contents of the cells identified in cell_list
         """
         return itertools.chain.from_iterable(
-            self[x][y] for x, y in cell_list if not self.is_cell_empty((x, y))
+            self._grid[x][y] for x, y in cell_list if self._grid[x][y]
```
rht, line 642: Replacing `not self.is_cell_empty((x, y))` with this is nonlocal and problematic, because when one changes the implementation of `is_cell_empty`, this line needs to be changed as well, which is very easy to miss.

Tortar: this makes me think that we can instead of iterating as is we can apply a custom filter decided by the user, this could be a useful feature.

Tortar: For example:

from:

` return itertools.chain.from_iterable(self._grid[x][y] for x, y in cell_list if not self.is_cell_empty((x, y)))`

we can pass to


```
if exclude is None:
    return itertools.chain.from_iterable(self._grid[x][y] for x, y in cell_list)
else:
    return itertools.chain.from_iterable(self._grid[x][y] for x, y in cell_list if filter(exclude, cell_list))
```

Note that in the first case we have an iterator which returns non-empty content because it chains the content of the cells. What do you think?

Tortar: But the problem remains, in Grid we have an `iter_cell_list_contents` without reference to `is_cell_empty`, here we could do the same with my implementation, so what to do? I think that we can change the docstring to make it more clear that we are returning the agents in `cell_list` and this would be enough, something like `Returns an iterator of the agents in the
cells identified in cell_list.`

Tortar: What about now? (I changed also the annotation in  `MultiGrid.iter_cell_list_contents` which happened to be wrong)

Tortar: I think that you are right, so I made everything consistent.

Tortar: But right now I'm thinking that the strategy with the custom filter could be actually doable, because in any case the overhead created is just the same as using the default case right now and then filtering from the outside, so we would have the same speed in the custom case, but it would be faster in the default one, with also the added functionality

Tortar: @rht I think this one as it is now doesn't change the behaviour of the methods, improving them (2 in performance 1 in homogeneity with the other methods), and so could be merged as it is, the filter functionality (which I would like to hear your opinion about) can be done in another pr.

mesa/space.py
```diff
@@ -1082,4 +1080,8 @@ def iter_cell_list_contents(self, cell_list: list[int]) -> Iterator[Agent]:
         """Returns an iterator of the agents contained in the nodes identified
         in `cell_list`; nodes with empty content are excluded.
         """
-        yield from self.get_cell_list_contents(cell_list)
+        return itertools.chain.from_iterable(
+            self.G.nodes[node_id]["agent"]
+            for node_id in cell_list
+            if not self.is_cell_empty(node_id)
```
rht, line 1086: Is there a particular reason why this is not `filterfalse`?

Tortar: slightly better in terms of performance because nodes are not passed differently in `self.is_cell_empty` and when accessing the agents in them (while in the other classes this happens)

rht: Simplicity takes precedence if the perf benefit is small. Let's keep the implementation consistent instead. And I need a up-to-date list of the perf improvement diff relative to main branch to be included in the commit message. I suppose you only need to re-measure for the network grid.

Tortar: Implemented with filterfalse. I recalculated the perf improvement and they are actually smaller than what It seemed to be before, probably something went wrong in the previous perf tests: 1.3x for MultiGrid and more or less the same as before for NetworkGrid (but now the iterator is a real iterator, before it wasn't). The filter functionality will make all of them much better, maybe I will try to implement it in another PR :-)

rht:
LGTM

PR #1573, opened 2022-12-31
remove already inherited methods from HexGrid

Tortar:
All of these methods are already present in Grid without any change, so they can be removed.


Tortar:
Indeed HexGrid should be entirelt deprecated because it implements only a new `get_neighborhood` but without changing the rectangular topology, this should be only a parameter of the `get_neighborhood` function in Grid.

Tortar:
ops, my bad, didn't see the different parameters on the function arguments, tests not failing deceived me :$

PR #1575, opened 2023-01-02
Make Grid class private

Tortar:
As discussed I made the Grid class private and removed from it the methods `place_agent` and `remove_agent`, so that they are implemented only in the child public classes. 

Things still to do:

1. Implement SingleHexGrid and MultiHexGrid classes (and make HexGrid class private maybe) -> Deferred to some breaking change PR for mesa 2.0
2. Change tests and class docstrings -> Done

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -79,7 +79,7 @@ def is_integer(x: Real) -> bool:
     return isinstance(x, (int, np.integer))
 
 
-class Grid:
+class _Grid:
     """Base class for a rectangular grid.
 
     Grid cells are indexed by [x][y], where [0][0] is assumed to be the
@@ -435,24 +435,6 @@ class Grid:
         self.remove_agent(agent)
         self.place_agent(agent, pos)
 
-    def place_agent(self, agent: Agent, pos: Coordinate) -> None:
-        """Place the agent at the specified location, and set its pos variable."""
-        x, y = pos
-        self._grid[x][y] = agent
-        if self._empties_built:
-            self._empties.discard(pos)
-        agent.pos = pos
-
-    def remove_agent(self, agent: Agent) -> None:
-        """Remove the agent from the grid and set its pos attribute to None."""
-        if (pos := agent.pos) is None:
-            return
-        x, y = pos
-        self._grid[x][y] = self.default_val()
-        if self._empties_built:
-            self._empties.add(pos)
-        agent.pos = None
-
     def swap_pos(self, agent_a: Agent, agent_b: Agent) -> None:
         """Swap agents positions"""
         agents_no_pos = []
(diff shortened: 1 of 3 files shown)


Tortar:
It should be ready for review!

mesa/space.py
```diff
@@ -435,24 +435,6 @@ def move_agent(self, agent: Agent, pos: Coordinate) -> None:
         self.remove_agent(agent)
         self.place_agent(agent, pos)
 
-    def place_agent(self, agent: Agent, pos: Coordinate) -> None:
```
rht, line 438: I think it should be an abstract method that raises `NotImplementedError` here.

rht: Actually, should be just `...`. It's an abstract method after all, and to be consistent with other methods.

mesa/space.py
```diff
@@ -537,8 +519,17 @@ def exists_empty_cells(self) -> bool:
         return len(self.empties) > 0
 
 
-class SingleGrid(Grid):
-    """Grid where each cell contains exactly at most one object."""
+class SingleGrid(_Grid):
+    """Rectangular grid where each cell contains exactly at most one agent.
+
+    Grid cells are indexed by [x][y], where [0][0] is assumed to be the
```
rht, line 525: While we are on it, we should recommend `[x, y]` instead.

mesa/space.py
```diff
@@ -643,7 +644,7 @@ def iter_cell_list_contents(
         )
 
 
-class HexGrid(Grid):
+class _HexGrid(_Grid):
```
rht, line 647: This will break people's code, and should be deferred to 2.0.

rht: We probably have enough breaking changes for 2.0. Maybe the next release should be 2.0? Let's discuss on Jan 7th meeting.

rht: Regardless, the PR should contain commits that have the same theme. Implementing `SingleHexGrid`and `MultiHexGrid` should be done in a separate PR. This will screw with the changelog otherwise.

Tortar: right, I made HexGrid inherit from SingleGrid so that it place agents correctly

mesa/space.py
Tortar: done!

mesa/space.py
Tortar: ok, used the `...` syntax

rht:
LGTM

EwoutH:
While I agree that the baseclass Grid is better suited as a private class, why was this conversion not properly deprecated? This is currently breaking examples we're still using to teach students, and it was removed between version 1.1.1 and 1.2.0 without any warning.

Maybe we need proper deprecation policy for future breaking changes. @tpike3 maybe we can add a deprecation policy in the governance document you're working on in #1836?

rht:
There is already semver in place. This was merged with 1.1.1 going to 2.0 in mind, but at some point it became 1.2.0 instead.

PR #1581, opened 2023-01-08
Make HexGrid private and create HexSingleGrid and HexMultiGrid

Tortar:
Made HexGrid private and created two new grid classes as discussed in the dev meeting. This is a breaking change to defer until 2.0.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -858,10 +858,16 @@ class HexGrid(HexSingleGrid):
         width, height: The grid's width and height.
         torus: Boolean which determines whether to treat the grid as a torus.
     """
+
     def __init__(self, width: int, height: int, torus: bool) -> None:
         super().__init__(width, height, torus)
-        warn(("HexGrid is being deprecated; use instead HexSingleGrid or HexMultiGrid "
-              "depending on your use case."), DeprecationWarning)
+        warn(
+            (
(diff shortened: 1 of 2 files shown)


jeremander:
To ease backward compatibility, can you keep `HexGrid` and deprecate it?  For example, have `HexGrid` subclass `HexSingleGrid` and issue a `DeprecationWarning` upon construction?

Tortar:
This seems like a good idea :+1:

Tortar:
But I have not much time to work on this now, also it should be done with care I think since if we issue it in `__init__` we will overwrite the `__init__` method in the superclass so this should be managed somehow

Tortar:
Actually did it

Tortar:
I don't see any merge conflict @rht , however I resolved one sometime ago

rht:
I have rebased the branch. RTD build has a failure that is unrelated to this PR
```
ImportError: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with OpenSSL 1.0.2n  7 Dec 2017. See: https://github.com/urllib3/urllib3/issues/2168
```

@EwoutH any idea about the situation?

rht:
Merged, thank you!

PR #1587, opened 2023-01-21
ci: Replace flake8 with Ruff

rht:
See https://notes.crmarsh.com/ruff-the-first-200-releases. Highlight:

> Probably the most complex and rewarding deployment (at least for me) thus far came from [Dagster](https://github.com/dagster-io/dagster), where they replaced 70 parallel CI jobs (to run Pylint) with a single pre-commit hook. It runs in 400 milliseconds, a ~1000x speed-up over their previous setup.
> It’s a perfect encapsulation of the impact that “faster tools” can have on a project, on a team, on a company. Ruff isn’t just faster; it’s so much faster that it completely changes the workflow.

@wang-boyu we should do this for Mesa-Geo. @EwoutH check this out.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,8 +1,0 @@
-[flake8]
-# Ignore list taken from https://github.com/psf/black/blob/master/.flake8
-# E203	Whitespace before ':'
-# E266	Too many leading '#' for block comment
-# E501	Line too long (82 > 79 characters)
-# W503	Line break occurred before a binary operator
-ignore = E203, E266, E501, W503
-exclude = docs, build
--- a/.github/workflows/build_lint.yml
+++ b/.github/workflows/build_lint.yml
@@ -60,7 +60,7 @@ jobs:
       name: Codecov
       uses: codecov/codecov-action@v3
 
-  lint-flake:
+  lint-ruff:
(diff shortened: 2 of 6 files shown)


rht:
It took 2 s to run flake8 on the codebase, and 1 s for Black. Now GH Actions says 0 s. Also, it is a single linter instead of separate. We should configure further to include `isort` etc.

rht:
I think it doesn't replace Black yet. It is meant to be used alongside Black.

tpike3:
This LGTM,  however as it is a significant dependency change, could I get a second @wang-boyu, @jackiekazil @EwoutH and then we can merge.

wang-boyu:
Looks great! Probably there'll be more tools written in Rust in the future. I'll update Mesa-Geo with the same.

rht:
Once this is merged, I will add more as configured in [Zulip](https://github.com/zulip/zulip/blob/d05f672132139c3e82a366f0d4c89bd732e736d4/pyproject.toml#L99-L125).

wang-boyu:
LGTM.

jackiekazil:
Ty @rht

PR #1590, opened 2023-01-23
Use getattr for attribute strings in model data collection

Tortar:
The partial conversion in model_reporters slows down the look-up for attributes values, tested with:

```
import timeit
default = timeit.timeit("s(a)", "from functools import partial; a = lambda x: x; a.y=10; _getattr= lambda name, _object: getattr(_object, name, None); s=partial(_getattr, 'y')")
new = timeit.timeit("getattr(a, 'y', None)", "a = lambda x: x; a.y=10")
print(default, new)
```

shows that the new way is almost twice as fast

### DIFF
--- a/mesa/datacollection.py
+++ b/mesa/datacollection.py
@@ -117,8 +117,6 @@ class DataCollector:
             reporter: Attribute string, or function object that returns the
                       variable when given a model instance.
         """
-        if type(reporter) is str:
-            reporter = partial(self._getattr, reporter)
         self.model_reporters[name] = reporter
         self.model_vars[name] = []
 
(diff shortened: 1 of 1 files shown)


mesa/datacollection.py
```diff
@@ -175,8 +173,8 @@ def collect(self, model):
                 if isinstance(reporter, types.LambdaType):
                     self.model_vars[var].append(reporter(model))
                 # Check if model attribute
-                elif isinstance(reporter, partial):
-                    self.model_vars[var].append(reporter(model))
+                elif isinstance(reporter, str):
+                    self.model_vars[var].append(getattr(model, reporter, None))
```
rht, line 177: `self._getattr` is shorter.

Tortar: but it is an indirection we shouldn't make, calling a function which only calls a function is not useful, this is done only for calling partial because it swaps the parameters

rht: I thought the point of this PR is the speedup due to removing the partial. If you include removing `self._getattr`, then this could be the cause of the bulk of the speedup. It's unclear.

Tortar: indeed, I verified that a big part of the speed-up is due to the removal of `self._getattr` :-)

rht: Yes, the commit message & the PR title needs to be changed.

rht: Also, always prefix the perf commit message with "perf:". It's really hard to distinguish from a refactor otherwise.

Tortar: amended +1:

PR #1591, opened 2023-01-24
Remove _reporter_decorator

Tortar:
This is not useful, we can just call reporter


rht:
This is the commit that introduced it: https://github.com/projectmesa/mesa/commit/b0bacf5095c34a1d26f05151206d4df4a9ae5922. In any case, removing the `_reporter_decorator` is strictly better.

PR #1592, opened 2023-01-24
Fix docstring of DataCollector

Tortar:
There was a bug since a @property decorated method is not callable, to be extra safe I used a callable check before that one for retro-compatibility

### DIFF
--- a/mesa/datacollection.py
+++ b/mesa/datacollection.py
@@ -178,7 +178,7 @@ class DataCollector:
                 elif isinstance(reporter, list):
                     self.model_vars[var].append(reporter[0](*reporter[1]))
                 # TODO: Check if method of a class, as of now it is assumed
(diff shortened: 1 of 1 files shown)


Tortar:
My bad, I misunderstood something I think

Tortar:
what do you thing @rht?

mesa/datacollection.py
```diff
@@ -177,6 +177,7 @@ def collect(self, model):
                 # Check if function with arguments
                 elif isinstance(reporter, list):
                     self.model_vars[var].append(reporter[0](*reporter[1]))
+                # Check if method of a class
```
rht, line 180: You meant to add "TODO" here? There doesn't seem to be any check if reporter is at least a callable.

Tortar: I also changed a bit the docstring

PR #1593, opened 2023-01-26
Not describe again the types for overload decorated __getitem__

Tortar:
Redeclare the types if the function uses the overload decorator should be avoided. Though we could also drop the overloaded methods and describe all just in the actual implementation of the function, losing a bit of accuracy in describing the returned value.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -138,7 +138,7 @@ class _Grid:
         self._empties_built = True
 
     @overload
-    def __getitem__(self, index: int) -> list[GridContent]:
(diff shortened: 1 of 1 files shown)


jeremander:
Looks good to me.  `overload` is a bit of a complicated mechanism, but this PR simplifies the code; omitting annotations on the implementation signature is consistent with the example in the `typing` [documentation](https://docs.python.org/3/library/typing.html).

jackiekazil:
After a review and discussion with @jeremander at Pycon 2023 -- moving to merge. 
Thank you @Tortar!

PR #1595, opened 2023-02-03
Create a Numpy MT generator to use in the simulation

Tortar:
This is the product of my exploration with the help and support of @rht on the topic of using different MT19937 random numbers generators in different parts of a simulation, seeding two MT generators with different seeds can sometimes be not enough since it can produce correlated results, see [here](https://scicomp.stackexchange.com/questions/23547/parallel-mersenne-twister-for-monte-carlo) and [here](https://en.wikipedia.org/wiki/Mersenne_Twister#Disadvantages). But there is a remedy to this, as described [in this paper](https://doi.org/10.1287/ijoc.1070.0251) by the original creator of the Mersenne Twister, from which I cite some relevant parts:

> Random number generators (RNGs) with multiple disjoint streams and substreams are an important component of any good general-purpose simulation or statistical software. they are very handy, for example, to obtain parallel RNGs and to support the implementation of variance reduction techniques (Hellekalek 1998, Kelton 2006, Law and Kelton 2000, L’Ecuyer et al. 2002). The most convenient way of getting these streams and substreams is to start with a backbone RNG having a huge period and partition its output sequence into long disjoint subse quences and subsubsequences whose starting points are at equidistant lags (Law and Kelton 2000, L’Ecuyer 1990, L’Ecuyer and Côté 1991). When a new stream is needed, we find its starting point by jumping ahead from the starting point of the current subsequence to the starting point of the next one. [...] We have developed a viable jump-ahead algorithm for large linear RNGs over 2 . With this technique, one can easily implement RNG packages with multiple streams and substreams, based on long-period generators such as the Mersenne twister and the WELL with a period length of 2^19937 − 1. For these generators, jumping ahead takes a few milliseconds with the proposed method. This is still significantly slower than for the MRG32k3a generator in L’Ecuyer et al. (2002), whose jump-ahead time is a few microseconds. On the other hand, MRG32k3a is slower to generate its numbers, by a factor of two or three on common 32-bit computers, and has a much shorter period length. For applications where jumping ahead is not required too frequently and where a fast long-period RNG is desired, the new jump-ahead algorithm comes in very handy.

So we can jump the state of one generator by an amount unreachable by the other so that extracting numbers from two generators is like having only one, since if it was true that the jumped generator has correlation with the non-jumped one, then also extracting numbers from only one generator would have correlation. We can use 1 jump since it shift the state as if 2^128 random numbers had been generated.

### DIFF
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -8,6 +8,7 @@ Core Objects: Model
 from __future__ import annotations
 
 import random
+import numpy as np
 
 from mesa.datacollection import DataCollector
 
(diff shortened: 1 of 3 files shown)


Tortar:
@rht and I discussed the topic and arrived to the conclusion that it is a bit too involved to have two rngs even if there a solution to manage them, so I close this one, but if anyone think that instead it can be good I can reopen it

PR #1596, opened 2023-02-05
ruff: Add more rules based on Zulip's config

rht:
@wang-boyu this is the last one to be added to the list in https://github.com/projectmesa/mesa-geo/issues/131.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,5 +10,5 @@ select = ["E", "F", "I"]
 # See https://github.com/charliermarsh/ruff/issues/1842#issuecomment-1381210185
 extend-ignore = ["E501"]
 extend-exclude = ["docs", "build"]
-# Hardcode to Python 3.10.
-target-version = "py310"
+# Hardcode to Python 3.8.
+target-version = "py38"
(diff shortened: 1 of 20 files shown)


wang-boyu:
> @wang-boyu this is the last one to be added to the list in [projectmesa/mesa-geo#131](https://github.com/projectmesa/mesa-geo/issues/131).

Got it. Thanks!

rht:
I updated the earliest target of Ruff to be 3.8, because I enabled the `pyupgrade` linter.

rht:
There was a merge conflict due to the pre-commit CI PR. But I have fixed it. @tpike3 this PR is ready for review. I'm not adding anything more.

pyproject.toml
```diff
@@ -1,5 +1,29 @@
 [tool.ruff]
-select = ["E", "F", "I"]
+select = [
```
tpike3, line 2: similar to [Zulip](https://github.com/zulip/zulip/blob/201debc1ea28a489b1c235d4936405b11d78d4b4/pyproject.toml#L100) and the extend-ignore comments below (line 27) can you add the reference to the annotations. 

Also for purely out of curiosity why did you use Zulip as the model and not Pandas or FastAPI?

rht: Added. 2 reasons: 1. because Zulip's config is more comprehensive, 2. I heard about Ruff from Zulip devs.

tpike3: Ver cool, thanks!

PR #1597, opened 2023-02-05
perf: faster is_integer function for common cases

Tortar:
Changing the `is_integer` function like this __getitem__ is faster(1.3x) to access a position as `grid[x, y]`


rht:
Black lint failed.

rht:
You need to rebase your branch for the Black stuff. Should be fixed in main already.

rht:
Make sure the commit message uses imperative sentence.

rht:
I think we need to add a comment explaining why we define the integer types, i.e. for perf purpose. Also, needs a rebase.

Tortar:
I added a comment and rebased. But maybe I messed a bit the history, I don't find a way to squash the commits right now. Can you do it?

PR #1605, opened 2023-03-04
Update resources in README

tpike3:
Per 4 March Dev session update README to updates available resources for using Mesa


rht:
LGTM

PR #1609, opened 2023-03-09
Fix Ruff errors and pin Ruff version

rht:


rht:
@wang-boyu we should pin the Ruff version in Mesa-Geo as well, so that we are not caught by unexpected new lint errors.

wang-boyu:
> @wang-boyu we should pin the Ruff version in Mesa-Geo as well, so that we are not caught by unexpected new lint errors.

Sure, I'll look into it.

tpike3:
Changes LGTM, but for the failing pre-commit.ci failure. Don't you need to commit the changes (trailing whitespace) that were made so pre-commit.ci passes (in this case) before merge. 

Can you also add messages for commit history on updates to main to fix cookiecutter and fixes to to barchart visualization.

rht:
Whitespaces fixed.

> Can you also add messages for commit history on updates to main to fix cookiecutter and fixes to to barchart visualization.

This is not a release to bump the version to 1.2.1.

tpike3:
LGTM

PR #1618, opened 2023-03-18
Fix history formatting.

jackiekazil:
Minor formatting issues that I found after the merge. 
I don't think we need to recut a release for this, since this is mostly viewed thru the github interface, but open to push back if others feel passionate about it.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,15 +3,16 @@
 Release History
 ---------------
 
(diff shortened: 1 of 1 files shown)


rht:
LGTM. I used Markdown syntax by accident.

jackiekazil:
> LGTM. I used Markdown syntax by accident.

No worries & Ty.

PR #1619, opened 2023-03-21
ContiniousSpace: Speedup `get_neighbors` by using KDTree

EwoutH:
Speedup the `get_neighbors()` function in `ContiniousSpace` by using a [SciPy KDTree](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.KDTree.html). It improves performance especially with a large number of agents. For small number of agents, the performance is roughly the same, without large regressions.

I've run some benchmarks on the boid_flockers model, with different number of agents, measuring the total runtime in seconds.

boxplot

Please review carefully, both changes in functionality/behaviour as performance.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ from codecs import open
 
 from setuptools import find_packages, setup
 
-requires = ["click", "cookiecutter", "networkx", "numpy", "pandas", "tornado", "tqdm"]
+requires = ["click", "cookiecutter", "networkx", "numpy", "pandas", "scipy", "tornado", "tqdm"]
 
 extras_require = {
     "dev": [


EwoutH:
We didn't have SciPy as an dependency before, so I don't know if that's an acceptable addition.

Tortar:
I actually evaluated the possibility to use KDTree sometime ago instead of the current implementation, unfortunately if I recall well they are really really slower to build and update, in this case this is not showing because flockers has a fixed number of agents so this doesn't apply, but a lot of performance degradation will happen if removal and adding of agents will happen. As a side note, I think that you didn't update the KDTree when agents are added, this would result in wrong calculations on the neighbors. But anyway this algorithm is not a good choice for the reasons mentione above.

EwoutH:
Thanks for this insights, unfortunate to hear but good to know.

I will close this PR.

PR #1620, opened 2023-03-21
ContiniousSpace: Speedup `get_heading`

EwoutH:
Speedup the get_heading calculations. This reduces the runtime for 500 boid_flockers runs from about 144ms (6.1% of runtime) to 83ms (3.6%).

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -864,6 +864,7 @@ class ContinuousSpace:
         self.height = y_max - y_min
         self.center = np.array(((x_max + x_min) / 2, (y_max + y_min) / 2))
         self.size = np.array((self.width, self.height))
+        self.half_size = self.size / 2
         self.torus = torus
 
         self._agent_points: npt.NDArray[FloatCoordinate] | None = None
@@ -962,15 +963,12 @@ class ContinuousSpace:
         Args:
             pos_1, pos_2: Coordinate tuples for both points.
         """
-        one = np.array(pos_1)
-        two = np.array(pos_2)
+        heading = np.asarray(pos_2) - np.asarray(pos_1)
+
         if self.torus:
-            one = (one - self.center) % self.size
-            two = (two - self.center) % self.size
-        heading = two - one
-        if isinstance(pos_1, tuple):
-            heading = tuple(heading)
-        return heading
+            heading = (heading + self.half_size) % self.half_size
+
+        return tuple(heading)
 
     def get_distance(self, pos_1: FloatCoordinate, pos_2: FloatCoordinate) -> float:
         """Get the distance between two point, accounting for toroidal space.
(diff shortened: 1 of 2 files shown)


mesa/space.py
```diff
@@ -962,15 +963,12 @@ def get_heading(
...
+        heading = np.asarray(pos_2) - np.asarray(pos_1)
+
         if self.torus:
-            one = (one - self.center) % self.size
-            two = (two - self.center) % self.size
-        heading = two - one
-        if isinstance(pos_1, tuple):
-            heading = tuple(heading)
-        return heading
+            heading = (heading + self.half_size) % self.half_size
```
rht, line 969: Does this mean `heading` is always < `self.half_size` for both axes? It doesn't seem to be the case for the old version.

EwoutH: Good catch, I missed the calculation order so I simplified `self.size - self.half_size` to `self.half_size`, but that shouldn't be the case, because of course the modulo is done first and only after that the subtraction.

EwoutH:
Fixed, and also added some more unittests.

mesa/space.py
```diff
@@ -962,15 +963,12 @@ def get_heading(
...
+        heading = np.asarray(pos_2) - np.asarray(pos_1)
+
         if self.torus:
-            one = (one - self.center) % self.size
-            two = (two - self.center) % self.size
-        heading = two - one
-        if isinstance(pos_1, tuple):
-            heading = tuple(heading)
-        return heading
+            heading = (heading + self.half_size) % self.size - self.half_size
```
rht, line 969: @EwoutH I was wondering what is the reason for this being this way instead of `heading % self.size` for the torus wrapping, if you have an explanation.

EwoutH: Yes, that's correct, it's for handeling torus spaces correctly.

EwoutH: This implementation ensures that the heading vector wraps around the edges of the toroidal space correctly, taking into account the shortest distance between the two points. In a toroidal space, each dimension wraps around like a circular loop, so the shortest distance between two points may involve wrapping around the edge.

By adding `self.half_size` to the `heading`, the space is translated so that the center of the torus is at the origin. This makes it easier to perform the modulo operation.

Next, the modulo operation with `% self.size` is applied. This operation ensures that the heading vector stays within the bounds of the toroidal space, wrapping around when it exceeds the size.

Finally, subtracting `self.half_size` translates the space back to its original position.

Simply using `heading % self.size` would not correctly account for the shortest distance between two points in a toroidal space. It would wrap around the edges, but it wouldn't consider cases where the shortest path involves going across the edge, leading to incorrect heading vectors.

jeremander: They differ but are still equivalent modulo `size`.

To my eye, this reimplementation seems to be correctly calculating the heading vector.

rht:
Superseded by #1686, which is more correct.

PR #1637, opened 2023-04-24
change other example locations

smubita:
Similar change - changed three other locations to point at new examples repo.

Also removed dollar sign $ from code blocks so copied commands can be run without editing.


jackiekazil:
LGTM

PR #1646, opened 2023-04-24
Handle stop server

jatinkhilnani:
Handle keyboard interrupt gracefully to end visualization.


rht:
Thank you; I didn't notice the traceback that gets printed out during the keyboard interrupt.

jackiekazil:
LGTM

PR #1652, opened 2023-04-25
Revert "Modify Intro Tutorial"

jackiekazil:
Reverts projectmesa/mesa#1650

We are rolling this back to resolve out of the docs branch.


jackiekazil:
Codespell errors are from image text, not spellings. Also, this is a revert, not a new add.

PR #1664, opened 2023-04-25
Correct toroidal heading - Issue #943

catherinedevlin:
Supersedes [PR 944](https://github.com/projectmesa/mesa/pull/944) - same effect with simpler code.  Both vectors (wrapped and non-wrapped) are valid, but the smaller-magnitude of the two is the desired one.

Fix for https://github.com/projectmesa/mesa/issues/943


jackiekazil:
@catherinedevlin one last request - can you squash your commits?

rht:
Superseded by #1686, which is more correct.

PR #1666, opened 2023-04-26
DO NOT MERGE: Add optional title feature to ChartModule

catherinedevlin:
Closes https://github.com/projectmesa/mesa/issues/954

Supersedes https://github.com/projectmesa/mesa/pull/955 - that code is fine, but old enough that rebasing with the current main seems hopeless.  

Also, it includes an example that needs to be added to mesa-examples in a separate PR.  When that PR is ready, will remove the "DO NOT MERGE" on this one.

### DIFF
--- a/docs/issue_954.rst
+++ b/docs/issue_954.rst
@@ -1,0 +1,32 @@
+issue: Optional title in ChartModule #954
+
+mesa.visualization.modules package
+==================================
+
+Submodules
+----------
+
+mesa.visualization.modules.ChartVisualization module
+----------------------------------------------------
+Extra parameter "chart_title" added to ChartModule __init__() .
+
+mesa\visualization\templates\js\ChartModule.js
+----------------------------------------------------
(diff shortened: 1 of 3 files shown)


catherinedevlin:
I'm not able to make this work after all.

An experiment in https://github.com/projectmesa/mesa-examples/blob/main/examples/charts/charts/server.py is showing 
`
elements.push(new ChartModule([{"Label": "Rich", "Color": "#2ca02c"}, {"Label": "Poor", "Color": "#d62728"}, {"Label": "Middle Class", "Color": "#1f77b4"}], 500,  200, 'sample line chart title'));`

showing up in the javascript of the running example, but no resulting title text, and I don't know enough JavaScript to figure out why.

Also, even if it worked, it looks not only like ChartVisualization.py, but also BarChartVisualization.py and PieChartVisualization.py (mabye everything in visualization/modules) would need corresponding changes.

Issue #1661, opened 2023-04-25
Fix codespell build for tutorial docs


jackiekazil:
Not sure if this is possible, but codespell will pick up image text and try to suggest text edits and says the build is broken. 
Hoping there is a setting somewhere that we can add to ignore the image text. 

Here is an example: https://github.com/projectmesa/mesa/pull/1656/files#diff-732a7eb31fbbd25075ee965837dc0092fb53c2cb64c068e60392d9bf69d9f9bbR877

<img width="1052" alt="Screenshot 2023-04-25 at 12 17 29 PM" src="https://user-images.githubusercontent.com/166734/234368855-d62da5a6-aff8-46f1-bae5-bcf8aa05607f.png">

<img width="708" alt="Screenshot 2023-04-25 at 12 26 06 PM" src="https://user-images.githubusercontent.com/166734/234369720-91c22dbb-f7b9-4b9c-8764-832eb6e63cad.png">

jackiekazil:
Build is fixed... not ideal, but fixed. Codespell did not do what we wanted it to. 
Thank you @catherinedevlin @martin-martin @glnnlhmn for all scrubbing in. 

@rht @tpike3 @wang-boyu - we should discuss during may meeting.

rht:
I looked around if anyone had done this:
- FastAI doesn't spellcheck their notebooks (yet): https://github.com/fastai/fastbook
- PyTorch tutorial does, but with `pyspelling` instead: https://github.com/pytorch/tutorials/blob/44bcbdc2dd349765077c767d433196832e262777/.github/workflows/spelling.yml#L17

If the issue with ignoring png blobs is fixed, it would be a solution that can be applied to other repos containing Jupyter notebooks.

Also worth trying: https://github.com/crate-ci/typos, which is much faster than `codespell`.

jackiekazil:
For tracking -- the ticket was moved here: https://github.com/codespell-project/actions-codespell/issues/42

rht:
Since #1694, the .ipynb outputs are always cleared, not committed to the repo. As such, there should be no more base64-encoded pngs in the repo. We should be able to re-enable codespell again.

jackiekazil:
@rht  --- I have a test branch for another ticket and it looks like the fix for notebooks broke the build - https://github.com/projectmesa/mesa/pull/1696/files

rht:
For some reason the output for intro_tutorial.ipynb hasn't been cleared. Will clear it now.

[...]

PR #1669, opened 2023-04-26
Try different argument --regex

martin-martin:
--ignore-regex IGNORE_REGEX
                        regular expression that is used to find patterns to ignore by treating as whitespace. When writing
                        regular expressions, consider ensuring there are boundary non-word chars, e.g., "\bmatch\b".
                        Defaults to empty/disabled.

...

  -r REGEX, --regex REGEX
                        regular expression that is used to find words. By default any alphanumeric character, the
                        underscore, the hyphen, and the apostrophe is used to build words. This option cannot be specified
                        together with --write-changes.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -16,4 +16,4 @@ jobs:
       with:
         ignore_words_file: .codespellignore
         skip: .*bootstrap.*,*.js,.*bootstrap-theme.css.map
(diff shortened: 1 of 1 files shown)


jackiekazil:
related #1661

jackiekazil:
This wasn't supposed to be merged... going to revert to what was working.

PR #1670, opened 2023-04-26
Exclude notebook files from linter

martin-martin:
Unfortunately, `codespell` doesn't pass on [the existing `--ignore-regex` flag](https://github.com/codespell-project/codespell/blob/master/codespell_lib/_codespell.py#L358) through the GitHub actions YAML file.


jackiekazil:
Related: #1661

PR #1674, opened 2023-04-26
Update slider to start at 1 FPS.

ItsQuinnMoore:
The FPS slider now starts at 1 FPS as 0 FPS is unnecessary with a stop button.


rht:
Tested. LGTM

Issue #481, opened 2018-03-04
Make Mesa easily deployable to a server


jackiekazil:
We have talked about making Mesa easily deployable to a server, but we are wondering what that means. We would like someone to test this and write up a "how to" deploy mesa to server.

Things that might help... 
- Example of mesa on a web server -- research.havven.io/#; https://github.com/Havven/simulation 
- https://github.com/projectmesa/mesa/issues/84 <-- At one point we talked about a settings file, but this really only made sense if we were deploying it to a server. I am not sure if we need it, because we could just use run.py.

rht:
I would assume that this is a web server that is accessible from either a local internet, or web public. The simplest solution would be to run a model server in a Docker instance. With this, it is easy for an nginx reverse proxy to make the server reachable from the internet.

HimathX:
I can work on this and have a potential fix in mind. Let me know if I should proceed.

jackiekazil:
@HimathX thanks for chiming it -- can you write up a proposal in discussions -- https://github.com/mesa/mesa/discussions/new?category=future-of-mesa 

I think this is one to discussion / propose before a lot of work is done and someone says - this isn't going to work for the project.  

Looking forward to seeing your ideas.

HimathX:
Thanks, @jackiekazil  That makes total sense. I'll draft up a deployment proposal and post it in the Discussions section shortly, then link it back here.

HimathX:
Hi @jackiekazil , thanks for the guidance! I've drafted the architectural proposal and posted it in the Discussions here: https://github.com/mesa/mesa/discussions/3371

Looking forward to hearing the team's thoughts before I start on the PR!

tpike3:
Closing to make issues more impactful and relevant to current Mesa issues (anything before Mesa 3.0 release Nov 9, 2024) per May 26, 2026 dev meeting.

PR #1675, opened 2023-04-26
feat: stop ModularServer closes #1642

### LINKED ISSUE #1642: Stop `ModularServer` asynchronously
Only way to stop `ModularServer` is when control is with the user in the terminal and pressing 'Control+c'. When running through IDE, control is no longer with user, there is no direct way to stop the server. Running visualization code throws error, 'Address already in use'.

Add a `stop()` method to `ModularServer` class.

**Additional context**
N/A

reka:
Implementation
-------------------

Python:

* ModularServer: add `stop` method
* SocketHandler: handle message type "shut_down" (only in debug mode)

UI: Add "Shut Down" button

* Disable the nav items.
* Add Bootstrap alert element.
* Send "shut_down" message to WebSocket.

Other
-------

SocketHandler.on_message: add error handling for message without a 'type' key

### DIFF
--- a/mesa/visualization/ModularVisualization.py
+++ b/mesa/visualization/ModularVisualization.py
@@ -223,19 +223,24 @@ class SocketHandler(tornado.websocket.WebSocketHandler):
         if self.application.verbose:
             print(message)
         msg = tornado.escape.json_decode(message)
+        try:
+            msg_type = msg["type"]
+        except KeyError:
+            print("Unexpected message. Key 'type' missing.")
+            return
 
-        if msg["type"] == "get_step":
+        if msg_type == "get_step":
             if not self.application.model.running:
                 self.write_message({"type": "end"})
             else:
                 self.application.model.step()
                 self.write_message(self.viz_state_message)
 
-        elif msg["type"] == "reset":
+        elif msg_type == "reset":
             self.application.reset_model()
             self.write_message(self.viz_state_message)
 
-        elif msg["type"] == "submit_params":
+        elif msg_type == "submit_params":
             param = msg["param"]
             value = msg["value"]
 
@@ -245,6 +250,10 @@ class SocketHandler(tornado.websocket.WebSocketHandler):
                     self.application.model_kwargs[param].value = value
                 else:
                     self.application.model_kwargs[param] = value
+
+        elif msg_type == "shut_down":
+            if self.application.is_debug_mode():
+                self.application.stop()
 
         else:
             if self.application.verbose:
@@ -407,7 +416,13 @@ class ModularServer(tornado.web.Application):
         try:
             tornado.ioloop.IOLoop.current().start()
         except KeyboardInterrupt:
-            tornado.ioloop.IOLoop.current().stop()
+            self.stop()
+
+    def stop(self):
+        tornado.ioloop.IOLoop.current().stop()
+
+    def is_debug_mode(self):
+        return bool(self.settings["debug"])
 
     @staticmethod
     def _is_stylesheet(filename):
--- a/mesa/visualization/ModularVisualization.py
+++ b/mesa/visualization/ModularVisualization.py
@@ -219,14 +219,22 @@ class SocketHandler(tornado.websocket.WebSocketHandler):
         return {"type": "viz_state", "data": self.application.render_model()}
 
     def on_message(self, message):
-        """Receiving a message from the websocket, parse, and act accordingly."""
+        """Receiving a message from the websocket, parse, and act accordingly.
+
+        Args:
+            message (dict): A dictionary containing a key 'type'.
+        """
         if self.application.verbose:
             print(message)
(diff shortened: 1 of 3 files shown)


reka:
Here is a 20-second video showing the "Shut Down" button.
[screencast_shut_down_button_2023-04-26T18-47-34.webm](https://user-images.githubusercontent.com/382113/234680414-f4e9beec-f626-42e5-a712-d8a9bd477c4b.webm)

rht:
@reka I have tested your PR and it works!

mesa/visualization/ModularVisualization.py
```diff
@@ -223,19 +223,24 @@ def on_message(self, message):
         if self.application.verbose:
             print(message)
         msg = tornado.escape.json_decode(message)
+        try:
+            msg_type = msg["type"]
+        except KeyError:
+            print("Unexpected message. Key 'type' missing.")
```
rht, line 229: I haven't encountered this branch point yet.
I think the safeguard is not necessary. Printing this error message instead of raising the exception results in an information loss about the location in which the error happens.
Replacing with `msg_type = msg["type"]` should do.

mesa/visualization/ModularVisualization.py
```diff
@@ -246,6 +251,10 @@ def on_message(self, message):
                 else:
                     self.application.model_kwargs[param] = value
 
+        elif msg_type == "shut_down":
```
rht, line 254: Why is it necessary to be in debug mode in order for the shutdown to be functional?

mesa/visualization/ModularVisualization.py
reka: > I haven't encountered this branch point yet.

I ran into this, as I was trying out various changes in `mesa/visualization/templates/js/runcontrol.js`.
When I sent a message that didn't contain a key "type" 

=>

* The SocketHandler threw an error
* The Tornado server shut down.

If we add an error handling, the server will continue running despite of the unexpected message.
I'm not sure whether this is a good or bad thing.

> Printing this error message instead of raising the exception results in an information loss about the location in which the error happens.

That's a good point.
Is there more info that should be included in the error message?
Or do you think that raising the exception and shutting down the server is the way to go?

mesa/visualization/ModularVisualization.py
reka: My way of thoughts was:
* It's generally not a good idea to allow a client to shut down a web server.
* In a local environment, in debug mode, this might be OK.

By the way, I have a general question:
Do I understand correctly that this Tornado server is expected to run only in a local environment, on the user's computer?
Or is it also deployed somewhere as an actual web server? :thinking:

reka:
> @reka I have tested your PR and it works!

@rht 

Thanks a lot. :+1: 

What are the next steps after we've discussed the 2 open questions and I've improved the coverage?

mesa/visualization/ModularVisualization.py
rht: > When I sent a message that didn't contain a key "type"

This is a change that is made by the developer instead of the user, as such, I believe the server should halt violently with tracebacks instead of continue running.

mesa/visualization/ModularVisualization.py
rht: > Or is it also deployed somewhere as an actual web server? :thinking:

Some people might have done this, that I am not aware of. The relevant issue is #481. We are also currently exploring the possibility of having the UI stack fully in Python (https://github.com/projectmesa/mesa/discussions/1622), in a way that can be run on a Jupyter notebook. Hosting Jupyter notebooks is so ubiquitous that there are canned ways to do it securely.

rht: I am in the opinion that if the pure Python implementation goes well, it might replace the current clunky Tornado server, in a way that is easy to extend, and based on existing established data interaction frameworks. cc: @ankitk50.

Another past attempt at revamping the frontend: #1334 by @Corvince.

rht:
We have shipped Mesa 2.1, and it includes an experimental Solara-based viz frontend that can either run standalone or in a Jupyter notebook. See the usage in https://mesa.readthedocs.io/en/stable/tutorials/visualization_tutorial.html. Your feedback would be appreciated.

rht:
https://mesa.readthedocs.io/en/stable/tutorials/visualization_tutorial.html doesn't tell how to do it standalone. See https://github.com/projectmesa/mesa-examples/tree/main/examples/boltzmann_wealth_model_experimental#how-to-run instead.

Corvince:
Since we moved vie tornado stuff to its own repo I am going to close this issue

But @reka  feel free to re-open a PR at https://github.com/projectmesa/mesa-viz-tornado

I think this change should still be implemented. The only blocker for me is that it seems like the shut-down button is always visible, even if not in debug-mode. This should be fixed, otherwise it is very bad UX to always see a button you can never use ;)

Issue #1678, opened 2023-04-27
JS part of advanced tutorial tries to assign to a `const`


electric-souperman:
**Describe the bug**
In the advanced tutorial in the docs (`docs/tutorials/adv_tutorial.rst`), the chart is initialized as: 

```Javascript
const chart = new Chart(context, {type: 'bar', data: data, options: options});
```

Further on, the `this.reset` method contains the line:
```Javascript
chart = new Chart(context, {type: 'bar', data: data, options: options});
```

Attempting to change the value of a constant results in an `Uncaught TypeError` which stops the visualisations from working correctly:

![image](https://user-images.githubusercontent.com/63853472/234949806-02e3894b-d428-4e3c-86a2-04c70827c5f1.png)

![image](https://user-images.githubusercontent.com/63853472/234950297-7cb5cea2-ffa7-4b4c-9d15-d9640e23c634.png)

![image](https://user-images.githubusercontent.com/63853472/234950689-d17a4916-2d4c-4b1e-8e81-9ac63fe086f4.png)

**Expected behavior**
Switching `const` for `var` fixed the visualisations - as you can see below, and didn't results in any errors being logged to the console. I would therefore propose to change this syntax in the tutorial.

![image](https://user-images.githubusercontent.com/63853472/234951562-9d312917-125b-49bd-b7f0-7abcccd794d3.png)

**To Reproduce**
The issue is with a tutorial, so following the tutorial as it appears on the main branch should be sufficient to reproduce the issue.

**Additional context**
Switching `const` for `var` in this tutorial fixed the issue.

rht:
Good catch! Fixed by #1679.

PR #1679, opened 2023-04-27
Replacing `const chart` for `var chart` in advanced tutorial JS

electric-souperman:
I raised in #1678 that the JS part of the advanced tutorial declares `chart` as a `const`, and subsequently tries to assign a new value to it. This breaks the visualizations for users following along with the tutorial. Declaring `chart` as a `var` instead solves this issue and removes the `Uncaught TypeError` previously logged in the console.

### DIFF
--- a/docs/tutorials/adv_tutorial.rst
+++ b/docs/tutorials/adv_tutorial.rst
@@ -412,7 +412,7 @@ created, we can create the chart object.
       };
 
       // Create the chart object
-      const chart = new Chart(context, {type: 'bar', data: data, options: options});
+      let chart = new Chart(context, {type: 'bar', data: data, options: options});
 
       // Now what?
     };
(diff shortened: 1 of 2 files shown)


docs/tutorials/adv_tutorial.rst
```diff
@@ -412,7 +412,7 @@ created, we can create the chart object.
       };
 
       // Create the chart object
-      const chart = new Chart(context, {type: 'bar', data: data, options: options});
+      var chart = new Chart(context, {type: 'bar', data: data, options: options});
```
rht, line 415: `let` should be safer from redefinition.

electric-souperman:
Ok, I'll change it to `let` when I get the chance.

rht:
No worries. I had edited your PR by replacing the `var` with `let`, doing the same in adv_tutorial.ipynb, and amending the commit message to be imperative and to say `let` instead.

rht:
Merged, thank you!

PR #1673, opened 2023-04-26
Arbitrary chart variable assignment by @gotkaren's PR714

catherinedevlin:
Supports passing arbitrary chart variables to the Chart Module.

Supersedes https://github.com/projectmesa/mesa/pull/714 by @gotkaren.  I apologize that I wasn't able to preserve any of her commits, but there have been too many changes since then and I wasn't able to unsnarl git.

I don't know if we have a way to test JavaScript, but I informally tried some examples mesa-examples; they didn't break, and I was able to change chart visual parameters.  For instance, changed https://github.com/projectmesa/mesa-examples/blob/68c8e87c8e9b0eed5cf87419bcec4de5839ff5a1/examples/forest_fire/forest_fire/server.py#L23 to

`    [{"Label": label, "Color": color, "borderWidth": 10, "pointRadius": 10} for (label, color) in COLORS.items()]`

and got the desired effect.

I didn't make any changes to the docs - as far as I can see, they already imply that you ought to be able to submit arbitrary parameters to a chart, so this PR just brings the code into sync with the docs.

PR #714, opened 2019-07-23
Modified Chart Module in order to add dinamically named properties.

gotkaren:
i.e borderWidth and pointRadius

schelling_chart = ChartModule([{"label": "happy", 
                                                    "borderColor": "Black", 
                                                    "borderWidth": "3", 
                                                    "pointRadius": "0"}],
                                                  data_collector_name="datacollector")

Before this it was only possible to modifiy label and borderColor for each serie.

PR #1685, opened 2023-05-09
Modify ChartModule series to allow dynamically named properties

rht:
Original work: https://github.com/projectmesa/mesa/pull/714 https://github.com/projectmesa/mesa/pull/1673


rht:
@tpike3 @jackiekazil my mistake for breaking the protocol and merging this PR by myself. Was too focused on clearing the unmerged PR counts that I merged my own PR. I think we should turn the main branch into protected branch?

rht:
Though technically, for this case, this was a re-open PR for #1664, because I wasn't able to edit that PR.

tpike3:
It's fine.. pretty straight forward change so no reason to create more work. But yes agree we should make main a protected branch.

PR #1686, opened 2023-05-09
Correct get_heading for toroidal space

rht:
This is an improved version of #1664 where each dimension is optimized independently.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -1006,11 +1006,21 @@ class ContinuousSpace:
         """
         one = np.array(pos_1)
         two = np.array(pos_2)
-        if self.torus:
-            one = (one - self.center) % self.size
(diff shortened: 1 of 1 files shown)


jackiekazil:
LGTM

EwoutH:
Did you measure performance on this PR? The `get_heading()` implementation in #1620 was almost twice as fast, but how fast is this one?

EwoutH:
Also, why weren't the additional tests from #1620 included? Those tests where explicitly written to test and check the unintended behaviour we were trying to fix with this PR.

rht:
This is based on the Agents.jl implementation https://github.com/projectmesa/mesa/discussions/1632#discussioncomment-5837207.

This has clearer logic than #1620, because each dimension is optimized separately.
The tests indeed should be updated.

PR #1689, opened 2023-05-09
breaking: Remove deprecated BatchRunner

rht:
Pending to update overview.rst to use `batch_run`.


docs/overview.rst
tpike3, line 1: I can't believe we didn't remove this.

tpike3 (approved):
LGTM

PR #1692, opened 2023-05-11
Remove Pipfile and Pipfile.lock

rht:
The Pipfile and Pipfile.lock are no longer supported for the upcoming Mesa 2.0. And the files haven't been updated in a long time anyway.


jackiekazil:
LGTM.

Issue #1228, opened 2022-04-03
Easier way to run Mesa examples


rht:
Consider if we can do this for Mesa examples:
```python
import mesa.examples

schelling = mesa.examples.Schelling
schelling.ensure_dependencies()  # will pip install the requirements
schelling.run()
# Without GUI
schelling.run(gui=False)
```

jackiekazil:
I like this.

tpike3:
Closing to make issues more impactful and relevant to current Mesa issues (anything before Mesa 3.0 release Nov 9, 2024) per May 26, 2026 dev meeting.

PR #1698, opened 2023-05-15
adv_tutorial: Add Binder and Colab badge

### LINKED ISSUE #1363: Docs: Remove .rst from docs for simpler build
Use of `.ipynb` transitioned to `.rst` for the Sphinx build can result in desync from readthedocs build and`.ipynb` files and large diff files.

We need to update to using just `.ipynb` with [Sphinx](https://docs.readthedocs.io/en/stable/guides/jupyter.html). 

Discussed in #1361

rht:
Fixes #1363

I added Colab because the Binder version is rather slow with the dependencies (will get even slower with Streamlit/Panel).


tpike3:
> 
What if we package up Moneymodel as a PyPI and then do pip install quiet for both? I think we will still need to test it because the jupyter collides with the server, but COLAB may not have that issue.  If there server doesn't launch then I would say we just do the intro tutorial until we get the python front end. I don't know? What are your thoughts?

rht:
One complication is that as the list of models grows, this becomes a code duplication of the code in mesa-examples, but in a separate repo. I think this is still an acceptable trade-off, regardless. The use case I imagined was a quick demo in a classroom setting. So this is only for simple models. If this is the case, maybe we should just bundle the small models together with `mesa`.

tpike3:
> > What if we package up Moneymodel as a PyPI and then do pip install quiet for both?
> 
> I had a similar idea in #1228. As of now, it has 2 more use cases: 1. adv tutorial, 2. @ankitk50's Streamlit example. Let's start small with a repo (Mesa-Models?) that contains only the MoneyModel.
> 
> ```
> !pip install mesa-models
> # Contains compute_gini, MoneyAgent, and MoneyModel
> from mesa_models import boltzmann_wealth_model as bwm
> bwm.MoneyModel
> ```

I think this is great idea!

rht:
See #1700.

rht:
In the notebook, I added `%pip install git+https://github.com/projectmesa/mesa-examples`. However, it's not usable since the advanced tutorial showcases the visualization server, which can't be run in a Jupyter notebook. We really need to nail the pure Python frontend first.

tpike3:
> In the notebook, I added `%pip install git+https://github.com/projectmesa/mesa-examples`. However, it's not usable since the advanced tutorial showcases the visualization server, which can't be run in a Jupyter notebook. We really need to nail the pure Python frontend first.

Concur

rht:
Further development in #1726

PR #1700, opened 2023-05-16
feat: Introduce mesa.models

rht:
Addresses #1228.
This is so that people can easily reuse these canned models, for experimentation, classroom demo, etc.
For now, it contains only boltzmann_wealth_model.

### DIFF
--- a/mesa/models/boltzmann_wealth_model.py
+++ b/mesa/models/boltzmann_wealth_model.py
@@ -1,0 +1,78 @@
+# Important: always sync the content of this file with
+# https://github.com/projectmesa/mesa-examples/blob/main/examples/boltzmann_wealth_model/boltzmann_wealth_model/model.py
+import mesa
+
+
+def compute_gini(model):
+    agent_wealths = [agent.wealth for agent in model.schedule.agents]
+    x = sorted(agent_wealths)
+    N = model.num_agents  # noqa: N806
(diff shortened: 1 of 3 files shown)


tpike3:
This is awesome! I see where you are going with this. Would we want to put mesa.models in mesa folder or do it as part of examples? The idea being there would be a library called mesa_models so if they wanted to demo they could just install the mesa_models library which would alleviate the duplication complication](https://github.com/projectmesa/mesa/pull/1698#issuecomment-1549326138) and then this would be a subset of mesa-examples.    What do you think?

tpike3 (commented):
See [comment](https://github.com/projectmesa/mesa/pull/1700#issuecomment-1550506568)

rht:
mesa-examples would be fine only if select a subset of the examples to be in the Python package. Some of the GIS examples have data files that would make the wheel large, though won't get to 50 MB large. The problem is that then we will have to maintain one more additional PyPI package.

rht:
I suppose maintaining one more PyPI package is less cumbersome than maintaining copies of the examples.

ankitk50:
> I suppose maintaining one more PyPI package is less cumbersome than maintaining copies of the examples.

And later of course we will keep adding more examples to the package.

rht:
Superseded by https://github.com/projectmesa/mesa-examples/pull/35

Issue #348, opened 2017-01-05
Multi-Agent data collection


Corvince:
Dear all,

This is actually how I came across this issue: I wanted to activate different agent types sequentially (but both randomly), so I used two different schedulers, but this broke the data collection.  Currently, agent reporters get their agents hard-coded from model.scheduler.agents, assuming it exists and failing if your scheduler is named differently. One way to fix this would be to (optionally?) supply the scheduler to the DataCollector.

The downside is that if you have agents that are not part of any schedule you still can't collect data for them. That's already a problem right now, so it wouldn't worsen the situation, but maybe someone has a better long-term solution to this?

Also, if you use the same scheduler, there seems to be no way to collect data from _different_ agents. If you for example want to count the wealth of some of your agents, but not all agents have a wealth attribute, it fails.

dmasad:
That's an interesting question. 

My inclination is to say that since the scheduler keeps track of time in a model, models should have just one scheduler. Check out the [custom scheduler in the Wolf-Sheep example ](https://github.com/projectmesa/mesa/blob/master/examples/wolf_sheep/wolf_sheep/schedule.py) (if you haven't already) to see how you can activate different types of agents in different orders.

That doesn't solve the problem of how to handle data collection for heterogeneous attributes, though. The easy but ugly way is to give all agents the attribute, and just set some of them to 0, or None, or some similar 'N/A' value.

I think the better way would be to give the DataCollector a default 'collect attribute' behavior, which would also let us get away from some of the ugliness with passing lambdas, etc. As part of that, the DataCollector would ensure that agents had the appropriate attribute, and only collect it if they did.

Corvince:
Aha! I did not look into the example, because I thought the order is not important for the wolf-sheep example. And I did not think about implementing a custom scheduler. But after reconsidering I strongly agree that there should be exactly one scheduler per model. 

Regarding the DataCollector, a "collect_attribute" would indeed be nice and more intuitive than the lambda functions. But I would still lean towards additionally defining the agent-type. Staying with the Wolf-Sheep example, one might be interested only in the position of the wolves, but querying the pos-attribute would still query all agents.

Corvince:
To advance on this, in the datacollection module I modified the `_new_agent_reporter` function and added the `_collect_attribute` function as follows:

```python
def _new_agent_reporter(self, reporter_name, reporter_function=None):
    """ Add a new agent-level reporter to collect.

    Args:
        reporter_name: Name of the agent-level variable to collect.
[...]

PR #1701, opened 2023-05-19
Collecting data from multiple differentiable classes of agents

GiveMeMoreData:
Curent version of mesa did allow for capturing data from multiple classes, but collected variables had to be present in all agents, which for large number of agent classes and variables, results in messy and unnecessary code. Modification present in this PR allows to write collection rules for specific class of agents. This changes the current code from 

```python
class MockAgent(Agent):
    def __init__(self, unique_id, model, val=0):
        self.val = val
        self.val2 = None # used only in DifferentMockAgent


class DifferentMockAgent(Agent):
    def __init__(self, unique_id, model, val=0):
        self.val = None # used only in MockAgent
        self.val2 = val

agent_reporters = {"value": "val", "value2": "val2"}
```

to 

```python
class MockAgent(Agent):
    def __init__(self, unique_id, model, val=0):
        self.val = val


class DifferentMockAgent(Agent):
    def __init__(self, unique_id, model, val=0):
        self.val2 = val

agent_reporters = { MockAgent: {"value": "val"}, DifferentMockAgent: {"value2": "val2"}}
```
Agent specific variables are collected in separate dictionary, which can later be retrieved with `get_agent_specific_vars_dataframe` as dataframe copyting the bahaviour of  `get_agent_vars_dataframe`, but with addition of one index column containg agent class info. 

```python
        agent_vars = data_collector.get_agent_vars_dataframe()
        specific_agent_vars = data_collector.get_agent_specific_vars_dataframe()
```


GiveMeMoreData:
pre-commit.ci autofix

rht:
This gives me an idea for a much simpler solution. Basically, modify https://github.com/projectmesa/mesa/blob/e3af2a508080d343eb44f07c5fe5cc9f487a19aa/mesa/datacollection.py#L163
to
```python
agent_records = []
for agent in model.schedule.agents:
    record = get_reports(agent)
    record = tuple(r for r in record if r is not None)
    agent_records.append(record)
```
You do `agent_reporters={"value": "val", "value2": "val2"}`. @tpike3 what do you think of this solution for trading Sugarscape?

rht:
In my solution, you can have multiple different agents with the same attribute name to be collected together.

tpike3:
> > In my solution, you can have multiple different agent classes with the same attribute name to be collected together.
> 
> Yea, but what about classes with different attributes, which is the sole reason for this PR? If you see possible changes in overall collection process, thats great, but for simplicity lets focus on the matter in hand.

This is awesome!! This has been a big gap (#1419) and if we do a quick release to PyPI, I can  incorporate into the Complexity Explorer tutorial for sugarscape. 

To @rht's point there may be a more elegant way to do this, but it has been brought up that Mesa needs to be easier to contribute to (Thank you @EwoutH for your candid feedback) and per open source etiquette we must be polite and positive (this why @jackiekazil  pinged me to get on the code of conduct)

I say we merge and then release Mesa 1.2.2. What do you say @rht?

rht:
The reason why I said my proposed solution in the first place is because it solves the problem brought up in this PR (classes with different attributes) in a much simpler way. I'd prefer to base on Occam's razor to have the codebase not contain complicated solution if possible.

rht:
> but it has been brought up that Mesa needs to be easier to contribute to (Thank you @EwoutH for your candid feedback) and per open source etiquette we must be polite and positive (this is why @jackiekazil pinged me to get on the code of conduct)

I don't think I have crossed any line regarding with the code of conduct. I was just pointing out a different way to solve the problem.

rht:
I suppose I need to elaborate on why my solution solves classes with different attributes. With `agent_reporters={"value": "val", "value2": "val2"}`, if class A has only `val`, and class B has only `val2`, they will be collected into `value` and `value2` separately. In another situation, where you want `val` and `val2` to be collected into `value`, you can do a custom lambda function `lambda agent: agent.val if isinstance(agent, A) else agent.val2`.

GiveMeMoreData:
> I suppose I need to elaborate on why my solution solves classes with different attributes. With `agent_reporters={"value": "val", "value2": "val2"}`, if class A has only `val`, and class B has only `val2`, they will be collected into `value` and `value2` separately. In another situation, where you want `val` and `val2` to be collected into `value`, you can do a custom lambda function `lambda agent: agent.val if isinstance(agent, A) else agent.val2`.

Okey, I might be missing something, but the get_report created from `agent_reporters={"value": "val", "value2": "val2"}` is in both cases `attrgetter` and it returns an AttributeError when you are trying to get an nonexisting attribute of a class. 

Even after your proposed change this code still results in an error

```python
from mesa import Agent, Model
from mesa.time import BaseScheduler

class MockAgent(Agent):
    def __init__(self, unique_id, model, val=0):
        self.val = val


class DifferentMockAgent(Agent):
    def __init__(self, unique_id, model, val=0):
        self.val2 = val


class MockModel(Model):
    schedule = BaseScheduler(None)

    def __init__(self):
        self.schedule = BaseScheduler(self)

        self.schedule.add(MockAgent(1, self))
        self.schedule.add(DifferentMockAgent(2, self))
        
        self.initialize_data_collector(
            agent_reporters={"value": "val", "value2": "val2"}
        )
        
    def step(self):
        self.schedule.step()
        
model = MockModel() 
model.step()
    
agent_vars = model.datacollector.get_agent_vars_dataframe()
```

Corvince:
> agent_reporters = { MockAgent: {"value": "val"}, DifferentMockAgent: {"value2": "val2"}}
> ```

I just wanted to say that I love the semantics here. Using the agent classes directly as an identifier is just so explicit that it should be obvious for what it does. Awesome

tpike3:
> 

@GiveMeMoreData  Take a look #1701. I believe this addresses your concerns but let us know.

Corvince:
Maybe we should revisit this PR after the discussions in #348 . It seems to be rather additive to #1702 rather than superseded by that PR.

Corvince:
> > I suppose I need to elaborate on why my solution solves classes with different attributes. With `agent_reporters={"value": "val", "value2": "val2"}`, if class A has only `val`, and class B has only `val2`, they will be collected into `value` and `value2` separately. In another situation, where you want `val` and `val2` to be collected into `value`, you can do a custom lambda function `lambda agent: agent.val if isinstance(agent, A) else agent.val2`.
> 
> Okey, I might be missing something, but the get_report created from `agent_reporters={"value": "val", "value2": "val2"}` is in both cases `attrgetter` and it returns an AttributeError when you are trying to get an nonexisting attribute of a class.

I think maybe it would be best to remove the performance optimization from `attrgetter`, i.e. the lines 

```python
        if all(hasattr(rep, "attribute_name") for rep in rep_funcs):
            # This branch is for performance optimization purpose.
            prefix = ["model.schedule.steps", "unique_id"]
            attributes = [func.attribute_name for func in rep_funcs]
            get_reports = attrgetter(*prefix + attributes)`
```
I think this would ease multi agent data collection, albeit with None values, which could be removed with the option added in #1701, if we find a way to fix it. 

Should i open such a PR?

rht:
Yeah, the performance optimization is not worth the added opaqueness of the code.

PR #1703, opened 2023-05-21
refactor: Simplify _new_agent_reporter

rht:
In addition to simplifying the code (no need to import and consult definition for `partial`), it is also 13% faster.
Benchmark code
```python
import timeit

repetition = 1000


def print_elapsed(label, setup, stmt):
    _elapsed = timeit.timeit(stmt, setup, number=repetition) * 10**6 / repetition
    elapsed = "{:.3f} μs".format(_elapsed)
    print(label, elapsed)
    return _elapsed

common_setup = """
class Foo:
    __slots__ = ("a", "b", "c")
    def __init__(self):
        self.a = 0
        self.b = 1
        self.c = 2
obj = Foo()
attrs = ("a", "b", "c")
"""

# getattr
setup = common_setup + """
def fn():
    tuple(getattr(obj, i, None) for i in attrs)
"""
print_elapsed("getattr", setup, "fn()")

# attrgetter
setup = common_setup + """
from operator import attrgetter

def fn():
    attrgetter(*attrs)(obj)
"""
print_elapsed("attrgetter", setup, "fn()")

# Mesa default
setup = common_setup + """
from functools import partial

def _getattr(name, _object):
    return getattr(_object, name, None)
def _new_reporter(attr_name):
    return partial(_getattr, attr_name)
reporters = [_new_reporter(i) for i in attrs]
def fn():
    tuple(r(obj) for r in reporters)
"""
print_elapsed("Mesa default", setup, "fn()")

# Mesa faster
setup = common_setup + """
reporters = [lambda o: getattr(o, n, None) for n in attrs]
def fn():
    tuple(r(obj) for r in reporters)
"""
print_elapsed("Mesa faster", setup, "fn()")
```
Output:
```
getattr 2.726 μs
attrgetter 1.009 μs
Mesa default 3.815 μs
Mesa faster 3.336 μs
```


Corvince:
Cool, but lambda functions are not pickleable, so I don't think this is the best solution.

rht:
It's not a blocker, you can pickle with `dill` if you want to serialize a model. Besides, users generally provide lambda functions for the reporters anyway. `batch_run`'s multiprocessing mode still works with this PR.

Corvince:
Easy, it's not a blocker, but something to consider. This is only marginally faster, but would introduce a breaking change so I am not sure if it is worth it.

rht:
I have replaced the lambda with a local function instead, so no more PEP 8 violation. `BatchrunnerMP` has been marked as deprecated and is going to be removed in this release, given that the next release is 2.0.

rht:
I think the value of this PR is in the code clarity and simplification, more than the performance.

rht:
To purge remaining concerns:
> BatchrunnerMP has been marked as deprecated and is going to be removed in this release, given that the next release is 2.0.

Even if `BatchrunnerMP` is not going to be deprecated, this PR doesn't break backward compatibility either. Again, there are lambda functions from the user-specified reporters. Saying that you *can* pickle the model in the case when you specify all the reporters as strings -- I would consider this as relying on hidden, undocumented feature.

Corvince:
That's a bit awkward to say, because I have several workflows that depend on pickleable models. For example I always pickle my models at the end of a run so I can analyze or continue them later. So a lambda function would definitely be a breaking change.

But nonetheless your solution without the lambda is definitely better than the current approach. Just curiously asking: is it still faster?

Corvince (approved):

Issue #1709, opened 2023-06-03
Requesting for a feature of model level functions to be included in StagedActivation class


subhamonsey:
**What's the problem this feature will solve?**

Right now the Staged Activation class only allows for functions executed at Agent level. However, one can well want to implement a function which is executed at model level (for ex, in an Economy, adjust the wage rate of all labourers, who are agents and share the same wage rate). Hence suggesting writing "model." before the function name in the object stage_list to indicate its a model level function.

**Describe the solution you'd like**
Solution is straightforward as will be suggested via a PR shortly.

tpike3:
Merged @subhamonsey  pull request on this. Thankyou!

PR #1710, opened 2023-06-03
Added a feature in mesa.time StagedActivation class

subhamonsey:
Suggested solution to issue #1709

Hi team Mesa, 

Right now the Staged Activation class only allows for functions executed at Agent level. However, one can well want to implement a function which is executed at model level (for ex, in an Economy, adjust the wage rate of all labourers, who are agents and share the same wage rate). Hence suggesting writing "model." before the function name in the object stage_list to indicate its a model level function.

### DIFF
--- a/tests/test_time.py
+++ b/tests/test_time.py
@@ -75,7 +75,7 @@ class MockModel(Model):
 
         # Make scheduler
         if activation == STAGED:
-            model_stages = ["stage_one", "stage_two"]
+            model_stages = ["stage_one", "model.model_stage", "stage_two"]
             self.schedule = StagedActivation(self, model_stages, shuffle=shuffle)
         elif activation == RANDOM:
             self.schedule = RandomActivation(self)
@@ -94,13 +94,16 @@ class MockModel(Model):
     def step(self):
         self.schedule.step()
 
+    def model_stage(self):
+        self.log.append("model_stage")
+
 
 class TestStagedActivation(TestCase):
     """
     Test the staged activation.
     """
 
-    expected_output = ["A_1", "B_1", "A_2", "B_2"]
+    expected_output = ["A_1", "B_1", "model_stage", "A_2", "B_2"]
 
     def test_no_shuffle(self):
         """
@@ -109,7 +112,7 @@ class TestStagedActivation(TestCase):
         model = MockModel(shuffle=False)
         model.step()
         model.step()
-        assert all(i == j for i, j in zip(model.log[:4], model.log[4:]))
+        assert all(i == j for i, j in zip(model.log[:5], model.log[5:]))
 
     def test_shuffle(self):
         """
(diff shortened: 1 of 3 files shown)


tpike3 (changes requested):
@subhamonsey This is great thanks! I have no issues with the code very clean and straighforward.

Could you make the following administrative adjustments: 
- make sure you use [Ruff so it passes the formatting test](https://github.com/projectmesa/mesa/blob/main/CONTRIBUTING.rst)
- put your commit message in the imperative so "Add feature in mesa.time StagedActivation"
- could you add a quick [useful snippet](https://github.com/projectmesa/mesa/blob/main/docs/useful-snippets/snippets.rst) maybe right after "Models with Discrete Time"; something like "Implementing Model Level Functions in Staged Activation"

Thanks again for identifying this features and adding it!

rht:
You can simplify the code into:
```python
        for stage in self.stage_list:
            if stage.startswith("model."):
                getattr(self.model, stage[6:])()
            else:
                for agent_key in agent_keys:
                    if agent_key in self._agents:
                        getattr(self._agents[agent_key], stage)()  # Run stage
            # We recompute the keys because some agents might have been removed
            # in the previous loop.
            agent_keys = list(self._agents.keys())
            if self.shuffle_between_stages:
                self.model.random.shuffle(agent_keys)
            self.time += self.stage_time
```

rht:
A different way of doing staged activation without abstraction:
```python
# Model step
def step(self):
    self.each_agents("stage1")
    self.shuffle_agents()
    self.adjust_wages()
    self.each_agents("stage2")
    self.shuffle_agents()
```

The `self.time` is not always needed in most cases. Sometimes. having fewer abstractions means it is more customizable, that the Mesa library's structure constraint doesn't get in the way of the user's model needs.

rht:
Random activation:
```python
# Model step
def step(self):
    self.shuffle_agents()
    self.do_each("step")
```

Simultaneous activation:
```python
def step(self):
    self.do_each("step")
    self.do_each("advance")
```

Random activation by type:
```python
def step(self):
    self.shuffle_agents()
    for agent_class in self.agent_types:
        self.shuffle_agents(agent_class)
        self.do_each("step", key=agent_class)
```

subhamonsey:
@tpike3 , @rht thanks for the helpful suggestions. I implemented the requested changes and simplified the code. Could you please run the workflows again?

rht:
One last thing, I think it needs a test case. The new code branch is not yet covered.

subhamonsey:
Hey, I added unit tests and rebased my commit over the latest changes. Please could you run the workflows again? Thank you.

tests/test_time.py
```diff
@@ -119,8 +122,9 @@ def test_shuffle(self):
         model.step()
         for output in self.expected_output[:2]:
             assert output in model.log[:2]
-        for output in self.expected_output[2:]:
-            assert output in model.log[2:]
+        for output in self.expected_output[3:]:
```
rht, line 125: I suppose the existing code isn't clear about saying that it checks that the shuffling happens within stages, and not across stages.

rht (approved):

tpike3 (approved):

tpike3:
Thanks @subhamonsey

PR #1711, opened 2023-06-04
feat: Implement do_each

rht:
This simplifies and clarifies the step logic of the existing schedulers implementation. And makes it easier to write custom schedulers.

### DIFF
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -28,7 +28,7 @@ from __future__ import annotations
 from collections import defaultdict
 
 # mypy
-from typing import Iterator, Union
+from typing import Union
 
 from mesa.agent import Agent
 from mesa.model import Model
@@ -78,8 +78,9 @@ class BaseScheduler:
 
     def step(self) -> None:
         """Execute the step of all the agents, one at a time."""
-        for agent in self.agent_buffer(shuffled=False):
-            agent.step()
+        # To be able to remove and/or add agents during stepping
+        # it's necessary for the keys view to be a list.
+        self.do_each("step")
         self.steps += 1
         self.time += 1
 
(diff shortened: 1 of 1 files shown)


rht:
I think we should deprecate `SimultaneousActivation`. With this PR, it is evident that it is a subset of `StagedActivation`.

Corvince:
Interesting and I agree that it makes writing custom schedulers easier, which is a big plus.

tpike3 (commented):
Interesting. Is there any performance difference?

We could add this to 2.0 with the deprecation warning.

rht:
```ipython
In [1]: import timeit

In [2]: class Foo:
   ...:     pass
   ...: 

In [3]: foo = Foo()

In [4]: foo.bar = "spam"

In [5]: timeit.timeit("getattr(foo, 'bar')", 'from __main__ import foo')
Out[5]: 0.24488897202536464

In [6]: timeit.timeit("foo.bar", 'from __main__ import foo')
Out[6]: 0.13063999998848885
```
About 2x slower. But usually, in a large-scale simulation, the most time consuming part would be the space operations, and the `step()` or stages method themselves. In this case, the readability improvement may outweigh the slight perf degradation.

> We could add this to 2.0 with the deprecation warning.

Did you mean for deprecating `SimultaneousActivation`? Is there an equivalent to this scheduler in MASON or NetLogo or Agents.jl?

rht:
I want to raise that one of Mesa's values is pedagogy, that people are reading its code to learn and understand how to implement their own customized ABM. The Mesa source code is meant to be read by researchers. This may come at odds with (some, not all) performance optimization. But I can see that `getattr` is something that can be optimized in Python >3.11, with the [Faster CPython Project](https://github.com/faster-cpython/ideas).

But Agents.jl [code](https://github.com/JuliaDynamics/ABM_Framework_Comparisons/blob/main/WolfSheep/Agents/WolfSheep.jl) can be both readable and performant.

wang-boyu:
I'm wondering what this means for end users who don't care about the internal implementations of schedulers. Is this a more general way of calling `agent.step()`?

That is, in `model.step()`, instead of doing

```self.schedule.step()``` which always calls `agent.step()`,

we can instead do

```self.schedule.do_each("arbitrary_agent_function")```

while `self.schedule` could be any of random/uniform/simultaneous activation as usual.

rht:
`StagedActivation` has existed for a while.

> On a side note, the name StagedActivation doesn't seem to imply simultaneously activated. What if I want the agents to be randomly or uniformly activated and also have multiple stages?

IIRC the naming comes from the needs to get around of the order advantage problem. The `step` and `advance` stages are written in such a way that the execution order of the agents does not matter. But the naming `step` and `advance` seems to be model-specific.

rht:
> Is this a more general way of calling agent.step()?

The "UI" change of replacing `schedule.step()` (consists of `agent.step()`) with `schedule.do_each("agent_function")` is too drastic of a change that I haven't thought the implication of yet. The staged activation could be made unnecessary, and the time.py library may consist of only `Activation` (covers `BaseScheduler`, `RandomActivation`, `SimultaneousActivation`, `StagedActivation`), and `ActivationByType` (covers `RandomActivationByType`, and the not yet implemented `SimultaneousActivationByType` and `StagedActivationByType`).

rht:
I forgot to say this: thank you @subhamonsey for #1710 that made me think of this idea.

tpike3 (approved):
LGTM

PR #1719, opened 2023-06-17
rtd: Use gruvbox-dark as style

rht:


rht:
Doesn't work because the Sphinx version is too old.

tpike3 (approved):
LGTM

PR #1723, opened 2023-06-21
breaking: Change coord_iter to return tuple[content, pos]

rht:
This PR recreates #1566.


tpike3 (approved):
LGTM. Mesa 2.0 here we come!

PR #1726, opened 2023-06-29
Initialize Solara-based adv_tutorial

rht:


rht:
You can try at https://colab.research.google.com/github/rht/mesa/blob/adv_tutorial_experimental/docs/tutorials/adv_tutorial_experimental.ipynb.

Currently, only the step button works fine. The play button doesn't properly update the grid and the timeseries.

tpike3 (approved):
LGTM

tpike3 (changes requested):
Looks great, but can you commit the lint-black edits so the build passes and then I will merge

rht:
The Colab issue was fixed by https://github.com/projectmesa/mesa-examples/pull/47. I have just added a tutorial to write your custom visualization object via Matplotlib. It's a histogram, just like the current advanced tutorial. It's only 6 lines of code (excluding comments). Compare this with ~35 LOC of JS (client-side) and ~15 LOC of Python (server-side).

rht:
> Compare this with ~35 LOC of JS (client-side) and ~15 LOC of Python (server-side).

Of the current advanced tutorial.

rht:
I have moved the experimental tutorial to be hidden in the "more" part of the menu bar. This is not discoverable at all. One option is to have the experimental tutorial be the default, and the old version to be linked from it instead.

rht:
I have updated so that the Solara-based advanced tutorial is the one that people will see first in the menu bar.

One concern is that people will have to install from the Git repo of mesa-examples in order to install mesa-models. Should we move the Solara viz code to core Mesa? Though it's not a huge problem because it's just an extra `pip install git+https...` away.

rht:
I don't think we should call it advanced tutorial anymore? It no longer requires writing hybrid Python-JS code, and the content is very concise. Maybe "visualization tutorial"?

tpike3:
> I don't think we should call it advanced tutorial anymore? It no longer requires writing hybrid Python-JS code, and the content is very concise. Maybe "visualization tutorial"?

I agree visualization tutorial

tpike3:
> I have updated so that the Solara-based advanced tutorial is the one that people will see first in the menu bar.
> 
> One concern is that people will have to install from the Git repo of mesa-examples in order to install mesa-models. Should we move the Solara viz code to core Mesa? Though it's not a huge problem because it's just an extra `pip install git+https...` away.

Yes, I think we should, this is consistent with using it for the tutorial and providing both JS and python option. Defiitely in the 

> I have updated so that the Solara-based advanced tutorial is the one that people will see first in the menu bar.
> 
> One concern is that people will have to install from the Git repo of mesa-examples in order to install mesa-models. Should we move the Solara viz code to core Mesa? Though it's not a huge problem because it's just an extra `pip install git+https...` away.

Yes, I think we should make it part of core mesa since it is becoming the main approach for the tutorial

rht:
Any objections with moving the Solara viz to core Mesa, @jackiekazil @Corvince ?

Corvince:
> Any objections with moving the Solara viz to core Mesa, @jackiekazil @Corvince ?

No objections from my side, I think solara is the perfect fit for mesa. That said both solara and the current implementation of the JupyterViz class seems quite rough around there edges. So it should definitely stay under experimental, like currently done in mesa_examples. And thus communicate that it is not currently part of our public API and not subject to semver, since I think there will be quite some changes in the near future. 

The alternative to bringing this to mesa would again be a separate mesa_viz package. I don't know how big a dependency is in this regard.

rht:
At this point, the many benefits of having it in core Mesa outweighs the fact that it may still be volatile. It is mainly to expose the viz to many people early on (and that it is convenient to install from `pip install mesa` and no more), so that we have feedbacks sooner. As it is currently, we will only get feedbacks from new users who happen to read the visualization tutorial, not existing users.

@tpike3 is going to do an intro tutorial session at CSSSA. The choice is whether to have the tutorial code does `pip install mesa`, or it does `pip install mesa` and `pip install mesa-models`.

An example feedback/feature request would be to add figure title to the chart output (#954), and the PR #955. It seems to be hard to debug.

Corvince:
Yes i mean the idea has always been to have a mesa meta package. So pip install mesa would get you mesa-core, mesa-examples and mesa-viz. And only an advanced Installation procedure could only install mesa-core or any other mix. So I think from a (standard) user perspective it would be even simpler. 

But that would require some substantial restructuring so I agree nothing for the foreseeable future. But we could later move from mesa.experimental to mesa-viz. If we put this directly into mesa.visualization I am worried it will lead to some false sense of stability.

rht:
Then I will add `mesa-models` as requirements of `mesa`. This means users can automatically import Boltzmann wealth model for quick demo. Depends on https://github.com/projectmesa/mesa-examples/pull/52 to be merged.

tpike3:
> Then I will add `mesa-models` as requirements of `mesa`. This means users can automatically import Boltzmann wealth model for quick demo. Depends on [projectmesa/mesa-examples#52](https://github.com/projectmesa/mesa-examples/pull/52) to be merged.

Done

jackiekazil:
This is intended to go into 2.0 yes? 
Are there any other solara tickets for 2.0?

tpike3:
> This is intended to go into 2.0 yes? Are there any other solara tickets for 2.0?

This should go with 2.0 and this is the only one. However, @rht the build is failing it looks like Solara's the problem, with its use of [pydantic as the problem](https://github.com/widgetti/solara/blob/6b79bfc867386750a040a6070be7a7204f4bc058/solara/settings.py#L14)

What do you think?

rht:
I have raised the issue on Solara Discord. Still waiting for their reply.

rht:
The Solara Pydantic issue has been fixed in https://github.com/widgetti/solara/pull/189, but there is another problem: `mesa-models` depends on `mesa`, and since I added `mesa-models` as a dependency of `mesa`, it is a circular dependency. As there is no `mesa-core` yet, I will have to remove `mesa` from `mesa-models` dependency. To save time, I will push directly the change to mesa-examples without a PR process.

rht:
@tpike3 this is ready to merge.

rht:
Don't forget to update the v2.0 branch or this PR is not incorporated into the release.

jackiekazil:
@rht V2.0 branch wasn't be updated.  When I move to release,I am going to cut a new v2.0 branch from what is in main... for easy access.

jackiekazil:
I am going to merge this. Looks good to me. Thank you @rht for your work on this.

PR #1743, opened 2023-07-26
intro_tutorial: Warn user the Mesa version must be up-to-date

rht:


Corvince:
I'm a bit worried about the maintenance burden of this change. This would need to be updated for every version change. 

The docs are already versioned. While not 100% optimal I think it is ultimately the users responsibility to check the correct version. Otherwise we would need this reminder on top of every documentation page. 

Also >= 2.1 is misleading. We don't know if it will work in 2.x (also it should ) or 3.x

Corvince:
That said some advice might actually be a good idea at least for the intro tutorial. But maybe just rephrase it so that users should check that the doc version and installed version matches.

rht:
I have made the warning to be more generic.

Corvince (approved):

PR #1745, opened 2023-07-29
fix: Remove race condition when pause button sometimes doesn't work

rht:
There is a typo: on_play should be on_playing. As such, the playing=playing.value argument of widgets.Play and `playing.value = True` inside `on_play_value` may sometimes cancel each other.

With this PR, the play/pause should be more robust.


tpike3 (approved):
LGTM

PR #1746, opened 2023-07-31
Move viz stuff to mesa-viz-tornado Git repo

rht:


rht:
This greatly simplifies the setup.py, and we should be able to move to setup.cfg+pyproject.toml very soon.

tpike3 (approved):
So my understanding is #1747 fixes most of the build fails; but what about the lint-ruff failure that the imports are unsorted?

tpike3 (commented):
So my understanding is #1747 fixes the build fails; but what about the lint-ruff fails that the imports are unsorted?

rht:
One main blocker to this PR is that the dependencies once again include a Git repo. mesa-viz-tornado needs to be released as a PyPI package first.

rht:
Another blocker is that the API documentation no longer shows the viz stuff: https://mesa--1746.org.readthedocs.build/en/1746/apis/visualization.html. Either create a separate RTD for mesa-viz-tornado, or just tell people to read the source code directly. Either way, I'd rather spend effort on completing the Solara viz features than on this.

rht:
> Either way, I'd rather spend effort on completing the Solara viz features than on this.

This may sounds like I'm running away from the responsibility of ensuring that the existing viz API documentation persists. I meant it as prioritization for my time allocation.

tpike3:
> > Either way, I'd rather spend effort on completing the Solara viz features than on this.
> 
> This may sound like I'm running away from the responsibility of ensuring that the existing viz API documentation persists. I meant it as prioritization for my time allocation.

It doesn't your fine, we are always running on limited bandwidth we have to prioritize and python front end has been our # 1 feedback so Solara has to be the priority

PR #1751, opened 2023-08-03
perf: Access grid only once

Corvince:
In our current implementation we access the grid first to check if the cell is empty and then a second time to iterate over the contents. Inlining the check and using the walrus operator allows to access the content only once. 

Using the ABM framework comparison i get the following speedups on my laptop for the Schelling model.

| Model | main| this PR|
|--------|--------|--------|
| Schelling (small) | 136ms | 97ms |
| Schelling (large) | 2120ms| 1558ms |

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -394,8 +394,9 @@ class _Grid:
         """
         # iter_cell_list_contents returns only non-empty contents.
         return (
-            self._grid[x][y]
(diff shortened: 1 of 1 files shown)


Corvince:
On the Schelling performance I also opened https://github.com/JuliaDynamics/ABM_Framework_Comparisons/issues/57 which could give us a 3-5x advantage

tpike3 (approved):
This is very cool, thanks!

LGTM

PR #1752, opened 2023-08-04
fix install for visualization tutorial in colab

tpike3:
I removed the `-e` argument in the pip install because for colab that stores the install in `src` directory and presents some other challenges of mesa-example not being imported easily. 

Removing the `-e` installs mesa-examples in the standard colab install location and prevents an `ImportError` without effecting performance if someone is using a local instantiation of their jupyter notebook

### DIFF
--- a/docs/tutorials/adv_tutorial_legacy.ipynb
+++ b/docs/tutorials/adv_tutorial_legacy.ipynb
@@ -9,7 +9,6 @@
    ]
   },
   {
-   "attachments": {},
    "cell_type": "markdown",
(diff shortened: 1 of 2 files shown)


tpike3:
Sorry I also opened the legacy advanced tutorial so their is a bunch of noise of metadata changing but no actual changes; all significant changes for visualization tutorial are at the end

rht (approved):

rht:
I would have not committed the legacy change, but it's fine.

PR #1756, opened 2023-08-07
Drop support for Python 3.8

rht:
NumPy had dropped it since Jan 31 2023: https://numpy.org/neps/nep-0029-deprecation_policy.html.

If this PR is merged, the next release needs to be 2.2.0 instead of 2.1.2.

### DIFF
--- a/docs/tutorials/intro_tutorial.ipynb
+++ b/docs/tutorials/intro_tutorial.ipynb
@@ -60,7 +60,7 @@
    "source": [
     "### Tutorial Setup\n",
     "\n",
-    "Create and activate a [virtual environment](http://docs.python-guide.org/en/latest/dev/virtualenvs/). *Python version 3.8 or higher is required*.\n",
+    "Create and activate a [virtual environment](http://docs.python-guide.org/en/latest/dev/virtualenvs/). *Python version 3.9 or higher is required*.\n",
     "\n",
     "Install Mesa:\n",
     "\n",
--- a/mesa/batchrunner.py
+++ b/mesa/batchrunner.py
@@ -2,11 +2,7 @@ import itertools
 from collections.abc import Iterable, Mapping
 from functools import partial
 from multiprocessing import Pool
-from typing import (
-    Any,
-    Optional,
-    Union,
-)
+from typing import Any, Optional, Union
 
 from tqdm.auto import tqdm
 
--- a/mesa/batchrunner.py
+++ b/mesa/batchrunner.py
@@ -1,16 +1,12 @@
 import itertools
+from collections.abc import Iterable, Mapping
 from functools import partial
 from multiprocessing import Pool
 from typing import (
     Any,
-    Dict,
-    List,
     Optional,
-    Tuple,
-    Type,
     Union,
 )
-from collections.abc import Iterable, Mapping
 
 from tqdm.auto import tqdm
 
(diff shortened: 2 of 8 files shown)


rht:
I think this shouldn't be merged until after the CSSSA conference. We should instead release 2.1.2 with bugfix releases for the Solara viz.

docs/tutorials/intro_tutorial.ipynb
```diff
@@ -60,7 +60,7 @@
    "source": [
     "### Tutorial Setup\n",
     "\n",
-    "Create and activate a [virtual environment](http://docs.python-guide.org/en/latest/dev/virtualenvs/). *Python version 3.8 or higher is required*.\n",
+    "Create and activate a [virtual environment](http://docs.python-guide.org/en/latest/dev/virtualenvs/). *Python version 3.9 or higher is required*.\n",
```
tpike3, line 63: Do we still want to do 3.9 as it seems Solara only works with 3.10 or higher?

Corvince: Quite the opposite, solara still supports Python 3.6. They are quite proud of it, although i cant find the source for that claim right now. But their unit tests still run on Python 3.6 https://github.com/widgetti/solara/blob/master/.github/workflows/unittest.yml

EwoutH: Following [SPEC 0](https://scientific-python.org/specs/spec-0000/), the successor of [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html), Python 3.9 can also be dropped at this point.

That would allow us to start using Structural Pattern Matching (also in examples) and use new typing features. See [What’s New In Python 3.10](https://docs.python.org/3/whatsnew/3.10.html).

Users still on Python 3.8 and 3.9 can of course still keep using the current Mesa versions, as always.

EwoutH:
Will there be another patch release (2.1.6) after the current 2.1.5? If not, we could merge this right and release 2.2.0 in a bit.

EwoutH (changes requested):
You can also update pyupgrade in the pre-commit config.

(in `.pre-commit-config.yaml`, change `--py38-plus` to `--py39-plus`)

tpike3 (approved):
LGTM

tpike3:
@rht could you please resolve the conflicts and then I will merge. Honestly, I am a little confused why it is saying these are conflicts

rht:
Merge conflicts resolved.

mesa/batchrunner.py
```diff
@@ -1,15 +1,10 @@
 import itertools
+from collections.abc import Iterable, Mapping
 from functools import partial
 from multiprocessing import Pool
 from typing import (
```
EwoutH, line 5: can be a single line

mesa/space.py
```diff
@@ -22,15 +22,11 @@
 import itertools
 import math
 import warnings
+from collections.abc import Iterable, Iterator, Sequence
 from numbers import Real
 from typing import (
```
EwoutH, line 27: single line

EwoutH (approved):

mesa/batchrunner.py
rht: Fixed.

mesa/space.py
rht: Fixed.

tpike3 (approved):
LGTM

PR #1758, opened 2023-08-07
fix: Ensure hooks are called at the top level

rht:
This fix is recommended by @mariobuikhuizen at Solara Discord. See https://solara.dev/docs/understanding/rules-of-hooks.

### DIFF
--- a/mesa/experimental/jupyter_viz.py
+++ b/mesa/experimental/jupyter_viz.py
@@ -116,7 +116,10 @@ def make_text(renderer):
     return function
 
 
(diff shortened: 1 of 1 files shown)


rht:
I just tested this. When I change the values in the slider in the Schelling example, the model no longer updates. The fix doesn't work at the moment.

tpike3 (approved):
LGTM

rht:
This needs to be reverted. As I said, it is recommended, but not yet working.

PR #1759, opened 2023-08-07
fix: Allow multiple connections in Solara

rht:


rht:
This PR addresses #856, but only for Solara viz.

tpike3 (approved):
LGTM

PR #1771, opened 2023-08-21
solara: Implement decorator to simplify custom space drawer

rht:


rht:
@tpike3 this is a very convenient feature to have, and would be great to have it in before the 2.1.2 release.

rht:
`solara.FigureMatplotlib(space_fig, dependencies=[viz.model, viz.df])` is a rather arcane line requiring people to understand Solara reactive component. With the indirection, people just need to focus on drawing Matplotlib.

EwoutH:
@rht which direction would you like this PR to go?

rht:
This PR is now obsolete. The current API is already simple enough to not require this wrapper.

Issue #1777, opened 2023-08-24
update JupyterViz code to add support for Solara select input and document supported inputs


rlskoeser:
**What's the problem this feature will solve?**
Only `SliderInt` and `SliderFloat` inputs are currently available, and neither of these are documented in the user-facing documentation.

**Describe the solution you'd like**
Support for solara select and improved documentation.

**Additional context**
Some discussion on #1772 

I have the Solara select mapping implemented and working in a local fork; I will be glad to add some suggested changes to the documentation and open a pull request with these changes.

rht:
Implemented in https://github.com/projectmesa/mesa/pull/1779.

PR #1779, opened 2023-08-25
Add support for select input type

rlskoeser:
Add support for `solara.Select` for model input parameters, per #1777 

I'd be glad to draft documentation or add more detail to one of the tutorials about supported input types, but not sure where to add that.

### DIFF
--- a/mesa/experimental/jupyter_viz.py
+++ b/mesa/experimental/jupyter_viz.py
@@ -150,6 +150,12 @@ def make_user_input(user_input, k, v):
             max=v.get("max"),
             step=v.get("step"),
         )
+    elif v["type"] == "Select":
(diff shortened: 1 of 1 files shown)


rht (approved):

rht:
Merged, thank you @rlskoeser !

rht:
> I'd be glad to draft documentation or add more detail to one of the tutorials about supported input types, but not sure where to add that.

The place to add documentation should be the docstring of `make_user_input`, so that it is automatically displayed in readthedocs.io in the future.

As with tutorial, maybe in an example in https://github.com/projectmesa/mesa-examples that has yet to be ported to Solara. There is `pd_grid`.

Issue #1778, opened 2023-08-24
JupyterViz grid display should be optional, since not all models use a grid or agent space


rlskoeser:
**What's the problem this feature will solve?**
currently JupyterViz breaks if `model.grid` is not defined

**Describe the solution you'd like**
`model.grid` should continue to be optional; space should only be drawn when the model has a grid.

From testing locally, it looks like it is sufficient to modify the `make_space` method to check for the `grid` attribute and do nothing when it is defined.  I'm willing to open a PR with this change if this solution is sufficient.

**Additional context**
Some discussion on https://github.com/projectmesa/mesa/issues/1772

rht:
I can see your point, where some models don't have an explicit grid/space, but my concern is that some users might blunder by forgetting to create a space, and the explicit error/breakage helps to remind them. The other option is to have a flag that lets user explicitly disable drawing space.

rlskoeser:
@rht that sounds fine to me; seems good to require users to explicitly opt out of displaying the space. Could you suggest a simple way to implement this? What about defining an argument for the `JupyterViz` `space_drawer` parameter that would disable it instead of getting the default display? Would `False` work, or is there a better way to explicitly indicate it should be disabled?

rht:
Fixed in #1783.

PR #1783, opened 2023-08-30
Add an option to disable drawing agent space #1778

rlskoeser:
as discussed on #1778

### DIFF
--- a/mesa/experimental/jupyter_viz.py
+++ b/mesa/experimental/jupyter_viz.py
@@ -20,9 +20,23 @@ def JupyterViz(
     measures=None,
     name="Mesa Model",
     agent_portrayal=None,
-    space_drawer=None,
+    space_drawer="default",
     play_interval=400,
 ):
+    """Initialize a component to visualize a model.
+    Args:
+        model_class: class of the model to instantiate
+        model_params: parameters for initializing the model
+        measures: list of callables or data attributes to plot
+        name: name for display
+        agent_portrayal: options for rendering agents (dictionary)
+        space_drawer: method to render the agent space for
+            the model; default implementation is :meth:`make_space`;
+            simulations with no space to visualize should
+            specify `space_drawer=False`
+        play_interval: play interval (default: 400)
+    """
+
     current_step, set_current_step = solara.use_state(0)
 
     solara.Markdown(name)
@@ -48,10 +62,14 @@ def JupyterViz(
 
     with solara.GridFixed(columns=2):
         # 4. Space
-        if space_drawer is None:
+        if space_drawer == "default":
+            # draw with the default implementation
             make_space(model, agent_portrayal)
-        else:
+        elif space_drawer:
+            # if specified, draw agent space with an alternate renderer
             space_drawer(model, agent_portrayal)
+        # otherwise, do nothing (do not draw space)
+
         # 5. Plots
         for measure in measures:
             if callable(measure):
--- a/tests/test_jupyter_viz.py
+++ b/tests/test_jupyter_viz.py
@@ -1,7 +1,9 @@
 import unittest
-from unittest.mock import patch
+from unittest.mock import Mock, patch
 
-from mesa.experimental.jupyter_viz import make_user_input
+import solara
+
+from mesa.experimental.jupyter_viz import JupyterViz, make_user_input
 
 
 class TestMakeUserInput(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


rht:
I have seen `None` and `False` being used as distinct possible values in a parameter in a library somewhere. Can't remember which one. But it's easy to confuse the 2, because `None` is considered falsy. I am not fully sure yet of the best way forward in this situation.

rht:
Maybe a string `"disabled"`? @Corvince thoughts?

mesa/experimental/jupyter_viz.py
```diff
@@ -160,6 +160,16 @@ def make_user_input(user_input, k, v):
...
 def MesaComponent(viz, space_drawer=None, play_interval=400):
+    """Initialize a component to display a mesa simulation.
+
+    Args:
+        viz: a component, as returned by :meth:`JupyterViz`
+        space_drawer: optional method to render the agent space for
+            the model; if not specified, will call :meth:`make_space`
+            to render; simulations with no space to visualize should
+            specify `space_drawer=False`
+        play_interval: play interval (default: 400)
```
tpike3, line 171: Thanks @rlskoeser!

This change is making me wonder if we should switch to a different naming convention as the False, None, Else took me a while to think through, which I worry will confuse users.

Maybe kwargs for different space and not space options?

Thoughts @rlskoeser?
Thoughts @rht?
Thoughts  @ankitk50

Corvince:
I think space_drawer=None reads pretty good and should be unambiguous.

That said I think we shouldn't assume what and how people want to plot there models.

I think it would be easier to let users just decide on which solara components they want to display (and provide some components)

rht:
> I think it would be easier to let users just decide on which solara components they want to display (and provide some components)

The reason why I set the space drawing to be automatic was to cover the 80% of the use case to be easy. The previous Tornado version does require users to specify manually, but I find this too much boilerplate code.

mesa/experimental/jupyter_viz.py
ankitk50: 'None' should also represent absence of a value which seems to be the case here.

rlskoeser:
I was thinking of `None` as the default / unspecified option, but it's plausible that someone would specify `specify_drawer=None` and expect to have no space and I agree that the code isn't as readable.

I was wondering if there was a different way to pass in the default `make_space` method instead of using `None` to flag the default, but when I tried that locally I couldn't pass through the `False` / `None` value; maybe because there's something I'm missing about how the `JupyterViz` method generates a `MesaComponent`.

I thought about a defined constant for either the default or the disabled state, but adding an optional kwarg parameter to disable drawing the space as @tpike3 suggested sounds easier to use and more readable to me.  Although I'm not sure of a brief, readable name for that argument.

rht:
What about passing a do-nothing `lambda: None` function and document this in the docstring?

rht:
As in, if user doesn't want the space to be drawn, they should pass in the empty lambda function.

rht:
Another option is to set `space_drawer=make_space` as the default argument, and if user specifies it with a falsy value (`None`, `False`, 0, etc), then don't draw anything.

rlskoeser:
> Another option is to set `space_drawer=make_space` as the default argument, and if user specifies it with a falsy value (`None`, `False`, 0, etc), then don't draw anything.

I tried this and it didn't work (the `None` turned into default behavior somewhere - I think related to the component creation). I will try again with the refactored jupyterviz code and see if I can get it to work. This seems like the cleanest option to me.

rht:
You could also do `space_drawer="default"` and then replace is with `make_space`, and do nothing for falsy values.

tpike3:
I don't want to stand in the way of merging this as it is an unequivocally an improvement;

The doc build issue was fixed with #1785 

But as I opened a can of worms, any issue merging @rht, @Corvince, @jackiekazil ?

rht:
>  But it's easy to confuse the 2, because None is considered falsy.

I stand by the statement of the problem of using `None` and `False` in a different way, so this PR can't be merged yet.

tpike3:
> You could also do `space_drawer="default"` and then replace is with `make_space`, and do nothing for falsy values.

I am good with this solution

Corvince:
> > You could also do `space_drawer="default"` and then replace is with `make_space`, and do nothing for falsy values.
> 
> I am good with this solution

me too

tpike3:
@rlskoeser This is your PR, so you are lead, let us know your thoughts or if there is any support we can provide.

jackiekazil:
> >>  You could also do `space_drawer="default"` and then replace is with `make_space`, and do nothing for falsy values.
>>  
> > I am good with this solution

> me too

+1

rlskoeser:
Thanks for the input, everyone. I've re-implemented based on the latest refactor of the jupyter viz code with the proposed 'default' option, which does seem to be the simplest solution. I added a docstring for `JupyterViz` (although not sure my description of the `measures` parameter is accurate/helpful) and a minimal unit test that confirms the expected behavior for various options for drawing the space (default, custom, none).

rht:
The `ruff` lint failed.

tests/test_jupyter_viz.py
```diff
@@ -1,7 +1,9 @@
 import unittest
-from unittest.mock import patch
+from unittest.mock import patch, Mock
 
-from mesa.experimental.jupyter_viz import make_user_input
+import solara
+
+from mesa.experimental.jupyter_viz import make_user_input, JupyterViz
```
tpike3, line 6: I think this is a super simple fix ... https://beta.ruff.rs/docs/rules/unsorted-imports/ and the only issue I see

tpike3:
@rlskoeser This is great and thanks for the tests! I think 2 commits is good, one for the improvement and one for the tests, but please squash the lint correction

Again thank you very much!

rht:
It's fine to not squash the commits. We can do "squash and merge" later instead.

rlskoeser:
Sorry I missed the linting error! I guess I've gotten too used to the isort pre-commit hook that I've started using on other projects.

Corrected now, and I amended & squashed the commit history.

jackiekazil:
I am excited about this addition. Thank you for your contribution.  <3 
I haven't had a chance to test, so I will defer to someone else to make the final review/merge.

rht (approved):

Issue #1772, opened 2023-08-22
JupyterViz questions and proposed enhancements


rlskoeser:
Hello and thanks for the great library.

I recognize that the JupyterViz is still experimental, but it seems like a great solution for allowing project collaborators to run simulations in Colab notebooks.  I have questions about using it and also about submitting a PR with some improvements.

I am having some difficulty customizing the display - it took me a bit to figure out that the agent portrayal for `color` needs to be lowercase when previously it was uppercase. I haven't yet figured out how to change the size; previously I was using `r` for radius; I see `size` in the jupyterviz code but setting that in my agent portrayal doesn't seem to have any effect.  I've looked at the code in `mesa.experimental.jupyter_viz` but I'm not clear on how that's generated; I guess it's part of a matplotlib figure?

I also wanted to use a `Select` input for a choice parameter and add an output that displays the current step.  I have a fork of mesa where I've implemented these - should I open a pull request with these changes to get feedback?  (I haven't written any tests or added documentation for the changes yet, but would do if these changes are welcome.)

I'm also interested in adding a reset button similar to the play/stop buttons, but I'm not sure how to programmatically restart the model; currently the only way to reset is to change one of the input parameters.

Another question: when running with jupyterviz, how should I access the model in order to get collected data? I figured out I could get it this way: `page.args[0].model.datacollector.get_model_vars_dataframe()` —  but that doesn't seem ideal!

## related issues
- #1773 / #1775 
- #1776 
-  #1777
- #1778

rht:
Glad to know that the experimental frontend is being used.

> I also wanted to use a Select input for a choice parameter and add an output that displays the current step. I have a fork of mesa where I've implemented these - should I open a pull request with these changes to get feedback? (I haven't written any tests or added documentation for the changes yet, but would do if these changes are welcome.)

You are welcome to do so. Note that the current user input code is not yet documented either: https://github.com/projectmesa/mesa/blob/24a1df30a12a399cc54d8ff2d578db57e0f37a4a/mesa/experimental/jupyter_viz.py#L136-L152. And they have only been tested manually.

> I'm also interested in adding a reset button similar to the play/stop buttons, but I'm not sure how to programmatically restart the model; currently the only way to reset is to change one of the input parameters.

It's easier if I do this instead. You should open an issue for this.

[...]

PR #1784, opened 2023-08-30
Add docstring for jupyterviz make_user_input that documents supported inputs

rlskoeser:
- add a docstring to jupyterviz `make_user_input` method that documents the supported solara inputs
- update parameters to `make_user_input` - drop unused parameter, use more readable variable names

some related conversation on #1772 and follow up work to #1777 

I haven't added any tests yet since there is no existing testing infrastructure for the jupyterviz stuff, but I did test locally with a simulation that uses a slider int and a switch parameter and the changes work


mesa/experimental/jupyter_viz.py
```diff
@@ -133,28 +133,37 @@ def function(viz):
     return function
 
 
-def make_user_input(user_input, k, v):
-    if v["type"] == "SliderInt":
+def make_user_input(user_input, options):
```
rht, line 136: I initially included the name ("k") because I figured it might be used in the future. Looking at the code again, I think the name can be used as a fallback label instead of the current generic "label".

rlskoeser: Ah, ok, that makes sense; I'll revise.

Should the method throw an exception if the type is not supported? If so, does `ValueError` seem appropriate to you?

rlskoeser: @rht updated with name fallback logic, exception for unsupported type, and preliminary unit test for this method

rht:
This PR LGTM, but I'm merging #1786 first because that PR is a huge refactor, will cause a merge conflict, and it is easier to redo this PR instead of that one. Sorry for this.

rht:
There is indeed a merge conflict.

rlskoeser:
@rht thanks for letting me know. I'll see what I can figure out; looking forward to seeing the updates.

rht (approved):

rht:
Merged, thank you @rlskoeser !

PR #1785, opened 2023-08-31
fix: Use .pytemplate for name for cookiecutter

rht:
This avoids the conda install problem.
Fixes https://github.com/projectmesa/mesa/discussions/1774


tpike3 (approved):
LGTM

Issue #1741, opened 2023-07-25
JupyterViz: the default grid space drawer doesn't scale to large size


rht:
**Describe the bug**

Here is Schelling space for 60x60:
![2023-07-24-102909_468x350_scrot](https://github.com/projectmesa/mesa/assets/395821/0949b124-dfeb-48ff-9205-4d242e99c833)

**Expected behavior**

Should either scale down the circle marker size automatically, or scale up the figure size automatically.

rlskoeser:
I vote for scaling up the figure size (which I'm understanding as allocating more space for the agent space figure on the page). Not sure how you would determine the threshold for this - would it have to be user configurable, or could a certain grid size automatically require 3/4 of the width, full width?

rht:
I think we should do both simultaneously.

ankitk50:
I started looking into this-- some upfront ideas here:

1.  If we need to make the figure scalable , I would be better if we make the plots one per row (so that the scaling can be done to the whole screen).

2. I think it would be better to implement it as a test feature that can be enabled via a checkbox. 

3. The agents size looks straight forward to scale based on number of agents. I look into ways to scale the figure. `GridDraggable` in solara looks like a promising feature in this regard.

rht:
There shouldn't be upfront customization from the user unless necessary. This should behave like how LaTeX renders documents, where people focus on the content of the writing, not the layout. Automatically deciding when to use the entire row or not, scaling down marker size, increasing figure size, should be the library's responsibility.

ankitk50:
Sure, I have open a draft PR #1820.

PR #1786, opened 2023-08-31
Simplify solara code

Corvince:
Over the past weeks, I've had the opportunity to delve into the Solara frontend and experiment with different ideas. A big thank you to @rht for laying the foundation of the Solara frontend. 

I'm experimenting with exciting concepts for the frontend and am eager to incorporate my React knowledge.

However, with this PR, my primary aim is to streamline the codebase before introducing any additional modifications.

At present, the frontend code is organized across three functions/classes:
  - `JupyterContainer`: This contains model control functions, grid plotting features, and logic to segregate model parameters into fixed and user-configurable ones.
  - `MesaComponent`: This establishes user-configurable controls, sets up the model, contains some model control functions, performs `JupyterContainer` manipulations, and holds the plotting logic.
  - `JupyterViz`: This Solara component essentially returns the MesaComponent and initiates the Container.

I've reorganized and simplified the structure. Now, `JupyterViz` stands as the primary component, managing user input (subject to further enhancement in an upcoming PR), initializing the model, and handling the plotting logic. The model control has been externalized to a separate component named `ModelController`.

While this PR might seem extensive, it's mainly due to the code repositioning to achieve a more intuitive flow. The actual modifications are minimal. Key changes include:
  - Removal of `FigureMatplotlib` dependencies. Presently, we're free from unnecessary re-renders, so this optimization isn't required.
  - Elimination of `df` from the container. Previously, I believe this was used to trigger re-renders. Now, we track the current step to achieve this. This approach also supports models without a data collector and is essential for upcoming enhancements to the model controller.
  - The `split_model_params` function has been made top-level and devoid of side effects.

I think that sums it up.

### DIFF
--- a/mesa/experimental/jupyter_viz.py
+++ b/mesa/experimental/jupyter_viz.py
@@ -1,6 +1,5 @@
 import threading
 
-import matplotlib.pyplot as plt
 import networkx as nx
 import reacton.ipywidgets as widgets
 import solara
@@ -9,48 +8,72 @@ from matplotlib.ticker import MaxNLocator
 
 import mesa
 
-# Avoid interactive backend
-plt.switch_backend("agg")
 
+@solara.component
+def JupyterViz(
+    model_class,
+    model_params,
+    measures=None,
+    name="Mesa Model",
+    agent_portrayal=None,
+    space_drawer=None,
+    play_interval=400,
+):
+    current_step, set_current_step = solara.use_state(0)
 
-class JupyterContainer:
-    def __init__(
-        self,
-        model_class,
-        model_params,
-        measures=None,
-        name="Mesa Model",
-        agent_portrayal=None,
-    ):
-        self.model_class = model_class
-        self.split_model_params(model_params)
-        self.measures = measures
-        self.name = name
-        self.agent_portrayal = agent_portrayal
-        self.thread = None
+    solara.Markdown(name)
 
-    def split_model_params(self, model_params):
-        self.model_params_input = {}
-        self.model_params_fixed = {}
-        for k, v in model_params.items():
-            if self.check_param_is_fixed(v):
-                self.model_params_fixed[k] = v
+    # 0. Split model params
+    model_params_input, model_params_fixed = split_model_params(model_params)
+
+    # 1. User inputs
+    user_inputs = {}
+    for k, v in model_params_input.items():
+        user_input = solara.use_reactive(v["value"])
+        user_inputs[k] = user_input.value
+        make_user_input(user_input, k, v)
+
+    # 2. Model
+    def make_model():
+        return model_class(**user_inputs, **model_params_fixed)
+
+    model = solara.use_memo(make_model, dependencies=list(user_inputs.values()))
+
+    # 3. Buttons
+    ModelController(model, play_interval, current_step, set_current_step)
+
+    with solara.GridFixed(columns=2):
+        # 4. Space
+        if space_drawer is None:
+            make_space(model, agent_portrayal)
+        else:
+            space_drawer(model, agent_portrayal)
+        # 5. Plots
+        for measure in measures:
+            if callable(measure):
+                # Is a custom object
+                measure(model)
             else:
-                self.model_params_input[k] = v
+                make_plot(model, measure)
 
-    def check_param_is_fixed(self, param):
-        if not isinstance(param, dict):
-            return True
-        if "type" not in param:
-            return True
 
-    def do_step(self):
-        self.model.step()
-        self.set_df(self.model.datacollector.get_model_vars_dataframe())
+@solara.component
+def ModelController(model, play_interval, current_step, set_current_step):
+    playing = solara.use_reactive(False)
+
+    def on_value_play(change):
+        if model.running:
+            do_step()
+        else:
+            playing.value = False
+
+    def do_step():
+        model.step()
+        set_current_step(model.schedule.steps)
 
     def do_play(self):
-        self.model.running = True
-        while self.model.running:
+        model.running = True
+        while model.running:
             self.do_step()
 
     def threaded_do_play(self):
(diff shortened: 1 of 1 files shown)


Corvince:
Just to give some perspective on what "future improvements" from my side are already locally implemented (for testing, need more polishing)
 - Use more components => All the `make_*` functions can and should be converted to solara components (reuseability)
 - Run models using `use_thread` hook - this works on its own, but matplotlib is not thread-safe, so the line-plots crash (much simpler code, but needs some investigating)
 - Move all the input parameter logic into its own component (challenge how to trigger re-initializing of the model)
 - Some layout/styling improvements

mesa/experimental/jupyter_viz.py
```diff
@@ -9,48 +8,72 @@
...
-
-    def split_model_params(self, model_params):
-        self.model_params_input = {}
-        self.model_params_fixed = {}
-        for k, v in model_params.items():
-            if self.check_param_is_fixed(v):
-                self.model_params_fixed[k] = v
+
+@solara.component
+def JupyterViz(
```
rht, line 13: Yeah, this is simpler. The reason why I split into `JupyterContainer`, `MesaComponent`, and `JupyterViz` was the answer to these questions on Discord:
- https://discord.com/channels/1106593685241614489/1106593686223069309/1123620981911597208

The question:
> Is it possible to do just
```python
model_params = {
    "N": {
        "type": "SliderInt",
        "value": 50,
        "label": "Number of agents:",
        "min": 10,
        "max": 100,
        "step": 1,
    },
    "width": 10,
    "height": 10,
}
JupyterViz(
    BoltzmannWealthModel, model_params, measures=["Gini"], name="Money Model"
)
```

The answer,
> @rht you can also assign to page = MesaComponent(...) but they you would have to reverse your model, so JupyterViz shouldn't call MesaComponent, but the other way around

This was the reason for the `MesaComponent` and `JupyterViz` split. But in hindsight, the split wasn't necessary.

This was the reason for the `JupyterContainer` split: https://discord.com/channels/1106593685241614489/1106593686223069309/1123583246307954828.

mesa/experimental/jupyter_viz.py
```diff
@@ -9,48 +8,72 @@
...
+        if model.running:
+            do_step()
+        else:
+            playing.value = False
+
+    def do_step():
+        model.step()
+        set_current_step(model.schedule.steps)
 
     def do_play(self):
```
rht, line 74: This still has `self`.

mesa/experimental/jupyter_viz.py
```diff
@@ -9,48 +8,72 @@
...
-        self.model.running = True
-        while self.model.running:
+        model.running = True
+        while model.running:
             self.do_step()
 
     def threaded_do_play(self):
```
rht, line 79: This shouldn't have `self`.

rht:
> Removal of FigureMatplotlib dependencies. Presently, we're free from unnecessary re-renders, so this optimization isn't required.
> Elimination of df from the container. Previously, I believe this was used to trigger re-renders. Now, we track the current step to achieve this. This approach also supports models without a data collector and is essential for upcoming enhancements to the model controller.

In this PR, only `solara.Markdown(md_text=f"**Step:** {current_step}")` has a dependency to `current_step`, but not the `FigureMatplotlib`'s. How are they notified of the step change then?

mesa/experimental/jupyter_viz.py
```diff
@@ -9,48 +8,72 @@
 
 import mesa
 
-# Avoid interactive backend
-plt.switch_backend("agg")
```
rht, line 13: This was added so that user code that does the pyplot interface (`plt.plot`, etc) can still work, even though not recommended.

mesa/experimental/jupyter_viz.py
Corvince: Yes, the power of hindsight... but still quite impressive to make it work with all the necessary basic features! Solara uses quite a different paradigm, so not everything is clear from the beginning

mesa/experimental/jupyter_viz.py
Corvince: ah, makes sense. I reverted the change

mesa/experimental/jupyter_viz.py
Corvince: removed

mesa/experimental/jupyter_viz.py
Corvince: yes, removed

Corvince:
> > Removal of FigureMatplotlib dependencies. Presently, we're free from unnecessary re-renders, so this optimization isn't required.
> > Elimination of df from the container. Previously, I believe this was used to trigger re-renders. Now, we track the current step to achieve this. This approach also supports models without a data collector and is essential for upcoming enhancements to the model controller.
> 
> In this PR, only `solara.Markdown(md_text=f"**Step:** {current_step}")` has a dependency to `current_step`, but not the `FigureMatplotlib`'s. How are they notified of the step change then?

Ah, I think you got the dependency list of FigureMatplotlib backwards. In Solara, when the state changes (here: current_step), the component and every child component is re-rendered. So we don't need to notify FigureMatplotlib explicitly to update the plot. Only if we **dont** want it to redraw we can add a dependency list and only if the values of the dependency list change a new figure is drawn. But even without a dependency change it technically still re-renders, but uses the old (cached) image.

rht:
I see, that makes sense. I will merge now.

rht (approved):

rht:
Forgot to mention that the visualization tutorial needs to be updated ASAP before the release (see https://mesa.readthedocs.io/en/stable/tutorials/visualization_tutorial.html). I can't do this within the next few days, and so, @ankitk50, if you have the time to do it, it would be great.

ankitk50:
Hi all, I started working on updating this tutorial. I get the following error while rendering the page:

```
Failed to load model class 'SheetModel' from module 'jupyter-vuetify'
makeError@http://localhost:8889/static/components/requirejs/require.js?v=d37b48bb2137faa0ab98157e240c084dd5b1b5e74911723aa1d1f04c928c2a03dedf922d049e4815f7e5a369faa2e6b6a1000aae958b7953b5cc60411154f593:168:26
```

Are we aware of this issue or I'm missing some config ?

rht:
Try https://discourse.jupyter.org/t/ipysheet-in-tljh-failed-to-load-model-class-sheetmodel-from-module-ipysheet/18714/2.

rht:
@ankitk50 I had already fixed the issue (#1792) because we wanted to release sooner. You should look for other issue instead. #1741 seems to be a low-hanging fruit.

ankitk50:
Thanks @rht, everything looked in order to me. Although I cannot get rid of the Jupyter notebook error. The error does not show on colab, so I use it for now.

rht:
I don't recommend using Colab to develop for long term. You have to fix your local env problem soon.

PR #1789, opened 2023-09-03
Convert make_space into solara components

Corvince:
In this PR I wanted to convert the `make_space` function into a `SpaceView` component that handles all the aspects of drawing the model space. 

However, as it is now, it doesn't work, because in contrast to React solara uses lazy rendering. That is child component only rerender if their props change. But with SpaceView I am only passing space_drawer, model and agent_portrayal. And those don't change with each step. I found two workarounds

1. pass current_step as additional prop. This works but then it needs to be passed again to SpaceViews child components as well
2. remove `@solara.component` from SpaceView and its children. Thus converting them to normal functions that return solara components. 

Both approaches seems rather hacky. Maybe someone else has any ideas?

### DIFF
--- a/mesa/experimental/jupyter_viz.py
+++ b/mesa/experimental/jupyter_viz.py
@@ -48,10 +48,7 @@ def JupyterViz(
 
     with solara.GridFixed(columns=2):
         # 4. Space
-        if space_drawer is None:
-            make_space(model, agent_portrayal)
-        else:
-            space_drawer(model, agent_portrayal)
+        SpaceView(space_drawer, model, agent_portrayal)
         # 5. Plots
         for measure in measures:
             if callable(measure):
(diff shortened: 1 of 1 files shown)


rht:
This is how Solara does it: https://github.com/widgetti/solara/blob/baa36623c3eb7db50672d8eb3d3cdab9220a50a6/solara/components/matplotlib.py#L11, add an optional argument `dependencies`, and the decorator will automatically handle it. Haven't tested whether it has to be passed recursively or not.

Corvince:
> This is how Solara does it: https://github.com/widgetti/solara/blob/baa36623c3eb7db50672d8eb3d3cdab9220a50a6/solara/components/matplotlib.py#L11, add an optional argument `dependencies`, and the decorator will automatically handle it. Haven't tested whether it has to be passed recursively or not.

No, this is the opposite. This is used to *limit* the number of re-renders, where I want to *trigger* a re-render. I mean the idea of passing and additional, changing prop works and this is what my solution 1) refers to (and what i got recommended on discord). But the dependencies argument in Solaras FigureMatplotlib in particular is meant to do the opposite.

Corvince:
Added the additional prop for now - still not completely happy with this solution but at least it works right now

mesa/experimental/jupyter_viz.py
```diff
@@ -182,15 +179,16 @@ def make_user_input(user_input, name, options):
         raise ValueError(f"{input_type} is not a supported input type")
 
 
-def make_space(model, agent_portrayal):
-    def portray(g):
+@solara.component
+def GridView(grid, agent_portrayal, dummy):
```
rht, line 183: `dummy` as a name is rather non-descriptive and confusing.

Corvince:
Reason to close this PR https://github.com/projectmesa/mesa/pull/1805#issuecomment-1737933085

PR #1791, opened 2023-09-05
ModelController: Use use_thread hook

Corvince:
This PR starts to simplify the ModelController logic and unify its controls (no mix between solara and reaction widgets).

I used the use_thread hook, but this doesn't seem to be the best option. 

  1. After the first couple of steps it gets frustrateingly slow. Don't know if that is due to the overhead of starting a new thread for every step or what else is causing this.
  2. The line plot doesn't work. I guess this is due to Matplotlib not being thread-safe and so the data gets changed while still drawing. Might be another issue though. The grid view seems to work fine though

Maybe someone has an idea on how to periodically call the step function.

### DIFF
--- a/mesa/experimental/jupyter_viz.py
+++ b/mesa/experimental/jupyter_viz.py
@@ -1,4 +1,4 @@
-import threading
+import time
 
 import matplotlib.pyplot as plt
 import networkx as nx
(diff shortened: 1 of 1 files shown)


Corvince:
1) will be auto-solved by the next version of solara. There have been some updates to use_thread and using this with the latest git version runs very smoothly and without any strange changes in speed.

2) is still unresolved. Altair works fine though

rht:
The main reason I use `ipywidgets.Play` is that it is the only one that works on Colab, because the play controller is run from the browser in JavaScript.

Have you tested that everything is working fine on Colab?

Corvince:
Reason t close this PR https://github.com/projectmesa/mesa/pull/1805#issuecomment-1737933085

To answer the question: It doesn't work on my local machine  (due to 2) so I haven't tested on colab yet.

PR #1796, opened 2023-09-10
fix: Don't continue playing when a model is reset

### LINKED ISSUE #1782: JupyterViz: Changing simulation parameters shouldn't automatically start the simulation
**Describe the bug**
Currently, changing the parameters auto-play the simulation.

**Expected behavior**
It shouldn't.

**To Reproduce**
Run any `JupyterViz` simulation.

rht:
Fixes #1782.

### DIFF
--- a/mesa/experimental/jupyter_viz.py
+++ b/mesa/experimental/jupyter_viz.py
@@ -89,15 +89,25 @@ def ModelController(
 ):
     playing = solara.use_reactive(False)
     thread = solara.use_reactive(None)
+    # We track the previous step to detect if user resets the model via
+    # clicking the reset button or changing the parameters. If previous_step >
+    # current_step, it means a model reset happens while the simulation is
+    # still playing.
+    previous_step = solara.use_reactive(0)
 
     def on_value_play(change):
-        if model.running:
+        if previous_step.value > current_step and current_step == 0:
+            # We add extra checks for current_step == 0, just to be sure.
+            # We automatically stop the playing if a model is reset.
+            playing.value = False
+        elif model.running:
             do_step()
         else:
             playing.value = False
 
     def do_step():
         model.step()
+        previous_step.value = current_step
         set_current_step(model.schedule.steps)
 
     def do_play():


mesa/experimental/jupyter_viz.py
```diff
@@ -84,15 +84,25 @@ def handle_change_model_params(name: str, value: any):
 def ModelController(model, play_interval, current_step, set_current_step):
     playing = solara.use_reactive(False)
     thread = solara.use_reactive(None)
+    # We track the previous step to detect if user resets the model via
+    # clicking the reset button or changing the parameters. If previous_step >
+    # current_step, it means a model reset happens while the simulation is
+    # still playing.
+    previous_step = solara.use_reactive(0)
```
Corvince, line 91: Have you seen the use previous hook? Might be usable here

https://solara.dev/api/use_previous

rht: I just tried, but with that, the simulation will always step once after the reset. Not sure why. At least this PR works, and the code is not that much different from if it were using `use_previous`.

Corvince: It's still a bit verbose. The step+1 got me curious and I was playing around a bit. 

I think a simpler solution would be to use `use_previous` on the model and then look if the previous model is equal to the current model. If it's not set playing to False and reset current step.

rht: It's problematic to store 2 copies of model with different specs at any given time. Consumes 2x RAM.

rht:
@tpike3 @jackiekazil can we disable the blocking the out-of-date branch? It is not necessary.

jackiekazil:
@rht not sure I understand.... what branch that is out of date is being blocked? and what are you trying to do?

rht:
This screenshot:
> This branch is out-of-date with the base branch

2023-09-13-171553_835x292_scrot

jackiekazil:
Thank you for the clarification. It is updated now. Let me know if anything else needs to be done.

rht:
Can this PR be merged first for the upcoming release that is by the end of this month? It fixes a very problematic autoplay bug. There is no obvious solution regarding with `use_previous`, and it shouldn't block this PR.

Corvince (approved):

Corvince:
Sure, no this was never blocking

PR #1805, opened 2023-09-15
Solara: Let users rewind sessions

Corvince:
Hi there!

I decided to re-implement one of my favorite experimental feature that I never managed to publish to the old visualization server. Its a feature I think very few ABM libraries are implementing, although I find it quite powerful and useful. 

It basically implements that you can go back in time and analyze each step. 

Let me show you a quick screencast of the feature

[screen-capture.webm](https://github.com/projectmesa/mesa/assets/13568919/7fc28528-467d-4a8e-9979-8c7d0f45012a)

As you can see with this change it is possible to go back to previous steps. @rlskoeser since you are very active at the moment, do you think this would be a useful feature for you? This is also a stepping stone on my way to implement another possibly interesting feature and that is running models (with different settings) side-by-side. 

And I know this will come up in the discussion: Yes this implementation seems to be a bit RAM hungry in that each model step is saved. In reality I couldn't find any notable spikes in my ram usage, so we shouldn't make this a problem before it is one.  Remember that we are dealing with visualisable models that tend to be fairly small (more testing would be needed for larger models)

### DIFF
--- a/mesa/experimental/jupyter_viz.py
+++ b/mesa/experimental/jupyter_viz.py
@@ -1,4 +1,4 @@
-import threading
+import copy
 
 import matplotlib.pyplot as plt
 import networkx as nx
@@ -37,32 +37,53 @@ def JupyterViz(
         play_interval: play interval (default: 400)
     """
 
-    current_step, set_current_step = solara.use_state(0)
-
     # 1. Set up model parameters
     user_params, fixed_params = split_model_params(model_params)
     model_parameters, set_model_parameters = solara.use_state(
         {**fixed_params, **{k: v["value"] for k, v in user_params.items()}}
     )
 
+    model, set_model = solara.use_state(
+        model_class(**model_parameters),
+        eq=model_eq,
+    )
+    model_cache, set_model_cache = solara.use_state({0: model})
+
     # 2. Set up Model
     def make_model():
         model = model_class(**model_parameters)
-        set_current_step(0)
-        return model
+        set_model(model)
+        set_model_cache({0: model})
 
-    reset_counter = solara.use_reactive(0)
-    model = solara.use_memo(
-        make_model, dependencies=[*list(model_parameters.values()), reset_counter.value]
+    solara.use_memo(
+        make_model,
+        dependencies=[
+            list(model_parameters.values()),
+        ],
     )
 
     def handle_change_model_params(name: str, value: any):
         set_model_parameters({**model_parameters, name: value})
 
+    def handle_step(step: int):
+        if step in model_cache:
+            previous_model = model_cache[step]
+            set_model(previous_model)
+        else:
+            model.step()
+            set_model_cache({**model_cache, step: copy.deepcopy(model)})
+            set_model(model)
+
     # 3. Set up UI
     solara.Markdown(name)
     UserInputs(user_params, on_change=handle_change_model_params)
-    ModelController(model, play_interval, current_step, set_current_step, reset_counter)
+    ModelControls(
+        model,
+        play_interval,
+        max_step=max(model_cache.keys()),
+        on_step=handle_step,
+        on_reset=make_model,
+    )
 
     with solara.GridFixed(columns=2):
         # 4. Space
@@ -84,82 +105,82 @@ def JupyterViz(
 
 
 @solara.component
-def ModelController(
-    model, play_interval, current_step, set_current_step, reset_counter
-):
+def ModelControls(model, play_interval, max_step, on_step, on_reset):
+    current_step = model.schedule.steps
     playing = solara.use_reactive(False)
-    thread = solara.use_reactive(None)
-    # We track the previous step to detect if user resets the model via
-    # clicking the reset button or changing the parameters. If previous_step >
-    # current_step, it means a model reset happens while the simulation is
-    # still playing.
-    previous_step = solara.use_reactive(0)
 
-    def on_value_play(change):
-        if previous_step.value > current_step and current_step == 0:
-            # We add extra checks for current_step == 0, just to be sure.
-            # We automatically stop the playing if a model is reset.
-            playing.value = False
-        elif model.running:
-            do_step()
-        else:
-            playing.value = False
+    def on_value_play(_):
+        if model.running and playing.value:
+            on_step(current_step + 1)
 
-    def do_step():
-        model.step()
-        previous_step.value = current_step
-        set_current_step(model.schedule.steps)
+    def change_step(value):
+        return lambda: on_step(max(0, value))
 
-    def do_play():
-        model.running = True
-        while model.running:
-            do_step()
+    def reset():
+        playing.value = False
+        on_reset()
 
-    def threaded_do_play():
-        if thread is not None and thread.is_alive():
-            return
-        thread.value = threading.Thread(target=do_play)
-        thread.start()
+    with solara.Card(), solara.Column(
+        gap="40px",
+    ):
+        with solara.Row(gap="2px", style={"align-items": "center"}):
+            with solara.Tooltip("Reset the model"):
+                solara.Button(icon_name="mdi-reload", color="primary", on_click=reset)
+            with solara.Tooltip("Step backward to the beginning"):
+                solara.Button(
+                    icon_name="mdi-skip-backward",
+                    color="primary",
+                    on_click=change_step(0),
+                )
+            with solara.Tooltip("Step backward"):
+                solara.Button(
+                    label="|◀",
+                    color="primary",
+                    on_click=change_step(current_step - 1),
+                )
 
-    def do_pause():
-        if (thread is None) or (not thread.is_alive()):
-            return
-        model.running = False
-        thread.join()
-
-    def do_reset():
-        reset_counter.value += 1
-
-    with solara.Row():
-        solara.Button(label="Step", color="primary", on_click=do_step)
-        # This style is necessary so that the play widget has almost the same
-        # height as typical Solara buttons.
-        solara.Style(
-            """
-        .widget-play {
-            height: 30px;
-        }
-        """
+            # This style is necessary so that the play widget has the same
+            # height as typical Solara buttons.
+            solara.Style(
+                """
+                .widget-play {
+                    height: 36px;
+                }
+                """
+            )
+            with solara.Tooltip("Start/Stop the model"):
+                widgets.Play(
+                    interval=play_interval,
+                    show_repeat=False,
+                    on_value=on_value_play,
+                    playing=playing.value,
+                    on_playing=playing.set,
+                )
+            with solara.Tooltip("Step forward"):
+                solara.Button(
+                    label="▶|",
+                    color="primary",
+                    on_click=change_step(current_step + 1),
+                )
+            with solara.Tooltip("Step forward to the end"):
+                solara.Button(
+                    icon_name="mdi-skip-forward",
+                    color="primary",
+                    on_click=change_step(max_step),
+                )
+        solara.SliderInt(
+            label="Timeline (current step)",
+            value=current_step,
+            min=0,
+            max=max_step,
+            thumb_label="always",
+            step=1,
+            on_value=on_step,
         )
-        widgets.Play(
-            value=0,
-            interval=play_interval,
-            repeat=True,
-            show_repeat=False,
-            on_value=on_value_play,
-            playing=playing.value,
-            on_playing=playing.set,
-        )
-        solara.Button(label="Reset", color="primary", on_click=do_reset)
-        solara.Markdown(md_text=f"**Step:** {current_step}")
-        # threaded_do_play is not used for now because it
-        # doesn't work in Google colab. We use
-        # ipywidgets.Play until it is fixed. The threading
-        # version is definite a much better implementation,
-        # if it works.
-        # solara.Button(label="▶", color="primary", on_click=viz.threaded_do_play)
-        # solara.Button(label="⏸︎", color="primary", on_click=viz.do_pause)
-        # solara.Button(label="Reset", color="primary", on_click=do_reset)
+
+
+def model_eq(a, b):
+    return id(a) == id(b) or (a.schedule.steps == b.schedule.steps)
 
 
 def split_model_params(model_params):
--- a/mesa/experimental/jupyter_viz.py
+++ b/mesa/experimental/jupyter_viz.py
@@ -66,6 +66,8 @@ def JupyterViz(
         set_model_parameters({**model_parameters, name: value})
 
     def handle_step(step: int):
+        if not model.running:
+            return
         if step in model_cache:
             previous_model = model_cache[step]
             set_model(previous_model)
@@ -78,8 +80,8 @@ def JupyterViz(
     solara.Markdown(name)
     UserInputs(user_params, on_change=handle_change_model_params)
     ModelControls(
-        model,
-        play_interval,
+        play_interval=play_interval,
+        current_step=model.schedule.steps,
         max_step=max(model_cache.keys()),
         on_step=handle_step,
         on_reset=make_model,
@@ -105,12 +107,11 @@ def JupyterViz(
 
 
 @solara.component
-def ModelControls(model, play_interval, max_step, on_step, on_reset):
-    current_step = model.schedule.steps
+def ModelControls(play_interval, current_step, max_step, on_step, on_reset):
     playing = solara.use_reactive(False)
 
     def on_value_play(_):
-        if model.running and playing.value:
+        if playing.value:
             on_step(current_step + 1)
 
     def change_step(value):
@@ -180,7 +181,7 @@ def ModelControls(model, play_interval, max_step, on_step, on_reset):
 
 
 def model_eq(a, b):
-    return id(a) == id(b) or (a.schedule.steps == b.schedule.steps)
+    return id(a) == id(b) and (a.schedule.steps == b.schedule.steps)
 
 
 def split_model_params(model_params):
--- a/mesa/experimental/jupyter_viz.py
+++ b/mesa/experimental/jupyter_viz.py
@@ -1,5 +1,4 @@
 import copy
-from typing import Optional
 
 import matplotlib.pyplot as plt
 import networkx as nx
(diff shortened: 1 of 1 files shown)


rht:
RAM issue is a solvable problem. You can store just enough info to recreate the model state at any given time, and this includes capturing only 1 step of data collector output, not the whole timeseries, for any particular step.

See https://github.com/Logende/mesa-replay/blob/main/mesa_replay/cacheable_model.py.

rht:
Though the space is probably the most RAM consuming part, unless it can be stored in a memory efficient way. That said, for huge model space, the space for each step can always be serialized to a file.

rht:
This is a very cool economy simulator: https://news.ycombinator.com/item?id=37527773, that has buttons to skip step by 4 or 200. And it is a space-less model that still has visualization. This would be a fun student project to recreate in Mesa.

tpike3:
This is super cool @Corvince!  I think the RAM challenge is just something to continually improve and mitigate as @rht commented. But this feature and its value added, makes the RAM concerns worth the cost.

Corvince:
Thanks for the kind words @tpike3 

And maybe my wording was a bit unclear, but I haven't even seen ram usage as an issue so far. So I totally agree this is something we can mitigate **if** it comes up as an issue. It is just something I was worried about while developing but so far it seems fine.

rht:
For this to be mergeable, it needs to be optional, as a separate component. @tpike3 commented that having a minimalist default UI helps a lot so as not to overwhelm students.

The current jupyter_viz.py is still small, but this seems to be a suitable time to split it into several files. I'd say the space drawers and plot drawers should be moved to a separate file instead of the UI controllers.

Corvince:
I agree with most of your comment, but lets dissect it a bit to see where we agree and possibly disagree. I think this PR actually can serve as a good base for the things you want.

> For this to be mergeable, it needs to be optional, as a separate component. @tpike3 commented that having a minimalist default UI helps a lot so as not to overwhelm students.

This gave me some ideas how to further simplify the code and now these controls are available as `TimelineControls` and a simplified version with only reset, step and play as `BaseControls`. This is toggleable via a new JupyterViz parameter with a boolean timeline. The further simplification makes it theoretically possible to use these controls without any model - they are fully control-agnostic for maximal reusability. 

> 
> The current jupyter_viz.py is still small, but this seems to be a suitable time to split it into several files. I'd say the space drawers and plot drawers should be moved to a separate file instead of the UI controllers.

I think the file is already quite large so I totally agree on splitting - but I wouldn't want to do it as part of this PR, this should be a separate PR. From React projects I am used to have one component per file. 

Also I would really like to have this PR merged soonish, because it bakes in some nice code simplifications. Foremost `model` is now a state variable that is newly set for every step. This means I could remove `reset_counter` and `current_step` to trigger any state changes. This also means that means problems like in #1789 are removed - we don't need something else to trigger a rerender. 

Also my vision for our frontend is to have each component act as independently of each other as possible. Thus `JupyterViz` would act as the default glue between them and most of the dependent logic should happen there. Especially something like a `SpaceView` should be callable on its own to display single components in a notebook environment.

mesa/experimental/jupyter_viz.py
```diff
@@ -1,9 +1,11 @@
-import threading
+import copy
+from typing import Optional
 
 import matplotlib.pyplot as plt
 import networkx as nx
 import reacton.ipywidgets as widgets
 import solara
+from ipywidgets import Layout
```
rht, line 8: In L6, the `widgets` (`reacton.ipywidgets`) already has `Layout`. This class is probably not reactive, but we should always use the API given by Solara dev, just to be safe.

mesa/experimental/jupyter_viz.py
```diff
@@ -22,6 +24,7 @@ def JupyterViz(
     agent_portrayal=None,
     space_drawer="default",
     play_interval=400,
+    timeline=False,
```
rht, line 27: Needs to update the docstring.

mesa/experimental/jupyter_viz.py
```diff
@@ -37,32 +40,58 @@ def JupyterViz(
...
-    ModelController(model, play_interval, current_step, set_current_step, reset_counter)
+    TimelineControls(
+        play_interval=play_interval,
+        on_step=handle_step,
+        on_reset=make_model,
+        current_step=model.schedule.steps,
+        max_step=max(model_cache.keys()),
+    ) if timeline else BaseControls(
+        play_interval=play_interval,
+        on_step=handle_step,
```
rht, line 92: `handle_step` does caching regardless of whether the control is `TimelineControls` or `BaseControls`. It's not a problem for models that can be visualized, but it's still consuming more RAM than necessary.

rht: Clarification: *It's not a problem for the **scale** of models that can be visualized, which is usually small.

mesa/experimental/jupyter_viz.py
```diff
@@ -83,83 +112,98 @@ def handle_change_model_params(name: str, value: any):
...
+    def reset():
+        playing.value = False
+        on_reset()
+
+    with solara.Card(), solara.Row(gap="2px", style={"align-items": "center"}):
+        with solara.Tooltip("Reset the model"):
+            solara.Button(icon_name="mdi-reload", color="primary", on_click=reset)
+
+        with solara.Tooltip("Step forward"):
+            solara.Button(label="▶|", color="primary", on_click=on_step)
```
rht, line 131: This symbol is not in https://en.wikipedia.org/wiki/Media_control_symbols. I think `+1` or `-1` (for step backward) is clearer?

mesa/experimental/jupyter_viz.py
rht: > but it's still consuming more RAM than necessary.

There is the use case when the viz server needs to be hosted in an AWS EC2 t2.nano (0.5 GB) or t2.micro (1 GB).

Corvince: No, handle_step only does caching if you specify a specific step. If it is called without a step (like the BaseControls do) the model is simply advanced. Added a docstring to handle_step to clarify this. There is still the model cache and to keep the code simple it always holds the initial step. But I think we can live with 1 additional copy. That might also be handy if someone wants to add the option to go to the first step later, to re-run the model. Currently it is a bit ambiguous if the reset button should just go back to the first step or regenerate a new model (It does the latter but there is no reason why it shouldn't just rewind).

rht: But in the hypothetical situation when I am not familiar with the frontend code, and want to read a small subset of this component, I have to read the docstring about caching and so on. Whereas if you have a simpler `handle_step` for the simpler controller, I can grasp the structure better before moving on to a more complicated component.

rht: As in, have 2 `handle_step`-like functions for each controller classes.

Corvince: > But in the hypothetical situation when I am not familiar with the frontend code, and want to read a small subset of this component, I have to read the docstring about caching and so on. 

That doesn't make sense. If you want to understand the code, why wouldn't you read the docstring? 
 And the whole function is only 10 lines of code, 5 of which would need to be duplicated. I think this would be more confusing and more difficult to maintain. And imagine someone writing a new Controls component. Now they have to decide which handle step method they want to use or add yet another one. If the function were more complex I would agree but I think it is not worth the complications.

Is this a blocker?

rht: > Now they have to decide which handle step method they want to use or add yet another one.

This is far better than fattening up `handle_step` with more if-else's if the existing function does not cover the use case.

Why do I need to go through the caching code and have to consciously read and then to ignore it, just to read about stepping for basic controls? Such a waste of effort. I strongly disagree.

Corvince: I think we need a third opinion. My final statement is that my experience as a software developer tells me that in 95% of cases one should aim for maintainability. And two functions are harder to maintain than one.

Corvince:
There is a discussion about whether to split handle_step. I revisited the code and needed to do a small adjustment for an upcoming feature. With that it is easy enough to split handle_step into two functions without code duplication. I still think it would not be necessary, but this way I am fine to follow @rht suggestion about splitting the function. 

So I will do that, but there is a bug I need to fix first. So, dont merge yet, update coming

Corvince:
Got it. @rht let me know if there is anything else you spot with this PR. 👍

Corvince:
Would be great if we could merge this soonish @tpike3 @rht . It blocks me a little bit right now

mesa/experimental/jupyter_viz.py
```diff
@@ -35,34 +36,66 @@ def JupyterViz(
...
     )
 
     def handle_change_model_params(name: str, value: any):
         set_model_parameters({**model_parameters, name: value})
 
+    def do_step():
+        if not model.running:
+            return
+
+        updated_model = copy.deepcopy(model)
```
rht, line 69: You are probably going to say it's fine for small models. But this recreates the grid, agent vars, and model vars from scratch every time. It will hit the resource ceiling sooner than before. Python objects are not meant to be performant for immutable paradigm, are they?

rht: Compare with e.g. Agents.jl, which is frugal with resource.

Corvince: > You are probably going to say it's fine for small models. But this recreates the grid, agent vars, and model vars from scratch every time. It will hit the resource ceiling sooner than before. Python objects are not meant to be performant for immutable paradigm, are they?

Yes this is exactly what I am going to say 😅 

Plus, also for large models that should be okay for a long time. For performance there is no right or wrong, but in terms of speed copying is much faster than actually recreating and for memory usage when we deactivate the timeline we only ever hold 2 models at the same time. 

I would say we should try it this way, because it makes the code much simpler and easier to reason about. If we eventually do hit a performance ceiling we can change things. But right now we should optimize for the 80% use case. 

And our solara frontend is still experimental. We should focus on moving forward instead of worrying too much about hypothetical scenarios.

rht: Can you describe what features you want to implement with this immutable model convenience? IIRC there was a GH issue where the OP requested for a feature to allow them to change model params (only if it makes sense to do so) without triggering a whole system reset. Is this one of them?

mesa/experimental/jupyter_viz.py
```diff
@@ -83,83 +116,98 @@ def handle_change_model_params(name: str, value: any):
...
-        else:
-            playing.value = False
 
-    def do_step():
-        model.step()
-        previous_step.value = current_step
-        set_current_step(model.schedule.steps)
+    def on_value_play(_):
+        if playing.value:
+            on_step()
```
rht, line 124: So basically, when the simulation runs until `model.running` is `False`, the buttons will look like they are still playing indefinitely, but that there are just no more viz state update?

mesa/experimental/jupyter_viz.py
```diff
@@ -83,83 +116,98 @@ def handle_change_model_params(name: str, value: any):
                 make_plot(model, measure)
 
 
-@solara.component
-def ModelController(
-    model, play_interval, current_step, set_current_step, reset_counter
-):
+def BaseControls(play_interval, on_step, on_reset):
```
rht, line 119: Why is there no `@solara.component` here?

mesa/experimental/jupyter_viz.py
rht: Note that some Solara servers are meant to be accessed by lots of users in parallel, and some are run on a small AWS instance. They could be considered to be within 80% of the use case. I have to weigh over this use case vs the features you want to implement.

mesa/experimental/jupyter_viz.py
Corvince: Good catch, hadn't thought about this case. But I would consider the current "play" button a temporary solution, because it also doesn't fit the style of the other buttons and it shows an unnecessary additional stop button that does the same as the pause button. Therefore I would resolve that issue once we implement a better "play" solution

mesa/experimental/jupyter_viz.py
Corvince: Forgot it

mesa/experimental/jupyter_viz.py
Corvince: > Can you describe what features you want to implement with this immutable model convenience? IIRC there was a GH issue where the OP requested for a feature to allow them to change model params (only if it makes sense to do so) without triggering a whole system reset. Is this one of them?

 - First of all right now instead of a single "model" variable, we have "model", "current_step" and "reset_counter". Both of the latter are used for triggering change detection, because changes to the model object will go unnoticed as far as solara is concerned. If we implement further model-changing functionalities (ex. interactivity) we always have to worry about keeping model up to date (by adding more and more dependency variables to the make_model function).

- This very PR that adds time traveling to mesa.

- PR #1791 will work without any dummy parameter to function calling trickery.

- Running multiple models in parallel will get messy when we have to store multiple models, model parameters, current_steps and reset_counters. 

- Saving visualization and model state to continue model runs accross sessions. 

These are the features I have currently planned or panned out (mostly following https://github.com/Corvince/mesa-viz). Please don't start any lenghty discussions about any of them.

rht: > First of all right now instead of a single "model" variable, we have "model", "current_step" and "reset_counter".

They all can be generalized to at most 2 reactive variables: to inform a model reset (`reset_counter`), and to inform a model update.

```python
model_update = solara.reactive(0)
```
This reactive variable is updated whenever you want to trigger components rerender (e.g. after a `model.step()`).

While `reset_counter` can be removed once `ipywidgets.Play` is replaced.

Corvince: Sorry I have to ask again for clarification. I am really not doing this annoy you. 

What you write is absolutely correct. I just don't know its purpose. Is there a hidden question or some form of conclusion I am not getting or is it a simple statement?

rht: I forgot to quote the remaining paragraph

> Both of the latter are used for triggering change detection, because changes to the model object will go unnoticed as far as solara is concerned. If we implement further model-changing functionalities (ex. interactivity) we always have to worry about keeping model up to date (by adding more and more dependency variables to the make_model function).

What I am saying is that the `model_update` variable can be used as a proxy for any interactive features to be based on, if they want to get a redraw signal from model.

Corvince: I understand that, but I am asking what you want to say. Are you seriously opposed to merging this PR, because it copies the model for state updates? This approach incidentally makes the code more straightforward, comprehensible, maintainable and easier to reason about.

rht: The current Solara code is already far simpler than mesa-viz-tornado, for what it can already do. Either:
- use `model_update` as a compromise
- we consider your current immutable model implementation as a training wheel so that it is easier for you to experiment with features, but eventually, this must be optimized in the future

Which one do you prefer? It's rather paradoxical that Agents.jl is more performant yet they are more resource conscious, while you are being more extravagant.

Corvince:
That is it. I’m done. @rht you won. The continuous hoops that I feel I’m being asked to jump through are becoming exhaustive.

I opened this PR with the intention of introducing a beneficial feature. Instead of being happy for the submission and helping to improve the PR, @rht makes me jump through hoops by constantly imagining, in my view, unlikely scenarios like people being overwhelmed by three additional buttons, people who want to understand the code but can’t be bothered to read a docstring and ten lines of code and finally people who want to serve their mesa models to lots users, but can only afford the smallest AWS instances. I want to clarify that I understand the importance of considering various user needs and scenarios, but the focus on such extensive hypothetical situations has been draining, leading to prolonged discussions that feel unproductive and detract from the substantive aspects of the PR. The bulk of the 40 comments on this PR revolve around these speculative situations rather than the concrete code changes proposed.

But to be clear this is not about any code changes. "It’s primarily about the tone of the discussion. I’ve yet to see comments like “nice feature, here are some suggestions,” or “thank you for implementing my comments.” Instead, the interactions often seem marked by a lack of trust and feel provocative, and the recent offer of “training wheels” felt quite personal and dismissive. I’ve been contributing to Mesa for six years and have several years of experience as a professional web developer; I believe I’m past needing training wheels

And, of course, this doesn’t only affect this PR. My last (non-trivial) merged PRs also needed lengthy discussions about absurd details. I am creating PRs because I have fun coding for mesa and want to add some nice features that I truly believe are useful and for example in this case are somewhat unique in the ABM landscape. Receiving positive feedback from @tpike, @jackiekazil and others has been reassuring, but with @rht comments it feels like I am a supplicant dependent on the grace of the master and not like a contributor on the same level. 
  
This has left me feeling undervalued and, frankly, exhausted. It seems my contributions are more subjected to scrutiny over seemingly inconsequential details than valued for their merit and potential impact. This atmosphere is becoming toxic for me, leading me to decide to contribute in other forms. I will pursue some of my ideas in another repo and we can see if at some point I will contribute back to main mesa or just showcase features in mesa-examples. I will keep watching issues and try to help users and review PRs, but I don’t think I will open any PRs soon.

rht:
That sounds like how you would describe yourself. In the past years, you had initiated dismissive remarks to me (e.g. [most recent one](https://github.com/projectmesa/mesa/pull/1814#discussion_r1332910391), which was definitely intentional). But when I somehow unintentionally irked you, you ragequitted.

My PR's were being scrutinized and I have been fine with it. This is a larger PR with non-atomic changes, you should be aware that of course larger surface has more comments.

> unlikely scenarios like people being overwhelmed by three additional buttons

My words were being twisted. This was from @tpike3's feedback when teaching a class.

jackiekazil:
From my brief passing this evening, I see how frustrating it is to be in either position in this discussion. There is an opportunity for us to consider possible procedural changes in governance to ensure everyone's expectations are met. I will spend a day or two thinking about this and then respond further. It is essential for us as a project. I personally value both of your contributions to Mesa.

jackiekazil:
We value these contributions. @tpike3 and I wanted to help the review on these to work through any issues -- @Corvince, please let us know if you are open to it. I will leave PRs closed unless you open and move forward OR unless someone else picks up your contributions to push forward. 

We are also proposing a Code of conduct discussion which is happening here -- https://github.com/projectmesa/mesa/discussions/1822 And will continue during our dev meeting Oct 14 in homes to provide guardrails to avoid situations like what occurred.

rht:
Note that I review this PR with the objective of getting this merged soon.

@Corvince if I understand your criticisms correctly:

> Instead of being happy for the submission and helping to improve the PR, @rht makes me jump through hoops by constantly imagining, in my view, unlikely scenarios ...

You understand that this is core Mesa repo, and so being careful with what is merged is justified. I intentionally wrote the current `JupyterViz` to be very similar in functionality as `mesa-viz-tornado`. As much as you ridiculed my concerns, I didn't find the scenarios I raised unlikely, and so, I disagreed. Am I not supposed to express disagreement? I have already said the rationale behind the 3 additional buttons earlier. Regarding with code clarity, in addition to the early version of the components not being orthogonal, I was confused by your code before you added the docstring, and this was coming from me who wrote the initial code, not some random users. Regarding with the AWS comment, this is rather inconvenient for me to disclose, but I am one of those people who can only afford the smallest AWS instances, and so, here is a concrete data point.

> But to be clear this is not about any code changes. "It’s primarily about the tone of the discussion. I’ve yet to see comments like “nice feature, here are some suggestions,” or “thank you for implementing my comments.” Instead, the interactions often seem marked by a lack of trust and feel provocative, and the recent offer of “training wheels” felt quite personal and dismissive. I’ve been contributing to Mesa for six years and have several years of experience as a professional web developer; I believe I’m past needing training wheels

I thought @tpike's comment was already sufficient? And given that you have been here for a long time, I wouldn't have to treat you as a first timer, because you know the worth of this PR. If you need an explicit opinion for contextualization, then here you go: I find the rewind feature very useful for exploration that was harder to do previously, where you would have to restart the simulation over and over. There would be a system which would undergo a phase transition several times, and so, just like a detective who would need to pause and rewind to do their investigation on a video recording, and so does a complexity researcher.

You extracted the "training wheels" remark out of context. I also said:
"It's rather paradoxical that Agents.jl is more performant yet they are more resource conscious, while you are being more extravagant."

This is in the context of your position that Mesa doesn't need performance optimization that badly: https://github.com/projectmesa/mesa/discussions/1610#discussioncomment-6126311. And you are entitled to it.

> And, of course, this doesn’t only affect this PR. My last (non-trivial) merged PRs also needed lengthy discussions about absurd details.

I couldn't find such evidence for "absurd details" for the last merged PRs: #1786 and #1788. On the other hand, my PR's have had lengthy discussions from you, and I was fine with it. Appreciate the eyeballs though.

Lastly, you should be aware that I have been upset with your dismissive remarks. If you, after an internal self-review, acknowledge that they were inappropriate, then we can move on.

Corvince:
@rht I value the effort you've made to address my criticisms. However, I believe that continuing to discuss specific points further may not be productive.

Corvince:
> Regarding with the AWS comment, this is rather inconvenient for me to disclose, but I am one of those people who can only afford the smallest AWS instances, and so, here is a concrete data point.
> 

Again, not wanting to discuss any details here, but this should be completely orthogonal and I find this quite interesting. Can you provide any details for how to run mesa on AWS? If your workflow is in any way reproducible it would be great if you could write up how you host your models on AWS, I think this would be of interest for some people. Also (for the future of mesa frontend), how does your user base look like, for example how many users to you typically serve/ max user count?

PR #1807, opened 2023-09-18
HISTORY.rst: Correct neighbor_iter() replacement in 2.0.0

EwoutH:
The 2.0.0 release notes said incorrectly to replace `neighbor_iter()` with `iter_neighborhood()` (which returns an iterator of Tuple coordinates) instead of the correct `iter_neighbors()` (which returns an iterator of Agents). This PR fixes this.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -47,7 +47,7 @@ Mesa 2.0 includes:
     * `find_empty()`: convert this to `move_to_empty()`
     * `num_agents`: removed parameter from `move_to_empty()`
     * `position_agent()`: convert this to `place_agent`
-    * `neighbor_iter()`: convert this to `iter_neighborhood()`
+    * `neighbor_iter()`: convert this to `iter_neighbors()`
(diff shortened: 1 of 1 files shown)


rht (approved):

rht:
This PR can't be merged because the current settings requires the tests to pass, but since this PR only modifies a non-Python file, the tests are not activated.

Corvince:
> This PR can't be merged because the current settings requires the tests to pass, but since this PR only modifies a non-Python file, the tests are not activated.

Interesting, not sure how we should handle this. For now I am merging by bypassing the branch protection rules.

Corvince:
Probably it doesn't happen that often. It feels a bit unease to bypass the protection, but that only happens if we have no code changes which shouldn't break things so easily. So I guess we can live with having to wait for a maintainer to merge

Corvince:
Oh and forgot: Thanks @EwoutH for catching an fixing this! 🚀

PR #1809, opened 2023-09-18
remove attrgetter performace optimization

Corvince:
Remove performace optimization to ease multi agent data collection.

This makes it possible to collect agent attributes with simple string values across multiple agent types.

Previously
```python
agent_reporters={'a': 'val_a', 'b': 'val_b'}
```

would error if val_a or val_b is missing in some agents. Now it just returns None for that value.

And the performance drop should be barely noticable.

### DIFF
--- a/mesa/datacollection.py
+++ b/mesa/datacollection.py
@@ -36,7 +36,6 @@ The default DataCollector here makes several assumptions:
 """
 import itertools
 import types
-from operator import attrgetter
 
 import pandas as pd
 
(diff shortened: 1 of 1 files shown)


rht:
Ruff error.

Corvince:
Oopsie, somehow my ruff extension for vscode only works roughly have the time and I almost always forget to check again. Should be fixed now.

rht:
Still Ruff error.

rht (approved):

PR #1757, opened 2023-08-07
Docker: Update to use Solara viz

rht:
Not yet tested. Note that the current Docker setup is broken because the examples folder is now gone (moved to mesa-examples repo).

PR #1811, opened 2023-09-19
Update version to 2.1.2

rht:
I included #1757, so it must be merged first.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,56 @@
 Release History
 ---------------
 
+2.1.2 (2023-09-20)
+++++++++++++++++++
+
+This release contains fixes, and several improvements and new features to
+the JupyterViz/Solara frontend. It's a patch release instead of a minor release
+because the JupyterViz frontend is still considered experimental.
+
+**Improvements**
+
+* perf: Access grid only once #1751
+* docs: compile notebooks at build time #1753
+* docs: Remove nbsphinx and explicit .ipynb suffix #1754
+* rtd: Use gruvbox-dark as style #1719
+* build(deps): bump actions/checkout from 3 to 4 #1790
+
+**Solara/JupyterViz**
+
+* solara: Implement visualization for network grid #1767
+* Add support for select input type #1779
+* Add step count display to JupyterViz #1775
+* Simplify solara code #1786
+* Add docstring for jupyterviz make_user_input that documents supported inputs #1784
+* Revise, test, & document JupyterViz options for drawing agent space #1783
+* Add UserInputs component #1788
+* Fix: Remove dict merge operator, python 3.8 compat #1793
+* feat: Add reset button to JupyterViz #1795
+* Add support for solara.Checkbox user input #1798
+* viz tutorial: Update custom plot to reflect new code #1799
+* fix: Don't continue playing when a model is reset #1796
+* Docker: Update to use Solara viz #1757
+
+**Refactors**
+
+* Move viz stuff to mesa-viz-tornado Git repo #1746
+* simplify get neighborhood #1760
+* remove attrgetter performance optimization #1809
+
+**Fixes**
+
+* fix: Add Matplotlib as dependency #1747
+* fix install for visualization tutorial in colab #1752
+* fix: Allow multiple connections in Solara #1759
+* Revert "Ensure sphinx>=7" #1762
+* fix README pic to remove line on left side #1763
+* space: Ensure get_neighborhood output & cache are immutable #1780
+* fix: Use .pytemplate for name for cookiecutter #1785
+* HISTORY.rst: Correct neighbor_iter() replacement in 2.0.0 #1807
+* docs: Always link to stable version #1810
+* Remove exclude_none_values #1813
+
 2.1.1 (2023-08-02)
 +++++++++++++++++++
 
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,7 +3,7 @@
 Release History
 ---------------
 
-2.1.2 (2023-09-20)
+2.1.2 (2023-09-23)
 ++++++++++++++++++
 
 This release contains fixes, and several improvements and new features to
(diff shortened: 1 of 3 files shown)


tpike3:
Can you fix the codespell error on line 41

Corvince (dismissed):
This won't be publishable to PyPi again because of the GitHub dependency of mesa-viz-tornado. We have to package this up first

Corvince:
But thanks a lot for putting this together @rht

rht:
Good catch. I was too hasty, because it has been days without progress.

HISTORY.rst
```diff
@@ -3,6 +3,55 @@
...
+* Add support for solara.Checkbox user input #1798
+* viz tutorial: Update custom plot to reflect new code #1799
+* fix: Don't continue playing when a model is reset #1796
+* Docker: Update to use Solara viz #1757
+
+**Refactors**
+
+* Move viz stuff to mesa-viz-tornado Git repo #1746
+* simplify get neighborhood #1760
+* remove attrgetter performace optimization #1809
```
rht, line 41: Note: I copy pasted this from the PR.

rht:
> This won't be publishable to PyPi again because of the GitHub dependency of mesa-viz-tornado. We have to package this up first

I don't have permission to do so. Please don't delay this any longer.

Corvince:
What kind of permissions do you mean? There isn't any package on PyPi named `mesa-viz-tornado` yet, so theoretically anyone can submit such package. I think you also already have some knowledge about packaging a python package, no? So maybe you would be willing to publish? You can also first test on test.pypi.org if everything works. 

Otherwise I myself won't have much time until next week. Don't know how it is with @tpike3 or @jackiekazil

Corvince:
I mean we have to somewhat wait for @tpike3 or @jackiekazil anyhow, because they have to finally publish the mesa package. So they might be willing to do the same with mesa-viz-tornado.

rht:
> There isn't any package on PyPi named mesa-viz-tornado yet, so theoretically anyone can submit such package.

You already know yourself that if it is not published by Mesa's PyPI account, it is bound to draw suspicion.

tpike3:
Just to recap.. 
- we want to release Mesa 2.1.2 today (9/20)
- we need to package up mesa-viz-tornado #1746 but it needs to be packaged and released under Mesa so not to be considered suspicious, but the the plan is to do that later (not today)

Corvince:
> > There isn't any package on PyPi named mesa-viz-tornado yet, so theoretically anyone can submit such package.
> 
> You already know yourself that if it is not published by Mesa's PyPI account, it is bound to draw suspicion.

There is no Mesa PyPi account. Previous versions of mesa have been published by @jackiekazil account. @tpike3 would have the same right. Mesa-geo is either published by myself or @wang-boyu. I wouldn't know **who** would be suspicious if you published that package @rht . After all you are probably the most active contributor at the moment and member of the core team.

jackiekazil:
I can publish right after work today.

jackiekazil:
Catching up on where we are ... please let me know if I need to correct anything. 

* This is blocked by needing mesa-viz-tornado release. 
  * I created a history ticket - does that need to be done as well -- https://github.com/projectmesa/mesa-viz-tornado/issues/29
   * I can do a release of this for speed's sake - we should talk about governance around releases. We discussed this in the past, but we have yet to conclude. (cc @tpike3 for next agenda)
   * readme - we can skip a history for now, but we need a readme; otherwise, people will wonder what is happening. Is there anything else that needs to happen? 
 
Are we imagining the process for now being release mesa-viz-tornado, then Mesa? 
 
 * Release of mesa -- 
   * We should update this requirement before the release -- https://github.com/projectmesa/mesa/blob/main/setup.py#L11

rht:
I think it's fine to not have the history file for mesa-viz-tornado, since technically it is no longer going to be developed anyway.

rht:
I have updated the changelog to include #1813.

tpike3:
@rht are you going to remove the requires from github--- https://github.com/projectmesa/mesa/blob/main/setup.py#L11 or do you want me to so we can release to PyPI?

rht:
You can do it, to save a roundtrip.

Corvince:
> @rht are you going to remove the requires from github--- https://github.com/projectmesa/mesa/blob/main/setup.py#L11 or do you want me to so we can release to PyPI?

Wait, but we can't just do that - that would would be a breaking change. Everyone's visualization would break.

rht:
@tpike3 is going to replace it with a PyPI release version instead of Git version: https://github.com/projectmesa/mesa/pull/1811#issuecomment-1727376679

Corvince:
Ok I wasn't sure from the wording and there isn't a package yet.

jackiekazil:
Quick read me for mesa-viz-tornado? 
I would do it, but know less about it than you all. It dosn't have to be fancy, I would cover the following -- 

TItle 
1 Paragraph description
Possible things to include - what is it? Why it exists? who would use it? what is the future state?

jackiekazil:
Suggested pattern for release -- 

- [ ] Update readme in mesa-viz-tornado, release to pypi
- [ ] Adjust setup.py in this repo to point to pypi, not the repo 
- [ ] Update history in this repo
- [ ] Release this repo

setup.py
```diff
@@ -8,7 +8,7 @@
     "click",
     "cookiecutter",
     "matplotlib",
-    "mesa_viz_tornado @ git+https://github.com/rht/mesa-viz-tornado@a4d9242e90ef7f1fcd388fb5c6d72a4177a76bdd",
+    "mesa_viz_tornado,
```
rht, line 11: Syntax error.

tpike3: Sorry about that, that's what I get for being lazy and seeing what it was like to edit the file in the browser
 now fixed

jackiekazil (approved):
LGTM

jackiekazil:
Merging for a release.

jackiekazil:
Released - https://pypi.org/project/Mesa/2.1.2/

PR #1813, opened 2023-09-20
Remove exclude_none_values

rht:
This feature is functional only if the agent records are in the form of
```
{'satisfication': (('A0', 1), ('B0', None), ('A1', 1), ('B1', None), ('A2', 1), ('B2', None)), 'unique_id': (('A0', 'A0'), ('B0', 'B0'), ('A1', 'A1'), ('B1', 'B1'), ('A2', 'A2'), ('B2', 'B2'))}
```

instead of
```
((1, 'A0', 1, 'A0'), (1, 'B0', None, 'B0'), (1, 'A1', 1, 'A1'), (1, 'B1', None, 'B1'), (1, 'A2', 1, 'A2'), (1, 'B2', None, 'B2'))
```

A more explicit solution instead of implicitly ignoring a group of agents just because of their attribute returns `None`, would be to add a filter to the data collector itself. i.e., instead of
```python
agent_records = map(get_reports, model.schedule.agents)
```

we should do
```python
agent_records = map(get_reports, filter(agent_filter, model.schedule.agents))
```
Where `agent_filter` can be `lambda a: isinstance(a, Trader)` . And this, too, is only a few lines of code of change.

### DIFF
--- a/mesa/datacollection.py
+++ b/mesa/datacollection.py
@@ -55,7 +55,6 @@ class DataCollector:
         model_reporters=None,
         agent_reporters=None,
         tables=None,
-        exclude_none_values=False,
     ):
         """Instantiate a DataCollector with lists of model and agent reporters.
         Both model_reporters and agent_reporters accept a dictionary mapping a
@@ -79,8 +78,6 @@ class DataCollector:
             model_reporters: Dictionary of reporter names and attributes/funcs
             agent_reporters: Dictionary of reporter names and attributes/funcs.
             tables: Dictionary of table names to lists of column names.
-            exclude_none_values: Boolean of whether to drop records which values
-            are None, in the final result.
 
         Notes:
             If you want to pickle your model you must not use lambda functions.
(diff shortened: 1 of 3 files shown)


rht:
With the latter, we may have multiple data collectors, but only 1 scheduler is needed.

Corvince:
On a first impression it is a really good idea to use a filter function. 

With respect to this PR. I am in favor of removing the function, but technically this is a breaking change so we have to go to mesa 3.0. 

Being very handwavy we could say this is a bug fix, since the functions is broken atm. And since I doubt many people (if any) use this function we should get away with it. But we have to be aware that we are breaking semver. I am fine with that in this case, but let's see what are other people's opinions about it.

rht:
This functionality hasn't been well advertised and without a tutorial/how-to guide, so in this case I think it is fine to consider this a bugfix.

tpike3:
I am good with a bugfix as well; we want to get this in 2.1.2 or would this be 2.2?

Corvince (approved):

Issue #1812, opened 2023-09-20
JupyterViz: display simulation seed on screen so that the result can be recreated if needed


rht:
**What's the problem this feature will solve?**
While the qualitative behavior of a simulation shouldn't depend on the seed, sometimes you'd observe a rare behavior that only a subset of the seed enables.

PR #1814, opened 2023-09-20
model: Ensure the seed is initialized with current timestamp when it is None

rht:
This is a precursor to #1812, and is useful info for debugging purpose, because apparently there is no way to get a `random.Random()` object's seed once it is initialized. At best, you can have its state instead.

### DIFF
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -21,7 +21,11 @@ class Model:
     def __new__(cls, *args: Any, **kwargs: Any) -> Any:
         """Create a new model object and instantiate its RNG automatically."""
         obj = object.__new__(cls)
-        obj._seed = kwargs.get("seed", None)
+        obj._seed = kwargs.get("seed")
+        if obj._seed is None:
+            # We explicitly specify the seed here so that we know its value in
+            # advance.
+            obj._seed = random.random()  # noqa: S311
         obj.random = random.Random(obj._seed)
         return obj
 


mesa/model.py
```diff
@@ -21,7 +22,16 @@ class Model:
     def __new__(cls, *args: Any, **kwargs: Any) -> Any:
         """Create a new model object and instantiate its RNG automatically."""
         obj = object.__new__(cls)
-        obj._seed = kwargs.get("seed", None)
+        obj._seed = kwargs.get("seed")
+        if obj._seed is None:
+            # See https://docs.python.org/3/library/random.html#random.seed
```
Corvince, line 27: The comment is not 100% accurate, but I think it can be removed completely anyway. I don't think the rational is important here (it should go into the commit message if you want to preserve it).

mesa/model.py
```diff
@@ -21,7 +22,16 @@ class Model:
...
+            # which says "If a is omitted or None, the current system time is used."
+            # We explicitly specify the seed here so that we know its value.
+            # datetime.now().timestamp() is more cross-platform than using
+            # time.time(), and guarantees that the precision is to the
+            # microsecond.
+            current_timestamp = datetime.datetime.now(datetime.timezone.utc).timestamp()
```
Corvince, line 33: I don't understand why we need that additional precision here - time.time seems to be simpler and should in any case be sufficient for just having a known seed.

Corvince (approved):

Corvince:
I approved, but added 2 comments. That should mean you can make suggested or leave it as is if you strongly disagree. Its not important. 

I haven't merged yet, because I don't know if this should go into the 2.1.2. release (which would need to be updated - might be easier to merge later)

mesa/model.py
rht: See https://stackoverflow.com/questions/27276135/python-random-system-time-seed#comment134270695_27276285.

mesa/model.py
rht: I can make it more concise. But I think a new rule of thumb is that if ChatGPT can't deduce the purpose of the code that matches this comment, then the comment is needed.

rht: Also, having to view the blame for a small helpful comment is inconvenient.

mesa/model.py
Corvince: Well at the beginning of https://docs.python.org/3/library/time.html#module-time it says that time() will always return the most accurate time available. There is simply no logical way for datetime to achieve a higher precision. This might have been different in the past, the answer is from 2014

rht: The answer is from 2014, but that comment is from 2023.

From https://docs.python.org/3/library/time.html#time.time

> Note that even though the time is always returned as a floating point number, not all systems provide time with a better precision than 1 second. While this function normally returns non-decreasing values, it can return a lower value than a previous call if the system clock has been set back between the two calls.

Corvince: Yes, but if the operating system doesn't provide a better accuracy, how should datetime have a higher precision?

mesa/model.py
```diff
@@ -21,7 +22,15 @@ class Model:
     def __new__(cls, *args: Any, **kwargs: Any) -> Any:
         """Create a new model object and instantiate its RNG automatically."""
         obj = object.__new__(cls)
-        obj._seed = kwargs.get("seed", None)
+        obj._seed = kwargs.get("seed")
+        if obj._seed is None:
+            # We explicitly specify the seed here so that we know its value in
+            # advance. This value matches random.Random's default seed, which
+            # is the current timestamp. Note: datetime.now().timestamp() is
```
Corvince, line 29: It usually isn't the default seed.

> If a is omitted or None, the current system time is used. If randomness sources are provided by the operating system, they are used instead of the system time (see the [os.urandom()](https://docs.python.org/3/library/os.html#os.urandom) function for details on availability).

Its hard to find a OS without a randomness source.

mesa/model.py
rht: The datetime, by specification, with the timestamp, will outputs the microseconds.

mesa/model.py
rht: I have removed the statement that it the timestamp is CPython's default. This is how they do it: https://github.com/python/cpython/blob/d4cea794a7b9b745817d2bd982d35412aef04710/Modules/_randommodule.c#L283. There is a PyObject interface to it, but it is a non-public module, `_random.Random`.

mesa/model.py
Corvince: Yes, but having the milliseconds (which are all 0 if no higher precision is available) provides no advantage. Why should a number with trailing zeros be better suitable as a random seed. What's your train of thought here?

rht: To make sure that simulations that are launched in parallel (that are within a second from each other) will have different seeds.

rht: The milliseconds/microseconds being 0's should only happen with `time.time()`.

Corvince: But the precision of datetime can't be better than that of time.time()! That's what I keep trying to tell you. 

There is just a difference between the display of your time. If you run on a machine with only second precision time.time() might for example return the value `1234`. 

If you run datetime.now() the datetime module will receive the same value - `1234`. The OS simply doesn't know better. But the  datetime is *required* to give you milliseconds. So it will return you `[...]34.000`. But you still end up with the same value within each second.

mesa/model.py
Corvince: I would say simply random.random() would be the appropriate choice. That also wouldn't require any comment, because the code itself would already "say": just use a random value. And that's what we want. If someone requires cryptographically secure seeds they can provide them. But I can't really imagine a use case for that.

mesa/model.py
rht: The elaborate SO answer wouldn't be worth it if it were to return 123.000000, but anyway, `random.random()` is the simplest way to go for it, hence, case closed.

mesa/model.py
```diff
@@ -21,7 +21,11 @@ class Model:
     def __new__(cls, *args: Any, **kwargs: Any) -> Any:
         """Create a new model object and instantiate its RNG automatically."""
         obj = object.__new__(cls)
-        obj._seed = kwargs.get("seed", None)
+        obj._seed = kwargs.get("seed")
+        if obj._seed is None:
+            # We explicitly specify the seed here so that we know its value in
+            # advance.
+            obj._seed = random.random()  # noqa: S311
```
rht, line 28: Had to add the ignore, otherwise Ruff said
> S311 Standard pseudo-random generators are not suitable for cryptographic purposes

mesa/model.py
Corvince: You demonstrated enough times that you are very capable of logical thinking. I don't know why you keep refusing to use that ability. I said or asked several times now: if the OS doesn't provide higher precision where should the milliseconds come from? There is no logical way to derive them. I'll add a comment on SO

rht: Why did you assume that the OS can't provide higher precision with `datetime.now()`? It's just that `time.time()` and `datetime.now()` may output different precision, depending on the system. Don't just conclude on your own whim that your assumption of the inner workings of `time.time()` and `datetime.now()` were correct in the first place.

Corvince: Well I worked a lot with datetimes and once with an embedded system that had no subsecond precision. Not at the same time, but I think it's fine to say my understanding is a bit more than a "whim". Plus,as I quoted, it says in front of the time docs that time() uses the best precision available. 

But I definitely agree that we can quit this conversation now. I just wanted to help you understand.

rht: https://docs.python.org/3/library/datetime.html#datetime.datetime.now

> If optional argument tz is None or not specified, this is like [today()](https://docs.python.org/3/library/datetime.html#datetime.datetime.today), but, if possible, supplies more precision than can be gotten from going through a [time.time()](https://docs.python.org/3/library/time.html#time.time) timestamp (for example, this may be possible on platforms supplying the C gettimeofday() function).

Is this not clear enough?

rht: I'm fine with moving on. I didn't see your last message before my comment.

tpike3: Please let's talk considerately to each other. I am going to resolve this conversation

PR #1817, opened 2023-09-23
Update mesa_viz_tornado in setup.py.

jackiekazil:
This repoints mesa_viz_tornado from the repo to pypi.


jackiekazil:
this is a duplicate... @tpike3 seems to have already done this.

PR #1818, opened 2023-09-23
frontend: Add experimental VegaGrid

Corvince:
Added an experimental VegaGrid via vega-altair. Here is a quick screencast of whats possible with this:

[screen-capture (3).webm](https://github.com/projectmesa/mesa/assets/13568919/efd86066-22ad-431a-8b97-862f0b43a75b)

So, we have the following unique features

- Tooltip for agent data - defaults to all JSON serializable agent data
- declaratively describe chart features instead of imperative portrayal functions
- Interactivity via click-handlers! 

The last one is made possible by vega-altair allowing to define an on_click handler. I than only had to wrap the click handler by first finding the corresponding agent by its unique_id and after any interactions updating the data values.

This means the example shown can be achieved with a simple click_handler that acts upon the clicked agent. That is we only have to define 
```python
def click_handler(agent):
    agent.type = 1 - agent.type
```

to achieve interactivity for changing the agent type!

Then we run the example with 

```python
page = JupyterViz(
    Schelling,
    model_params,
    measures=["happy", make_text(get_happy_agents)],
    space_drawer=make_vega_grid("type:N", click_handler=click_handler),
    name="Schelling",
)
page  # noqa
```

In total I would say the current implementation might be a bit rough, but I think a lot can be improved over time, adding more functions and allowing more customizations. This is just a start.

### DIFF
--- a/mesa/experimental/VegaGrid.py
+++ b/mesa/experimental/VegaGrid.py
@@ -1,0 +1,56 @@
+import json
+
+import altair as alt
+import solara
+
+
+def get_agent_data_from_coord_iter(data):
+    for agent, (x, y) in data:
+        if agent:
+            agent_data = json.loads(
+                json.dumps(agent.__dict__, skipkeys=True, default=str)
+            )
+            agent_data["x"] = x
+            agent_data["y"] = y
+            agent_data.pop("model", None)
+            agent_data.pop("pos", None)
+            yield agent_data
+
+
+def make_vega_grid(color=None, click_handler=None):
+    return lambda model, _: VegaGrid(model, _, color, click_handler)
+
+
+def VegaGrid(model, _agent_portrayal, color=None, on_click=None):
+    if color is None:
+        color = "unique_id:N"
+
+    data = solara.reactive(
+        list(get_agent_data_from_coord_iter(model.grid.coord_iter()))
+    )
+
+    def update_data():
+        data.value = list(get_agent_data_from_coord_iter(model.grid.coord_iter()))
+
+    def click_handler(datum):
+        if datum is None:
+            return
+        agent = model.schedule._agents[tuple(datum["unique_id"])]
+        on_click(agent)
+        update_data()
+
+    default_tooltip = [f"{key}:N" for key in data.value[0]]
+    chart = (
+        alt.Chart(alt.Data(values=data.value))
+        .mark_rect()
+        .encode(
+            x=alt.X("x:N", scale=alt.Scale(domain=list(range(model.grid.width)))),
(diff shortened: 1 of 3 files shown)


rht:
> simple, modern look

It's not that much different than Matplotlib's output. I use scatter plot for Matplotlib so that the markers can be replaced with custom pictures.

- Is this faster than the Matplotlib with png instead of svg?
- Does this support scatter plot with custom images, e.g. wolf-sheep?
- It doesn't seem to have network support: https://github.com/altair-viz/altair/issues/340

rht:
That Altair issue pointed to https://github.com/Zsailer/nx_altair, but this package hasn't been maintained for a while.

Corvince:
> > simple, modern look
> 
> It's not that much different than Matplotlib's output. I use scatter plot for Matplotlib so that the markers can be replaced with custom pictures.

Sorry I put that poorly yesterday and updated the original post. This isn't meant as a replacement for matplotlib, so I removed the comparative phrasing.

Corvince:
I don't know if these questions are specifically addressed at me and how they relate to this PR, but here you go


> - Is this faster than the Matplotlib with png instead of svg?

Probably not, haven't fully tested

> - Does this support scatter plot with custom images, e.g. wolf-sheep?

What do you mean by this? This PR? Then no. Altair? Then yes.

rht:
I don't have time to argue with your nitpicking over semantics (such as in https://github.com/projectmesa/mesa/pull/1811#issuecomment-1731267188).

I can see this PR as a successor to the ideas in #631 and https://github.com/Corvince/mesa-viz. So, we are clear that the `VegaGrid` is a complement to the default Matplotlib-based ones, for the use case when user wants interactivity, and that it is possibly faster (but needs benchmark). Given that this is added in a separate file, it doesn't affect default components. I'm fine with merging this soon.

Corvince:
> I can see this PR as a successor to the ideas in #631 and https://github.com/Corvince/mesa-viz. So, we are clear that the `VegaGrid` is a complement to the default Matplotlib-based ones, for the use case when user wants interactivity, and that it is possibly faster (but needs benchmark). Given that this is added in a separate file, it doesn't affect default components. I'm fine with merging this soon.

Thank you for the clarification. Maybe next time you can make such a statement up front, so it is clearer to me what you think in general about a PR. I sincerely didn't know if you think this is mergeable, you don't like it or if either depends on the answers to the questions you raised.

Corvince:
Reason to close this PR https://github.com/projectmesa/mesa/pull/1805#issuecomment-1737933085

Issue #1797, opened 2023-09-10
JupyterViz: Allow user to customize plots via Matplotlib's Axes.set API


rht:
**What's the problem this feature will solve?**
Plot customization via https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.set.html.

**Describe the solution you'd like**
As an example, users could specify plot title within the model's fixed parameters `{"xlabel": year, "title": "Price timeseries"}`, and the plotter could do
```python
ax.set(**user_configs)
```

Not sure if this is necessary, because currently, users could just draw Matplotlib plots by themselves if they need customization.

See also https://github.com/projectmesa/mesa-viz-tornado/issues/23.

quaquel:
#2430 adds a posprocess keyword that takes a callable and is called with the ax on which stuff is being plotted. This gives users a hook to do any further refinement.

PR #1819, opened 2023-09-24
perf: Speed up Solara space render

rht:
This speeds up space render for Schelling with 80x80 grid size from 1.2 s to 80 ms.

### DIFF
--- a/mesa/experimental/jupyter_viz.py
+++ b/mesa/experimental/jupyter_viz.py
@@ -270,7 +270,7 @@ def make_space(model, agent_portrayal):
     else:
         space_ax.scatter(**portray(model.grid))
     space_ax.set_axis_off()
-    solara.FigureMatplotlib(space_fig)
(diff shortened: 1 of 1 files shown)


Corvince:
Thanking for investigating the bottleneck and providing this fix. 

While I don't observe such a drastic speedup on my machine, this seems to be a better default for mesas use case.

I am just wondering if it would be possible to make the format user-providable, so if someone cares more about quality (to right click and save for example) they can choose. Or maybe even generalize so any kwargs are forwarded to FigureMatplotlib.

rht:
> I am just wondering if it would be possible to make the format user-providable, so if someone cares more about quality (to right click and save for example) they can choose. Or maybe even generalize so any kwargs are forwarded to FigureMatplotlib.

Sounds good. See also #1797. Regardless, this feature should be in a separate PR.

Corvince (approved):

Issue #1806, opened 2023-09-16
JupyterViz: Simulation step is very slow for huge grid size (e.g. 80x80)


rht:
This is a continuation of the discussion started in https://github.com/projectmesa/mesa/issues/1772#issuecomment-1713805080.
@rlskoeser suggested Altair, which might be faster than Solara's Matplotlib backend.

I did manual benchmark. I found that on my laptop (i5-1345U), the portray generation and `ax.scatter` took 80 ms, but the slowest part is actually Solara's `savefig`: https://github.com/widgetti/solara/blob/a747a680478653ab73c3f9323aeb5fee45147b60/solara/components/matplotlib.py#L54, which took 1.2 s.

I changed the output format from ~~png to svg~~svg to png, and the `savefig` elapsed went down from 1.2 s to 180 ms. I additionally experimented with updating the scatter plot with `set_offsets` (for x, y), and `set_sizes` for size: the portrayal generation and scatter update went down from 80 ms to 6 ms. The branch I used to experiment can be found at https://github.com/rht/mesa/tree/solara_perf.

This is very promising.

rht:
According to https://discuss.streamlit.io/t/plot-library-speed-trial/4688, Altair is orders of magnitude faster than Matplotlib.

Corvince:
Altair is also suitable for interactive explorations, where matplotlib only displays a static image in solara. (https://solara.dev/examples/libraries/altair)

Also we don't have to worry about thread safety since the actual plotting of altair happens in javascript (vega)

Corvince:
Here is some code for a very simple altair grid chart

```python
def altair_space(model, test):
    def get_data(agent, pos):
        if agent:
            return {"x": pos[0], "y": pos[1], "type": agent.type}

    data = list(
        filter(None, (get_data(agent, pos) for agent, pos in model.grid.coord_iter()))
    )
    chart = (
        alt.Chart(alt.Data(values=data))
        .mark_rect()
        .encode(x="x:O", y="y:O", color="type:N")
    )
    return solara.FigureAltair(chart)
```

But it does appear to be slower than Matplotlib :( Could you test this out @rht ?

rht:
Just tested. Altair is no-go for now. At first, I got this error
```
File /venv/lib/python3.11/site-packages/altair/utils/data.py:81, in limit_rows.<locals>.raise_max_rows_error()
     80 def raise_max_rows_error():
---> 81     raise MaxRowsError(
[...]

PR #1820, opened 2023-09-25
Agent size change based on size of figure

ankitk50:
If the number of agents in greater than 50, I use a linearly decreasing function to calculate the size of the agent. The lower limit for the agent size is 5.

### DIFF
--- a/mesa/experimental/jupyter_viz.py
+++ b/mesa/experimental/jupyter_viz.py
@@ -52,7 +52,9 @@ def JupyterViz(
         return model
 
     def calculate_space_size():
-        pass
+        if model_parameters["N"] > 50:
+            fixed_size = int(model_parameters["N"] / 5)
+            plt.rcParams["figure.figsize"] = (fixed_size, fixed_size)
 
     def calculate_agent_size():
         lower_limit = 5
@@ -76,21 +78,24 @@ def JupyterViz(
     # 3. Set up UI
     solara.Markdown(name)
     # calculate agent size based on number of users
-    agent_size_delta = calculate_agent_size()
+    # agent_size_delta = calculate_agent_size()
+    calculate_space_size()
+
     UserInputs(user_params, on_change=handle_change_model_params)
     ModelController(model, play_interval, current_step, set_current_step, reset_counter)
 
-    with solara.GridFixed(columns=2):
+    with solara.Row():
         # 4. Space
         if space_drawer == "default":
             # draw with the default implementation
-            make_space(model, agent_portrayal, delta=agent_size_delta)
+            make_space(model, agent_portrayal, delta=0)
         elif space_drawer:
             # if specified, draw agent space with an alternate renderer
             space_drawer(model, agent_portrayal)
         # otherwise, do nothing (do not draw space)
 
         # 5. Plots
+    with solara.Row():
         for measure in measures:
             if callable(measure):
                 # Is a custom object
--- a/mesa/experimental/jupyter_viz.py
+++ b/mesa/experimental/jupyter_viz.py
@@ -51,6 +51,20 @@ def JupyterViz(
         set_current_step(0)
         return model
 
+    def calculate_space_size():
+        pass
+
+    def calculate_agent_size():
+        lower_limit = 5
+        upper_limit = 50
+        num_agents = model_parameters["N"]
+        delta = 0
+        if model_parameters["N"] > 50:
+            new_agent_size = max(lower_limit, int(-0.9 * num_agents + 90))
+            delta = new_agent_size - upper_limit
+
+        return delta
+
     reset_counter = solara.use_reactive(0)
     model = solara.use_memo(
         make_model, dependencies=[*list(model_parameters.values()), reset_counter.value]
@@ -61,6 +75,8 @@ def JupyterViz(
 
     # 3. Set up UI
     solara.Markdown(name)
+    # calculate agent size based on number of users
+    agent_size_delta = calculate_agent_size()
     UserInputs(user_params, on_change=handle_change_model_params)
     ModelController(model, play_interval, current_step, set_current_step, reset_counter)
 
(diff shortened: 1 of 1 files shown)


mesa/experimental/jupyter_viz.py
```diff
@@ -51,6 +51,20 @@ def make_model():
         set_current_step(0)
         return model
 
+    def calculate_space_size():
+        pass
+
+    def calculate_agent_size():
```
rht, line 57: The Matplotlib marker size shouldn't depend on the number of agents. It should depend only on the grid size and the computer's screen size.

Corvince:
Thank you for starting to work on this 👍

mesa/experimental/jupyter_viz.py
rht: It should be more abstract, like screen width, because in a Jupyter notebook, the width should be even smaller.

ankitk50: hmm, let me think how this can be implemented. Currently the size of each agent is parsed via `agent_portrayal` and hence it's easier(also less compute expensive) to manipulate.

ankitk50: I think it would also be important to show all agents in a smaller window size, hence the implementation can work there. I try to come up with some scaling to fit the screen width.

ankitk50:
I have restructured the UI layout using some of the features by Solara. The sidebar can be minimised via to expand the user view: 

Screenshot 2023-09-26 at 15 57 57

rht:
That looks nice. We could adaptively change the layout by detecting whether the Solara component is run in a Jupyter notebook or not: https://stackoverflow.com/questions/15411967/how-can-i-check-if-code-is-executed-in-the-ipython-notebook. But for this PR, you should focus on solving the grid render for #1806. The layout restyling should be in a separate PR.

ankitk50:
I have now updated to change the size of the figure based on number of users. This is possibly the easiest way. This avoids user markers to come close as the number of users increase.

Changing the size of the markers itself does not look neat, although its possible.

Sure I can move future layout modifications to another PR.

ankitk50:
Thanks for the explaination, I can try this out. Meanwhile UI related changes is moved to PR #1825.

ankitk50:
The change maintains the size of the agents based on size of the figure.

mesa/experimental/jupyter_viz.py
```diff
@@ -51,6 +51,23 @@ def make_model():
...
+            plt.rcParams["figure.figsize"] = (fixed_size, fixed_size)
+
+    def calculate_agent_size():
+        aspect_ratio_high = (
+            0.5  # maintain this aspect ration b/w agent size and space size
+        )
+        aspect_ratio_low = 0.1  #
+        print()
+        size_of_agent = aspect_ratio_high * (
+            model_parameters["width"] * model_parameters["height"]
```
rht, line 67: (Must be GitHub hiccup. I remembered already commenting about this) the size is length^1, but the width * height is length^2, so it scales quadratically instead. It makes more sense to scale linearly over `min(width, height)`.

ankitk50: yes, the idea is to have a constant ratio b/w size of agents (for e.g. area of a circle) and size of the figure (area of a square). Hence the scaling is still linear i.e. the size of agent = aspect_ratio * size of figure.

rht: I just realized that the `s` param in Matplotlib scatter is length^2: https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.scatter.html. It's `markersize**2` without a multiplication by pi. We should document that size refers to Matplotlib size, and it is an area.

mesa/experimental/jupyter_viz.py
```diff
@@ -51,6 +51,23 @@ def make_model():
...
         return model
 
+    def calculate_space_size():
+        if model_parameters["N"] > 50:
+            fixed_size = int(model_parameters["N"] / 5)
+            model_parameters.update({"width": fixed_size, "height": fixed_size})
+            plt.rcParams["figure.figsize"] = (fixed_size, fixed_size)
+
+    def calculate_agent_size():
+        aspect_ratio_high = (
```
rht, line 61: Definition of aspect ratio is width/height: https://en.wikipedia.org/wiki/Aspect_ratio_(image). `fill_fraction` is a more descriptive term.

mesa/experimental/jupyter_viz.py
```diff
@@ -51,6 +51,23 @@ def make_model():
         set_current_step(0)
         return model
 
+    def calculate_space_size():
+        if model_parameters["N"] > 50:
```
rht, line 55: Once again, the grid size is independent of the number of agents. Having few agents and having many agents should result in the same marker size and figure size.

mesa/experimental/jupyter_viz.py
ankitk50: sure, makes more sense, thanks!

mesa/experimental/jupyter_viz.py
ankitk50: Yes, agents less than 50 the figure size of default used by matplotlib. I start to scale the it more num of agents. 

Ideal would be scale the image size based on screen size, but this doesn't seem feasible. Any other ideas?

rht: > I start to scale the it more num of agents.

Number of agents shouldn't have effect on image size for the grid.

EwoutH:
Really like this idea! What's needed to move this forward?

rht:
1. Remove number of agents from the formula to calculate size. I have said **several times** that grid cell size has nothing to do with the number of agents. You can have a huge grid with sparse agents, and yet the agent marker still needs to be not bigger than the grid cell size
2. https://github.com/projectmesa/mesa/pull/1825#discussion_r1349484373. Given that we haven't figured out a way to get screen size info to be sent to the Solara server, auto-scaling figure size needs to be removed for now.
3. Print functions haven't been removed

rht:
Secondary reason for https://github.com/projectmesa/mesa/pull/1825#discussion_r1349484373 is that it is not thread safe. Anything that is global operation via `plt.*` is not thread safe.

EwoutH:
@ankitk50 can we help you move this forward?

rht:
Superseded by #2049.

Issue #1831, opened 2023-10-07
Von Neumann grid hits `TypeError` on agents in corners


maskarb:
**Describe the bug**

I am recreating the Forest Fire example with Mesa 2, and sometimes I hit the following traceback when the model iterates on an agent in the corner of the grid, specifically with `Von Neumann neighborhood` (when using `Moore neighborhood`, I haven't seen this exception).

```
(99, 99) <<< coordinates of agent when exception is raised
Uncaught exception GET /ws (127.0.0.1)
HTTPServerRequest(protocol='http', host='127.0.0.1:8521', method='GET', uri='/ws', version='HTTP/1.1', remote_ip='127.0.0.1')
Traceback (most recent call last):
  File "/Users/mskarbek/projects/opinion-dynamics-leaderboard/.venv/lib/python3.11/site-packages/tornado/websocket.py", line 630, in _run_callback
    result = callback(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mskarbek/projects/opinion-dynamics-leaderboard/.venv/lib/python3.11/site-packages/mesa_viz_tornado/ModularVisualization.py", line 230, in on_message
    self.application.model.step()
  File "/Users/mskarbek/projects/opinion-dynamics-leaderboard/mesa_examples/forest_fire/forest_fire/model.py", line 50, in step
    self.schedule.step()
  File "/Users/mskarbek/projects/opinion-dynamics-leaderboard/.venv/lib/python3.11/site-packages/mesa/time.py", line 128, in step
    self.do_each("step", shuffle=True)
  File "/Users/mskarbek/projects/opinion-dynamics-leaderboard/.venv/lib/python3.11/site-packages/mesa/time.py", line 110, in do_each
    getattr(self._agents[agent_key], method)()
  File "/Users/mskarbek/projects/opinion-dynamics-leaderboard/mesa_examples/forest_fire/forest_fire/agent.py", line 32, in step
    for neighbor in self.model.grid.iter_neighbors(self.pos, False):
  File "/Users/mskarbek/projects/opinion-dynamics-leaderboard/.venv/lib/python3.11/site-packages/mesa/space.py", line 401, in <genexpr>
    if (cell := self._grid[x][y]) != self.default_val()
                ~~~~~~~~~~^^^
TypeError: list indices must be integers or slices, not tuple
```

**Expected behavior**

I do not expect a traceback when stepping on agents in the corners of the grid.

[...]

PR #1832, opened 2023-10-07
check if positions values are tuples

maskarb:
#1831 

Add extra check to the `accept_tuple_argument` wrapper. If `positions` is a tuple of tuples of length 2, then the `iter_cell_list_contents` raises a TypeError.

Example, given a cell_list of `((0, 98), (1, 99))`, this will unpack here `x, y in cell_list` as `x = (0, 98) and y = (1, 99)`.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -68,10 +68,9 @@ def accept_tuple_argument(wrapped_function: F) -> F:
     single-item list rather than forcing user to do it."""
 
     def wrapper(grid_instance, positions) -> Any:
-        if isinstance(positions, tuple) and len(positions) == 2:
-            return wrapped_function(grid_instance, [positions])
-        else:
-            return wrapped_function(grid_instance, positions)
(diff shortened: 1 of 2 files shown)


mesa/space.py
```diff
@@ -68,10 +68,9 @@ def accept_tuple_argument(wrapped_function: F) -> F:
     single-item list rather than forcing user to do it."""
 
     def wrapper(grid_instance, positions) -> Any:
-        if isinstance(positions, tuple) and len(positions) == 2:
-            return wrapped_function(grid_instance, [positions])
-        else:
-            return wrapped_function(grid_instance, positions)
+        if len(positions) == 2 and not isinstance(positions[0], tuple):
```
rht, line 71: This makes more sense, because the previous code mistakenly detects ((1, 2), (2, 3)) as `[positions]`.

rht (approved):

rht:
Merged, thank you!

PR #1717, opened 2023-06-17
Tutorial improvement

rht:

PR #1835, opened 2023-10-15
intro tutorial: Explain how to plot reporter of multiple agents

rht:
This is a section taken from #1717, in particular, "You can also plot a reporter of multiple agents over time".


rht:
cc: @EwoutH

tpike3 (approved):
Excellent thank you

Issue #1837, opened 2023-10-18
Document possible agent reporter inputs in the datacollector


EwoutH:
Currently in the data collector documentation its defined very well which inputs model reporters can take:

https://github.com/projectmesa/mesa/blob/d40bc5b475f667f65efc3c6dd256f56b1fe3f724/mesa/datacollection.py#L82-L100

However, this isn't done for the agent reports. Could this documentation be added?

rht:
I found it in https://github.com/projectmesa/mesa/blob/d40bc5b475f667f65efc3c6dd256f56b1fe3f724/mesa/datacollection.py#L179. Was added in https://github.com/projectmesa/mesa/commit/80511e2d71de85fd8d84035712ad0078fce32be6#diff-72396865b218fb20682afecba66d3b794e1306841d39490ed4037e26cba31b29L165. @tpike3 is the reason why you added the feature was for pickling issue in `multiprocessing`?

EwoutH:
I would be fine with removing the "Functions with parameters placed in a list" behaviour by the way, as long as we keep the syntax for agent_reporters and model_reporters identical (and thus remove it from both).

rht:
I agree with removing it in both agent and model reporters. I haven't compiled the list of breaking changes for Mesa 3.0 yet.

EwoutH:
Then I would propose removing it for the agent-reporter immediately, and adding a deprecation warning to the model-reporter ASAP.

To summerize, we propose to remove method nr 4 as it's practically identical to nr 1.

```
Model reporters can take four types of arguments:
        1. Lambda function:
           {"agent_count": lambda m: m.schedule.get_agent_count()}
        2. Method of a class/instance:
           {"agent_count": self.get_agent_count} # self here is a class instance
           {"agent_count": Model.get_agent_count} # Model here is a class
        3. Class attributes of a model:
           {"model_attribute": "model_attribute"}
        4. Functions with parameters that have been placed in a list:
           {"Model_Function": [function, [param_1, param_2]]}

        Agent reporters can similarly take:
        1. Attribute name (string) referring to agent's attribute:
           {"energy": "energy"}
        2. Lambda function:
           {"energy": lambda a: a.energy}
        3. Method of an agent class/instance:
           {"agent_action": self.do_action} # self here is an agent class instance
           {"agent_action": Agent.do_action} # Agent here is a class
        4. Functions with parameters placed in a list:
           {"Agent_Function": [function, [param_1, param_2]]}
```

[...]

PR #1838, opened 2023-10-18
DataCollector: Allow agent reporters to take class methods and functions with parameter lists

EwoutH:
Modify the `DataCollector` class to allow agent reporters to take methods of a class/instance and functions with parameters placed in a list (like model reporters), by extending the `_new_agent_reporter` method.

This implementation starts by checking if the reporter is an attribute string. If so, it creates a function to retrieve the attribute from an agent. Next, it checks if the reporter is a list. If it is, this indicates that we have a function with parameters, so it wraps that function to pass those parameters when called.

For any other type (like lambdas or methods), we assume they're directly suitable to be used as reporters.

Now, with this modification, agent reporters in the `DataCollector` class can take:

1. Attribute strings
2. Function objects (like lambdas)
3. Methods of a class/instance
4. Functions with parameters placed in a list

This approach ensures backward compatibility because the existing checks for attribute strings and function objects remain unchanged. The added functionality only extends the capabilities of the class without altering the existing behavior.

Closes: #1837

**Todo**
 - [x] Update and fix tests
 - [x] Add more docstring

Can be done in separate PRs:
 - Update examples
 - Update tutorial (optional)

### DIFF
--- a/mesa/datacollection.py
+++ b/mesa/datacollection.py
@@ -125,37 +125,37 @@ class DataCollector:
         self.model_reporters[name] = reporter
         self.model_vars[name] = []
 
-def _new_agent_reporter(self, name, reporter):
-    """Add a new agent-level reporter to collect.
+    def _new_agent_reporter(self, name, reporter):
+        """Add a new agent-level reporter to collect.
 
-    Args:
-        name: Name of the agent-level variable to collect.
-        reporter: Attribute string, function object, method of a class/instance, or
-                  function with parameters placed in a list that returns the
-                  variable when given an agent instance.
-    """
-    # Check if the reporter is an attribute string
-    if isinstance(reporter, str):
-        attribute_name = reporter
+        Args:
+            name: Name of the agent-level variable to collect.
+            reporter: Attribute string, function object, method of a class/instance, or
+                      function with parameters placed in a list that returns the
+                      variable when given an agent instance.
+        """
+        # Check if the reporter is an attribute string
+        if isinstance(reporter, str):
+            attribute_name = reporter
 
-        def attr_reporter(agent):
-            return getattr(agent, attribute_name, None)
+            def attr_reporter(agent):
+                return getattr(agent, attribute_name, None)
 
-        reporter = attr_reporter
+            reporter = attr_reporter
 
-    # Check if the reporter is a function with arguments placed in a list
-    elif isinstance(reporter, list):
-        func, params = reporter[0], reporter[1]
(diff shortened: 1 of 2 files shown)


EwoutH:
Code is ready to review. Will still work a bit on documentation.

> Is this a syntax sugar for an agent reporter of previously specified as lambda agent: func(agent, *params), but now with [func, params]?

Yes, that's completely right, the modification is essentially syntactic sugar, offering an alternative way to specify agent reporters that involve functions with additional parameters.

The big advantage of this however, is that it's now completely consistent with specifying model reporters. This way, if you can use the syntax for a model reporter, you can use it for an agent reporter, and visa versa.

A disadvantage could of course be the increased complexity of the code base. But, then I would argue, we should also report one of those syntax types from the model reporter. Because than that would be consistent, but I also don't think that's a desired option due to breaking backwards compatibility.

EwoutH:
Updated the docs, this PR is ready for review.

I might update the tutorial and some examples in separate PRs.

rht:
I personally think that passing `lambda agent: func(agent, *params)` doesn't require you to read extra documentation and is most likely what people would have come up with naturally.

I think it has nothing to do with the pickling issue in `multiprocessing` as I initially guessed in https://github.com/projectmesa/mesa/issues/1837#issuecomment-1769869159, because it will create a local anonymous function that is unpickleable anyway.

@jackiekazil @tpike3 should we merge this PR?

mesa/datacollection.py
```diff
@@ -130,16 +135,31 @@ def _new_agent_reporter(self, name, reporter):
...
         """
+        # Check if the reporter is an attribute string
         if isinstance(reporter, str):
             attribute_name = reporter
 
-            def reporter(agent):
+            def attr_reporter(agent):
                 return getattr(agent, attribute_name, None)
 
-            reporter.attribute_name = attribute_name
```
rht, line 142: This seems to be a bug introduced in https://github.com/projectmesa/mesa/commit/bc5a5cd949992fcaa750ab5c6a8fc1cdb3b0c968, and you fixed in this PR.

rht:
Content: LGTM. Concept: I prefer passing `lambda agent: func(agent, *params)`, but others might think otherwise.

jackiekazil:
I am good with this. @rht thank you for doing the review. 
LGTM.

jackiekazil (approved):

EwoutH:
Thanks for reviewing and merging!

PR #1841, opened 2023-10-23
Update to Ruff 0.1.1

EwoutH:
Note that Ruff has adopted a new version policy (https://docs.astral.sh/ruff/versioning/), similar to SemVer. So we can now do >=0.1.1,<0.2.0 to get bugfixes and deprecation warnings, but don't get new (syntax) features.

### DIFF
--- a/.github/workflows/build_lint.yml
+++ b/.github/workflows/build_lint.yml
@@ -74,7 +74,7 @@ jobs:
     - name: Lint with ruff
       # Include `--format=github` to enable automatic inline annotations.
       # Use settings from pyproject.toml.
-      run: ruff . --format=github --extend-exclude 'mesa/cookiecutter-mesa/*'
+      run: ruff . --output-format=github --extend-exclude 'mesa/cookiecutter-mesa/*'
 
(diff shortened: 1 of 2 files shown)


EwoutH:
Cleaned up the commits, ready for review.

See https://github.com/astral-sh/ruff/pull/8014 for the change from `--format=github` to `--output-format=github`.

.github/workflows/build_lint.yml
```diff
@@ -70,11 +70,11 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: "3.12"
-    - run: pip install ruff==0.0.275
+    - run: pip install "ruff>=0.1.1,<0.2.0"  # Update periodically
```
rht, line 73: I think `ruff~=0.1.1` should do.

rht:
The `ruff~=0.1.1` is syntactically equivalent to `>=0.1.1,<0.2.0` regardless of the versioning scheme used by Ruff.

rht (approved):

PR #1855, opened 2023-11-01
Update release.yml for trusted publishing

Corvince:
This uses [trusted publishing](https://docs.pypi.org/trusted-publishers/) to use auto-generated short-lived tokens exchanged directly between PyPi and GitHub. This is the currently recommended way to upload releases from Github Actions to PyPi. With this in place we don't need to generate and deploy api keys to GitHub, but someone with access to mesa on PyPi has to follow [this guide](https://docs.pypi.org/trusted-publishers/adding-a-publisher/) (I think that is either @jackiekazil or @tpike3 )

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,6 +20,8 @@ jobs:
   release:
     name: Deploy release to PyPI
     runs-on: ubuntu-latest
+    permissions: 
+      id-token: write
     steps:
       - name: Checkout source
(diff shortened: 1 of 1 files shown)


EwoutH (approved):

EwoutH:
Looks good!

tpike3 (approved):
LGTM --- Sorry for the delay on this we were trying to get 2.1.3 then 2.1.4 out the door.

PR #1857, opened 2023-11-02
Update History.rst for 2.1.3 release

tpike3:


jackiekazil:
we need to add #1860, which was just merged and we can release.

jackiekazil (approved):
LGTM

jackiekazil:
Squashing and merging... the other builds are unnecessary b/c the only file that was edited was the history.... I wonder if something in our build prevents the builds from happening when none of the Python files are changed.

PR #1870, opened 2023-11-08
Migrate setup from setup.py to pyproject.toml

rht:


rht:
This PR is needed so that the requirements can be pinned, and scanned declaratively by Dependabot https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions, and to seamlessly migrate to Poetry with a lockfile.

tpike3 (approved):
LGTM

PR #1873, opened 2023-11-13
Enhance Docstrings for Scheduler Classes in Mesa Time Module

EwoutH:
This commit improves the docstrings for each scheduler class and some key methods in the Mesa Time Module. The enhancements include detailed descriptions of the classes and methods, clarifying their functionality, usage, and the expected behavior of each scheduler type. 

Improvements include:
- Added comprehensive class descriptions for `BaseScheduler`, `RandomActivation`, `SimultaneousActivation`, `StagedActivation`, and `RandomActivationByType`, highlighting their specific roles and behaviors in agent activation within simulations.
- Included detailed method descriptions for `add`, `remove`, `step`, and other relevant methods, explaining their functionality and the parameters they use.
- Specified attributes for each class, providing clarity on the data each scheduler maintains and manipulates.
- Enhanced readability and understanding for users who need to implement or modify schedulers in their agent-based models.

No code or functional changes, only documentation.

### DIFF
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -152,8 +152,8 @@ class SimultaneousActivation(BaseScheduler):
     A scheduler that simulates the simultaneous activation of all agents.
 
     This scheduler is unique in that it requires agents to have both `step` and `advance` methods.
-    The `step` method is for activating the agent and staging any changes without applying them immediately.
-    The `advance` method then applies these changes, simulating simultaneous action.
+    - The `step` method is for activating the agent and staging any changes without applying them immediately.
+    - The `advance` method then applies these changes, simulating simultaneous action.
 
     This scheduler is useful in scenarios where the interactions between agents are sensitive to the order
     of execution, and a quasi-simultaneous execution is more realistic.
@@ -178,13 +178,13 @@ class SimultaneousActivation(BaseScheduler):
 class StagedActivation(BaseScheduler):
     """
     A scheduler allowing agent activation to be divided into several stages, with all agents executing one stage
-    before moving on to the next.
+    before moving on to the next. This class is a generalization of SimultaneousActivation.
 
     This scheduler is useful for complex models where actions need to be broken down into distinct phases
     for each agent in each time step. Agents must implement methods for each defined stage.
 
     The scheduler also tracks steps and time separately, allowing fractional time increments based on the number
-    of stages.
+    of stages. Time advances in fractional increments of 1 / (# of stages), meaning that 1 step = 1 unit of time.
 
     Inherits all attributes and methods from BaseScheduler.
 
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -46,17 +46,16 @@ class BaseScheduler:
     It assumes that each agent added has a `step` method which takes no arguments and executes the agent's actions.
 
     Attributes:
-        model (Model): The model instance associated with the scheduler.
-        steps (int): The number of steps the scheduler has taken.
-        time (TimeT): The current time in the simulation. Can be an integer or a float.
-        _agents (dict[int, Agent]): A private dictionary mapping agent IDs to agent instances.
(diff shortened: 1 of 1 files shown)


mesa/time.py
```diff
@@ -39,12 +39,24 @@
...
-    Assumes that each agent added has a *step* method which takes no arguments.
-
-    (This is explicitly meant to replicate the scheduler in MASON).
+    """
+    A simple scheduler that activates agents one at a time, in the order they were added.
+
+    This scheduler is designed to replicate the behavior of the scheduler in MASON, a multi-agent simulation toolkit.
+    It assumes that each agent added has a `step` method which takes no arguments and executes the agent's actions.
+
+    Attributes:
```
rht, line 48: It currently renders without formatting: https://mesa--1873.org.readthedocs.build/en/1873/apis/time.html.

It seems that you could use info fields: https://stackoverflow.com/questions/8649105/how-to-show-instance-attributes-in-sphinx-doc, which renders better. This is not a blocker to this PR getting merged. Only if you have the time for it.

mesa/time.py
```diff
@@ -39,12 +39,24 @@
...
+        model (Model): The model instance associated with the scheduler.
+        steps (int): The number of steps the scheduler has taken.
+        time (TimeT): The current time in the simulation. Can be an integer or a float.
+        _agents (dict[int, Agent]): A private dictionary mapping agent IDs to agent instances.
```
rht, line 52: This shouldn't be documented. The private implementation detail may change, while the API (`BaseScheduler.agents`) doesn't.

mesa/time.py
```diff
@@ -131,11 +149,20 @@ def step(self) -> None:
...
 class SimultaneousActivation(BaseScheduler):
-    """A scheduler to simulate the simultaneous activation of all the agents.
+    """
+    A scheduler that simulates the simultaneous activation of all agents.
+
+    This scheduler is unique in that it requires agents to have both `step` and `advance` methods.
+    The `step` method is for activating the agent and staging any changes without applying them immediately.
+    The `advance` method then applies these changes, simulating simultaneous action.
+
+    This scheduler is useful in scenarios where the interactions between agents are sensitive to the order
```
rht, line 159: This paragraph is informative.

mesa/time.py
```diff
@@ -150,15 +177,25 @@ def step(self) -> None:
...
-    """A scheduler which allows agent activation to be divided into several
-    stages instead of a single `step` method. All agents execute one stage
+    """
+    A scheduler allowing agent activation to be divided into several stages, with all agents executing one stage
     before moving on to the next.
 
-    Agents must have all the stage methods implemented. Stage methods take a
-    model object as their only argument.
+    This scheduler is useful for complex models where actions need to be broken down into distinct phases
+    for each agent in each time step. Agents must implement methods for each defined stage.
```
rht, line 185: May add that this class is a generalization of `SimultaneousActivation`.

mesa/time.py
EwoutH: `model.schedule.steps` (or `self.schedule.steps`) is frequently used to get the current timestep. I think this should be documented somewhere. Or is there another best practice to get the current timestep?

mesa/time.py
EwoutH: I just added list markers (`- `), that seems to do the trick

mesa/time.py
rht: Just `_agents`. Everything else is fine.

EwoutH: I removed `_agents` from the docs.

mesa/time.py
EwoutH: Done!

rht (approved):

rht:
Ruff and Black CI failed.

EwoutH:
Fixed

rht:
Merged, thank you.

PR #1875, opened 2023-11-15
refactor: Define current_step as reactive

rht:
This should reduce the number of arguments that needs to be passed to the components.


tpike3 (approved):
LGTM

PR #1879, opened 2023-11-16
space_drawer: Default to have model as the only argument

rht:
The new function signature requires fewer assumptions on the structure of space_drawer. See https://github.com/projectmesa/mesa-geo/issues/157#issuecomment-1793307510.


rht:
Closing because the custom space_drawer should be a Solara component instead of a function.

PR #1880, opened 2023-11-17
Replace Black with ruff format

rht:
See https://astral.sh/blog/the-ruff-formatter, with `ruff format` being
30x faster.

### DIFF
--- a/.github/workflows/build_lint.yml
+++ b/.github/workflows/build_lint.yml
@@ -84,6 +84,6 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: "3.12"
-    - run: pip install black[jupyter]
-    - name: Lint with black
-      run: black --check --exclude=mesa/cookiecutter-mesa/* .
+    - run: pip install ruff~=0.1.1  # Update periodically
+    - name: Lint with ruff format
+      run: ruff format --check .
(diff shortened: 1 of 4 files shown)


rht:
I marked this as draft because #1870 hasn't been merged yet.

rht:
@EwoutH the pre-commit Ruff and the GH Actions Ruff are redundant. Which one do you think should be kept? I find pre-commit to be faster.

.github/workflows/build_lint.yml
```diff
@@ -84,6 +84,6 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: "3.12"
-    - run: pip install black[jupyter]
-    - name: Lint with black
-      run: black --check --exclude=mesa/cookiecutter-mesa/* .
+    - run: pip install ruff~=0.1.1  # Update periodically
+    - name: Lint with ruff format
+      run: ruff format --check .
```
rht, line 89: The exclude is no longer necessary because the files in the cookiecutter folder have been renamed with .pytemplate as an extension.

rht:
And the versions will be automatically updated, and hence doesn't require manual maintenance #1864.

EwoutH:
> the pre-commit Ruff and the GH Actions Ruff are redundant. Which one do you think should be kept? I find pre-commit to be faster.

I would say pre-commit, since linting is primarily a pre-commit action and not necessarily tests.

rht:
OK, I have removed it in this PR.

rht:
Rebased. This PR is ready.

tpike3 (approved):
LGTM

PR #1882, opened 2023-11-17
Migrate from setuptools to hatch

rht:
Depends on #1870 to be merged.

Several resources recommended to replace setuptools with either Flit, Hatch, Poetry, or PDM. Both Flit and Hatch are maintained by PyPA. Flit is restricted to pure Python projects. According to [pyOpenSci](https://www.pyopensci.org/python-package-guide/package-structure-code/python-package-build-tools.html), PDM's documentation can be confusing, and has only 1 maintainer. That leaves Hatch and Poetry. There was a discussion about dropping Pipenv and instead support Poetry. I wasn't able to find the thread from the issue search. However, it should be fine to go from setuptools to Hatch first, because the jump is much smaller: the diff for pyproject.toml in this PR is <10 LOC.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,9 +30,9 @@ jobs:
         with:
           python-version: "3.12"
       - name: Install dependencies
-        run: pip install -U pip build wheel setuptools
+        run: pip install -U pip hatch
       - name: Build distributions
-        run: python -m build
+        run: hatch build
       - name: Upload package as artifact to GitHub
         if: github.repository == 'projectmesa/mesa' && startsWith(github.ref, 'refs/tags')
         uses: actions/upload-artifact@v3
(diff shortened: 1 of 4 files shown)


rht:
Rebased. This PR is ready.

tpike3:
Sorry this took so long, I always get nervous messing with the project management tools. 

I am good with this. It is faster and cleaner and as you said we can always switch to poetry if that becomes the standard. 

The only thing I would ask is can you add the badge to the readme --- [Hatch project](https://github.com/pypa/hatch)

rht:
Rebased, and the Hatch badge added.

Corvince (approved):

Corvince:
I am merging this, I had good (albeit very small) experience with hatch

PR #1888, opened 2023-11-27
space: Document empties property

EwoutH:
Document the empties property in the Mesa Space module for the different Grids.

I might have overdone it a little, but that's in the spirit of [PEP 20](https://peps.python.org/pep-0020/):

Only documentation changes, no functional ones.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -492,9 +492,18 @@ class SingleGrid(_Grid):
     bottom-left and [width-1, height-1] is the top-right. If a grid is
     toroidal, the top and bottom, and left and right, edges wrap to each other.
 
+    This class provides a property `empties` that returns a set of coordinates
+    for all empty cells in the grid. It is automatically updated whenever
+    agents are added or removed from the grid. The `empties` property should be
+    used for efficient access to current empty cells rather than manually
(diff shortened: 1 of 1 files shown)


mesa/space.py
```diff
@@ -526,6 +535,9 @@ class MultiGrid(_Grid):
     bottom-left and [width-1, height-1] is the top-right. If a grid is
     toroidal, the top and bottom, and left and right, edges wrap to each other.
 
+    Unlike SingleGrid, MultiGrid does not inherently track empty cells due to
```
rht, line 538: `empties` are tracked in `MultiGrid`. `MultiGrid` inherits from `_Grid`, which defines the `empties` property.
```
In [1]: import mesa
g =
In [2]: g = mesa.space.MultiGrid(5,5,False)

In [3]: g.empties
Out[3]:
{(0, 0),
 (0, 1),
...
```

EwoutH: You're right, will update that

EwoutH: Updated!

rht (approved):

rht:
Merged, thank you.

PR #1890, opened 2023-11-27
Add DiscreteEventScheduler

EwoutH:
This PR introduces the DiscreteEventScheduler, an addition to the Mesa time module. This new scheduler is designed for discrete event simulation, where agent steps are scheduled to occur at specific simulation times, rather than at regular, step-wise intervals.

Key Features:
- The scheduler advances the simulation to the time of the next scheduled event, handling multiple events efficiently.
- If multiple events are scheduled for the same time, their execution order is randomized. This is achieved by adding a secondary sorting criterion based on a random value. This ensures fairness and reduces bias that might occur from consistent ordering.
- The `model.step()` works the same by default: It advances time by 1. The DiscreteEventScheduler has a keyword argument `time_step` (default 1) which is the period in which the model is advanced by each model.step() call. In that period, agent may execute their step function zero, one or multiple times, depending on the amount of eventa scheduled for that agent.
  - This allows easy adaption from current models, and to gather data at regular intervals.

Modifications:
- Added the DiscreteEventScheduler class with methods to schedule and process events.
- Modified the `schedule_event` method to include a random value for secondary sorting in the event queue.
- Adjusted the `step` method to handle events within a given time frame, advancing the simulation in fixed time steps.
- Modified the `add` method to have a `schedule_now` keyword, which is True by default. When True, the first event for that agent is immediately added to the schedule. 

Two example models are available, a very simple one and a slightly more complex one.

```Python
from mesa import Model, Agent
from mesa.time import DiscreteEventScheduler
import random

class MyAgent(Agent):
        # Agent initialization code here

    def step(self):
        # Agent action code here
        # For example, schedule next action after a random delay using schedule_in
        print(f"Agent {self.unique_id} executing step at time {self.model.schedule.time}")
        delay = random.uniform(1, 2)
        self.model.schedule.schedule_in(delay, self)

class MyModel(Model):
    def __init__(self, num_agents):
        self.schedule = DiscreteEventScheduler(self, time_step=1)
        # Initialize and add agents
        for i in range(num_agents):
            agent = MyAgent(i, self)
            self.schedule.add(agent, schedule_now=True)

    def step(self):
        self.schedule.step()  # Execute events for the next time step

# Create and run the model
model = MyModel(num_agents=5)
for i in range(10):  # Run for 20 steps
    model.step()
```
More complex example:

<details>

```Python
from mesa import Agent, Model
from mesa.time import DiscreteEventScheduler
import random


class Prey(Agent):
        self.energy = 5

    def step(self):
        # print(f"Prey {self.unique_id} executing step at time {self.model.schedule.time}")
        if self.energy <= 0:
            self.model.schedule.remove(self)
        else:
            self.move()
            self.energy -= 1
            self.eat()
            # Schedule next move at a random time in the future
            next_event_time = self.model.schedule.time + random.uniform(1, 2)
            self.model.schedule.schedule_event(next_event_time, self)

    def eat(self):
        self.energy += random.uniform(1, 2)
    def move(self):
        # Move to a random neighboring cell
        possible_steps = self.model.grid.get_neighborhood(
            self.pos,
            moore=True,
            include_center=False
        )
        new_position = random.choice(possible_steps)
        self.model.grid.move_agent(self, new_position)

    def reproduce(self):
        # If another prey is in the same cell, reproduce
        this_cell = self.pos
        x, y = this_cell
        this_cell_agents = self.model.grid[x][y]
        for agent in this_cell_agents:
            if isinstance(agent, Prey):
                child = Prey(self.model.next_id(), self.model)
                self.model.schedule.add(child)
                self.model.grid.place_agent(child, self.pos)
                return


class Predator(Agent):
        self.hunger = 5

    def step(self):
        # print(f"Predator {self.unique_id} executing step at time {self.model.schedule.time}")
        if self.hunger <= 0:
            self.model.schedule.remove(self)
        else:
            self.hunt()
            self.hunger -= 1
            # Schedule next hunt at a random time in the future
            next_event_time = self.model.schedule.time + random.uniform(2, 4)
            self.model.schedule.schedule_event(next_event_time, self)

    def hunt(self):
        # If a prey is in the same cell, eat it
        this_cell = self.pos
        x, y = this_cell
        this_cell_agents = self.model.grid[x][y]
        for agent in this_cell_agents:
            if isinstance(agent, Prey):
                self.model.schedule.remove(agent)
                self.model.grid.remove_agent(agent)

                self.hunger += random.uniform(5, 10)
                return
        else:
            # Move to a neighboring cell
            self.move()

    def move(self):
        # Move to a random neighboring cell
        possible_steps = self.model.grid.get_neighborhood(
            self.pos,
            moore=True,
            include_center=False
        )
        new_position = random.choice(possible_steps)
        self.model.grid.move_agent(self, new_position)

class EcosystemModel(Model):
        super().__init__()
        self.num_agents = N
        self.grid = MultiGrid(width, height, True)
        self.schedule = DiscreteEventScheduler(self)
        self.prey = []
        self.predators = []
            model_reporters={
                "N": lambda m: m.schedule.time,
                "Prey": lambda m: len(m.prey),
                "Predators": lambda m: len(m.predators),
                "Prey average energy": lambda m: sum([a.energy for a in m.prey]) / len(m.prey) if len(m.prey) > 0 else 0,
                "Predator average hunger": lambda m: sum([a.hunger for a in m.predators]) / len(m.predators) if len(m.predators) > 0 else 0,
            }
        )

        # Create predators and prey
        for i in range(self.num_agents):
            if random.random() < 0.5:
                a = Prey(self.next_id(), self)
                self.prey.append(a)
            else:
                a = Predator(self.next_id(), self)
                self.predators.append(a)
            self.schedule.add(a, schedule_now=True)
            # Place the agent on the grid
            x = random.randrange(self.grid.width)
            y = random.randrange(self.grid.height)
            self.grid.place_agent(a, (x, y))

    def step(self):
        self.schedule.step()
        # Update prey and predator lists
        self.prey = [a for a in self.schedule.agents if isinstance(a, Prey)]
        self.predators = [a for a in self.schedule.agents if isinstance(a, Predator)]


model = EcosystemModel(N=25, width=10, height=10)
for i in range(100):  # Run for 25 steps
    model.step()
print(model.datacollector.get_model_vars_dataframe())
```

</details>

Both are also available in this ZIP, and below: [discrete_event_example_models.zip](https://github.com/projectmesa/mesa/files/13479627/discrete_event_example_models.zip)

#### Todo:
- [x] Check if manually needing to schedule first event can be integrated (yes!)
- [x] Add tests

Ready for review!

### DIFF
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -26,7 +26,6 @@ Key concepts:
 from __future__ import annotations
 
 import heapq
-import random
 from collections import defaultdict
 
 # mypy
@@ -386,7 +385,7 @@ class DiscreteEventScheduler(BaseScheduler):
             raise ValueError(f"Scheduled time ({time}) must be >= the current time ({self.time})")
         event = (
             time,
-            random.random(),
+            self.model.random.random(),
             agent,
         )  # Add a random value for secondary sorting
         heapq.heappush(self.event_queue, event)
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -362,6 +362,17 @@ class DiscreteEventScheduler(BaseScheduler):
         schedule_in(delay, agent): Schedule an event after a specified delay.
         step(): Execute all events within the next time_step period.
         get_next_event_time(): Returns the time of the next scheduled event.
+
+    Usage:
+        1. Instantiate the DiscreteEventScheduler with a model instance and a time_step period.
+        2. Add agents to the scheduler using schedule.add(). With schedule_now=True (default),
+              the first event for the agent will be scheduled immediately.
+        3. In the Agent step() method, schedule the next event for the agent
+              (using schedule_in or schedule_event).
+        3. Add self.schedule.step() to the model's step() method, as usual.
+
+    Now, with each model step, the scheduler will execute all events within the
+    next time_step period, and advance time one time_step forward.
     """
 
     def __init__(self, model: Model, time_step: TimeT = 1) -> None:
@@ -413,3 +424,16 @@ class DiscreteEventScheduler(BaseScheduler):
         if not self.event_queue:
             return None
         return self.event_queue[0][0]
+
+    def add(self, agent: Agent, schedule_now: bool = True) -> None:
+        """Add an Agent object to the schedule and optionally schedule its first event.
+
+        Args:
+            agent: An Agent to be added to the schedule. Must have a step() method.
+            schedule_now: If True, schedules the first event for the agent immediately.
+        """
+        super().add(agent)  # Call the add method from BaseScheduler
+
+        if schedule_now:
+            # Schedule the first event immediately
+            self.schedule_event(self.time, agent)
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -354,8 +354,8 @@ class DiscreteEventScheduler(BaseScheduler):
     Attributes:
         model (Model): The model instance associated with the scheduler.
         event_queue (list): A priority queue of scheduled events.
-        time_step (int): The fixed time period by which the model advances on
-                         each step. Defaults to 1.
+        time_step (int or float): The fixed time period by which the model advances
+                                  on each step. Defaults to 1.
 
     Methods:
         schedule_event(time, agent): Schedule an event for a specific time.
@@ -364,13 +364,15 @@ class DiscreteEventScheduler(BaseScheduler):
         get_next_event_time(): Returns the time of the next scheduled event.
     """
 
-    def __init__(self, model, time_step=1):
+    def __init__(self, model: Model, time_step: TimeT = 1) -> None:
         super().__init__(model)
-        self.event_queue = []
-        self.time_step = time_step  # Fixed time period for each step
+        self.event_queue: list[tuple[TimeT, float, Agent]] = []
+        self.time_step: TimeT = time_step  # Fixed time period for each step
 
-    def schedule_event(self, time, agent):
+    def schedule_event(self, time: TimeT, agent: Agent) -> None:
         """Schedule an event for an agent at a specific time."""
+        if time < self.time:
+            raise ValueError(f"Scheduled time ({time}) must be >= the current time ({self.time})")
         event = (
             time,
             random.random(),
@@ -378,16 +380,18 @@ class DiscreteEventScheduler(BaseScheduler):
         )  # Add a random value for secondary sorting
         heapq.heappush(self.event_queue, event)
 
-    def schedule_in(self, delay, agent):
+    def schedule_in(self, delay: TimeT, agent: Agent) -> None:
         """ Schedule an event for an agent after a specified delay. """
+        if delay < 0:
+            raise ValueError(f"Delay must be non-negative")
         event_time = self.time + delay
         self.schedule_event(event_time, agent)
 
-    def step(self):
+    def step(self) -> None:
         """Execute the next event and advance the time."""
         end_time = self.time + self.time_step
 
-        while self.event_queue and self.event_queue[0][0] < end_time:
+        while self.event_queue and self.event_queue[0][0] <= end_time:
             # Get the next event
             time, _, agent = heapq.heappop(
                 self.event_queue
@@ -404,7 +408,7 @@ class DiscreteEventScheduler(BaseScheduler):
         self.time = end_time
         self.steps += 1
 
-    def get_next_event_time(self):
+    def get_next_event_time(self) -> TimeT | None:
         """Returns the time of the next scheduled event."""
         if not self.event_queue:
             return None
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -385,11 +385,8 @@ class DiscreteEventScheduler(BaseScheduler):
             raise ValueError(
                 f"Scheduled time ({time}) must be >= the current time ({self.time})"
             )
-        event = (
-            time,
-            self.model.random.random(),
-            agent,
-        )  # Add a random value for secondary sorting
+        # Create an event, sorted first on time, secondary on a random value
+        event = (time, self.model.random.random(), agent)
         heapq.heappush(self.event_queue, event)
 
     def schedule_in(self, delay: TimeT, agent: Agent) -> None:
@@ -404,10 +401,8 @@ class DiscreteEventScheduler(BaseScheduler):
         end_time = self.time + self.time_step
 
         while self.event_queue and self.event_queue[0][0] <= end_time:
-            # Get the next event
-            time, _, agent = heapq.heappop(
-                self.event_queue
-            )  # Ignore the random value during unpacking
+            # Get the next event (ignore the random value during unpacking)
+            time, _, agent = heapq.heappop(self.event_queue)
 
             # Advance time to the event's time
             self.time = time
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -359,6 +359,7 @@ class DiscreteEventScheduler(BaseScheduler):
 
     Methods:
         schedule_event(time, agent): Schedule an event for a specific time.
+        schedule_in(delay, agent): Schedule an event after a specified delay.
         step(): Execute all events within the next time_step period.
         get_next_event_time(): Returns the time of the next scheduled event.
     """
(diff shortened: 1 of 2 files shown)


rht:
How do you implement Poisson activation with this?

mesa/time.py
```diff
@@ -327,3 +329,111 @@ def get_type_count(self, type_class: type[Agent]) -> int:
...
+        self.event_queue: list[tuple[TimeT, float, Agent]] = []
+        self.time_step: TimeT = time_step  # Fixed time period for each step
+
+    def schedule_event(self, time: TimeT, agent: Agent) -> None:
+        """Schedule an event for an agent at a specific time."""
+        if time < self.time:
+            raise ValueError(f"Scheduled time ({time}) must be >= the current time ({self.time})")
+        event = (
+            time,
+            random.random(),
```
rht, line 389: I think this needs to be `self.model.random.random()`. Because the random source could be from NumPy, and has been seeded. Hardcoding it to `random.random` is a gotcha that may cause non-deterministic behavior.

EwoutH:
> How do you implement Poisson activation with this?

You're probably talking about an arrival process right? Interesting question, could you expand a bit on what you're exactly thinking about?

EwoutH:
Right, I thought about it a bit and I think I understand what you mean (please expand a bit next time. PEP 20, explicit is better than implicit).

We are thinking at this from two different ways:
- I'm thinking from an agent-central behavior principle: An agent takes a certain, stochastic time to do something, and then wants to continue with the next thing, irrespectabele from if the model time step is done or not.
- You're thinking from a model-coordinated perspective, in which the model draws the next (arrival) time for an event and schedules that.

So currently, the model_step uses a fixed time step. I build it that way to keep it consistent with the other schedulers (you can still do `model.step` and it advances some predicable time period `time_step`). This way, data can also be collected in regular intervals, as currently can be done with all other schedulers.

Basically I made the (implicit) design decision: Agents are scheduled with discrete events, but the model is not.

However, if you want to _centrally_ schedule events based on some distribution, like drawing from the Poisson distribution to simulate stochastic arrival times, the fixed time_step of the model doesn't allow that (currently).

If we are going to modify this, the most logical option is for the model to step at each event. But then what does a model step mean, since it's the same as an agent step.

I can see a few options:
- Just don't support it (for now / in this scheduler)
- Ask another entity (Agent) to schedule the next stochastic event
- Drop the fixed model time step

I tried a few quick and dirty implementations, but each option with dropping the fixed model time step looks more complicated an note really like a Mesa scheduler anymore. Therefore, this also becomes a bit of a scoping decision.

I think we need input from @jackiekazil, @tpike3 and @Corvince for that. To what extend do we want to support discrete-event simulation, and at what costs?

EwoutH:
As for the second option, letting arriving agent schedule the next one, we could add a `schedule_poisson` method to the DiscreteEventScheduler:

```Python
def schedule_poisson(self, rate: float, agent: Agent) -> None:
    """ Schedule an event based on Poisson process. """
    if rate <= 0:
        raise ValueError("Rate must be positive")
    delay = random.expovariate(rate)
    self.schedule_in(delay, agent)
```
Then, an arrival process could look something like this:

```Python
class CustomerAgent(Agent):

    def step(self):
        # Implement customer's behavior here
        # Example: print("Customer", self.unique_id, "is being served.")
        
        # Schedule the next customer's arrival
        next_customer = CustomerAgent(self.unique_id + 1, self.model)
        self.model.schedule.add(next_customer)
        self.model.schedule.schedule_poisson(self.model.lambda_arrival, next_customer)

class ServiceCenterModel(Model):
    def __init__(self, lambda_arrival):
        self.schedule = DiscreteEventScheduler(self)
        self.lambda_arrival = lambda_arrival
        self.next_customer_id = 0

        # Schedule the first customer
        first_customer = CustomerAgent(self.next_customer_id, self)
        self.schedule.add(first_customer)
        self.schedule.schedule_poisson(self.lambda_arrival, first_customer)
        self.next_customer_id += 1

    def step(self):
        self.schedule.step()

# Example usage
lambda_arrival = 0.5  # Average rate of customer arrival
model = ServiceCenterModel(lambda_arrival)
for _ in range(10):  # Simulate 10 time steps
    model.step()
```

tpike3:
Wow! This is great thanks @EwoutH!

Regarding the Poisson activation, I don't think it is necessary for this class.  Although for a different `mesa.time`  feature, we should add a `PoissonActivation` scheduler. 

I am fine with merging....thanks!

EwoutH:
Thanks!

> Regarding the Poisson activation, I don't think it is necessary for this class. Although for a different mesa.time feature, we should add a PoissonActivation scheduler.

I think adding the `schedule_poisson` is fairly easy (5 lines of code) and helps some of the situations. On the other had, then we might also want to Gaussian, Uniform, Exponential, etc., so maybe we can better just give some examples on how to do that with the current implementations.

What do you think? (also @rht)

What do you think would be useful for documentation. I attempted to get the docstring as clear as possible, but it might be nice to have a small tutorial or example model somewhere. What do you think would be most useful?

EwoutH:
Had a nice long bike ride, and thought about it some.

In the core, Mesa is an agent-based modelling library. Agent-based modelling is built from the postulate that relatively simple behavior of many individual entities can lead to complex, emergent behavior. In this postulate, the environment is mostly just a shared state, with some dimensions like space and time, though which agent communicate. The agent is active, while the model is passive. 

The current `DiscreteEventScheduler` fits in that narrative, an agent undertakes some task, which takes some fixed or variable amount of time, and then decides what's going to do next. Task/step scheduling is agent driven. It might fit it even better than the other schedulers.

In central discrete-event scheduler, events are mostly driven by a shared entity or model. So flips the principle around: The model decides what's happening to the agents, the model is active while the agents are passive. So while that's extremely interesting, it start to become another domain.

Therefore, I would say that the current implementation is very much in line with what Mesa tries to attempt, and other extension might not be.

The scheduler might need a different name though.

EwoutH:
One potential issue could be that at some point multiple events are (unintended) added to the stack for one agent, in which case it might do too many steps. I tried some mechanisms to circumvent this, but it isn't easy to get events for specific agents out of the `heapq` queue. So I will consider that future improvement.

As long as you schedule one event for each agent from the model, and then just let the agent itself schedule it's next event when it's done, it works great.

So this thing is far from perfect and far from done, but I think it's a good start, and can be reasonable well extended with future desired behavior. I am mostly thinking about tasks interruptions and system-level events. I've made an issue there at #1893.

mesa/time.py
EwoutH: Updated

mesa/time.py
```diff
@@ -327,3 +328,113 @@ def get_type_count(self, type_class: type[Agent]) -> int:
...
+    A scheduler for discrete event simulation in Mesa.
+
+    This scheduler manages events where each event is associated with a
+    specific time and agent. The scheduler advances time not in fixed
+    increments, but to the moment the next event is scheduled to occur.
+
+    This implementation uses a priority queue (heapq) to manage events. Each
+    event is a tuple of the form (time, random_value, agent), where:
+        - time (float): The scheduled time for the event.
+        - random_value (float): A secondary sorting criterion to randomize
```
rht, line 344: Would have been clearer to say that the `random_value` is used only once for the `heapq.push` operation. I had to check elsewhere in the code for its usage, until I realized of this.

rht (approved):

rht:
Implementation LGTM. Waiting for agreement from the others for the merge. The Poisson activation can be showcased in an example model or in the how-to guide.

mesa/time.py
EwoutH: Read it for again, I think it’s quite clear, especially together with the comments lower in the code..

EwoutH:
Thanks! Tom already approved right? Anyone else you would like agreement from? (if so, maybe tag them)

tpike3:
I think 2 is sufficient, so I will merge. Thanks!

If you could also put the poisson activation into a [how to](https://github.com/projectmesa/mesa/blob/main/docs/howto.rst) I think that  would be very useful

EwoutH:
Awesome, thanks for merging!

Will work on a How To entry. By the way, I saw we're converting some docs from rst to md, is that also the plan for the howto.rst file?

Issue #1889, opened 2023-11-27
Proposal: Adding Cell Properties to Grids in Mesa


EwoutH:
_This post is now slightly out of date, but preserved for historical context. See #1898 for the proposed implementation._

This proposal suggests two potential methods for introducing cell properties to grids in Mesa: a base class extension and a Mixin class. These enhancements will enable users to assign and manage custom properties (such as color or type) to grid cells, similar to [patches](https://ccl.northwestern.edu/netlogo/bind/primitive/patches.html) in NetLogo. This functionality can be beneficial for a wide range of simulations where the environment's characteristics are as crucial as the agents' behaviors.

### Motivation
Introducing properties to grid cells in agent-based models adds significant depth and realism to simulations. These properties can represent various environmental factors like terrain type, resources, or pollution levels, impacting how agents interact and behave. For example, in ecological simulations, cell properties might dictate animal movement based on land fertility or water availability. In urban models, properties could represent different neighborhood characteristics, influencing decisions made by agents representing residents or developers. This enhancement not only allows for more detailed and realistic modeling scenarios but also makes simulations more versatile and applicable to a wide range of fields, from ecology and urban planning to social sciences.

### Why not just a layer of agents (in a MultiGrid)?
- From a technical standpoint, cell properties can reduce computational overhead, as they avoid the need to create and manage numerous agent objects for static or slow-changing environmental features. They don't need to be added to a schedule, since they're passive entities.
- Conceptually, cell properties provide a more direct and intuitive representation of the environment, enhancing model clarity and interpretability. They allow for a straightforward distinction between dynamic agents and the static or semi-static environment, reducing unnecessary complexity and improving the ease of maintenance and analysis of the model.

### Implementation
Here are two high-level options how to integrate grid cell properties to the Mesa Space module.

#### Option 1: Base Class Extension

**Implementation:**

Extend the existing grid classes (`SingleGrid`, `MultiGrid`, `HexSingleGrid`, `HexMultiGrid`) to include a dictionary for cell properties.

**Advantages:**
- Direct and straightforward implementation.
- Clear inheritance structure.

**Disadvantages:**
- Modifies the original Mesa grid classes.
- Less flexible in terms of mixing with other potential extensions.

**Usage Example:**

```python
from mesa import Model
from mesa.space import MultiGrid

class MultiGridWithProperties(MultiGrid):
    def __init__(self, width, height, torus):
        super().__init__(width, height, torus)
[...]

PR #1898, opened 2023-12-04
space: Implement PropertyLayer and _PropertyGrid

EwoutH:
### Summary
This PR introduces a system for managing cell properties within Mesa's grid classes. It adds a `PropertyLayer` class and extends the `SingleGrid` and `MultiGrid` classes to support cell properties, enhancing the capability of Mesa for more detailed environmental modeling in simulations.

### Motive
The addition of cell properties brings a new dimension to agent-based models, allowing them to simulate environmental factors (e.g., terrain types, resources, pollution levels) more realistically. 

From a technical perspective, cell properties reduce computational overhead by negating the need for numerous agent objects for static environmental features. Conceptually, they provide a clearer and more intuitive representation of the environment, streamlining model complexity and facilitating analysis.

This feature enhances Mesa's modeling capabilities by enabling dynamic interaction between agents and their environment. It allows for simulations where environmental conditions directly influence agent behavior, thus broadening the scope and depth of possible models.

### Implementation
pako_eNrdVd9vmzAQ_lcsP9GNROR3gqpK0yrtZZUm5W1CQi6-EKtgI2Pasiz_-w5CgqGk27S38ZDA3Xd3n-8-2wcaKQ7Up1HC8vxesFizNJAEn9pCvmmVgTblV1aCJoeTq3pGW6OFjIlkKVhWIQ15Edzse7Y9iHhvLKMs0qwcS860ZiXhzLDW-TEHE0aQJM5npTQXkhkgmcqFEUq65JklB

#### PropertyLayer Class
The `PropertyLayer` class represents a layer of properties associated with the cells of a grid. It includes the following functionalities:

- **Initialization:** The constructor initializes a 2D NumPy array to store property values for each cell, with a specified default value and data type.
- **Cell Value Management:** Methods to set or modify the value of individual cells or batches of cells, either unconditionally or based on specified conditions.
- **Property Manipulation:** Functions to apply operations (such as lambda functions or NumPy universal functions) to modify cell values, either individually or in a batch mode.
- **Property Querying:** Capabilities to select cells based on certain conditions and perform aggregate operations across all cells (e.g., sum, average).

#### _PropertyGrid Class
The `_PropertyGrid` class extends the existing `_Grid` class, adding support for managing multiple `PropertyLayer` instances. Key features include:

- **Property Layer Management:** Methods to add and remove property layers from the grid. Each property layer is identified by a unique name and is managed independently.
- **Interaction with Agents:** Integration of property layers with agent actions, allowing agents to interact with and modify cell properties as part of their behavior.
- **Enhanced Grid Functionalities:** Extension of existing grid functionalities to incorporate property-based conditions. For example, enabling agents to move to cells with specific properties or query neighboring cells based on property values.

#### Integration with Grid Classes
The existing grid classes, such as `SingleGrid` and `MultiGrid`, are derived from `_PropertyGrid`. This inheritance allows them to leverage the property layer functionalities seamlessly. These enhanced grid classes enable modelers to define and manipulate cell properties directly, enriching the modeling capabilities of Mesa.

#### Implementation Details
- The implementation is compatible with existing Mesa models, ensuring no disruption to current users.
- It supports lambda functions and NumPy ufuncs, providing flexibility in defining cell operations.

### API examples
Here are some code examples of how these features could be utilized.

Initializing a Property Layer and Setting Cell Values
```python
import numpy as np
from mesa.space import SingleGrid, PropertyLayer

# Initialize a SingleGrid with a PropertyLayer
grid = SingleGrid(10, 10, False)
property_layer = PropertyLayer("elevation", 10, 10, default_value=0)
grid.add_property_layer(property_layer)

# Set a specific cell's value
grid.properties["elevation"].set_cell((5, 5), 100)

# Set multiple cells' values based on a condition
condition = lambda x: x < 50  # Condition to check each cell's value
grid.properties["elevation"].set_cells(60, condition)
```
Modifying Cell Values Using Operations.
Note: If performance is important, NumPy [Universal functions](https://numpy.org/doc/stable/reference/ufuncs.html) (ufunc) are often faster.
```python
# Modify a single cell's value with a lambda function
grid.properties["elevation"].modify_cell((3, 3), lambda x: x + 20)

# Modify multiple cells' values with a NumPy ufunc
grid.properties["elevation"].modify_cells(np.multiply, 2)

```
Conditional Operations Based on Cell Properties
```python
# Define a condition function
high_pollution = lambda x: x > 3.0

# Select cells that meet the condition
high_pollution_cells = grid.properties['pollution'].select_cells(condition=high_pollution)

# Reduce pollution in highly polluted cells using a lambda function
reduce_pollution = lambda x: x - 2.0 if x > 3.0 else x
grid.properties['pollution'].modify_cells(operation=reduce_pollution)
```
Aggregating Property Values
We can perform an aggregate operation, such as calculating the sum or mean of all values in the property layer.
```python
# Calculate the total pollution level
total_pollution = grid.properties['pollution'].aggregate_property(np.sum)

# Calculate the average pollution level
average_pollution = grid.properties['pollution'].aggregate_property(np.mean)
```
Complex Cell Selection Based on Condition
We can select cells that meet a complex condition, such as finding cells within a specific range of values.
```python
# Select cells where the value is between 5 and 15
selected_cells = layer.select_cells(lambda x: (x > 5) & (x < 15))

# Print the coordinates of the selected cells
print("Selected Cells:", selected_cells)
```

##### Moving an Agent to a Cell with the Highest Value
In this example, we'll move an agent to the cell with the highest value in a specified property layer.
```Python
from mesa import Agent, Model
from mesa.space import SingleGrid, PropertyLayer
import numpy as np

class MyAgent(Agent):
    def move_to_highest(self):
        # Move the agent to the cell with the highest value in 'temperature' layer
        target_cells = self.model.grid.select_cells(extreme_values={"temperature": "highest"})
        self.model.grid.move_agent_to_one_of(self, target_cells)

class MyModel(Model):
    def __init__(self):
        self.grid = SingleGrid(10, 10, False)
        temperature_layer = PropertyLayer("temperature", 10, 10, default_value=20, dtype=int)
        self.grid.add_property_layer(temperature_layer)

        # Randomly increase temperature in some cells
        for _ in range(30):
            x, y = self.random.randint(0, 9), self.random.randint(0, 9)
            temperature_layer.set_cell((x, y), self.random.randint(20, 40))

        # Add an agent
        self.agent = MyAgent(0, self)
        self.grid.place_agent(self.agent, (5, 5))

# Create model and move agent
model = MyModel()
model.agent.move_to_highest()

print(f"Agent moved to: {model.agent.pos}")
```
##### Example 2: Moving an Agent Based on Multiple Property Conditions
Here, we move an agent to a random cell that satisfies multiple conditions based on different property layers.

```Python
from mesa import Agent, Model
from mesa.space import SingleGrid, PropertyLayer
import numpy as np

class MyAgent(Agent):
    def move_to_optimal_cell(self):
        conditions = {
            "temperature": lambda x: x < 30,
            "humidity": lambda x: x > 50
        }
        target_cells = self.model.grid.select_cells(conditions)
        if target_cells:
            self.model.grid.move_agent_to_one_of(self, target_cells)

class MyModel(Model):
    def __init__(self):
        self.grid = SingleGrid(10, 10, False)
        self.grid.add_property_layer(PropertyLayer("temperature", 10, 10, 20, dtype=int))
        self.grid.add_property_layer(PropertyLayer("humidity", 10, 10, 60.0, dtype=float))

        # Add an agent
        self.agent = MyAgent(0, self)
        self.grid.place_agent(self.agent, (5, 5))

# Create model and move agent
model = MyModel()
model.agent.move_to_optimal_cell()

print(f"Agent moved to: {model.agent.pos}")
```
### Scenario: Forest Fire Simulation

In this simulation, we have a forest with varying levels of temperature, humidity, and elevation. Agents represent firefighters trying to contain a forest fire. The goal is for the firefighters to move towards areas with the highest temperature (indicative of fire) but also consider safety constraints such as avoiding extremely low humidity areas (high fire risk), inaccessible high elevation zones or extremely high temperature zones. They also have a maximum movement radius of 3 (using the neighbor mask) and won't move to a cell where already a firefighter is (using the empty mask).

```python
import numpy as np

from mesa import Agent, Model
from mesa.space import PropertyLayer, SingleGrid


class FirefighterAgent(Agent):
    def step(self):
        current_location = self.pos

        # Move to the hottest eligible cell
        self.move_to_hottest_eligible_cell()

        current_temperature = self.model.grid.properties["temperature"].data[self.pos]

        # Decrease the temperature (in Kelvin) by 20% at the agent's current cell
        self.reduce_temperature()

        print(
            f"Firefighter {self.unique_id} moved from {current_location} to {self.pos}, and decreased the temperature there from {current_temperature:.2f} to {self.model.grid.properties['temperature'].data[self.pos]:.2f}"
        )

    def move_to_hottest_eligible_cell(self):
        # Define additional conditions
        conditions = {
            "humidity": lambda x: x > 30,  # Humidity in %
            "elevation": lambda x: x < 75,  # Elevation in meters
            "temperature": lambda x: x < 800,  # Temperature in Kelvin
        }

        # Create masks for the neighborhood
        neighborhood_mask = self.model.grid.get_neighborhood_mask(
            self.pos, moore=True, include_center=False, radius=3
        )

        # Select target cells based on conditions, extreme values, and masks
        target_cells = self.model.grid.select_cells(
            conditions=conditions,
            extreme_values={"temperature": "highest"},
            masks=neighborhood_mask,
            only_empty=True,
        )
        # Move the agent to one of the selected cells
        self.model.grid.move_agent_to_one_of(self, target_cells)

    def reduce_temperature(self):
        self.model.grid.properties["temperature"].modify_cell(
            self.pos, np.multiply, 0.8
        )


class ForestFireModel(Model):
        super().__init__()
        self.num_agents = N
        self.grid = SingleGrid(width, height, False)

        # Create property layers
        self.grid.add_property_layer(
            PropertyLayer("temperature", width, height, 300.0, dtype=float)
        )
        self.grid.add_property_layer(
            PropertyLayer("humidity", width, height, 60, dtype=int)
        )
        self.grid.add_property_layer(
            PropertyLayer("elevation", width, height, 0, dtype=int)
        )

        # Initialize property layers with random values
        self._init_property_layers()

        # Add agents
        for i in range(self.num_agents):
            agent = FirefighterAgent(i, self)
            self.grid.move_to_empty(agent)
            self.schedule.add(agent)

    def _init_property_layers(self):
        # Access the NumPy arrays of each property layer
        temperature_data = self.grid.properties["temperature"].data
        humidity_data = self.grid.properties["humidity"].data
        elevation_data = self.grid.properties["elevation"].data

        # Assign random values to the entire arrays at once
        temperature_data[:] = np.random.randint(270, 370, temperature_data.shape)
        humidity_data[:] = np.random.randint(20, 80, humidity_data.shape)
        elevation_data[:] = np.random.randint(0, 100, elevation_data.shape)

    def step(self):
        # Increase the temperature of all cells by 5
        print(
            f"Starting step {self.schedule.steps}. Average temperature: {self.grid.properties['temperature'].aggregate_property(np.mean):3f}"
        )
        self.grid.properties["temperature"].modify_cells(np.add, 5)

        self.schedule.step()


# Create and run the model
forest_fire_model = ForestFireModel(25, 100, 100)
for _i in range(100):  # Run for 10 steps
    forest_fire_model.step()
```

### Key Features of This Example

1. **Multiple Property Layers**: Temperature, humidity, and elevation layers model different aspects of the environment.
2. **Agent Behavior**: Firefighter agents move towards high-temperature zones while avoiding low humidity and high elevation areas, simulating strategic movement in a forest fire scenario.
3. **Dynamic Environment**: The model demonstrates how agents can interact with and respond to a multi-layered, dynamically changing environment.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -24,6 +24,7 @@ import collections
 import inspect
 import itertools
 import math
+import warnings
 from numbers import Real
 from typing import (
     Any,
@@ -587,6 +588,8 @@ class PropertyLayer:
         aggregate_property(operation): Performs an aggregate operation over all cells.
     """
 
+    agentset_experimental_warning_given = False
+
     def __init__(
         self, name: str, width: int, height: int, default_value, dtype=np.float32
     ):
@@ -631,6 +634,15 @@ class PropertyLayer:
             )
 
         self.data = np.full((width, height), default_value, dtype=dtype)
+
+        if not self.__class__.agentset_experimental_warning_given:
+            warnings.warn(
+                "The new PropertyLayer and _PropertyGrid classes experimental. It may be changed or removed in any and all future releases, including patch releases.\n"
+                "We would love to hear what you think about this new feature. If you have any thoughts, share them with us here: https://github.com/projectmesa/mesa/discussions/1932",
+                FutureWarning,
+                stacklevel=2,
+            )
+            self.__class__.agentset_experimental_warning_given = True
 
     def set_cell(self, position: Coordinate, value):
         """
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -767,10 +767,6 @@ class _PropertyGrid(_Grid):
             optionally with a mask, returning either a list of coordinates or a mask.
         select_extreme_value_cells(property_name, mode, mask, return_list): Selects cells with extreme values of a property,
             optionally with a mask, returning either a list of coordinates or a mask.
-        move_agent_to_cell_by_properties(agent, conditions, mask): Moves an agent to a random cell meeting specified property
-            conditions, optionally with a mask.
-        move_agent_to_extreme_value_cell(agent, property_name, mode, mask): Moves an agent to a cell with extreme value of
-            a property, optionally with a mask.
 
     Mask Usage:
         Several methods in this class accept a mask as an input, which is a NumPy ndarray of boolean values. This mask
@@ -926,33 +922,6 @@ class _PropertyGrid(_Grid):
         else:
             return combined_mask
 
-    def move_agent_to_cell_by_properties(
-        self, agent: Agent, conditions: dict, mask: np.ndarray = None
-    ) -> None:
-        """
-        Move an agent to a random cell that meets specified property conditions, optionally with a mask.
-        If no eligible cells are found, issue a warning and keep the agent in its current position.
-
-        Args:
-            agent (Agent): The agent to move.
-            conditions (dict): Conditions for selecting the cell.
-            mask (np.ndarray, optional): A boolean mask to restrict the selection.
-        """
-        eligible_cells = self.select_cells_by_properties(
-            conditions, mask, return_list=True
-        )
-
-        if not eligible_cells:
-            warn(
-                f"No eligible cells found. Agent {agent.unique_id} remains in the current position.",
-                RuntimeWarning,
-                stacklevel=2,
-            )
-            return  # Agent stays in the current position
-
-        new_pos = agent.random.choice(eligible_cells)
-        self.move_agent(agent, new_pos)
-
     def select_extreme_value_cells(
         self,
         property_name: str,
@@ -990,35 +959,6 @@ class _PropertyGrid(_Grid):
             return list(zip(*np.where(target_mask)))
         else:
             return target_mask
-
-    def move_agent_to_extreme_value_cell(
-        self, agent: Agent, property_name: str, mode: str, mask: np.ndarray = None
-    ) -> None:
-        """
-        Move an agent to a cell with the highest or lowest property value,
-        optionally with a mask.
-
-        Args:
-            agent (Agent): The agent to move.
-            property_name (str): The name of the property layer.
-            mode (str): 'highest' or 'lowest'.
-            mask (np.ndarray, optional): A boolean mask to restrict the selection.
-        """
-        target_cells = self.select_extreme_value_cells(
-            property_name, mode, mask, return_list=True
-        )
-
-        # If no eligible cells are found, issue a warning and keep the agent in its current position.
-        if len(target_cells) == 0:
-            warn(
-                f"No eligible cells found. Agent {agent.unique_id} remains in the current position.",
-                RuntimeWarning,
-                stacklevel=2,
-            )
-            return  # Agent stays in the current position
-
-        new_pos = tuple(agent.random.choice(target_cells))
-        self.move_agent(agent, new_pos)
 
 
 class SingleGrid(_PropertyGrid):
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -584,7 +584,7 @@ class PropertyLayer:
     agentset_experimental_warning_given = False
 
     def __init__(
-        self, name: str, width: int, height: int, default_value, dtype=np.float32
+        self, name: str, width: int, height: int, default_value, dtype=np.float64
     ):
         """
         Initializes a new PropertyLayer instance.
@@ -595,7 +595,7 @@ class PropertyLayer:
             height (int): The height of the grid (number of rows). Must be a positive integer.
             default_value: The default value to initialize each cell in the grid. Should ideally
                            be of the same type as specified by the dtype parameter.
-            dtype (data-type, optional): The desired data-type for the grid's elements. Default is np.float32.
+            dtype (data-type, optional): The desired data-type for the grid's elements. Default is np.float64.
 
         Raises:
             ValueError: If width or height is not a positive integer.
@@ -603,7 +603,7 @@ class PropertyLayer:
         Notes:
             A UserWarning is raised if the default_value is not of a type compatible with dtype.
             The dtype parameter can accept both Python data types (like bool, int or float) and NumPy data types
-            (like np.int16 or np.float32). Using NumPy data types is recommended (except for bool) for better control
+            (like np.int64 or np.float64). Using NumPy data types is recommended (except for bool) for better control
             over the precision and efficiency of data storage and computations, especially in cases of large data
             volumes or specialized numerical operations.
         """
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -12,9 +12,6 @@ ContinuousSpace: a two-dimensional space where each agent has an arbitrary
                  position of `float`'s.
 NetworkGrid: a network where each node contains zero or more agents.
 """
-# Instruction for PyLint to suppress variable name errors, since we have a
-# good reason to use one-character variable names for x and y.
-# pylint: disable=invalid-name
 
 # Mypy; for the `|` operator purpose
 # Remove this __future__ import once the oldest supported Python is 3.10
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -21,6 +21,7 @@ NetworkGrid: a network where each node contains zero or more agents.
 from __future__ import annotations
 
 import collections
+import inspect
 import itertools
 import math
 from numbers import Real
@@ -548,7 +549,153 @@ class _Grid:
         return len(self.empties) > 0
 
 
-class SingleGrid(_Grid):
+def is_lambda_function(function):
+    """Check if a function is a lambda function."""
+    return (
+        inspect.isfunction(function)
+        and len(inspect.signature(function).parameters) == 1
+    )
+
+
+class PropertyLayer:
+    def __init__(
+        self, name: str, width: int, height: int, default_value, dtype=np.float32
+    ):
+        self.name = name
+        self.width = width
+        self.height = height
+        self.data = np.full((width, height), default_value, dtype=dtype)
+
+    def set_cell(self, position: Coordinate, value):
+        """
+        Update a single cell's value in-place.
+        """
+        self.data[position] = value
+
+    def set_cells(self, value, condition=None):
+        """
+        Perform a batch update either on the entire grid or conditionally, in-place.
+
+        Args:
+            value: The value to be used for the update.
+            condition: (Optional) A callable that returns a boolean array when applied to the data.
+        """
+        if condition is None:
+            np.copyto(self.data, value)  # In-place update
+        else:
+            # Ensure condition is a boolean array of the same shape as self.data
+            if (
+                not isinstance(condition, np.ndarray)
+                or condition.shape != self.data.shape
+            ):
+                raise ValueError(
+                    "Condition must be a NumPy array with the same shape as the grid."
+                )
+            np.copyto(self.data, value, where=condition)  # Conditional in-place update
+
+    def modify_cell(self, position: Coordinate, operation, value=None):
+        """
+        Modify a single cell using an operation, which can be a lambda function or a NumPy ufunc.
+        If a NumPy ufunc is used, an additional value should be provided.
+
+        Args:
+            position: The grid coordinates of the cell to modify.
+            operation: A function to apply. Can be a lambda function or a NumPy ufunc.
+            value: The value to be used if the operation is a NumPy ufunc. Ignored for lambda functions.
+        """
+        current_value = self.data[position]
+
+        # Determine if the operation is a lambda function or a NumPy ufunc
+        if is_lambda_function(operation):
+            # Lambda function case
+            self.data[position] = operation(current_value)
+        elif value is not None:
+            # NumPy ufunc case
+            self.data[position] = operation(current_value, value)
+        else:
+            raise ValueError("Invalid operation or missing value for NumPy ufunc.")
+
+    def modify_cells(self, operation, value=None, condition_function=None):
+        """
+        Modify cells using an operation, which can be a lambda function or a NumPy ufunc.
+        If a NumPy ufunc is used, an additional value should be provided.
+
+        Args:
+            operation: A function to apply. Can be a lambda function or a NumPy ufunc.
+            value: The value to be used if the operation is a NumPy ufunc. Ignored for lambda functions.
+            condition_function: (Optional) A callable that returns a boolean array when applied to the data.
+        """
+        if condition_function is not None:
+            condition_array = np.vectorize(condition_function)(self.data)
+        else:
+            condition_array = np.ones_like(self.data, dtype=bool)  # All cells
+
+        # Check if the operation is a lambda function or a NumPy ufunc
+        if is_lambda_function(operation):
+            # Lambda function case
+            modified_data = np.vectorize(operation)(self.data)
+        elif value is not None:
+            # NumPy ufunc case
+            modified_data = operation(self.data, value)
+        else:
+            raise ValueError("Invalid operation or missing value for NumPy ufunc.")
+
+        self.data = np.where(condition_array, modified_data, self.data)
+
+    def select_cells(self, condition, return_list=True):
+        """
+        Find cells that meet a specified condition using NumPy's boolean indexing, in-place.
+
+        Args:
+            condition: A callable that returns a boolean array when applied to the data.
+            return_list: (Optional) If True, return a list of (x, y) tuples. Otherwise, return a boolean array.
+
+        Returns:
+            A list of (x, y) tuples or a boolean array.
+        """
+        condition_array = condition(self.data)
+        if return_list:
+            return list(zip(*np.where(condition_array)))
+        else:
+            return condition_array
+
+    def aggregate_property(self, operation):
+        """Perform an aggregate operation (e.g., sum, mean) on a property across all cells.
+
+        Args:
+            operation: A function to apply. Can be a lambda function or a NumPy ufunc.
+        """
+
+        # Check if the operation is a lambda function or a NumPy ufunc
+        if is_lambda_function(operation):
+            # Lambda function case
+            return operation(self.data)
+        else:
+            # NumPy ufunc case
+            return operation(self.data)
+
+
+class _PropertyGrid(_Grid):
+    def __init__(self, width: int, height: int, torus: bool):
+        super().__init__(width, height, torus)
+        self.properties = {}
+
+    # Add and remove properties to the grid
+    def add_property_layer(self, property_layer: PropertyLayer):
+        self.properties[property_layer.name] = property_layer
+
+    def remove_property_layer(self, property_name: str):
+        if property_name not in self.properties:
+            raise ValueError(f"Property layer {property_name} does not exist.")
+        del self.properties[property_name]
+
+    # TODO:
+    # - Select cells conditionally based on multiple properties
+    # - Move random cells conditionally based on multiple properties
+    # - Move to cell with highest/lowest/closest property value
+
+
+class SingleGrid(_PropertyGrid):
     """Rectangular grid where each cell contains exactly at most one agent.
 
     Grid cells are indexed by [x, y], where [0, 0] is assumed to be the
@@ -591,7 +738,7 @@ class SingleGrid(_Grid):
         agent.pos = None
 
 
-class MultiGrid(_Grid):
+class MultiGrid(_PropertyGrid):
     """Rectangular grid where each cell can contain more than one agent.
 
     Grid cells are indexed by [x, y], where [0, 0] is assumed to be at
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -564,6 +564,22 @@ class PropertyLayer:
         self.name = name
         self.width = width
         self.height = height
+
+        # Check that width and height are positive integers
+        if (not isinstance(width, int) or width < 1) or (
+            not isinstance(height, int) or height < 1
+        ):
+            raise ValueError(
+                f"Width and height must be positive integers, got {width} and {height}."
+            )
+        # Check if the dtype is suitable for the data
+        if not isinstance(default_value, dtype):
+            warn(
+                f"Default value {default_value} ({type(default_value).__name__}) might not be best suitable with dtype={dtype.__name__}.",
+                UserWarning,
+                stacklevel=2,
+            )
+
         self.data = np.full((width, height), default_value, dtype=dtype)
 
     def set_cell(self, position: Coordinate, value):
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -710,10 +710,73 @@ class _PropertyGrid(_Grid):
             raise ValueError(f"Property layer {property_name} does not exist.")
         del self.properties[property_name]
 
-    # TODO:
-    # - Select cells conditionally based on multiple properties
-    # - Move random cells conditionally based on multiple properties
-    # - Move to cell with highest/lowest/closest property value
+    def select_cells_multi_properties(self, conditions: dict) -> List[Coordinate]:
+        """
+        Select cells based on multiple property conditions using NumPy.
+
+        Args:
+            conditions (dict): A dictionary where keys are property names and values are
+                               callables that take a single argument (the property value)
+                               and return a boolean.
+
+        Returns:
+            List[Coordinate]: A list of coordinates where the conditions are satisfied.
+        """
+        # Start with a mask of all True values
+        combined_mask = np.ones((self.width, self.height), dtype=bool)
+
+        for prop_name, condition in conditions.items():
+            prop_layer = self.properties[prop_name].data
+            # Apply the condition to the property layer
+            prop_mask = condition(prop_layer)
+            # Combine with the existing mask using logical AND
+            combined_mask = np.logical_and(combined_mask, prop_mask)
+
+        # Extract coordinates from the combined mask
+        selected_cells = list(zip(*np.where(combined_mask)))
+        return selected_cells
+
+    def move_agent_to_random_cell(self, agent: Agent, conditions: dict) -> None:
+        """
+        Move an agent to a random cell that meets specified property conditions.
+        If no eligible cells are found, issue a warning and keep the agent in its current position.
+
+        Args:
+            agent (Agent): The agent to move.
+            conditions (dict): Conditions for selecting the cell.
+        """
+        eligible_cells = self.select_cells_multi_properties(conditions)
+        if not eligible_cells:
+            warn(f"No eligible cells found. Agent {agent.unique_id} remains in the current position.", RuntimeWarning)
+            return  # Agent stays in the current position
+
+        # Randomly choose one of the eligible cells and move the agent
+        new_pos = agent.random.choice(eligible_cells)
+        self.move_agent(agent, new_pos)
+
+    def move_agent_to_extreme_value_cell(self, agent: Agent, property_name: str, mode: str) -> None:
+        """
+        Move an agent to a cell with the highest, lowest, or closest property value.
+
+        Args:
+            agent (Agent): The agent to move.
+            property_name (str): The name of the property layer.
+            mode (str): 'highest', 'lowest', or 'closest'.
+        """
+        prop_values = self.properties[property_name].data
+        if mode == 'highest':
+            target_value = np.max(prop_values)
+        elif mode == 'lowest':
+            target_value = np.min(prop_values)
+        elif mode == 'closest':
+            agent_value = prop_values[agent.pos]
+            target_value = prop_values[np.abs(prop_values - agent_value).argmin()]
+        else:
+            raise ValueError(f"Invalid mode {mode}. Choose from 'highest', 'lowest', or 'closest'.")
+
+        target_cells = list(zip(*np.where(prop_values == target_value)))
+        new_pos = agent.random.choice(target_cells)
+        self.move_agent(agent, new_pos)
 
 
 class SingleGrid(_PropertyGrid):
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -821,7 +821,6 @@ class _PropertyGrid(_Grid):
 
         # Initialize an empty mask as a boolean NumPy array
         self._empty_mask = np.ones((self.width, self.height), dtype=bool)
-        print(self._empty_mask)
 
         # Handle both single PropertyLayer instance and list of PropertyLayer instances
         if property_layers:
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -710,17 +710,50 @@ class _PropertyGrid(_Grid):
             raise ValueError(f"Property layer {property_name} does not exist.")
         del self.properties[property_name]
 
-    def select_cells_multi_properties(self, conditions: dict) -> List[Coordinate]:
+    def get_neighborhood_mask(
+        self, pos: Coordinate, moore: bool, include_center: bool, radius: int
+    ) -> np.ndarray:
         """
-        Select cells based on multiple property conditions using NumPy.
+        Generate a boolean mask representing the neighborhood.
+
+        Args:
+            pos (Coordinate): Center of the neighborhood.
+            moore (bool): True for Moore neighborhood, False for Von Neumann.
+            include_center (bool): Include the central cell in the neighborhood.
+            radius (int): The radius of the neighborhood.
+
+        Returns:
+            np.ndarray: A boolean mask representing the neighborhood.
+        """
+        neighborhood = self.get_neighborhood(pos, moore, include_center, radius)
+        mask = np.zeros((self.width, self.height), dtype=bool)
+
+        # Convert the neighborhood list to a NumPy array and use advanced indexing
+        coords = np.array(neighborhood)
+        mask[coords[:, 0], coords[:, 1]] = True
+        return mask
+
+    def select_cells_multi_properties(
+        self,
+        conditions: dict,
+        only_neighborhood: bool = False,
+        pos: None | Coordinate = None,
+        moore: bool = True,
+        include_center: bool = False,
+        radius: int = 1,
+    ) -> list[Coordinate]:
+        """
+        Select cells based on multiple property conditions using NumPy, optionally within a neighborhood.
 
         Args:
             conditions (dict): A dictionary where keys are property names and values are
                                callables that take a single argument (the property value)
                                and return a boolean.
+            only_neighborhood (bool): If True, restrict selection to the neighborhood.
+            pos, moore, include_center, radius: Optional neighborhood parameters.
 
         Returns:
-            List[Coordinate]: A list of coordinates where the conditions are satisfied.
+            List[Coordinate]: Coordinates where conditions are satisfied.
         """
         # Start with a mask of all True values
         combined_mask = np.ones((self.width, self.height), dtype=bool)
@@ -732,50 +765,107 @@ class _PropertyGrid(_Grid):
             # Combine with the existing mask using logical AND
             combined_mask = np.logical_and(combined_mask, prop_mask)
 
+        if only_neighborhood and pos is not None:
+            neighborhood_mask = self.get_neighborhood_mask(
+                pos, moore, include_center, radius
+            )
+            combined_mask = np.logical_and(combined_mask, neighborhood_mask)
+
         # Extract coordinates from the combined mask
         selected_cells = list(zip(*np.where(combined_mask)))
         return selected_cells
 
-    def move_agent_to_random_cell(self, agent: Agent, conditions: dict) -> None:
+    def move_agent_to_random_cell(
+        self,
+        agent: Agent,
+        conditions: dict,
+        only_neighborhood: bool = False,
+        moore: bool = True,
+        include_center: bool = False,
+        radius: int = 1,
+    ) -> None:
         """
-        Move an agent to a random cell that meets specified property conditions.
+        Move an agent to a random cell that meets specified property conditions, optionally within a neighborhood.
         If no eligible cells are found, issue a warning and keep the agent in its current position.
 
         Args:
             agent (Agent): The agent to move.
             conditions (dict): Conditions for selecting the cell.
+            only_neighborhood, moore, include_center, radius: Optional neighborhood parameters.
         """
-        eligible_cells = self.select_cells_multi_properties(conditions)
+        pos = agent.pos if only_neighborhood else None
+        eligible_cells = self.select_cells_multi_properties(
+            conditions,
+            only_neighborhood,
+            pos,
+            moore,
+            include_center,
+            radius,
+        )
         if not eligible_cells:
-            warn(f"No eligible cells found. Agent {agent.unique_id} remains in the current position.", RuntimeWarning)
+            warn(
+                f"No eligible cells found. Agent {agent.unique_id} remains in the current position.",
+                RuntimeWarning,
+                stacklevel=2,
+            )
             return  # Agent stays in the current position
 
         # Randomly choose one of the eligible cells and move the agent
         new_pos = agent.random.choice(eligible_cells)
         self.move_agent(agent, new_pos)
 
-    def move_agent_to_extreme_value_cell(self, agent: Agent, property_name: str, mode: str) -> None:
+    def move_agent_to_extreme_value_cell(
+        self,
+        agent: Agent,
+        property_name: str,
+        mode: str,
+        only_neighborhood: bool = False,
+        moore: bool = True,
+        include_center: bool = False,
+        radius: int = 1,
+    ) -> None:
         """
-        Move an agent to a cell with the highest, lowest, or closest property value.
+        Move an agent to a cell with the highest, lowest, or closest property value,
+        optionally within a neighborhood.
 
         Args:
             agent (Agent): The agent to move.
             property_name (str): The name of the property layer.
             mode (str): 'highest', 'lowest', or 'closest'.
+            only_neighborhood, moore, include_center, radius: Optional neighborhood parameters.
         """
+        pos = agent.pos if only_neighborhood else None
         prop_values = self.properties[property_name].data
-        if mode == 'highest':
-            target_value = np.max(prop_values)
-        elif mode == 'lowest':
-            target_value = np.min(prop_values)
-        elif mode == 'closest':
-            agent_value = prop_values[agent.pos]
-            target_value = prop_values[np.abs(prop_values - agent_value).argmin()]
-        else:
-            raise ValueError(f"Invalid mode {mode}. Choose from 'highest', 'lowest', or 'closest'.")
 
-        target_cells = list(zip(*np.where(prop_values == target_value)))
-        new_pos = agent.random.choice(target_cells)
+        if pos is not None:
+            # Mask out cells outside the neighborhood.
+            neighborhood_mask = self.get_neighborhood_mask(
+                pos, moore, include_center, radius
+            )
+            # Use NaN for out-of-neighborhood cells
+            masked_prop_values = np.where(neighborhood_mask, prop_values, np.nan)
+        else:
+            masked_prop_values = prop_values
+
+        # Find the target value
+        if mode == "highest":
+            target_value = np.nanmax(masked_prop_values)
+        elif mode == "lowest":
+            target_value = np.nanmin(masked_prop_values)
+        elif mode == "closest":
+            agent_value = prop_values[agent.pos]
+            target_value = masked_prop_values[
+                np.nanargmin(np.abs(masked_prop_values - agent_value))
+            ]
+        else:
+            raise ValueError(
+                f"Invalid mode {mode}. Choose from 'highest', 'lowest', or 'closest'."
+            )
+
+        # Find the coordinates of the target value(s)
+        target_cells = np.column_stack(np.where(masked_prop_values == target_value))
+        # If there are multiple target cells, randomly choose one
+        new_pos = tuple(agent.random.choice(target_cells, axis=0))
         self.move_agent(agent, new_pos)
 
 
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -747,11 +747,11 @@ class _PropertyGrid(_Grid):
         remove_property_layer(property_name): Removes a property layer from the grid by its name.
         get_empty_mask(): Generates a boolean mask indicating empty cells on the grid.
         get_neighborhood_mask(pos, moore, include_center, radius): Generates a boolean mask of the neighborhood.
-        select_cells_multi_properties(conditions, mask, return_list): Selects cells based on multiple property conditions,
+        select_cells_by_properties(conditions, mask, return_list): Selects cells based on multiple property conditions,
             optionally with a mask, returning either a list of coordinates or a mask.
         select_extreme_value_cells(property_name, mode, mask, return_list): Selects cells with extreme values of a property,
             optionally with a mask, returning either a list of coordinates or a mask.
-        move_agent_to_random_cell(agent, conditions, mask): Moves an agent to a random cell meeting specified property
+        move_agent_to_cell_by_properties(agent, conditions, mask): Moves an agent to a random cell meeting specified property
             conditions, optionally with a mask.
         move_agent_to_extreme_value_cell(agent, property_name, mode, mask): Moves an agent to a cell with extreme value of
             a property, optionally with a mask.
@@ -877,7 +877,7 @@ class _PropertyGrid(_Grid):
         mask[coords[:, 0], coords[:, 1]] = True
         return mask
 
-    def select_cells_multi_properties(
+    def select_cells_by_properties(
         self, conditions: dict, mask: np.ndarray = None, return_list: bool = True
     ) -> list[Coordinate] | np.ndarray:
         """
@@ -910,7 +910,7 @@ class _PropertyGrid(_Grid):
         else:
             return combined_mask
 
-    def move_agent_to_random_cell(
+    def move_agent_to_cell_by_properties(
         self, agent: Agent, conditions: dict, mask: np.ndarray = None
     ) -> None:
         """
@@ -922,7 +922,7 @@ class _PropertyGrid(_Grid):
             conditions (dict): Conditions for selecting the cell.
             mask (np.ndarray, optional): A boolean mask to restrict the selection.
         """
-        eligible_cells = self.select_cells_multi_properties(
+        eligible_cells = self.select_cells_by_properties(
             conditions, mask, return_list=True
         )
 
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -915,22 +915,24 @@ class _PropertyGrid(_Grid):
         new_pos = agent.random.choice(eligible_cells)
         self.move_agent(agent, new_pos)
 
-    def move_agent_to_extreme_value_cell(
-        self, agent: Agent, property_name: str, mode: str, mask: np.ndarray = None
-    ) -> None:
+    def select_extreme_value_cells(
+        self,
+        property_name: str,
+        mode: str,
+        mask: np.ndarray = None,
+        return_list: bool = True,
+    ) -> list[Coordinate] | np.ndarray:
         """
-        Move an agent to a cell with the highest or lowest property value,
-        optionally with a mask.
-
-
-        If multiple cells have the same extreme value, one of them is chosen randomly.
-        If no eligible cells are found, issue a warning and keep the agent in its current position.
+        Select cells with the highest or lowest property value, optionally with a mask.
 
         Args:
-            agent (Agent): The agent to move.
             property_name (str): The name of the property layer.
             mode (str): 'highest' or 'lowest'.
             mask (np.ndarray, optional): A boolean mask to restrict the selection.
+            return_list (bool, optional): If True, return a list of coordinates, otherwise return an ndarray.
+
+        Returns:
+            list[Coordinate] or np.ndarray: Coordinates of cells with the extreme property value.
         """
         prop_values = self.properties[property_name].data
         if mask is not None:
@@ -938,7 +940,6 @@ class _PropertyGrid(_Grid):
         else:
             masked_prop_values = prop_values
 
-        # Find coordinates of target cells directly
         if mode == "highest":
             target_cells = np.column_stack(
                 np.where(masked_prop_values == np.nanmax(masked_prop_values))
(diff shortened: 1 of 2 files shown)


EwoutH:
This PR is now feature complete! Please take a proper look at the code, it's ready for review.

Next up:
- [ ] Tests
- [ ] Updating PR message
- [ ] Adding more examples

mesa/space.py
```diff
@@ -485,7 +486,444 @@ def exists_empty_cells(self) -> bool:
...
+
+        # Convert the neighborhood list to a NumPy array and use advanced indexing
+        coords = np.array(neighborhood)
+        mask[coords[:, 0], coords[:, 1]] = True
+        return mask
+
+    def select_cells_multi_properties(
+        self,
+        conditions: dict,
+        only_neighborhood: bool = False,
```
Corvince, line 794: Maybe this could be simplified to a an optional neighborhood argument that precalculated the neighborhood? And _get_neighborhood_mask only masks a neighborhood. That way we don't need all the extra arguments for the neighborhood and this could also be applied for other spaces, where neighborhoods can be defined differently

mesa/space.py
```diff
@@ -485,7 +486,444 @@ def exists_empty_cells(self) -> bool:
...
+
+        Args:
+            position: The grid coordinates of the cell to modify.
+            operation: A function to apply. Can be a lambda function or a NumPy ufunc.
+            value: The value to be used if the operation is a NumPy ufunc. Ignored for lambda functions.
+        """
+        current_value = self.data[position]
+
+        # Determine if the operation is a lambda function or a NumPy ufunc
+        if is_lambda_function(operation):
```
Corvince, line 602: does it really have to be a lambda function? Wouldn't a normal function also work?

mesa/space.py
```diff
@@ -485,7 +486,444 @@ def exists_empty_cells(self) -> bool:
...
+        else:
+            # Ensure condition is a boolean array of the same shape as self.data
+            if (
+                not isinstance(condition, np.ndarray)
+                or condition.shape != self.data.shape
+            ):
+                raise ValueError(
+                    "Condition must be a NumPy array with the same shape as the grid."
+                )
+            np.copyto(self.data, value, where=condition)  # Conditional in-place update
```
Corvince, line 587: cant we update the values in-place, why the copy?

mesa/space.py
```diff
@@ -485,7 +486,444 @@ def exists_empty_cells(self) -> bool:
...
+        Args:
+            value: The value to be used for the update.
+            condition: (Optional) A callable that returns a boolean array when applied to the data.
+        """
+        if condition is None:
+            np.copyto(self.data, value)  # In-place update
+        else:
+            # Ensure condition is a boolean array of the same shape as self.data
+            if (
+                not isinstance(condition, np.ndarray)
```
Corvince, line 581: Wait, is condition a callable that returns a boolean array or already a boolean array?

mesa/space.py
```diff
@@ -485,7 +486,444 @@ def exists_empty_cells(self) -> bool:
...
+            operation: A function to apply. Can be a lambda function or a NumPy ufunc.
+        """
+
+        # Check if the operation is a lambda function or a NumPy ufunc
+        if is_lambda_function(operation):
+            # Lambda function case
+            return operation(self.data)
+        else:
+            # NumPy ufunc case
+            return operation(self.data)
```
Corvince, line 668: both branches do the same, so I think we dont need the branching?

mesa/space.py
```diff
@@ -485,7 +486,439 @@ def exists_empty_cells(self) -> bool:
...
+            combined_mask = np.logical_and(combined_mask, neighborhood_mask)
+
+        # Extract coordinates from the combined mask
+        selected_cells = list(zip(*np.where(combined_mask)))
+        return selected_cells
+
+    def move_agent_to_random_cell(
+        self,
+        agent: Agent,
+        conditions: dict,
```
Corvince, line 836: cool idea to randomly move, but only to a specified neighborhood.

Corvince (commented):
Super excited for this! Obviously still some rough edges, but I think this has very high potential.

I left some comments. One thing that got me thinking. You often define methods that potentially only affect a neighborhood. For this you always have the optional arguments only_neighborhoods, moore, include_center and radius. This somewhat predetermines what a neighborhood can be. But maybe we can define "neighborhood" just as a list of coordinates. Then we can define neighborhoods also based on conditions and only pass in the neighborhoods to these functions. Maybe that would be more general?

I also think some method names are not optimal. But I think to fully judge that we would need some toy model where we can "read" how models are written. 

Anyway, again, great work!

EwoutH:
All docs and tests are added. That should reviewing a bit easier.

 - I'm quite certain I went for the right approach with PropertyLayer. I like the code base and the interface, and think it's quite performant and inuative
 - I also like the basic construct of adding one or more PropertyLayers to a grid. I think using a private _PropertyGrid of which SingleGrid and MultiGrid inherit is the right approach.
 - I'm not fully content with the build in methods of _PropertyGrid, especially the last three:
   - select_cells_multi_properties
   - move_agent_to_random_cell
   - move_agent_to_extreme_value_cell
   They work well but everything about them is just a little complicated, and maybe too specific. Any feedback on those will be extra appreciated.

While this PR is fully implemented and quite specific, feel also free to talk about the big picture stuff. I'm proposing a major conceptual change to the Mesa space stack, and thus discussion on all levels is appreciated.

<hr>

@rht Ruff doesn't allow me to specify a condition as a lambda function:

```python
condition = lambda x: x == 0
```
```bash
tests/test_space.py:557:9: E731 Do not assign a `lambda` expression, use a `def`
Found 7 errors.
No fixes available (7 hidden fixes can be enabled with the `--unsafe-fixes` option).
```
Direct insertion is okay, so without using the condition variable in between.

Being able to use lambda function to update the property layer is fundamental to this PR, and I like the current explicit `condition = lambda` approach. How do you think we should approach this.

EwoutH:
> One thing that got me thinking. You often define methods that potentially only affect a neighborhood. For this you always have the optional arguments only_neighborhoods, moore, include_center and radius. This somewhat predetermines what a neighborhood can be. But maybe we can define "neighborhood" just as a list of coordinates. Then we can define neighborhoods also based on conditions and only pass in the neighborhoods to these functions. Maybe that would be more general?

Yes, I like this. Just pass a sub selection of cells, which can be a neighbourhood but also can be something else. That would be extremely flexible, you can define rooms, areas, basically anything you like.

mesa/space.py
EwoutH: Nice catch, fixed in 7b9310c.

mesa/space.py
EwoutH: Thanks! I think generalizing that is indeed a good idea, I replied below.

mesa/space.py
EwoutH: Yeah it probably would. No idea why I limited to lambda's.

NumPy functions are the most performant however, so when possible it should be encouraged to use those.

mesa/space.py
EwoutH: Agreed, will think about how to make this modular.

If you have any suggestions, let me know!

wang-boyu:
Great work!

I'm wondering what type of indices is used for cells. In grid space the coordinate system is in `(x, y)` with origin at the lower left corner, which is typical Cartesian coordinates (if I remember correctly, but I might be wrong about this). However in matrices (or numpy arrays) we often use the `(row, column)` format, with an origin at the upper left corner. In mesa-geo I ended up using both: https://github.com/projectmesa/mesa-geo/blob/95cdcebb7490b0e832308ce8bb6161915c5de875/mesa_geo/raster_layers.py#L177C13-L180

Another question is that it seems to be possible to have property layers with different dimensions than that of the grid space. If this is the case, how do we align them? This also links to the coordinate system mentioned above. In mesa-geo, raster layers can have different dimensions and resolutions, and everything was taken care of through a `crs` setting in `GeoSpace`. But I think this is gis specific and may not apply here.

As an example, if you have a 10x10 grid, and a 20x20 property layer, where would you put it on the grid?

If, alternatively, you would prefer to make sure that property layers always have the same dimension as the grid, perhaps we could simplify the API to something like this?

```python
# Initialize a SingleGrid
grid = SingleGrid(width=10, height=10, torus=False)

# Create a PropertyLayer for a specific property (e.g., pollution level)
# and add the property layer to the grid
grid.add_property_layer(name='pollution', default_value=0.0)

# or
grid.add_property_layer(name='another_property', values=some_numpy_array_of_the_same_shape)
```

EwoutH:
Based on Corvince’s insight and 
- Instead of neighborhood arguments, take a mask. It can be an empty cell mask, neighborhood mask, or any mask you like.
  - Add utility method to create mask from list of cells
  - Add utility methods to create neighborhood and empty cell masks
- Maybe split move_agent_to_extreme_value_cell into a select target and move to target method. That way you can mix and match them.

EwoutH:
> I'm wondering what type of indices is used for cells. In grid space the coordinate system is in (x, y) with origin at the lower left corner, which is typical Cartesian coordinates (if I remember correctly, but I might be wrong about this). However in matrices (or numpy arrays) we often use the (row, column) format, with an origin at the upper left corner.

Not even thought of this. 😅

> Another question is that it seems to be possible to have property layers with different dimensions than that of the grid space. If this is the case, how do we align them?

Was also thinking about this for ContinuousSpace. I would say out of scope for this PR, but in general, it would be nice to support this.

> If, alternatively, you would prefer to make sure that property layers always have the same dimension as the grid, perhaps we could simplify the API to something like this?

Personally I wouldn’t go this way, because now property layers are completely independent. It would be nice to keep them that way.

EwoutH:
Quite productive morning so far!

- In d731e2f I removed the neighborhood stuff from the PropertyGrid methods. They now take a mask argument
- In 478506b I allowed returning a mask or a list, for added flexibility
- In 5857c4f I split move_agent_to_extreme_value_cell into two methods: A select and a move method
- In c2f1b2d I wrote a utility function to create an empty mask
- In ca72bee I renamed two methods. I hope they're clearer now.
- In the other commits, I added docstring, updated the tests and fixed the remaining issues.

<hr>

@wang-boyu I added a test in 68992a0 to check the coordinate systems between agents and layers. It looks like they match. Can you check if I tested correctly and haven't forgotten an edge case?

<hr>

@rth I would really like to know what you think about Ruff [not allowing assigned lambda functions](https://github.com/projectmesa/mesa/pull/1898#issuecomment-1839508312).

<hr>

Code, test and docs wise, this PR is done. I will now work on some examples to show functionality. I would invite everyone to do that with me and play around with it a bit. This branch can be installed with:

```
pip install -U -e git+https://github.com/EwoutH/mesa@property_layer#egg=mesa
```

EwoutH:
I added 3 examples to the main PR message. I think is shows (almost) all functionality. Really curious what everybody thinks!

mesa/space.py
EwoutH: Implemented! It now takes a mask with a utility method `get_neighborhood_mask()`

mesa/space.py
EwoutH: Done, you can now insert any mask. Utility methods `get_neighborhood_mask()` and `get_empty_mask` are provided.

wang-boyu:
> @wang-boyu I added a test in [68992a0](https://github.com/projectmesa/mesa/commit/68992a0f3d634b45234932a7d9db3c9b467664f6) to check the coordinate systems between agents and layers. It looks like they match. Can you check if I tested correctly and haven't forgotten an edge case?

Thanks for adding a test case for it. But it's not really what I'm trying to say.

I made a small visualization for your forest fire model:

<details>
  <summary>code</summary>

```python
import numpy as np
import solara
from mesa.experimental import JupyterViz


def space_drawer(model, agent_portrayal):
    agents_portrayal = {
        "firefighter": {"x": [], "y": [], "c": "tab:red", "marker": "o", "s": 10},
    }
    for agent in model.schedule.agents:
        if isinstance(agent, FirefighterAgent):
            agents_portrayal["firefighter"]["x"].append(agent.pos[0])
            agents_portrayal["firefighter"]["y"].append(agent.pos[1])

    fig, ax = plt.subplots()
    im = ax.imshow(model.grid.properties["temperature"].data, cmap="spring")
    # im = ax.imshow(model.grid.properties["temperature"].data, cmap="spring", origin="lower")
    fig.colorbar(im, orientation="vertical")

    ax.scatter(**agents_portrayal["firefighter"])
    ax.set_axis_off()
    solara.FigureMatplotlib(fig)


model_params = {
    "N": 1,
    "width": 4,
    "height": 4,
}

page = JupyterViz(
    ForestFireModel,
    model_params,
    measures=[],
    name="Forest Fire",
    space_drawer=space_drawer,
    play_interval=1500,
)
page  # noqa
```

</details>

For simplicity I also changed your `_init_property_layers()` function, from

```python
temperature_data[:] = np.random.randint(270, 370, temperature_data.shape)
```

to

```python
temperature_data[0, 0] = 500
```

so that the cell at position (0, 0) has a temperature of 500, with other cells having the default temperature of 300. Now the temperature property becomes

```python
model = ForestFireModel(1, 4, 4)
model.grid.properties["temperature"].data

array([[500., 300., 300., 300.],
       [300., 300., 300., 300.],
       [300., 300., 300., 300.],
       [300., 300., 300., 300.]])
```

If we plot it, it looks like

Screenshot 2023-12-06 at 10 53 01

which isn't what we want, because (0, 0) is supposed to be at the lower left corner. This is essentially what I was trying to explain earlier.

I just realized that the same problem has been dealt with in the sugarscape demo: https://github.com/projectmesa/mesa-examples/blob/7275b172aad8905584dcffc131d9bbd96992abb7/examples/sugarscape_g1mt/app.py#L36-L38. In the demo code above I have a line commented out, that specifies `origin="lower"` when plotting the temperature layer. In that case, the visualization will look correct.

However I should mention that this only makes the model _looks_ correct. The underlying numpy array still has its _upper_ left corner with the 500 temperature. If we go for this approach, we probably need to make sure that

1. we and the users will use position in (x, y) format for consistency
2. all visualizations of property layers should have `origin="lower"`
3. the underlying numpy array is not the same as what it looks like in visualizations

Sorry about all these. I hope this helps clarifying my concern.

EwoutH:
I already hate this, but adding a translation method that flips the y coordinate of every position like this:

```Python
    def _translate_coords(self, position):
        """
        Translates Mesa coordinates to NumPy array coordinates.
        """
        x, y = position
        return self.height - 1 - y, x
```
And then using that everywhere where a position is inputed:
```Python
    def set_cell(self, position, value):
        """
        Update a single cell's value in-place.
        """
        self.data[self._translate_coords(position)] = value
```
Would probably fix it right.

**Except** if users start manipulating the class directly, without using the methods...

I already hate this.

Corvince:
Your hate is very understandable. What we could potentially do is use a numpy array with dtype=object for the agents as well. Last time I checked it was marginally slower, but would allow easier consistency.

EwoutH:
> We treat the numpy array as an implementation detail that is hidden from the users. Indexing is through (x, y) coordinate everywhere. We only need to make sure that, the right way to interpret the array (e.g., when visualizing it) is to set origin="lower". I believe this is what we did in the sugarscape example.

I thought about it for a bit, and I think this is the right approach. It's basically just a convention for visualisation. In future efforts, we might want to create a visualisation function that's 

> Regarding the first option, it is also possible to make the numpy array as a Python property without a setter.

I like this. Let me work on a implementation. We basically define it like this, right?

```Python
    @property
    def data(self):
        return self._data
```
<hr>

As a side note, I really would like feedback on the API and the variable names. This is the stuff that's difficult to change after the first release. So are there any names of methods that veel unintuitive, or could be improved?

EwoutH:
To be honest, I don't know if we want data as a strict property. Doing some initializations as `np.arange(100).reshape(10, 10)`, `np.full((10, 10), 2)` is pretty powerful, and (as far as I understand) making `data` a property without a setter would disable that.

Or should we include a setter and do some checks there?

wang-boyu:
> I thought about it for a bit, and I think this is the right approach. It's basically just a convention for visualisation. In future efforts, we might want to create a visualisation function that included the up-down flip.

Yes I prefer this approach too, which is also consistent with the sugarscape example model . To recap, we assume indexing such as `data[1, 2]` means `x=1` and `y=2`. This is different from the usual way of `(row, col)` indexing, but it is what we specify in Mesa.

A side effect of this approach is the internal representation using the numpy array will be upside down. To mitigate this problem, there are again a few options which are essentially design decisions:

1. Do nothing with the setter and getter. When plotting the layer, use `origin="lower"`. This is what we did in the sugarscape model.
2. Either transform the coordinate in the setter, or flip the numpy array in the getter. In this way the array representation will be correct, and we don't need to do `origin="lower"` in plots. An example of changing the getter would be (while the setter does nothing but returning the `_data`)
   ```python
   @property
   def data(self):
       return np.flipud(self._data)  # flip the layer; not very sure whether it works as expected
   ```
   So is the user does `data[0, 0] = 500`, then the cell at the _lower_ left corner gets updated. Might come as a surprise to those who are familiar with matrix indexing.

Regarding the APIs I don't have any preferences. My only questions is whether you would like to call it `cell` or `patch` or anything else.

mesa/space.py
```diff
@@ -485,7 +486,463 @@ def exists_empty_cells(self) -> bool:
...
+        # If no eligible cells are found, issue a warning and keep the agent in its current position.
+        if len(target_cells) == 0:
+            warn(
+                f"No eligible cells found. Agent {agent.unique_id} remains in the current position.",
+                RuntimeWarning,
+                stacklevel=2,
+            )
+            return  # Agent stays in the current position
+
+        new_pos = tuple(agent.random.choice(target_cells))
```
Corvince, line 941: This might be surprising that a cell is chosen randomly

Corvince:
> Regarding the APIs I don't have any preferences. My only questions is whether you would like to call it `cell` or `patch` or anything else.

This is also something I noticed and I wonder if we should drop a concrete name altogether. I would be in favor of using more generic names for most of the functions. Anything with cell or even patch in the name implies a grid. If we later extend to 3 dimensional spaces or for networks the terminology doesn't fit anymore.

I would prefer to exchange cell with value in most cases. For example instead of `set_cell` we could have `set_value`, which fits for all kind of property layers. 

And `move_to_extreme_value_cell` could just drop the cell, which would also make it a bit shorter. 

In cases where value doesn't fit maybe we can use pos or position, which is still generic. 

Otherwise I am quite happy with this PR. Really excellent intro post, great examples.

Currently on mobile so I can't to a proper code review, but I agree that having a proper API is currently more important and implementation details can also be improved later (if there are any improvements)

EwoutH:
Thanks! Let met think a bit about the `cell` / `patch` / `value` naming discussion. 

<hr>

As a demo, I've implemented Conway's Game of Life using the PropertyLayer:

```Python
import numpy as np
from mesa import Model
from mesa.space import SingleGrid, PropertyLayer
from scipy.signal import convolve2d

class GameOfLifeModel(Model):
    def __init__(self, width, height):
        self.grid = SingleGrid(width, height, torus=True)
        # Initialize the property layer for cell states
        self.cell_layer = PropertyLayer("cells", width, height, False, dtype=bool)
        self.grid.add_property_layer(self.cell_layer)

    from scipy.signal import convolve2d

    def step(self):
        # Define a kernel for counting neighbors. The kernel has 1s around the center cell (which is 0).
        # This setup allows us to count the live neighbors of each cell when we apply convolution.
        kernel = np.array([[1, 1, 1],
                           [1, 0, 1],
                           [1, 1, 1]])

        # Count neighbors using convolution.
        # convolve2d applies the kernel to each cell of the grid, summing up the values of neighbors.
        # boundary="wrap" ensures that the grid wraps around, simulating a toroidal surface.
        neighbor_count = convolve2d(self.cell_layer.data, kernel, mode="same", boundary="wrap")

        # Apply Game of Life rules:
        # 1. A live cell with 2 or 3 live neighbors survives, otherwise it dies.
        # 2. A dead cell with exactly 3 live neighbors becomes alive.
        # These rules are implemented using logical operations on the grid.
        self.cell_layer.data = np.logical_or(
            np.logical_and(self.cell_layer.data, np.logical_or(neighbor_count == 2, neighbor_count == 3)),
            # Rule for live cells
            np.logical_and(~self.cell_layer.data, neighbor_count == 3)  # Rule for dead cells
        )

    def count_live_neighbors(self, x, y):
        neighbors = self.grid.get_neighborhood((x, y), moore=True, include_center=False, radius=1)
        return sum(self.cell_layer.data[nx, ny] for nx, ny in neighbors)
```

EwoutH:
I benchmarked it against [the GoL implementation](https://github.com/projectmesa/mesa-examples/tree/main/examples/conways_game_of_life) in the Mesa examples repo, and this implementation using a PropertyLayer is about 100x faster for the same grid size.

runtime_comparison

- [benchmark_gol.zip](https://github.com/projectmesa/mesa/files/13628343/benchmark_gol.zip)

EwoutH:
I thought a bit about it, and while I like this discussion I have to disagree on this premise:

> Anything with cell or even patch in the name implies a grid.  If we later extend to 3 dimensional spaces

I think a cell implies a contained space, which can be 2D, 3D or otherwise. It also doesn’t imply a square grid, it could be hexagonal, multipolygon or else.

But it does imply a surface of volume is made up of discrete units, and each of the units takes some space, and all space is covered by exactly one unit.

> or for networks the terminology doesn't fit anymore.

I find networks difficult to be compatible with the PropertyLayer as implemented, because they don’t represent physical space. So I don’t think we’re ever going to mix those, since the PropertyLayer inherently implies physical space.

So personally, I currently prefer `cell`. I think it will also make teaching about it easier, there is something concrete you can talk about.

EwoutH:
Initializing things randomly feels a bit weird:
```Python
class HabitatModel(Model):
    def __init__(self, n_prey, n_predators, food_probability):
        super().__init__()
        self.running = True
        self.food_layer = PropertyLayer("food", 32, 32, default_value=False, dtype=bool)
        self.grid = MultiGrid(32, 32, True, property_layers=self.food_layer)
        self.food_probability = food_probability

        condition = lambda val: np.random.random_sample(val.shape) < self.food_probability
        self.food_layer.set_cells(True, condition)
```
I think random is important enough to built in. Doesn't have to be as part of this PR though.

mesa/space.py
EwoutH: What would your alternative be, if we have multiple cells with the same highest or lowest value?

wang-boyu:
To recap my comments during today's dev meeting, which are mainly about how things are named. Apologies for the tediousness.

1. I'm also in favor of the following suggestion. Plus, `layer.data` could instead be `layer.values` to be consistent.

   > I would prefer to exchange cell with value in most cases. For example instead of `set_cell` we could have `set_value`, which fits for all kind of property layers.
   > 
   > And `move_to_extreme_value_cell` could just drop the cell, which would also make it a bit shorter.
   > 
   > In cases where value doesn't fit maybe we can use pos or position, which is still generic.

2. Regarding the function names:

   ```python
   set_cell(Coordinate position, value)
   set_cells(value, condition)
   ```
   Perhaps `set_cells()` could be `set_cells(condition, value)` (with parameters places swapped)? This is because `set_cell()` also has `value` as the second parameter.

   ```python
   modify_cell(Coordinate position, operation, value)
   modify_cells(operation, value, condition_function)
   ```

   This is where I miss function overloading in Python. I understand why these two functions exist. It's just that users need to remember when to use `set_cell()` and `modify_cell()` for similar things. Would it be too much to ask for something like `set_value(position, value=None, operation=None)`, and users are expected to provide either `value` or `operation`?

   For the case of both `operation` and `value` are needed in the current implementation, users can combine them as a single `operation`, maybe in the form of a lambda or partial function.

   We can even go further to just have one `set_value(position=None, condition=None, value=None, operation=None)` where one of `position` and `condition`, and one of `value` and `operation` is needed.

3. This PropertyLayer approach is much more efficient than before, and is equivalent to a simultaneous activation. We probably need to highlight this somewhere in the documentation, in case users aren't aware of it (e.g., if they really need a random activation for PropertyLayer, although I can't think of any particular example at the moment).

EwoutH:
@quaquel curious what you think about this one. Going in the right direction, or shouldn't we entrench in (numpy) arrays to keep track of and manipulate environmental properties? Also considering future enhancements in data structures to the Grid itself.

quaquel:
I haven't read through the entire discussion, so some points might have been discussed.

As I explained yesterday in person, I once rigged up my own SingleGrid implementation using NumPy. I had the grid itself and a 3d index array twice the size of the actual grid. Moore, von Neumann, and hex grids can all be done with this by changing how you slice on the indexing array.  Next, you flatten the return from the slice on the indexing array and have the neighborhood indices you then use on the actual grid to get the objects referenced here. I don't recall the details, but handling neighborhoods this way was several orders of magnitude faster. So, I am all in favor of building on NumPy for gridded spaces. 

I would avoid allowing property layers to have different gridding from the main grid. Likewise, I would standardize the origin to avoid the obnoxious translation between two different coordinate systems. By reimplementing grids themselves on top of NumPy, this problem would disappear.

Another thing to consider is that you might want to make it possible for PropertyLayers to have behavior. For example, to simulate grass regrowth. Conceptually, this would turn property layers into agents because they would have both state and behavior. So either property layers would subclass Agent, or you rely on duck typing through an Agent protocol to implement this cleanly. It also means you want to be able to schedule PropertyLayer updates.

EwoutH:
That pretty much aligns with the earlier discussion in this PR and in #1889. Only problem I see is even if you implement the Grid as a NumPy array, you still want to store a list of agents in each cell. Those data structured aren't optimized or vectorized, contrary to the numerical arrays here in the PropertyLayers.

> Another thing to consider is that you might want to make it possible for PropertyLayer to have behavior. 

In https://github.com/projectmesa/mesa/issues/1889#issuecomment-1831968467 we bisected this issues in two parts:
- A layer of patches that's passive and only has some property that can be read and modified.
- A proper Patch (or Cell) Agent

This PR solves the former, for the latter there isn't a proposal yet.

<hr>

As for this PR, I'm going to do a simple implementation of proposal #1903 to be able to move agents. Then we can remove the explicit movement functions from the `_PropertyGrid`.

@tpike3 and @jackiekazil after that I think this feature is in good enough state to be included in 2.2.0 as experimental.

quaquel:
> That pretty much aligns with the earlier discussion in this PR and in #1889. Only problem I see is even if you implement the Grid as a NumPy array, you still want to store a list of agents in each cell. Those data structured aren't optimized or vectorized, contrary to the numerical arrays here in the PropertyLayers.
> 

True. However, the main performance difference comes from the slicing of the indexing array to get the indices of the neighboring cells. So you could still store stuff in the list of list things and only use NumPy to generate the indices given a neighborhood function, size, and whether you are on a torus or not.

EwoutH:
Updated move_agent method, which now can take a list of positions and choose either one randomly or the closest:

- #1920

mesa/space.py
EwoutH: Normal function works!

EwoutH:
With #1920 merged, I could remove the the move methods, making this PR a little less insanely big. 

```Python
target_cells = grid.select_cells_by_properties(conditions, mask=neighborhood_mask)
grid.move_agent_to_one_of(agent, target_cells)
```
Of course, you could make this an oneliner, but it might be a bit ugly.

```Python
grid.move_agent_to_one_of(agent, grid.select_cells_by_properties(conditions, mask=neighborhood_mask))
```

See the tests changed in a119cd9 for some more examples.

This PR is now fully ready for review, but please take your time, it's Christmas!🎄

EwoutH:
@quaquel and @Corvince, any thoughts on my last two points?

quaquel:
1. sounds like a good idea
2. Not sure. Why take NetLogo as your source of inspiration? In essence a grid is a discretization, so why not try to support continuous properties were possible?

EwoutH:
Will try an implementation for a single select function tomorrow.

> Not sure. Why take NetLogo as your source of inspiration?

We're standing on the shoulders of giants, right?

No mainly because many models and modelers are used to NetLogo, we should try to offer a superset of their capabilities. That makes porting models and enabling modelers to utilize Mesa easier.

> In essence a grid is a discretization, so why not try to support continuous properties were possible?

In the end that would be nice. Not sure how compatible that is with our current data structure (NumPy arrays). Maybe we can add some (optional) interpolation functionality when working with an continuous space.

I'm really trying to look hard for the the unknown unknowns.
- With a continuous space I imagine each agent having a (float, float) position. We will probably round (or floor) that when mapping to a cell as currently implemented in the PropertyLayer.
- When getting a value (so the other way around) we can either ask for the mapped value or an interpolated one.
- Setting values with some interpolation would probably be complicated, right?
- Radius checks (for neighbours etc.) could work like a float, and just check if the centers of patches are in it's radius

So I think the current PropertyLayer would work fine for most behavior. Still curious if I'm missing anything big.

We will be adding it as experimental, so maybe we should just start building it and then see if it works sufficiently or not.

Corvince:
> With #1920 merged, I could remove the the move methods, making this PR a little less insanely big.
> 
> ```python
> target_cells = grid.select_cells_by_properties(conditions, mask=neighborhood_mask)
> grid.move_agent_to_one_of(agent, target_cells)
> ```
> 
> Of course, you could make this an oneliner, but it might be a bit ugly.
> 
> ```python
> grid.move_agent_to_one_of(agent, grid.select_cells_by_properties(conditions, mask=neighborhood_mask))
> ```
> 

Of course outside this PR, but this got me thinking. Maybe analog to our new AgentSet we can define something like a PositionSet and this could become 
```python
grid.select_cells_by_properties(conditions, mask=neighborhood_mask).move_to(agent)
```

Although even better to read would be 
´´´python
agent.move_to(grid.select_cells_by_properties(conditions, mask=neighborhood_mask).select(n=1))
```

But we haven't explored agents space function much yet.

**Comment by EwoutH** [comment]
> Maybe analog to our new AgentSet we can define something like a PositionSet 

Also thought about it a few times. Each time really quickly tried to stop thinking about it due to the huge rewrite it would require.

But yeah, definitely worth thinking about.

**Comment by EwoutH** [comment]
I proposed in https://github.com/projectmesa/mesa/pull/1917#issuecomment-1870625103 to cut this feature out of the 2.2.0 release. We can discuss both the PositionSet as the ContiniousSpace implementation.

**Comment by EwoutH** [comment]
Thought about it some more, and I think it's best to first figure out how to port this to a `ContinousSpace`, and then think about a `PositionSet` kind of class. Because I think we first need to decide what a point is in a continuousspace and how to interpret that.

This means the HexGrid will be the only grid that doesn't work with the PropertyLayer. I don't like it, but I also don't see an logical way around it while keeping the PropertyLayer an ndarray.

**Comment by quaquel** [comment]
A set of points in continuous space is infinite by definition. So, if you want a `PositionSet` that works in continuous space, you have to somehow specify the boundary of the set. You cannot enumerate all positions, but in that way, you can at least check if points are in the set, and also generate any point that is within the set. I would suggest figuring out the API for the PositionSet first, starting for a simple grid. Once you have a fair idea of the API, we can worry about how to port this to a ContinuousSpace. 

What is the problem with hex spaces? In principle, you can implement hexspaces on top of an ndarray. you just get a slightly funny way of generating indices. I have done it a while ago when playing with Self Organizing Maps and learning numpy indexing and slicing.

**Comment by EwoutH** [comment]
Time for a quick call?

**Comment by EwoutH** [comment]
Okay, I worked on this PR some more:
- 3d752a4 I merged the `select_cells_by_properties()` and `select_extreme_value_cells()` methods into a single, more flexible single `select_cells()` method that incorperates the functionality of both. It now also allows multiple extreme values and multiple masks. Further, it adds a boolean to only include empty cells. 
- 26d2c80 I replaced the slow `get_empty_mask()` method with a very fast `empty_mask` property. Also fixed a bug.
- 598bdea Marked the PropertyLayer as experimental. Feedback will be collected in #1932.

I also updated the UML and all the examples in the PR description.

I think this PR is ready for final review and to be merged. Let's see how it works in the field!

@jackiekazil @tpike3 @rht @Corvince @quaquel

File: mesa/space.py [hunk 8]
```diff
@@ -548,7 +550,428 @@ def exists_empty_cells(self) -> bool:
         return len(self.empties) > 0
 
 
-class SingleGrid(_Grid):
+def is_lambda_function(function):
```
Corvince, line 553: Maybe call it is_single_argument_function!?

mesa/space.py
```diff
@@ -548,7 +550,428 @@ def exists_empty_cells(self) -> bool:
...
+    """Check if a function is a lambda function."""
+    return (
+        inspect.isfunction(function)
+        and len(inspect.signature(function).parameters) == 1
+    )
+
+
+def is_numpy_ufunc(function):
```
Corvince, line 561: I think this function is unnecessary, it's not really shorter or easier to read

mesa/space.py
```diff
@@ -548,7 +550,428 @@ def exists_empty_cells(self) -> bool:
...
+    return isinstance(function, np.ufunc)
+
+
+def ufunc_requires_additional_input(ufunc):
```
Corvince, line 565: Maybe this can be merged with the currently named is_lqmbda_function?

mesa/space.py
```diff
@@ -548,7 +550,428 @@ def exists_empty_cells(self) -> bool:
...
+    Methods:
+        set_cell(position, value): Sets the value of a single cell.
+        set_cells(value, condition): Sets the values of multiple cells, optionally based on a condition.
+        modify_cell(position, operation, value): Modifies the value of a single cell using an operation.
+        modify_cells(operation, value, condition_function): Modifies the values of multiple cells using an operation.
+        select_cells(condition, return_list): Selects cells that meet a specified condition.
+        aggregate_property(operation): Performs an aggregate operation over all cells.
+    """
+
+    agentset_experimental_warning_given = False
```
Corvince, line 591: Agentset?

mesa/space.py
```diff
@@ -548,7 +550,428 @@ def exists_empty_cells(self) -> bool:
...
+                condition_array = condition_function(self.data)
+            else:
+                vectorized_condition = np.vectorize(condition_function)
+                condition_array = vectorized_condition(self.data)
+
+        # Check if the operation is a lambda function or a NumPy ufunc
+        if is_numpy_ufunc(operation):
+            # Check if the ufunc requires an additional input
+            if ufunc_requires_additional_input(operation):
+                if value is None:
```
Corvince, line 729: A bit hard to read/understand with 3 nested ifs

mesa/space.py
```diff
@@ -548,7 +550,428 @@ def exists_empty_cells(self) -> bool:
...
+
+        Raises:
+            ValueError: If a property layer's dimensions do not match the grid dimensions.
+        """
+        super().__init__(width, height, torus)
+        self.properties = {}
+
+        # Initialize an empty mask as a boolean NumPy array
+        self._empty_mask = np.ones((self.width, self.height), dtype=bool)
+        print(self._empty_mask)
```
Corvince, line 827: Remove print statement

mesa/space.py
```diff
@@ -548,7 +550,428 @@ def exists_empty_cells(self) -> bool:
...
+                a list of PropertyLayer instances, or None to initialize without any property layers.
+
+        Raises:
+            ValueError: If a property layer's dimensions do not match the grid dimensions.
+        """
+        super().__init__(width, height, torus)
+        self.properties = {}
+
+        # Initialize an empty mask as a boolean NumPy array
+        self._empty_mask = np.ones((self.width, self.height), dtype=bool)
```
Corvince, line 826: I wonder if this should be inside Grid, since it is independent of properties. And then it is suspiciously similar to the empties property.

Corvince (commented):
Reviewed on mobile, so might have missed some things. But found only minor points, good work

EwoutH:
Thanks for reviewing! I’m on vacation until the 7th, I will try to process your feedback the 8th.

mesa/space.py
```diff
@@ -548,7 +550,428 @@ def exists_empty_cells(self) -> bool:
...
+        modify_cell(position, operation, value): Modifies the value of a single cell using an operation.
+        modify_cells(operation, value, condition_function): Modifies the values of multiple cells using an operation.
+        select_cells(condition, return_list): Selects cells that meet a specified condition.
+        aggregate_property(operation): Performs an aggregate operation over all cells.
+    """
+
+    agentset_experimental_warning_given = False
+
+    def __init__(
+        self, name: str, width: int, height: int, default_value, dtype=np.float32
```
quaquel, line 594: Why would you need to pass the width and height? Should this not be the same as the space to which this property layer belongs?

mesa/space.py
```diff
@@ -21,8 +21,10 @@
 from __future__ import annotations
```
tpike3, line 21: Can you please update the space module (and can you update the pylint as well as we switched to ruff, if I am reading the documentation correctly you should just be able to remove line 15-17)

mesa/space.py
tpike3: I would agree we want to try an consolidate these into one function. As you call it numerous times

tpike3:
This is very cool and particularly making more compatible with Mesa-Geo should really help link together the ecosystem. Can you also condense you commit history a little bit. (Not a must have a nice to have)

mesa/space.py
EwoutH: I think in preparation for when used with the `ContiniousSpace`, you need to have some width and height.

I’m fine with both leaving it or removing it.

mesa/space.py
EwoutH: If you allow multiple grids, it should right? Because each grid can have cells that are empty or not.

(Imagine a grid as a 2D layer, and you can have multiple 2D layers?)

mesa/space.py
EwoutH: Agreed

mesa/space.py
EwoutH: Correct

EwoutH:
@Corvince and @quaquel thanks for reviewing!

> Can you also condense you commit history a little bit. (Not a must have a nice to have)

You can use the “Squash and merge” method on GitHub to merge everything into a single commit to the main branch. That preserves the commit history in the PR, but merges one single commit into the main branch.

<hr>

I’m on vacation until the 7th, I can process the other feedback 8th. @quaquel, you have access to my fork, so if you want you can also do it.

mesa/space.py
quaquel: I agree you need a space and width.  However, my point was that these _always_ should be the same as the owning Space. So, rather than passing width and height, I guess you should pass the owning space instead and next retrieve the height and width from this.

mesa/space.py
```diff
@@ -548,7 +550,428 @@ def exists_empty_cells(self) -> bool:
...
+        """
+        Initializes a new PropertyLayer instance.
+
+        Args:
+            name (str): The name of the property layer.
+            width (int): The width of the grid (number of columns). Must be a positive integer.
+            height (int): The height of the grid (number of rows). Must be a positive integer.
+            default_value: The default value to initialize each cell in the grid. Should ideally
+                           be of the same type as specified by the dtype parameter.
+            dtype (data-type, optional): The desired data-type for the grid's elements. Default is np.float32.
```
rht, line 605: `np.float64` as the default would make more sense, as it is the default for Python's `float`.
```
In [4]: %timeit data = np.full((1000, 1000), 10, dtype=np.float32)
339 µs ± 998 ns per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [5]: %timeit data = np.full((1000, 1000), 10, dtype=np.float64)
813 µs ± 6.21 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```
It's ~2x slower, but the vectorized version would still be much faster in https://github.com/projectmesa/mesa/pull/1898#issuecomment-1849000346.

mesa/space.py
```diff
@@ -548,7 +550,428 @@ def exists_empty_cells(self) -> bool:
...
+class PropertyLayer:
+    """
+    A class representing a layer of properties in a two-dimensional grid. Each cell in the grid
+    can store a value of a specified data type.
+
+    Attributes:
+        name (str): The name of the property layer.
+        width (int): The width of the grid (number of columns).
+        height (int): The height of the grid (number of rows).
+        data (numpy.ndarray): A NumPy array representing the grid data.
```
rht, line 580: In the beginning, I found the new terms overwhelming. Would have eased the understanding if the documentation says that in essence, that the class's concrete representation is of a 2D NumPy array, with methods to perform bulk operation on them.

rht:
I find the [vectorized implementation of GoL](https://github.com/projectmesa/mesa/pull/1898#issuecomment-1848962861) fascinating. Reminds me of the other extreme of brevity of a GoL implementation, [in APL](https://aplwiki.com/wiki/Conway%27s_Game_of_Life) (in which NumPy's vectorized operations are based on).

mesa/space.py
EwoutH: I think for much more users the extra performance wil, be notable, and I don’t see any current models where float64 is needed.

We could also make `dtype` an required argument (without default).

mesa/space.py
EwoutH: Thanks for the feedback. If you have concrete suggestions for added docstring, please suggest them.

EwoutH:
@rht thanks for reviewing. We were a bit worried about your sudden disappearance. Good to have you back!

mesa/space.py
rht: > I think for much more users the extra performance wil, be notable, and I don’t see any current models where float64 is needed.

I'd be surprised if the default is `float32` in this case. Because I think the precision used should be consistent regarding with the floating point error accummulation, either `float32` everywhere, or `float64` everywhere. Users who want the speedup at the tradeoff of using `float32`, or using a mix of both, can manually choose to do so. We can document this in the [how-to guide](https://github.com/projectmesa/mesa/blob/main/docs/howto.rst).

> We could also make dtype an required argument (without default).

I'd say it should be optional, much like how it is for the `np.ndarray` constructors in NumPy.

mesa/space.py
```diff
@@ -548,7 +550,428 @@ def exists_empty_cells(self) -> bool:
...
+
+    Methods:
+        set_cell(position, value): Sets the value of a single cell.
+        set_cells(value, condition): Sets the values of multiple cells, optionally based on a condition.
```
rht, line 584: `set_cells(value, condition=None)` is clearer, because otherwise, as @wang-boyu said, the position of the `value` argument seems off.

mesa/space.py
rht: You could say something along the line of "a mental model of this class would be a 2D NumPy array with methods to update each cells individually or in bulk".

Related: https://stackoverflow.com/questions/6016271/why-is-it-better-to-have-100-functions-operate-on-one-data-structure-than-10-fun. In that the class only has 1 data structure, and so, is simple.

mesa/space.py
EwoutH: You mean you would like the `set_cells` method `condition` argument to be `None` by default?

mesa/space.py
EwoutH: @Corvince @jackiekazil @tpike3 @wang-boyu @quaquel I would like additional opinions on this.

I’m leaning towards making `dtype` a required argument so that users make a conscious choice. However, that requires familiarity with NumPy’s [data types](https://numpy.org/doc/stable/user/basics.types.html) for a proper choice.

Also I’m afraid users will, just pass `float`, which equates `np.float64` to making their grid twice as slow, while `np.float32` suffices in almost every situation. I can’t think of a single model or ABM use case in which 64-bit floating point precision is required, so I don’t think it would be a good default.

quaquel: I agree with @rht. Just follow numpy and use float 64 as default. Having `dtype` as an keyword argument would be convenient if users have good reasons to deviate they can easily do so.

EwoutH:
I notice I’m mentally overloading a bit on this PR. My brain is running out of RAM to track all the discussions and comments. This PR now tracks 85 comments and over a month’s worth of discussion.

I would highly recommend merging this **experimental** feature as-is. Everything works, is tested, and is documented. Making iterative, [atomic](https://www.pauline-vos.nl/atomic-commits/) modifications on each of the potential improvement.

This feature is **exactly** for which #1909 was meant. We can break everything about it an hundred times and scrap it fully if we want to.

Please use **Squash and merge** while merging. I don’t want 30 commits in the commit history. The PR serves as historical reference and grounding.

rht (approved):

rht:
@EwoutH I can take over applying the rest of the change requests. Should be done in less than an hour and then I will squash and merge. Well done!

mesa/space.py
rht: https://github.com/projectmesa/mesa/pull/1898/commits/6e12379dc1194522be515c7373974a2d79016b7f

mesa/space.py
rht: https://github.com/projectmesa/mesa/pull/1898/commits/627eccb59777b43bd36e7f81c04a5b290f1a438e

mesa/space.py
rht: Given that `is_lambda_function` has been renamed to `is_single_argument_function`, I will leave this one as is.

mesa/space.py
rht: Resolving this, because Ruff requires lowercase for the naming here.

mesa/space.py
rht: Indeed, this is a bug:
```python
        if isinstance(operation, np.ufunc):
            # Check if the ufunc requires an additional input
            if ufunc_requires_additional_input(operation):
                if value is None:
                    raise ValueError("This ufunc requires an additional input value.")
                modified_data = operation(self.data, value)
        else:
            # Vectorize non-ufunc operations
            vectorized_operation = np.vectorize(operation)
            modified_data = vectorized_operation(self.data)
```
Should have been
```python
        if isinstance(operation, np.ufunc) and ufunc_requires_additional_input(operation):
                if value is None:
                    raise ValueError("This ufunc requires an additional input value.")
                modified_data = operation(self.data, value)
        else:
            # Vectorize non-ufunc operations
            vectorized_operation = np.vectorize(operation)
            modified_data = vectorized_operation(self.data)
```
because in the original code, when `operation` doesn't require additional input, `modified_data` is undefined, and would raise an error.

rht: Actually, should have been
```python
        if isinstance(operation, np.ufunc):
            if ufunc_requires_additional_input(operation):
                if value is None:
                    raise ValueError("This ufunc requires an additional input value.")
                modified_data = operation(self.data, value)
            else:
                modified_data = operation(self.data)
        else:
            # Vectorize non-ufunc operations
            vectorized_operation = np.vectorize(operation)
            modified_data = vectorized_operation(self.data)
```
This can be simplified in the future.

rht: https://github.com/projectmesa/mesa/pull/1898/commits/f5f5065387715c1bd75de7ced36835e4a692cacd

mesa/space.py
rht: https://github.com/projectmesa/mesa/pull/1898/commits/f96a9ef011f70b753034a27ff7d9530c2d6e4091

mesa/space.py
rht: https://github.com/projectmesa/mesa/pull/1898/commits/3e7bd7b6c016e6c9957969f00fd74d8c46f371c5

mesa/space.py
rht: https://github.com/projectmesa/mesa/pull/1898/commits/c71c9f54722c0a24bbbfb37248b7209c283efd44

mesa/space.py
rht: https://github.com/projectmesa/mesa/pull/1898/commits/363fb0b6469808581c9e37b10934eba6e67bb3af

rht:
@EwoutH these are the remaining unaddressed reviews:
- https://github.com/projectmesa/mesa/pull/1898#discussion_r1439742992
- https://github.com/projectmesa/mesa/pull/1898#discussion_r1440205916

EwoutH:
Thanks for taking care of the comments and merging!

Very glad it's in!

dylan-munson:
Apologies if this is the incorrect place to ask this question, I am relatively new to this community and Github in general, but one issue I have encountered using this awesome feature is how to use it with the datacollector method.

I'm trying to implement the sugar/spice model with pollution in Mesa (modifying the sugarscape_g1mt model from mesa-examples), and currently have the following in my datacollector definition:

`        self.datacollector = mesa.DataCollector(
            model_reporters={ #add externality measures here once pollution is included?
                "Trader": lambda m: m.schedule.get_type_count(Trader),
                "Trade Volume": lambda m: sum(
                    len(a.trade_partners)
                    for a in m.schedule.agents_by_type[Trader].values()
                ),
                "Price": lambda m: geometric_mean(
                    flatten(
                        [a.prices for a in m.schedule.agents_by_type[Trader].values()]
                    )
                ),
                "Pollution": lambda m: m.grid.properties["pollution"].data},
            agent_reporters={"Trade Network": lambda a: get_trade(a),
                             "Welfare": lambda a: get_welfare(a)},
        )`

Everything seems to work as intended, as in the base model, except for the pollution data; it seems that for all steps, only the final pollution property layer is being stored (i.e. in the final results dataframe, the "Pollution" column is identical across model steps).  Perhaps this is just me misunderstanding how the data collector works, but I wanted to see if this was potentially just not something that the PropertyLayer class is intended to handle.  I imagine there should be some way to do it (for this model, we of course expect the distribution over space of pollution to change with each step), but I'm struggling to figure it out.  Any help is greatly appreciated!

EwoutH:
Thanks for reaching out! Can you share a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example)?

dylan-munson:
Yes, apologies for not doing so initially.  Below is an extremely simplified version of what I am trying to do:

```
import mesa

class testAgent(mesa.Agent):
    def __init__(self, unique_id, model, pos):
        self.pos = pos
    def pollute(self):
        self.model.grid.properties["pollution"].modify_cell(self.pos, lambda p: p+1)

class testModel(mesa.Model):
    def __init__(self, width = 20, height = 20, initial_population=200):
        super().__init__()
        self.width = width
        self.height = height
        self.initial_population = initial_population
        
        # initiate activation schedule
        self.schedule = mesa.time.RandomActivationByType(self) #activate different type of agents in different order each time
        # initiate mesa grid class
        self.grid = mesa.space.MultiGrid(self.width, self.height, torus=False)
        #add pollution property layer
        pollution_layer = mesa.space.PropertyLayer(name = "pollution", width = self.width, height = self.height, default_value = 0)
        self.grid.add_property_layer(pollution_layer)
        
        self.datacollector = mesa.DataCollector(model_reporters = {"Pollution": lambda m: m.grid.properties["pollution"].data})
        
        agent_id = 0
        for i in range(self.initial_population):
            # get agent position
            x = self.random.randrange(self.width)
            y = self.random.randrange(self.height)
            agent = testAgent(agent_id, self, (x, y))
            # place agent
            self.grid.place_agent(agent, (x, y))
            self.schedule.add(agent)
            agent_id += 1
    
    def randomizeAgents(self):
        agents_shuffle = list(self.schedule.agents_by_type[testAgent].values())
        self.random.shuffle(agents_shuffle)

        return agents_shuffle
    
    def step(self):
        agents_shuffle = self.randomizeAgents()
        for agent in agents_shuffle:
            agent.pollute()
        
    def run_model(self, step_count=20):
        for i in range(step_count):
            self.step()
            
model = testModel()
model.run_model()
model_results = model.datacollector.get_model_vars_dataframe()
```
If you check the model_results object after running, you will see that all of the arrays in the Pollution column are exactly the same--specifically, they seem to be the final step levels of pollution in each grid cell.  The pollution at step 0, at least, should be 0 in every entry, which is not the case.
So I am wondering if the issue is with the way I am calling the pollution property layer in DataCollector (perhaps with the lambda function I am using? or the way the step function has been defined?), or if use with the DataCollector is just not part of the functionality for PropertyLayer.  Thanks.

EwoutH:
Thanks for the minimal example. I can reproduce this bug, and it's indeed unintended behavior. Could you open a new issue for this?

I will investigate a bit further in the meantime.

dylan-munson:
Thanks for looking into this, I've opened an issue and will check back for updates/to see if anything else is needed.

PR #1910, opened 2023-12-17
time: Use weak references to agents in schedulers

EwoutH:
_This PR builts on the `agents` defaultdict introduced in #1894. See 89b61ec for the code changes proposed in this PR._

This PR modifies the schedulers to use weak references for agent management. The update aims to improve memory efficiency and garbage collection, especially in scenarios where agents are dynamically added and removed during model execution.

It also ensures that the Agent `remove()` method (introduced in #1894) also removes the Agent from the schedule.

**Key changes**
- Agents within schedulers are now stored as weak references (`weakref.ref[Agent]`), reducing the memory footprint and allowing for more efficient garbage collection.
- The `add` method in `BaseScheduler` and derived classes has been updated to store agents as weak references.
- The `remove` method has been modified to handle both direct agent objects and weak references, improving flexibility and robustness.
- Updated `do_each` method to iterate over a copy of agent keys and check the existence of agents before calling methods, ensuring stability during dynamic agent removal.
- Updated the `agents` property to return a list of live agent instances, maintaining backward compatibility for user code that iterates over agents.
- Ensured all existing tests pass, confirming the stability and correctness of the changes.

The changes result in a more memory-efficient scheduler implementation in Mesa, while preserving existing functionality and ensuring backward compatibility.

All 27 current tests pass without any modifications, so there shouldn't be changes breaking compatibility.

### DIFF
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -26,6 +26,7 @@ Key concepts:
 from __future__ import annotations
 
 import heapq
+import weakref
 from collections import defaultdict
 
 # mypy
@@ -64,7 +65,7 @@ class BaseScheduler:
         self.model = model
         self.steps = 0
         self.time: TimeT = 0
-        self._agents: dict[int, Agent] = {}
+        self._agents: dict[int, weakref.ref[Agent]] = {}
 
     def add(self, agent: Agent) -> None:
         """Add an Agent object to the schedule.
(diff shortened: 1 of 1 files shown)


EwoutH:
Rebased on main, now that #1894 is merged.

quaquel:
I think you can simply the code a bit by using [WeakrefValueDictionary](https://www.google.com/search?client=safari&rls=en&q=python+weakrefdict&ie=UTF-8&oe=UTF-8) directly. This will make sure that items in the dict are automagically removed if referenced value no longer exists. So, some of the None checks you now have would become redundant because this is handled within the data structure for you.

EwoutH:
Thanks! If we go this direction, I will certainly look into it.

Right now I would like to propose a different direction:

- https://github.com/projectmesa/mesa/discussions/1912

Because this just looks way too good to ignore.
```python
class MyModel(Model):
    def step(self):
        self.do_each("move")
        self.do_each("eat")
        self.do_each("sleep")
```

quaquel:
My point was primarily about the fact that there are some useful collections in weakref that you want to use as much as possible. This point applies also when using these in spaces. It is always better to use a data structure that handles an issue (refs to None in this case) rather than solving it in your code (as you did in this draft). 

The scheduler discussion is interesting and I am still thinking about it.

EwoutH:
Fair point, thanks.

quaquel:
How do you want to proceed with this now that #1916 is ready to be merged?

EwoutH:
Probably close it. The next effort should be to refactor the schedulers using AgentSet functionality, and possibly deprecate them.

quaquel:
Ok, if I get to it, I'll open a separate PR reimplementing the schedulers on top of AgentSet. I already had a quick look. It can be done in a backward compatible way while raising some warnings here and there.

EwoutH:
Yeah it should be pretty easy, all the examples all already in #1916.

I have another deal: If you handle my email I will do this implementation 😇.

EwoutH:
Closing this, succeeded by #1926.

PR #1911, opened 2023-12-18
Add select_agents method to Model

EwoutH:
This PR introduces a new `select_agents` method to the Model class. It's inspired by NetLogo's [`n-of`](https://ccl.northwestern.edu/netlogo/docs/dict/n-of.html) functions and an earlier concept from @rht as discussed in #1894. It aims to provide a flexible and efficient way to select and filter agents in a model, based on a wide range of criteria.

#### Features
The `select_agents` method offers several key functionalities:

1. **Selective quantity (`n`):** Specify the number of agents to select. If `n` is omitted, all matching agents are chosen.
2. **Criteria-based sorting (`sort` and `direction`):** Sort agents based on one or more attributes. The sorting order can be set individually for each attribute.
3. **Functional filtering (`filter_func`):** Use a lambda function or a callable to apply complex filter conditions.
4. **Type-based filtering (`agent_type`):** Filter agents by their class type, allowing for selection among specific subclasses.
5. **Flexible size handling (`up_to`):** When `True`, the method returns up to `n` agents, which is useful when the available agent count is less than `n`.

### Usage examples
1. **Random Selection of a Specific Number of Agents:**
   ```python
   selected_agents = model.select_agents(n=10)
   ```

2. **Sorting and Selecting Agents Based on Multiple Criteria:**
   ```python
   selected_agents = model.select_agents(n=5, sort=['wealth', 'age'], direction=['highest', 'lowest'])
   ```

3. **Using a Custom Filter Function:**
   ```python
   selected_agents = model.select_agents(filter_func=lambda agent: agent.energy >= 5)
   ```

4. **Selecting Agents of a Specific Type:**
   ```python
   selected_agents = model.select_agents(agent_type=SpecificAgentClass)
   ```

5. **Selecting Agents with a Complex Conditional Filter:**
   ```python
   selected_agents = model.select_agents(filter_func=lambda agent: agent.age > 30 and agent.wealth > 50)
   ```

6. **Combining Type Filtering with Attribute Sorting:**
   ```python
   selected_agents = model.select_agents(agent_type=WorkerAgent, sort=['experience'], direction=['highest'])
   ```

7. **Selecting a Maximum Number of Agents when Available Agents are Fewer:**
   ```python
   selected_agents = model.select_agents(n=15, up_to=True)
   ```

8. **Inverse Functional Filtering for Excluding Certain Agents:**
   ```python
   selected_agents = model.select_agents(filter_func=lambda agent: not agent.is_retired)
   ```

#### Implementation Details
[...]

Issue #1905, opened 2023-12-13
Proposal: Agent selection method


EwoutH:
A built-in method for agent selection and filtering could help in many situations to select an appropriate subset of agents. This feature request proposes the introduction of a versatile `Model` method, named `select_agents`. This function will allow users to select agents based on a combination of attributes, agent types, and functional filters, with the flexibility of specifying the number of agents to select.

This proposal is inspired by the various variants of the [`n-of`](http://ccl.northwestern.edu/netlogo/docs/dict/n-of.html) functions in NetLogo and @rht smart idea in https://github.com/projectmesa/mesa/pull/1894#issuecomment-1834206436.

**Functionality Overview:**
The `select_agents` function is designed to be highly flexible and adaptable to various modeling scenarios. Its key features include:

1. **Optional Quantity of Agents (`n`):**
   - `n` (Optional): An integer specifying the desired number of agents to select. If not specified, all agents meeting the criteria will be selected.

2. **Criteria-Based Sorting (`sort` and `direction`):**
   - `sort` (Optional): A list of attribute names to sort agents. It defines the criteria for selection.
   - `direction` (Optional): Accompanying `sort`, this list specifies the sorting direction ('lowest' or 'highest') for each criterion.

3. **Agent Filtering (`filter`):**
   - `filter` (Optional): A lambda function or callable that applies additional conditions to filter agents. It allows for complex logical expressions to refine the selection.

4. **Type-Based Filtering (`agent_type`):**
   - `agent_type` (Optional): A class or list of classes specifying the type(s) of agents to be included in the selection. This ensures that only agents of certain types or subclasses are considered.
   - _This feature depends on #1894._

5. **Flexible Size Handling (`up_to`):**
   - `up_to` (Optional): A boolean that, when True, allows the function to return up to `n` agents, preventing errors if `n` exceeds the available agent count. It defaults to True for convenience.

**Usage Scenarios:**

1. **Selecting a Specific Number of Agents:**
   - To select a specific number of agents randomly:
     ```python
     select_agents(n=10)
     ```

2. **Complex Multi-Criteria Selection:**
   - Select agents based on multiple attributes, like 'wealth' and 'age':
     ```python
     select_agents(n=5, sort=['wealth', 'age'], direction=['highest', 'lowest'])
     ```

3. **Selection with Functional Filtering:**
   - Apply a condition, such as selecting agents with 'energy' above a certain threshold:
     ```python
[...]

PR #1916, opened 2023-12-19
Introduce AgentSet class

EwoutH:
_This PR is based on the discussion in https://github.com/projectmesa/mesa/discussions/1912#discussioncomment-7890448 for the initial discussion._

This PR introduces the `AgentSet` class in the Mesa agent-based modeling framework, along with significant changes to the agent management process. The `AgentSet` class is designed to encapsulate and manage a collection of agents, providing methods for efficient selection, sorting, shuffling, and applying actions to groups of agents. This addition aims to enhance the framework's scalability and flexibility in handling agent operations.

### Key Changes

1. **Agent Class Modifications**: The `Agent` class has been updated to manage agent registration and removal within the model's `_agents` attribute directly. This approach eliminates the need for separate registration and removal methods.
2. **Model Class Enhancements**: The `Model` class now utilizes the `AgentSet` class. The `agents` property returns an `AgentSet` instance, representing all agents in the model, thus streamlining agent access and manipulation.
3. **AgentSet Functionality**: The `AgentSet` class includes methods like `select`, `shuffle`, `sort`, and `do` for more intuitive operations on agent collections. These methods facilitate tasks like filtering agents, randomizing their order, sorting, and applying actions to each agent.
4. **Efficient Agent Management**: `AgentSet` maintains weak references to agents, enabling efficient management of agent lifecycles and garbage collection. It uses a `WeakKeyDictionary` to store agents.

### Usage examples

1. **Selecting and Filtering Agents**:
    ```python
    # Select agents with a specific attribute
    selected_agents = model.agents.select(lambda agent: agent.attribute > threshold)
    ```

2. **Shuffling Agents**:
    ```python
    # Randomize agent order
    shuffled_agents = model.agents.shuffle()
    ```

3. **Sorting Agents**:
    ```python
    # Sort agents by wealth, high to low
    sorted_agents = model.agents.sort("wealth")
    ```

4. **Applying Actions to Agents**:
    ```python
    # Apply a method to each agent
    model.agents.do('method_name')
    ```

5. **Retrieving Agent Attributes**:
    ```python
    # Get a specific attribute from all agents
    attributes = model.agents.get('attr_name')
    ```

### Advanced Chaining examples

1. **Filter, Shuffle, and Apply an Action**:
    Chaining `select`, `shuffle`, and `do` allows for filtering, randomizing, and then performing an action on the selected agents.
    ```python
    # Select agents meeting a condition, shuffle them, and then apply a method
    model.agents.select(lambda a: a.condition).shuffle().do('action_method')
    ```

2. **Sort, Select Top N, and Get Attribute**:
    This chain sorts agents, selects the top N agents based on the sort, and retrieves a specific attribute from these agents.
    ```python
    # Sort agents, select the top N, and get an attribute
    top_agents_attributes = model.agents.sort(key=lambda a: a.rank).select(n=5).get('attribute')
    ```

3. **Complex Filtering and Action Application**:
    Combining multiple filters and then applying an action. This chain can be particularly useful for complex conditional operations.
    ```python
    # Apply a method to agents that meet multiple conditions
    model.agents.select(lambda a: a.condition1).select(lambda a: a.condition2).do('method')
    ```

4. **Combined Set Operations and Action Application**:
    If set-like operations (union, intersection, difference) are implemented, they can be combined with action application.
    ```python
    # Combine two AgentSets and apply a method to the combined set
    (agentset1 + agentset2).do('collaborative_method')
    ```

5. **Shuffle, Sort, and Sequential Action Execution**:
    This chain first randomizes the order of agents, then sorts them based on an attribute, and finally applies a method sequentially.
    ```python
    # Shuffle, sort, and then apply a method in sequence
    model.agents.shuffle().sort(key=lambda a: a.priority).do('sequential_method')
    ```

6. **Nested Chain Operations**:
    Combining chaining with nested operations for more sophisticated scenarios.
    ```python
    # Nested operation: Select agents, then within that selection, shuffle and apply an action
    model.agents.select(lambda a: a.group == 'X').shuffle().do('group_specific_method')
    ```

### Replacing schedulers 
You can use the new AgentSet functionality to replacer the current schedulers and create entirely new ones.

#### 1. BaseScheduler

The `BaseScheduler` activates each agent once per step, in the order they were added.

```python
class MyModel(Model):
    def step(self):
        self.agents.do('step')
```

#### 2. RandomActivation

`RandomActivation` activates each agent once per step, in random order.

```python
class MyModel(Model):
    def step(self):
        self.agents.shuffle().do('step')
```

#### 3. SimultaneousActivation

In `SimultaneousActivation`, all agents are activated simultaneously. This can be simulated by first collecting all agents' decisions and then applying them.

```python
class MyModel(Model):
    def step(self):
        self.agents.do("step")
        self.agents.do("advance")
```

#### 4. StagedActivation

`StagedActivation` allows for multiple stages per step. For instance, agents first "move", then "eat", then "sleep".

```python
class MyModel(Model):
    def step(self):
        self.agents.do('move')
        self.agents.do('eat')
        self.agents.do('sleep')
```

#### 5. RandomActivationByType

`RandomActivationByType` activates each type of agent in random order.

```python
class MyModel(Model):
    def step(self):
        # Assuming AgentType1 and AgentType2 are subclasses of Agent
        self.agents.select(agent_type=AgentType1).shuffle().do('step')
        self.agents.select(agent_type=AgentType2).shuffle().do('step')
```

These examples demonstrate how the `AgentSet` class can be used to replicate the behavior of different schedulers in Mesa. The `AgentSet` class provides a more flexible and concise way to define the activation and interaction patterns among agents within the model.

#### Future work
- Adding more Sequence and MutableSet methods (e.g., `__reversed__`, `index`, `count`, `clear`, `pop`, `__ior__`) for enhanced performance and functionality.
- Rewriting current schedulers using AgentSets.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -115,6 +115,7 @@ class AgentSet(MutableSet, Sequence):
         filter_func: Callable[[Agent], bool] | None = None,
         n: int = 0,
         inplace: bool = False,
+        agent_type: type[Agent] | None = None,
     ) -> AgentSet:
         """
         Select a subset of agents from the AgentSet based on a filter function and/or quantity limit.
@@ -124,14 +125,18 @@ class AgentSet(MutableSet, Sequence):
                 agent should be included in the result. Defaults to None, meaning no filtering is applied.
             n (int, optional): The number of agents to select. If 0, all matching agents are selected. Defaults to 0.
             inplace (bool, optional): If True, modifies the current AgentSet; otherwise, returns a new AgentSet. Defaults to False.
+            agent_type (type[Agent], optional): The class type of the agents to select. Defaults to None, meaning no type filtering is applied.
 
         Returns:
             AgentSet: A new AgentSet containing the selected agents, unless inplace is True, in which case the current AgentSet is updated.
         """
+        agents = list(self._agents.keys())
+
         if filter_func is not None:
-            agents = [agent for agent in self._agents if filter_func(agent)]
-        else:
-            agents = list(self._agents.keys())
+            agents = [agent for agent in agents if filter_func(agent)]
+
+        if agent_type is not None:
+            agents = [agent for agent in agents if isinstance(agent, agent_type)]
 
         if n:
             agents = agents[:n]
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -10,7 +10,7 @@ from __future__ import annotations
 from random import Random
 
 # mypy
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Callable, Iterator
 
 if TYPE_CHECKING:
     # We ensure that these are not imported during runtime to prevent cyclic
@@ -41,12 +41,15 @@ class Agent:
         self.model = model
         self.pos: Position | None = None
 
-        # Register the agent with the model using defaultdict
-        self.model.agents[type(self)][self] = None
+        # Directly register the agent with the model
+        if type(self) not in self.model._agents:
+            self.model._agents[type(self)] = set()
+        self.model._agents[type(self)].add(self)
 
     def remove(self) -> None:
         """Remove and delete the agent from the model."""
-        self.model.agents[type(self)].pop(self)
+        if type(self) in self.model._agents:
+            self.model._agents[type(self)].discard(self)
 
     def step(self) -> None:
         """A single step of the agent."""
@@ -57,3 +60,41 @@ class Agent:
     @property
     def random(self) -> Random:
         return self.model.random
+
+
+class AgentSet:
+    def __init__(self, agents: set[Agent], model: Model):
+        self._agents = agents
+        self.model = model
+
+    def __len__(self):
+        return len(self._agents)
+
+    def __iter__(self) -> Iterator[Agent]:
+        return iter(self._agents)
+
+    def __contains__(self, agent: Agent) -> bool:
+        """Check if an agent is in the AgentSet."""
+        return agent in self._agents
+
+    def select(self, filter_func: Callable[[Agent], bool] | None = None) -> AgentSet:
+        if filter_func is None:
+            return AgentSet(set(self._agents), self.model)
+        return AgentSet(
+            {agent for agent in self._agents if filter_func(agent)}, self.model
+        )
+
+    def shuffle(self) -> AgentSet:
+        shuffled_agents = list(self._agents)
+        self.model.random.shuffle(shuffled_agents)
+        return AgentSet(set(shuffled_agents), self.model)
+
+    def sort(self, key: Callable[[Agent], Any], reverse: bool = False) -> AgentSet:
+        sorted_agents = sorted(self._agents, key=key, reverse=reverse)
+        return AgentSet(set(sorted_agents), self.model)
+
+    def do_each(self, method_name: str):
+        for agent in self._agents:
+            getattr(agent, method_name)()
+
+    # Additional methods like union, intersection, difference, etc., can be added as needed.
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -50,11 +50,8 @@ class Agent:
 
     def remove(self) -> None:
         """Remove and delete the agent from the model."""
-        try:
-            # remove agent
+        with contextlib.suppress(KeyError):
             self.model._agents[type(self)].pop(self)
-        except KeyError:
-            pass
 
     def step(self) -> None:
         """A single step of the agent."""
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -42,15 +42,16 @@ class Agent:
         self.model = model
         self.pos: Position | None = None
 
-        # Directly register the agent with the model
-        if type(self) not in self.model._agents:
-            self.model._agents[type(self)] = set()
-        self.model._agents[type(self)].add(self)
+        # register agent
+        self.model._agents[type(self)][self] = None
 
     def remove(self) -> None:
         """Remove and delete the agent from the model."""
-        if type(self) in self.model._agents:
-            self.model._agents[type(self)].discard(self)
+        try:
+            # remove agent
+            self.model._agents[type(self)].pop(self)
+        except KeyError:
+            pass
 
     def step(self) -> None:
         """A single step of the agent."""
@@ -140,6 +141,9 @@ class AgentSet:
         # TODO::
         # TBD:: it is a bit tricky to make this work
         # part of the problem is that there is no weakreflist
+        #  might also be fixable through weakref.finalize
+        # TODO:: make slice also work
+        # item can be int or slice
         agent = self._indices[item]()
         if agent is None:
             # the agent has been garbage collected
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -7,8 +7,9 @@ Core Objects: Agent
 # Remove this __future__ import once the oldest supported Python is 3.10
 from __future__ import annotations
 
+import operator
 import weakref
-from collections.abc import MutableSet
+from collections.abc import MutableSet, Sequence
 from random import Random
 
 # mypy
@@ -65,7 +66,7 @@ class Agent:
         return self.model.random
 
 
-class AgentSet(MutableSet):
+class AgentSet(MutableSet, Sequence):
     """Ordered set of agents"""
 
     def __init__(self, agents: Iterable[Agent], model: Model):
@@ -85,53 +86,72 @@ class AgentSet(MutableSet):
         """Check if an agent is in the AgentSet."""
         return agent in self._agents
 
-    def select(self, filter_func: Callable[[Agent], bool] | None = None, inplace: bool = False) -> AgentSet:
-        if filter_func is None:
-            return self if inplace else AgentSet(iter(self), self.model)
+    def select(self, filter_func: Callable[[Agent], bool] | None = None, n: int = 0, inplace: bool = False) -> AgentSet:
+        """select agents from AgentSet
 
-        agents = [agent for agent in self._agents.keys() if filter_func(agent)]
+        Args:
+            filter_func (Callable[[Agent]]): function to filter agents. Function should return True if agent is to be
+                                             included, false otherwise
+            n (int, optional): number of agents to return. Defaults to 0, meaning all agents are returned
+            inplace (bool, optional): updates agentset inplace if True, else return new Agentset. Defaults to False.
 
-        if inplace:
-            self._reorder(agents)
-            return self
+
+        """
+        if filter_func is not None:
+            agents = [agent for agent in self._agents.keys() if filter_func(agent)]
         else:
-            return AgentSet(
-                agents,
-                self.model
-            )
+            agents = list(self._agents.keys())
 
-    def shuffle(self, inplace: bool = False)-> AgentSet:
+        if n:
+            agents = agents[:n]
+
+        return AgentSet(agents, self.model) if not inplace else self._update(agents)
+
+    def shuffle(self, inplace: bool = False) -> AgentSet:
         shuffled_agents = list(self)
         self.model.random.shuffle(shuffled_agents)
 
-        if inplace:
-            self._reorder(shuffled_agents)
-            return self
-        else:
-            return AgentSet(shuffled_agents, self.model)
+        return AgentSet(shuffled_agents, self.model) if not inplace else self._update(shuffled_agents)
 
-    def sort(self, key: Callable[[Agent], Any], reverse: bool = False, inplace: bool = False)-> AgentSet:
+    def sort(self, key: Callable[[Agent], Any]|str, reverse: bool = False, inplace: bool = False) -> AgentSet:
+        if isinstance(key, str):
+            key = operator.attrgetter(key)
+
         sorted_agents = sorted(self._agents.keys(), key=key, reverse=reverse)
 
-        if inplace:
-            self._reorder(sorted_agents)
-            return self
-        else:
-            return AgentSet(sorted_agents, self.model)
+        return AgentSet(sorted_agents, self.model) if not inplace else self._update(sorted_agents)
 
-    def _reorder(self, agents: Iterable[Agent]):
+    def _update(self, agents: Iterable[Agent]):
         _agents = weakref.WeakKeyDictionary()
         for agent in agents:
             _agents[agent] = None
 
         self._agents = _agents
+        return self
 
-    def do_each(self, method_name: str, *args, **kwargs) -> list[Any]:
-        """invoke method on each agent"""
-        return [getattr(agent, method_name)(*args, **kwargs) for agent in self._agents]
+    def do(self, method_name: str, *args, return_results: bool = False, **kwargs) -> AgentSet | list[Any]:
+        """invoke method on each agent
 
-    def get_each(self, attr_name: str) -> list[Any]:
-        """get attribute value on each agent"""
+        Args:
+            method_name (str): name of the method to call on each agent
+            return_results (bool): whether to return the result from the method call or
+                                   return the AgentSet itself. Defaults to False, so you can
+                                   continue method chaining
+
+        Additional arguments and keyword arguments will be passed to the method being called
+
+        """
+        res = [getattr(agent, method_name)(*args, **kwargs) for agent in self._agents]
+
+        return res if return_results else self
+
+    def get(self, attr_name: str) -> list[Any]:
+        """get attribute value on each agent
+
+        Args:
+            attr_name (str): name of the attribute to get from eahc agent in the set
+
+        """
         return [getattr(agent, attr_name) for agent in self._agents]
 
     def __getitem__(self, item: int | slice) -> Agent:
@@ -160,4 +180,9 @@ class AgentSet(MutableSet):
 
     def __setstate__(self, state):
         self.model = state['model']
-        self._reorder(state['agents'])
+        self._update(state['agents'])
+
+
+# consider adding for performance reasons
+# for Sequence: __reversed__, index, and count
+# for MutableSet clear, pop, remove, __ior__, __iand__, __ixor__, and __isub__
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 
 import contextlib
 import operator
+import warnings
 import weakref
 from collections.abc import MutableSet, Sequence
 from random import Random
@@ -66,6 +67,12 @@ class Agent:
 
 class AgentSet(MutableSet, Sequence):
     """
+    .. warning::
+        The AgentSet is experimental. It may be changed or removed in any and all future releases, including
+        patch releases.
+        We would love to hear what you think about this new feature. If you have any thoughts, share them with
+        us here: https://github.com/projectmesa/mesa/discussions/1919
+
     A collection class that represents an ordered set of agents within an agent-based model (ABM). This class
     extends both MutableSet and Sequence, providing set-like functionality with order preservation and
     sequence operations.
@@ -92,11 +99,20 @@ class AgentSet(MutableSet, Sequence):
             agents (Iterable[Agent]): An iterable of Agent objects to be included in the set.
             model (Model): The ABM model instance to which this AgentSet belongs.
         """
+        self.model = model
+
+        if not self.model.agentset_experimental_warning_given:
+            self.model.agentset_experimental_warning_given = True
+            warnings.warn(
+                "The AgentSet is experimental. It may be changed or removed in any and all future releases, including patch releases.\n"
+                "We would love to hear what you think about this new feature. If you have any thoughts, share them with us here: https://github.com/projectmesa/mesa/discussions/1919",
+                FutureWarning,
+                stacklevel=2,
+            )
+
         self._agents = weakref.WeakKeyDictionary()
         for agent in agents:
             self._agents[agent] = None
-
-        self.model = model
 
     def __len__(self) -> int:
         """Return the number of agents in the AgentSet."""
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -109,7 +109,7 @@ class AgentSet(MutableSet, Sequence):
 
     def shuffle(self, inplace: bool = False) -> AgentSet:
         shuffled_agents = list(self)
-        self.model.random.shuffle(shuffled_agents)
+        self.random.shuffle(shuffled_agents)
 
         return AgentSet(shuffled_agents, self.model) if not inplace else self._update(shuffled_agents)
 
@@ -181,6 +181,10 @@ class AgentSet(MutableSet, Sequence):
     def __setstate__(self, state):
         self.model = state['model']
         self._update(state['agents'])
+
+    @property
+    def random(self) -> Random:
+        return self.model.random
 
 
 # consider adding for performance reasons
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -7,10 +7,11 @@ Core Objects: Agent
 # Remove this __future__ import once the oldest supported Python is 3.10
 from __future__ import annotations
 
+import weakref
 from random import Random
 
 # mypy
-from typing import TYPE_CHECKING, Any, Callable, Iterator
+from typing import TYPE_CHECKING, Any, Callable, Iterator, Iterable
 
 if TYPE_CHECKING:
     # We ensure that these are not imported during runtime to prevent cyclic
@@ -63,38 +64,70 @@ class Agent:
 
 
 class AgentSet:
-    def __init__(self, agents: set[Agent], model: Model):
-        self._agents = agents
+    def __init__(self, agents: Iterable[Agent], model: Model):
+        self._agents = weakref.WeakKeyDictionary()
+
+        for agent in agents:
+            self._agents[agent] = None
+
         self.model = model
 
-    def __len__(self):
+    def __len__(self) -> int:
         return len(self._agents)
 
     def __iter__(self) -> Iterator[Agent]:
-        return iter(self._agents)
+        return iter(self._agents.keys())
 
     def __contains__(self, agent: Agent) -> bool:
         """Check if an agent is in the AgentSet."""
         return agent in self._agents
 
-    def select(self, filter_func: Callable[[Agent], bool] | None = None) -> AgentSet:
+    def select(self, filter_func: Callable[[Agent], bool] | None = None, inplace: bool = False) -> AgentSet:
         if filter_func is None:
-            return AgentSet(set(self._agents), self.model)
-        return AgentSet(
-            {agent for agent in self._agents if filter_func(agent)}, self.model
-        )
+            if inplace:
+                return self
+            else:
+                return AgentSet(list(self._agents.keys()), self.model)
+        else:
+            agents = [agent for agent in self._agents.keys() if filter_func(agent)]
 
-    def shuffle(self) -> AgentSet:
-        shuffled_agents = list(self._agents)
+            if inplace:
+                self._reorder(agents)
+                return self
+            else:
+                return AgentSet(
+                    agents,
+                    self.model
+                )
+
+    def shuffle(self, inplace: bool = False) -> AgentSet:
+        shuffled_agents = list(self._agents.keys())
         self.model.random.shuffle(shuffled_agents)
-        return AgentSet(set(shuffled_agents), self.model)
 
-    def sort(self, key: Callable[[Agent], Any], reverse: bool = False) -> AgentSet:
-        sorted_agents = sorted(self._agents, key=key, reverse=reverse)
-        return AgentSet(set(sorted_agents), self.model)
+        if inplace:
+            self._reorder(shuffled_agents)
+            return self
+        else:
+            return AgentSet(shuffled_agents, self.model)
 
-    def do_each(self, method_name: str):
-        for agent in self._agents:
-            getattr(agent, method_name)()
+    def sort(self, key: Callable[[Agent], Any], reverse: bool = False, inplace: bool = False) -> AgentSet:
+        sorted_agents = sorted(list(self._agents.keys()), key=key, reverse=reverse)
+
+        if inplace:
+            self._reorder(sorted_agents)
+            return self
+        else:
+            return AgentSet(sorted_agents, self.model)
+
+    def _reorder(self, agents: Iterable[Agent]):
+        _agents = weakref.WeakKeyDictionary()
+        for agent in agents:
+            _agents[agent] = None
+        self._agents = _agents
+
+    def do_each(self, method_name: str, *args, **kwargs) -> list[Any]:
+        return [getattr(agent, method_name)(*args, **kwargs) for agent in self._agents]
+
+
 
     # Additional methods like union, intersection, difference, etc., can be added as needed.
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -165,7 +165,7 @@ class AgentSet(MutableSet, Sequence):
     def sort(
         self,
         key: Callable[[Agent], Any] | str,
-        reverse: bool = False,
+        ascending: bool = False,
         inplace: bool = False,
     ) -> AgentSet:
         """
@@ -173,7 +173,7 @@ class AgentSet(MutableSet, Sequence):
 
         Args:
             key (Callable[[Agent], Any] | str): A function or attribute name based on which the agents are sorted.
-            reverse (bool, optional): If True, the agents are sorted in descending order. Defaults to False.
+            ascending (bool, optional): If True, the agents are sorted in ascending order. Defaults to False.
             inplace (bool, optional): If True, sorts the agents in the current AgentSet; otherwise, returns a new sorted AgentSet. Defaults to False.
 
         Returns:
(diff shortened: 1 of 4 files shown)


mesa/agent.py
```diff
@@ -57,3 +60,41 @@ def advance(self) -> None:
     @property
     def random(self) -> Random:
         return self.model.random
+
+
+class AgentSet:
+    def __init__(self, agents: set[Agent], model: Model):
+        self._agents = agents
```
quaquel, line 67: Consider turning this into weak refs?

mesa/agent.py
```diff
@@ -57,3 +60,41 @@ def advance(self) -> None:
...
+        self.model = model
```
quaquel, line 68: why do you need a model attribute? I guess for model.random, but then can't you just pass the reference to the seeded random number generator itself?

mesa/agent.py
```diff
@@ -57,3 +60,41 @@ def advance(self) -> None:
...
+        if filter_func is None:
+            return AgentSet(set(self._agents), self.model)
+        return AgentSet(
+            {agent for agent in self._agents if filter_func(agent)}, self.model
+        )
+
+    def shuffle(self) -> AgentSet:
+        shuffled_agents = list(self._agents)
+        self.model.random.shuffle(shuffled_agents)
+        return AgentSet(set(shuffled_agents), self.model)
```
quaquel, line 90: by turning it into a set, you loose the ordering again....

mesa/agent.py
```diff
@@ -57,3 +60,41 @@ def advance(self) -> None:
...
+
+    def sort(self, key: Callable[[Agent], Any], reverse: bool = False) -> AgentSet:
+        sorted_agents = sorted(self._agents, key=key, reverse=reverse)
+        return AgentSet(set(sorted_agents), self.model)
```
quaquel, line 94: see previous point.

It is worth checking if WeakrefDictionary is order preserving, just like a normal dict.

mesa/agent.py
```diff
@@ -57,3 +60,41 @@ def advance(self) -> None:
...
+
+    def do_each(self, method_name: str):
+        for agent in self._agents:
+            getattr(agent, method_name)()
```
quaquel, line 98: add support for args and kwargs here, and make it possible to get returns.

So 

def do_each(self, method_name:str, *args, **kwargs): --> list[any]
    return [getattr(agent, method_name)(*args, **kwargs) for agent in self._agents]

quaquel (commented):
some quick suggestions

EwoutH:
I won’t be working on it between now and out meeting tomorrow, so take it as far as you like. Maybe there is also still some stuff from #1911 that can be ported over.

quaquel:
I'll see what I can do tomorrow morning. I just addressed my comments on your first submission and added the inplace keyword argument.

mesa/agent.py
```diff
@@ -57,3 +61,73 @@ def advance(self) -> None:
...
+        self.model.random.shuffle(shuffled_agents)
+
+        if inplace:
+            self._reorder(shuffled_agents)
+            return self
+        else:
+            return AgentSet(shuffled_agents, self.model)
+
+    def sort(self, key: Callable[[Agent], Any], reverse: bool = False, inplace: bool = False) -> AgentSet:
+        sorted_agents = sorted(list(self._agents.keys()), key=key, reverse=reverse)
```
Corvince, line 114: List is superfluous here

Corvince:
Awesome PR

mesa/agent.py
```diff
@@ -57,3 +61,73 @@ def advance(self) -> None:
...
+            agents = [agent for agent in self._agents.keys() if filter_func(agent)]
+
+            if inplace:
+                self._reorder(agents)
+                return self
+            else:
+                return AgentSet(
+                    agents,
+                    self.model
+                )
```
Corvince, line 101: I think this code can be improved a bit, it's a bit hard to read at the moment. I'm on mobile so please check this again, but here is what ChatGPT made of it
```python
def select(self, filter_func: Callable[[Agent], bool] | None = None, inplace: bool = False) -> AgentSet:
    """
    Select agents based on a filter function.

    Args:
    filter_func: A callable that returns True for agents to be included.
    inplace: If True, modify the current AgentSet; otherwise, return a new one.

    Returns:
    AgentSet: A new or modified AgentSet based on the filter criteria.
    """

    # Handle the case where no filter function is provided
    if filter_func is None:
        return self if inplace else AgentSet(list(self._agents.keys()), self.model)

    # Apply filter function to select agents
    agents = [agent for agent in self._agents.keys() if filter_func(agent)]

    # Modify in place or return a new AgentSet
    if inplace:
        self._reorder(agents)
        return self
    else:
        return AgentSet(agents, self.model)
```

quaquel: chatgpt version does indeed look a bit nicer.

mesa/agent.py
EwoutH: Agreed, thanks for implementing!

mesa/agent.py
EwoutH: Good point. Did you find out if it’s order preserving?

mesa/agent.py
EwoutH: Good catch, thanks for fixing!

mesa/agent.py
EwoutH: Good question. Let me think about it and possible cases.

mesa/agent.py
EwoutH: Thanks for implementing, curious how it looks, will review the implementation tomorrow.

EwoutH:
Thanks for working on It. I will try to do an initial review tomorrow between 9 and 10 CET, otherwise around noon (feel free to continue regardless).

mesa/agent.py
```diff
@@ -63,38 +64,87 @@ def random(self) -> Random:
...
+            _agents[agent] = None
+            _indices.append(weakref.ref(agent))
+        self._agents = _agents
+        self._indices = _indices
+
+    def do_each(self, method_name: str, *args, **kwargs) -> list[Any]:
+        """invoke method on each agent"""
+        return [getattr(agent, method_name)(*args, **kwargs) for agent in self._agents]
+
+    def get_each(self, attr_name: str) -> list[Any]:
```
EwoutH, line 135: Nice idea. We have to think a bit more of useful constructs to access agent data efficiently and flexibly. Maybe we want to (also or optionally) return a dict, so you can check which agent has which value. And maybe we want to do aggregation (min, mean, median, max). There was also an `as_dataframe` idea.

EwoutH:
I’m leaning towards just documenting properly that you can use `agentset.copy()` or `copy(agentset)’ instead of the whole `inplace` thing everywhere. It just seems to add unnecessary complexity to every method while `copy` is a general solution.

quaquel:
I disagree with the copy thing. First, it adds another method call into any chain. Second, the API now follows Pandas, so users of Mesa are likely to be familiar with it. Third, in particular, in the case of a large number of agents, there are substantial memory advantages to using `inplace=True.`.

mesa/agent.py
quaquel: still to be tested, but my hunch is that it will be.

EwoutH:
Thanks for your perspective. Sound arguments, let me further think about it. Also curious what others think.

I would propose all the "getting data back" things (like `get_each`) to be done in a future PR. That could be done together with the redesign of the data collector (see #348).

Corvince:
> Thanks for your perspective. Sound arguments, let me further think about it. Also curious what others think.
> 
> I would propose all the "getting data back" things (like `get_each`) to be done in a future PR. That could be done together with the redesign of the data collector (see #348).

I concur with keeping inplace in the API, since I think copying should be the default. We will see how that turns out performance wise, but we shouldnt worry about that just yet.

I also agree we should keep this PR small at first and add further functions later. This is a very solid basis and can/should be used for lots of things so not overcomplicate things now.

mesa/model.py
```diff
@@ -51,12 +52,22 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         self.running = True
         self.schedule = None
         self.current_id = 0
-        self.agents: defaultdict[type, dict] = defaultdict(dict)
+        self._agents: defaultdict[type, dict] = defaultdict(dict)
+
+    @property
+    def agents(self) -> AgentSet:
+        all_agents = set()
```
Corvince, line 59: I know the discussion about a set have been all over the place, but I thought we create AgentSet specifically to be deterministic. Using a set here destroys this advantage, no? I think we can use some itertools to chain the individual types of the _agents dict and pass that to AgentSet, so we don't need to create an intermediate list

mesa/agent.py
```diff
@@ -41,12 +42,15 @@ def __init__(self, unique_id: int, model: Model) -> None:
         self.model = model
         self.pos: Position | None = None
 
-        # Register the agent with the model using defaultdict
-        self.model.agents[type(self)][self] = None
+        # Directly register the agent with the model
+        if type(self) not in self.model._agents:
+            self.model._agents[type(self)] = set()
```
Corvince, line 47: this should be a dict, right?

mesa/model.py
quaquel: I agree that this should not be a set. I have focussed on AgentSet so far, but I will take a look at this next.

mesa/agent.py
quaquel: I have focussed on AgentSet so far, but I will take a look at this next.

quaquel:
> I also agree we should keep this PR small at first and add further functions later. This is a very solid basis and can/should be used for lots of things so not overcomplicate things now.


Keeping the PR focused is the right way to go. However, it should also be feature-complete, and the basic structure should be solid. Otherwise, the code base keeps changing repeatedly to reflect new ideas. My suggestion would be to focus on making sure AgentSet has the following things:

- Complete behavior of an ordered set 
- inclusion of `shuffle`, `select`, and `sort`
- ensure that you can call methods and retrieve attributes on the agents in the set. So, for me, `do_each` and `get_each` are within scope, although I am unsure about their naming. It also makes sense to ensure `do_each` has a return. This makes `do_each` much more useful across the board.

What can be discussed or kept for later, in my view, are:

- support for listlike indexing. Since `AgentSet` is ordered, so, conceptually, it makes sense to be able to retrieve by index. It also would mean there is no need for a separate `AgentList` class (contra AgentPy). It is possible to add this by implementing `__getitem__`. There are, however, some tricky questions because of the use of weak references. Should the index just refer to whatever is the fifth item in the ordered set, or should it return None or raise an exception if the agent has been removed by the garbage collector?
- the sophistication of `select`. The present implementation is bare bones because it only accepts a boolean callable. So selecting the first _n_ items  is not supported through `select` at the moment. It can however be done through slicing `agents[0:5]`, if indexing is supported
- Adding methods like `to_dataframe` or `as_dataframe`.

EwoutH:
Inheriting from [`MutableSet`](https://docs.python.org/3/library/collections.abc.html#collections.abc.MutableSet) is an interesting approach. Since we're using it as a mixin, do we comply with these notes?

Screenshot_315

Also remarkably, there is a [reference](https://code.activestate.com/recipes/576694/) to an `OrderedSet` recipe in the official Python docs.

EwoutH:
If either of you has anything to add to the Python discussion on `OrderedSets`, please do so. I'm trying to make the case but your probably have more specialized knowledge about this:  https://discuss.python.org/t/add-orderedset-to-stdlib/12730/92

quaquel:
We follow 1, but I am not sure we need to worry about 2 (it involves subset and superset comparisons as far as I can tell); 3 is likewise unnecessary.

While googling this, I came across some discussions on pickling the set. Is this something we want to support?

Corvince:
> > I also agree we should keep this PR small at first and add further functions later. This is a very solid basis and can/should be used for lots of things so not overcomplicate things now.
> 
> Keeping the PR focused is the right way to go. However, it should also be feature-complete, and the basic structure should be solid. Otherwise, the code base keeps changing repeatedly to reflect new ideas. My suggestion would be to focus on making sure AgentSet has the following things:
> 
> * Complete behavior of an ordered set
> * inclusion of `shuffle`, `select`, and `sort`
> * ensure that you can call methods and retrieve attributes on the agents in the set. So, for me, `do_each` and `get_each` are within scope, although I am unsure about their naming. It also makes sense to ensure `do_each` has a return. This makes `do_each` much more useful across the board.

The question is should do_each return the return values the called function or return the AgentSet itself. The latter would allow to keep on chaining, for example:

```python
agents.do_each(step).select(n=10).do_each(move)
```

But getting the return values could also be valueable, so maybe we can also add `agents.apply(func)`, which returns whatever the function returns. Just thinking out loud here.

mesa/agent.py
```diff
@@ -57,3 +63,94 @@ def advance(self) -> None:
...
+
+    def do_each(self, method_name: str, *args, **kwargs) -> list[Any]:
+        """invoke method on each agent"""
+        return [getattr(agent, method_name)(*args, **kwargs) for agent in self._agents]
+
+    def get_each(self, attr_name: str) -> list[Any]:
+        """get attribute value on each agent"""
+        return [getattr(agent, attr_name) for agent in self._agents]
+
+    def __getitem__(self, item: int | slice) -> Agent:
```
Corvince, line 137: Converting all agents to a list seems a bit inefficient for retrieving a single agent

```suggestion
    def __getitem__(self, item: int | slice) -> "Agent":
            if isinstance(item, slice):
                return list(self._agents.keys())[item]
            return next(itertools.islice(self._agents.keys(), item, item+1))
```

quaquel:
Unit tests have been added, and it seems pretty complete.
 
However, there are several outstanding issues:
1. `select` is very barebones because it only takes a callable -> bool. Do we want to support other arguments, such as `n=5` for the first five items?
2. `sort` likewise depends on a callable, so you cannot do a simple sort on attributes. In my view, this is such a common use case that we should support it. It is straightforward to do. if `key` is a string, use `getattr` with the key. 
3. The current class is not pickle-able. This can be an issue if you want to run in parallel and have created AgentSets somewhere during the initialization of the model. 
4.  Naming for `do_each` invokes the given method on each agent in the set, and `get_each` retrieves the given attribute from each agent in the set.

mesa/agent.py
```diff
@@ -57,3 +63,94 @@ def advance(self) -> None:
...
+        return list(self._agents.keys())[item]
```
Corvince, line 138: Converting all agents to a list seems a bit inefficient for retrieving a single agent

```suggestion
        if isinstance(item, slice):
            return list(self._agents.keys())[item]
        return next(itertools.islice(self._agents.keys(), item, item+1))
```

mesa/agent.py
```diff
@@ -57,3 +63,94 @@ def advance(self) -> None:
...
+    def __iter__(self) -> Iterator[Agent]:
+        return self._agents.keys()
+
+    def __contains__(self, agent: Agent) -> bool:
+        """Check if an agent is in the AgentSet."""
+        return agent in self._agents
+
+    def select(self, filter_func: Callable[[Agent], bool] | None = None, inplace: bool = False) -> AgentSet:
+        if filter_func is None:
+            return self if inplace else AgentSet(list(self._agents.keys()), self.model)
```
Corvince, line 90:
```suggestion
            return self if inplace else AgentSet(iter(self), self.model)
```

mesa/agent.py
```diff
@@ -57,3 +63,94 @@ def advance(self) -> None:
...
+            self._reorder(agents)
+            return self
+        else:
+            return AgentSet(
+                agents,
+                self.model
+            )
+
+    def shuffle(self, inplace: bool = False)-> AgentSet:
+        shuffled_agents = list(self._agents.keys())
```
Corvince, line 104:
```suggestion
        shuffled_agents = list(self)
```

quaquel:
> The question is should do_each return the return values the called function or return the AgentSet itself. The latter would allow to keep on chaining, for example:
> 
> ```python
> agents.do_each(step).select(n=10).do_each(move)
> ```
> 
> But getting the return values could also be valueable, so maybe we can also add `agents.apply(func)`, which returns whatever the function returns. Just thinking out loud here.

Good point that do_each will stop chaining which might be undesirable. So having one method that calls a method but returns the agents, and another which calls the method and returns the results might be a nice way to enable both use cases. We could use 'do_each`, 'apply', and 'retrieve', where the latter 2 return the value from the method call or attribute call, while do_each returns the agents itself to allow continued chaining.

Corvince:
> Unit tests have been added, and it seems pretty complete.
> 
> However, there are several outstanding issues:

Thanks for putting these together, great discussion points.

> 
> 1. `select` is very barebones because it only takes a callable -> bool. Do we want to support other arguments, such as `n=5` for the first five items?

I think n would be nice and also selecting by type would be a great addition, since this is what got all this started I think

> 2. `sort` likewise depends on a callable, so you cannot do a simple sort on attributes. In my view, this is such a common use case that we should support it. It is straightforward to do. if `key` is a string, use `getattr` with the key.

I think its a bit of a personal taste, but for sorting by attributes I could imagine a dedicated `sortBy` method.

> 3. The current class is not pickle-able. This can be an issue if you want to run in parallel and have created AgentSets somewhere during the initialization of the model.

This is a downer for me, because I used to love pickling objects for varying use cases. What would be required to make it pickleable? 

> 4. Naming for `do_each` invokes the given method on each agent in the set, and `get_each` retrieves the given attribute from each agent in the set.

I think do_each is pretty good in this context, because it does exactly what the name suggests. get_each could also be called collect_attribute, but I am not sure that is better. But having suggested that a generic collect function could also be nice in a future iteration.

quaquel:
> > 3. The current class is not pickle-able. This can be an issue if you want to run in parallel and have created AgentSets somewhere during the initialization of the model.
> 
> This is a downer for me, because I used to love pickling objects for varying use cases. What would be required to make it pickleable?
> 

It seems that we need to add `__setstate__` and `__getstate__`. See [the docs](https://docs.python.org/3/library/pickle.html#pickling-class-instances)

EwoutH:
Great progress! 
> 1. `select` is very barebones because it only takes a callable -> bool. Do we want to support other arguments, such as `n=5` for the first five items?

It’s the end, the `AgentSet` should include all functionally as proposed in #1905 and implemented in #1911. Doesn’t all have to be with `select`, but should be possible though chaining or slicing.

> 2. `sort` likewise depends on a callable, so you cannot do a simple sort on attributes. In my view, this is such a common use case that we should support it. It is straightforward to do. if `key` is a string, use `getattr` with the key.

Agreed, also aligns with my remark on 1. Should take multiple sorting criteria and directions.

> 3. The current class is not pickle-able. This can be an issue if you want to run in parallel and have created AgentSets somewhere during the initialization of the model.

Already fixed, really nice!

> 4. Naming for `do_each` invokes the given method on each agent in the set, and `get_each` retrieves the given attribute from each agent in the set.

> We could use 'do_each`, 'apply', and 'retrieve', where the latter 2 return the value from the method call or attribute call, while do_each returns the agents itself to allow continued chaining.

I would argue for only one method that asks agents to do something. Otherwise it will just confuse the hell out of everyone. That method can have an optional keyword argument with what it returns. By default I think that should be an `AgentSet`, but we might want to support other types of data (list, maybe dict).

Since we might want to “get” other things or data formats later, I don’t think the method to get an attribute from all agents should be called `get_each`. `get_attribute` or `get_attribute_list` sounds more logical.

For me this will get more clear as we’re going to work on #348.

EwoutH:
Maybe we can just call `do_each` simply `do`. In the `AgentSet`, it might be implied that the methods apply on all agents. Then we can also make multiple variants of `get`.

```Python
agents.do(step).select(n=10).shuffle().do(move)
```

EwoutH:
[`IndexedSet`](https://boltons.readthedocs.io/en/latest/setutils.html#boltons.setutils.IndexedSet) was suggested on discuss.python.org. Functionality wise it looks great, but I don’t know how fast it is since it looks like pure Python.

quaquel:
1. Ok, I'll have a look and make sure it all is covered
2. Ok. I do like `sort_by` as well, but given your later points, I agree that having many methods that do slightly different things is not desirable. My hunch is to just have several kwargs, each with a different clear use case.
3. Was a bit tricky because of weakrefs, but yes it works.
4. I'll probably go with `do` and `get` or `get_attribute`. For do, I'll add a kwarg to control the return.

quaquel:
I believe I have completely covered the various comments and suggestions given so far. In short, `AgentSet` can now be pickled. I have changed the method names to `do` for calling methods on agents and `get` for retrieving attributes. `do` by default returns the agentset itself to enable continued changing: `agents.shuffle().do("stage").do("execute")`. However, you can use a keyword argument to return the results from the method call instead. `get` evidently always returns the results.

`select` and `sort` have some additional optional keyword arguments. `select` now accepts `n=` to select the __n__ first items. This can be used in conjunction with a callable. `sort` now accepts a callable or str. In the case of a str, it is assumed that you want to sort on an attribute, so it gets wrapped into `operator.attrgetter`. 

Selecting by agent type can already be handled by the callable. Moreover, there is also `model.get_agents_of_type()`, which returns an AgentSet. Since `model._agents` stores by agent type, this will likely be faster. However, if there is a need or desire to make `select` work with agent types as well, it would be relatively easy to add this analogous to how sort handles the situation where `key` is a string.

EwoutH:
Okay, I think this is mature enough to get out in the wild. In my opinion we try to move fast and break things, by releasing this with an experimental warning to 2.2.0. Then we can gather feedback while we develop more advanced features on top of the `AgentSet` in the mean time, and aim to release it stable in 2.3.0 (or 3.0.0 if that's next).

What we still need to do in this PR, that I can think of:
- [ ] Make sure tests are covering all current features and use cases
- [ ] Update docstring
- [ ] Update PR description and examples

quaquel:
I have added tests that cover most of the stuff. I haven't tested with-coverage yet to be completely sure, however. 
I added docstrings to the key user-facing methods. 
go ahead on the PR text.

mesa/agent.py
EwoutH: This way it’s explicit we use the same `model.random` thing as elsewhere. Are there disadvantages about including it / advantages of removing?

mesa/agent.py
EwoutH: @quaquel what do you think?

quaquel: This concerns how we want to handle agents disappearing due to the garbage collector. 

I checked various other ordered set and indexed set implementations. They either do something like the current code or have a way of keeping track of changes to the internal dict, which then triggers a rebuilding of the index. I can do some performance tests on the suggested code and my version to see if it makes a big difference.

mesa/agent.py
quaquel: Ok, let's mimic how Agent does it.

mesa/agent.py
quaquel: Just to elaborate a bit on the problem.

Because of weakrefs, agents can magically disappear from `self._agent`. Normal OrderedSet implementations typically have a list of keys, and indexing is done on this. If items are added or removed from the set, the index is updated. Because of weakrefs, we don't fully know when agents are removed from the set. Thus we cannot do the default updating. 

One option is to replace the list typically used with a WeakValueDict, but you cannot conveniently slice on this. You would have to turn the slice into a list of indices and next for each index retrieve the key. Also, you can potentially retrieve None's if the index refers to an agent that has been garbage collected. So, how to handle such None's?

The other option, which I don't know will work, but I can check, is to use weakref. finalizers. In my understanding this allows us to detect when agents are removed by the garbage collector, in turn triggering a rebuilding of the index.

Whether any of this makes a big difference in performance, I also don't know yet.

Corvince: I see, its more complicated than I thought. I would say keep the current code, performance optimization (if needed!) can come later.

Corvince:
> I believe I have completely covered the various comments and suggestions given so far. In short, `AgentSet` can now be pickled. I have changed the method names to `do` for calling methods on agents and `get` for retrieving attributes. `do` by default returns the agentset itself to enable continued chaining: `agents.shuffle().do("stage").do("execute")`. However, you can use a keyword argument to return the results from the method call instead. `get` evidently always returns the results.
> 
> `select` and `sort` have some additional optional keyword arguments. `select` now accepts `n=` to select the **n** first items. This can be used in conjunction with a callable. `sort` now accepts a callable or str. In the case of a str, it is assumed that you want to sort on an attribute, so it gets wrapped into `operator.attrgetter`.

That sounds pretty good to me so far. 
> 
> Selecting by agent type can already be handled by the callable. Moreover, there is also `model.get_agents_of_type()`, which returns an AgentSet. Since `model._agents` stores by agent type, this will likely be faster. However, if there is a need or desire to make `select` work with agent types as well, it would be relatively easy to add this analogous to how sort handles the situation where `key` is a string.

While it is true that it can already be done by the callable, I think a a dedicated type keyword would make sense anyway. The advantage of having this inside the agentset and not only in get_agents_of_type is that we can subselect the agentset by type and not only select all the agents of a specified type. For example if a grid returns an Agentset for the agents in a grid cell and we only want to select a specific type there.

EwoutH:
Thanks for all the work @quaquel! I'm going to clean the last things up and make it ready for review.

mesa/agent.py
EwoutH: Resolving for now

EwoutH:
Todo:
- Easy selection of agents by type in the AgentSet
- Rename `reverse` to something more descriptive like `descending`.

I think by default, sorting should happen from high to low. So `descending=True` as default.

quaquel:
1. done
2. There are two Python defaults. `reverse` is used in default Python sort functions. `ascending` is used in pandas. I deliberately went with the default Python label.

EwoutH:
Sorry, was also already working on it, overwritten your changes.

EwoutH:
@jackiekazil @tpike3 @rht @wang-boyu and others,

Thanks to the hard work of @quaquel and critical eye of @Corvince, I'm quite proud to announce this PR is ready for review! Please look though the code and examples and share your thoughts.

I added the experimental warning in 7888007, and it works beautifully.

Screenshot_317

mesa/model.py
```diff
@@ -51,12 +65,27 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+        self._agents: defaultdict[type, dict] = defaultdict(dict)
+
+        # Warning flags for current experimental features. These make sure a warning is only printed once per model.
+        self.agentset_experimental_warning_given = False
+
+    @property
+    def agents(self) -> AgentSet:
+        """Provides an AgentSet of all agents in the model, combining agents from all types."""
+        all_agents = itertools.chain(
+            *[agents_by_type.keys() for agents_by_type in self._agents.values()]
```
Corvince, line 77: the syntax itertools.chain(*[]) creates a list and immediately unpacks it. It should work better with itertools.chain(()), i.e. use a generator expression instead of a list comprehension

mesa/agent.py
```diff
@@ -57,3 +63,274 @@ def advance(self) -> None:
...
+            filter_func (Callable[[Agent], bool], optional): A function that takes an Agent and returns True if the
+                agent should be included in the result. Defaults to None, meaning no filtering is applied.
+            n (int, optional): The number of agents to select. If 0, all matching agents are selected. Defaults to 0.
+            inplace (bool, optional): If True, modifies the current AgentSet; otherwise, returns a new AgentSet. Defaults to False.
+            agent_type (type[Agent], optional): The class type of the agents to select. Defaults to None, meaning no type filtering is applied.
+
+        Returns:
+            AgentSet: A new AgentSet containing the selected agents, unless inplace is True, in which case the current AgentSet is updated.
+        """
+        agents = list(self._agents.keys())
```
Corvince, line 149: Maybe I am overly cautious with respect to performance implications, but this function creates at worst 4 lists of agents, when we actually don't need any list, but just an iterable to pass to Agentset. I think we could use a generator to only iterate once and only up to n agents

```python
    def agent_generator():
        count = 0
        for agent in self:
            if filter_func and not filter_func(agent):
                continue
            if agent_type and not isinstance(agent, agent_type):
                continue
            yield agent
            count += 1
            if n is not None and count >= n:
                break

    agents = agent_generator()
```

The code might be a bit more difficult to understand, but I also think it might be worth it.

Corvince:
Just two more comments from my side, otherwise I think we are good to go!

mesa/agent.py
quaquel: `if n is not None and count >= n:` should be  `if n and count >= n`

Otherwise, I agree with the suggestions. 

@EwoutH  and me both independently added support for selecting on AgentType. The current implementation adds it as an additional argument, while my implementation replaced filter_func with key, where key was either a callable or agent type. The reason for my suggested implementation was to mimic the sort method signature and partially avoid multiple criteria for selection. I am fine either way.

quaquel:
> Just two more comments from my side, otherwise I think we are good to go!

both have been done.

mesa/agent.py
EwoutH: Did your implementation allow both an AgentType and a callable? Or would the preferred usage be calling `.select()` twice?

quaquel: It was either or, and deliberately so, because of the point made by @Corvince (which is now solved).

In my view, if you have multiple criteria on which to select, it is cleaner to implement them all together into a single selection function. This, by the way would avoid also multiple calls to select, which really is not desirable from a performance point.

EwoutH: Great, we align in that regard. Will mark this as resolved.

mesa/model.py
EwoutH: Great point, it's implemented in 22b99cd.

Corvince (approved):

Corvince:
I approved, but would like to have 2 more eyes on this before I feel comfortable merging. But I think @EwoutH you can start to adapt this in your other prs

EwoutH:
Thanks, I agree on at least another reviewer. No other PRs I have currently open have to be modified based on this PR (we can modify the schedulers, datacollector, and maybe grid in the future).

tpike3:
@EwoutH This is great!!! 

Thanks @Corvince and @quaquel this is truly an amazing effort and incredible discussion.

ankitk50:
@EwoutH , I get this error while running one of the examples which looks related to this change:
```
  File "/Users/ankitkumar/mesa/venvMesa/lib/python3.11/site-packages/reacton/core.py", line 1647, in _render
    root_element = el.component.f(*el.args, **el.kwargs)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ankitkumar/mesa/mesa/mesa/experimental/jupyter_viz.py", line 57, in JupyterViz
    reset_counter = solara.use_reactive(0)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ankitkumar/mesa/venvMesa/lib/python3.11/site-packages/reacton/core.py", line 828, in use_memo
    return rc.use_memo(f, dependencies, debug_name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ankitkumar/mesa/venvMesa/lib/python3.11/site-packages/reacton/core.py", line 1193, in use_memo
    value = f()
            ^^^
  File "/Users/ankitkumar/mesa/mesa/mesa/experimental/jupyter_viz.py", line 52, in make_model
    def make_model():
                ^^^^^^
  File "/Users/ankitkumar/mesa/mesa-examples/examples/boltzmann_wealth_model_experimental/model.py", line 29, in __init__
    a = MoneyAgent(i, self)
        ^^^^^^^^^^^^^^^^^^^
  File "/Users/ankitkumar/mesa/mesa-examples/examples/boltzmann_wealth_model_experimental/model.py", line 53, in __init__
  File "/Users/ankitkumar/mesa/mesa/mesa/agent.py", line 45, in __init__
    self.model.agents[type(self)][self] = None
    ^^^^^^^^^^^^^^^^^
AttributeError: 'BoltzmannWealthModel' object has no attribute 'agents'
```

EwoutH:
Thanks for reporting, that's because the example (for some reason) doesn't run `super().__init__()` in the BoltzmannWealthModel init, and thus the Mesa Model initialization is never ran, and thus `model.agents` never initialized.

It should be fixed in #1928.

quaquel:
I used the BoltzmannWealthModel when testing #1928 so I know that one works fine with #1928.

rht:
(Still catching up with the influx of exciting constructive developments with the new APIs.)

Modifying slightly the example:
```
# Get a specific attribute from all agents
attributes = model.agents.select(...).get('attr_name')
# Or
agent_attrs_df = model.agents.select(...).get('attr_name1', 'attr_name2').to_df()
# Or
agent_attrs_df = to_df(model.agents.select(...).get('attr_name1', 'attr_name2'))
```
As far as I understand, this overlaps with the declarative data collecting specification in https://github.com/projectmesa/mesa/issues/348#issuecomment-1789064622? I find it excessive to have 2 APIs, where one is for agent dynamics description and another for data collection. I am wondering if the 2 APIs are inevitable. The constraint seems to be to make it easy to build complex scheduler and data collection, while to minimize the amount of documentation needed.

rht:
How do I select `model.agents` by agent type in a way that is more performant than `model.agents.select(lambda agent: isinstance(agent, AgentType))`?

quaquel:
1. I agree that any revision of data collection should built on AgentSet to avoid as much as possible having 2 different API's. I have been thinking about this in light also of #348 and #1933. In short, I believe all data collection can be simplified to retrieval of one or more attributes from an object (or collection of objects) and optionally applying a function to it. This is consistent with @Corvince  proposal in #348. So, `target` in this proposal should also accept an AgentSet.
2. There is a seperate method on the mode: `get_agents_of_type`. `model` stores the agents internally (i.e., `_agents`) in a `dict` with `type` as key and `AgentSet` as value.

EwoutH:
Just thought of this now, but their might be one edge-case in which we might break user models: If `model.agents` is already defined in their models. Some model might use `agents` to track a number of agents or some set of them.

PR #1920, opened 2023-12-21
space: Let move_agent choose from multiple positions

EwoutH:
This PR enables the `move_agent` method to take a list of potential positions and offering two selection strategies: random and closest.

This allows any subset of potential positions to be inputted, like an neighborhood, list of empty cells, etc.

Furthermore, it will help with #1898 when cells are selected based on properties.

### New features

1. **Flexible Position Handling**: The `move_agent` method can now accept a single position or a list of possible positions for agent movement.

2. **Selection Strategies**: 
    - `random`: Selects a position randomly from the provided list.
    - `closest`: Chooses the position closest to the agent's current location. In case of ties, a position is selected randomly from the closest ones.

### Usage examples

1. **Moving to a Random Position**:
   ```python
   space.move_agent(agent, possible_positions, selection='random')
   ```

2. **Moving to the Closest Position**:
   ```python
   space.move_agent(agent, possible_positions, selection='closest')
   ```

### Limitations
- This currently only works for Grids, not for ContiniousSpace.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -431,6 +431,54 @@ class _Grid:
         self.remove_agent(agent)
         self.place_agent(agent, pos)
 
+    def move_agent_to_one_of(
+        self,
+        agent: Agent,
+        pos: list[Coordinate],
+        selection: str = "random",
+    ) -> None:
+        """
+        Move an agent to one of the given positions.
+
+        Args:
+            agent: Agent object to move. Assumed to have its current location stored in a 'pos' tuple.
+            pos: List of possible positions.
+            selection: String, either "random" (default) or "closest". If "closest" is selected and multiple
+                       cells are the same distance, one is chosen randomly.
+        """
+        # Handle list of positions
+        if selection == "random":
+            chosen_pos = agent.random.choice(pos)
+        elif selection == "closest":
+            current_pos = agent.pos
+            # Find the closest position without sorting all positions
+            closest_pos = None
+            min_distance = float("inf")
+            for p in pos:
+                distance = self._distance_squared(p, current_pos)
+                if distance < min_distance:
+                    min_distance = distance
+                    closest_pos = p
+            chosen_pos = closest_pos
+        else:
+            raise ValueError(
+                f"Invalid selection method {selection}. Choose 'random' or 'closest'."
+            )
+
+        # Move agent
+        self.move_agent(agent, chosen_pos)
+
+    def _distance_squared(self, pos1: Coordinate, pos2: Coordinate) -> float:
+        """
+        Calculate the squared Euclidean distance between two points for performance.
+        """
+        # Use squared Euclidean distance to avoid sqrt operation
+        dx, dy = abs(pos1[0] - pos2[0]), abs(pos1[1] - pos2[1])
+        if self.torus:
+            dx = min(dx, self.width - dx)
+            dy = min(dy, self.height - dy)
+        return dx**2 + dy**2
+
     def swap_pos(self, agent_a: Agent, agent_b: Agent) -> None:
         """Swap agents positions"""
         agents_no_pos = []
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -436,6 +436,7 @@ class _Grid:
         agent: Agent,
         pos: list[Coordinate],
         selection: str = "random",
+        handle_empty: str | None = None,
     ) -> None:
         """
         Move an agent to one of the given positions.
@@ -445,28 +446,42 @@ class _Grid:
             pos: List of possible positions.
             selection: String, either "random" (default) or "closest". If "closest" is selected and multiple
                        cells are the same distance, one is chosen randomly.
+            handle_empty: String, either "warning", "error" or None (default). If "warning" or "error" is selected
+                          and no positions are given (an empty list), a warning or error is raised respectively.
         """
-        # Handle list of positions
-        if selection == "random":
-            chosen_pos = agent.random.choice(pos)
-        elif selection == "closest":
-            current_pos = agent.pos
-            # Find the closest position without sorting all positions
-            closest_pos = None
-            min_distance = float("inf")
-            for p in pos:
-                distance = self._distance_squared(p, current_pos)
-                if distance < min_distance:
-                    min_distance = distance
-                    closest_pos = p
-            chosen_pos = closest_pos
-        else:
-            raise ValueError(
-                f"Invalid selection method {selection}. Choose 'random' or 'closest'."
-            )
+        # Only move agent if there are positions given (non-empty list)
+        if pos:
+            if selection == "random":
+                chosen_pos = agent.random.choice(pos)
+            elif selection == "closest":
+                current_pos = agent.pos
+                # Find the closest position without sorting all positions
+                closest_pos = None
+                min_distance = float("inf")
+                for p in pos:
+                    distance = self._distance_squared(p, current_pos)
+                    if distance < min_distance:
+                        min_distance = distance
+                        closest_pos = p
+                chosen_pos = closest_pos
+            else:
+                raise ValueError(
+                    f"Invalid selection method {selection}. Choose 'random' or 'closest'."
+                )
+            #  Move agent to chosen position
+            self.move_agent(agent, chosen_pos)
 
-        # Move agent
-        self.move_agent(agent, chosen_pos)
+        # If no positions are given, throw warning/error if selected
+        elif handle_empty == "warning":
+            warn(
+                f"No positions given, could not move agent {agent.unique_id}.",
+                RuntimeWarning,
+                stacklevel=2,
+            )
+        elif handle_empty == "error":
+            raise ValueError(
+                f"No positions given, could not move agent {agent.unique_id}."
+            )
 
     def _distance_squared(self, pos1: Coordinate, pos2: Coordinate) -> float:
         """
--- a/tests/test_space.py
+++ b/tests/test_space.py
@@ -408,6 +408,30 @@ class TestSingleGridTorus(unittest.TestCase):
                 agent, possible_positions, selection="invalid_option"
             )
 
+    def test_move_agent_empty_list(self):
+        agent = self.agents[0]
+        possible_positions = []
+        agent.pos = (3, 3)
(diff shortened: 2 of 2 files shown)


EwoutH:
@jackiekazil @tpike3 @Corvince @rht @wang-boyu @quaquel please review

Corvince:
I definitely like this PR and this is something we should implement. 
However the first thing that comes to my mind is that this would be a breaking change and could potentially really break a lot of models, because I don't think it's uncommon to pass the coordinates as a list of x, y instead of a tuple. It's probably also used in some of the examples. Maybe there is a way to circumvent this issue.

Corvince:
One thing that could be done is to introduce this as a now function that is simply called "move" and at some time deprecate the otherk one.

EwoutH:
> because I don't think it's uncommon to pass the coordinates as a list of x, y instead of a tuple.

Good point, I also notice this in testing.

What if we check if it's a sequence of numbers (sequence of int/float). That way it would catch lists coordinate input, but still allow multiple coordinates (which would be sequences of sequences).

EwoutH:
Implemented in 0e8e42f. None of the test are now needed to be modified.

We should still steer to inputting tuples, it's also type hinted that way. But that can be another effort.

EwoutH:
This can be quite slow for long lists of agents. Could we just check the first element?

Corvince:
That sounds correct. Either the first element is just a number or another iterable.

EwoutH:
Implemented, all tests pass.

We have to think about how to process empty lists of possible positions. Either:
- Don't allow it (by crashing) (current behavior)
- Don't move the agent

Corvince:
Good question. My first instinct would say don't move but issue a warning. I guess most of the time something went wrong when trying to move to an empty list of cells, but crashing the program seems a bit harsh.

EwoutH:
Having thought about it a bit, let’s keep it this way for now. Before this PR, you where required to input a location and your agent would always move. Now, that’s still the case, only multiple is also allowed.

If empty lists become an issue that we can’t solve in usage, we can revisit this.

mesa/space.py
```diff
@@ -419,17 +419,61 @@ def place_agent(self, agent: Agent, pos: Coordinate) -> None:
     def remove_agent(self, agent: Agent) -> None:
         ...
 
-    def move_agent(self, agent: Agent, pos: Coordinate) -> None:
-        """Move an agent from its current position to a new position.
+    def move_agent(
+        self,
+        agent: Agent,
+        pos: Coordinate | list[Coordinate],
+        selection: str = "random",
```
tpike3, line 426: One thought on UX here. 

I am thinking it should be explicit that the user has to make a choice (i.e. `random` or `closest` when they pass in a list of coordinates. I think this will ensure they understand at least part if their movement behavior. 

So maybe the default is `move_agent(agent, user_specified_position, selection ="user_specified") and if the user doesn't specify (i.e. passes in a list of coordinates) then it tells they need to pass in a coordinate or choose `random` or `closest`

 Thoughts?

quaquel: In my view, if pos is a single location, selection becomes irrelevant. So, the default value is `selection=None`. If pos is a collection and `selection=None`,  you raise an error (probably a `ValueError`).

The interesting corner case is a list with 1 position and how to handle this. In my view, you handle this as a list, so `selection` should not be `None`.

EwoutH: Personally, I think random is a good default. You input a list, it chooses one without bias or favor.

I was in doubt of even including closest, but I thought it was an interesting addition for some edge-cases.

tpike3: My vote is to go with `selection=None`  and throw a `ValueError` if they pass a list without changing selection. This ensures the users choices are explicit and understood (as much as possible). 

@EwoutH My assumption is this is not a heavy lift, would you be able to make that update? If my assumption is wrong please let me know or if you feel passionately about not doing this.

EwoutH: Technically, it's an easy change.

I just don't like the boiler plate we're adding, while in my opinion an unbiased (thus random) selection is a very obvious default. If you specify some criteria (like lowest distance), you specify it, if not, it selects without a criteria randomly.

Most of the use cases I can think of, I will already have narrowed down the possible locations I can go, for example to a neighbourhood or all values above a threshold in the property layer. So I want to just pick one of the, without favor. That makes random in my opinion the obvious default choice.

quaquel: But that is only true if the position is a list. If you pass a single position, the keyword argument random is meaningless. Moreover, I'll be happy to debate what is obvious in this case. For you, it might be random, but for others working in other domains, closest might be the obvious. So, the pythonic response, in my view, is to refuse to guess.

EwoutH:
Still thinking about the empty lists case. Currently, if you think you might have no target locations, you have to check manually if this is the case:
```Python
possible_positions = some_selection_function()
if possible_positions >= 1:
  move_agent(agent, possible_positions)
```
Advantages:
- Explicit
- You can do a `else` if no location if found
- Maybe easier to debug

Disatvantages:
- Added boilerplate code

We could add a `allow_no_move` boolean to `move_agent`, which if true allows to input an empty list, which results in the agent not moving (either with or without warning).

However, then we fundamentally change the behavior of `move_agent`: Instead of always moving, we now maybe move the agent.

Another solution could be just to create a new method for these cases, and leave `move_agent` as it originally was. Then `move_agent` requires a single position and you can be sure your agents moves there, while `move_agent_to_one_of` moves an agent to one of a list. If the list if empty, it doesn't move (and we could add a bool for a warning if that happens).

mesa/space.py
EwoutH: In my view, you have two categories of selection:
- Unfavored/undirected: Random
- Favored/directed: Closest, farthest, highest, quietest, etc.

Note that unfavored only has one option: completely random. Any other selection you need some selection criteria. I would argue that unfavored selection a good default, and if you want to introduce favor/direction you have to specify the criteria what you are going to favor.

Maybe we should first discuss if this PR needs to modify `move_agent` or should introduce a new method (see my comment below).

quaquel:
> Another solution could be just to create a new method for these cases, and leave `move_agent` as it originally was. Then `move_agent` requires a single position and you can be sure your agents moves there, while `move_agent_to_one_of` moves an agent to one of a list. If the list if empty, it doesn't move (and we could add a bool for a warning if that happens).

I like the clarity of `move_agent_to_one_off` and keeping it separate from `move`. In this case, you can keep the default behavior as `random`, like you want. I am unsure about the behavior of the following two cases.

1. Empty list; I am inclined to raise an error in this case. You cannot move, so it is an error, and errors should not be passed over silently. It is up to the user to catch the exception and handle it properly.  Clearly, this must be documented well. 
2. list with one item; in principle, you can simply handle this identically to a list larger than one.

mesa/space.py
quaquel: Closest and farthest just require a distance metric. So, if you want to be complete, you probably would need to pass some additional thing(s) in case of non-random selection. 

I agree that the additional method warrants its own discussion.

Corvince: I agree with @Ewouth's viewpoint. When given a list of potential locations, selecting one at random should be the default approach. Why? If the intention was to always pick the first item, there wouldn't be a need for a list. And if specific criteria for selection exist, they must be clearly defined. In scenarios lacking such criteria, a random choice is the most logical. This logic applies even to a list with a single item; selecting randomly from one option inevitably results in that option being chosen.

Regarding an empty list, the concept is straightforward: if you randomly select from an empty list, you end up with no new position, meaning you remain in your current location. It's conceivable to have a system that generates a list of possible positions, which could be empty in some cases.

tpike3: I have found random to be the easy explain away with ABMs but in real life I don't see it, it is typically not random, typically it is "the path of least resistance." (not to mention the endless debates of computer scientists on the "randomness" of different random generators) So I am firmly with @quaquel on this and will emphasize his quote  "So, the pythonic response, in my view, is to refuse to guess."

However we are tied 2 to 2. 

If everyone is ok with it I can ping @jackiekazil Jackie and have her tiebreak.

EwoutH: > If everyone is ok with it I can ping Jackie and have her tiebreak.

Yes, but I would like the discussion if moving from a list (and possibly non-moving on an empty list) should be a separate method or not. Because it will influence my standpoint and it might influence others.

If we go with allowing an empty list input and thus allowing non-movements, I think it should be a separate method.

tpike3: To make sure I understand, @EwoutH your proposal is just to make two separate methods. 
1 - move agent where parameter is the position to move to
2 - move agent where a list is passed in as the argument and then random is the default selection

If I am understanding, I think that is great, I would totally support that. That solves the entire discussion IMHO.

EwoutH: Yes, that’s indeed what I’m proposing.

I will adjust this PR to reflect that.

EwoutH:
Fully refactored the PR. There is now a new method `move_agent_to_one_of` (suggestions for other names are welcome), and the `move_agent` method is unchanged. Docs are updated and tests pass.

Random is the default selection method and by default, empty lists are allowed. You can optionally enable warnings or errors on empty list input.

Curious what everybody thinks!

mesa/space.py
jackiekazil: It seems like this is resolved, and if I understand it correctly I am in agreement. Let me know if I am misunderstanding.

jackiekazil (commented):
Is there an example that demonstrates this functionality change beyond the testing? 
(I am thinking of how people discover this.)

tpike3 (approved):
LGTM! Thanks @EwoutH

tpike3:
> Is there an example that demonstrates this functionality change beyond the testing? (I am thinking of how people discover this.)

@jackiekazil I can add this to the CE tutorial. @EwoutH if you want to add examples to the `How To Guide` when you get the chance this will wrap it up I think

EwoutH:
Thanks for merging! Also nice to see that we can have critical discussions and still keep a certain velocity. It feels like a nice balance and flow in general.

Agreed on the examples. I will try to rebase #1898 soon after Christmas. And then I think we can have an amazing release to start 2024!

Have a great Christmas everyone!

PR #1926, opened 2023-12-23
Reimplement schedulers to use AgentSet

quaquel:
This PR builds on #1916 and updates the schedulers to use the new AgentSet class and/or weakrefs. It also adds a new feature: schedulers take an optional agents keyword argument in `__init__` so you can spawn a scheduler with all agents. 

My aim with this PR is to be fully backward compatible but raise warnings when methods that have become redundant are being used. The current code passes all unit tests.

Still to do:

- [x] - update docs
- [x] - check type annotations
- [x] - double check to ensure backward compatible behavior
- [x] - code formatting
- [x] - add some additional tests

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -90,6 +90,7 @@ class AgentSet(MutableSet, Sequence):
         interactions with the model's environment and other agents.The implementation uses a WeakKeyDictionary to store agents,
         which means that agents not referenced elsewhere in the program may be automatically removed from the AgentSet.
     """
+    agentset_experimental_warning_given = False
 
     def __init__(self, agents: Iterable[Agent], model: Model):
         """
@@ -101,8 +102,8 @@ class AgentSet(MutableSet, Sequence):
         """
         self.model = model
 
-        if not self.model.agentset_experimental_warning_given:
-            self.model.agentset_experimental_warning_given = True
+        if not self.__class__.agentset_experimental_warning_given:
+            self.__class__.agentset_experimental_warning_given = True
             warnings.warn(
                 "The AgentSet is experimental. It may be changed or removed in any and all future releases, including patch releases.\n"
                 "We would love to hear what you think about this new feature. If you have any thoughts, share them with us here: https://github.com/projectmesa/mesa/discussions/1919",
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -90,6 +90,7 @@ class AgentSet(MutableSet, Sequence):
         interactions with the model's environment and other agents.The implementation uses a WeakKeyDictionary to store agents,
         which means that agents not referenced elsewhere in the program may be automatically removed from the AgentSet.
     """
+
     agentset_experimental_warning_given = False
 
     def __init__(self, agents: Iterable[Agent], model: Model):
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -61,7 +61,13 @@ class BaseScheduler:
     """
 
     def __init__(self, model: Model, agents: Iterable[Agent] | None = None) -> None:
-        """Create a new, empty BaseScheduler."""
+        """Create a new BaseScheduler.
+
+        Args:
+            model (Model): The model to which the schedule belongs
+            agents (Iterable[Agent], None, optional): An iterable of agents who are controlled by the schedule
+
+        """
         self.model = model
         self.steps = 0
         self.time: TimeT = 0
@@ -226,11 +232,12 @@ class StagedActivation(BaseScheduler):
         """Create an empty Staged Activation schedule.
 
         Args:
-            - model: Model object associated with the schedule.
-            - stage_list: List of strings of names of stages to run, in the
+            model (Model): The model to which the schedule belongs
+            agents (Iterable[Agent], None, optional): An iterable of agents who are controlled by the schedule
+            stage_list (:obj:`list` of :obj:`str`): List of strings of names of stages to run, in the
                          order to run them in.
-            - shuffle: If True, shuffle the order of agents each step.
-            - shuffle_between_stages: If True, shuffle the agents after each
+            shuffle (bool, optional): If True, shuffle the order of agents each step.
+            shuffle_between_stages (bool, optional): If True, shuffle the agents after each
                                     stage; otherwise, only shuffle at the start
                                     of each step.
         """
@@ -281,6 +288,12 @@ class RandomActivationByType(BaseScheduler):
 
     def __init__(self, model: Model, agents: Iterable[Agent] | None = None) -> None:
         super().__init__(model, agents)
+        """
+
+        Args:
+            model (Model): The model to which the schedule belongs
+            agents (Iterable[Agent], None, optional): An iterable of agents who are controlled by the schedule
+        """
 
         # can't be a defaultdict because we need to pass model to AgentSet
         self.agents_by_type: [type, AgentSet] = {}
@@ -399,8 +412,14 @@ class DiscreteEventScheduler(BaseScheduler):
     """
 
     def __init__(self, model: Model, time_step: TimeT = 1) -> None:
+        """
+
+        Args:
+            model (Model): The model to which the schedule belongs
+            time_step (TimeT): The fixed time step between steps
+
+        """
         super().__init__(model)
-        # TODO:: should go to weakrefs as well...
         self.event_queue: list[tuple[TimeT, float, weakref.ref]] = []
         self.time_step: TimeT = time_step  # Fixed time period for each step
 
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -249,8 +249,7 @@ class StagedActivation(BaseScheduler):
             else:
                 self.do_each(stage, shuffle=shuffle)
 
-            if self.shuffle_between_stages:
-                shuffle = False
+            shuffle = self.shuffle_between_stages
             self.time += self.stage_time
 
         self.steps += 1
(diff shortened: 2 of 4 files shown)


EwoutH:
Looks like a nice start!

I think a big conceptual design decision we need to make is if we want to support multiple schedulers. Did we do so in the past?

If we only have one scheduler, we could simplify and assume a lot, for example that all agents are in the scheduler. We don't have to explicitly add or remove them, or even keep track of them in the scheduler. There would also be one single step/time clock. It would probably allow deprecating all schedulers, and moving a lot of the requesting of behavior to `AgentSet.do()`. It would be a flexible approach with lot possibilities, like described in #1912.

However, if we do want to support multiple schedulers, we need to leave a lot of stuff in the scheduler, at least the time of that scheduler and which agents are part of it. Personally, I can't try of that many cases you can't solve with the AgentSet approach, but there might be any.

quaquel:
I have used multiple schedulers in the past (pre `RandomActivationByType`), and more generally, I like its flexibility. Also, this PR's only aim is to keep the current Schedulers up to date by using AgentSet. Discussing the future of schedulers is beyond the scope of this PR. 

As indicated in #1912, one possible direction could be to shift to a central EventList. This creates a central understanding of time. All existing functionality of the various schedulers could be rebuilt on such an event list.

mesa/time.py
```diff
@@ -74,20 +81,28 @@ def add(self, agent: Agent) -> None:
...
         """Remove all instances of a given agent from the schedule.
 
         Args:
             agent: An agent object.
         """
-        del self._agents[agent.unique_id]
+        if not self._remove_warning_given:
+            self._remove_warning_given = True
+        warnings.warn(
+            "Because of the shift to using weakrefs, it is no longer needed to explicitly remove"
```
EwoutH, line 99: I think this holds up as long a you use a single scheduler, and assume agents in model = agents in schedulers. However, as you let that assumption go you might want to remove agents from a scheduler but not from a model.

EwoutH: However, I find it weird you still have to explicitly add agents to the scheduler, but don't explicitly have to remove them.

quaquel: This comes back to having one or more than one scheduler. If you have only one, then yes, there is no need to remove them. But that would be a backward incompatible change and thus is outside the scope of this PR.

quaquel:
I have reviewed the code several times, fixed the docs, and tested against various examples. This is ready for a proper check. @EwoutH, @Corvince, @jackiekazil, or @tpike3 do you have a chance to check this?

Corvince:
Probably not in the coming days, but I really look forward to review this!

EwoutH:
Will try to review it later today!

mesa/time.py
```diff
@@ -60,12 +60,19 @@ class BaseScheduler:
         - agents (property): Returns a list of all agent instances.
     """
 
-    def __init__(self, model: Model) -> None:
+    def __init__(self, model: Model, agents: Iterable[Agent] | None = None) -> None:
```
EwoutH, line 63: Does this also mean you can now initialize an scheduler with an existing `AgentSet`? If so, great feature, please update the docs that you can now do this.

mesa/time.py
```diff
@@ -102,25 +112,31 @@ def get_agent_count(self) -> int:
         return len(self._agents)
 
     @property
-    def agents(self) -> list[Agent]:
-        return list(self._agents.values())
+    def agents(self) -> AgentSet:
+        # a bit dirty, but returns a copy of the internal agent set
+        return self._agents.select()
```
EwoutH, line 117: This is indeed the way to do it for backwards compatibility. I don't like that `model.agents` is now an `AgentSet` and `schedule.agents` is a list (this PR doesn't change that). We should discuss how that align this in the long term.

mesa/time.py
```diff
@@ -102,25 +112,31 @@ def get_agent_count(self) -> int:
...
+
+        if not self._agents_key_warning_given:
+            self._agents_key_warning_given = True
+            warnings.warn(
+                "Because of the shift to using weakrefs, this method will be removed in a future version",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+
+        agent_keys = [agent.unique_id for agent in self._agents]
```
EwoutH, line 131: Agreed on deprecating in this case.

However, while in this way it's probably fine, we're technically deprecating something for which the replacement (AgentSet) is still experimental. Might be something to keep in mind.

mesa/time.py
```diff
@@ -202,6 +218,7 @@ class StagedActivation(BaseScheduler):
     def __init__(
         self,
         model: Model,
+        agents: Iterable[Agent] | None = None,
```
EwoutH, line 221: Document that Agents is now an input.

mesa/time.py
```diff
@@ -265,9 +279,18 @@ class RandomActivationByType(BaseScheduler):
         - get_type_count: Returns the count of agents of a specific type.
     """
 
-    def __init__(self, model: Model) -> None:
-        super().__init__(model)
-        self.agents_by_type = defaultdict(dict)
+    def __init__(self, model: Model, agents: Iterable[Agent] | None = None) -> None:
```
EwoutH, line 282: Document agents input

EwoutH (commented):
Looks good in general, and agree with the intend of this PR.

However, I think that we only should make minimal changes until we decide which direction to go with the schedulers (None, one per model, or multiple). So while this PR is good, I think the next step is a broad discussion about how we see the future of activating agents to do things.

mesa/time.py
quaquel: Again, this is outside the scope of this PR, but I want to do a separate discussion on unique_id. Now that it is no longer needed for the schedulers, its only function is to identify agents when gathering statistics.

mesa/time.py
quaquel: done

mesa/time.py
quaquel: done

mesa/time.py
quaquel: `AgentSet.select` returns an AgentSet, not a list. So, this PR already gives the desired alignment.

Because AgentSet is a Sequence, it should be backward compatible with places were lists are currently being used.

EwoutH (approved):

mesa/time.py
EwoutH: Wait but then we have a behavior change, right? That makes this PR a little bit less backwards compatible I think.

quaquel: Not really, because AgentSet behaves like a list. You can get its length, index it, slice it, and iterate over it. That is why I made sure AgentSet implements all methods of Sequence.

mesa/agent.py
```diff
@@ -91,6 +91,8 @@ class AgentSet(MutableSet, Sequence):
         which means that agents not referenced elsewhere in the program may be automatically removed from the AgentSet.
     """
 
+    agentset_experimental_warning_given = False
```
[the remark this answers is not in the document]
quaquel: The question is how to handle warnings in general. The problem this fixed was that I ran into a circular reference between model.py and agent.py. Perhaps warning flags should be moved into a separate util.py to have a general place for them.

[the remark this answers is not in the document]
EwoutH: I think in the class itself (where it's now) is okay. It's relatively close to the warning message itself and it makes sure your CLI isn't flooded.

EwoutH:
@quaquel What's your guidance on this regarding the next release. Would you recommend having this in the 2.2.0 release, or to release 2.2.0 first and merge this on the 2.3 development branch?

mesa/time.py
jackiekazil: Agree - outside of this pr - suggested place for the discussion?

jackiekazil:
@quaquel  

RE: Release -- would like to hear your thoughts on the release integration mentioned by @EwoutH. 
Re: scheduler - +1 to what you said -- multiple schedulers is a requirements. 

RE: Review -- I looked over the code this evening. I will do a test run tomorrow evening.

mesa/time.py
quaquel: Given that #1930 also relies on unique_id, we can perhaps discuss it there. Alternatively, we can have a broader discussion on the future of data collection and make it part of that.

quaquel:
1. @EwoutH : Ideally, make it part of 2.2
2. @jackiekazil:  I will look at the release approach, but given that it is modeled on the one we use for the ema workbench, I probably like it.
3. @jackiekazil: thanks for the clarity on having multiple schedulers as a requirement. This does, however, leads into a next question: what tracks the global simulation time?

tpike3 (approved):
LGTM! Thanks at @quaquel I think this really highlights a new and positive path for Mesa.

EwoutH:
Unfortunately, this PR, together with an implementation error in https://github.com/projectmesa/mesa/pull/1916, broke RandomActivation, and more important, went completely unnoticed for multiple weeks, until an user reported it in https://github.com/projectmesa/mesa/issues/2006. We should evaluate very critically how such a major unintended behavior could slip through the cracks.

quaquel:
1. we did not test AgentSet.shuffle properly. 
2. We moved the schedulers to AgenSet for 2.2. We should have waited with this for 2.3 (although I doubt we would have seen these bugs before then)

EwoutH:
(and thus the next questions are, should we alter our review process, testing standards or something else, and if so, how)

quaquel:
See #2007. The fix can be found there.

PR #1937, opened 2024-01-06
model: Initialize _agents attribute on __new__

rht:
This is to avoid the warning:
```
/projectmesa/mesa-examples/examples/boltzmann_wealth_model_experimental/model.py:53: FutureWarning: The Mesa Model class wasn’t initialized. In the future, you need to explicitly initialize the Model by calling super().__init__() on initialization.
```
So that users won't have to call `super().__init__()` in their model code.


EwoutH:
Looks useful, thanks! I don’t have the Python expertise to judge the code changes, but high-level, I approve!

<hr>

Not necessary for this PR, but maybe we should also consider on a broader spectrum if we want to require `super().__init__()` calls on initializing a subclass of the Agent or the Model. Can’t we just move everything to `__new__`? I started a discussion on this here:
- #1938

PR #1942, opened 2024-01-07
refactor: Remove dependence on model.schedule, add clock to Model

rht:
The current AgentSet API allows users to define the structure and order of agent actions without initializing a scheduler object. However, a scheduler object is currently necessary until this pull request addresses the issue by incorporating the steps and time attributes directly into the model (which were previously only tracked in the scheduler).

A complication arises for users who still use the scheduler object, as they now need to manually specify `model.advance_time()` within the model's `step()` function. To resolve this, the pull request proposes a solution with assistance from ChatGPT, involving the injection of `model.advance_time()` into the scheduler's `step()` function. This modification aims to streamline the process for users and enhance the overall functionality of the AgentSet API.

### DIFF
--- a/mesa/datacollection.py
+++ b/mesa/datacollection.py
@@ -178,7 +178,7 @@ class DataCollector:
         rep_funcs = self.agent_reporters.values()
 
         def get_reports(agent):
-            _prefix = (agent.model.steps_, agent.unique_id)
+            _prefix = (agent.model._steps, agent.unique_id)
             reports = tuple(rep(agent) for rep in rep_funcs)
             return _prefix + reports
 
@@ -205,7 +205,7 @@ class DataCollector:
 
         if self.agent_reporters:
             agent_records = self._record_agents(model)
-            self._agent_records[model.steps_] = list(agent_records)
+            self._agent_records[model._steps] = list(agent_records)
 
     def add_table_row(self, table_name, row, ignore_missing=False):
         """Add a row dictionary to a specific table.
--- a/mesa/datacollection.py
+++ b/mesa/datacollection.py
@@ -14,8 +14,7 @@ name.
 
 When the collect() method is called, each model-level function is called, with
 the model as the argument, and the results associated with the relevant
-variable. Then the agent-level functions are called on each agent in the model
-scheduler.
+variable. Then the agent-level functions are called on each agent.
 
 Additionally, other objects can write directly to tables by passing in an
 appropriate dictionary object for a table row.
@@ -30,8 +29,7 @@ The DataCollector then stores the data it collects in dictionaries:
 Finally, DataCollector can create a pandas DataFrame from each collection.
 
 The default DataCollector here makes several assumptions:
-    * The model has a schedule object called 'schedule'
-    * The schedule has an agent list called agents
+    * The model has an agent list called agents
     * For collecting agent-level variables, agents must have a unique_id
 """
 import contextlib
@@ -67,7 +65,7 @@ class DataCollector:
 
         Model reporters can take four types of arguments:
         1. Lambda function:
-           {"agent_count": lambda m: m.schedule.get_agent_count()}
+           {"agent_count": lambda m: len(m.agents)}
         2. Method of a class/instance:
            {"agent_count": self.get_agent_count} # self here is a class instance
            {"agent_count": Model.get_agent_count} # Model here is a class
@@ -180,11 +178,11 @@ class DataCollector:
         rep_funcs = self.agent_reporters.values()
 
         def get_reports(agent):
-            _prefix = (agent.model.schedule.steps, agent.unique_id)
+            _prefix = (agent.model.steps, agent.unique_id)
             reports = tuple(rep(agent) for rep in rep_funcs)
             return _prefix + reports
 
-        agent_records = map(get_reports, model.schedule.agents)
+        agent_records = map(get_reports, model.agents)
         return agent_records
 
     def collect(self, model):
@@ -207,7 +205,7 @@ class DataCollector:
 
         if self.agent_reporters:
             agent_records = self._record_agents(model)
-            self._agent_records[model.schedule.steps] = list(agent_records)
+            self._agent_records[model.steps] = list(agent_records)
 
     def add_table_row(self, table_name, row, ignore_missing=False):
         """Add a row dictionary to a specific table.
--- a/mesa/datacollection.py
+++ b/mesa/datacollection.py
@@ -182,7 +182,10 @@ class DataCollector:
             reports = tuple(rep(agent) for rep in rep_funcs)
             return _prefix + reports
 
-        agent_records = map(get_reports, model.agents)
+        agent_records = map(
+            get_reports,
+            model.schedule.agents if hasattr(model, "schedule") else model.agents,
+        )
         return agent_records
 
     def collect(self, model):
--- a/mesa/datacollection.py
+++ b/mesa/datacollection.py
@@ -178,7 +178,7 @@ class DataCollector:
         rep_funcs = self.agent_reporters.values()
 
         def get_reports(agent):
-            _prefix = (agent.model._steps, agent.unique_id)
+            _prefix = (agent.model.steps_, agent.unique_id)
             reports = tuple(rep(agent) for rep in rep_funcs)
             return _prefix + reports
 
@@ -205,7 +205,7 @@ class DataCollector:
 
         if self.agent_reporters:
             agent_records = self._record_agents(model)
-            self._agent_records[model._steps] = list(agent_records)
+            self._agent_records[model.steps_] = list(agent_records)
 
     def add_table_row(self, table_name, row, ignore_missing=False):
         """Add a row dictionary to a specific table.
--- a/mesa/datacollection.py
+++ b/mesa/datacollection.py
@@ -178,7 +178,7 @@ class DataCollector:
         rep_funcs = self.agent_reporters.values()
 
         def get_reports(agent):
-            _prefix = (agent.model.steps, agent.unique_id)
+            _prefix = (agent.model._steps, agent.unique_id)
             reports = tuple(rep(agent) for rep in rep_funcs)
             return _prefix + reports
 
@@ -205,7 +205,7 @@ class DataCollector:
 
         if self.agent_reporters:
             agent_records = self._record_agents(model)
-            self._agent_records[model.steps] = list(agent_records)
+            self._agent_records[model._steps] = list(agent_records)
 
     def add_table_row(self, table_name, row, ignore_missing=False):
         """Add a row dictionary to a specific table.
--- a/mesa/experimental/jupyter_viz.py
+++ b/mesa/experimental/jupyter_viz.py
@@ -181,7 +181,7 @@ def ModelController(model, play_interval, current_step, reset_counter):
     def do_step():
         model.step()
         previous_step.value = current_step.value
-        current_step.value += 1
+        current_step.value = model.steps
 
     def do_play():
         model.running = True
--- a/mesa/experimental/jupyter_viz.py
+++ b/mesa/experimental/jupyter_viz.py
@@ -181,7 +181,7 @@ def ModelController(model, play_interval, current_step, reset_counter):
     def do_step():
         model.step()
         previous_step.value = current_step.value
-        current_step.value = model.steps
+        current_step.value = model.steps_
 
     def do_play():
         model.running = True
--- a/mesa/experimental/jupyter_viz.py
+++ b/mesa/experimental/jupyter_viz.py
@@ -181,7 +181,7 @@ def ModelController(model, play_interval, current_step, reset_counter):
     def do_step():
         model.step()
         previous_step.value = current_step.value
-        current_step.value = model.steps_
+        current_step.value = model._steps
 
     def do_play():
         model.running = True
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -13,10 +13,12 @@ import warnings
 from collections import defaultdict
 
 # mypy
-from typing import Any
+from typing import Any, Union
 
 from mesa.agent import Agent, AgentSet
 from mesa.datacollection import DataCollector
+
+TimeT = Union[float, int]
 
 
 class Model:
@@ -68,6 +70,9 @@ class Model:
         self.current_id = 0
         self.agents_: defaultdict[type, dict] = defaultdict(dict)
 
+        self.steps = 0
+        self.time: TimeT = 0  # the model's clock
+
         # Warning flags for current experimental features. These make sure a warning is only printed once per model.
         self.agentset_experimental_warning_given = False
 
@@ -111,6 +116,11 @@ class Model:
 
     def step(self) -> None:
         """A single step. Fill in here."""
+
+    def advance_time(self, deltat: TimeT = 1):
+        """Increment the model's steps counter and clock."""
+        self.steps += 1
+        self.time += deltat
 
     def next_id(self) -> int:
         """Return the next unique ID for agents, increment current_id"""
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -70,8 +70,8 @@ class Model:
         self.current_id = 0
         self.agents_: defaultdict[type, dict] = defaultdict(dict)
 
-        self.steps_: int = 0
-        self.time_: TimeT = 0  # the model's clock
+        self._steps: int = 0
+        self._time: TimeT = 0  # the model's clock
 
         # Warning flags for current experimental features. These make sure a warning is only printed once per model.
         self.agentset_experimental_warning_given = False
@@ -119,8 +119,8 @@ class Model:
 
     def _advance_time(self, deltat: TimeT = 1):
         """Increment the model's steps counter and clock."""
-        self.steps_ += 1
-        self.time_ += deltat
+        self._steps += 1
+        self._time += deltat
 
     def next_id(self) -> int:
         """Return the next unique ID for agents, increment current_id"""
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -71,7 +71,7 @@ class Model:
         self.agents_: defaultdict[type, dict] = defaultdict(dict)
 
         self._steps: int = 0
-        self._time: TimeT = 0  # the model's clock
+        self.time_: TimeT = 0  # the model's clock
 
         # Warning flags for current experimental features. These make sure a warning is only printed once per model.
         self.agentset_experimental_warning_given = False
@@ -120,7 +120,7 @@ class Model:
     def _advance_time(self, deltat: TimeT = 1):
         """Increment the model's steps counter and clock."""
         self._steps += 1
-        self._time += deltat
+        self.time_ += deltat
 
     def next_id(self) -> int:
         """Return the next unique ID for agents, increment current_id"""
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -70,8 +70,8 @@ class Model:
         self.current_id = 0
         self.agents_: defaultdict[type, dict] = defaultdict(dict)
 
-        self.steps = 0
-        self.time: TimeT = 0  # the model's clock
+        self._steps = 0
+        self._time: TimeT = 0  # the model's clock
 
         # Warning flags for current experimental features. These make sure a warning is only printed once per model.
         self.agentset_experimental_warning_given = False
@@ -119,8 +119,8 @@ class Model:
 
     def advance_time(self, deltat: TimeT = 1):
         """Increment the model's steps counter and clock."""
-        self.steps += 1
-        self.time += deltat
+        self._steps += 1
+        self._time += deltat
 
     def next_id(self) -> int:
         """Return the next unique ID for agents, increment current_id"""
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -72,8 +72,6 @@ class Model:
 
         self.steps_: int = 0
         self.time_: TimeT = 0  # the model's clock
-        self._original_step = self.step
-        self.step = self._wrapped_step
 
         # Warning flags for current experimental features. These make sure a warning is only printed once per model.
         self.agentset_experimental_warning_given = False
@@ -118,11 +116,6 @@ class Model:
 
     def step(self) -> None:
         """A single step. Fill in here."""
-
-    def _wrapped_step(self):
-        """Wrapper for the step method to include time and step updating."""
-        self._original_step()
-        self._advance_time()
 
     def _advance_time(self, deltat: TimeT = 1):
         """Increment the model's steps counter and clock."""
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -117,7 +117,7 @@ class Model:
     def step(self) -> None:
         """A single step. Fill in here."""
 
-    def advance_time(self, deltat: TimeT = 1):
+    def _advance_time(self, deltat: TimeT = 1):
         """Increment the model's steps counter and clock."""
         self._steps += 1
         self._time += deltat
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -72,6 +72,8 @@ class Model:
 
         self.steps_: int = 0
         self.time_: TimeT = 0  # the model's clock
+        self._original_step = self.step
+        self.step = self._wrapped_step
 
         # Warning flags for current experimental features. These make sure a warning is only printed once per model.
         self.agentset_experimental_warning_given = False
@@ -116,6 +118,11 @@ class Model:
 
     def step(self) -> None:
         """A single step. Fill in here."""
+
+    def _wrapped_step(self):
+        """Wrapper for the step method to include time and step updating."""
+        self._original_step()
+        self._advance_time()
 
     def _advance_time(self, deltat: TimeT = 1):
         """Increment the model's steps counter and clock."""
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -70,7 +70,7 @@ class Model:
         self.current_id = 0
         self.agents_: defaultdict[type, dict] = defaultdict(dict)
 
-        self._steps = 0
+        self._steps: int = 0
         self._time: TimeT = 0  # the model's clock
 
         # Warning flags for current experimental features. These make sure a warning is only printed once per model.
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -70,7 +70,7 @@ class Model:
         self.current_id = 0
         self.agents_: defaultdict[type, dict] = defaultdict(dict)
 
-        self._steps: int = 0
+        self.steps_: int = 0
         self.time_: TimeT = 0  # the model's clock
 
         # Warning flags for current experimental features. These make sure a warning is only printed once per model.
@@ -119,7 +119,7 @@ class Model:
 
     def _advance_time(self, deltat: TimeT = 1):
         """Increment the model's steps counter and clock."""
-        self._steps += 1
+        self.steps_ += 1
         self.time_ += deltat
 
     def next_id(self) -> int:
--- a/tests/test_batch_run.py
+++ b/tests/test_batch_run.py
@@ -85,7 +85,6 @@ class MockModel(Model):
     def step(self):
         self.datacollector.collect(self)
         self.schedule.step()
-        self._advance_time()
 
 
 def test_batch_run():
--- a/tests/test_batch_run.py
+++ b/tests/test_batch_run.py
@@ -85,6 +85,7 @@ class MockModel(Model):
     def step(self):
         self.datacollector.collect(self)
         self.schedule.step()
+        self.advance_time()
 
 
 def test_batch_run():
--- a/tests/test_batch_run.py
+++ b/tests/test_batch_run.py
@@ -85,7 +85,7 @@ class MockModel(Model):
     def step(self):
         self.datacollector.collect(self)
         self.schedule.step()
-        self.advance_time()
+        self._advance_time()
 
 
 def test_batch_run():
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -114,8 +114,6 @@ class BaseScheduler:
         self.do_each("step")
         self.steps += 1
         self.time += 1
-        if self.model.steps_ < self.steps:
-            self.model._advance_time()
 
     def get_agent_count(self) -> int:
         """Returns the current number of agents in the queue."""
@@ -173,8 +171,6 @@ class RandomActivation(BaseScheduler):
         self.do_each("step", shuffle=True)
         self.steps += 1
         self.time += 1
-        if self.model.steps_ < self.steps:
-            self.model._advance_time()
 
 
 class SimultaneousActivation(BaseScheduler):
@@ -203,8 +199,6 @@ class SimultaneousActivation(BaseScheduler):
         self.do_each("advance")
         self.steps += 1
         self.time += 1
-        if self.model.steps_ < self.steps:
-            self.model._advance_time()
 
 
 class StagedActivation(BaseScheduler):
@@ -268,8 +262,6 @@ class StagedActivation(BaseScheduler):
             self.time += self.stage_time
 
         self.steps += 1
-        if self.model.steps_ < self.steps:
-            self.model._advance_time()
 
 
 class RandomActivationByType(BaseScheduler):
@@ -370,8 +362,6 @@ class RandomActivationByType(BaseScheduler):
             self.step_type(agent_class, shuffle_agents=shuffle_agents)
         self.steps += 1
         self.time += 1
-        if self.model.steps_ < self.steps:
-            self.model._advance_time()
 
     def step_type(self, agenttype: type[Agent], shuffle_agents: bool = True) -> None:
         """
@@ -497,8 +487,6 @@ class DiscreteEventScheduler(BaseScheduler):
         # After processing events, advance time by the time_step
         self.time = end_time
         self.steps += 1
-        if self.model.steps_ < self.steps:
-            self.model._advance_time()
 
     def get_next_event_time(self) -> TimeT | None:
         """Returns the time of the next scheduled event."""
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -108,7 +108,6 @@ class BaseScheduler:
             agent: An agent object.
         """
         self._agents.remove(agent)
-        agent.remove()
 
     def step(self) -> None:
         """Execute the step of all the agents, one at a time."""
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -114,7 +114,7 @@ class BaseScheduler:
         self.do_each("step")
         self.steps += 1
         self.time += 1
-        if self.model.steps < self.steps:
+        if self.model._steps < self.steps:
             self.model.advance_time()
 
     def get_agent_count(self) -> int:
@@ -173,7 +173,7 @@ class RandomActivation(BaseScheduler):
         self.do_each("step", shuffle=True)
         self.steps += 1
         self.time += 1
-        if self.model.steps < self.steps:
+        if self.model._steps < self.steps:
             self.model.advance_time()
 
 
@@ -203,7 +203,7 @@ class SimultaneousActivation(BaseScheduler):
         self.do_each("advance")
         self.steps += 1
         self.time += 1
-        if self.model.steps < self.steps:
+        if self.model._steps < self.steps:
             self.model.advance_time()
 
 
@@ -268,7 +268,7 @@ class StagedActivation(BaseScheduler):
             self.time += self.stage_time
 
         self.steps += 1
-        if self.model.steps < self.steps:
+        if self.model._steps < self.steps:
             self.model.advance_time()
 
 
@@ -370,7 +370,7 @@ class RandomActivationByType(BaseScheduler):
             self.step_type(agent_class, shuffle_agents=shuffle_agents)
         self.steps += 1
         self.time += 1
-        if self.model.steps < self.steps:
+        if self.model._steps < self.steps:
             self.model.advance_time()
 
     def step_type(self, agenttype: type[Agent], shuffle_agents: bool = True) -> None:
@@ -497,7 +497,7 @@ class DiscreteEventScheduler(BaseScheduler):
         # After processing events, advance time by the time_step
         self.time = end_time
         self.steps += 1
-        if self.model.steps < self.steps:
+        if self.model._steps < self.steps:
             self.model.advance_time()
 
     def get_next_event_time(self) -> TimeT | None:
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -115,7 +115,7 @@ class BaseScheduler:
         self.steps += 1
         self.time += 1
         if self.model._steps < self.steps:
-            self.model.advance_time()
+            self.model._advance_time()
 
     def get_agent_count(self) -> int:
         """Returns the current number of agents in the queue."""
@@ -174,7 +174,7 @@ class RandomActivation(BaseScheduler):
         self.steps += 1
         self.time += 1
         if self.model._steps < self.steps:
-            self.model.advance_time()
+            self.model._advance_time()
 
 
 class SimultaneousActivation(BaseScheduler):
@@ -204,7 +204,7 @@ class SimultaneousActivation(BaseScheduler):
         self.steps += 1
         self.time += 1
         if self.model._steps < self.steps:
-            self.model.advance_time()
+            self.model._advance_time()
 
 
 class StagedActivation(BaseScheduler):
@@ -269,7 +269,7 @@ class StagedActivation(BaseScheduler):
 
         self.steps += 1
         if self.model._steps < self.steps:
-            self.model.advance_time()
+            self.model._advance_time()
 
 
 class RandomActivationByType(BaseScheduler):
@@ -371,7 +371,7 @@ class RandomActivationByType(BaseScheduler):
         self.steps += 1
         self.time += 1
         if self.model._steps < self.steps:
-            self.model.advance_time()
+            self.model._advance_time()
 
     def step_type(self, agenttype: type[Agent], shuffle_agents: bool = True) -> None:
         """
@@ -498,7 +498,7 @@ class DiscreteEventScheduler(BaseScheduler):
         self.time = end_time
         self.steps += 1
         if self.model._steps < self.steps:
-            self.model.advance_time()
+            self.model._advance_time()
 
     def get_next_event_time(self) -> TimeT | None:
         """Returns the time of the next scheduled event."""
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -114,7 +114,7 @@ class BaseScheduler:
         self.do_each("step")
         self.steps += 1
         self.time += 1
-        if self.model._steps < self.steps:
+        if self.model.steps_ < self.steps:
             self.model._advance_time()
 
     def get_agent_count(self) -> int:
@@ -173,7 +173,7 @@ class RandomActivation(BaseScheduler):
         self.do_each("step", shuffle=True)
         self.steps += 1
         self.time += 1
-        if self.model._steps < self.steps:
+        if self.model.steps_ < self.steps:
             self.model._advance_time()
 
 
@@ -203,7 +203,7 @@ class SimultaneousActivation(BaseScheduler):
         self.do_each("advance")
         self.steps += 1
         self.time += 1
-        if self.model._steps < self.steps:
+        if self.model.steps_ < self.steps:
             self.model._advance_time()
 
 
@@ -268,7 +268,7 @@ class StagedActivation(BaseScheduler):
             self.time += self.stage_time
 
         self.steps += 1
-        if self.model._steps < self.steps:
+        if self.model.steps_ < self.steps:
             self.model._advance_time()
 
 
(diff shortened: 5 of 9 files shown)


quaquel:
The use of model.agents looks good to me.

One quick question: you also changed how time is handled. It was retrieved from the schedule, and now you track it internally. This is because you don't want to depend on schedule, which makes sense. However, from a conceptual point of view, is this the nicest, longer-term way of handling time?

rht:
> It was retrieved from the schedule, and now you track it internally.
> This is because you don't want to depend on schedule, which makes sense. However, from a conceptual point of view, is this the nicest, longer-term way of handling time?

Thank you for raising the issue. In this PR, I track it within the `datacollector` object, which signifies the observer's clock. Which is not ideal. Still looking for a better solution. One alternative would be to define a method in the `mesa.Model`:
```python
def advance_time(self):
    self.time += 1
```

In the user's model step
```python
def step(self):
    self.advance_time()
```
This is how it is done in [abcEconomics](https://github.com/ab-CE/abce).

quaquel:
I think tying time to `model.step` is indeed the way to go. The simple solution would be to handle it through a `super` call or through some kind of annotation. To be discussed separately at some point.

EwoutH:
> I think tying time to `model.step` is indeed the way to go.

At least by default. I think some concept of actual time would also be very useful. See https://github.com/projectmesa/mesa/discussions/1912#discussioncomment-7916666

rht:
> The simple solution would be to handle it through a super call or through some kind of annotation. To be discussed separately at some point.

I had thought about `super().step()`. The problem is that it is very easy to forget doing so. It doesn't have a direct meaning of incrementing the time.

> At least by default. I think some concept of actual time would also be very useful. See https://github.com/projectmesa/mesa/discussions/1912#discussioncomment-7916666

There would be the `ContinuousSpace` equivalent of time in that the system may advance a `float`-type amount of time, and indeed that `steps` needs to remain separate.

rht:
Regarding with the term, `advance_time` is more general than `increment_time` or `step_time`, with the latter 2 are specific for discrete time step. But at the same time, the term needs to also encompass incrementing the model steps count. As such, `advance_time` is not sufficient.

rht:
To be pedantic, `step_and_advance_time` would be the term for incrementing both `time` and `steps`.

quaquel:
> There would be the ContinuousSpace equivalent of time in that the system may advance a float-type amount of time

In my view, the moment you allow for this, you should go all the way and have a discrete event-style event list at the heart of everything. If you want traditional ABM behavior, you just schedule evenly-spaced events. Each event would be then a call to `model.step`. If you want full-blown discrete event behavior, you can schedule events (i.e., a combination of a time instant and callable) at any other non-discrete time instant. In fact, you can simply hybridize this by allowing both. Time then is held by the event list. 

No idea what a resulting clear API could look like.

EwoutH:
> Another thought I have was giving `Model.step()` a `timestep` argument. Default would be `timestep=1`, but you can change that (once or every step if you want).

This could integrate the advance time part into the step, right?

quaquel:
There is a way to track the number of calls to `model.step` without relying on super. It involves the use of metaclasses.

```python
from collections import defaultdict
from functools import wraps

def count_calls(func):
    name = func.__name__

    @wraps(func)
    def wrapper(self, *args, **kwargs):
        # creates the instance counter if necessary
        counter = getattr(self, "time", None)
        if counter is None:
            counter = 0
        setattr(self,"time", counter + 1)
        return func(self, *args, **kwargs)

    wrapper._is_count_call_wrapper = True
    return wrapper


class CountStep(type):
    def __new__(cls, name, bases, attrs):
        try:
            step_method = attrs["step"]
        except KeyError:
            print(attrs)
        else:
            attrs["step"] = count_calls(step_method)
        return super(CountStep, cls).__new__(cls, name, bases, attrs)

class Model(metaclass=CountStep):
    pass
    
class MyModel(Model):

    def step(self):
        print(self.time)

```

If we run this

```python
model = MyModel()
for _ in range(10):
    model.step()
```

we nicely get 1 ... 10.

quaquel:
> > Another thought I have was giving `Model.step()` a `timestep` argument. Default would be `timestep=1`, but you can change that (once or every step if you want).
> 
> This could integrate the advance time part into the step, right?

Not sure how to read something like this. What would it mean if you say `step(timestep=3.1415)`? Do you mean to take the current time and add the timestep to it and execute all events scheduled between the current time and the new endtime? Or something else?

rht:
(I'm probably digressing too much here on DES...)

Considering https://github.com/projectmesa/mesa/discussions/1912#discussioncomment-7920886
> In case of ABM, there are only fixed time intervals, or ticks, between sets of events.
> MESA lacks an eventlist. Instead, it is up to the user to advance the eventlist by one tick at a time by calling model.step().

I'm drawing example from the [Eurace@Unibi model](https://www.uni-bielefeld.de/fakultaeten/wirtschaftswissenschaften/lehrbereiche/etace/eurace@unibi/), one of the most elaborate macroeconomic model that has existed.

Taking an excerpt from the [paper on Eurace@Unibi](https://pub.uni-bielefeld.de/download/2622498/2643911/Eurace_Unibi_OUP_Handbook_dec_2013.pdf)
```
Concerning the activation of agents, the actions can be calendar-based (time-driven) or event-based, where the former can follow either subjective or objective time schedules (agent-time vs.
clock-time). Furthermore, the economic activities take place on a hierarchy of time-scales: yearly,
monthly, weekly and daily activities all take place following the calendar-time or subjective agent-time. Agents are activated asynchronously according to their subjective time schedules that are
anchored on an individual activation day. These activation days are uniformly and randomly
distributed among the agents at the start of the simulation but may change endogenously.
```

If we were to be able to model a reduced version of Eurace@Unibi in Mesa, for pedagogical purpose. Extending Mesa to describe events would be necessary.

quaquel:
Colleagues of mine have been doing pandemic modeling with models involving between 150 thousand and 25 million agents. The only way to make this computationally feasible was by switching from calender-based to event-based activation. So at some point figuring out how to support this in MESA would be great.

In the meantime, however, there is still the issue of tracking the time of the simulation. Would it make sense to do it along the lines of the metaclass example I have given above?

rht:
> In the meantime, however, there is still the issue of tracking the time of the simulation. Would it make sense to do it along the lines of the metaclass example I have given above?

While it is more convenient to the user, I find the implementation to be too complex for the reader of Mesa code. The library code needs to be simple enough without requiring one to spend an effort to decipher the implementation to what amount to tracking the time automatically.

rht:
> Not sure how to read something like this. What would it mean if you say step(timestep=3.1415)? Do you mean to take the current time and add the timestep to it and execute all events scheduled between the current time and the new endtime? Or something else?

I see it as the step period that is not necessarily an integer. All the events that happen within the step, are sorted based on their activation times, and are executed in order. That event1 fires at time 0.06674, event2 at 0.1054, event3 at 2.9979, which information is used to decide their execution order.

rht:
What do you think of this approach instead?
```python
class Clock:
    def __init__(self):
        self.steps = 0
        self.time = 0

    def step(deltat=1):  # deltat is more mnemonic than timestep
        self.steps += 1
        self.time += deltat

class MyModel(mesa.Model):
    def __init__(self, ...):
        self.clock = Clock()

    def step(self):
        # This is sufficiently mnemonic, as a replacement of self.schedule.step()
        self.clock.step()
```
And so, we reuse the existing ABM terms without having to invent new terms. It's FSM all the way down.

EwoutH:
That's exactly what I had in mind. Only instead of creating a new class, I would just integrate it in the Model class.

quaquel:
Why do I use a library for something? To avoid having to write boilerplate code. All models need to track time, so if I use a library, the library should handle this for me. With this suggestion, I must add Clock myself and remember to advance it. It also adds two lines of code to any model I make. Also, speaking from experience teaching MESA for the last 3 years at the MSc level, this is something that will easily trip up new users. So, no, I don't like this suggestion.

> While it is more convenient to the user, I find the implementation to be too complex for the reader of Mesa code. The library code needs to be simple enough without requiring one to spend an effort to decipher the implementation to what amount to track the time automatically.

So, here I have a different view. For me, the cleanliness of the API and the use of the library come first. So be it if a clean and easy-to-use API requires some more obscure Python machinery. Because, who is going to read the source code? Only users who are invested in the library and already have some programming background. So, as long as the code is well documented and explained at a high level (i.e., what does it do) and with some detail on how this is achieved, I prefer such a solution over forcing my user always to add boilerplate code.

rht:
I appreciate the criticism and see the point regarding with the annoyance of having to manually carry along the `Clock` object to wherever the dynamics of the model happens. But I do think that having a simple library encourages users to read the source code, to extend and experiment with them, and to more likely contribute back to the code. From the maintainers' perspective, obscure code only works if there are only a small number of maintainers who understand the code but no one else does. If https://github.com/projectmesa/mesa/pull/1942#issuecomment-1880024330 were to be incorporated, from the perspective of an uninitiated developer new to this section of the code, the Git archaelogy would have been more involved than the issue I encountered with `__new__` and `__init__` in the model initialization.

At the very least, `model.clock` simply replaces `model.schedule` in the previous code. And the students may understand what is going on under the hood, instead of using a library that "just works".

Regarding with [PEP 20](https://peps.python.org/pep-0020/):
> Simple is better than complex.

I interpret it as overall simplicity, instead of simple API but obscure implementation.

rht:
That said, I still think there might be a solution where a `model.clock` or `model.steps` & `model.time` is not needed, depending on how `model.datacollector`'s definition, as an observer to the system, should be modified.

rht:
> That's exactly what I had in mind. Only instead of creating a new class, I would just integrate it in the Model class.

This seems the simplest approach for now. `model.steps` and `model.time` are automatically initialized during `super().__init__()`. Users then just need to remember to do `self.step_and_advance_time()` inside the model `step()`, as the only boilerplate line.

quaquel:
I am fine with a simple solution, although I would advocate calling super over adding another method to the model class. It is still only one additional line, but at least for me and in my teaching, I always advocated calling super anyway. 

Some other thoughts
1. The discussion for me is not about simplicity versus complexity. I agree that PEP 20 should guide all Python projects. Here, however, there is a trade-off between the simplicity of the API and the simplicity of the implementation. I am personally also doubtful whether using metaclasses for something trivial like tracking time is defensible. However, it is the only solution I could devise which avoids forcing the user to write additional boilerplate code. 
2. The argument that implementation simplicity would stimulate users to contribute back to MESA makes no sense to me. In my experience, other factors drive that choice. WRT to mesa, the fact that ruff is not automated, that there are many open issues where it is unclear whether the maintainers actually want to address them (see. e.g., my discussion with @Corvince in @1933 on #574), the lack of milestones, and the many stale pull requests, are much more important in my decision on whether to continue to contribute than readability. In fact, having spent the better part of yesterday getting my head around the current space implementation, there are more important parts of the current code base that are hard to comprehend than a relatively small part of the code that could be quite easily explained (also PEP 20: `If the implementation is easy to explain, it may be a good idea.`) with some comments in the code (i.e., the suggestion is a combination of an annotation (count_calls) and a metaclass to automatically assign this annotation to the step method of the user's model). 
3. Yes, datacollection is in need of an overhaul. However, I believe the data collector should retrieve time from the model rather than be responsible for maintaining time. Because, as @rht stated, the data collector observers the model and a model should run fine without any data collection.

rht:
> The argument that implementation simplicity would stimulate users to contribute back to MESA makes no sense to me. In my experience, other factors drive that choice. WRT to mesa, the fact that ruff is not automated, that there are many open issues where it is unclear whether the maintainers actually want to address them ...

I think the emphasis on the implementation simplicity of the code should be orthogonal to the questioning of the maintainers' time commitment. There would definitely be a situation where both the code is simple and readable, together with active maintenance. (On my end, #1933 on #574 is definitely on my radar; I just need some time to digest them.)

To localize the discussion on the system clock: that said, https://github.com/projectmesa/mesa/pull/1942#issuecomment-1880024330 handles the `model.steps` update by counting the number of `step()` calls, but it hasn't taken into account of `model.time` update. You could specify the `timestep` at model `__init__`, but there is an implicitness in this design choice.

tests/test_batch_run.py
```diff
@@ -85,6 +85,7 @@ def get_local_model_param(self):
     def step(self):
         self.datacollector.collect(self)
         self.schedule.step()
+        self.advance_time()
```
EwoutH, line 88: If we can remove this line from being needed, we are fully compliant with current user methods. I would see `deltat` as an input argument of step itself (defaulting to one).

EwoutH:
Let’s separate some issues here:

1. Tracking time in the model
2. Decoupling the data collector from the scheduler
3. State tracking
4. Event based activation
5. Complexity in implementation vs user API
6. Mesa maintenance

1 and 2 are implementation discussions. I think everyone agrees they should be done, so let’s (continue) discussing how. Maybe in separate issues or PRs though, and I think it might be useful to do 1 first and then 2.

3 and 4 are long term and conceptual. In any case, you probably want a central clock in the model, right? So it doesn’t block 1 or 2, and we can continue discussing 3 and 4 in their respective discussions.

5 important, but can quickly get very broad. If it’s not about this specific implementation anymore I would say spin off into a new discussion.

6 also important, but can get personal, and thus maybe face to face stuff (or very well-thought out written out).

(might still be missing some stuff)

<hr>

In general, I would suggest issues and PRs to be atomic, and only focus on one coherent issue. Of course it can touch other stuff, and therefore spin-off new discussions (which is great in general, also about meta things like user/contributor friendliness), but let’s try to spin-off those discussions in separate threads, or discuss them in a face to face dev meeting. That helps to keep the PRs on topic.

tests/test_batch_run.py
rht: My take on the alternative was:

> I had thought about super().step(). The problem is that it is very easy to forget doing so. It doesn't have a direct meaning of incrementing the time.

I suppose it is the choice being remembering to always:
- call super on an inherited method
- advance the clock and steps

I pick the latter because it is more explicit on what it does. But I can see the former being lower on the LOC needed, because I can also sweep `self.datacolllector.collect(self) if hasattr(self, "datacollector") else None` into `super().step()`.

rht: After writing it more explicitly,
```python
def step(self, deltat=1):
    super().step(deltat)
```
One problem is that the time update needs to happen last. And the super call being called last doesn't seem conventional.

Corvince:
Thanks a lot for summarizing the sometimes confusing discussion @EwoutH !

However I disagree on 

> 2. Decoupling the data collector from the scheduler
> [...]
> 1 and 2 are implementation discussions. I think everyone agrees they should be done, so let’s (continue) discussing how. Maybe in separate issues or PRs though, and I think it might be useful to do 1 first and then 2.

From the discussion in #1912 I think it is still unclear if we want to actually get rid of schedulers or not. I think we need to continue to discuss this first before we continue this path here. Because if we want to keep schedulers I think they are the right place to keep track of time and so there is no need to decouple the logic. I mean the whole discussion can be viewed as an advantage of schedulers - it is clear that they need to track time. For example, if we remove schedulers, but add a Clock instance we don't really gain anything. Same for tracking time inside the model instance - we just further clutter the model namespace, but keep a tied coupling. This isn't necessarily bad, but we should really first discuss about the future of schedulers before arguing about implementation details. And the best place for this discussion is #1912.

rht:
> For example, if we remove schedulers, but add a Clock instance we don't really gain anything.

The gain: `model.agents.shuffle().do("step") and model.agents.shuffle().do("advance")` is conceptually clearer than the term `StagedActivation` with lots of boilerplate code. The `Clock` instance has a very specific purpose and is easy to conceptualize and explain.

EwoutH:
> From the discussion in https://github.com/projectmesa/mesa/discussions/1912 I think it is still unclear if we want to actually get rid of schedulers or not.

Agreed, but it’s clear we want to make them optional, right? Because if they’re not optional you’re still bound by the currently implemented schedulers.

Corvince:
My point was purely on the purpose of this PR. You remove the dependency on scheduler, but add a dependency on Clock. 

Hope that clears this up

tests/test_batch_run.py
EwoutH: Could we use a decorator for time advancement?

```Python
def time_advancement_decorator(step_func):
    def wrapper(self, *args, **kwargs):
        # Extract deltat from kwargs or default to 1 if not provided
        deltat = kwargs.get('deltat', 1)
        step_func(self, *args, **kwargs)
        self.advance_time(deltat)
    return wrapper

class Model:
    @time_advancement_decorator
    def step(self, deltat=1):
        # Base step implementation
        pass

class MyModel(Model):
    def step(self):
        # Subclass-specific logic
        pass
```

Corvince:
> > From the discussion in #1912 I think it is still unclear if we want to actually get rid of schedulers or not.
> 
> Agreed, but it’s clear we want to make them optional, right? Because if they’re not optional you’re still bound by the currently implemented schedulers.

Is this clear? It wasn't for me, but I might have missed part of that discussion. I don't think we should keep it optional. Otherwise we have more than 1 way of doing things and especially if we track time outside of the scheduler that will be very confusing. I would strongly recommend to either stick with a dedicated scheduler or deprecate it. Personally I lean towards deprecating them, but @quaquel makes some good points in #1912

quaquel:
Some additional thoughts, although I still haven't fully made up my mind on the future of schedulers.

1. @jackiekazil indicated she was in favor of supporting multiple schedulers. At the moment, this is already possible. However, this also means that you don't have a single global truth for time. This to me is less than ideal. 
2. The current "schedulers" are not really schedulers in a discrete event sense. They are a way of ordering the execution of agents at a given tick. 
3. If schedulers are deprecated, I would like a replacement. I am inclined to go with a full event list. This makes it possible to start supporting full discrete event simulation (through event scheduling, which is a particular flavour of des) while also allowing ABM-style activation. You can schedule something like the current "model.step" as events that are evenly and discretely spaced. A call to `model.advance(delta=5`) would then execute 5 model.step events. The eventlist would then also be the single source for the current simulation time and stuff like the current `model.running` attribute. Put differently, the current `model.step` and scheduler approach works fine, and something like this should remain easy to do (doing agentset.shuffle etc would qualify for me as simple enough), but it would be great if a richer form of discrete event simulation is available if you want and need it.

tests/test_batch_run.py
rht: That should do. It frees the user from having to deal with the location of the `advance_time` within the `step()` specification (that it should be called last). I'm in favor of using a decorator.

rht: I didn't see the 2 models in the code. What I had in mind was to specify the decorator in `MyModel`.
```python
class MyModel(Model):
    @time_advancement_decorator
    def step(self):
        # Subclass-specific logic
        pass
```

quaquel: > That should do. It frees the user from having to deal with the location of the `advance_time` within the `step()` specification (that it should be called last). I'm in favor of using a decorator.

Just for my understanding, why should time be advanced last?

rht: I suppose it would be fine to advance time first before everything else, at the expense of the steps starting from 1, and the time starting from `deltat`, instead of 0.

EwoutH: The challenge here is to find a solution which requires no changes to existing user models. Otherwise we have to bump to 3.0.

quaquel: > The challenge here is to find a solution which requires no changes to existing user models

That is easy: leave the schedulers alone for now for backward compatibility while slowly putting in place more sophisticated machinery (as we did with the implementation of the scheduler on top of AgentSet). 

> I think it's just convention to maintain roughly this order:

> Do things
Collect data
Increase time

Interesting. For me, the obvious order is increment time, do stuff, and gather data. But indeed, It does not really matter.

EwoutH: Yeah that's also an obvious one. If you then collect data once after initialisation you also capture the initial model state at timestep 0, and the final model state at timestep n, giving a database of n+1 length.

Maybe we should formalize a good default for this somewhere.

(note that the order of do stuff and increase time not matter, only collect data does)

EwoutH:
> @jackiekazil indicated she was in favor of supporting multiple schedulers. At the moment, this is already possible. 

True, but currently datacollection is limited to the time tracked by one scheduler.

> The default DataCollector here makes several assumptions:
>     * The model has a schedule object called 'schedule'

And there might be more caveats (and I think multiple schedulers doesn't make sense if you can use a flexible AgentSet calls instead).

<hr>

But to get back on track, I would propose that we first add a time and step counter to the Model object. That way there is a global simulation time, with the model `step()` function increasing those by default. That would also allow using only the AgentSet API, without schedulers. I think there is consensus that that's useful, right?

Then, a second PR can be opened to decouple the datacollector from the scheduler and use the global time.

@rht do you agree that it would be useful to do this in two separate PRs? If so, are you willing to implement them?

rht:
> But to get back on track, I would propose that we first add a time and step counter to the Model object.

It's already the case. See the 2nd commit. The current PR does it with `advance_time()` at the end. And regardless, this PR can't be merges unless the next release is 3.0, which is very likely to be the case given the rehaul of most of the components in Mesa.

> Then, a second PR can be opened to decouple the datacollector from the scheduler and use the global time.

The clock has no longer been in the datacollector for days.

rht:
Actually, I argue that this PR needs to be merged for 2.2.0. The model clock exists independently of the scheduler's clock. Users who still use `model.schedule.step()` won't have to call `advance_time()`, but users who don't use the schedulers at all will have to call `advance_time()`, or use the decorator.

@tpike3 I forgot to mark https://github.com/projectmesa/mesa-examples/pull/80 as a draft, because I removed the scheduler in that experimental Schelling example.

rht:
Never mind. This PR is still a breaking change regardless. I will revert https://github.com/projectmesa/mesa-examples/pull/80.

rht:
I have minimized the scope of the breaking change in the 3rd commit. In the 3rd commit, for the `step()` definition of all the schedulers, I added `self.model.advance_time()`, so that both clocks are in sync, and I can use `model.steps` inside the datacollector. But this is still a breaking change in that users' custom scheduler will have to include `self.model.advance_time()` in their `step()` method.

EwoutH:
Only if A) users need to change their models or B) current behavior changes it's a breaking change. We can do the following things without a breaking changes:
- Add time and step counter to the model
- Update those when the model step() is called (without requiring users to do anything different - this might be difficult)
- Let the datacollector depend on the step in the model instead of the step in the scheduler, if either:
  - no `model.scheduler` object is detected
  - the users request that with a keyword argument (`use_model_steps=True`)

A keyword argument also allows to deprecate behavior at some point if we want to.

What I find somewhat weird, is that the current datacollector records the step number, and not the time. We might want to add a keyword argument `use_time`, which if `True` records the time instead of the step number.

_(it might be useful to open a discussion thread first to agree on new API features etc. before we continue with implementations)_

EwoutH:
For 769f58d, I think it's a smart solution, but it only works with one scheduler right? Can we also make it work for models with multiple schedulers?

The datacollector now relies on a variable called `model.schedule`. I think as a default the time from that schedulers should be synced with the model clock.

rht:
> An alternative non-breaking change would be to simply add an optional steps_counter keyword-only argument to the datacollector. 

Disagreed on mixing the clock with the data collector. It's conceptually confusing. Also, while they are fewer in number, some models might not have a `model.datacollector`, because this is not a mandatory attribute as per the model specification.

> Otherwise I wouldn't be comfortable merging this PR which adds two new public API attributes (steps and time), that at least I feel we haven't fully discussed yet.

They can always be turned private. Not a blocker.

> There are still some interesting ideas by @quaquel around event sourcing and basing schedulers on an event list, that might be incompatible with this new API.

Still needs careful discussion on the event sourcing. Needs to be pragmatic for this PR so that `model.schedule` is optional, given that AgentSet is classified as experimental feature.

---

> For https://github.com/projectmesa/mesa/commit/769f58d57e294355c7648e973d58e9ba5ba07275, I think it's a smart solution, but it only works with one scheduler right? Can we also make it work for models with multiple schedulers?

I suppose for to make it work for multiple schedulers, you do this instead:
```python
if self.model.steps < self.steps:
    self.model.advance_time()
```
I will add a commit that does this. For sanity check, they should have differed only by 1.

Corvince:
> > An alternative non-breaking change would be to simply add an optional steps_counter keyword-only argument to the datacollector.
> 
> Disagreed on mixing the clock with the data collector. It's conceptually confusing. Also, while they are fewer in number, some models might not have a `model.datacollector`, because this is not a mandatory attribute as per the model specification.
> 

No I think you misunderstood my proposal. Here is a more complete version to see how it would work for a new model that doesn't use a scheduler

```python
class MyModel(Model):

    def __init__(self):
        super().__init__()
        self.steps = 0
        self.data_collector = Datacollector({...}, steps_counter=self.steps)

    def step(self):
        self.agents.do("step")
        self.steps += 1
        self.data_collector.collect()

# In DataCollector.py
...
if steps_counter is None:
     steps_counter = self.model.schedule.steps
```
And then use steps_counter where we currently hardcode model.schedule.steps

This way we would stay 100% backwards compatible, but also make the use of schedulers optional without prematurely adding new APIs. Of course this would only be a temporary solution until we settle our discussions.

EwoutH:
I like this idea, but if we release it as stable I think it would be more difficult to add `steps` in to the Model by default in the end.

So maybe we first need the complete picture before we can more on steps like this.

I think this will benefit from some face-to-face brainstorming and discussion.

rht:
> self.data_collector = Datacollector({...}, steps_counter=self.steps)
> This way we would stay 100% backwards compatible, but also make the use of schedulers optional without prematurely adding new APIs.

> I like this idea, but if we release it as stable I think it would be more difficult to add steps in to the Model by default in the end.

That seems to be fine as an optional argument, but `model.steps` is still a new public API, and `steps_counter=self.steps` is less readable than `model.advance_time()`.

> If I understand correctly, the API's additional public part is tiny (model._step and model.advance_time). Any future, more comprehensive solution should be able to either fit with this or work around this.

I suppose the API can be marked explicitly as experimental and private: `model._experimental_steps`, `model._experimental_time`, `model._advance_time`. Or maybe `model.experimental.steps`, `model.experimental.time`, `model.experimental.advance_time`?

EwoutH:
Is there consensus that in all cases (no schedulers, single schedulers, multiple schedulers) a single clock in the model would be useful? In both time and absolute steps?

In that case, can we do the same trick we did in #1976 to reserve `model.steps` and `model.time`? Then we can implement them with as `_steps` and `_time` before 3.0, and switch to `steps` and `time` in 3.0.

rht:
The current state of this PR is my vote on how it should be done, for now. I could rename the attributes to `self.time_` and `self.steps_` following #1976. I haven't fixed the CI failures though. But I have to say that the current experimental Schelling example is broken because https://github.com/projectmesa/mesa-examples/pull/80 (which removed the scheduler) got merged. I'd personally want this PR to be in 2.2.2.

quaquel:
I agree with @EwoutH and @rht. This is the right direction to go. Adopting the simple underscore thing for 2.2.2 is a good idea, but we need to develop a general strategy for this type of problem. Feel free to let me know if you need any help on fixing the tests.

mesa/model.py
```diff
@@ -67,6 +69,9 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         self.current_id = 0
         self._agents: defaultdict[type, dict] = defaultdict(dict)
 
+        self._steps = 0
```
EwoutH, line 72: Can we assume this is an `int`?

EwoutH:
The only thin I currently still dislike is that time and step have to explicitly increased (by using `_advance_time`). It should just keep track of the number of steps by default, and optionally can be overwritten if you want to do something else.

tests/test_batch_run.py
```diff
@@ -85,6 +85,7 @@ def get_local_model_param(self):
     def step(self):
         self.datacollector.collect(self)
         self.schedule.step()
+        self._advance_time()
```
EwoutH, line 88: Shouldn’t `self.schedule.step()` take care of this now? (also in the other tests)

rht: Good catch.

mesa/model.py
rht: Annotated.

rht:
I renamed `_time` -> `time_`, and `_steps` -> `steps_`.
>  Feel free to let me know if you need any help on fixing the tests.

Yes, a help would be appreciated.

EwoutH:
> I renamed `_time` -> `time_`, and `_steps` -> `steps_`.

I think this was an unintended side effect of having already used _agent, so something else had to be used. I like using `_var` more than `var_`. The former is the default for a "private" variable in Python, and I think as a library we can make it very defensible that we want to reserve all private variables for future use without breaking API (at all).

EwoutH:
What do we think about a [dynamic wrapper approach](https://chat.openai.com/share/d9b9c6c6-17d0-4eb9-9eae-484402bed756)? Because the wrapper is added on the `super().__init__()` (which is now required to be called), it will work for subclassed models.

> One solution is to use a wrapper function around the user's `step` method. This wrapper can be set up during the initialization of the model object, ensuring that each call to `step` automatically includes the time and step updating logic. This can be achieved with the following steps:
> 
> 1. **Modify the `__init__` Method:** In the `__init__` method of the `Model` class, dynamically wrap the `step` method with a custom wrapper that first calls `_advance_time` and then the original `step`.
> 
> 2. **Create a Wrapper Function:** This function will be responsible for calling `_advance_time` and then the user's `step` method.
> 
> Here is how you can implement this approach:
> 
> ```python
> class Model:
>     # ... existing code ...
> 
>     def __init__(self, *args: Any, **kwargs: Any) -> None:
>         # ... existing initialization code ...
>         self._steps: int = 0
>         self._time: TimeT = 0  # the model's clock
>         self._wrap_step()
> 
>     def _wrap_step(self):
>         """Wrap the step method to include time and step updating."""
>         original_step = self.step
> 
>         def step_wrapper():
>             self._advance_time()
>             original_step()
> 
>         self.step = step_wrapper
> 
>     def step(self) -> None:
>         """A single step. To be overridden in subclasses."""
>         pass
> 
>     def _advance_time(self, deltat: TimeT = 1):
>         """Increment the model's steps counter and clock."""
>         self._steps += 1
>         self._time += deltat
> ```
> 
> In this revised approach:
> 
> - The `step` method in user-defined subclasses remains named `step`.
> - When an instance of `Model` (or a subclass) is created, the `step` method is automatically wrapped with a function that updates the time and steps before calling the original `step`.
> - The original functionality of user-defined `step` methods is preserved without any changes required from the user.
> 
> This solution ensures backward compatibility and automates the time and step updating process without requiring any changes to existing user code.

rht:
> > Feel free to let me know if you need any help on fixing the tests.

> Yes, a help would be appreciated.

On the test fail of `TestDataCollector.test_agent_records`, it seems that the actual value of 14 is the correct behavior (i.e. the previous value from step 3), given that `schedule._agents[3]` is removed at step 4.

rht:
https://github.com/projectmesa/mesa/pull/1942#issuecomment-1903419315

That's a very simple solution! Let me add more commits to implement it. I will credit you as the author of the commits.

rht:
Actually, I need to additionally credit GPT-3.5 / GPT-4 (which one? Also the training cutoff date) in the commit message body.

rht:
Implemented. But encountered this pickle errror: https://github.com/projectmesa/mesa/actions/runs/7608147424/job/20716756266?pr=1942#step:8:102.

Corvince:
> [#1942 (comment)](https://github.com/projectmesa/mesa/pull/1942#issuecomment-1903419315)
> 
> That's a very simple solution! Let me add more commits to implement it. I will credit you as the author of the commits.

I am not sure this works according to ChatGPT itself

> Calling super().__init__() in the subclasses of Model will ensure that the initialization code in the Model class, including the wrapping of the step method, is executed. However, this alone may not be sufficient to ensure that overridden step methods in subclasses also get wrapped properly. This is because the wrapping happens at the time of the Model class's initialization, which occurs before the subclass defines its own step method.
> 
> Here's a breakdown of the situation:
> 
> When super().__init__() is Called: In a subclass, when you call super().__init__(), it executes the __init__ method of the Model class. At this point, the Model class's step method (which does nothing by default) is wrapped.
> 
> Overriding step in Subclasses: After super().__init__() is called, if the subclass defines its own step method, this new step method is not automatically wrapped. The wrapping has already occurred, and it was applied to the original step method from the Model class.

But I am not sure this is correct Hope it works, would be elegant.

rht:
> Don't think so. Otherwise we can start crediting every time we use stackoverflow, google, that one book I once read, my first-grade teacher, etc.

I'd say it's more like if you ask StackOverflow directly about this PR, and getting a targeted answer from it. It just happens that the answer is from an AI language model instead of a human. Additionally, this PR is likely going to be used for training data for future language/multimodal models anyway, and so it's like contributing back to the free (as in free beer) help by us annotating whether the answer is labeled as useful or not.

EwoutH:
Looks like the pickle issue is resolved this way, right? Only errors are remaining are numerical in the DataCollector, and one in the test_agent_lifetime.

rht:
Right, no more pickle problem.

mesa/datacollection.py
```diff
@@ -180,11 +178,11 @@ def _record_agents(self, model):
         rep_funcs = self.agent_reporters.values()
 
         def get_reports(agent):
-            _prefix = (agent.model.schedule.steps, agent.unique_id)
+            _prefix = (agent.model.steps_, agent.unique_id)
             reports = tuple(rep(agent) for rep in rep_funcs)
             return _prefix + reports
 
-        agent_records = map(get_reports, model.schedule.agents)
+        agent_records = map(get_reports, model.agents)
```
rht, line 185: This could be the cause of the `test_datacollector.py` problems.

rht: It appears that `model.schedule.remove(agent)` doesn't remove the agent in `model.agents`.

mesa/experimental/jupyter_viz.py
```diff
@@ -181,7 +181,7 @@ def on_value_play(change):
     def do_step():
         model.step()
         previous_step.value = current_step.value
-        current_step.value += 1
+        current_step.value = model.steps
```
EwoutH, line 184: should this be `model.steps_` ?

rht: Yes, good catch. Almost broke JupyterViz with this line.

mesa/datacollection.py
rht: Do you agree with https://github.com/projectmesa/mesa/pull/1942/commits/c4ad2b9e10c2507e502cd8cf4ff3272823275783 ? It is necessary for the test_agent_lifetime to pass.

EwoutH:
Really like the current code changes!

With https://github.com/projectmesa/mesa-examples/pull/89 we went from 5 errors to 3 in the examples.

EwoutH:
There are three errors:
- (2s) Bank reserves: In the datacollector [wealth is collected](https://github.com/projectmesa/mesa-examples/blob/21624513c6ad94faabbb1cb4b97d73b897f9afae/examples/bank_reserves/bank_reserves/model.py#L137), Person agents have wealth but the Bank agent does not. Really strange why that starts failing now.
- In sugarscape_g1mt there is a keyerror on [model.schedule.steps = 1](https://github.com/projectmesa/mesa-examples/blob/21624513c6ad94faabbb1cb4b97d73b897f9afae/examples/sugarscape_g1mt/sugarscape_g1mt/model.py#L198) in datacollector._agent_records. Should that be step 0? Or something else?
  ```
  >       agent_trades = self.datacollector._agent_records[self.schedule.steps]
  E       KeyError: 1
  ```

EwoutH:
But why should an existing model with a scheduler break? Probably means we have the number of steps wrong somewhere right?

rht:
2 of the 3 errors is fixed by https://github.com/projectmesa/mesa-examples/pull/90. The remaining one is fixed by #1996. This PR needs to be merged soon because currently the Mesa Docker build uses the experimental Schelling. And we will have to release 2.2.4.

rht:
One more PR to be merged and the examples test should pass: https://github.com/projectmesa/mesa-examples/pull/91.

rht:
New examples test error should be fixed by https://github.com/projectmesa/mesa-examples/pull/92.

rht:
Wait, never mind. `self.schedule.steps()` never gets called. This is probably not the right fix.

rht:
This PR is ready to merge. But it has to be merged soon after/before https://github.com/projectmesa/mesa-examples/pull/95. I have tested locally that the remaining test error is fixed by both.

tests/test_batch_run.py
EwoutH: I think we have this resolved with the wrapper step right?

EwoutH:
Thanks a lot for this. Will review.

Could you please update the PR description to represent the final state of this PR? Most important are:
- Motivation / problems it solves
- Why this problem was solved the way it is (and not another)
- Any limitations, implementation nuances, and other stuff we have to keep in mind

mesa/datacollection.py
EwoutH: No, unfortunately not. The other way around is okay:
- Agent removed from model --> Should be removed from schedule
But the other way around would block multiple schedulers.

EwoutH: No, unfortunatly not. The other way around is okay:
- Agent removed from model --> Should be removed from schedule
But the other way around would block multiple schedulers.

EwoutH: Maybe the test just need to be updated

mesa/model.py
```diff
@@ -68,6 +70,9 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         self.current_id = 0
         self.agents_: defaultdict[type, dict] = defaultdict(dict)
 
+        self.steps_: int = 0
+        self.time_: TimeT = 0  # the model's clock
```
EwoutH, line 74: I would really like this to be `_steps` and `_time`. I know agents currently is the other way around, but the long term goals should be having `_var` as mesa reserved variables.

EwoutH (commented):
I really like how clean this PR now is. Thanks a lot for this effort, I know it was a long journey.

I have three points left (this is just the overview):
- PR description
- Agent removal
- Private variables as `_var`

mesa/model.py
rht: Done.

rht:
PR description updated.

rht:
I request to prioritize reviewing this PR, because of the reason stated in https://github.com/projectmesa/mesa/pull/1942#issuecomment-1907693600.

Corvince:
Thanks for your endless efforts on this one @rht! I think it is good to go, Just one thing:

> * Agent removal

I don't know what this refers to, is this addressed now?

tests/test_batch_run.py
EwoutH: Marking as resolved, we solved this with the wrapper idea in https://github.com/projectmesa/mesa/pull/1942#issuecomment-1903419315.

rht:
> > Agent removal
> I don't know what this refers to, is this addressed now?

@EwoutH disagreed with this line: https://github.com/projectmesa/mesa/pull/1942/files#diff-2ae534e35a540eca09bc3f43b6e3ffd91695f0b178f4824869c8975a0c0423a5R111. His [comment](https://github.com/projectmesa/mesa/pull/1942#discussion_r1465583066). I'm fine with changing the test if this is the consensus.

mesa/datacollection.py
EwoutH: @rht @Corvince this is the last point need addressing. I would propose:
- Adjusting the failing test, agent lifetime now just means something else. Remove from the schedule is like “deactivated”, removed from the model is “terminated/dead”.
- We should document this clearly, with maybe a comment in the code, PR message and release notes.

rht:
I removed the `agent.remove()` line in the commit `
time: Remove agent.remove in remove` so that you can see the test errors again.

rht:
This line is the cause: https://github.com/projectmesa/mesa/pull/1942/files#diff-72396865b218fb20682afecba66d3b794e1306841d39490ed4037e26cba31b29R68. because it is `m.agents`, it is not tied to any (multiple) schedulers, and should be semantically better than relying on a scheduler's agent count. And also that it removes the need to hardcode-assume that the scheduler attribute is called `m.schedule`.

jackiekazil:
This looks like it is near a wrap but I still wanted to chime and say thank you to rht for the work on the contribution and to quaqel for sharing concerns and ideas and @EwoutH for helping break down the problem and review, and @Corvince for review as well as thoughtful contributions.

mesa/datacollection.py
rht: This basically means that the 2.1.5 behavior is such that it doesn't allow multiple schedulers, because the data collector is hardcoded to only 1 scheduler. I suggest to just use my solution (`agent.remove()`) for backward compatibility, and worry about being proper in Mesa 3.0.

quaquel:
I looked through the code. It seems fine for me. The wraped_step is a clever solution. I also agree with the last point of @rht regarding data collection via model.agents.

I don't fully understand the discussion on remove. In my view, the desired behavior is that `schedule.remove` only removes the agent from the scheduler but not from the model as a whole. So like @EwoutH said: it deactivates the agent. 

Yes, this is a very subtle change in overal behavior compared. But, since 2.2, it is recommended to call agent.remove() if you want to kill an agent (which automatically removes an agent from any agentset and thus any scheduler), I am fine with this. 

For clarity, it might be a an idea to issue an explicit warning once when calling schedule.remove to the effect that this only removes the agent from the scheduler but not from the model. This will make it easy for users to adjust their model code if needed.

EwoutH:
Thanks! An agent isn’t killed currently (in 2.1.5) if it’s removed from the schedule right? You still have to do that manually, so 2.1.5 models will have done that.

quaquel:
In 2.1.5, it depends. There is no explicit mechanism in place. Only if all references to the agent are removed will it be garbage collected. In simple cases, this would mean removal from schedule and space. However, MESA offers no guarantee for this.

EwoutH:
Right, I now understand the complication: 
- The data collector depended on the schedule.
- If an agent is only removed from the schedule (and not from the model) the data collection used to stop.
- Now it doesn’t

While it isn’t best practice, there might be models out there that rely on this behavior.

A workaround might be:
- switch an private attribute called _legancy_datacollector_behavior = True

EwoutH:
(sorry misclicked)

Let me think a bit about possible solutions. Maybe we can get away with throwing a clear warning in the right place.

EwoutH:
Another option could be adding some switch like `old_datacollector_behaviour` which we flip on 3.0.

EwoutH:
Sorry but for me this doesn't solve the issue:
- There might be models which remove the agent after removing it from the schedule. Those will now crash, since the agent is already removed.
- There might be models that remove it from the schedule temporarily and then add it back again.
- There might be models that remove it from the schedule but keep it on the grid

I'm not happy about altering `time` module behavior. I would like to solve it in the datacollector. Some kind of flag or switch that "if an agents is removed from model.schedule, we will stop collecting data. With Mesa 3.0 this might change. We recommend explicitly removing your agent from the model with agent.remove() if you want to completely remove the agent."

Corvince:
I think the simplest solution would be to check if model.schedule exists and if it does collect data from its agents. Otherwise use model.agents. This would be backwards compatible, but allow removing the scheduler. And agree that in a future datacollector it should be made explicit which agent data is collected.

quaquel:
I agree with @Corvince proposed solution.

EwoutH:
Good idea, also agreed. @rht would you like to implement it?

rht:
Done. You can check the last commit of this PR.

mesa/datacollection.py
```diff
@@ -180,11 +178,14 @@ def _record_agents(self, model):
...
-            _prefix = (agent.model.schedule.steps, agent.unique_id)
+            _prefix = (agent.model._steps, agent.unique_id)
             reports = tuple(rep(agent) for rep in rep_funcs)
             return _prefix + reports
 
-        agent_records = map(get_reports, model.schedule.agents)
+        agent_records = map(
+            get_reports,
+            model.schedule.agents if hasattr(model, "schedule") else model.agents,
+        )
```
EwoutH, line 188: Really nice! Could you add tests for this behaviour? Just to be 100% certain it works as expected.

rht: This is such a simple line (far simpler than AgentSet's shuffle) and is overkill to add such. That said, there was and is a test:
1. you have seen the CI tests passing with either of the if case.
2. there is already the experimental Schelling in place, which has no scheduler.

rht: What is missing is test of multiple schedulers and datacollectors pairs

rht: Yes, correcting for multiple scheduler-datacollector pairs is out of scope for this PR.

quaquel (approved):

EwoutH (approved):
Good to go! Thanks a lot, we churned out a lot of conceptual things together with this PR. I will try to do a quick write up tomorrow.

I can merge later today if preferred. I recommend either squashing or cleaning up the commits.

EwoutH:
> The only thing I currently still dislike is that time and step have to explicitly increased (by using `_advance_time`). It should just keep track of the number of steps by default, and optionally can be overwritten if you want to do something else.

This is still the case in the current codebase right? If so, it's a bit weird that users should call an private function to increase the time.

PR #1945, opened 2024-01-08
JupyterViz: Simplify SliderInt and SliderFloat specification using slice

rht:
With this PR, instead of the lengthy [`model_params`](https://github.com/projectmesa/mesa-examples/blob/8f6e0437541e66f247b34c97bb732b4cbb73a254/examples/schelling_experimental/app.py#L17-L44)

```python
model_params = {
    "density": {
        "type": "SliderFloat",
        "value": 0.8,
        "label": "Agent density",
        "min": 0.1,
        "max": 1.0,
        "step": 0.1,
    },
    "minority_pc": {
        "type": "SliderFloat",
        "value": 0.2,
        "label": "Fraction minority",
        "min": 0.0,
        "max": 1.0,
        "step": 0.05,
    },
    "homophily": {
        "type": "SliderInt",
        "value": 3,
        "label": "Homophily",
        "min": 0,
        "max": 8,
        "step": 1,
    },
    "width": 20,
    "height": 20,
}
```
This becomes
```python
model_params = {
    "density": {
        "type": "SliderFloat",
        "value": 0.8,
        "label": "Agent density",
        "values": slice(0.1, 1.0, 0.1),
    },
    "minority_pc": {
        "type": "SliderFloat",
        "value": 0.2,
        "label": "Fraction minority",
        "values": slice(0.0, 1.0, 0.05),
    },
    "homophily": {
        "type": "SliderInt",
        "value": 3,
        "label": "Homophily",
        "values": slice(0, 8, 1),
    },
    "width": 20,
    "height": 20,
}
```

### DIFF
--- a/mesa/experimental/jupyter_viz.py
+++ b/mesa/experimental/jupyter_viz.py
@@ -282,22 +282,30 @@ def UserInputs(user_params, on_change=None):
             on_change(name, value)
 
         if input_type == "SliderInt":
+            values = options.get("values")
+            min, max, step = None, None, None
+            if values:
+                min, max, step = values.start, values.stop, values.step
             solara.SliderInt(
                 label,
                 value=options.get("value"),
                 on_value=change_handler,
-                min=options.get("min"),
-                max=options.get("max"),
-                step=options.get("step"),
+                min=min,
+                max=max,
+                step=step,
             )
         elif input_type == "SliderFloat":
+            values = options.get("values")
+            min, max, step = None, None, None
+            if values:
+                min, max, step = values.start, values.stop, values.step
             solara.SliderFloat(
                 label,
                 value=options.get("value"),
                 on_value=change_handler,
-                min=options.get("min"),
-                max=options.get("max"),
-                step=options.get("step"),
+                min=min,
+                max=max,
+                step=step,
             )
         elif input_type == "Select":
             solara.Select(
--- a/tests/test_jupyter_viz.py
+++ b/tests/test_jupyter_viz.py
@@ -31,9 +31,7 @@ class TestMakeUserInput(unittest.TestCase):
             "type": "SliderInt",
             "value": 10,
             "label": "number of agents",
-            "min": 10,
-            "max": 20,
-            "step": 1,
+            "values": slice(10, 20, 1),
         }
         user_params = {"num_agents": options}
         _, rc = solara.render(Test(user_params), handle_error=False)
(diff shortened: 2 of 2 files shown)


EwoutH:
It’s shorter, but also less explicit. So I’m a bit mixed about this.

rht:
It's reminiscent of `range(1, 24, 2)`, and so users already have a shared context of what the 3 numbers refer to.

quaquel:
I like the brevity of this because it is similar to how, _e.g._, range and slicing work; it is easy to explain. I am a bit less sure about the choice to reuse `slice` for this. `slice` is normally used to produce a sequence of indices. Indices, for me, are integers, but that is not the case here. So why did you decide to reuse `slice` for this?

rht:
I agree that using `slice` does not make sense when the arguments are `float`.
```
In [8]: slice(0.1, 0.9, 0.2).indices(1)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[8], line 1
----> 1 slice(0.1, 0.9, 0.2).indices(1)

TypeError: slice indices must be integers or None or have an __index__ method
```
`range` is more suitable, except that it doesn't support `float` arguments.

It seems that I will have to define
```python
class arange:
    def __init__(self, start, stop, step):
        self.start = start
        self.stop = stop
        self.step = step
        self.current = start

    def __iter__(self):
        return self

    def __next__(self):
        if self.current >= self.stop:
            raise StopIteration
        else:
            result = self.current
            self.current += self.step
            return result
```

quaquel:
I indeed see two options:
* add a small helper class as described by @rht, although I would suggest trying to find a different name because arrange clashes with numpy.arrange.
* use a tuple or namedtuple with start, end, step, and handle the rest internally within jupyterviz.

rht:
> use a tuple or namedtuple with start, end, step, and handle the rest internally within jupyterviz.

Regardless of how it should be implemented under the hood, `np.arange` has the closest definition to how the tuple `(start, stop, step)` should mean. I think it's fine to reuse the name, as long as the namespace is `mesa.experimental.arange` or `mesa.arange`.

rht:
https://emaworkbench.readthedocs.io/en/latest/examples/example_mesa.html has a more concise description. Instead of
```python
{
    "type": "SliderFloat",
    "value": 0.8,
    "label": "Agent density",
    "min": 0.1,
    "max": 1.0,
    "step": 0.1,
}
```
we have
```python
from mesa.experimental import `Slider`
Slider("Agent density", 0.8, 0.1, 1.0, 0.1)
# optionally
# min_value instead of min in order to avoid collision with reserved keywords
Slider(name="Agent density", value=0.8, min_value=0.1, max_value=1.0, step=0.1)
```

1. `Slider` encompasses both `int` and `float`, and to translate it to `Solara`, there would be `any(insinstance(e, float) for e in (value, min_value, max_value, step))`
2. This ties back to the original Tornado viz interface

rht:
Superseded by #1972.

PR #1950, opened 2024-01-09
make mesa runable without some dependencies

Corvince:
I was playing around with Jupyter Lite to create a first version of a Netlogo Web-like experience, where users can try out some models on a statically hosted webpage. 

However, Pyodide (the underlying tech) needs either pure python packages or specialized wheels. Mesa itself is pure python, but it depends on mesa-viz-tornado, which depends on tornado, which is not yet compatible with Pyodide. I was trying to install mesa without dependencies, but the way it is currently build it requires mesa_viz_tornado to be present to import mesa.

This PR removes that hard dependency by only trying to import the module. So if users don't use the tornado backend they don't need to have that dependency installed.

I did the same for pandas, which is only used for converting the output of the datacollector and networkx, which is only used for NetworkGrid. 

The other dependencies are either already only used on access or an integral part of mesa.


rht (approved):

mesa/datacollection.py
```diff
@@ -34,11 +34,13 @@
     * The schedule has an agent list called agents
     * For collecting agent-level variables, agents must have a unique_id
 """
+import contextlib
 import itertools
 import types
 from functools import partial
 
-import pandas as pd
+with contextlib.suppress(ImportError):
```
rht, line 42: Reading https://stackoverflow.com/questions/34566806/why-use-contextlib-suppress-as-opposed-to-try-except-with-pass, I suppose that using `contextlib.suppress` is more concise than try-except.

rht:
It seems that the dependencies in Solara (`watchdog`, etc) can be optionally imported as well: https://github.com/widgetti/solara/issues/449#issue-2068946338.

PR #1951, opened 2024-01-10
Update __init__.py - change to 2.2.0 for release

jackiekazil:
The current deploy process uses this line for generating the version.


jackiekazil:
I am going to force a merge to do the release to not delay it. 
I double checked that this version matches the history -- https://github.com/projectmesa/mesa/blob/main/HISTORY.md

We should talk more about how we update __init__.py version since the current build depends on it.

PR #1957, opened 2024-01-11
Refer to just Python instead of Python 3

rht:


EwoutH (approved):

Corvince:
thanks for the PR and the review

PR #1958, opened 2024-01-12
AgentSet: Ensure do() operates on unchanged agents list

rht:


quaquel:
I can conform this worked (I made precisely this fix last night on a plane).

rht:
> I can conform this worked (I made precisely this fix last night on a plane).

The weakref bug indeed no longer showed up from the newly added mesa-examples tests: https://github.com/projectmesa/mesa/actions/runs/7499202630/job/20415560935?pr=1958.

quaquel:
Excellent. Do you think explicitly adding something recreating the bug to the unit tests would be a good idea (as I suggested in #1959?

rht:
Agreed to adding a test case. It is similar to how [SQLite is tested](https://www.sqlite.org/testing.html).
> Whenever a bug is reported against SQLite, that bug is not considered fixed until new test cases that would exhibit the bug have been added to either the TCL or TH3 test suites. Over the years, this has resulted in thousands and thousands of new tests. These regression tests ensure that bugs that have been fixed in the past are not reintroduced into future versions of SQLite.

Corvince:
> I can conform this worked (I made precisely this fix last night on a plane).

Well not really, no. This creates a list of agents before their actions are called and then do the actions on the list. But this means if Agent A removes Agent B, Agent B still fulfills its action.  This should not happen and is luckily already part of our tests (and correctly failing https://github.com/projectmesa/mesa/actions/runs/7499202630/job/20415561300?pr=1958)

quaquel:
> > I can conform this worked (I made precisely this fix last night on a plane).
> 
> Well not really, no. This creates a list of agents before their actions are called and then do the actions on the list. But this means if Agent A removes Agent B, Agent B still fulfills its action. This should not happen and is luckily already part of our tests (and correctly failing https://github.com/projectmesa/mesa/actions/runs/7499202630/job/20415561300?pr=1958)

You are absolutely correct. However, based on my reading of the old scheduler code, this bug already existed in 2.1.5 as well. I don't understand the reference to the tests because I thought that I ran the new implementation of schedulers against the original tests, so this should have been caught before.

rht:
The 2.1.5 code has this extra check: https://github.com/projectmesa/mesa/blob/53e8761b47109b18ff77fae78a9024376b14d0e2/mesa/time.py#L120.

EwoutH:
Thanks for the fix, but this will have more performance implications right? Are there alternative solutions?

Corvince:
Performance implications should be minimal. I just don't understand why it is still not working. Maybe we have to check the test itself?

quaquel:
I have a hunch but no access to a machine atm. Will check tomorrow.

rht:
It's because this operation https://github.com/projectmesa/mesa/blob/36da14542fa09e53497bb46a4e59bf757e29dcb0/mesa/agent.py#L67 still leaves the agent to remain in the `WeakKeyDictionary`, because the agent still exists.

quaquel:
> It's because this operation
> 
> https://github.com/projectmesa/mesa/blob/36da14542fa09e53497bb46a4e59bf757e29dcb0/mesa/agent.py#L67
> 
> still leaves the agent to remain in the `WeakKeyDictionary`, because the agent still exists.

The problem is a bit more complicated. I think that the test is not entirely correct in combination with AgentSet (why test against `model.log`?). The root cause, however, seems to be that when you turn the `weakkeyDict` into a list, the weakrefs are automagically changed back into hardrefs. Using `self._agents.copy()` instead of `list(self._agents)` seems to at least fix the problem with the tests. I will investigate and explain in more detail once I have time later this weekend.

quaquel:
this can now be closed because it has been fixed with #1960

Issue #1959, opened 2024-01-12
RuntimeError when removing agents from model with new schedulers


quaquel:
**Describe the bug**
WolfSheep and similar models where agents are removed during `agent.step` throw a Runtime Error:

```python
RuntimeError: dictionary changed size during iteration
```

The cause is quite straightforward: in `AgentSet.do`, one iterates over the agents. If `do` triggers the removal of agents that are in the AgentSet over which you are iterating, you get the error. 

**Expected behavior**
The Runtime error should not occur

**Proposed fix**
Change line 254 in agent.py to `res = [getattr(agent, method_name)(*args, **kwargs) for agent in list(self._agents)]`. You no longer get the error by turning the keys into a list and iterating over this list. This is also identical to how the error was prevented in the [2.1.5 version of the schedulers](https://github.com/projectmesa/mesa/blob/53e8761b47109b18ff77fae78a9024376b14d0e2/mesa/time.py#L106-L112). 

In addition, it would be good to also add an explicit unit test for this error so that it won't happen in the future. Something like below should be all that is needed

```python

MockAgent(agent):
    def step(self):
        self.model.remove()

agents = [MockAgent(i, mockmodel) for i in range(10)]
agentset = AgentSet(agents, mockmodel)
agentset.do("step")

```

**To Reproduce**
Run WolfSheep with the latest stable version of MESA

rht:
An attempt to fix was made at #1958.

quaquel:
I have done some additional diagnostics on this.

Specifically, with respect to the WolfSheep model, there are two ways in which the scheduler can be changed while iterating over it. First, the error happens because a sheep or wolf is born and added to the model. Second,  the error happens when a wolf or sheep dies and thus is removed from the model.

WeakkeyDict runs with a form of protection. See [StackOverflow](https://stackoverflow.com/questions/12428026/safely-iterating-over-weakkeydictionary-and-weakvaluedictionary#:~:text=It%20is%20actually%20safe%20to,the%20docs%20never%20got%20updated.) for more details. If I change, in the example model, the code from `self.model.schedule.remove` to `self.remove`, this prevents errors because a wolf or sheep has died. However, this proper fix is not backward compatible and can only be relied upon partially.

What remains, thus, is the addition of new agents while iterating and casting the keys to list or copying them before iterating solves this problem but can trigger other errors as discussed in #1958.

[...]

PR #1960, opened 2024-01-13
Bugfix in agentset to handle addition and removal correctly

quaquel:
Fix for #1959 and additional unittests that test specifically on this

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -252,7 +252,7 @@ class AgentSet(MutableSet, Sequence):
             AgentSet | list[Any]: The results of the method calls if return_results is True, otherwise the AgentSet itself.
         """
         # we iterate over the actual weakref keys and check if weakref is alive before calling the method
-        res = [getattr(agent(), method_name)(*args, **kwargs) for agent in self._agents.keyrefs() if agent()]
+        res = [getattr(agentref(), method_name)(*args, **kwargs) for agentref in self._agents.keyrefs() if agentref()]
 
         return res if return_results else self
 
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -252,7 +252,11 @@ class AgentSet(MutableSet, Sequence):
             AgentSet | list[Any]: The results of the method calls if return_results is True, otherwise the AgentSet itself.
         """
         # we iterate over the actual weakref keys and check if weakref is alive before calling the method
-        res = [getattr(agentref(), method_name)(*args, **kwargs) for agentref in self._agents.keyrefs() if agentref()]
+        res = [
+            getattr(agentref(), method_name)(*args, **kwargs)
+            for agentref in self._agents.keyrefs()
+            if agentref()
+        ]
 
         return res if return_results else self
 
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -251,7 +251,8 @@ class AgentSet(MutableSet, Sequence):
         Returns:
             AgentSet | list[Any]: The results of the method calls if return_results is True, otherwise the AgentSet itself.
         """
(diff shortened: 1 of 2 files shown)


mesa/agent.py
```diff
@@ -251,7 +251,8 @@ def do(
         Returns:
             AgentSet | list[Any]: The results of the method calls if return_results is True, otherwise the AgentSet itself.
         """
-        res = [getattr(agent, method_name)(*args, **kwargs) for agent in self._agents]
+        # we iterate over the actual weakref keys and check if weakref is alive before calling the method
+        res = [getattr(agent(), method_name)(*args, **kwargs) for agent in self._agents.keyrefs() if agent()]
```
EwoutH, line 255: Does this need to be a list comprehension `[]`, or can ir also be a regular comphrension `()`, and that we return `list(res)` if requested.

quaquel: No, it most be a list comprehension because otherwise it becomes a generator and the method does not get executed.

EwoutH: Right, good to know

rht, line 255: I think it'd be clearer to name `agent_ref` instead of `agent`. Because the call `agent_ref()` is the actual agent themselves.

Corvince:
pre-commit.ci autofix

Corvince (approved):
Seems good to me! Thanks for investigating. I must admit I haven't wrapped my head around weakrefs completely. This error went a bit above my head

quaquel:
It was a bit of a learning experience for me as well. I am familiar with weakrefs from Java so that helped. 

By the way, the code could  have been even more elegant by using the walrus operator (first time I came across a reason to use it).

```python
res = [getattr(agent, method_name)(*args, **kwargs) for agentref in self._agents.keyrefs() if (agent := agentref())]
```

Corvince:
> By the way, the code could  have been even more elegant by using the walrus operator (first time I came across a reason to use it).
> 

That's true! I actually stumbled a bit while reading the code thinking about agentref being executed twice. But I didn't think of the walrus! But it's actually used in the space code for a similar list comprehension in get_cell_list_contents

quaquel:
Should this result in a 2.2.1 release?

EwoutH:
In my opinion, yes.

EwoutH:
To inform you, we're currently blocked on our PyPI pipeline being broken (see https://github.com/projectmesa/mesa/discussions/1962). I was supposed to meet with @jackiekazil to resolve that issue, but I unfortunately missed that meeting. Working on it.

Corvince:
There was another "bug" we should fix first imho. `Scheduler.agents_by_type` changed its return type, which is a breaking change. This should be reverted for 2.2 and then we should release a bugfix release.

quaquel:
> There was another "bug" we should fix first imho. `Scheduler.agents_by_type` changed its return type, which is a breaking change. This should be reverted for 2.2 and then we should release a bugfix release.

To make sure I understand: `RandomActivationByType.agents_by_type` is now a `[type, AgentSet]` but it should have been `[type, [agent][int]]`? So, a simple property solution is all that is needed, and moving the current `self.agents_by_type` to `self._agents_by_type`.


```python
@property
def agents_by_type(self):
	agentsbytype = defauldict(dict)
	for k, v in self._agents_by_type.items():
		agentsbytype[k] = {agent:agent.unique_id for agent in v}

	return agentsbytype

```

Corvince:
Yes, but you might add a warning that the return type is deprecated and mesa 3.0 will return an Agentset (which imho makes much more sense)

EwoutH:
> add a warning that the return type is deprecated and mesa 3.0 will return an Agentset

Agreed

PR #1964, opened 2024-01-15
Perf agentset

Corvince:
I fell into a rabbit hole this weekend that I didn't want to fall in anymore, but here we are: I was looking at mesa performance and noticed that the new AgentSet feature took quite some burden on run times. Here are the results on my local machine, manually running 3 model runs for each (so not completely accurate, but indicative enough). I also included the results for this PR

| Mesa Version| Schelling 100x100 for 100 steps |
| --- | --- |
| Mesa 2.1.5 | ~11.8 sec |
| Mesa 2.2.0 | ~18.2 sec|
| Mesa-PR | ~8.9 sec | 

So I think AgentSet had some noticable effect, but through some modifications to AgentSet and also some stuff in Space.py I was able to increase overall performance. 

I think the changes are non-intrusive, but here they are

## Agentset 
I first inherit the weakref from a normal dict instead of adding all agents sequentially
I also added a short-cut for the select function. If you call `AgentSet.select()` you immediately receive a copy the agentset, without triggering any logic for filter-func, etc. The scheduler does this, so it applies to all models using the scheduler.

## Space
I modified iter_cell_list_contents to retrieve the default_val of the space only once. I also rewrote the generator expression to a normal generator, because perfomance is about the same, but it is easier to understand.
Additionally iter_neighbors now returns the content directly instead of the call to iter_cell_list_contents, which is annotated by accept_tuple_arguments, which costed some performance even though we knew its a list of positions. The difference might appear small, but in my model run this generator gets called 6 million times, so even a microsecond faster execution translates to a whole second. 

## Observation: Weakref
With these modifications the use of Weakref became the new bottleneck. Since weakKeyDictionaries are implemented in pure python and not in C, they considerable slower than Agent dictionaries. But I think it is still manageable.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -8,6 +8,7 @@ Core Objects: Agent
 from __future__ import annotations
 
 import contextlib
+import copy
 import operator
 import warnings
 import weakref
@@ -124,9 +125,7 @@ class AgentSet(MutableSet, Sequence):
                 stacklevel=2,
             )
 
-        self._agents = weakref.WeakKeyDictionary()
-        for agent in agents:
-            self._agents[agent] = None
+        self._agents = weakref.WeakKeyDictionary({agent: None for agent in agents})
 
     def __len__(self) -> int:
         """Return the number of agents in the AgentSet."""
@@ -161,20 +160,21 @@ class AgentSet(MutableSet, Sequence):
             AgentSet: A new AgentSet containing the selected agents, unless inplace is True, in which case the current AgentSet is updated.
         """
 
-        def agent_generator():
+        if filter_func is None and agent_type is None and n == 0:
+            return self if inplace else copy.copy(self)
+
+        def agent_generator(filter_func=None, agent_type=None, n=0):
             count = 0
             for agent in self:
-                if filter_func and not filter_func(agent):
-                    continue
-                if agent_type and not isinstance(agent, agent_type):
-                    continue
-                yield agent
-                count += 1
-                # default of n is zero, zo evaluates to False
-                if n and count >= n:
-                    break
+                if (not filter_func or filter_func(agent)) and (
+                    not agent_type or isinstance(agent, agent_type)
+                ):
+                    yield agent
+                    count += 1
+                    if 0 < n <= count:
+                        break
 
-        agents = agent_generator()
+        agents = agent_generator(filter_func, agent_type, n)
 
         return AgentSet(agents, self.model) if not inplace else self._update(agents)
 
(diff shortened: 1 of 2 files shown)


EwoutH:
Thanks a lot for this effort!

> I also added a short-cut for the select function. If you call `AgentSet.select()` you immediately receive a copy the agentset, without triggering any logic for filter-func, etc. The scheduler does this, so it applies to all models using the scheduler.

I honestly thought we did this already. This will probably be the largest speedup, right?

mesa/agent.py
```diff
@@ -161,23 +160,35 @@ def select(
...
+                    count += 1
+                    if 0 < n <= count:
+                        break
 
-        agents = agent_generator()
+        agents = agent_generator(filter_func, agent_type, n)
 
         return AgentSet(agents, self.model) if not inplace else self._update(agents)
 
+    def _agent_generator(self, filter_func=None, agent_type=None, n=0):
```
rht, line 181: This seems to be identical to `agent_generator`. And if there is a way to deduplicate the definition.

mesa/space.py
```diff
@@ -1045,6 +1046,17 @@ def remove_agent(self, agent: Agent) -> None:
...
         agent.pos = None
 
+    def iter_neighbors(
+        self,
+        pos: Coordinate,
+        moore: bool,
+        include_center: bool = False,
+        radius: int = 1,
+    ) -> Iterator[Agent]:
+        return itertools.chain.from_iterable(
```
rht, line 1056: While I can see caching `default_val()` as a speedup, how does this line speed up the code?

rht:
> Mesa-PR	~8.9 sec

I'd be curious to know if the PR can be broken down into Mesa-PR-1st-commit and Mesa-PR-2nd-commit. But only if you have the time. It seems to me that the AgentSet commit is the one that provides the most speedup, but not sure.

quaquel:
Thanks for this. I saw this rabbit hole last week as well. I am glad you went down it and came up with some solutions. I'll try to find time for a closer look at the code later today.

mesa/space.py
Corvince: This just chains the iterable from the parent function - which is already cache-optimized. So now we don't have to call iter_cell_list_contents, which is wrapped in an annotation

mesa/agent.py
Corvince: whoopsie, you are correct, this shouldn't be here anymore. Removed it now

Corvince:
> > Mesa-PR	~8.9 sec
> 
> I'd be curious to know if the PR can be broken down into Mesa-PR-1st-commit and Mesa-PR-2nd-commit. But only if you have the time. It seems to me that the AgentSet commit is the one that provides the most speedup, but not sure.

Tried it just now, but my laptop seems to be doing some stuff in the background, the results are very flaky. So just from my memory I started out with the modifications to space.py and  that imroved things by about 2 seconds. The rest was indeed from Agentset, with dict update vs dict set about 1 second differene and the rest coming from the shortcut and the updated agent_generator

quaquel:
I looked through the code, and all changes seem sensible and non-intrusive.

EwoutH (approved):
Approving based on @quaquel's insights.

@rht could you do final review and merge if satisfied?

rht (approved):

PR #1965, opened 2024-01-15
Make RandomActivationByType.agents_by_type backward compatible

quaquel:
As #pointed out by @Corvince, `RandomActivationByType.agents_by_typ`e in 2.2 behaved differently from 2.1.5. This is a simple fix for this, with a warning added to raise awareness of future changes.

### DIFF
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -28,6 +28,7 @@ from __future__ import annotations
 import heapq
 import warnings
 import weakref
+from collections import defaultdict
 from collections.abc import Iterable
 
 # mypy
@@ -287,6 +288,22 @@ class RandomActivationByType(BaseScheduler):
         - get_type_count: Returns the count of agents of a specific type.
     """
 
+    @property
+    def agents_by_type(self):
+        warnings.warn(
+            "Because of the shift to using AgentSet, in the future this attribute will return a dict with"
+            "type as key as AgentSet as value",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+
+        agentsbytype = defaultdict(dict)
+        for k, v in self._agents_by_type.items():
+            agentsbytype[k] = {agent: agent.unique_id for agent in v}
+
+        return agentsbytype
+
+
     def __init__(self, model: Model, agents: Iterable[Agent] | None = None) -> None:
         super().__init__(model, agents)
         """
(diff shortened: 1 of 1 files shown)


quaquel:
pre-commit.ci autofix

EwoutH:
Thanks for this. Can we add an optional keyword argument named something like `return_agentset`, that's now `False` by default and emits the deprecation warning? Because then users can already set it to `True` to get the new behavior.

Or should users migrate to `model.agents`? But then if we want to keep schedulers it can mean that the scheduler only has a subset of agents, and thus isn't equivalent to `model.agents`.

(in general, when deprecating something a replacement should be available)

quaquel:
> Thanks for this. Can we add an optional keyword argument named something like `return_agentset`, that's now `False` by default and emits the deprecation warning? Because then users can already set it to `True` to get the new behavior.
> 
> Or should users migrate to `model.agents`? But then if we want to keep schedulers it can mean that the scheduler only has a subset of agents, and thus isn't equivalent to `model.agents`.
> 
> (in general, when deprecating something a replacement should be available)

It's a property so it behaves like an attribute from the outside (as in 2.1.5). The actual data is available via `_agents_by_type`. We might include this information in the warning.

EwoutH:
Adding that information to the warning sounds good!

Corvince:
Thanks a lot @quaquel ! I approved the PR, but noticed that we have the same Problem for `schedule.agents`, which now also returns an AgentSet instead of list[Agent]. I am afraid we have to change this one as well. You can decide if you want to make it part of this PR or a new one

Corvince (approved):

quaquel:
I'll fix it here as soon as possible (have some meetings coming up)

quaquel:
> Thanks a lot @quaquel ! I approved the PR, but noticed that we have the same Problem for `schedule.agents`, which now also returns an AgentSet instead of list[Agent]. I am afraid we have to change this one as well. You can decide if you want to make it part of this PR or a new one

Actually, I'm not sure this needs fixing at all. We discussed it earlier in the original PR. AgentSet is a sequence, so it behaves like a list.

quaquel:
What dunder method would be called when you do that?

EwoutH (approved):

PR #1968, opened 2024-01-16
Run release workflow on the creation of a tag

EwoutH:
Currently the release workflow doesn't run on the creation of a tag, and therefore the wheel isn't deployed to PyPI. This PR should fix that.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,8 @@
 name: Release
 on:
   push:
+    tags:
+    - '*'
     branches:
     - main
     - release**


EwoutH:
Note: I [forced](https://github.com/projectmesa/mesa/actions/runs/7541496990/job/20528344215) a run for the 2.2.1 release, so this PR isn't blocking that. 2.2.1 is successfully released and deployed to PyPI.

.github/workflows/release.yml
```diff
@@ -1,6 +1,8 @@
 name: Release
 on:
   push:
+    tags:
+    - '*'
```
rht, line 5: What about restricting to 'v*' for release tags?

rht: I checked the output of `git tag`. They all are release tags, so maybe it should be fine.

rht (approved):

PR #1971, opened 2024-01-17
run release workflow only when tag is created

wang-boyu:
Do not run release workflow when
- there is a push to main/release branch, or
- there is a pull request

I'm not very familiar with GitHub Actions so a review will be appreciated.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,16 +3,6 @@ on:
   push:
     tags:
     - '*'
-    branches:
-    - main
-    - release**
-    paths-ignore:
-      - '**.md'
-      - '**.rst'
-  pull_request:
-    paths-ignore:
-      - '**.md'
-      - '**.rst'
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -3,16 +3,6 @@ on:
   push:
     tags:
     - '*'
-    branches:
```
rht, line 6: The reason these are enabled is that so that they (building the wheel) are run for every PR. Then, the publishing to PyPI happens only when a tag is created.

wang-boyu: Should the building part be inside the `build_lint` workflow then? So that this workflow only cares about publishing.

This PR is created because the manually triggered release workflow uses a different file: https://github.com/projectmesa/mesa/actions/runs/7541496990/workflow (see also https://github.com/projectmesa/mesa/commit/15e064ead82e3483122b86394c6c197df017736e and https://github.com/projectmesa/mesa/commit/dce482e86710cdb79f5cab475b042d29a5e6ac1e). I guess there are some difficulties in checking if a tag is created during jobs, instead of triggers.

wang-boyu: @EwoutH pls feel free to correct me if I'm wrong : )

EwoutH: Let me explain what happened:
1. Normally, a wheel is built (to test) on every push and pull.
2. On a tag, that wheel should also be deployed to PyPI
3. However, the workflow didn't get triggered on a tag, the wheel wasn't uploaded to PyPI. #1971 should fixed that.
4. So, for 2.2.1, I "forced" a deploy to PyPI by removing the "only deploy on tag"-requirement on a temporary branch, and running it exactly once. That's why that workflow file is different.

So with #1971 merged, I think the workflow is already configured as it should be, and this PR is not necessary. It's useful to test wheel building regularly.

wang-boyu: Thanks for the info. Let me go ahead and close this PR.

EwoutH: Thanks for looking into it anyway!

PR #1972, opened 2024-01-18
feat: Implement Slider class for JupyterViz

rht:
This provides API parity with the previous Tornado viz. Supersedes #1945.
For the Schelling experimental example, from
```python
model_params = {
    "density": {
        "type": "SliderFloat",
        "value": 0.8,
        "label": "Agent density",
        "min": 0.1,
        "max": 1.0,
        "step": 0.1,
    },
    "minority_pc": {
        "type": "SliderFloat",
        "value": 0.2,
        "label": "Fraction minority",
        "min": 0.0,
        "max": 1.0,
        "step": 0.05,
    },
    "homophily": {
        "type": "SliderInt",
        "value": 3,
        "label": "Homophily",
        "min": 0,
        "max": 8,
        "step": 1,
    },
    "width": 20,
    "height": 20,
}
```
to
```python
model_params = {
    "density": Slider("Agent density", 0.8, 0.1, 1.0, 0.1),
    "minority_pc": Slider("Fraction minority", 0.2, 0.0, 1.0, 0.05),
    "homophily": Slider("Homophily", 3, 0, 8, 1),
    "width": 20,
    "height": 20,
}
```
I haven't added test because still waiting for agreement. But I have tested this manually on the Schelling experimental example.

### DIFF
--- a/mesa/experimental/UserParam.py
+++ b/mesa/experimental/UserParam.py
@@ -1,0 +1,56 @@
+class UserParam:
+    _ERROR_MESSAGE = "Missing or malformed inputs for '{}' Option '{}'"
+
+    def maybe_raise_error(self, param_type, valid):
+        if valid:
+            return
+        msg = self._ERROR_MESSAGE.format(param_type, self.label)
+        raise ValueError(msg)
+
+
+class Slider(UserParam):
+    """
+    A number-based slider input with settable increment.
+
+    Example:
+
+    slider_option = Slider("My Slider", value=123, min=10, max=200, step=0.1)
+
+    Args:
+        label: The displayed label in the UI
+        value: The initial value of the slider
+        min: The minimum possible value of the slider
+        max: The maximum possible value of the slider
+        step: The step between min and max for a range of possible values
+        dtype: either int or float
+    """
+
+    def __init__(
+        self,
+        label="",
+        value=None,
+        min=None,
+        max=None,
+        step=1,
+        dtype=None,
+    ):
+        self.label = label
+        self.value = value
+        self.min = min
+        self.max = max
+        self.step = step
+
+        # Validate option type to make sure values are supplied properly
+        valid = not (self.value is None or self.min is None or self.max is None)
+        self.maybe_raise_error("slider", valid)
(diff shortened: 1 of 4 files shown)


quaquel:
I really like this API.

rht:
Added test. This PR is ready for review.

EwoutH (commented):
Really like the API, this is good stuff.

Do we want to drop the explicit distinction between float and int? Is that implied with the values (mainly step) that you input now?

Will try to do a full review later today.

quaquel:
> Do we want to drop the explicit distinction between float and int? Is that implied with the values (mainly step) that you input now?

I am fine with dropping the distinction, but I can see arguments to have explicit int and float sliders. If desired, one possibility would be to just add a dtype kwarg to optionally control this behavior. If dtype is none, you derive the dtype from the arguments.

rht:
Added an optional dtype argument.

EwoutH:
@ankitk50 since you’re also working of visualization, what’s your take?

ankitk50:
I would agree with being explicit with float/int data type. This would also later cater to the possibility of a custom slider with strings` ['min', 'mid', 'max']`.

quaquel:
With the optional dtype argument, it should be quite easy to expand to categorical sliders with string labels later down the line. I can't fully judge the code because I am unfamiliar with solara. But from an API and extensibility point of view, this looks all fine to me.

rlskoeser:
Really like how much more concise this is to configure.

EwoutH:
Should it have an `initial_value` and a `current_value` (or just `value`)? So that you can reset it to it's original state, if wanted? (@Corvince would you think that is useful?)

rht:
> Should it have an initial_value and a current_value (or just value)? So that you can reset it to it's original state, if wanted?

That would need an extra button "reset to initial value" in order to be usable, and would be out of the scope of this PR. This PR wraps the same number of params in Solara's slider.

quaquel:
I agree with @rth. Let's merge this.

Corvince (approved):

PR #1973, opened 2024-01-18
Add radius argument to NetworkGrid.get_neighbors()

EwoutH:
For some reason `get_neighborhood()` did have the `radius` input argument, but `get_neighbors()` did not have it in the NetworkGrid. This commit resolves that inconsistency and allows `get_neighbors()` to take a radius as input and return the agents in that radius. Tests are included.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -1539,9 +1539,11 @@ class NetworkGrid:
             neighborhood = sorted(neighbors_with_distance.keys())
         return neighborhood
 
-    def get_neighbors(self, node_id: int, include_center: bool = False) -> list[Agent]:
-        """Get all agents in adjacent nodes."""
-        neighborhood = self.get_neighborhood(node_id, include_center)
+    def get_neighbors(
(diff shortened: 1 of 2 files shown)


rht (approved):

rht:
Can't be merged before 2.2.2.

EwoutH:
Thanks, agree shouldn't be included (doesn't break anything though).

We can merge it to the main branch though, it's the development branch. For 2.2.2 we can create a maintenance branch from 2.2.1 and cherry pick what we want in there.

rht:
Needs to notify @jackiekazil that we are doing a branch with cherry picks for a patch release, because #1951 happened.

jackiekazil:
RE: #1951 - I am okay if we change the versioning process -- for the manual build - the version was being pulled from the __init__ ... we can change the process. I just want to insure that both a manual build and an automated build would work. If the auto build failed for some reason, the manual would be the back up now.

PR #1978, opened 2024-01-19
Add performance benchmarking scripts and run in CI

EwoutH:
_Let's see how many attempts this is going to take_

General usage (once this is merged to main):
1. Go to the main branch.
2. Run `global_benchmarks.py`
3. Go to your development branch
4. Run `global_benchmarks.py` again
5. Run `compare_timings.py`

It should output a table like this:

| Model          | Size   | Init time [95% CI]            | Run time [95% CI]         |
|----------------|--------|-------------------------------|---------------------------|
| SchellingModel | small  | 🔵 +5.7% [+1.7%, +10.8%]      | 🔵 +5.8% [+1.1%, +12.0%]  |
| SchellingModel | large  | 🔵 -1.0% [-2.1%, -0.2%]       | 🔵 -0.5% [-2.1%, +1.2%]   |
| WolfSheep      | small  | 🔵 +0.0% [-0.6%, +0.6%]       | 🔵 -0.0% [-0.5%, +0.4%]   |
| WolfSheep      | large  | 🔴 +211.2% [+172.4%, +255.0%] | 🔵 +4.6% [-8.0%, +22.4%]  |
| BoidFlockers   | small  | 🔵 +0.1% [-9.7%, +11.6%]      | 🔵 +3.1% [-0.7%, +7.4%]   |
| BoidFlockers   | large  | 🔵 -11.9% [-25.2%, -1.8%]     | 🟢 -18.2% [-31.6%, -4.7%] |

Positive indicates increate in runtime, negative a decrease.

PR #1983, opened 2024-01-20
Add CI workflow for performance benchmarks

EwoutH:
This PR introduces a new GitHub Actions workflow for performance benchmarking. The workflow is triggered on `pull_request_target` events and can be triggered with a "/rerun-benchmarks" comment in the PR. It should be compatible with PRs from both forks and the main repository. It includes steps for setting up the environment, checking out the PR and main branches, installing dependencies, and running benchmarks on both branches. The results are then compared, encoded, and posted as a comment on the PR.

It's largely based and supersedes https://github.com/projectmesa/mesa/pull/1978 and is reasonably well tested on my own fork in https://github.com/EwoutH/mesa/pull/22. It shouldn't need any additional permissions.

The results is a comment on a PR like this:

Screenshot_349

This PR will start working when it's merged onto the main branch.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,0 +1,77 @@
+name: Performance benchmarks
+
+on:
+  pull_request_target:
+    types: [opened, ready_for_review]
+    branches:
+      - main
+  issue_comment:
+    types: [created]
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  run-benchmarks:
+    if: >
+      github.event_name == 'pull_request_target' ||
+      (github.event_name == 'issue_comment' && contains(github.event.comment.body, '/rerun-benchmarks'))
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout PR branch
+        uses: actions/checkout@v4
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+      - name: Add project directory to PYTHONPATH
+        run: echo "PYTHONPATH=$PYTHONPATH:$(pwd)" >> $GITHUB_ENV
+      - name: Install dependencies
+        run: pip install numpy pandas tqdm tabulate
+      - name: Checkout main branch
+        uses: actions/checkout@v4
+        with:
+          ref: main
+          repository: EwoutH/mesa
+      - name: Run benchmarks on main branch
+        working-directory: benchmarks
+        run: python global_benchmark.py
+      - name: Upload benchmark results
+        uses: actions/upload-artifact@v4
+        with:
+          name: timings-main
+          path: benchmarks/timings_1.pickle
+      - name: Checkout PR branch
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          fetch-depth: 0
+          persist-credentials: false
+          clean: false
+      - name: Download benchmark results
+        uses: actions/download-artifact@v4
+        with:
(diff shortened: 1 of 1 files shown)


rht (approved):

rht:
I might have missed some detail, but how did you fix the permission problem, if you could summarize? This is for my learning purpose.

jackiekazil:
Catching up - this looks like a great addition. Since I have been tracking less -- i am going to defer to others for the review and merge. Let me know if you need anything.

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,77 @@
...
+          python-version: '3.12'
+      - name: Add project directory to PYTHONPATH
+        run: echo "PYTHONPATH=$PYTHONPATH:$(pwd)" >> $GITHUB_ENV
+      - name: Install dependencies
+        run: pip install numpy pandas tqdm tabulate
+      - name: Checkout main branch
+        uses: actions/checkout@v4
+        with:
+          ref: main
+          repository: EwoutH/mesa
```
Corvince, line 36: You forgot to update this line

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,77 @@
...
+
+on:
+  pull_request_target:
+    types: [opened, ready_for_review]
+    branches:
+      - main
+  issue_comment:
+    types: [created]
+
+permissions:
```
Corvince, line 11: Are these needed with pull_request_target?

Corvince (commented):
So happy you got it working! Added two comments

.github/workflows/benchmarks.yml
EwoutH: I think these are the default values, but explicit might be better than implicit here, especially if they ever change the defaults.

tpike3: This does make me a little nervous with "write" permissions. [Did you try it with "none" or "read" ](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs)  or is it even necessary, looking through some other projects with `.github/workflows/benchmarks.yml` I did not see others with permissions but it was a cursory look and there are many variants of repo setups and workflows.

EwoutH: > [Did you try it with "none" or "read" ](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs)or is it even necessary,

Yes, doesn't work. Basically the whole discussion of Friday: https://github.com/projectmesa/mesa/pull/1978#issuecomment-1901104588. This is because there is write permission needed to place a comment.

EwoutH:
> I might have missed some detail, but how did you fix the permission problem, if you could summarize? This is for my learning purpose.

Using the [`pull_request_target`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target). That does have write permissions, but only initiated from the main branch of a repo. So it only takes effect once it's merged into main, protecting it from attacks on CI infrastructure.

.github/workflows/benchmarks.yml
EwoutH: Good catch, done

EwoutH:
So let's see if this works: Let's see if the CI is working: /rerun-benchmarks

EwoutH:
It works! Still a bit noisy, but enough to get a general picture.

For future improvements we could:
- Ensure seeds are correctly working
- Run on a own, custom runner without hyperthreading, clock speed turbo boosting and conflicting tasks
- Optimize configurations, replications and statistical methods
- Add models that test relevant components

EwoutH:
Note to self: It looks like the comment-triggerd workflows can't find the PR branch correctly. Needs a fix.

PR #1984, opened 2024-01-21
fix: Explicitly specify JupyterViz space view limits

### LINKED ISSUE #1977: JupyterViz space view limits
At the moment, when drawing spaces with JupyterViz, the axes limits are left to be automatically determined by Matplotlib. Would it not make more sense (in particular of continuous spaces, to derive xlim and xmax from the space itself?

The simplest fix would be to adapt `_draw_grid` to use `space.width` and `space.height` while `_draw_continuos_space` uses `space.x_min`, `space.x_max`, `space.y_min`, and `space.y_max`. Are there reasons where the user might want to rely on matpltolib's automatic determination of the bounds? If so, then a slightly more sophisticated solution would be needed where kwargs are passed to from jupyterviz to the underlying space_drawer function. So you would add a `space_drawer_kwargs` keyword argument.

rht:
Fixes #1977. Manually tested on the Schelling (regular 2D grid) and Boid flocker (continuous) example.

### DIFF
--- a/mesa/experimental/components/matplotlib.py
+++ b/mesa/experimental/components/matplotlib.py
@@ -55,6 +55,8 @@ def _draw_grid(space, space_ax, agent_portrayal):
             out["c"] = c
         return out
 
+    space_ax.set_xlim(-1, space.width)
+    space_ax.set_ylim(-1, space.height)
     space_ax.scatter(**portray(space))
 
 
@@ -90,6 +92,14 @@ def _draw_continuous_space(space, space_ax, agent_portrayal):
         if len(c) > 0:
             out["c"] = c
         return out
+
+    width = space.x_max - space.x_min
+    x_padding = width / 20
+    height = space.y_max - space.y_min
+    y_padding = height / 20
+    space_ax.set_xlim(space.x_min - x_padding, space.x_max + x_padding)
+    space_ax.set_ylim(space.y_min - y_padding, space.y_max + y_padding)
+    space_ax.scatter(**portray(space))
 
     space_ax.scatter(**portray(space))
 


mesa/experimental/components/matplotlib.py
```diff
@@ -55,6 +55,8 @@ def portray(g):
             out["c"] = c
         return out
 
+    space_ax.set_xlim(-1, space.width)
```
rht, line 58: This has to start at -1, because otherwise the markers at position 0 are cropped out.

quaquel: Probably go with `space.width+1` as well for the same reason?

and for `_draw_continuous_space` (which is were my problem started)

```python
space_ax.scatter(**portray(space))
space_ax.set_xlim(xmin=space.x_min, xmax=space.x_max)
space_ax.set_ylim(ymin=space.y_min, ymax=space.y_max)
```

quaquel:
Code looks good to me.

mesa/experimental/components/matplotlib.py
rht: `space.width+1` is not necessary, because the position is 0-index based, so technically it already has +1 in it by doing `space.width` instead of `space.width-1`.

EwoutH:
For clarity, could you add a before and after screenshot?

rht:
There is no perceptible change in the before and after screenshot for both the Schelling and Boid flocker, and hence why I didn't upload anything. However, I have tested the negative cases by intentionally setting the xlim/ylim to (-1, 2), and the displayed scatter plots are cut off as expected. It's standard Matplotlib.

quaquel:
> For clarity, could you add a before and after screenshot?

I am making an exam model with a random walk in continuous space. Here this makes a big difference. If I have time, I'll try to make a short movie.

quaquel:
old behavior:
Screenshot 2024-01-22 at 10 37 52

new behavior
Screenshot 2024-01-22 at 10 37 20

EwoutH:
@quaquel Thanks! I have difficulty seeing what I should notice (as differences) here.

rht:
The plot on the right side is a guide of how far the positions should be relative to the [0, 1] range. As in, they should be too close to the top/right edge of the space.

EwoutH:
Right, it's about the blue dots representing the agents. Why isn't anything like a grid, axis or canvas visible that represents the space? Now the agents just sit there in total isolation.

rht:
> Why isn't anything like a grid, axis or canvas visible that represents the space? Now the agents just sit there in total isolation.

A grid is misleading, because then it would look like a discrete grid. But a box border or axis would make sense.

EwoutH:
What about a light grey background representing the total space? With maybe a hard black border if torus=False and no border (or a dotted one?) if torus=True.

rht:
I think it is out of scope for this PR, but we can continue discussing after the PR has been merged.

EwoutH:
@quaquel you have validated this fixes your problem, without know side-effects or breaking changes as of now?

If so, @rht go ahead and merge.

quaquel:
@EwoutH, I can more easily show you in person.  The current problem is that ax.scatter autoscales (default matplotlib behavior), so when you have agents moving in space, you don't properly see their movement because the view limits change each step. You see the movement properly by fixing the view limits based on the space limit. And yes, this fix works as intended and no side effects (it really is just a few matplotlib commands, don't overthink this).

I personally would like to see just the spines of the axes and, in the case of a continuous grid, which is my example, perhaps some ticks. But I agree with @rht; this is probably beyond this simple PR, which is really just a bug fix.

EwoutH (approved):

EwoutH:
> so when you have agents moving in space, you don't properly see their movement because the view limits change each step.

I now understand the problem. @rht go ahead with merging if you want.

I will try a simple PR for a background color and some axis.

Issue #1900, opened 2023-12-06
Proposal: Formal neighborhood definition for grids


EwoutH:
When implementing #1898, I was inherently playing with spaces. A lot of times closeness (as a spatial relationship) is important, and a lot of time, there is a hard cut-off with a binary result: You are either a neighbor or not.

We have a few nice mechanisms built-in the current neighbor logic:
- `moore`: If True, return Moore neighborhood (including diagonals)
               If False, return Von Neumann neighborhood (exclude diagonals)
- `include_center`: If True, return the (x, y) cell as well. Otherwise, return surrounding cells only.
- `radius`: radius, in cells, of neighborhood to get.

However, it's a bit tedious to pass all those specifications each time you want to use a neighborhood of something. So therefor I would like to have some formal neighborhood specification, so that you can pass one object or variable throughout Mesa and knowing consistently what you need to send and what you get.

EwoutH:
Now really full solution of this, but some thoughts:
 - In it's simplest form, a neighborhood definition can just be a dictionary:
  ```Python
  neighborhood_def = {include_center: False, radius: 1, moore: True}
  ```
 - It could also be a simple object.
  ```Python
  class NeighborhoodDefinition:
      def __init__(self, include_center=False, radius=1, moore=None, grid_type=None):
          self.include_center = include_center
          self.radius = radius
          self.moore = moore
          self.grid_type = grid_type
  ```
   - A dataclass could make this more elegant
 - However, we have to consider different grid types (like hex and network) that take different keywords 
 - Each grid could have a `get_neighborhood_definition()` method that gives a definition that can be used in all methods of that grid type (whether it's a dict, object or something else).

Corvince:
But this issue has let me to think about how we define theesa space API and if we should switch to a more "chainable" API?

So for example instead of doing `grid.get_neighbors(pos, moore, radius, center)`

We do something like this
```python
grid[pos] # get the current position (a Position)
    .neighborhood(moore, radius, center) # get the neighborhood (a PositionList)
    .agents # get the agents (an AgentList)
```

So we would have all those intermediate classes with their own attributes and methods. 

In this case it might seem a bit verbose, but I think especially combined with properties we can have a much cleaner and more powerful way of extracting the things we want. I mean to say that we can have simpler functions as building blocks for complex logic and not having highly specialized functions like `move_agent_to_extreme_value_cell`.

What do you think?

EwoutH:
Thanks for the insights!

[...]

PR #1985, opened 2024-01-21
Minor edits to benchmarking code

quaquel:
* fix for < python 3.12
* Insert the current repo of mesa into sys.path to ensure you are using the correct version
* pep8 name changes of files
* added code to run each benchmark in isolation (useful for profiling and line profiling, used it extensively already for https://github.com/projectmesa/mesa/issues/1900#issuecomment-1902643678)


EwoutH:
Let's see if the CI is working: /rerun-benchmarks

.gitignore
```diff
@@ -88,3 +88,5 @@ pythonenv*/
 # JS dependencies
 mesa/visualization/templates/external/
 mesa/visualization/templates/js/external/
+*.pclprof
+*.pickle
```
EwoutH, line 92: On line 2 `benchmarking/**/*.pickle` is already defined. Is it needed to ignore pickles everywhere?

EwoutH:
Let me know when this is ready for review

.gitignore
quaquel: I thought, too, that you added them to .gitignore, but somehow it was not picked up on my machine. I will take a closer look.

I also have a few more minor changes coming due to the cell space comparison (see also https://github.com/projectmesa/mesa/issues/1900#issuecomment-1902643678)

EwoutH: Maybe rebase your branch on main, it looks like it's lagging a bit behind.

quaquel:
> Let me know when this is ready for review

It should be alright now. Please squash and merge this because I made a bit of a mess of the commit history.

EwoutH:
Could you update the PR message with a bit of motivation for the changes made?

/rerun-benchmarks

quaquel:
> Could you update the PR message with a bit of motivation for the changes made?
> 
> /rerun-benchmarks

I already did so, or do you want to know more? But if so, what?

benchmarks/WolfSheep/__init__.py
```diff
@@ -0,0 +1,14 @@
+from wolf_sheep import WolfSheep
```
quaquel, line 1:
```suggestion
from wolf_sheep import WolfSheep
```

.gitignore
EwoutH: It should indeed have been `benchmarks/**/*.pickle`, good catch

EwoutH:
LGTM, thanks for this effort. @rht or @Corvince could one of you do the final review and squash merge if ok?

Did you also observe the Schelling being faster on your system? Or is that an anomaly?

In the long run we should try to keep these models in sync with [ABMFrameworksComparison](https://github.com/JuliaDynamics/ABMFrameworksComparison). Do we want to make further improvements on these models in the short run?

rht:
> In the long run we should try to keep these models in sync with [ABMFrameworksComparison](https://github.com/JuliaDynamics/ABMFrameworksComparison). Do we want to make further improvements on these models in the short run?

The sync with Agents.jl devs needs to happen soon, if possible. I think they would appreciate to know if we are now monitoring the performance as well.

rht:
> Did you also observe the Schelling being faster on your system? Or is that an anomaly?

Haven't tested, but the finding is probably not a blocker to this PR being merged.

quaquel:
> LGTM, thanks for this effort. @rht or @Corvince could one of you do the final review and squash merge if ok?
> 
> Did you also observe the Schelling being faster on your system? Or is that an anomaly?
> 
> In the long run we should try to keep these models in sync with [ABMFrameworksComparison](https://github.com/JuliaDynamics/ABMFrameworksComparison). Do we want to make further improvements on these models in the short run?

1. I think there is still quite some noise in the benchmarks. But one issue with the old schelling code was that it did not use self.random properly in all places. This probably contributes to even more noise.
2. In my view we should indeed keep our versions in sync with [ABMFrameworksComparison](https://github.com/JuliaDynamics/ABMFrameworksComparison). It would be good to push any changes to them in the context of a new stable release of MESA.

I also noticed that, in particular, WolfSheep is not a good Python implementation. There is a lot of unnecessary code duplication going on. I have the impression that some implementations were done to mimic NetLogo's version. Which is fine for a tutorial for people coming from NetLogo. However, I would argue that the benchmark (and most examples) should reflect the Pythonic and Mesa way of doing things. I am happy to push a WolfSheep reimplementation if desired (But also see [https://github.com/quaquel/mesa/blob/cell-space/benchmarks/WolfSheep/wolf_sheep.py](https://github.com/quaquel/mesa/blob/cell-space/benchmarks/WolfSheep/wolf_sheep.py), for the direction in which I would change it.

EwoutH:
Sounds good, would you like to do both (pushing to [ABMFrameworksComparison](https://github.com/JuliaDynamics/ABMFrameworksComparison) and speeding up WolfSheep yourself?

quaquel:
Yes, I'm fine with doing both (although it is more about cleaning up wolfsheep code; I have no idea about performance, but it is probably small).

rht:
I think the first step is to simplify the WolfSheep at https://github.com/projectmesa/mesa-examples, then next all the other WolfSheep's (in Mesa's benchmark and Agents.jl's benchmark) should be updated to be almost the same as this reference implementation.

PR #1989, opened 2024-01-22
fix: Remove duplicate scatter plotting in ContinuousSpace

rht:


EwoutH:
Good catch, I also had included it in #1988

PR #1991, opened 2024-01-23
feat: Implement Altair version of grid visualization

rht:
Demo for Boltzmann wealth model:

https://github.com/projectmesa/mesa/assets/395821/a74aaf4f-1547-41f8-88f5-466500506317

### DIFF
--- a/mesa/experimental/components/altair.py
+++ b/mesa/experimental/components/altair.py
@@ -1,7 +1,6 @@
 import contextlib
 from typing import Optional
 
-import pandas as pd
 import solara
 
 with contextlib.suppress(ImportError):
@@ -36,25 +35,24 @@ def _draw_grid(space, agent_portrayal):
         return all_agent_data
 
     all_agent_data = portray(space)
-    df = pd.DataFrame(all_agent_data)
+    encoding_dict = {
+        # no x-axis label
+        "x": alt.X("x", axis=None, type="ordinal"),
+        # no y-axis label
+        "y": alt.Y("y", axis=None, type="ordinal"),
+    }
+    has_color = "color" in all_agent_data[0]
+    if has_color:
+        encoding_dict["color"] = alt.Color("color", type="nominal")
+    has_size = "size" in all_agent_data[0]
+    if has_size:
+        encoding_dict["size"] = alt.Size("size", type="quantitative")
+
     chart = (
-        alt.Chart(df)
+        alt.Chart(alt.Data(values=all_agent_data), encoding=alt.Encoding(**encoding_dict))
         .mark_point(filled=True)
-        .encode(
-            # no x-axis label
-            x=alt.X("x", axis=None),
-            # no y-axis label
-            y=alt.Y("y", axis=None),
-        )
+        .properties(width=space.width * 15, height=space.height * 15)
         # .configure_view(strokeOpacity=0)  # hide grid/chart lines
     )
-
-    has_color = hasattr(all_agent_data[0], "color")
-    if has_color:
-        chart = chart.encode(color=alt.Color("color"))
-
-    has_size = hasattr(all_agent_data[0], "size")
-    if has_size:
-        chart = chart.encode(size=alt.Size("size"))
 
     return chart
--- a/mesa/experimental/components/altair.py
+++ b/mesa/experimental/components/altair.py
@@ -7,6 +7,7 @@ import solara
 with contextlib.suppress(ImportError):
     import altair as alt
 
+
 @solara.component
 def SpaceAltair(model, agent_portrayal, dependencies: Optional[list[any]] = None):
     space = getattr(model, "grid", None)
--- a/mesa/experimental/components/altair.py
+++ b/mesa/experimental/components/altair.py
@@ -1,0 +1,52 @@
+import contextlib
+from typing import Optional
+
+import pandas as pd
+import solara
+
+with contextlib.suppress(ImportError):
+    import altair as alt
+
+@solara.component
+def SpaceAltair(model, agent_portrayal, dependencies: Optional[list[any]] = None):
+    space = getattr(model, "grid", None)
+    if space is None:
+        # Sometimes the space is defined as model.space instead of model.grid
+        space = model.space
+    chart = _draw_grid(space, agent_portrayal)
+    solara.FigureAltair(chart)
+
+
+def _draw_grid(space, agent_portrayal):
+    def portray(g):
+        all_agent_data = []
+        for content, (x, y) in space.coord_iter():
+            if not content:
+                continue
+            if not hasattr(content, "__iter__"):
+                # Is a single grid
+                content = [content]  # noqa: PLW2901
+            for agent in content:
+                # use all data from agent portrayal, and add x,y coordinates
+                agent_data = agent_portrayal(agent)
+                agent_data["x"] = x
+                agent_data["y"] = y
+                all_agent_data.append(agent_data)
+        return all_agent_data
+
+    all_agent_data = portray(space)
+    df = pd.DataFrame(all_agent_data)
+    chart = (
+        alt.Chart(df)
+        .mark_point(filled=True)
+        .encode(
+            # no x-axis label
+            x=alt.X("x", axis=None),
+            # no y-axis label
+            y=alt.Y("y", axis=None),
+            size=alt.Size("size"),
+            color=alt.Color("color"),
+        )
+        # .configure_view(strokeOpacity=0)  # hide grid/chart lines
+    )
+    return chart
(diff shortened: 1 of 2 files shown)


rht:
The animation jitter could be fixed with a `plt.tight_layout()` equivalent of Altair. I just haven't figured it out yet.

rht:
Fixed for multigrid, where now you may see overlapping markers:
2024-01-22T23:22:59,903471196-05:00

EwoutH:
Thanks for this effort, will try to review it later today.

Also curious what @Corvince and @ankitk50 think

mesa/experimental/components/altair.py
```diff
@@ -0,0 +1,68 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+def _draw_grid(space, agent_portrayal):
+    def portray(g):
+        all_agent_data = []
+        for i in range(space.width):
+            for j in range(space.height):
+                content = space._grid[i][j]
```
Corvince, line 39: You could simplify this to
```for content, x, y in space.coord_iter() ```

**Comment by Corvince** [comment]
Nice and simple solution, but the obvious question is how this relates to https://github.com/projectmesa/mesa/pull/1902

This PR has a agent_portrayal function, but I don't see how it is used. Since altair uses a declarative approach vs an imperative in matplotlib, I think its confusing to have different agent_portrayal functions. But would have to see the agent_portrayal. 

Otherwise drawing agents on the edge itself looks strange to me, but this shouldn't be a deciding factor.

**Comment by rht** [comment]
> Since altair uses a declarative approach vs an imperative in matplotlib, I think its confusing to have different agent_portrayal functions. But would have to see the agent_portrayal.

I need to say that there is no modification needed in https://github.com/projectmesa/mesa-examples/blob/21624513c6ad94faabbb1cb4b97d73b897f9afae/examples/boltzmann_wealth_model_experimental/app.py#L5-L11, i.e. the `agent_portrayal` API is the same for both Matplotlib and Altair version. I just need to specify `space_drawer="altair"`.

File: mesa/experimental/components/altair.py (hunk 1 above)
Comment (rht) [reply mesa/experimental/components/altair.py]: Refactored.

File: mesa/experimental/components/altair.py [hunk 2]
```diff
@@ -0,0 +1,66 @@
...
+        # Sometimes the space is defined as model.space instead of model.grid
+        space = model.space
+    chart = _draw_grid(space, agent_portrayal)
+    solara.FigureAltair(chart)
+
+# _draw_grid is derived from
+# https://github.com/Princeton-CDH/simulating-risk/blob/907c290e12c97b28aa9ce9c80ea7fc52a4f280ae/simulatingrisk/hawkdove/server.py#L114-L171
+# Copyright 2023 The Center for Digital Humanities of Princeton
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
```
EwoutH, line 23: Adding this line requires a review from @jackiekazil @tpike3 @Corvince and maybe even a dev meeting with a vote. I’m totally fine with attribution, but now we’re changing our license to a mixed one.

rht: It's technically a compatible license, with Mesa being Apache 2.0 as well, instead of GPLv3. This is how proper attribution of a code snippet should be, according to ChatGPT 3.5. I have only seen examples of attributing an entire file(s) in the past, so I haven't looked up for any precedence yet.

Corvince:
> > Since altair uses a declarative approach vs an imperative in matplotlib, I think its confusing to have different agent_portrayal functions. But would have to see the agent_portrayal.
> 
> I need to say that there is no modification needed in https://github.com/projectmesa/mesa-examples/blob/21624513c6ad94faabbb1cb4b97d73b897f9afae/examples/boltzmann_wealth_model_experimental/app.py#L5-L11, i.e. the `agent_portrayal` API is the same for both Matplotlib and Altair version. I just need to specify `space_drawer="altair"` in the `JupyterViz` initialization.

Not exactly. For size this works as you described, but for color in matplotlib you set the actual colors (in your example tab:red and tab:blue ). In altair/vega you set the color encoding. So you are just saying give the value "tab:red" one color and the value "tab:blue" another one. The actual colors that are used are just the default categorical ones (lightblue and orange)

mesa/experimental/components/altair.py
rlskoeser: Thanks, all, for being so thoughtful about this. 

I don't think I have capacity right now for a PR (unless there's something tightly scoped I can easily provide), and I think maybe the development has already moved beyond that anyway. (Apologies for the hidden axis thing in my code that tripped you up!)

An acknowledgement in the release notes sounds like a good solution to me.

rht: I am switching to co-authorship in this PR for now.

rht:
I was about to switch the Altair viz color behavior to match Matplotlib (i.e. color means color, instead of category encoded in color), but decided against it, because the underlying code will look more opaque for the users to modify. As such, I'm leaving the code in this PR as is, to limit its scope. The scope is to have a feature parity with Matplotlib JupyterViz's `_draw_grid` as much as possible.

We need something like the 2nd point in https://github.com/projectmesa/mesa/pull/1988#issuecomment-1904528049 that @quaquel raised. This can be in the form of abstracting out the chart definition https://github.com/projectmesa/mesa/pull/1991/files#diff-681727772982fb0400d413fc9253bbc5e1af9b51d5a9d8581f2ab9d39a3cd516R39-R51 to a function that can be specified by the user.

rht:
> The animation jitter could be fixed with a plt.tight_layout() equivalent of Altair. I just haven't figured it out yet.

I fixed this by converting the data from DF to `alt.Data`.

rht:
New version of the video, with no more jitter.

https://github.com/projectmesa/mesa/assets/395821/d0bbaefb-bfa8-43e4-b29e-684bf0ae396f

The only concern in this PR is that the `agent_portrayal` function differs from the Matplotlib version in that color is categorical instead of literal. I have chosen to keep the discrepancy, to adapt to Altair whenever it makes sense instead.

Are there any other concerns?

quaquel:
I don't know why, but I can't view the movie.

rht:
Can we move forward with this PR ASAP? I want to implement #2049 but for Altair and many other things.
There is nothing controversial in this PR that warrants a delay in review for a month. And I'm not pleased with the slowdown in the supposedly experimental feature.

quaquel:
There have been various requests for modifications to the code in #1902. The last response from the author of that PR is from 2 weeks ago and in no way acknowledges the requested changes. I am sympathetic to therefore moving forward with this one instead.

rht:
> I understand your frustration, but for me the main question is still and you haven't commented on this yet, how this relates to https://github.com/projectmesa/mesa/pull/1902 and how to deal with this situation, as I asked on Matrix. I think this is very controversial

@Corvince  I already replied almost immediately on your question on Matrix, on Matrix, more than a week ago.

quaquel:
> I really don't know what to do with the two competing jupyter PRs implementing the same feature. How can we resolve this without devaluing the work of one of the PRs? 

I guess this is @Corvince's question. I looked through your responses @rht, but I don't see a clear answer to this specific question. You did write

>in #1991 i attribute @rlskoeser because that's what it is mainly based on, and we had already made a fanfare about including her in the next mesa release (https://github.com/projectmesa/mesa/pull/1991#discussion_r1463378129). i can't exactly attribute the student (ankit) either, because his pr is mainly based on corvince's mesa-interactive, almost an exact copy. i'm not sure if ankit intentionally ignored my reviews or something else. https://github.com/projectmesa/mesa/pull/1820 seems to suggest the former. it's meant to be a bugfix, where i have outlined a suggestion with a very simple fix (https://github.com/projectmesa/mesa/pull/1820#issuecomment-1774838319), but hasn't been responded to *for ages*. if i shouldn't be frustrated about this, i don't know what else.

So, are you saying that your answer to @Corvince is to just merge this PR and close the other without any acknowledgement to the author of #1902?

rht:
> So, are you saying that your answer to @Corvince is to just merge this PR and close the other without any acknowledgement to the author of https://github.com/projectmesa/mesa/pull/1902?

Yes, mesa-interactive already exists in a separate repo, https://github.com/Corvince/mesa-interactive.

quaquel:
I don't know the backstory to any of this. At face value, merging this PR without any acknowledgement to the author of #1902 seems a s bit harsh even though the author has not replied to various reasonable modifications to the code. What is wrong with a simple mention in the PR message itself of the work done in #1902?

rht:
Thank you for moving this forward. Finally.

Issue #1953, opened 2024-01-11
move_agent_to_neighborhood


rht:
**What's the problem this feature will solve?**
Currently, there are several examples [1] where the agents are supposed to move to a random location within the neighborhood.
This is implemented with a rather verbose class, [`RandomWalker`](https://github.com/projectmesa/mesa-examples/blob/main/examples/wolf_sheep/wolf_sheep/random_walk.py), .
```python
next_moves = self.model.grid.get_neighborhood(self.pos, self.moore, True)
next_move = self.random.choice(next_moves)
# Now move:
self.model.grid.move_agent(self, next_move)
```
If you look at https://github.com/JuliaDynamics/ABMFrameworksComparison/tree/main/WolfSheep/Mesa, the random_walk.py is one of the reason why Mesa has a long LOC.

**Describe the solution you'd like**

It would be great if there is a method `move_agent_to_neighborhood` that encapsulates the random walk procedure defined in the previous section.

[1] bank reserve, Boltzmann wealth model, wolf sheep

quaquel:
This might or might not be a possible direction, but one I am curious to explore if I have a bit of time:

In various contexts, adding a dedicated "patch" or "grid cell" class to MESA has been suggested. This would be a way to give agent-like behavior to the grid (e.g., grass regrowth dynamics). If there is a dedicated grid cell class, you could specify within it, its neighbors (e.g., left, right, top bottom for a Moore neighborhood). You can even do larger neighborhood sizes on top of this by recursively querying these attributes. Moreover, you can eliminate boundary checking when dealing with non-toroidal spaces. 

Long story short, if each grid cell class has a collection of neighbors, the random walk becomes trivial:

```python
neighbors = self.my_gridcell.neighbors(size=1)
next_move = self.random.choice(neighbors)
```

Corvince:
> If you look at https://github.com/JuliaDynamics/ABMFrameworksComparison/tree/main/WolfSheep/Mesa, the random_walk.py is one of the reason why Mesa has a long LOC.

In my opinion its a bit idiotic to rate LOC, but if you are concerned with respect to the framework comparison you could submit a PR where it is condensed in a single line:
```python
self.model.grid.move_agent(self, self.random.choic(self.model.grid.get_neighborhood(self.pos, True, 1)))
```
And rermove the RandomWalker class. For me it shows how you can easily define a base class for your agents, but in a comparison that counts LOC it is probably too verbose.

[...]

Issue #1903, opened 2023-12-11
Proposal: Flexible move_agent method that allows multiple movement strategies


EwoutH:
The Mesa Space module allows a few specific agent movements and interactions within a grid environment. However, the current methods lack a consistent way to apply different movement strategies. This proposal suggests integrating a single, versatile `move_agent` method, accepting a parameter that can either be a tuple for specific coordinates, a string indicating a movement strategy (like "random" or "empty"), or an object defining a neighborhood.

In the future, it can be extended with movement strategies based on properties (see #1898).

#### Proposed Method Enhancement
1. **Unified Move Method**:
   - `move_agent(agent: Agent, destination) → None`
   - The `destination` parameter is versatile:
     - It can be a tuple `(x, y)` for moving the agent to specific coordinates.
     - It can be a string, such as `"random"` for a random cell, `"empty"` for a random empty cell.
     - It can be an object or a dictionary defining a neighborhood, allowing custom definitions of neighboring cells. For this we need a formal neighborhood definition, see #1900.

2. **Retiring Redundant Methods**:
   - Methods like `move_to_empty` and `place_agent` would be redundant and can be removed, as their functionalities are integrated into the new `move_agent`.

3. **Enhanced Out-of-Bounds and Validity Checking**:
   - Maintain `out_of_bounds` method, but possibly enhance it to include checks for cell occupancy, ensuring valid movement destinations.

4. **Agent Removal and Position Swapping**:
   - `remove_agent` and `swap_pos` methods remain useful and unchanged.

#### Example Implementations
- **Moving to a Specific Cell**:
  ```python
  space.move_agent(agent, (x, y))
  ```
- **Moving to a Random Cell**:
  ```python
  space.move_agent(agent, "random")
  ```
- **Moving to a Random Empty Cell**:
  ```python
  space.move_agent(agent, "empty")
  ```
- **Moving to a Custom-Defined Neighborhood**:
   (see #1900)
  ```python
  neighborhood_def = {"type": "Moore", "radius": 2}
  space.move_agent(agent, neighborhood_def)
  ```

#### Conclusion
This proposal aims to simplify and unify the movement methods in the Mesa Space module. By consolidating various movement strategies into a single method, we enhance the API's usability and flexibility, allowing users to execute complex movements with minimal and more intuitive code.

EwoutH:
[...]

PR #1994, opened 2024-01-23
Add cell-centric discrete spaces (experimental)

Corvince:
## Summary
This PR introduces an alteranative implementation for discrete spaces. This implementation centers on the explicit inclusion of a Cell class. Agents can occupy cells. Cells have connections, specifying their neighbors. The resulting classes offer a cell centric API where agents interact with a cell, and query the cell for its neighbors. 

To capture a collection of cells, and their content (_i.e._, Agents), this PR adds a new CellCollection class. This is an immutable collection of cells with convenient attribute accessors to the cells, or their agents. 

This PR also includes a CellAgent class which extends the default Agent class by adding a `move_to` method that works in conjunction with the new discrete spaces. 

From a performance point of view, the current code is a bit slower in building the grid and cell data structure, but in most use cases this increase in time for model initialization will be more than offset by the faster retrieval of neighboring cells and the agents that occupy them.

## Motive
The PR emerged out of various experiments aimed at improving the performance of the current discrete space code. Moreover, it turned out that a cell centric API resolved various open issues (_e.g._, #1900, #1903, #1953). 

## Implementation
The key idea is to have Cells with connections, and using this to generate neighborhoods for a given radius. So all discrete space classes are in essence a [linked data structure](https://en.wikipedia.org/wiki/Linked_data_structure).

The cell centric API idea is used to implement 4 key discrete space classes: OrthogonalMooreGrid, OrthogonalVonNeumannGrid (alternative for SingleGrid and MultiGrid, and moore and von Neumann neighborhood) , HexGrid (alternative for SingleHexGrid and MultiHexGrid), and Network (alternative for NetworkGrid). Cells have a capacity, so there is no longer a need for seperating Single and Multi grids. Moore and von Neumann reflect different neighborhood connections and so are now implemented as seperate classes. 

## Usage examples
Below are a few examples demonstrating the functionality of the cell spaces.

#### Creating a Grid and Adding Agents
```python
from mesa import Model
from mesa.experimental.cell_space import OrthogonalMooreGrid, CellAgent

class MyAgent(CellAgent):

# Initialize a model and grid
model = Model()
grid = OrthogonalMooreGrid([10, 10], torus=True, capacity=1, random=model.random)

# Create an agent and add it to a cell
agent = MyAgent(1, model)
cell = grid.select_random_empty_cell()
agent.move_to(cell)
```

#### Agent Movement
```python
# Assuming `agent` and `grid` are already defined
# Move the agent to a random neighboring cell
new_cell = agent.cell.neighborhood().select_random_cell()
agent.move_to(new_cell)
```

#### Querying the Neighborhood
```python
# Get all neighboring agents within a radius of 1
neighbors = agent.cell.neighborhood(radius=1).agents

# Iterate over neighbors and perform actions
for neighbor in neighbors:
    # Example action: print neighbor's ID
    print(neighbor.unique_id)
```

#### Agent Reproduction
```python
# Agent reproduces, creating a new agent in the same cell
def reproduce(agent):
    offspring = MyAgent(model.next_id(), model)
    offspring.move_to(agent.cell)
    model.schedule.add(offspring)

# Call reproduce function for an agent
reproduce(agent)
```

#### Removing an Agent
```python
# Remove an agent from the simulation
def remove_agent(agent):
    agent.cell.remove_agent(agent)  # Remove from cell
    model.schedule.remove(agent)    # Remove from the scheduler

# Call remove_agent function for an agent
remove_agent(agent)
```

#### Checking Cell Properties
```python
# Check if a cell is empty or full before moving an agent
if cell.is_empty:
    agent.move_to(cell)
elif cell.is_full:
    print("Cell is full. Agent cannot move here.")
```


EwoutH:
There’s _a lot_ in here that I like!

let me know when you want any input / a review.

EwoutH:
If you add `matplotlib` here, the current error in the benchmarks is resolved.

https://github.com/projectmesa/mesa/blob/7b0d3f12f0fb2a0f2dc3ec1bcca4a95a82ff6cee/.github/workflows/benchmarks.yml#L31

Corvince:
> If you add `matplotlib` here, the current error in the benchmarks is resolved.
> 
> https://github.com/projectmesa/mesa/blob/7b0d3f12f0fb2a0f2dc3ec1bcca4a95a82ff6cee/.github/workflows/benchmarks.yml#L31

Thanks, I now tried to bypass the jupyter stuff, but I am not sure if this is working. If I change the benchmark.yml won't the action stop working? Because of the pull_request_target thingy?

EwoutH:
> If I change the benchmark.yml won't the action stop working?

Yeah it might, honestly I don't know.

mesa/agent.py
```diff
@@ -46,6 +47,7 @@ def __init__(self, unique_id: int, model: Model) -> None:
         """
         self.unique_id = unique_id
         self.model = model
+        self.cell: Cell | None = None
```
EwoutH, line 50: Is it logical it has both a Cell and a Position? Maybe it is, or can one be an property of the other. Mostly to prevent them from going out of sync.

mesa/experimental/cell_space.py
```diff
@@ -0,0 +1,204 @@
...
+        """Returns a bool of the contents of a cell."""
+        return len(self.agents) == self.capacity
+
+    def __repr__(self):
+        return f"Cell({self.coordinate}, {self.agents})"
+
+    @cache
+    def neighborhood(self, radius=1, include_center=False):
+        if radius == 0:
+            return {self: self.agents}
```
EwoutH, line 60: So this is just returning agents at the current location? But if include_center=False, that's conceptually incorrect right?

So should we allow it at all? I think for clearness I would be in favor of not allowing radius==0 and making a separate function for it.

EwoutH: Ah I understand, that ends the loop on line 67-68.

Smart, but I'm not sure I fully like it.

mesa/experimental/cell_space.py
```diff
@@ -0,0 +1,204 @@
...
+        else:
+            neighborhood = {}
+            for neighbor in self._connections:
+                neighborhood.update(neighbor.neighorhood(radius - 1, include_center))
+            if not include_center:
+                neighborhood.pop(self, None)
+            return CellCollection(neighborhood)
+
+
+class CellCollection:
```
EwoutH, line 74: We now have AgentSet and CellCollection. Mixed feelings about CellSet.

mesa/experimental/cell_space.py
```diff
@@ -0,0 +1,204 @@
...
+        return itertools.chain.from_iterable(self._cells.values())
+
+    def select_random_cell(self):
+        return random.choice(self.cells)
+
+    def select_random_agent(self):
+        return random.choice(list(self.agents))
+
+    def select(self, filter_func: Optional[Callable[[Cell], bool]] = None, n=0):
+        if filter_func is None and n == 0:
```
EwoutH, line 109: In which case will this happen / be useful?

mesa/experimental/cell_space.py
```diff
@@ -0,0 +1,204 @@
...
+
+    def __iter__(self):
+        return iter(self.cells.values())
+
+    def __getitem__(self, key):
+        return self.cells[key]
+
+    def move_agent(self, agent: Agent, pos: Coordinate) -> None:
+        """Move an agent from its current position to a new position."""
+        self.cells[pos].add_agent(agent)
```
EwoutH, line 139: This is incomplete right?

mesa/experimental/cell_space.py
```diff
@@ -0,0 +1,204 @@
...
+
+    @property
+    def empties(self) -> CellCollection:
+        return self.all_cells.select(lambda cell: cell.is_empty)
+
+    def move_to_empty(self, agent: Agent) -> None:
+        while True:
+            new_cell = self.all_cells.select_random_cell()
```
EwoutH, line 147: This is faster (in most cases) than getting the empty cells and selecting one of those I guess?

Would be interesting what happens with a 90% and 99% full grid.

mesa/experimental/cell_space.py
```diff
@@ -0,0 +1,204 @@
...
+            if new_cell.is_empty:
+                new_cell.add_agent(agent)
+                return
+
+        # TODO: Adjust cutoff value for performance
+        for _ in range(len(self.all_cells) // 10):
+            new_cell = self.all_cells.select_random_cell()
```
EwoutH, line 154: Oh you already looked at the performance. Won't the while loop run forever if no empty cells are found though?

mesa/experimental/cell_space.py
```diff
@@ -0,0 +1,204 @@
...
+
+        try:
+            self.empties.select_random_cell().add_agent(agent)
+        except IndexError as err:
+            raise Exception("ERROR: No empty cell found") from err
+
+
+class Grid(Space):
+    def __init__(
+        self, width: int, height: int, torus: bool = False, moore=True, capacity=1
```
EwoutH, line 167: I would say the default is no capacity restriction.

mesa/experimental/cell_space.py
```diff
@@ -0,0 +1,204 @@
...
+                ( 0, -1),          ( 0, 1),
+                ( 1, -1), ( 1, 0), ( 1, 1),
+            ]
+        else: # Von Neumann neighborhood
+            directions = [
+                          (-1, 0),
+                ( 0, -1),          (0, 1),
+                          ( 1, 0),
+            ]
+        # fmt: on
```
EwoutH, line 197: This is an absolute work of art.

(Didn't know about `# fmt: off`, I'm going to use that everywhere now. Sorry @rht)

EwoutH (commented):
Just some random thoughts from reading the code. On a high level, I think I really like it but I feel like I don't have enough comprehension of all the nuances and implications to say for sure.

Guess I would need to start make models with it to see what happens!

mesa/experimental/cell_space.py
```diff
@@ -0,0 +1,204 @@
...
+        if radius == 1:
+            return CellCollection(
+                {neighbor: neighbor.agents for neighbor in self._connections}
+            )
+        else:
+            neighborhood = {}
+            for neighbor in self._connections:
+                neighborhood.update(neighbor.neighorhood(radius - 1, include_center))
+            if not include_center:
+                neighborhood.pop(self, None)
```
EwoutH, line 70: Does this go right in the Grid? Because the centers aren't included there artwork. So shouldn't it be added if include_center is true, instead of popped if false?

rht:
Had 3 comments at https://github.com/projectmesa/mesa/commit/2e78940728f2b946abcdca571fd4e85343849a50#comments.

Corvince:
@EwoutH it looks like the benchmark wasn't run for this PR after the comment trigger. You can see that from the action. Maybe you can look into that?

EwoutH:
>  it looks like the benchmark wasn't run for this PR after the comment trigger.

You forgot the dash between rerun and benchmarks. It has to be the exact string (see rht's one).

mesa/experimental/cell_space.py
```diff
@@ -0,0 +1,204 @@
...
+        i, j = cell.coordinate
+
+        # fmt: off
+        if self.moore:
+            directions = [
+                (-1, -1), (-1, 0), (-1, 1),
+                ( 0, -1),          ( 0, 1),
+                ( 1, -1), ( 1, 0), ( 1, 1),
+            ]
+        else: # Von Neumann neighborhood
```
rht, line 191: Should be "von Neumann neighborhood".

mesa/experimental/cell_space.py
```diff
@@ -0,0 +1,204 @@
...
+
+    @property
+    def is_full(self) -> bool:
+        """Returns a bool of the contents of a cell."""
+        return len(self.agents) == self.capacity
+
+    def __repr__(self):
+        return f"Cell({self.coordinate}, {self.agents})"
+
+    @cache
```
rht, line 57: I tried `@cache` as a decorator of `get_cell_list_contents` in #1995. It gave the error message that looks like https://github.com/projectmesa/mesa/pull/1995#issuecomment-1907405611. Decorators seem to be no-go in Cython.

EwoutH: Are you sure? Can’t imagine such a large Python feature not being allowed.

mesa/experimental/cell_space.py
rht: Can't be CellSet because the elements in this object is ordered. Some agents might want to actually loop the Cell elements according to the given non-random order, e.g. a bee that scans in clockwise/counterclockwise direction starting from the north direction.

mesa/experimental/cell_space.py
rht: `black` is focused only on IT code, not tailored for scientific code. I know that folks at https://github.com/SRI-International/QC-App-Oriented-Benchmarks resist `black` because of this.

mesa/experimental/cell_space.py
Corvince: Exactly, this is usually the fastest option. But yes once you get an almost full grid it gets increasingly slow and if your grid is full it indeed runs forever. Thats why i added the (currently dead) code below, but it was much slower and didn't have the time yet to look into it more.

mesa/experimental/cell_space.py
Corvince: I have to give ChatGPT credit for the idea, I just adjusted it to look nicer

mesa/agent.py
Corvince: No, its kind of reduntant. Pos could be a property that just links to `cell.coordinate`, but that would break exisiting models.

mesa/experimental/cell_space.py
Corvince: Exactly, its for ending the recursion. I would say radius=0 and include_self=False dont make sense together. Actually radius=0 already doesnt make sense from a user perspective, but is used for the recursion

mesa/experimental/cell_space.py
```diff
@@ -0,0 +1,204 @@
...
+    def neighborhood(self, radius=1, include_center=False):
+        if radius == 0:
+            return {self: self.agents}
+        if radius == 1:
+            return CellCollection(
```
Corvince, line 62: this currently breaks neighborhoods with radius>1 Made some errors while refactoring

mesa/experimental/cell_space.py
Corvince: This is correct. The "center"cell enters the neighborhood indirectly from the neighboring cells (assuming we always have two-way connections between cells)

mesa/experimental/cell_space.py
Corvince: None, it should at least return a new CellCollection in that case. But the function is also broken because n is not considered

Corvince:
> > it looks like the benchmark wasn't run for this PR after the comment trigger.
> 
> You forgot the dash between rerun and benchmarks. It has to be the exact string (see rht's one).

Ha, I see how my comment was completely miswritten. I already realized after @rht made the correct comment. But in the action that was actually triggered this PR isn't checked out, hence no difference in performance. Performance should actually be about 20% faster.

Here is the exact line of the Action. It should check out this PR, but instead creates a new branch "main"
https://github.com/projectmesa/mesa/actions/runs/7632442972/job/20792667533#step:9:149

mesa/experimental/cell_space.py
quaquel: We debated some of this in #1953. I see several solutions

1. Use cell properties on the agent and handle the remove from other cell within the `cell.setter` property. So

```python
# within an agent 
self.cell = self.cell.neighborhood(radius=1).select_random()
```

I like this code. It is clean, expressive and concise. A different version would be to have a `move` method on the agent Either version however breaks fundamentally how MESA has been handling movement in spaces. 

2. Have a separate `move_agent_to(agent:Agent, cell:Cell)` function. This handles the removing from cell and adding to new cell. This is less breaking. It is in line with NetLogo (a reason in my view to not do this :grin:). And is not as potentially tying movement to cells or grids where it does not conceptually belong. 

3. Do it in spaces, as currently done.
4. handle movement within add_agent. Again implicit so bad idea.

mesa/experimental/cell_space.py
quaquel: I have looked into this somewhere last week. It should be possible to make it work, but will look at this again later.

quaquel:
Thanks for this! I will take a deeper dive at this, hopefully, tonight or otherwise over the weekend. I quickly checked the code this morning and have a few additional comments and suggestions in addition to the various points already raised

I will also move my versions of Schelling and WolfSheep into this (do I have write access)? So we can properly see the change in performance.

EwoutH:
> It should check out this PR, but instead creates a new branch "main"

I tried some stuff. Let's see if it works. /rerun-benchmarks

EwoutH:
Here we go: /rerun-benchmarks

EwoutH:
Doesn't work unfortunately. Will research further at a later point. For now, just do the "mark as ready for review" trick.

mesa/experimental/cell_space.py
quaquel: The current draft is also implicitly immutable because of the `@cached_property`.

The AgentSet is also ordered by the way, so that is not really a good reason in my view. 

The direction is interesting but we currently don't explicitly store this information in cell.connections. In my proof of principle, I was storing everything separately [topleft, top., toplight, left, etc.]. The current code is much more concise, cleaner, and it would be rare to need the directions. However, I am still wondering whether connections should not be a dict with the direction as key and the cell as value. You can do stupid stuff like

```python
diagonal = []
cell = self.cell.connections[l"top left"] 
while cell is not self:
    diagonal.append(cell)
    cell = cell.connections["top left"] 
```

Which would give you the diagonal cells on a torus space. This is an example of how the current data structure is like a linked list.

mesa/experimental/cell_space.py
quaquel: I think it is possible to get rid of the radius=1 check but instead raise an exception if radius < 1. I think I might even have the code in that way in my work from last weekend.

EwoutH: What will happen with float radius in the continiousspace? We override the method?

quaquel: We haven't thought about continuous space yet. All this really only works for discrete spaces. So I suggest we don't worry about that yet.

mesa/experimental/cell_space.py
quaquel: The entire issue starts from not wanting a coupling between cells and the grid. In my version, I did allow for such coupling. Basically, cells update their owning grid when they are empty or when they are no longer empty. It makes move_to_empty in the grid simple because it can reuse the old optimized code. 

I understand @Corvince motivation (see https://github.com/projectmesa/mesa/issues/1900#issuecomment-1905592341) but I am not sure whether the price of this is worth it. Also, conceptually, I don't see how cells can exist without an "owning" DiscreteSpace.

mesa/experimental/cell_space.py
```diff
@@ -0,0 +1,204 @@
...
+        return CellCollection(
+            {
+                cell: agents
+                for cell, agents in self._cells.items()
+                if filter_func is None or filter_func(cell)
+            }
+        )
+
+
+class Space:
```
quaquel, line 121: I would rename this to DiscreteSpace to be a bit clearer and is nicer in opposition to ContinuousSpace.

mesa/experimental/cell_space.py
```diff
@@ -0,0 +1,204 @@
...
+    cells: dict[Coordinate, Cell] = {}
+
+    def _connect_single_cell(self, cell):
+        ...
+
+    @cached_property
```
quaquel, line 127: So the grid is immutable.  I agree, but will be important to communicate

Corvince:
> I will also move my versions of Schelling and WolfSheep into this (do I have write access)? So we can properly see the change in performance.

I invited you as a collaborator just now, so you should have write access. Schelling and WolfSheep are already part of this PR, but the benchmarks are currently not working because I moved the CellSpace into experimental. That means on importing cell_space (for the benchmarks) jupyter_viz is also imported and this leads to a unresolved dependency. Quick solution would be to move cell_space into mesas main directory. Then we can properly retrigger the benchmarks and we should finally see the difference.

rht:
I merged #2001. /rerun-benchmarks

mesa/experimental/cell_space.py
quaquel: I tested in on Schelling, radius 2 and seems fine?

mesa/experimental/cell_space.py
```diff
@@ -0,0 +1,247 @@
...
+
+Coordinate = tuple[int, int]
+
+
+class Cell:
+    __slots__ = ["coordinate", "_connections", "agents", "capacity", "properties"]
+
+    def __init__(self, i: int, j: int, capacity: int | None = 1) -> None:
+        self.coordinate = (i, j)
+        self._connections: list[Cell] = []
```
quaquel, line 17: See the discussion elsewhere with @rht, but what about turning this into a dict with the direction of the connection as keys, and the cells as values? So `self._connections["left"]` would return the cell to my left? You can do all kinds of funky stuff with that. it makes row and column retrieval trivial. But you can also express more complex moves (e.g., a knight move from chess) very concisely if you have the direction information on the connections somewhere.

mesa/experimental/cell_space.py
quaquel: its actually trivial. A decorator is just syntactic sugar. So just do

```python
def some_function():
    ....
some_function = cache(some_function)

```

Which is identical to what the decorator does anyway.

mesa/experimental/cell_space.py
```diff
@@ -0,0 +1,247 @@
...
+        try:
+            self.empties.select_random_cell().add_agent(agent)
+        except IndexError as err:
+            raise Exception("ERROR: No empty cell found") from err
+
+
+class Grid(Space):
+    def __init__(
+        self, width: int, height: int, torus: bool = False, moore: bool = True, capacity=1
+    ) -> None:
```
quaquel, line 170: capacity is now very nicely done!. No need for MultiGrid and SingleGrid anymore.

mesa/experimental/cell_space.py
rht: I wonder if the method used in #1942 to ensure that the class is pickle-able
```python
def __init__(...):
   ...
   self._original_sum_function = self.some_function
   self.some_function = cache(self._original_some_function)
```
still works.

mesa/experimental/cell_space.py
rht: I would encode the left-right, down-up in a "Python turtle" agent's move methods, as an interpretation to the connection indices. This interpretation is specific to a 2D rectangular grid. While a cell collection signifying a geographical map wouldn't always have exactly 4 neighbors at all times (west/east neighbor, south/north neighbor), but sometimes would have 1 neighbor (Vatican cell and Italy cell).

quaquel: I think I understand your point. If we decide to go down this route, naming the connections (i.e., the keys in my proposal) should be handled in the `_connect_single_cell`. For a 2D grid and HexGrid, this is relatively easy and unambiguous. With networks or geographical maps, it becomes a bit more difficult.

Corvince: I am sympathetic towards the idea of named connections, but I think it's not required for a first version. Connections is a private variable so we change the data type to a dict later. The reason is that I think finding good names is not so trivial. For example for a knights move is it up, up, left or right, right, up, or any of the other possibilities? It would be unambiguous from an agents point of view (forward, forward, left or right), but i don't see an immediate translation to the directions yet. 

But in principle the idea is super cool, you would even be able to name some long-range connections between distant cells.
Also networks-on-grid would be easy

quaquel:
I have made some choices in light of all the comments and feedback. Feel free to discuss if you disagree. In short:

* added NetworkGrid
* Renamed various classes
* Added an intermediate Grid class from which OrthogonalGrid and HexGrid derive
* Added a CellAgent class. This isolates this experimental code from the rest of MESA. I also added a `move_to` method to this agent to handle movement.
* I changed the handling of empties closer to the old code. This does introduce coupling between a Cell and its owning "Space" with the various drawbacks that it entails. However, `select_random_empty_cell` now works across OrthogoonalGrid, HexGrid, and NetworkGrid.

compared to my code from last week, this is still a bit slower. I'll try to investigate later.

mesa/experimental/cell_space.py
quaquel: Ok, let's leave this for later. 

I guess that there most be some proper terminology from the cellular automata world that we should be able to reuse.

quaquel:
> compared to my code from last week, this is still a bit slower. I'll try to investigate later.

I fixed the performance issues. The code base is mainly complete. @EwoutH, if possible, let's take 5 minutes tomorrow to discuss property grids and how they might be made to work with this. It's not on the critical path (just as with named connections), but if there is a simple solution, it would be nice to have.

random number generation is still a problem. Once #1981 is addressed, we should use that solution here. 

I still want to add at least some tests before even thinking about merging this.

Corvince:
> I have made some choices in light of all the comments and feedback. Feel free to discuss if you disagree. In short:
> 

Thanks for all the changes!

> * Added a CellAgent class. This isolates this experimental code from the rest of MESA. I also added a `move_to` method to this agent to handle movement.

I don't think this isolates it too much, this is something that has been on several wish lists. But if we manage to find an appropriate implementation for ContinousSpace that is also somewhat "cell-based" (im working on this issue) , these changes could also be propagated to the basic Agent. I know there are ABMs without a space component, but its cell-centric methods don't need to be used. and I don't think there will be too many.

However, I wonder how we will go along with distinguishing cells from positions, esp. in such methods. Currently it only supports cells directly, but you would have to go through model.space to get a reference. I guess thats another point for coupling cells to their owners space. 

> * I changed the handling of empties closer to the old code. This does introduce coupling between a Cell and its owning "Space" with the various drawbacks that it entails. However, `select_random_empty_cell` now works across OrthogoonalGrid, HexGrid, and NetworkGrid.

I still have hopes that we don't need this coupling. I think we can use raise your select_random_empty_cell function from NetworkGrid to the space base class, because it should work for all spaces. Then we can have more optimized implementations for the grids. But re: coupling. I don't like that the cells manipulates the owners attributes (_empties) directly. This is already based on some assumptions how the owner is implemented. This is basically a performance optimization, but those shouldn't dictate how we implement our methods. But yes, without it, it can be quite slow. Not sure how to proceed with this one.

benchmarks/WolfSheep/wolf_sheep.py
```diff
@@ -49,55 +149,69 @@ def __init__(
             grass_regrowth_time: How long it takes for a grass patch to regrow
                                  once it is eaten
             sheep_gain_from_food: Energy sheep gain from grass, if enabled.
+            moore:
```
rht, line 152: Missing documentation.

benchmarks/WolfSheep/wolf_sheep.py
```diff
@@ -49,55 +149,69 @@ def __init__(
...
                 self.random.randrange(self.height),
             )
-            energy = self.random.randrange(2 * self.wolf_gain_from_food)
-            wolf = Wolf(self.next_id(), pos, self, True, energy)
-            self.grid.place_agent(wolf, pos)
+            energy = self.random.randrange(2 * wolf_gain_from_food)
+            wolf = Wolf(
+                self.next_id(), self, energy, wolf_reproduce, wolf_gain_from_food
+            )
+            wolf.move_to(self.grid.cells[pos])
```
rht, line 190: Thank you for agreeing to https://github.com/projectmesa/mesa/issues/1953#issuecomment-1891098411 & to be as intuitive as NetLogo.

mesa/experimental/cell_space.py
```diff
@@ -0,0 +1,390 @@
...
+        "_connections",
+        "owner",
+        "agents",
+        "capacity",
+        "properties",
+    ]
+
+    def __init__(self, coordinate, owner, capacity: int | None = None) -> None:
+        self.coordinate = coordinate
+        self._connections: list[Cell] = []  # TODO: change to CellCollection?
```
rht, line 52: I find CellCollection to be different from this. To draw analogy from graphs. A CellCollection is a collection of nodes, and we don't care about their edges structure. But in this case, The `_connections` consists of nodes that are nearest neighbors to this Cell.

If you want to be pedantic, `self._nearest_neighbors`, because `connections` signifies edges instead of nodes.

mesa/experimental/cell_space.py
```diff
@@ -0,0 +1,390 @@
...
+        return f"Cell({self.coordinate}, {self.agents})"
+
+    @cache
+    def neighborhood(self, radius=1, include_center=False):
+        return CellCollection(
+            self._neighborhood(radius=radius, include_center=include_center)
+        )
+
+    @cache
+    def _neighborhood(self, radius=1, include_center=False):
```
rht, line 109: What about reusing some of the optimized algorithms in NetworkX? https://stackoverflow.com/questions/18393842/k-th-order-neighbors-in-graph-python-networkx

quaquel:
> I don't think this isolates it too much, this is something that has been on several wish lists.

I just meant that it keeps all experimental code within a single location. That is, in this way, we don't touch any legacy code and thus don't have to worry about backward compatibility and breaking existing features. We can simply go to 2.4 with this as new experimental code.

I look forward to seeing your work on continous spaces and I agree that having some space based movement methods within the base agent class is not a problem perse. 

> I still have hopes that we don't need this coupling. I think we can use raise your select_random_empty_cell function from NetworkGrid to the space base class, because it should work for all spaces. 

Will do.

> But re: coupling. I don't like that the cells manipulates the owners attributes (_empties) directly. This is already based on some assumptions how the owner is implemented. This is basically a performance optimization, but those shouldn't dictate how we implement our methods. 

Conceptually, I agree. Within the current structure I don't really see another good solution. Some possible directions I still see are (1) to track the number of tries and use this to estimate when we have to construct the full collection of empty cells; (2) base it on the number of agents in the model, but this introduces tight coupling between model and space; (3) pub/sup where Cells fire events when changing their occupancy. Spaces can then subscribe if needed.

mesa/experimental/cell_space.py
quaquel: I tested that. It is slower because it requires us to first generate the networkx graph based on our DiscreteSpace. 

Also, NX uses weighted dijkstra for ego_graphs. In our case there are no weights because we don't represent the edges.

rht: The pimplementation](https://networkx.org/documentation/stable/_modules/networkx/generators/ego.html#ego_graph) of `nx.ego_graph` reduces to [`nx.single_source_shortest_path_length`](https://networkx.org/documentation/stable/_modules/networkx/algorithms/shortest_paths/unweighted.html#single_source_shortest_path_length) when it is unweighted.

rht: > I tested that. It is slower because it requires us to first generate the networkx graph based on our DiscreteSpace.

What if it is initialized as a NetworkX graph to begin with? But then again, there are cythonized NetworkX alternatives. I know there is a [Rust-based NetworkX clone](https://github.com/Qiskit/rustworkx) that powers Qiskit, the most widely used quantum computing library in Python.

Corvince:
> Conceptually, I agree. Within the current structure I don't really see another good solution. Some possible directions I still see are (1) to track the number of tries and use this to estimate when we have to construct the full collection of empty cells; (2) base it on the number of agents in the model, but this introduces tight coupling between model and space; (3) pub/sup where Cells fire events when changing their occupancy. Spaces can then subscribe if needed.

If you agree conceptually I think we shouldn't already break this concept at such an early stage. I tried 1) and think thats a good intermediate solution. Just didn't had the time to find good values. 2) is similar to what we already have, but that actually doesn't always work, because in a MultiGrid you can have lots of agents and still lots of empty cells (this is also problematic in the legacy implementation). 3) sounds like the best solution so far, but requires some work and depends on a general approach towards event sourcing. 

Again, in my opinion we shouldn't break good concepts yet just for the sake of performance optimizations. Until the grid is starting to get full an unoptimized version of 1) already leads to good performance.

mesa/experimental/cell_space.py
rht: Rustworkx is about 20x faster than NetworkX for the weighted Dijkstra according to their [benchmark](https://www.rustworkx.org/benchmarks.html#single-source-shortest-path). I'm currently cooking up for the unweighted version of the benchmark.

quaquel:
> If you agree conceptually I think we shouldn't already break this concept at such an early stage

It happened largely because of me trying to chase down the performance difference between this branch and my code from over a week ago. It turned out that this performance difference was unrelated to this code and I did not change it back. 

> I tried 1) and think thats a good intermediate solution. Just didn't had the time to find good values

I guess we can use the current legacy code to create a number. Will take a look once I have time. The current cutoff implies some expected value, so it should not be too hard.

mesa/experimental/cell_space.py
quaquel: While developing the proof of principle out of which this PR emerged, I tried basing everything on networkx. It was not fast and the code became horrible. Feel free to experiment with other ideas and let us know what you find.

rht: I tested. Rustworkx doesn't implement an equivalent of `nx.single_source_shortest_path_length`, nor does the [`dijkstra_shortest_paths`](https://www.rustworkx.org/apiref/rustworkx.dijkstra_shortest_paths.html#rustworkx.dijkstra_shortest_paths) have a cutoff for radius.
Next, I tried `igraph`, but it is 2x-4x slower than `nx.single_source_shortest_path_length`.

> It was not fast and the code became horrible.

How much slower is it? If it is hypothetically multiplied by 20, would it be faster? This is to consider on whether to invest time in implementing `nx.single_source_shortest_path_length` in Rustworkx, or even cythonizing the said NetworkX function.

quaquel:
I have started adding tests for this code. Coverage is not complete yet. I have focused on the initialization of OrthogonalGrid, HexGrid, and NetworkGrid, as well as the neighborhood calculation for various radii and types of grids. This has resulted in a few minor changes in the code. I think this code is getting close to the point that it is ready for review. Still on my to-do list:

- [ ] make Cell and DiscreteSpace again independent and change `Grid.get_random_empty_cell` code as discussed above.
- [ ] add tests for CellCollection, Cell, and CellAgent
- [ ] make neighborhood tests more rigorous. At the moment I test explicitly the size and content of `_connections`, but I only test the size of the `neighborhood` for radius 1-3 on OrthogonalGrid (Moore, von Neumann) and HexGrids. I want to check explicitly the content of the return `neighborhood` as well. 
- [ ] Add some tests for directed graphs and NetworkGrid. 

Further down the line, we can worry about 
1. the integration of property grids. I don't think it is overly complicated, but I don't have the bandwidth for it right now
2. make `_connections` a dict with named connections to enable funky space traversal. Again, see the ideas above

Did I miss anything? If so, please let me know.

mesa/experimental/cell_space.py
quaquel: I don't recall the details of the speed. It was at least an order of magnitude slower. 

However, as important for not pursuing a graph library route any further is that the linked list style data structure now presented here is in my view much cleaner, shorter, more elegant, and easier to use than what I was able to come up with while playing around with something on top of networkx.

benchmarks/WolfSheep/wolf_sheep.py
quaquel: Once we added a CellAgent, adding a dedicated move method was a small step. Random walks are now just a one-liner as well.

Corvince:
That's great, thank you for your effort! Tomorrow I want to restructure the file, or better to say divide it into different files, because I think it already has become quite unwieldy. Is that okay for you? I don't want to interfere with your efforts

quaquel:
It's 400 lines, so 25% of the current space module. But indeed, it is getting to a point where it is a bit unwieldy. So I am okay with a reorganization. I am not sure, however, how to cut it up cleanly. Also, if possible can you then reorganize the tests as well so they follow whatever new organization you impose?

I won't work on this tomorrow daytime. I might start working on this again around 19:00 CET. Would that work for you?

Corvince:
Yes that should work, otherwise I'll postpone it. Thanks for the reminder to also adjust the test structure

Corvince:
> I am not sure, however, how to cut it up cleanly. 

Simple, I cut it up into each class. Its somewhat less used in the Python world for whatever reason, but I really like to have one file per class. I found it makes switching between code points much easier, because you don't loose your scroll position when jumping between classes. But its still all under the "cell_space" namespace, so the models and tests still use the same import structure. I don't think this is something we need to impose, but I would like to keep it this way for this experimental feature. 

The only exception I made to this is the grid.py file. I think from the content it is quite imminent that its three classes can potentially be merged into a single grid. The only real difference between them is the "directions" list, so I think its rather a question of good naming (we only need a parameter that differentiates between moore, von Neumann and hex).

I kept the tests together for now, I think currently it still makes sense to keep them together.

quaquel:
Thanks for this. I already had a quick look and it makes sense to me (although it gives me a bit of a JAVA feel). 

I have to think about merging the three grid classes. I see your point. Only the `_connect_single_cell` is really different, and thus, it might be handled through a keyword argument. My pushback would be that orthogonal and hex grids are conceptually quite different and distinct notions. Moreover, I am typically fine with a slightly richer set of methods/functions/classes if they represent distinct thinks over solving everything through keyword arguments.

rht:
> Simple, I cut it up into each class. Its somewhat less used in the Python world for whatever reason, but I really like to have one file per class. I found it makes switching between code points much easier, because you don't loose your scroll position when jumping between classes. But its still all under the "cell_space" namespace, so the models and tests still use the same import structure. I don't think this is something we need to impose, but I would like to keep it this way for this experimental feature.

Partially disagree with the way you split the classes here. Either:
- the abstract definitions should be in 1 file, but the specific instances (grid, continuous space, network) be in separate file
- the other way around

While I agree with the general direction of splitting up the classes here, but not all:
1. your reasoning is due to your specific IDE choice/configuration/taste. Several IDE's have gone beyond the boundary of files as organization of code, and have class-based navigation in the left sidebar.
2. as you said, this is a Python project, not Java or MATLAB, and should be less surprising to people who are used to Python project organization.

rht:
> the abstract definitions should be in 1 file, but the specific instances (grid, continuous space, network) be in separate file

I meant this:
- cell_space.py
- grid.py
- continuous_space.py
- network.py
Not
- cell_space.py
- implementations.py

quaquel:
I suggest we keep the current organization while this is experimental. The code base will likely expand to make this work with property grids and when we add names to the connections. Once the code base is stable, we can worry about the exact organization. 

Note: for the user of MESA it does not matter because, from the outside, it appears as a single namespace. For developers, it means clicking between files instead of through a file with some structured outline. Again, it's not something to overly obsess about.

rht:
The code organization concern can be quickly resolved, once for all, for the code convention throughout the codebase. I find it tiring having to keep track of unresolved concerns until much later.

> For developers, it means clicking between files instead of through a file with some structured outline.

For developers using [VS Code](https://stackoverflow.com/questions/67476757/object-browser-class-browser-for-visual-studio-code), [IntelliJ](https://stackoverflow.com/questions/32511392/in-intellij-is-there-any-package-class-browser-like-eclipses-java-browsing-pers), and [Vim/Neovim](https://www.reddit.com/r/neovim/comments/y9g4is/is_there_a_plugin_similar_to_visual_studios/), one can configure the code browser to be class-based browser instead of file-based.

I know this is bikeshedding, but once again, needs to be resolved once for all. And I'd prefer to follow the standard convention if there is such thing like `black`/`gofmt`/PEP/NEP of this.

quaquel:
> I know this is bikeshedding, but once again, needs to be resolved once for all. And I'd prefer to follow the standard convention if there is such thing like black/gofmt/PEP/NEP of this.

Happy to discuss this, but can we do so in a separate place so this stays focussed on the PR? Like you, it would be nice to have some agreements (also on imports) that we can just use as a general code style. 

I don't know of, nor could I find anything so far that gives clear guidance. I can only find very general advice:

1. Up to roughly 400 lines of code in a single module
2. group related classes and functions
3. group modules into packages and sub-packages.

Corvince:
> I know this is bikeshedding, but once again, needs to be resolved once for all. And I'd prefer to follow the standard convention if there is such thing like `black`/`gofmt`/PEP/NEP of this.

Yes it is bikeshedding and therefore I would kindly ask you to have this discussion elsewhere if you think this is an important matter. I'll reply once here but for future discussions please open a dedicated topic. 

> Partially disagree with the way you split the classes here. Either:
> 
> * the abstract definitions should be in 1 file, but the specific instances (grid, continuous space, network) be in separate file
> * the other way around

I think we are not too far away on this point. The current implementation has all the specific instances in seperate files. As I understand it you would only merge Cell, cellcollection and DiscreteSpace into 1 file. But these are distinct features and therefore shouldn't be in 1 file. Cell, CellCollection and DiscreteSpace follow a hierachical structure and are not tightly coupled. This just leaves cell_agent, but this needs to go near agents anyway once this is out of experimental.

> 
> While I agree with the general direction of splitting up the classes here, but not all:
> 
> 1. your reasoning is due to your specific IDE choice/configuration/taste. Several IDE's have gone beyond the boundary of files as organization of code, and have class-based navigation in the left sidebar.

Its the other way around. All IDEs offer file based structure, some (most) offer class-based navigation. So file-based is more general and less dependent on my specific IDE choice (which I never mentioned).

> 2. as you said, this is a Python project, not Java or MATLAB, and should be less surprising to people who are used to Python project organization.

I think code organisation and language choice are orthogonal to each other, one shouldn't depend on the other. There have been historical reasons for keeping the number of files small, but if you look at modern libraries they tend to spread out their features into many small files and/or group their files per feature.

quaquel:
I have added additional tests. Coverage is pretty complete, but I want to add a few more neighborhood tests just to be sure it's all working correctly. 

I also added two keyword arguments. First, random is now passed around correctly. So, all classes, in principle, use the same random number generator. As long as the DiscreteSpace is initialized with the models rng, everything else works fine. 

I also added `CellKlass` as an optional argument to DiscreteSpace and its subclasses. This is a form of dependency injection. It also makes it trivial for users to subclass Cell while still using MESA predefined DiscreteSpaces.  If, for example, you want Cells with Agent-like behavior, this is now easy to do. 

I still need time to look carefully at the `select_random_empty_cell` code. I have, however, removed already the tight coupling.

benchmarks/Schelling/schelling.py
```diff
@@ -16,23 +18,22 @@ def __init__(self, unique_id, model, agent_type):
         """
         super().__init__(unique_id, model)
         self.type = agent_type
+        self.radius = radius
```
EwoutH, line 21: Why was radius moved from the model to the agent?

mesa/experimental/cell_space/grid.py
```diff
@@ -0,0 +1,179 @@
...
+        if self._try_random:
+            while True:
+                cell = self.all_cells.select_random_cell()
+                if cell.is_empty:
+                    return cell
+        else:
+            return super().select_random_empty_cell()
+
+
+class OrthogonalGrid(Grid):
```
EwoutH, line 51: Maybe I'm nitpicking too much, but isn't a grid by definition `Orthogonal`? If we look at the cambridge definition:

> a [pattern](https://dictionary.cambridge.org/dictionary/english/pattern) or [structure](https://dictionary.cambridge.org/dictionary/english/structure) made from [horizontal](https://dictionary.cambridge.org/dictionary/english/horizontal) and [vertical](https://dictionary.cambridge.org/dictionary/english/vertical) [lines](https://dictionary.cambridge.org/dictionary/english/line) [crossing](https://dictionary.cambridge.org/dictionary/english/crossing) each other to [form](https://dictionary.cambridge.org/dictionary/english/form) [squares](https://dictionary.cambridge.org/dictionary/english/square).

I would expect the maybe the (super)classes `ContiniousSpace` and `DiscreteSpace`, where `Grid` and `Hexagonal` (or `GridSpace` and `HexogonalSpace`) can be subclasses of `DiscreteSpace`.

mesa/experimental/cell_space/grid.py
```diff
@@ -0,0 +1,179 @@
+from random import Random
+
+from mesa.experimental.cell_space import Cell, DiscreteSpace
+
+
+class Grid(DiscreteSpace):
```
EwoutH, line 6: Is it really this useful to make another layer between `DiscreteSpace` and the user accessible grids?

Also do we want stuff like DiscreteSpace and Grid private (so with an underscore)?

mesa/experimental/cell_space/cell_collection.py
```diff
@@ -0,0 +1,67 @@
...
+
+if TYPE_CHECKING:
+    from mesa.experimental.cell_space.cell import Cell
+    from mesa.experimental.cell_space.cell_agent import CellAgent
+
+
+class CellCollection:
+    def __init__(
+        self,
+        cells: dict[Cell, list[CellAgent]] | Iterable[Cell],
```
EwoutH, line 17: If a `CellCollection` can be both an dictionary and iterable of cells, doesn't that cause unexpected behavior? I understand that when iterating though a dict, you iterate though the keys, but now you can't be sure if there are agents in there or not. Mabye we need two seperate constructs.

mesa/experimental/cell_space/grid.py
```diff
@@ -0,0 +1,179 @@
...
+
+        for cell in self.all_cells:
+            self._connect_single_cell(cell)
+
+    def _connect_single_cell(self, cell):
+        i, j = cell.coordinate
+
+        # fmt: off
+        if i % 2 == 0:
+            directions = [
```
EwoutH, line 161: We had nice formatting for this, right? I found that quite intuitive.

EwoutH (commented):
Great effort so far! I left a few high level comments, will take another more in depth look tomorrow.

benchmarks/Schelling/schelling.py
quaquel: Because it is an attribute of the agent. Any time you start repeatedly doing in an agent self.model.{some_attribute}, you should just move it to the agent itself.

mesa/experimental/cell_space/grid.py
quaquel: I am adopting cellular automata terminology here according to which "A grid is called orthogonal if all grid lines intersect at a right angle". So in case of hex tiling, you don't have right angle and no orthogonal grid.

mesa/experimental/cell_space/grid.py
quaquel: Using an underscore in class names is atypical. The better and correct way to declare stuff public is through `__all__`. Which, indeed we could add.

mesa/experimental/cell_space/cell_collection.py
quaquel: Check the implementation if statement. 

You cannot overload initializers (note:: __init__ is **not** a constructor, __new__ is the constructor).

mesa/experimental/cell_space/grid.py
quaquel: not sure what happend.  I thought it was still there?

benchmarks/Schelling/schelling.py
```diff
@@ -49,22 +48,21 @@ def __init__(
...
         # We use a grid iterator that returns
         # the coordinates of a cell as well as
         # its contents. (coord_iter)
-        for _cont, pos in self.grid.coord_iter():
+        for cell in self.grid:
             if self.random.random() < self.density:
                 agent_type = 1 if self.random.random() < self.minority_pc else 0
-                agent = SchellingAgent(self.next_id(), self, agent_type)
-                self.grid.place_agent(agent, pos)
+                agent = SchellingAgent(self.next_id(), self, agent_type, radius)
```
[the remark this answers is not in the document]
quaquel: For now, I prefer`radius` as it is the same as in the current implementation of spaces.

mesa/experimental/cell_space/grid.py
EwoutH: We discussed this, and for now exposing everything is fine.

mesa/experimental/cell_space/discrete_space.py
```diff
@@ -0,0 +1,54 @@
...
+
+from functools import cached_property
+from random import Random
+
+from mesa.experimental.cell_space.cell import Cell
+from mesa.experimental.cell_space.cell_collection import CellCollection
+from mesa.space import Coordinate
+
+
+class DiscreteSpace:
```
EwoutH, line 11: Docstring needed

mesa/experimental/cell_space/network.py
```diff
@@ -0,0 +1,36 @@
+from random import Random
+from typing import Any
+
+from mesa.experimental.cell_space.cell import Cell
+from mesa.experimental.cell_space.discrete_space import DiscreteSpace
+
+
+class Network(DiscreteSpace):
```
EwoutH, line 8: Docstring needed

EwoutH (commented):
Okay, we're getting there. Main thing needed now is:
- [x] Adres the two previous open conversations
  - [x] naming
  - [x] formatting (of neighbours)
- [ ] The main PR message needs to be updated. Take inspiration from #1898 (doesn't need to be as extensive).
- [ ] At some minimal docstring.

mesa/experimental/cell_space/grid.py
EwoutH: First you can choose between `DiscreteSpace` and `ContiniousSpace`. `DiscreteSpace` can than be divided in two kinds: `Network` and what's now called `Grid`. Only grid implies squares, so I don't find that a logical name. What the collection of (`Orthogonal`)`Grid`, `Hexagonal` and `Triangular` is, is a division of space into discrete units in 2D space. We could call those Cells, Polygons, Tiles, or anything else. But the superclass of those shouldn't be called Grid, since that's the square thing (what it currently already is).

quaquel: So, grid does not necessarily imply squares. I first thought so as well, but I dove into some math definitions and turned out to be wrong. In various branches of mathematics, a grid is merely a set of intersecting lines.    
According to [Wolfram](https://mathworld.wolfram.com/Grid.html), "a grid usually refers to two or more infinite sets of evenly-spaced parallel lines at particular angles to each other in a plane, or the intersections of such lines".

Orthogonal Grids are grids where the lines intersect at a right angle. Square grids are a subclass of Orthogonal Grids (but this only makes sense if you are tesselating a continuous space). Isometric, or triangular grids have 3 lines intersecting at 60 degrees. Hex grids are formed by tessellating the space with regular hexagons.

EwoutH: So than `SquareGrid` would be best, right?

quaquel: No, orthogonal is correct. To establish whether the grid is square or not, you require a cartesian space upon which you impose an orthogonal grid. Depending on the number of grid cells along the x and y axis can you establish the shape of the grid cell, and only then can you establish whether it is a square grid or not.

quaquel:
I have added all the docstrings @EwoutH requested.

Corvince:
I know we are already late in the review process, but I just added two major updates. Both relate to Grids.

1) I added an optional `neighbord_func` argument to the init function. This function can take a cell as its argument and derive the neighborhood directions for it. This makes it a lot more flexible with respect to custom neighborhoods. Based on this change I split OrthogonalGrid into OrthogonalMooreGrid and OrthogonalVonNeumannGrid. This simplifies the code quite a bit, because they now just define their own default neighborhood_func. 

2) I modified Grid to allow n-dimensional grids! API-wise the only, but major change is that instead of `Grid(width, height, ...)` we now have `Grid(dimensions, ...)`. So for the examples I had to change `Grid(width, height, ...)` to `Grid([width, height], ...)`. The internal code changes are quite minimal, but of course this opens up a lot of possibilities. 


This should be up for discussion. I think the code can be improved a bit, maybe even automatically calculate n-dimensional moore/von-Neumann neighborhoods. But I didn't want to go this route further until we agree to keep those changes.

quaquel:
I already had a quick look. 

1. I am not sure I fully understand `neighborhood_func`. Why make it an optional keyword argument if it could be handled through subclassing? I am fine with the Moore vs. von Neuman split into separate classes, which makes good sense to me. It is more explicit this way.
2. I like this direction, and the use of `dimensions` makes sense. Does it already work for, say, 3-D grids?

Corvince:
1. I actually am not a big fan of oop and for me it's more natural to just define a custom function and pass that to Grid. But that was just the start of my journey to the implementation. But looking at how it is now I fully agree that it should just be handled through subclassing, which feels more natural and less out of place in this context. 

2. I honestly haven't tested it 😅 but it should, the general code is actually quite simple. I just assumed it would work for higher dimensions when it already works for 2 dimensions. The code just iterates over all possible combinations of its dimensions.

quaquel:
1. I broadly agree that composition is often preferable over inheritance, and deep inheritance trees or extensive use of multiple inheritance are a recipe for disaster. But in this case, inheritance is indeed the obvious way to go. 
2. Cell._connections is currently a list. So, one thing worth testing is that cells are not repeated in this list across dimensions. For example, in an orthogonal grid (Moore or Neuman), the cell above the current cell will occur twice if you make the connections naively across three dimensions.

Corvince:
Updated the code to only support inheritance, added parameter validation (mainly to keep hex-grid two-dimensional) and updated moore and von Neumann to work in n dimensions. Added some tests for 3 and 4 dimensions. 

Not sure if Moore and von Neumann is the right nomenclature for n-dimensional grids?

quaquel:
> Not sure if Moore and von Neumann is the right nomenclature for n-dimensional grids?

according to [wikipedia](https://en.wikipedia.org/wiki/Von_Neumann_neighborhood) it seems possible to go to higher dimensions. Von Neumann is also applicable to 1D, Moore requires at least 2D.

Corvince:
Funny, never thought about the implications for 1-D. I added a test for moore-1d. It still works, although as you shown its technically a von-Neumann grid. But at least it doesn't run into any problems when used for 1d. 

I also updated the the space to be truly generic for different Cell classes. 

Finally I messed up the tests and the commit history ;) Will fix this later, but the goal was actually to make the CI happy.

quaquel:
I had a quick look. The problem with the tests is caused by circular references. In both cell_agent.py and cell_collection.py, you import Cell from cell.py, which in turn imports both cell_agent and cell_collection. 

The reason for importing Cell is that you want to define `T = TypeVar("T", bound=Cell)`. Type hinting is quite new to me, so I am not entirely sure how to fix this. Changing this to "T = TypeVar("T", bound="Cell")" seems to fix the problem. I am however not sure if using a string in this context is allowed. It is in other contexts of type hinting. 

Once done, there is one more test that fails related to Network. I'll try to take a closer look after dinner.

Corvince:
Seems to be finally fixed and all checks passing!

The type system of Python is indeed... interesting. I work with TypeScript all day with no problems, but in Python... it sometimes feels really awkward. But you were right, "Cell" instead of Cell did the trick and just required the special TYPE_CHECKING imports

quaquel:
I guess this is basically ready to be merged then or not?

Type hinting is just weirdly bolted onto the language over the last few releases. It's getting better and better however. Coming from Java myself, I first liked the dynamic nature of Python, but I more and more long for static typing again.

EwoutH:
Awesome stuff so far!

I will do final review tomorrow morning. Can I squash when merging?

> /rerun-benchmarks

Unfortunatly, I never got that to work properly, so it was changed to be triggered when a label got added in https://github.com/projectmesa/mesa/pull/2002.

quaquel:
> I will do final review tomorrow morning. Can I squash when merging?

There are 2 things left: write a more elaborate commit message when merging and indeed squash and merge. I'll try to write up the message later today.

quaquel:
draft PR message, feel free to edit

# Summary
This PR introduces an alteranative implementation for discrete spaces. This implementation centers on the explicit inclusion of a Cell class. Agents can occupy cells. Cells have connections, specifying their neighbors. The resulting classes offer a cell centric API where agents interact with a cell, and query the cell for its neighbors. The resulting iplementation is about 20% than the current implementation, has a more expressive API, and is much shorter in terms of lines of code. 

To capture a collection of cells, and their content (_i.e._, Agents), this PR adds a new CellCollection class. This is an immutable collection of cells with convenient attribute accessors to the cells, or their agents. 

This PR also includes a CellAgent class which extends the default Agent class by adding a `move_to` method that works in conjunction with the new discrete spaces. 


# Motive
The PR emerged out of various experiments aimed at improving the performance of the current discrete space code. Moreover, it turned out that a cell centric API resolved various open issues (_e.g._, #1900, #1903, #1953). 


# Implementation
The key idea is to have Cells with connections, and using this to generate neighborhoods for a given radius. So all discrete space classes are in essence a [linked data structure](https://en.wikipedia.org/wiki/Linked_data_structure).

The cell centric API idea is used to implement 4 key discrete space classes: OrthogonalMooreGrid, OrthogonalVonNeumannGrid (alternative for SingleGrid and MultiGrid, and moore and von Neumann neighborhood) , HexGrid (alternative for SingleHexGrid and MultiHexGrid), and Network (alternative for NetworkGrid). Cells have a capacity, so there is no longer a need for seperating Single and Multi grids. Moore and von Neumann reflect different neighborhood connections and so are now implemented as seperate classes.

EwoutH:
Thanks! PR message looks good, I added some usage examples. Could you review if they are correct?

What do we think about the performance? WolfSheep got faster, Schelling slower. I think the 13x to 17x increase in init time for Schelling it quite a lot.

quaquel:
I have tried to confirm the results locally and I cannot. Locally, the performance difference for Schelling and WolfSheep is roughly the same.

Corvince:
For me the local results are pretty much the same as on GitHub

quaquel:
Ok, good to know. I'll try to investigate as soon as possible. At a minimum, I want to understand why Schelling is slower in runtime while it was initially about 15-20% faster,  in line with WolfSheep.

quaquel:
So, the single biggest loss in performance in the initialization is due to the supporting n-dimensional grids. If a revert back to the older `_connect_single_cell` code, this more than half the initialization time. This is still a factor 2 slower than my old code from a month ago, so there are some other places where we have lost performance.

quaquel:
So, changing e.g., `OrthogonalMooreGrid` to the following code below gives me an order of magnitude speedup and we are back to about a factor 4 slower init times compared to main. I have done 2 things: get the offsets only once (hardcoded here for 2D) and _connect_single_cell is likewise 2D only. 

So, the question, at least for init times seems to be: how do we want to handle N-dimensional grids? Writing generic code for N dimensions is an order of magnitude slower than writing dedicated code for 2D. Given that 2D is the most common use case, unless we find a way to speed up the generic code, I guess we have to separate 2D grids from n-dimensional grids. 


```python
class OrthogonalMooreGrid(Grid[T]):
    """Grid where cells are connected to their 8 neighbors.

    Example for two dimensions:
    directions = [
        (-1, -1), (-1, 0), (-1, 1),
        ( 0, -1),          ( 0, 1),
        ( 1, -1), ( 1, 0), ( 1, 1),
    ]
    """

    def __init__(
            self,
            dimensions: Sequence[int],
            torus: bool = False,
            capacity: float | None = None,
            random: Random | None = None,
            cell_klass: type[T] = Cell,

    ) -> None:
        super().__init__(dimensions, torus=torus, capacity=capacity, random=random, cell_klass=cell_klass)
        # offsets = self._calculate_neighborhood_offsets(None)
        # fmt: off
        offsets = [
            (-1, -1), (-1, 0), (-1, 1),
            (0, -1), (0, 1),
            (1, -1), (1, 0), (1, 1),
        ]
        # fmt: on

        for cell in self.all_cells:
            self._connect_single_cell(cell, offsets)

    def _connect_single_cell(self, cell: T, offsets: list[tuple[int, int]]) -> None:
        height, width = self.dimensions

        i, j = cell.coordinate
        for di, dj in offsets:
            ni, nj = (i + di, j + dj)
            if self.torus:
                ni, nj = ni % height, nj % width
            if 0 <= ni < height and 0 <= nj < width:
                cell.connect(self._cells[ni, nj])
```

EwoutH:
The current grid also only supports 2D right? For now, and this PR, I would say let's do optimized 2D only. We can always extend the use case further in future.

Corvince:
I was worried that performance might not be optimal, but didn't expect (and didn't test) that it would be that bad. 

Still I think it would be a cool feature to keep, but totally agree we should just have an optimized 2D implementation as well

quaquel:
I am trying to get to a clean 2d / ND implementation. It will be different across Moore, von Neuman and Hex (which is only defined for 2d anyway. I hope to finish the code in the next hour or so.

quaquel:
The new benchmark figures are quite a bit better. I still don't fully understand why for Schelling the code is slower while in previous tests it was faster than the current implementation. Although one possible explanation is that with the update to AgentSet.shuffle the bottlenecks in the code have shifted quite a bit and in particular percentages have changed a lot as well.

quaquel:
So I did a bit more digging. Changing Cell to not use a list if `Cell.capacity` is 1, and updating CellCollection.agent accordingly is an order of magnitude performance difference.

quaquel:
I did some more testing. I can't reproduce my initial performance differences so I guess I made a mistake when I thought there was a massive difference. I also tested a dedicated SingleAgentCell next to the normal cell and updated CellCollection.agents to take advantage of this. Again, this did not produce a massive speedup. So, I suggest we leave the code as is for now and just merge it. I have updated the PR text to better reflect the actual performance figures.

EwoutH:
Agreed, we can always optimize more later if needed.

On thing I'm not sure about is already porting our main benchmarks over. Maybe we could add those models to [mesa-examples](https://github.com/projectmesa/mesa-examples), and leave the benchmarks here as is.

quaquel:
In both my datacollection and devs branch, I have added a simple examples folder within the folder with the new experimental code. I am fine with using either mesa-examples or doing what I have been doing in devs and datacollection. Let me know what you prefer and why.

EwoutH:
I would suggest this PR shouldn't change the models in `benchmarks`. I can do those last bits however, I will try to do that and merge it tomorrow.

Corvince:
I think an example folder inside the experiments sounds good and like the place to be.


But maybe it would still be nice to have one of the benchmarks use this feature, since I think we all agree that cell space is going to be the recommended space implementation eventually

quaquel:
I like @Corvince suggestions. It also allows us to track how the performance of the new grid stuff is affected going forward.

EwoutH:
We could have two sets of benchmarks, regular ones and experimental ones

quaquel:
I am also fine with adding additional benchmarks, but how would that work out if we add additional experimental stuff? For example, if I add devs, which benchmarks would I edit to showcase their impact?

EwoutH:
Thought about it a bit, let's keep it simple. Benchmarks are an internal development tool, not user examples. So let's keep them on the cutting edge.

I'm good with merging (please clean ip or squash). I would like a second aproval before doing so.

Thanks for the thremendous effort!

EwoutH (approved):

Corvince:
@quaquel can you handle the merging, since I opened the PR? I'm swamped with work until next work and won't be able to do so until then

EwoutH:
I can also do it

PR #1996, opened 2024-01-24
fix: Reverse dict key and value for agents_by_type

rht:
This change is so that the agents_by_type API is compatible with Mesa 2.1.5 again.


rht:
The other commits, when done in separate PR, will be ignored in the release note anyway. This PR can be rebase and merged, if the goal is not to mix lint changes into 1 commit.

quaquel:
Fine with me either way.

Corvince (approved):

EwoutH:
Thanks!

PR #2001, opened 2024-01-24
benchmarks.yml: Another attempt at fixing PR branch checkout

EwoutH:
Let's try passing the information though a `step.id.outputs` [context](https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context). Maybe it helps, maybe not.

> This commit addresses the issue where GitHub Actions incorrectly checked out the main branch of the base repository when triggered by `issue_comment` events, instead of the correct branch from a forked repository. The solution involves using a GitHub script to fetch the pull request details associated with the issue comment, extracting the branch and repository information of the head (fork). This data is then used to explicitly checkout the correct branch from the forked repository. This ensures that the workflow operates on the intended branch, regardless of whether it's triggered by `pull_request_target` or `issue_comment` events, thereby maintaining the correct workflow context and enhancing security.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -57,29 +57,35 @@ jobs:
           fetch-depth: 0
           persist-credentials: false
           clean: false
-      - name: Checkout PR branch from issue comment
+      - name: Get PR info for Issue Comment
         if: github.event_name == 'issue_comment'
-        uses: actions/github-script@v5
+        id: get-pr-info
+        uses: actions/github-script@v7
         with:
           script: |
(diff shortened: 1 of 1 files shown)


EwoutH:
Curious which code is running if I do /rerun-benchmarks

EwoutH:
Again, this PR is only testable after merging to main, unfortunately.

rht (approved):

EwoutH:
I think I can still test on this PR, even closed: /rerun-benchmarks

EwoutH:
I’m going to try to think about another strategy then a comment trigger. Maybe a label?

PR #2002, opened 2024-01-25
benchmarks.yml: Run on addition of label instead of comment

EwoutH:
Instead of triggering on a specific comment, which caused issues with running in a different scope ([`issue_comment`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment) instead of [`pull_request_target`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target)), the workflow can now instead be triggered by adding a label "trigger-benchmarks" to the PR.

The label can be repeatedly added and removed to rerun benchmarks.

Screenshot_363

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -2,14 +2,9 @@ name: Performance benchmarks
 
 on:
   pull_request_target:
-    types: [opened, ready_for_review]
+    types: [opened, ready_for_review, labeled]
     branches:
       - main
-    paths:
-      - '**.py'
-      - '.github/workflows/benchmarks.yml'
-  issue_comment:
-    types: [created]
     paths:
       - '**.py'
       - '.github/workflows/benchmarks.yml'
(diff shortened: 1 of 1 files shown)


Corvince:
If this works this is probably a good solution. While your changing the trigger, maybe also now deactivate the trigger on open pull requests? I think a lot of PRs a not related to performance and the performance benchmark is a bit distracting there. Or maybe run on every PR, but don't comment without the label?

rht:
What about reacting to the existing benchmark comment with a rocket emoji to retrigger the benchmark? Figuring out a more convenient way to do the retrigger than removing and readding the label.

EwoutH:
@rht Good idea, but unfortunatly this only works if it's specifically in the PR scope. Everything related to comments isn't. Just an limitation of events in GitHub Actions.

A label is related to the PR scope, so thus it works.

@Corvince I like the explicit confirmation that a PR doesn't change performance. But I also recognize it can add noise. Let me think about it a bit.

quaquel:
looks good to me

PR #2003, opened 2024-01-25
type annotations: Replace Union[a, b] with a | b, Optional[a] with a | None #1341

Aryan1Mahahjan:
Solving the issue of replacing Union[a, b] with a | b and Optional[a] with a | None in type annotations:

Replacing Union[a, b] with a | b:

Improves readability and clarity of type hints.
Makes it easier to understand what types are allowed for a variable or function parameter.
Provides better type checking by the static type checker.

It may CI issuse in  build / build (ubuntu, 3.9) (pull_request) 

Aryan1Mahajahn


.pre-commit-config.yaml
```diff
@@ -9,6 +9,7 @@ repos:
   hooks:
     # Run the linter.
     - id: ruff
+      args: [ --fix ]
```
EwoutH, line 12: I don't necessarily disagree this might be useful, but it should be in an separate PR.

mesa/time.py
```diff
@@ -23,23 +23,20 @@
 
 # Mypy; for the `|` operator purpose
 # Remove this __future__ import once the oldest supported Python is 3.10
```
EwoutH, line 25: These two lines can also be removed

EwoutH (commented):
Aside from two minor things, looks good!

Now we just have to wait until we drop Python 3.9 support. I think that will happen after the Mesa 2.3 release, which will probably release in February or March. If you want, you can help with that when the time comes.

mesa/time.py
Aryan1Mahahjan: The reveiw is showing a failed check at codecov/patch regarding lines i haven't touched or changed. Do you have any idea related to that?

.pre-commit-config.yaml
EwoutH: This was done in #2004!

rht:
> Now we just have to wait until we drop Python 3.9 support. I think that will happen after the Mesa 2.3 release, which will probably release in February or March. If you want, you can help with that when the time comes.

What's blocking us from doing it for Mesa 2.3? See also https://github.com/projectmesa/mesa/pull/1756#discussion_r1368411042.

Aryan1Mahahjan:
Hi, is there any way I can contribute in any similar issue.

EwoutH:
I think improving the example models in [Mesa-examples](https://github.com/projectmesa/mesa-examples) and adding new ones is now the most useful. Especially if you include new features like the AgentSet or PropertyLayer.

rht:
Adding more Mypy annotations would also be useful.

rht:
@Aryan1Mahahjan here are several type checking errors for mesa/agent.py
```
mesa/agent.py:45: error: Type argument "float" of "NDArray" must be a subtype of "generic"  [type-var]
mesa/agent.py:79: error: "Model" has no attribute "random"  [attr-defined]
mesa/agent.py:139: error: Argument 1 of "__contains__" is incompatible with supertype "AbstractSet"; supertype defines the argument type as "object"  [override]
mesa/agent.py:139: note: This violates the Liskov substitution principle
mesa/agent.py:139: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
mesa/agent.py:139: error: Argument 1 of "__contains__" is incompatible with supertype "Sequence"; supertype defines the argument type as "object"  [override]
mesa/agent.py:139: error: Argument 1 of "__contains__" is incompatible with supertype "Container"; supertype defines the argument type as "object"  [override]
mesa/agent.py:200: error: "WeakKeyDictionary[Agent, None]" has no attribute "data"  [attr-defined]
mesa/agent.py:279: error: Signature of "__getitem__" incompatible with supertype "Sequence"  [override]
mesa/agent.py:279: note:      Superclass:
mesa/agent.py:279: note:          @overload
mesa/agent.py:279: note:          def __getitem__(self, int, /) -> Any
mesa/agent.py:279: note:          @overload
mesa/agent.py:279: note:          def __getitem__(self, slice, /) -> Sequence[Any]
mesa/agent.py:279: note:      Subclass:
mesa/agent.py:279: note:          def __getitem__(self, int | slice, /) -> Agent
mesa/agent.py:289: error: Incompatible return value type (got "Agent | list[Agent]", expected "Agent")  [return-value]
mesa/agent.py:359: error: "Model" has no attribute "random"  [attr-defined]
```
are you interested in looking to fixing it? You should start with opening a PR with small changes first, so that it can get merged earlier.

Aryan1Mahahjan:
Yes I am interested I will look into it.

Aryan1Mahahjan:
Screenshot from 2024-01-15 14-27-18
It works on my machine no errors found

quaquel:
Virtually all these changes have been implemented by now, so I am closing this PR.

PR #2004, opened 2024-01-25
Quality of life: automatically fix ruff errors

Corvince:
Two small changes that should make it easier to contribute.

1. Run ruff pre-commit with the fix argument - I can't think why this shouldn't be desirable. Otherwise you just have to fix errors manually. Thanks @Aryan1Mahahjan for the inspiration
3. Run pre-commit.ci automatically. This makes it much easier for new contributors to not fail on formatting errors. The downside is that this might introduce an additional commit, but they are easy to spot and ignore while looking in the git history (there is no worry that features were changed in contrast to a commit that was just named "fix lint" manually)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,5 @@
 ci:
     autoupdate_schedule: 'monthly'
-    autofix_prs: false
 
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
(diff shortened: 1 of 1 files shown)


rht:
> Run ruff pre-commit with the fix argument - I can't think why this shouldn't be desirable. Otherwise you just have to fix errors manually. Thanks @Aryan1Mahahjan for the inspiration

SGTM

> Run pre-commit.ci automatically. This makes it much easier for new contributors to not fail on formatting errors.

SGTM, this is also helpful for existing contributors so as not to skip auto-fixable pre-commit errors.

rht (approved):

EwoutH:
Fully agree, I did propose this a few months ago. Glad to see it in!

Issue #2006, opened 2024-01-24
RandomActivation scheduler not working after upgrade 2.1 -> 2.2.3


profConradi:
I upgraded 2.1 to 2.2.3 and in this simple test model RandomActivation scheduler does not work anymore: agent activation is done sequentially.
Furthermore if I not call super().__init__() in the model class, the first agent is not added to the scheduler. Is this a bug or a mistake on my part?

```python
    class CognitiveAgent(mesa.Agent):
        """Agents make decisions using automatic (e.g., reflexive) versus controlled (e.g., deliberative) cognition,
        interact with each other, and influence the environment (i.e., game payoffs)."""
    
        def __init__(self, unique_id, model):
            # Pass the parameters to the parent class.
            super().__init__(unique_id, model)
    
            # Each agent is defined by a single parameter x that represents its probability of controlled processing
            self.xdx = np.random.uniform(0., 1.)
    
        def step(self):
            # The agent's step will go here.
            # For demonstration purposes we will print the agent's unique_id
            other_agent = self.random.choice(self.model.schedule.agents)
            print(f"I'm {self.unique_id} and I choose {other_agent.unique_id}")
            
    class CognitiveModel(mesa.Model):
        """A model with some number of agents."""
    
        def __init__(self, N):
            super().__init__()
            self.num_agents = N
            self.schedule = mesa.time.RandomActivation(self)
            # Create agents
            for i in range(self.num_agents):
                a = CognitiveAgent(i, self)
                # Add the agent to the scheduler
                print(a,i)
                self.schedule.add(a)
    
        def step(self):
            """Advance the model by one step."""
            # The model's step will go here for now this will call the step method of each agent and print the agent's unique_id
            self.schedule.step() 
```

EwoutH:
I've narrowed it down to the fact that the AgentSet's `shuffle(inplace=True)` doesn't shuffle the agents. Could be because the shuffle isn't working, or the in place updating isn't.

https://github.com/projectmesa/mesa/blob/99b35fcacadacae1d2860c86ac370f262a70da93/mesa/time.py#L144-L147

![Screenshot_366](https://github.com/projectmesa/mesa/assets/15776622/85c66aff-fe7d-4f63-8fe2-f026d46f0943)

EwoutH:
It's the inplace updating that isn't working. Without it, it works correctly:

[...]

PR #2007, opened 2024-01-26
Fix AgentSet inplace shuffle (and thus RandomActivation), add tests

EwoutH:
Adds a test that checks if the `RandomActivation` doesn't trigger agents in a sequential order. This test now fails, because as noted in #2006, it currently does do it in sequential order.

In theory this could give false positives (a test passing when it shouldn't, but that chance is around ~0.1^18).

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -54,6 +54,7 @@ class Agent:
         except AttributeError:
             # model super has not been called
             self.model.agents_ = defaultdict(dict)
+            self.model.agents_[type(self)][self] = None
             self.model.agentset_experimental_warning_given = False
 
             warnings.warn(
--- a/tests/test_agent.py
+++ b/tests/test_agent.py
@@ -255,7 +255,7 @@ def test_agentset_select_by_type():
 
 def test_agentset_shuffle():
     model = Model()
-    test_agents = [TestAgent(model.next_id(), model) for _ in range(4)]
+    test_agents = [TestAgent(model.next_id(), model) for _ in range(12)]
 
     agentset = AgentSet(test_agents, model=model)
     agentset = agentset.shuffle()
--- a/tests/test_agent.py
+++ b/tests/test_agent.py
@@ -251,3 +251,15 @@ def test_agentset_select_by_type():
     # Test with no type specified (should select all agents)
     all_agents = agentset.select()
     assert len(all_agents) == len(mixed_agents)
+
+def test_agentset_shuffle():
+    model = Model()
+    test_agents = [TestAgent(model.next_id(), model) for _ in range(4)]
+    agentset = AgentSet(test_agents, model=model)
+
+    agentset = agentset.shuffle()
+    assert not all([a1==a2 for a1, a2 in zip(test_agents, agentset)])
+
+    agentset = AgentSet(test_agents, model=model)
+    agentset.shuffle(inplace=True)
+    assert not all([a1 == a2 for a1, a2 in zip(test_agents, agentset)])
--- a/tests/test_agent.py
+++ b/tests/test_agent.py
@@ -258,8 +258,8 @@ def test_agentset_shuffle():
     agentset = AgentSet(test_agents, model=model)
 
     agentset = agentset.shuffle()
-    assert not all([a1==a2 for a1, a2 in zip(test_agents, agentset)])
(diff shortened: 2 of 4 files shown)


quaquel:
If you change `do_each` to

```python
    def do_each(self, method, shuffle=False):
        if shuffle:
            self._agents.shuffle(inplace=True)
        self._agents.do(method)
```

it works correctly. I am going to check RandomActivationByType as well.

Do I have write access to this?

quaquel:
I also think this bug fix will actually produce a speedup because we avoid a needless copy operation of the AgentSet.

EwoutH:
Thanks! For clarity, we should document the difference between `agents`, `_agents` and `agents_` somewhere

EwoutH:
@quaquel could you add unittests for `shuffle`? And maybe we should check coverage of the AgentSet again at some point (not this PR).

quaquel:
do you mean for AgentSet.shuffle?

quaquel:
Can you trigger the benchmarks? I am curious to see what this fix has done with those given that it removes an unnecessary copy operation.

EwoutH:
Looks good! Benchmarks depend on #2002.

Can’t approve this PR, but looks good to me.

tests/test_agent.py
```diff
@@ -251,3 +251,16 @@ def test_agentset_select_by_type():
     # Test with no type specified (should select all agents)
     all_agents = agentset.select()
     assert len(all_agents) == len(mixed_agents)
+
+
+def test_agentset_shuffle():
+    model = Model()
+    test_agents = [TestAgent(model.next_id(), model) for _ in range(4)]
```
EwoutH, line 258: Maybe add a few more agents to minimize the chance that the orders are accidentally the same.

EwoutH: Forgot this was my branch, can do this myself!

mesa/agent.py
```diff
@@ -54,6 +54,7 @@ def __init__(self, unique_id: int, model: Model) -> None:
         except AttributeError:
             # model super has not been called
             self.model.agents_ = defaultdict(dict)
+            self.model.agents_[type(self)][self] = None
```
rht, line 57: This fix is out of the scope of this PR, but it's fine to add it here.

rht (approved):

quaquel:
I was torn. It was raised in the same issue, and it was just 1 line.

yannickoswald:
Hi, was any of these bugs present at 1.1.1? Could you clarify? Or are you sure the randomactivation scheduler used to work in 1.1.1 correctly? I cannot produce certain modelling results that I published with 1.1.1 in any of the subsequent versions. So far I tested 2.3.2 and 3.1.3. Thanks a lot that would be super helpful

quaquel:
AgentSet was added near the end of MESA 2, so it most definately was not there for MESA 1.1.1.

yannickoswald:
Ok great. Thanks for the quick reply. I used the time module still here like this in 1.1.1 mesa.time.RandomActivation(self). I am just trying to trace back what changed in (random) agent activation exactly compared to later versions. Even in 3+ versions I get strange results that I do not get using 1.1.1

quaquel:
Are you sure the issue is just agent activation?

yannickoswald:
Not 100% of course, but given that with 1.1.1 the results look smooth, as expected given model specifications and are reproducible entirely using this version, but created abrupt strange behavior from time step 1 onwards in later versions I am also definitely not sure it is not agent activation in mesa.

yannickoswald:
Of course I consider the likelihood for mesa to have a bug there quite low, as compared to my own model or how i used scheduler and time. Could you pinpoint me to the agent scheduling /sequencing in the newest release? I cannot find which module it is part of. Just so i understand what is happening in my 1.1.1 model.

quaquel:
Agent scheduling was removed in MESA 3. All possible schedulers that existed can be implemented on top of `AgentSet`.

Random activation in the simplest case is just `model.agents.shuffle_do("step")`

EwoutH:
Could also be in changes in seed / rng?

If you want to get to the root of the issue, open a new issue with a [Minimal, Reproducible Example](https://stackoverflow.com/help/minimal-reproducible-example
)

PR #2008, opened 2024-01-26
Drop support for Python 3.9

rht:
https://github.com/projectmesa/mesa/pull/2003#issuecomment-1912654261

### DIFF
--- a/.github/workflows/build_lint.yml
+++ b/.github/workflows/build_lint.yml
@@ -36,8 +36,6 @@ jobs:
             python-version: "3.11"
           - os: ubuntu
             python-version: "3.10"
-          - os: ubuntu
-            python-version: "3.9"
           # Disabled for now. See https://github.com/projectmesa/mesa/issues/1253
           #- os: ubuntu
           #  python-version: 'pypy-3.8'
--- a/docs/tutorials/intro_tutorial.ipynb
+++ b/docs/tutorials/intro_tutorial.ipynb
@@ -60,7 +60,7 @@
    "source": [
     "### Tutorial Setup\n",
     "\n",
-    "Create and activate a [virtual environment](http://docs.python-guide.org/en/latest/dev/virtualenvs/). *Python version 3.9 or higher is required*.\n",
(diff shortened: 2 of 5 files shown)


EwoutH:
Let's wait until after 2.3 is released, all major scientific packages still support 3.9 and we don't have a pressing reason to drop it.

rht:
This is to enable the process of the cythonization of space.py and time.py.
Either:
- Wait until it makes sense to drop Python 3.9. My objection: I don't think Mesa needs to be necessarily more conservative than NumPy/SciPy. Users can always stay at Mesa 2.2.x if needed.
- Remove the need of `from __future__ import annotations`, but via downgrading the Mypy syntax of space.py, instead of dropping support for 3.9.

What are the planned new features in 2.3.0 that you think users of Python 3.9 really need to have?

EwoutH:
Right, this is an actual reason for us to drop 3.9. We pushed a lot of experimental stuff in 2.2, I would like to end 3.9 on a bit more stable platform. On the other hand we can always issue more patch releases.

Let's say we would target 2.3 mid-way February. Immediately after we drop 3.9 support. Would that work for you, or would that significantly slow down development?

rht:
> We pushed a lot of experimental stuff in 2.2, I would like to end 3.9 on a bit more stable platform. On the other hand we can always issue more patch releases.

Yeah, stability concern is not a real blocker given the patch releases.

> Let's say we would target 2.3 mid-way February. Immediately after we drop 3.9 support. Would that work for you, or would that significantly slow down development?

SGTM

quaquel:
I am running into 3.9 vs 3.10 now as well. I can fix the type annotations in #1994 to make it work with 3.9 or first move to 3.10, so I don't need to do the tedious work of fixing all type annotations (and then changing them back to 3.10 syntax).

Corvince:
I am fine with dropping support for 3.9 soonish, but I think we still shouldn't immediately make mesa incompatible with 3.9 just for the sake of it. If there a good 3.10+ features that we want to use, we should not hesitate. But I don't think slightly nicer mypy syntax qualifies.

PR #2009, opened 2024-01-26
Improve readability of badges

rht:
Can be viewed at https://github.com/rht/mesa/blob/readme/README.md. Based on the README.md of https://github.com/pypa/hatch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
 # Mesa: Agent-based modeling in Python
 
-[![PyPi](https://img.shields.io/pypi/v/mesa.svg)](https://pypi.python.org/pypi/Mesa)
-[![GitHub Actions build status](https://github.com/projectmesa/mesa/workflows/build/badge.svg)](https://github.com/projectmesa/mesa/actions)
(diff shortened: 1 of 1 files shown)


quaquel:
Looks good to me.

EwoutH (approved):
Looks good!

Noticed the Python version badge only displays 3.9 and 3.10, probably needs updating in the pyproject.toml.

EwoutH:
Can we disable those required CI runs? They don't work because those jobs don't run on documentation changes.

Corvince:
> Can we disable those required CI runs? They don't work because those jobs don't run on documentation changes.

Yes that would be good, this has happened several times already. But this requires an admin to make the change, I think that's currently @jackiekazil or @tpike3

EwoutH:
Another item for:
- https://github.com/projectmesa/mesa/discussions/1962

tpike3:
> Another item for:
> 
> * [Increase bus factor to at least 3 among all Mesa components #1962](https://github.com/projectmesa/mesa/discussions/1962)

Fair point, I will raise the issue

PR #2010, opened 2024-01-27
Speedup of Agentset.shuffle

quaquel:
This speeds up shuffle by shuffling the actual weakrefs, rather than first turning them back into hard refs, shuffling them, and turning them back into weakness (which is what the old code did). 

One probably controversial choice is that I have shifted to default inplace shuffle. Feel free to let me know what you think. I am fine either way but inplace is quite a bit faster (and the default behavior of random.shuffle).


quaquel:
How can I trigger the benchmark stuff?

mesa/agent.py
```diff
@@ -179,7 +179,7 @@ def agent_generator(filter_func=None, agent_type=None, n=0):
 
         return AgentSet(agents, self.model) if not inplace else self._update(agents)
 
-    def shuffle(self, inplace: bool = False) -> AgentSet:
+    def shuffle(self, inplace: bool = True) -> AgentSet:
```
EwoutH, line 182: This is an API change - not necessarily against it, but we have to go thought a proper announce and release process. Do you think that's worth it? Also, is it consistent with other `inplace` default values in the AgentSet methods?

quaquel: Hence my comment. I haven't checked the rest.

The issue is that for e.g., select defaulting to new makes sense. While for shuffle inplace might make more sense. 

One solution is to go back to the old behavior and just document the performance benefit of inplace. Let's not overthink this.

quaquel:
The benchmark init numbers make little sense to me. My changes only relate to `shuffle` and `do`, neither of which is touched by the model initialization.

EwoutH:
Can you replicate them locally? Maybe with some more replications/seeds (modify `benchmarks/configurations.py`)?

mesa/agent.py
EwoutH: I would go for that approach. `select()` and `sort()` currently also default to `inplace=False`, and if I recall correctly (most of) Pandas does that to. The consistency there is worth something.

But yes, let's document in the docstring that inplace shuffling is faster and demonstrate that also in our public examples (if we have any already).

quaquel: I'll do that later today.

quaquel:
> Can you replicate them locally? Maybe with some more replications/seeds (modify `benchmarks/configurations.py`)?

I'll try. 

One possible solution, more generally, is to ensure the same seeds are used in both benchmarks. I'll look at the code and open a separate PR if needed.

EwoutH:
> One possible solution, more generally, is to ensure the same seeds are used in both benchmarks.

Should already be happening:
https://github.com/projectmesa/mesa/blob/6a4ea54d6dade48591db512d3f7ada18ec4d42ff/benchmarks/global_benchmark.py#L38
So if `"seeds"` in the config say `25`, it will use seed 1 to 25. It then does use the paired samples to compare it.

But if you see room for improvement, please do! Also check: https://github.com/projectmesa/mesa/pull/1983#issuecomment-1902637496

quaquel:
> Can you replicate them locally? Maybe with some more replications/seeds (modify `benchmarks/configurations.py`)?

I ran it locally with the default settings. The results are more in line with what I would expect. It seems the init is particularly noisy. Not really surprising but good to know.

| Model        | Size   | Init time [95% CI]        | Run time [95% CI]          |
|--------------|--------|---------------------------|----------------------------|
| Schelling    | small  | 🔵 -0.2% [-0.7%, +0.2%]   | 🟢 -9.5% [-9.7%, -9.4%]    |
| Schelling    | large  | 🔵 -1.6% [-3.5%, +0.4%]   | 🟢 -25.0% [-29.1%, -20.9%] |
| WolfSheep    | small  | 🔵 -3.8% [-4.6%, -2.9%]   | 🟢 -20.5% [-21.2%, -19.7%] |
| WolfSheep    | large  | 🔵 -11.0% [-24.2%, -1.5%] | 🟢 -24.4% [-28.0%, -20.7%] |
| BoidFlockers | small  | 🔵 -3.2% [-5.6%, -0.6%]   | 🔵 -2.9% [-4.1%, -1.9%]    |
| BoidFlockers | large  | 🟢 -7.1% [-8.1%, -6.2%]   | 🟢 -5.0% [-6.8%, -3.2%]    |

mesa/agent.py
rht: Do I change https://github.com/projectmesa/mesa-examples/blob/479eaf389e1a98bda0bdba5aa8f24ede0d84e764/examples/schelling_experimental/model.py#L70 to `self.agents.shuffle(inplace=True).do("step")`?

quaquel:
This is ready for review and merging. When merging, please squash and merge.

mesa/agent.py
```diff
@@ -188,15 +188,30 @@ def shuffle(self, inplace: bool = False) -> AgentSet:
...
-        self.random.shuffle(shuffled_agents)
+        weakrefs = list(self._agents.keyrefs())
+        self.random.shuffle(weakrefs)
+
+        if inplace:
+            self._agents.data = {entry: None for entry in weakrefs}
+            return self
+        else:
+            return AgentSet(
+                (agent for ref in weakrefs if (agent := ref()) is not None), self.model
```
rht, line 204: Cython doesn't support Walrus operator yet. Though not sure if AgentSet can/should be cythonized.

quaquel: Let's worry about that when we cross that bridge.

mesa/agent.py
quaquel: Yes that should save some time. I haven't fully benchmarked it but inplace avoids unpacking and repacking of weakrefs.

mesa/agent.py
```diff
@@ -188,15 +188,30 @@ def shuffle(self, inplace: bool = False) -> AgentSet:
...
+            )
 
-        return (
-            AgentSet(shuffled_agents, self.model)
-            if not inplace
-            else self._update(shuffled_agents)
-        )
+        # shuffled_agents = list(self)
```
Corvince, line 207: There are some comments left here

Corvince (commented):
I am wondering if a similar logic could be applied to there other methods or in place of the current _update method. I would think similar performance improvements might be possible?

quaquel:
> I am wondering if a similar logic could be applied to there other methods or in place of the current _update method. I would think similar performance improvements might be possible?

It's a bit tricky. Both `select` and `sort` typically need the actual agent object and thus cannot operate only on the weakrefs as is possible with `shuffle`.

mesa/agent.py
EwoutH: Agreed, this now makes if faster. If no walrus + Cython make it even faster, we will do that when the time comes.

EwoutH:
This PR is starting to look ready to me. How good is the test coverage of this part of the code?

EwoutH:
Do we have any idea why this could slow down the shelling large init? It happened on both CI runs, with remarkable consistency.

quaquel:
> This PR is starting to look ready to me. How good is the test coverage of this part of the code?

We added code as part of #2007. I will expand that to test both `inplace=True` and `inplace=False`.

> Do we have any idea why this could slow down the shelling large init? It happened on both CI runs, with remarkable consistency.

Yes, but not on my machine. Moreover, the init code is unrelated to this change. Shuffle is not called in `model.__init__`. Remember, the init is very short so tiny changes result in large percentages quickly.

Corvince (approved):

rht:
@quaquel protip: with https://github.com/projectmesa/mesa/blob/6a4ea54d6dade48591db512d3f7ada18ec4d42ff/CONTRIBUTING.rst?plain=1#L30, when you do `git pull`, it will automatically be `git pull --rebase`.

rht:
Squashed and merged with the commit message `perf: Speed up Agentset.shuffle (#2010)`.

mesa/agent.py
```diff
@@ -188,15 +188,21 @@ def shuffle(self, inplace: bool = False) -> AgentSet:
...
-        )
+        Note:
+            Using inplace = True is more performant
+
+        """
+        weakrefs = list(self._agents.keyrefs())
+        self.random.shuffle(weakrefs)
+
+        if inplace:
+            self._agents.data = {entry: None for entry in weakrefs}
```
rht, line 200: How come in this line there is no need for `(agent := ref()) is not None)`, but it is needed in L204?

quaquel: When you create a new AgentSet, it has its own internal WeakkeyDict. Each WeakkeyDict creates its own weakrefs.

As can be seen in [the docs](https://docs.python.org/3/library/weakref.html#weakref.ref), `weakref.ref` takes an optional callback. In the case of WeakkeyDict, this callback is a method on the dict itself (i.e., `self.remove`). So, when not shuffling in place, I have to unpack the weakrefs to ensure the new WeakkeyDict behaves correctly. 

Moreover, AgentSet takes an iteratable of agents as an argument. Not an iterable of weakref. ref instances.

Hope this clarifies.

PR #2011, opened 2024-01-27
More pythonic implementation of wolf sheep

quaquel:
This is a more Pythonic and single-file implementation of Wolf Sheep for benchmarking.

The single file is easier for profiling and line profiling.


benchmarks/WolfSheep/wolf_sheep.py
```diff
@@ -8,15 +8,13 @@
     Center for Connected Learning and Computer-Based Modeling,
     Northwestern University, Evanston, IL.
 """
-import math
 
-from mesa import Model, Agent
+from mesa import Agent, Model
```
rht, line 12: These 3 lines can be condensed to be just `import mesa`, then calling them would be `mesa.Agent`, `mesa.Model`, `mesa.space.MultiGrid`, in line with `np.array` or `pd.DataFrame`. See https://github.com/projectmesa/mesa-examples/blob/479eaf389e1a98bda0bdba5aa8f24ede0d84e764/examples/wolf_sheep/wolf_sheep/model.py#L18.

quaquel: but that is slower

EwoutH:
Thanks! Would it be possible to split into two commits: One that makes it a single file, and then the actual changes? This diff is very difficult to review.

quaquel:
There is no real point looking at the diff here. It's just a clean reimplementation.

benchmarks/WolfSheep/wolf_sheep.py
rht: It's about the same:
```
In [1]: import mesa

In [2]: m = mesa.Model()

In [3]: %timeit mesa.Agent(0, m)
627 ns ± 125 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [4]: from mesa import Agent, Model

In [5]: m2 = Model()

In [6]: %timeit Agent(0, m2)
627 ns ± 148 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

rht: Except for the import lines, this PR LGTM.

quaquel: I really don't have strong opinions on this so if others want to change this as well, fine. 

Note, however, that the Schelling model already uses the same style of importing as I have used here. I also personally prefer _e.g._, `SchellingAgent(Agent)` over `SchellingAgent(mesa.Agent)`. This is really just a matter of taste.

rht: The convention in the [examples repo](https://github.com/projectmesa/mesa-examples/blob/479eaf389e1a98bda0bdba5aa8f24ede0d84e764/examples/wolf_sheep/wolf_sheep/model.py#L18) all uses `mesa.x`. The ones in the Agents.jl benchmark code are from the outdated version of the examples.

> This is really just a matter of taste.

But in the case of Agents.jl benchmark, they count and compare the LOC.

quaquel: > The convention in the [examples repo](https://github.com/projectmesa/mesa-examples/blob/479eaf389e1a98bda0bdba5aa8f24ede0d84e764/examples/wolf_sheep/wolf_sheep/model.py#L18) all uses mesa.x.

I ment our [benchmarks](https://github.com/projectmesa/mesa/blob/2837cc15fd0017523ea58358950d3eb32d1ad421/benchmarks/Schelling/schelling.py)

> But in the case of Agents.jl benchmark, they count and compare the LOC.

And we agree that this is only a rough proxy we should not obsess about.

rht: > I ment our [benchmarks](https://github.com/projectmesa/mesa/blob/2837cc15fd0017523ea58358950d3eb32d1ad421/benchmarks/Schelling/schelling.py)

The 3 versions of the code (mesa-examples, core Mesa benchmark, Agents.jl) ideally need to be kept in sync.

quaquel: I agree. Once this is merged, I'll happily make PRs on the other libraries.

tpike3:
As the build is failing at SugarScape_g1mt, I am going to merge.

PR #2012, opened 2024-01-28
benchmark: Install main & PR branch of Mesa explicitly

rht:
This is needed to test the Cython PR's


Corvince (approved):

Corvince:
Very weird performance results. Wonder if there is something wrong with it, its consistently slower for all the models that use randomactivation scheduler, which was recently changed. Maybe some weird form of caching issue?

Issue #1269, opened 2022-04-09
Simplification: merge run.py into server.py


rht:
run.py always contains a few lines of code. Should we merge it into server.py to simplify the coding convention? If people need to call `server.launch` from a different files (e.g. running Mesa simulation as a part of a larger web application), they can define their separate run.py. I'm citing the 80-20 UX rule (see https://lawsofux.com/pareto-principle/) that we should optimize for the 80% of the use case, and hide the remaining 20% until when it is actually needed.

jackiekazil:
I think the run.py was originally created for 2 reasons... 
1. It was easy to identify which file to run. 
2. The run.py was to explicitly hold more run-type settings but it never did. 

I am not against your suggestion especially now that the runserver command exists.

rht:
I simplified the Boltzmann wealth model layout in https://github.com/projectmesa/mesa-examples/pull/44 into
examples/boltzmann_wealth_model/:
- model.py
- app.py
- __init__.py  (so that the model is importable)

To run the GUI, simply do `solara run app.py`.

jackiekazil:
How does this work with mesa runserver?

rht:
`mesa runserver` could later be a proxy to `solara run`, but probably later on in 3.0.

quaquel:
this is all deprecated with MESA 3.

PR #2015, opened 2024-01-28
feat: Let mesa runserver detect server.py as fallback

rht:
This is part of the ongoing effort to obsolete run.py, to simplify the current model template. See #1269.

### DIFF
--- a/mesa/main.py
+++ b/mesa/main.py
@@ -30,11 +30,14 @@ def runserver(project):
     PROJECT is the path to the directory containing `run.py`, or the current
     directory if not specified.
     """
-    run_path = Path(project) / "run.py"
-    if not run_path.exists():
-        sys.exit(f"ERROR: file {run_path} does not exist")
-    args = [sys.executable, str(run_path)]
-    call(args)
+    run_files = ["run.py", "server.py"]
+    for run_file in run_files:
+        run_path = Path(project) / run_file
+        if not run_path.exists():
+            continue
+        args = [sys.executable, str(run_path)]
+        call(args)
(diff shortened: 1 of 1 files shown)


rht:
Tested manually:
- In the presence of both run.py and server.py
- In the presence of server.py
- In the presence of neither run.py nor server.py
All the 3 cases works as expected.

rht:
@EwoutH do you know why did the performance comment happen twice?

EwoutH:
Once for opening the PR, once for marking it "Ready for review"

jackiekazil:
I will take a look this evening or tomorrow. This is exciting.

tpike3 (approved):

EwoutH:
Can someone write a 3 sentence summary of this PR for the 2.2.5 release notes?

(I would find it really useful if each PR message is readable in isolation. Just from the PR message I would like to understand what a PR does/solves/enhances/adds without having to dig into code or other issues)

rht:
This is a new feature, even though @tpike3 said that further older models used server.py instead of run.py, but the feature wasn't present in 2.2.4 nor 2.1.x and 2.0.x. And so can't be included in a patch release.

rht:
But I suppose it is fine if the feature has once existed in the past? @EwoutH updated PR description. Up to you if you want to include in 2.2.5 or not.

rht:
Wait, I just realized of a bug in this PR
> In the presence of both run.py and server.py

The code says that both run.py and server.py are executed in sequence instead. Not sure why I didn't see it. Let me double check.

rht:
I see why the code passed my manual tests: it is because the first run call blocks the execution of the next line indefinitely, and the user always presses `ctrl-c` to exit before the next line gets executed. All is good.

Issue #2017, opened 2024-01-29
Update tutorials to include super().__init__() when subclassing mesa.model


davidbailey:
**Describe the bug**
The tutorials (e.g. https://mesa.readthedocs.io/en/stable/tutorials/intro_tutorial.html) instruct users to subclass mesa.model. However they do not say to include super().__init__(). In https://github.com/projectmesa/mesa/blob/main/mesa/model.py#L62-L77 it states you should "Always start with super().__init__() to initialize the model object properly."

**Expected behavior**
The documentation should tell users how to properly use the mesa.model class. The examples should include super().__init__() when subclassing mesa.model

**To Reproduce**
Not applicable.

**Additional context**
None.

rht:
Closed via #2024.

PR #2025, opened 2024-02-04
Adding super().__init__() to MoneyModel tutorial

sw23:
Noticed this error in my project, so turned to the tutorial for reference, which had the same error!
```
AttributeError: 'MoneyModel' object has no attribute '_steps'
```

I poked around and found that model sub-classes now require a call to `super().__init__()`, which wasn't originally in the tutorial. I also saw someone opened an issue ( #2017 ) to fix the tutorial, so opening this if it's helpful.

### DIFF
--- a/docs/tutorials/MoneyModel.py
+++ b/docs/tutorials/MoneyModel.py
@@ -43,6 +43,7 @@ class MoneyModel(mesa.Model):
     """A model with some number of agents."""
 
     def __init__(self, N, width, height):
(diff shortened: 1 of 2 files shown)


Corvince:
Thank you for bringing this up! This is quite important, because using super().__init__() should be encouraged (and thus in the tutorial), but not strictly required before mesa 3.0. So it is actually a bug you found.

rht (approved):

PR #2026, opened 2024-02-04
fix: Initialize model _steps and _time during __new__

rht:
This is to retain backward compatibility with Mesa 2.1.x behavior where a model `super().__init__()` is not necessary.

### DIFF
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -57,6 +57,9 @@ class Model:
             # advance.
             obj._seed = random.random()  # noqa: S311
         obj.random = random.Random(obj._seed)
+        # TODO: Remove these 2 lines just before Mesa 3.0
+        obj._steps = 0
+        obj._time = 0
         return obj
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:


rht:
Tested manually by running https://github.com/projectmesa/mesa-examples/blob/34e92d2c29a99e950bc17f7cac860c5b13156b21/examples/bank_reserves/batch_run.py without model `super().__init__()`.

mesa/model.py
```diff
@@ -57,6 +57,9 @@ def __new__(cls, *args: Any, **kwargs: Any) -> Any:
             # advance.
             obj._seed = random.random()  # noqa: S311
         obj.random = random.Random(obj._seed)
+        # TODO: Remove these 2 lines after Mesa 2.3
```
Corvince, line 60: Should be before Mesa 3.0, we can't know if 3.0 comes after 2.3

Corvince (approved):
Good, quick fix.

mesa/model.py
rht: Fixed.

rht:
cc: @EwoutH for patch release.

EwoutH:
Will try to do tonight, and will make screen recording so others can do it as well.

Only this PR needs to be included right?

EwoutH:
Could someone write a 3-sentence summary with which problem this fixes for the 2.2.5 release notes?

rht:
The problem is that #1942 added `_steps` and `_time` during model `__init__`. `_steps` is used in the data collector and `batch_run`, so that both support a scheduler-less model implementation. #1942 broke backward compatibility because it requires `super().__init__()` inside user's model `__init__`. This PR is to have the backward compatibility back.

rht:
@EwoutH  amended PR description. You may pick whichever version you prefer.

rht:
@EwoutH I know about doing the cherry picking and adding a tag for a release, but I'm not sure about the auto-draft for the release note. Still pending a release checklist (which has more automation than in the past) from your end.

EwoutH:
Checklist is here: https://github.com/projectmesa/mesa/pull/1917#issuecomment-1871352058

Could you also convert https://github.com/projectmesa/mesa/blob/main/CONTRIBUTING.rst to markdown? Then I will add the checklist to it.

Go ahead of the patch release. Make sure to cherry pick from `main` to [`2.2.x-maintenance`](https://github.com/projectmesa/mesa/tree/2.2.x-maintenance). If you have any questions, hit me up! (Matrix is probably the fastest)

rht:
I looked at the checklist. Step 1 definitely can be scripted in a code as an automated check. Step 2 can be automated as well ([API call](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release) with draft boolean). Step 3 can be input as a CLI flag. And so on.

mesa/model.py
```diff
@@ -57,6 +57,9 @@ def __new__(cls, *args: Any, **kwargs: Any) -> Any:
             # advance.
             obj._seed = random.random()  # noqa: S311
         obj.random = random.Random(obj._seed)
+        # TODO: Remove these 2 lines just before Mesa 3.0
+        obj._steps = 0
+        obj._time = 0
```
EwoutH, line 62: Could you remind me why we need to remove these lines? Seems like the time.

rht: It's because they are already specified in `__init__`. This was a hack to people who forget to `super().__init__` their model.

PR #2035, opened 2024-02-16
docs: Convert api/*.rst -> api/*.md via rst2myst

rht:
This has been brought up in dev meetings for a while, to have 1 format for documentation everywhere, not a mixture of .rst and .md. This is actually a long overdue change; @tpike3 could confirm this. The secondary reason is I want to try ChatGPT in making the docs clearer and more concise. See https://github.com/projectmesa/mesa/pull/2037.


Corvince:
Thank you for mentioning uv, the news haven't reached me yet. I think this is quite exciting and I will start using uv today, but I strongly discourage its use for mesa just yet. The first public release is less than 24 hours ago. This potentially decreases our workflows by a few seconds, but I don't think that justifies the possible hurdles that are attached to bleeding edge software.

Corvince:
You opened several similar PRs. None of the gives any reasoning to the change. My comment relates to all of those. What issue does this address? What benefits does this provide? Its quite tedious work to review this, so what reasoning am I missing?

rht:
This has been brought up in dev meetings for a while, to have 1 format for documentation everywhere, not a mixture of .rst and .md. This is actually a long overdue change; @tpike3 could confirm this. The secondary reason is I want to try ChatGPT in making the docs clearer and more concise. See #2037.

I made several orthogonal PR's so that they can be reviewed piecemeal, and I have self-reviewed each myself. Where manual fixes are applied in the subsequent commits.

Corvince:
Ok good, thanks for the reply. Would be good to include something like that in at least one of the PRs as a description in the future.

EwoutH:
But @rht, please **always** add a PR description. Not only for us as maintainers, but also for others looking around.

Structure for example it like this:
- Motivation or problem it solves
- Why it solves the problem this way
- (optionally) how it solves the problem (if non-trivial)

It's really important for reviews **and** users and contributors the PRs are accessible and as easy as possible to understand.

rht:
> Required statuses must pass before merging

EwoutH:
> Required statuses must pass before merging

Another #1962 bus factor issue!

@tpike3 or @jackiekazil could one of you remove all the required status checks? Approval by maintainers is already needed, we can interpret ourselves if the CI does what we want or not. We don't need a hard check for that.

jackiekazil:
@EwoutH  - 
Do you have this option? or is this only admins? 

>  Merge without waiting for requirements to be met (bypass branch protections)

Corvince:
> @EwoutH  - 
> Do you have this option? or is this only admins? 
> 
> >  Merge without waiting for requirements to be met (bypass branch protections)

This is only visible for admins.


Just for reference I added the required checks in response to https://github.com/projectmesa/mesa/pull/1788#issuecomment-1710291120 
Sorry that this was done a bit haphazardly

But I agree we should remove them again.

EwoutH (approved):
Tested and look good to me! Thank you both!

PR #2038, opened 2024-02-16
ci: Use uv pip for faster build

rht:
[`uv`](https://astral.sh/blog/uv) is a fast drop-in `pip`, `pip-compile`, `virtualenv` etc replacement created by the creator of Ruff.

### DIFF
--- a/.github/workflows/build_lint.yml
+++ b/.github/workflows/build_lint.yml
@@ -46,30 +46,12 @@ jobs:
       uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python-version }}
-    - if: ${{ runner.os == 'Windows' }}
-      # This is needed so that restoring cache on Windows is fast.
-      # See until https://github.com/actions/cache/issues/752 is resolved.
-      name: Use GNU tar
-      shell: cmd
-      run: |
-        echo "Adding GNU tar to PATH"
-        echo C:\Program Files\Git\usr\bin>>"%GITHUB_PATH%"
-    - uses: actions/cache@v4
-      with:
-        path: ${{ env.pythonLocation }}
-        key: ${{ runner.os }}-pip-${{ env.pythonLocation }}-${{ hashFiles('setup.py') }}
-    - name: Install dependencies
-      # Only if the cache misses
-      # Based on https://github.com/pypa/pip/issues/8049#issuecomment-633845028
-      # read_requirements.py should be removed once
-      # https://github.com/pypa/pip/issues/11440 is resolved.
-      if: steps.cache.outputs.cache-hit != 'true'
-      run: |
-        pip install toml
-        python tests/read_requirements.py > requirements.txt
-        pip install -r requirements.txt
+        cache: 'pip'
+    - name: Install uv
+      run: pip install uv
     - name: Install Mesa
-      run: pip install --no-deps .
+      # See https://github.com/astral-sh/uv/issues/1945
+      run: uv pip install --system .[dev]
     - name: Test with pytest
       run: pytest --durations=10 --cov=mesa tests/ --cov-report=xml
     - if: matrix.os == 'ubuntu'
@@ -84,22 +66,11 @@ jobs:
       uses: actions/setup-python@v5
       with:
         python-version: "3.12"
-    - uses: actions/cache@v4
-      with:
-        path: ${{ env.pythonLocation }}
-        key: test-examples-pip-${{ env.pythonLocation }}-${{ hashFiles('setup.py') }}
-    - name: Install dependencies
-      # Only if the cache misses
-      # Based on https://github.com/pypa/pip/issues/8049#issuecomment-633845028
-      # read_requirements.py should be removed once
-      # https://github.com/pypa/pip/issues/11440 is resolved.
-      if: steps.cache.outputs.cache-hit != 'true'
-      run: |
-        pip install toml
-        python tests/read_requirements.py > requirements.txt
-        pip install -r requirements.txt
+        cache: 'pip'
+    - name: Install uv
+      run: pip install uv
     - name: Install Mesa
-      run: pip install --no-deps .
+      run: uv pip install --system .[dev]
     - name: Checkout mesa-examples
       uses: actions/checkout@v4
       with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,6 +31,7 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: "3.12"
+          cache: 'pip'
       - name: Install dependencies
         run: pip install -U pip hatch
       - name: Build distributions


Corvince:
Thank you for mentioning uv, the news haven't reached me yet. I think this is quite exciting and I will start using uv today, but I strongly discourage its use for mesa just yet. The first public release is less than 24 hours ago. This potentially decreases our workflows by a few seconds, but I don't think that justifies the possible hurdles that are attached to bleeding edge software.

rht:
I have yet to measure by how much this shaves off the build time, because the CI still fails.

According to the blog post:
> [uv](https://github.com/astral-sh/uv) is designed as a drop-in replacement for pip and pip-tools, and is ready for production use today in projects built around those workflows.

If it builds and runs the code just fine, I see no reason not to use it. Besides, Ruff is considered bleeding edge, yet many huge projects use it in their CI. Not all "bleeding edge" public-released tools are the same in terms of reliability.

rht:
I have overcome https://github.com/astral-sh/uv/issues/1374. Waiting for https://github.com/astral-sh/uv/issues/1355 to be fixed, which should happen really soon.

Corvince:
>  Not all "bleeding edge" public-released tools are the same in terms of reliability.

Yes but there have already been 2 bugfix releases in the first 12 hours of its existence and you already linked two issues that you had to overcome/waiting for a fix. And these are just the obvious bugs, I am more worried about the subtle bugs, where it appears to be working, but some details are broken. This is inevitable for such large projects. We don't know if it will affect us, but why risk it?

rht:
Didn't you want to know how much this will speed up the macOS and Windows build? My sense is that it will be a lot, and to have a CI duration that is < 1min.

quaquel:
I agree with @Corvince. Let's give UV a few weeks to sort out the initial problems before we consider shifting.

Also, what is the problem with a CI that takes a bit over a minute?

rht:
> Let's give UV a few weeks to sort out the initial problems before we consider shifting.

That sounds like a good timeline for Ruff developers. Ruff used to have a daily (if not more frequent) release cycle. And during this high frequency releases, it already had adopters by huge projects. Faster CI does matter, to the point they have been willing to adopt an alpha-release linter.

> Also, what is the problem with a CI that takes a bit over a minute?

https://github.com/projectmesa/mesa/pull/2039#issuecomment-1948115515

EwoutH:
> but I strongly discourage its use for mesa just yet. The first public release is less than 24 hours ago.

Agreed. Let's wait a few week until uv stabilizes a bit more.

.github/workflows/build_lint.yml
```diff
@@ -60,18 +60,23 @@ jobs:
...
+      run: pip install uv
     - name: Install dependencies
       # Only if the cache misses
       # Based on https://github.com/pypa/pip/issues/8049#issuecomment-633845028
       # read_requirements.py should be removed once
       # https://github.com/pypa/pip/issues/11440 is resolved.
       if: steps.cache.outputs.cache-hit != 'true'
       run: |
-        pip install toml
+        rm -fr ${Python_ROOT_DIR}/lib/python${{ matrix.python-version}}/site-packages/pip-23.0.1.dist-info
```
Corvince, line 74: Why is this line needed? Seems like it would break easily when the pip version changes.

.github/workflows/build_lint.yml
```diff
@@ -60,18 +60,24 @@ jobs:
...
+        rm -fr ${Python_ROOT_DIR}/lib/python${{ matrix.python-version}}/site-packages/pip-23.3.1.dist-info
+        uv pip install toml
         python tests/read_requirements.py > requirements.txt
-        pip install -r requirements.txt
+        uv pip install -r requirements.txt
     - name: Install Mesa
-      run: pip install --no-deps .
+      run: uv pip install --no-deps "mesa @ ."
```
Corvince, line 80: Is the "mesa @" part required? Locally it works just fine without that

Corvince:
I think its funny that pip takes 2 seconds to install just uv and then uv continues to install the dependencies in a few milliseconds. Curious how it will perform with outdated caches.

.github/workflows/build_lint.yml
rht: `uv` breaks if there are multiple `pip` installed. There isn't a bug report for this yet, though if I were to open one, would be resolved within days because of the use case in GH Actions.

.github/workflows/build_lint.yml
rht: Used to require such. It could have been fixed by 0.1.5. Let me try it.

rht:
Everything builds except for Windows, which is a matter of using `rm -r -fo` instead of `rm -fr`.
The macOS build is sped up 2x (which should be sped up further after catching). The examples build is sped up from 27s in #2044 to 17s.

I should be able to revert my advanced (but complex) caching back to `setup-python` pip cache, and it would be simpler and even faster overall.

.github/workflows/build_lint.yml
Corvince: Are you sure about this one? Specifically I don't see how why multiple pips should be installed at all and why uv should interact with them at all. Also I haven't seen this in other workflow files that already use `uv`

rht: That's because they use `uv venv`, e.g. https://github.com/astral-sh/uv/issues/1386#issuecomment-1952420541.

rht: I was thinking of waiting until the non-venv solution has stabilized, but `uv venv` is blazing fast, so it shouldn't matter to create the venv.

.github/workflows/build_lint.yml
rht: Still requires "mesa @ .".

.github/workflows/build_lint.yml
rht: `uv venv` failed https://github.com/projectmesa/mesa/actions/runs/7980718647/job/21790983916?pr=2038 because Solara has a hardcoded assumption on the virtualenv information.

rht:
Tests all passed.

rht:
It's now even more concise with `uv pip install --system ...`.

EwoutH:
Nice work, I like the reduced complexity of our CI configuration. I think we can try it.

rht:
There has been ~100 PRs opened that contain `uv pip install`, including from the astral-sh repo itself. How about we wait until at least ~300? https://github.com/search?q=%22uv+pip+install%22&type=pullrequests.

.github/workflows/build_lint.yml
```diff
@@ -46,30 +46,11 @@ jobs:
...
-      run: |
-        pip install toml
-        python tests/read_requirements.py > requirements.txt
-        pip install -r requirements.txt
+        cache: 'pip'
+    - name: Install uv
+      run: pip install uv
     - name: Install Mesa
-      run: pip install --no-deps .
+      run: uv pip install --system -e .[dev]
```
EwoutH, line 53: Is the edible install needed here?

rht: There are 2 options as described in https://github.com/astral-sh/uv/issues/1945.

charliermarsh: I saw this linked from uv -- just as a heads up, this is supported as of `v0.1.24`. You can `uv pip install .` directly, without using an editable.

rht: @charliermarsh thank you for the awesome tool! I have amended the PR with `uv pip install --system .[dev]`.

EwoutH (approved):
```
Run uv pip install --system .[dev]
Resolved 110 packages in 903ms
Downloaded 108 packages in 1.76s
Installed 110 packages in 267ms
```
This is quite awesome. Thanks!

PR #2039, opened 2024-02-16
tests: Speed up test_batch_run

rht:
This is an attempt to speed up overall CI testing time, as much as possible without reducing comprehensiveness. Currently, it takes 12s on Python 3.12 on Ubuntu.

### DIFF
--- a/.github/workflows/build_lint.yml
+++ b/.github/workflows/build_lint.yml
@@ -73,7 +73,7 @@ jobs:
     - name: Install Mesa
       run: pip install --no-deps .
     - name: Test with pytest
-      run: pytest --cov=mesa tests/ --cov-report=xml
+      run: pytest --durations=10 --cov=mesa tests/ --cov-report=xml
     - if: matrix.os == 'ubuntu'
       name: Codecov
       uses: codecov/codecov-action@v4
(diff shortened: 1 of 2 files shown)


quaquel:
Can you explain why you are concerned about CI testing time?

rht:
1. It speeds up the review feedback loop between the reviewers and PR (that has been marked as ready) authors. The sub 1 min vs 2 min+ speedup benefit compounds for a single PR that requires several testing such as #1994. The Ubuntu CI runs used to be ~1 min, now about 2x faster at 30s.
2. More immediate feedback to contributors.

quaquel:
Fair enough, although this does not make a massive difference for me. I always think of this in a more categorical manner: can I wait behind my desk, can I go for a cup of coffee, can it run over lunch, or does it take even longer? The difference between 30 seconds and 2 minutes is still very much in the behind-my-desk category. 

Also, I see CI stuff as stuff that runs after having tested everything locally anyway, so I don't really see the feedback loop argument.

To be clear: speedups are still nice.

EwoutH (approved):
I don't see significant degradation in test functionality, so looks good. Faster CI is always nice, thanks!

PR #2044, opened 2024-02-20
new feature: AgentSet.get can retrieve one or more then one attribute

quaquel:
This adds a small additional feature to AgentSet. The get method now takes a string or list of strings. This allows you to retrieve multiple attributes in one go. I also added a test to cover this new feature.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -274,6 +274,10 @@ class AgentSet(MutableSet, Sequence):
         Returns:
             list[Any]: A list with the attribute value for each agent in the set if attr_names is a str
             List[List[Any]]: A list with a list of attribute values for each agent in the set if attr_names is a list of str
+
+        Raises:
+            AttributeError if an agent does not have the specified attribute(s)
+
         """
 
         if isinstance(attr_names, str):
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -269,7 +269,7 @@ class AgentSet(MutableSet, Sequence):
         Retrieve the specified attribute(s) from each agent in the AgentSet.
 
         Args:
-            attr_names (str | List[str]): The name(s) of the attribute(s) to retrieve from each agent.
+            attr_names (str | list[str]): The name(s) of the attribute(s) to retrieve from each agent.
 
         Returns:
             list[Any]: A list with the attribute value for each agent in the set if attr_names is a str
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -272,7 +272,8 @@ class AgentSet(MutableSet, Sequence):
             attr_names (str | List[str]): The name(s) of the attribute(s) to retrieve from each agent.
 
         Returns:
-            list[Any]: A list of attribute values for each agent in the set.
+            list[Any]: A list with the attribute value for each agent in the set if attr_names is a str
+            List[List[Any]]: A list with a list of attribute values for each agent in the set if attr_names is a list of str
         """
 
         if isinstance(attr_names, str):
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -273,7 +273,7 @@ class AgentSet(MutableSet, Sequence):
 
         Returns:
             list[Any]: A list with the attribute value for each agent in the set if attr_names is a str
-            List[List[Any]]: A list with a list of attribute values for each agent in the set if attr_names is a list of str
+            list[list[Any]]: A list with a list of attribute values for each agent in the set if attr_names is a list of str
 
         Raises:
             AttributeError if an agent does not have the specified attribute(s)
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -17,7 +17,7 @@ from collections.abc import Iterable, Iterator, MutableSet, Sequence
 from random import Random
 
 # mypy
-from typing import TYPE_CHECKING, Any, Callable
+from typing import TYPE_CHECKING, Any, Callable, List
 
 if TYPE_CHECKING:
     # We ensure that these are not imported during runtime to prevent cyclic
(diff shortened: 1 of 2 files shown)


mesa/agent.py
```diff
@@ -264,17 +264,24 @@ def do(
...
-        Retrieve a specified attribute from each agent in the AgentSet.
+        Retrieve the specified attribute(s) from each agent in the AgentSet.
 
         Args:
-            attr_name (str): The name of the attribute to retrieve from each agent.
+            attr_names (str | List[str]): The name(s) of the attribute(s) to retrieve from each agent.
 
         Returns:
-            list[Any]: A list of attribute values from each agent in the set.
+            list[Any]: A list of attribute values for each agent in the set.
```
EwoutH, line 275: Is this still correct? Would it be nested in some sort when a list of attrs is inputted?

Maybe make a little bit more explicit what is returned when a str is inputted for attr_names, and what when a list of strings is inputted.

EwoutH (commented):
Thanks, looks useful! One minor questions on the output.

mesa/agent.py
quaquel: see the updated text

quaquel:
I went back and forth on that. In the end, I stuck with the current design which also does not return a dict in case of just one attribute. Moreover, the caller knows the attributes and their order so it has all the information it needs and can easily create a list of dicts if desired.

mesa/agent.py
```diff
@@ -264,17 +264,25 @@ def do(
...
 
         Returns:
-            list[Any]: A list of attribute values from each agent in the set.
+            list[Any]: A list with the attribute value for each agent in the set if attr_names is a str
+            List[List[Any]]: A list with a list of attribute values for each agent in the set if attr_names is a list of str
         """
-        return [getattr(agent, attr_name) for agent in self._agents]
+
+        if isinstance(attr_names, str):
+            return [getattr(agent, attr_names) for agent in self._agents]
```
Corvince, line 280: Maybe we should default to None (getattr(agent, attr_names, None)). This way it would be easier to collect across different agent types, which might or might not have the specified attributes.

quaquel: I see your point but I am torn. I personally probably would prefer getting the error explicitly rather than being confronted with None values downstream.

EwoutH:
Since the output format is different, and there are two separate code paths anyway, I was also thinking if it shouldn't be a separate method. Something like `get_multiple` or so.

quaquel:
I came to this code while playing with the data collection. Somewhere you have to go done the either path (str, list of str). In the end, I saw pushing it into the agentset was the most convenient, but I am fine by separating it here into 2 methods and moving the if upwards

mesa/agent.py
rht: Past discussion: https://github.com/projectmesa/mesa-examples/pull/90. The summary is that it depends on the coder's background: in JavaScript, it is not surprising for `obj.attr` to return `undefined` and not raise an error. However, Python doesn't distinguish between `null` and `undefined`, and so there is an information loss when returning `None`. It is ambiguous and could mean the modeler intentionally chooses to return `None` instead of an absence of attribute.
I prefer no fallback to `None`.

EwoutH: Maybe we could add a boolean switch (something like `allow_undefined`) that defaults to `False`. When `False` it isn't allowed and an error is returned, when True it is allowed and `None` values are returned.

quaquel: that would make the code internally more complicated. You would get two additional nested if/else. 

The easier solution is for the user to do agentset.select("condition to ensure attribute exists").get(). 

Also, why allow None in the first place? It only gives more downstream problems.

Corvince: I think its not necesarily an "error" if we collect None instead of triggering an AttributeError. It depends on the use case. If you quickly want to get some attributes on all of your agents, but some agents don't have that attribute you still get some result. Otherwise you first have to pre-select the right AgentSet and then collect the attributes.

I think the cleanest solution would probably be to mimic getattr in that users can provide a default value for `agentset.get()` and pass that to getattr, if set.

quaquel: implementation wise this is a bit tricky. `default` in `getattr` is not a keyword argument but an argument. So you cannot use `None` to signal you don't specify default.

rht: We can defer this discussion and implementation in a separate PR. The current behavior in this PR is consistent with the existing API, which does not fall back to `undefined`/`None`.

mesa/agent.py
```diff
@@ -264,17 +264,25 @@ def do(
...
 
-    def get(self, attr_name: str) -> list[Any]:
+    def get(self, attr_names: str | list[str]) -> list[Any]:
         """
-        Retrieve a specified attribute from each agent in the AgentSet.
+        Retrieve the specified attribute(s) from each agent in the AgentSet.
 
         Args:
-            attr_name (str): The name of the attribute to retrieve from each agent.
+            attr_names (str | List[str]): The name(s) of the attribute(s) to retrieve from each agent.
```
rht, line 272: The `List` here shouldn't be capitalized.

rht (approved):

mesa/agent.py
rht: The return value also has a capitalized `List`. This may be a small typo, but it can't be automatically fixed by `pyupgrade`.

rht:
Other than the `List` typo, LGTM.

quaquel:
> Other than the `List` typo, LGTM.

The List typo is fixed and I made the attribute error behavior explicit in the docstring

rht:
> The List typo is fixed and I made the attribute error behavior explicit in the docstring

I still see capitalized `List`, and the last commit https://github.com/projectmesa/mesa/pull/2044/commits/76bb633047127920ccc60d659f6a9ed0b918a11f only added the AttributeError documentation.

rht:
Yes, that's why pyupgrade can't parse it, because it's a string. There is 1 more: https://github.com/projectmesa/mesa/pull/2044/files#diff-1dffe661c1ef6a8c750afb07bd029251b7c35f7793b044c5dde893c59236f0c3R272.

quaquel:
> Yes, that's why pyupgrade can't parse it, because it's a string. There is 1 more: https://github.com/projectmesa/mesa/pull/2044/files#diff-1dffe661c1ef6a8c750afb07bd029251b7c35f7793b044c5dde893c59236f0c3R272.

And they are a pain to find in the IDE. But its fixed now.

PR #2048, opened 2024-02-22
Correct wolf energy gained from eating sheep

JackAtOmenApps:
Implementing the already existing class attribute that was correctly used in Sheep, but not in Wolf.


rht (approved):

rht:
Merged, thank you.

EwoutH:
Good catch, thanks!

PR #2050, opened 2024-02-22
fix: Use model.schedule only when it is not None

rht:
This fixes the error that shows up in https://github.com/projectmesa/mesa-examples/pull/79 and https://github.com/projectmesa/mesa-examples/pull/99,
```
        agent_records = map(
            get_reports,
>           model.schedule.agents if hasattr(model, "schedule") else model.agents,
        )
E       AttributeError: 'NoneType' object has no attribute 'agents'
```

### DIFF
--- a/mesa/datacollection.py
+++ b/mesa/datacollection.py
@@ -184,7 +184,9 @@ class DataCollector:
 
         agent_records = map(
             get_reports,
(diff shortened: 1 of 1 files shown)


EwoutH (approved):
Thanks, looks good!

PR #2053, opened 2024-02-25
Schelling

coderbeta1:
Changes Made:

1. Seed: Now it is not compulsory to give a seed value (Random runs every time is now possible)
2. Doc String: Added information about Model Class


benchmarks/Schelling/schelling.py
```diff
@@ -76,8 +93,10 @@ def step(self):
 if __name__ == "__main__":
     import time
 
-    # model = Schelling(15, 40, 40, 3, 1, 0.625)
-    model = Schelling(15, 100, 100, 8, 2, 0.8)
+    # model = Schelling(seed=15, height=40, width=40, homophily=3, radius=1, density=0.625)
```
rht, line 96: Reminder: should change to width first while we have the chance.

rht:
There are still more in the code that has the `height, width` order.

PR #2054, opened 2024-02-25
Improve flocking benchmark 

coderbeta1:
Changes Made:

1. Changed `self.velocity` to `self.direction`: Both Velocity and Speed are agent attributes. This leads to unnecessary confusion. So I changed `velocity` to `direction`
2. Seed: It is now not compulsory to give a `seed` value. This allows for easy randomness if needed.
3. Default values: Code will work even if model values are not given
4. General Changes: Made the code more descriptive for beginners

### DIFF
--- a/benchmarks/Flocking/flocking.py
+++ b/benchmarks/Flocking/flocking.py
@@ -50,7 +50,7 @@ class Boid(mesa.Agent):
             separation: Minimum distance to maintain from other Boids.
             cohere: the relative importance of matching neighbors' positions
             separate: the relative importance of avoiding close neighbors
-            match: the relative importance of matching neighbors' headings
+            match: the relative importance of matching neighbors' directions
 
         """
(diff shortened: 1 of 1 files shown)


benchmarks/Flocking/flocking.py
```diff
@@ -136,15 +137,15 @@ def make_agents(self):
             x = self.random.random() * self.space.x_max
             y = self.random.random() * self.space.y_max
             pos = np.array((x, y))
-            velocity = np.random.random(2) * 2 - 1
+            direction = np.random.random(2) * 2 - 1
```
rht, line 140: I just saw the docstring says "headings" and this needs to be replaced with "directions" as well.

rht:
@coderbeta1 thank you for the needed corrections and improvements to the examples so far.

rht (approved):

Issue #2060, opened 2024-02-28
Make connections in experiment cell spaces named


quaquel:
**What's the problem this feature will solve?**
Currently the connections of Cell are stored as a list in `_connections`. While developing cell space, it was suggested to replace this with a dict with named connections. This would make it possible to get the cell to, e.g., the left or right, top or bottom in case of a von Neumann grid. Once connections are named, all kinds of other movements over the cell space are easily expressed. 

For example, it becomes trivial to retrieve all cells on a diagonal in a Moore grid.

```python
# within an agent class
startcell = self.cell
cell = None
diagonal = []
while cell is not startcell:
    cell = cell.connections['top_left']
    diagonal.append(cell)

```

**Describe the solution you'd like**
The largest challenge is to come to a clear naming of the connections. This is likely to be different across OrthogonalMooreGrid, OrthogonalVonNeumannGrid, and HexGrid. Moreover, it might not work for Network, and become tricky for n-dimensional grids

quaquel:
closed via #2296

PR #2061, opened 2024-02-28
Make connections in experiment cell space named #2060 #features

Abhijit1102:
The proposed feature resolves the limitation in the current representation of cell connections. By replacing the existing list of connections with a dictionary containing named connections, the code enables more expressive and flexible movement over the cell space. Specifically, this change allows for easy retrieval of neighboring cells in different directions, such as left, right, top, or bottom in the context of a von Neumann grid. The enhanced structure opens the door to straightforward expression of various movements and relationships within the cell space.

```python
self.connections: dict[str, Cell] = {}

### DIFF
--- a/mesa/experimental/cell_space/cell.py
+++ b/mesa/experimental/cell_space/cell.py
@@ -24,7 +24,7 @@ class Cell:
 
     __slots__ = [
         "coordinate",
-        "_connections",
+        "connections",
         "agents",
         "capacity",
         "properties",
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by quaquel** [comment]
As indicated in #2060, the real issue is with [`OrthogonalMooreGrid,` `OrthogonalVonNeumannGrid,` and `HexGrid`](https://github.com/projectmesa/mesa/blob/main/mesa/experimental/cell_space/grid.py) and specifically in `_connect_single_cell2d`, assuming we first address this for 2-dimensional grids. This PR, so far, does not address the code there (which is also, in part, why the builds are failing).

File: mesa/experimental/cell_space/cell.py [hunk 1]
```diff
@@ -24,7 +24,7 @@ class Cell:
 
     __slots__ = [
         "coordinate",
-        "_connections",
+        "connections",
```
quaquel, line 27: the choice to keep connections "private" was deliberate because many of the inner workings of discrete space classes rely on it. So, I would not change it here unless there are good reasons for it.

Corvince:
Honestly this PR looks very much AI-generated. Sorry if I am wrong about this @Abhijit1102 , but if there isn't any further progress soon I will close this PR (but feel free to open another one if this was a genuine first version)

PR #2062, opened 2024-02-29
Remove JupyterViz Altair marker overlap for huge grid size

rht:
This is #2049, but for the Altair visualization instead.
20x20
20x20
60x60
60x60
100x100
100x100


quaquel (approved):
LGTM

Issue #1893, opened 2023-11-28
Enable DiscreteEventScheduler to handle system-level events and task interruptions


EwoutH:
The current implementation of the `DiscreteEventScheduler` (as of #1890) could be improved by allowing task interruptions and resumptions, and integrating system-level event handling.

#### Motivation
System-level events, such as policy changes or environmental shifts, can be crucial for simulating real-world scenarios. By enabling agents to interrupt and resume tasks and react to system-wide changes, we allows new systems to be modelled.

#### Description
The current implementation of the `DiscreteEventScheduler` in Mesa works great for simulations where agents operate based on discrete events. However, for more complex scenarios, especially those requiring dynamic interruption and resumption of tasks, as well as system-level event responses, the scheduler could benefit from additional functionalities.

#### Proposed Enhancements:
1. **Enhanced Event Scheduling**:
    - Modify `schedule_event` to check an agent's current task status before scheduling new tasks.
    - Implement logic to handle scenarios where an agent is already engaged in a task (e.g., delay or reschedule the new task).

2. **Basic Interruption Handling**:
    - Introduce a method in `DiscreteEventScheduler` to interrupt an agent's current task.
    - Update the agent's status to reflect the interruption and, optionally, reschedule the interrupted task for a later time.

3. **System-Level Event Integration**:
    - Implement a feature to schedule system-wide events that can affect multiple or all agents.
    - These events could trigger task interruptions or other significant changes in agent behavior.

4. **Resumption Logic**:
    - Develop a mechanism within agents to resume interrupted tasks.
    - This would involve checking their status at the beginning of their `step` method and deciding whether to continue a previous task or start a new one.

There is probably more that could fit in there. I think the most important is to think about the concepts of **tasks** in Mesa, a thing that an agent is currently working on, has some future benefit and can be interrupted by different things.

EwoutH:
I wrote this issue before the whole new DEVS scheduler was in. Now that it is, I would like to revisit it.

I want to introduce especially the concept of a `Task`. A Task takes a non-fixed time, which can be determined stochastically. It might also be able to be interrupted by an other agent or a system level event, and if so, there are multiple options to handle that.

Expect some work on this the next few days/weeks.

CC @quaquel and @Corvince

PR #2066, opened 2024-02-29
support for discrete event scheduling

quaquel:
# Summary
This PR adds an experiment feature that puts event scheduling at the heart of how MESA works. An earlier draft of this code was shown and discussed in #2032. This code generalizes #1890 by making discrete event scheduling central to how mesa models are run. This makes it trivial to maintain discrete time progression while allowing for event scheduling, thus making it possible to build hybrid ABM-DEVS models. 

# Motive
Agent-based models can be quite slow because, typically, they involve the activation of all agents at each tick. However, this can become very expensive if it can be known upfront that the agent is not active. Combining ABM tick-based activation with event scheduling makes it easy to avoid activating dormant agents. 

For example, in Epstein's civil violence model, agents who are in jail need not be activated until they are released from jail. This release from jail can be scheduled as an event, thus avoiding unnecessary agent activations. Likewise, in Wolf-Sheep with grass, the regrowth of grass patches can be scheduled instead of activating all patches for each tick only to decrement a counter. 

# Implementation
The experimental feature adds three core new classes: Simulator, EventList, and SimulationEvent. 

The simulator is analogous to a numerical solver for ODEs. Theoretically, the idea of a Simulator is rooted in the work of [Zeigler](https://www.sciencedirect.com/book/9780128133705/theory-of-modeling-and-simulation), The Simulator is responsible for controlling and advancing time. The EventList is a heapq sorted list of SimulationEvents. SimulationEvents are sorted based on their time of execution, their priority, and their unique_id. A SimulationEvent is, in essence, a callable that is to be executed at a particular simulation time instant. 

This PR adds two specific simulators: ABMSimulator and DEVSSimulator. ABMSimulator uses integers as the base unit of time and automatically ensures that `model.step` is scheduled for each tick. DEVSSimulator uses float as the base unit of time. It allows for full discrete event scheduling. 

Using these new classes requires a minor modification to a Model instance. It needs a simulator attribute to be able to schedule events. 

# Usage
The basic usage is straightforward as shown below. We instantiate an ABMSimulator, instantiate the model, and call `simulator.setup`. Next, we can run the model for, e.g., 100 time steps). 

```python
    simulator = ABMSimulator()
    
    model = WolfSheep(simulator,25, 25, 60, 40, 0.2, 0.1, 20, seed=15,)

    simulator.setup(model)
    simulator.run(100)
    print(model.time)  # prints 100
    simulator.run(50)  
    print(model.time)  # prints 150
```

The simulator comes with a whole range of methods for scheduling events: `schedule_event_now`, `schedule_event_relative`, `schedule_event_absolute`, and the ABMSimulator also has a `schedule_event_next_tick`. See `experimental/devs/examples/*.*` for more details on how to use these methods.


EwoutH:
This looks like an remarkable effort. I will try to review after the weekend.

quaquel:
The benchmarks are currently failing because they rely on a for loop. I'll have to update all benchmarks to use a simulator. Will probably do this tomorrow.

mesa/experimental/devs/eventlist.py
```diff
@@ -0,0 +1,191 @@
...
+            return 1
+
+        if self.unique_id < other.unique_id:
+            return -1
+        if self.unique_id > other.unique_id:
+            return 1
+
+        # theoretically this should be impossible unless it is the
+        # exact same event
+        return 0
```
rht, line 103: Should an exception be raised instead? A duplicate event comparison seems to be a bug to me.

quaquel: I have been thinking about this code myself as well. This is one of the things I want to check when writing unit tests.

1. `__cmp__` should return -1, 0, or 1, so an exception should be raised in the heapq because that needs to produce a complete ordering
2. I think this code is fully redundant and probably not even called (to be tested) because the event lists consist of tuples (time, priority, unique_id, actual_event) which are sorted by heapq.

Corvince: I haven't used heapq before, so not an expert either, but from my understanding and reading this indeed is used by heapq for its sorting, so I do thing its needed and implemented correctly. I think even the 'return 0'  doesn't necessarily require the comment. Because this indeed just compares two events and it is a perfectly valid question to ask "are these 2 events the same event"

Corvince: Okay I should have read the whole implementation first. You are just adding the _to_tuple to the heap, so it might indeed never be called. Is this required for heap? Or could you not add just the event and then it uses the comparison?

Corvince:
Just started to finally look into this, but already want to say great effort and awesome work! I'll have a few questions, but first impression is very good.

mesa/experimental/devs/eventlist.py
quaquel: I have been otherwise occupied over the last few days. Hopefully, I have some time in the coming week to finally test this. Once testes, I'll report back here.

mesa/experimental/devs/eventlist.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def _to_tuple(self):
+        return self.time, self.priority, self.unique_id, self
+
+
+class EventList:
+    """An event list
+
+    This is a heap queue sorted list of events. Events are allways removed from the left. The events are sorted
```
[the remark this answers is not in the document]
quaquel: happy to add the info.

In case of pydsol the main reason for this extra info is that dsol (java) was using a different data structure (this has changed now as far as I know).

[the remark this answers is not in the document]
rht: An info stating the diff between the DEVS design choice in this PR and pydsol would be helpful. It seems that your implementation of EventList is simpler here, or at least, the ordering scheme is clearer.

mesa/experimental/devs/eventlist.py
```diff
@@ -0,0 +1,191 @@
+import itertools
+from enum import IntEnum
+from heapq import heapify, heappop, heappush
+from types import MethodType
+from typing import Any, Callable
+from weakref import WeakMethod, ref
+
+
+class InstanceCounterMeta(type):
+    """Metaclass to make instance counter not share count with descendants
```
Corvince, line 10: Why is it desired that the counter is not shared with descendants? This is used for the unique_id, which, as the name suggests, should be unique. Not sharing the counter might break this

Corvince:
So, just for my understanding, apart from the wonderful event scheduling this PR also introduces, to my understanding, two additional features that are not strictly required to make this work. 

1. ) Simulators that help manage the eventlist 
2. ) A required model.setup function

Its easy to see the motivation for 1) and I totally agree on its abstraction upon Event and EventList, its what makes those features easy to use. I am not so sure on number 2). As far as I can see it this isn't strctly required, right? Maybe I am overlooking something, but otherwise I think this deserves its own discussion and maybe even PR.

Previously I also imagined our other schedulers be built using the EventList, so RandomActivation and SimulatenousActivation built on top. Right now I am not sure this really makes sense, or can be reasonably built from this, but I have to think a bit more about this.

Lastly, I was surprised but the benchmark results. There seems to be a high cost involved in using the eventlist. For some models this is certainly regained by its advantages, but currently the price seems quite high. Have you started to look into this and any idea this could be prevented?

quaquel:
1. Great to hear you agree on this distinction
2. I am not sure about the split between init and setup either and agree it warrants possibly a separate discussion. I'll see if it is easy to refactor the code so it is not needed.
3. I am unsure about rebuilding the other schedulers on top of the event list. I must say that, as evidenced by wolf sheep, the agent set works like a charm for this already. I am slowly coming around to the idea voiced by @EwoutH to deprecate the various schedulers at some point. Some of the ongoing ideas like conditional agent sets, and the group_by method for agent set suggested on matrix only enhance the power of agent set and the declining use case for schedulers. 
4. Like you, I was surprised by the benchmark results. I hope to have some time this week to do some profiling to see where it comes from and how it might be remedied. My hunch is that there is quite some costs in heapifying and heap pushing events. Ideally, the event list/simulator should be optional, at least for now. So the performance of Schelling and Flocking, who don't use the even list is less of an issue. But still, Large WolfSheep should run faster than the current naive implementation without event scheduling.

Corvince:
Found the error on 4). Two, actually. First you had
´´´python
if self.fn() == self.model.step
```
Which always returned false, since it actually executed the function (step) instead of comparing it. And late down the event was executed, meaning the model was stepped twice. And it was always executing the same event.

But even after fixing that the model *still* took ~twice the time. The second issue was that the next event was now scheduled correctly, but before the `simulator.time` was updated, which still caused each step to be executed twice. 

I think the latter issue points towards some code obscurity and hard to follow logic. Even ChatGPT couldn't figure it out from just looking at the code (it immediately spotted the first error). Maybe we can improve the flow of information.

**Comment by Corvince** [comment]
Did some simplification of the eventlist, especially the comparison (now solved by using pythons total ordering and tuple comparisons). Also reintroduced weakreferenced functions (see edit comment on my previous comment)

**Comment by quaquel** [comment]
Thanks for this! The overhead is now small even for Schelling and boidflockers, while WolfSheep has a nice speedup.

File: mesa/experimental/devs/simulator.py [hunk 4]
```diff
@@ -0,0 +1,265 @@
...
+        self.event_list.clear()
+        self.model = model
+        model.setup()
+
+    def reset(self):
+        """Reset the simulator by clearing the event list and removing the model to simulate"""
+        self.event_list.clear()
+        self.model = None
+
+    def run(self, time_delta: int | float):
```
Corvince, line 56: I think the naming isn't optimal here. Run sounds very generic, but the condition is very specific to run for a specified amount of time. 

The implementation also isn't correct, because it calls step() until the current time is smaller than the end time. But the step just calls the next event, which might be later than the end time. This already happens in the benchmarks example, which run for 1 step too much.

mesa/experimental/devs/simulator.py
```diff
@@ -0,0 +1,265 @@
...
+    """
+
+    # TODO: add replication support
+    # TODO: add experimentation support
+
+    def __init__(self, time_unit: type, start_time: int | float):
+        # should model run in a separate thread,
+        # and we can then interact with start, stop, run_until, and step?
+        self.event_list = EventList()
+        self.time = start_time
```
EwoutH, line 28: In #1942 we added a clock to the model, in `self._steps` and `self._time`. How do you - conceptually - see this integrating on the long term? i.e. can there be one scheduler that commands multiple models, and is then the scheduler time leading? Is there one model with one scheduler? If so, which is the "atomic clock" so to speak.

mesa/experimental/devs/simulator.py
```diff
@@ -0,0 +1,265 @@
...
+        self.time_unit = time_unit
+        self.model = None
+
+    def check_time_unit(self, time: int | float) -> bool:
+        ...
+
+    def setup(self, model: "Model") -> None:
```
EwoutH, line 35: Does this mean every model that uses a Simulator should have a setup() function? What's the motivation by separating this from the init?

mesa/experimental/devs/simulator.py
```diff
@@ -0,0 +1,265 @@
...
+            The basic assumption of the simulator is that a Model has a model.setup method that sets up the
+            model.
+
+        """
+
+        self.event_list.clear()
+        self.model = model
+        model.setup()
+
+    def reset(self):
```
EwoutH, line 51: Should the time reset when resetting the simulator? Maybe as an option?

mesa/experimental/devs/simulator.py
EwoutH: Agreed. Maybe `simulate_time` or `progress_time`?

mesa/experimental/devs/simulator.py
```diff
@@ -0,0 +1,265 @@
...
+            self.time,
+            function,
+            priority=priority,
+            function_args=function_args,
+            function_kwargs=function_kwargs,
+        )
+        self._schedule_event(event)
+        return event
+
+    def schedule_event_absolute(
```
EwoutH, line 104: Maybe `schedule_event_at`?

mesa/experimental/devs/simulator.py
```diff
@@ -0,0 +1,265 @@
...
+            time,
+            function,
+            priority=priority,
+            function_args=function_args,
+            function_kwargs=function_kwargs,
+        )
+        self._schedule_event(event)
+        return event
+
+    def schedule_event_relative(
```
EwoutH, line 138: Maybe `schedule_event_in`?

mesa/experimental/devs/simulator.py
```diff
@@ -0,0 +1,265 @@
...
+        end_time = self.time + time_delta
+        while self.time < end_time:
+            self.step()
+
+    def step(self):
+        event = self.event_list.pop_event()
+        self.time = event.time
+        event.execute()
+
+    def schedule_event_now(
```
EwoutH, line 74: Do you think it would improve code quality to make one base function and let the other 2 or 3 depend on it? There is quite some replication between the different event functions.

Schedule now can also just be relative schedule with `0` I think.

mesa/experimental/devs/simulator.py
```diff
@@ -0,0 +1,265 @@
...
+    for hybrid ABM-DEVS simulations.
+
+    """
+
+    def __init__(self):
+        super().__init__(int, 0)
+
+    def setup(self, model):
+        super().setup(model)
+        self.schedule_event_now(self.model.step, priority=Priority.HIGH)
```
EwoutH, line 204: Are we always directly stepping after setup? Could you expand on this a bit?

mesa/experimental/devs/simulator.py
```diff
@@ -0,0 +1,265 @@
...
+        if not self.check_time_unit(event.time):
+            raise ValueError(
+                f"time unit mismatch {event.time} is not of time unit {self.time_unit}"
+            )
+
+        # check timeunit of events
+        self.event_list.add_event(event)
+
+
+class ABMSimulator(Simulator):
```
EwoutH, line 189: This should be a bit comparable to the current `DiscreteEventScheduler` right?

benchmarks/WolfSheep/wolf_sheep.py
```diff
@@ -94,26 +93,42 @@ class GrassPatch(CellAgent):
...
 
-    def step(self):
         if not self.fully_grown:
-            if self.countdown <= 0:
-                # Set as fully grown
-                self.fully_grown = True
-                self.countdown = self.model.grass_regrowth_time
-            else:
-                self.countdown -= 1
+            self.model.simulator.schedule_event_relative(
```
EwoutH, line 126: If the grow time is longer than the tick time, won't this schedule multiple events before it's fully grown?

benchmarks/WolfSheep/wolf_sheep.py
```diff
@@ -94,26 +93,42 @@ class GrassPatch(CellAgent):
     A patch of grass that grows at a fixed rate and it is eaten by sheep
     """
 
-    def __init__(self, unique_id, model, fully_grown, countdown):
+    @property
+    def fully_grown(self):
+        return self._fully_grown
+
+    @fully_grown.setter
+    def fully_grown(self, value: bool) -> None:
```
EwoutH, line 101: This feels duplicate with the part below somehow. Why do we need these properties and setters?

EwoutH (commented):
I haven't fully internalised all the implications and possibilities yet, but in my initial view, I like it. It seems to offer a powerful framework for both standard ABM, DEVS, and everything in between. It's also nice having everything in the experimental space and I like the powerful functionality like being able to set priorities to events. That might also help with #1893.

I have a few nitpicks, in the separate comments.

mesa/experimental/devs/eventlist.py
```diff
@@ -0,0 +1,162 @@
...
+    def pop_event(self) -> SimulationEvent:
+        """pop the first element from the event list"""
+        while self._events:
+            event = heappop(self._events)
+            if not event.CANCELED:
+                return event
+        raise IndexError("Event list is empty")
+
+    def is_empty(self) -> bool:
+        return len(self) == 0
```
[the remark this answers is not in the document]
quaquel: I prefer to be explicit in this particular case so I prefer to explicitly check if the length is zero rather than rely on the implicit boolean behavior.

mesa/experimental/devs/eventlist.py
```diff
@@ -0,0 +1,162 @@
...
+        time (float): The simulation time of the event
+        function  (Callable): The function to execute for this event
+        priority (Priority): The priority of the event
+        unique_id (int) the unique identifier of the event
+        function_args (list[Any]): Argument for the function
+        function_kwargs (Dict[str, Any]): Keyword arguments for the function
+
+    """
+
+    _ids = itertools.count(1)
```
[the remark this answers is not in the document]
quaquel: we can fully remove any numbers here. `itertools.count` takes an optional start and step number which default to 0 and 1 respectively.

mesa/experimental/devs/eventlist.py
```diff
@@ -0,0 +1,162 @@
...
+    def __lt__(self, other):
+        # Define a total ordering for events to be used by the heapq
+        return (self.time, self.priority, self.unique_id) < (
+            other.time,
+            other.priority,
+            other.unique_id,
+        )
+
+
+class EventList:
```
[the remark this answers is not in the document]
quaquel: no, and I would not make this part of this PR. If you want to track events, you better add some kind of memory to your agents.

mesa/experimental/devs/simulator.py
quaquel: In the long term, this should replace `self._steps` and `self._time` in `Model`. The simulator would become the sole truth for time.

mesa/experimental/devs/simulator.py
quaquel: I prefer relative and absolute because it is clearer to me.

mesa/experimental/devs/simulator.py
quaquel: I prefer relative and absolute because it is clearer to me.

mesa/experimental/devs/simulator.py
quaquel: `schedule_event_now` can indeed be simpler. I don't see any other obvious simplifications

mesa/experimental/devs/simulator.py
quaquel: It generalizes it.

benchmarks/WolfSheep/wolf_sheep.py
quaquel: it is not a nice solution I agree. The issue is that events take a callable, so I used the set_fully_grown function for this. It can be done cleaner by using `setattr` and `function_args`.

mesa/experimental/devs/simulator.py
quaquel: Happy to change the name, but not sure yet how. Basically this method will run the model for the provided time_delta. So, `run_for` or `progress_time_by`, or `simulate_for` all seem ok.

@Corvince, I am not sure I fully follow. It currently runs while the clock time (i.e., `self.time`) is smaller than `end_time`. So, yes, it will change the time until self.time equals end_time in case of ABMSimulator. This means that time_delta is inclusive. If you say `run(10)`, it will run until time equals 10 which is in fact 11 steps. I don't see an easy solution around this. 

A few options:
1. shift to using `eventlist.peak_ahead()` to check the timestamp of each next event before we execute this event. This creates a bit of overhead. 
2. Schedule the first step for next tick (rather than the current `schedule_event_now`). This implicitly means we shift from zero based to 1 based indexing because the first step happens at `time=1`.
3. solve it in simulator.step by not executing the event if the time becomes to large and instead reschedule this event for its specified time. It needs to be rescheduled because we have gotten it from the eventlist using pop.

EwoutH:
```Python
run_for(time_delta=something)
````
seems good

File: mesa/experimental/devs/eventlist.py (hunk 17 above)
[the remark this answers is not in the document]
Comment (EwoutH) [reply mesa/experimental/devs/eventlist.py:85]: That's fine, could you add a note somewhere in the PR message this could be future work?

File: mesa/experimental/devs/simulator.py (hunk 5 above)
Comment (EwoutH) [reply mesa/experimental/devs/simulator.py]: - Right, in that case, can we still https://github.com/projectmesa/mesa/discussions/1912?

File: mesa/experimental/devs/eventlist.py (hunk 17 above)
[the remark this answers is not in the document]
Comment (quaquel) [reply mesa/experimental/devs/eventlist.py:85]: I am unsure whether you want to handle this within mesa by default.

File: mesa/experimental/devs/simulator.py (hunk 5 above)
Comment (quaquel) [reply mesa/experimental/devs/simulator.py]: Yes. It will probably make it even easier.

File: mesa/experimental/devs/simulator.py (hunk 8 above)
Comment (EwoutH) [reply mesa/experimental/devs/simulator.py:98]: I think "now", "at" and "in" are more natural concepts of time. Beautiful Python reads like English. It would also be consistent with the current implementation.

https://github.com/projectmesa/mesa/blob/57d9cb565564cf8d4df9f68e5ca1ebb20dd99397/mesa/time.py#L472

@Corvince break the tie for us?

File: mesa/experimental/devs/simulator.py (hunk 11 above)
Comment (EwoutH) [reply mesa/experimental/devs/simulator.py:198]: But do I want to always schedule a model step? Maybe I want to do something else first? Or have no model steps at all? It's not obvious to me that initializing a simulator should result in the model doing it's first step. But maybe I'm missing something here.

File: benchmarks/WolfSheep/wolf_sheep.py (hunk 13 above)
Comment (EwoutH) [reply benchmarks/WolfSheep/wolf_sheep.py:128]: I guess I have no idea how this works.

- The step method is removed.
- One event is scheduled in the init
- There are properties and setters
- there is a set_fully_grown() method

I don't see where after the init (which happens only on model initialisation right?) grass is regrown when it's being eaten. Shouldn't that be a method that can be triggered when the grass is eaten?

File: mesa/experimental/devs/simulator.py (hunk 11 above)
Comment (quaquel) [reply mesa/experimental/devs/simulator.py:198]: You can always do something else first. If you want to execute events before the first step, you can give it a higher priority than `model.step`. If you don't want model steps at all, you should not use the `ABMSimulator,` the entire point of which is to have discrete-time advancement as normally encountered in ABM models.  So, everything you want is possible with the functionality provided.

File: benchmarks/WolfSheep/wolf_sheep.py (hunk 13 above)
Comment (EwoutH) [reply benchmarks/WolfSheep/wolf_sheep.py:128]: Oh wait does grass die if it's eaten? And will a new grass be created when not?

I was under the presumption that grass exists on each cell, is either grown or now, and covers the full space. But it's actually sparse, and some grass is created on initialisation, and spreads to other cells, creating new Agents on those locations. When it's eaten it's removed. Just like the sheep if they are eaten by a wolf.

If this is right never mind my earlier comments, this can be marked resolved.

**Comment by EwoutH** [comment]
Performance improvement looks really nice for wolf-sheep! I should have replied to all open conversations.

File: benchmarks/WolfSheep/wolf_sheep.py (hunk 13 above)
Comment (quaquel) [reply benchmarks/WolfSheep/wolf_sheep.py:128]: I have simplified the code to just

```python
    @fully_grown.setter
    def fully_grown(self, value: bool) -> None:
        self._fully_grown = value

        if not value:
            self.model.simulator.schedule_event_relative(
                setattr,
                self.grass_regrowth_time,
                function_args=[self, "fully_grown", True],
            )
```

This is used in conjunction with a `Sheep`, who when eating grass does `grass_patch.fully_grown = False`. So, this triggers the setter. The setter, in turn, schedules the event for when it becomes fully grown again. This is just a `setattr` call with the provided arguments. Because value in this case is True, we don't schedule a new event. 

Hope this helps.

EwoutH: Right, looks good. Thanks!

mesa/experimental/devs/simulator.py
quaquel: In the latest update, I added run_for and run_until. This also fixes the problem of running one step too much.

Still need to check a few things, but its almost there.

mesa/experimental/devs/simulator.py
quaquel: Sorry, but I disagree with this fundamentally, "In" is very ambiguous because it can mean in a specific tick (so the same as "at"). Absolute and relative are just clear.

quaquel:
I have added unit tests and most of the code is covered. This should be almost ready to be merged.

EwoutH:
These benchmarks ❤️!

Do you know why the BoidFlockers is so much faster?

I will try to review tomorrow.

quaquel:
> Do you know why the BoidFlockers initialization is so much faster?

This branch was not up to date with main, so something in main should explain that. My hunch is that it is #2083.

tpike3:
This is amazing!

The only real block I really saw was the merge conflict so I resolved the merge conflict in flockers benchmark. 

As  Ewout, rht and Corvince have all reviewed this, I don't think I will provide any more value added. 

I would ask one of them to do the actual merge if they are good with it and then per the dev session we can release 2.3.0.

quaquel:
I am meeting with @EwoutH in a bit. I'll discuss the PR and any possible open issues.

mesa/experimental/devs/simulator.py
EwoutH: Okay, let's go with the more explicit ones.

EwoutH:
I think the only thing that needed to be done was to remove the old DiscreteEventScheduler and put a warning there, right?

quaquel:
That is the most important thing indeed. I might have some time over the weekend to finalize this.

quaquel:
> I think the only thing that needed to be done was to remove the old DiscreteEventScheduler and put a warning there, right?

`time.DiscreteEventScheduler` is now largely removed and an Exception is raised if a user tries to instantiate it. I have also removed the associated tests and fixed some outstanding ruff issues. I think this is ready to be merged.

EwoutH (approved):

EwoutH:
@quaquel in my model I'm doing:
```Python
simulator1 = DEVSimulator()
model1 = UrbanModel(n_agents=5, simulator=simulator1)
simulator1.setup(model1)
```
In the initialisation the first events are scheduled. However, when I then do `simulator1.setup(model1)`, to associate the model with the simulator, it clears the event queue:

https://github.com/projectmesa/mesa/blob/1202e2e28a3bfc00f285b66087bd2ab8efce1cc1/mesa/experimental/devs/simulator.py#L48

Luckily this is quite easily circumvented by replacing the last line with `simulator1.model = model1`, instead of doing the setup.

However, I'm quite afraid most users won't discover this, since all the examples use the `setup()` method. So my concrete questions:

- What is the rationale between clearing the event queue on `simulator.setup()`?
- Should this clearing be identical between the `ABMSimulator` and `DEVSimulator`? (since the ABMSimulator does schedule an initial event, but the DEVSimulator doesn't)
- How do we communicate this to users in our documentation and examples?

EwoutH:
Another question: Is simulator time (`model.simulator.time`) kept in some way in sync with model time (`model._time`)? Or are your expected to only use simulator time when using a `DEVSimulator()`?

quaquel:
> What is the rationale between clearing the event queue on simulator.setup()?

Running a model multiple times in a row. I also just followed dsol here.

> Should this clearing be identical between the ABMSimulator and DEVSimulator?

Ideally yes, there should be a common way to reset a model to its initial state. This could be done on the model or the simulator. I need to think whether I have a preference.

> How do we communicate this to users in our documentation and examples?

Let's first figure out what we want. I personally have in the passed pushed for a model.init and model.reset method. This would avoid the problem you ran into because any event scheduling would go to model.init. So you get something like.

```python
simulator = DevsSimulator()
model = myModel()
simulator.setup(model)
model.init()
simulator.run()
```

> Another question: Is simulator time (model.simulator.time) kept in some way in sync with model time (model._time)? Or are you expected to only use simulator time when using a DEVSimulator()?

I don't recall, so I would need to check the code. In my vision for MESA 3, time should be kept by the simulator, not by the model.

EwoutH:
Thanks for the comments, they are useful and it's good to have this documented. I got it working as need for my model right now, but it also shows it's not directly ready to be stabilized. It requires a few more iterations of building models --> discovering limitations --> implementing feedback --> building models.

When you figure out how the internals work it's quite powerful though!

PR #2067, opened 2024-03-01
Enhance AgentSet.get to handle undefined attributes with fallback opt…

ai-naymul:
Issue no : #2045 
These chanegs provides a more user friendly approach to attribute retrieval, accommodating scenarios where some agents might not have all the specified attributes.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -90,7 +90,7 @@ class AgentSet(MutableSet, Sequence):
 
     Methods:
         __len__, __iter__, __contains__, select, shuffle, sort, _update, do, get, __getitem__,
-        add, discard, remove, __getstate__, __setstate__, random, group_by
+        add, discard, remove, __getstate__, __setstate__, random
 
     Note:
         The AgentSet maintains weak references to agents, allowing for efficient management of agent lifecycles
@@ -254,7 +254,7 @@ class AgentSet(MutableSet, Sequence):
     def get(self, attr_names: str | list[str], fallback_value=sentinel) -> list[Any]:
         """
         Retrieve the specified attribute(s) from each agent in the AgentSet.
-
+        
         Args:
             attr_names (str | list[str]): The name(s) of the attribute(s) to retrieve from each agent.
             fallback_value (Any, optional): The value to return if the attribute is not found. If not provided, an AttributeError is raised.
@@ -266,21 +266,17 @@ class AgentSet(MutableSet, Sequence):
         Raises:
             AttributeError if an agent does not have the specified attribute(s) and no fallback_value is provided.
         """
-
-        def get_attr(agent, attr_name):
-            try:
+        def get_attr(agent, attr_name, fallback_value=None):
+            if hasattr(agent, attr_name):
                 return getattr(agent, attr_name)
-            except AttributeError:
-                if fallback_value is not self.sentinel:
-                    return fallback_value
-                else:
-                    raise
+            else:
+                return fallback_value if fallback_value is not self.sentinel else None
 
         if isinstance(attr_names, str):
-            return [get_attr(agent, attr_names) for agent in self._agents]
+            return [get_attr(agent, attr_names, fallback_value) for agent in self._agents]
         else:
             return [
-                [get_attr(agent, attr_name) for attr_name in attr_names]
+                [get_attr(agent, attr_name, fallback_value) for attr_name in attr_names]
                 for agent in self._agents
             ]
 
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -254,7 +254,7 @@ class AgentSet(MutableSet, Sequence):
     def get(self, attr_names: str | list[str], fallback_value=sentinel) -> list[Any]:
         """
         Retrieve the specified attribute(s) from each agent in the AgentSet.
-        
+
         Args:
             attr_names (str | list[str]): The name(s) of the attribute(s) to retrieve from each agent.
             fallback_value (Any, optional): The value to return if the attribute is not found. If not provided, an AttributeError is raised.
@@ -266,6 +266,7 @@ class AgentSet(MutableSet, Sequence):
         Raises:
             AttributeError if an agent does not have the specified attribute(s) and no fallback_value is provided.
         """
+
         def get_attr(agent, attr_name):
             try:
                 return getattr(agent, attr_name)
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -90,7 +90,7 @@ class AgentSet(MutableSet, Sequence):
 
     Methods:
         __len__, __iter__, __contains__, select, shuffle, sort, _update, do, get, __getitem__,
-        add, discard, remove, __getstate__, __setstate__, random
+        add, discard, remove, __getstate__, __setstate__, random, group_by
 
     Note:
         The AgentSet maintains weak references to agents, allowing for efficient management of agent lifecycles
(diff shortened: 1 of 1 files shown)


Corvince:
I think the pr generally looks good, thanks for the contribution!

But I still think we can get away with just having a fallback value. The way this works is we have to set it's default value to a sentinel value and then check if the value is still the sentinel value and if not, return the fallback value. This way the fallback value can also be None.

The sentinel value could look just like
```python
sentinel = object()
```

quaquel:
I like the basic proposed solution with the inner function. This is quite elegant. A small addition from my side: could you add some unit tests as well? The test code is [here](https://github.com/projectmesa/mesa/blob/main/tests/test_agent.py).

mesa/agent.py
```diff
@@ -249,27 +249,42 @@ def do(
...
+        attr_names: str | list[str],
+        handle_undefined: bool = True,
+        fallback_value=None,
+    ) -> list[Any]:
         """
         Retrieve the specified attribute(s) from each agent in the AgentSet.
 
         Args:
             attr_names (str | list[str]): The name(s) of the attribute(s) to retrieve from each agent.
+            handle_undefined (bool, optional): If False, use fallback_value for undefined attributes. Defaults to True.
```
rht, line 263: I think it would be clearer is it is called `error_when_undefined`, or if there is a more concise naming for it.

ai-naymul:
> I like the basic proposed solution with the inner function. This is quite elegant. A small addition from my side: could you add some unit tests as well? The test code is [here](https://github.com/projectmesa/mesa/blob/main/tests/test_agent.py).


Hi there,
I will add those shortly...
Thanks a lot for your feedback 💯

mesa/agent.py
ai-naymul: Hi @rht ,

Made some changes into the latest commit could you please have a look and let me know if is there anything to change :)

ai-naymul:
> I think the pr generally looks good, thanks for the contribution!
> 
> But I still think we can get away with just having a fallback value. The way this works is we have to set it's default value to a sentinel value and then check if the value is still the sentinel value and if not, return the fallback value. This way the fallback value can also be None.
> 
> The sentinel value could look just like
> 
> ```python
> sentinel = object()
> ```

Hi @Corvince ,

Could you please have a look at the latest commit and let me know if is that fine, let me know...

Corvince:
Hi, yes! Thats exactly how I meant it. Now you could remove the "error_when_undefined" boolean, because you can either raise an Error if the fallback_value is undefined or use the fallback value. 

I also think its not necessary to include the sentinel value in the docstrings and documentation. Its purely an implementation detail and not something the user needs to be aware of. So its enough to state that, if provided, the fallback value will be used if an attribute error is encountered and otherwise an error is raised.

Corvince:
You might be wondering: If you leave fallback_value undefined and turn off error when undefined, it might appear the code is just returning nothing. But it will actually return None, so its the same as setting fallback_value to None, so error_when_undefined doesn't provide any additional functionality and thats why I think it can be removed.

ai-naymul:
> You might be wondering: If you leave fallback_value undefined and turn off error when undefined, it might appear the code is just returning nothing. But it will actually return None, so its the same as setting fallback_value to None, so error_when_undefined doesn't provide any additional functionality and thats why I think it can be removed.

Yeah, you are right!

ai-naymul:
> Hi, yes! Thats exactly how I meant it. Now you could remove the "error_when_undefined" boolean, because you can either raise an Error if the fallback_value is undefined or use the fallback value.
> 
> I also think its not necessary to include the sentinel value in the docstrings and documentation. Its purely an implementation detail and not something the user needs to be aware of. So its enough to state that, if provided, the fallback value will be used if an attribute error is encountered and otherwise an error is raised.

Hey I have done some changes according to your preference.Let me know if its perfect or not...

Thank you 👍

mesa/agent.py
```diff
@@ -90,7 +90,7 @@ class AgentSet(MutableSet, Sequence):
 
     Methods:
         __len__, __iter__, __contains__, select, shuffle, sort, _update, do, get, __getitem__,
-        add, discard, remove, __getstate__, __setstate__, random
+        add, discard, remove, __getstate__, __setstate__, random, group_by
```
Corvince, line 93: I guess this is from another branch?

mesa/agent.py
```diff
@@ -249,27 +249,38 @@ def do(
...
         Raises:
-            AttributeError if an agent does not have the specified attribute(s)
-
+            AttributeError if an agent does not have the specified attribute(s) and no fallback_value is provided.
         """
 
+        def get_attr(agent, attr_name):
+            try:
+                return getattr(agent, attr_name)
+            except AttributeError:
```
Corvince, line 273: Catching an exception is comparatively costly in terms of performance. Normally this isn't something to worry about, but if we run into this frequently (which users might do) it could become a problem. I think it would be better to first check for the sentinel value and then either return `getattr(agent, attr_name)` or ` getattr(agent, attr_name, fallback_value)`.

Corvince:
> > Hi, yes! Thats exactly how I meant it. Now you could remove the "error_when_undefined" boolean, because you can either raise an Error if the fallback_value is undefined or use the fallback value.
> > I also think its not necessary to include the sentinel value in the docstrings and documentation. Its purely an implementation detail and not something the user needs to be aware of. So its enough to state that, if provided, the fallback value will be used if an attribute error is encountered and otherwise an error is raised.
> 
> Hey I have done some changes according to your preference.Let me know if its perfect or not...
> 
> Thank you 👍

Almost 🙈 One more comment/consideration

mesa/agent.py
ai-naymul: I think thats was a mistake by me. Let me fix it :)

ai-naymul:
> > > Hi, yes! Thats exactly how I meant it. Now you could remove the "error_when_undefined" boolean, because you can either raise an Error if the fallback_value is undefined or use the fallback value.
> > > I also think its not necessary to include the sentinel value in the docstrings and documentation. Its purely an implementation detail and not something the user needs to be aware of. So its enough to state that, if provided, the fallback value will be used if an attribute error is encountered and otherwise an error is raised.
> > 
> > 
> > Hey I have done some changes according to your preference.Let me know if its perfect or not...
> > Thank you 👍
> 
> Almost 🙈 One more comment/consideration

Hi could you please have a look at the recent commit and let me know if is there any issue...

mesa/agent.py
ai-naymul: Hey facing some attribute error in the build after the changes, any ideas.....

quaquel:
> Hi could you please have a look at the recent commit and let me know if is there any issue...

All existing tests are failing, and no tests have been added to the new functionality. Both issues need addressing. If you need help or tips, please let us know.

ai-naymul:
> > Hi could you please have a look at the recent commit and let me know if is there any issue...
> 
> All existing tests are failing, and no tests have been added for the new functionality. Both issues need addressing. If you need help or tips, please let us know.

Actually after this commit(https://github.com/projectmesa/mesa/pull/2067/commits/fe6fea37558200f9f2e1e7b640b4726156fe7d4e) all test are failing. And regardin the test need to add for the changes I will add that shortly after the build is succeed. Sorry for the inconvenience :(

ai-naymul:
@EwoutH,
any ideas on how can fix the build issue?

EwoutH:
Yes, it seems that currently one of the tests [fail](https://github.com/projectmesa/mesa/actions/runs/8207353176/job/22448467124?pr=2067#step:8:37). It’s `test_agentset_get_attribute`, which indicates something is changed in the AgentSet.get() method that breaks that test. The test code is located here:

https://github.com/projectmesa/mesa/blob/be5fc0147c8c6f557f4a340b04556095251df3b8/tests/test_agent.py#L213
Most likely a failing test points to a break in current functionality. But it could also be that the test is unnecessary (strict).

so either the code needs to be fixed to pass the test, or the test modified.

I hope this helps. If you have further questions (or want me to just fix the test). You will learn a lot working though these issues though!

ai-naymul:
> Yes, it seems that currently one of the tests [fail](https://github.com/projectmesa/mesa/actions/runs/8207353176/job/22448467124?pr=2067#step:8:37). It’s `test_agentset_get_attribute`, which indicates something is changed in the AgentSet.get() method that breaks that test. The test code is located here:
> 
> https://github.com/projectmesa/mesa/blob/be5fc0147c8c6f557f4a340b04556095251df3b8/tests/test_agent.py#L213
> 
> 
> Most likely a failing test points to a break in current functionality. But it could also be that the test is unnecessary (strict).
> so either the code needs to be fixed to pass the test, or the test modified.
> 
> I hope this helps. If you have further questions (or want me to just fix the test). You will learn a lot working through these issues though!

Yeah, Let me try to fix that issue ...

mesa/agent.py
```diff
@@ -249,27 +249,37 @@ def do(
...
             list[list[Any]]: A list with a list of attribute values for each agent in the set if attr_names is a list of str
 
         Raises:
-            AttributeError if an agent does not have the specified attribute(s)
-
+            AttributeError if an agent does not have the specified attribute(s) and no fallback_value is provided.
         """
 
+        def get_attr(agent, attr_name, fallback_value=None):
+            if hasattr(agent, attr_name):
```
quaquel, line 271: replace this with try except, that is the python way.

mesa/agent.py
```diff
@@ -249,27 +249,37 @@ def do(
 
         return res if return_results else self
 
-    def get(self, attr_names: str | list[str]) -> list[Any]:
+    sentinel = object()
+
+    def get(self, attr_names: str | list[str], fallback_value=sentinel) -> list[Any]:
         """
```
quaquel, line 255: I would model this on dict.get, so don't use a keyword argument but use *args.

PR #2069, opened 2024-03-05
feat: Display model seed & allow user to specify it in JupyterViz

rht:
Fixes: #1812. The motivation can be found in that issue. It looks like this
2024-03-05T03:36:30,978108357-05:00
The current behavior is that the seed is not changed even when you press the reset button. After all, it is now a user input param. However, if you force refresh the browser, the seed is still not changed. A more sustainable solution (so that user doesn't accidentally lose the seed) is to add a button to generate a new seed.

### DIFF
--- a/mesa/experimental/jupyter_viz.py
+++ b/mesa/experimental/jupyter_viz.py
@@ -157,7 +157,7 @@ def JupyterViz(
         if measures:
             layout_types += [{"Measure": elem} for elem in range(len(measures))]
 
-        grid_layout_initial = get_initial_grid_layout(layout_types=layout_types)
+        grid_layout_initial = make_initial_grid_layout(layout_types=layout_types)
         grid_layout, set_grid_layout = solara.use_state(grid_layout_initial)
 
         with solara.Sidebar():
@@ -385,20 +385,15 @@ def make_text(renderer):
     return function
 
 
-def get_initial_grid_layout(layout_types):
-    grid_lay = []
-    y_coord = 0
-    for ii in range(len(layout_types)):
-        template_layout = {"h": 10, "i": 0, "moved": False, "w": 6, "y": 0, "x": 0}
-        if ii == 0:
-            grid_lay.append(template_layout)
-        else:
-            template_layout.update({"i": ii})
-            if ii % 2 == 0:
-                template_layout.update({"x": 0})
-                y_coord += 16
-            else:
-                template_layout.update({"x": 6})
-            template_layout.update({"y": y_coord})
-            grid_lay.append(template_layout)
-    return grid_lay
+def make_initial_grid_layout(layout_types):
+    return [
+        {
+            "i": i,
+            "w": 6,
+            "h": 10,
+            "moved": False,
+            "x": 6 * (i % 2),
+            "y": 16 * (i - i % 2),
+        }
+        for i in range(len(layout_types))
+    ]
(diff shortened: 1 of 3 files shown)


EwoutH:
Sounds useful, thanks! I will leave the formal review to Corvince.

One thought: Maybe we can have separate buttons for requesting a new seed and resetting the simulation to it's initial state.

mesa/experimental/jupyter_viz.py
```diff
@@ -71,6 +71,7 @@ def JupyterViz(
             simulations with no space to visualize should
             specify `space_drawer=False`
         play_interval: play interval (default: 150)
+        seed: the random seed used to initialize the model
```
tpike3, line 74: Can we explicitly pass in the random model seed so it is obvious to the user where the seed is coming form and what they are changing is they pass in their own?

rht: This is already an explicit argument. Specified by adding a "seed" entry to the `model_params` dictionary: https://github.com/projectmesa/mesa-examples/blob/main/examples/schelling_experimental/app.py#L21.

rht: I forgot to remove that line from the docstring.

rht:
Update:
- simplified the implementation by separating the seed from model parameters. The diff is smaller now.
- changed the play widget color to be blue
- added back the seed as an optional argument
- added a reseed button

2024-03-06T12:15:51,712164823-05:00


To test this PR, you need to modify your model.py, `__init__`, e.g. for Schelling, to accept seed as an argument
```python
    def __init__(self, width=20, height=20, density=0.8, minority_pc=0.2, homophily=3, seed=0):
        super().__init__(seed=seed)
```

Corvince:
> To test this PR, you need to modify your model.py, `__init__`, e.g. for Schelling, to accept seed as an argument
> 
> ```python
>     def __init__(self, width=20, height=20, density=0.8, minority_pc=0.2, homophily=3, seed=0):
>         super().__init__(seed=seed)
> ```

Hm, does that mean all the models need to accept a `seed` parameter? I don't think thats optimal

rht:
> Hm, does that mean all the models need to accept a seed parameter? I don't think thats optimal

`make_model` can be modified to have a try-except to catch the `TypeError: Schelling.__init__() got an unexpected keyword argument 'seed'` error to do a fallback to not showing the seed controls.

mesa/experimental/jupyter_viz.py
rht: > Scenario 1: User does not specify the seed so we want the UI to show the seed that was selected. Shouldn't the display of the random seed reference [obj._seed](https://github.com/projectmesa/mesa/blob/8a50f9f36392b95824e0e2768dcf6f16515514d8/mesa/model.py#L58)

The direction of the cause and effect should flow from the controls you see on the screen, and then the model properties. It can't be the `obj._seed` to have a value before the control.

If you notice, the seed value is displayed twice. The one below the information card is `model._seed`, displayed for debugging purpose. I will remove this once you agree to merge.

tpike3:
LGTM - This is very cool, the model seed is one of those critical details that have burned many a modeler and I think making it explicit is an excellent idea. 

To merge it is (1) fixing the tests and (2) removing the seed below the information card. Thanks @rht

rht:
@tpike3 sorry for the slow response, still recovering. I fixed the test by using `pytest-mock` instead of `unittest` mock. And removed the seed below the information card. I request this PR to be rebased & merged instead of squashed & merged, because it has the 2 unrelated commits:
1. Change the play/stop button to blue
2. Code cleanup

tpike3 (approved):
LGTM - I also like the change to the `current_step` as a parameter, at least for me much more intuitive.

PR #2072, opened 2024-03-06
Better Typing support for (Cell)agents

Corvince:
I recently enabled Pyright for my editor, its a python type checker which is much faster then mypy and therefore actually usable. At least thats my experience. 

Anyway, my experience with Mesa was very humbling. Take for example this code snippet from the schelling benchmark model

```python
        neighborhood = self.cell.neighborhood(radius=self.radius)
        for neighbor in neighborhood.agents:
            if neighbor.type == self.type:
                similar += 1
            [...]
            self.model.happy += 1
```

Lets start from bottom to top. The first typechecking error comes from `self.model.happy`. Since we hardcoded the model to be of type mesa.Model, it doesn't know that model.happy exists and is an integer. This was rather easy to fix with my first commit, where the Agent/CellAgent now can take a generic parameter for the model class. So you just have to define your Agent as 
```python
class SchellingAgent(CellAgent["Schelling"])
```
The next issue was that the ´neighbor.type` was unknown. This was much harder to fix and occupied me for quite a while. There are now two hoops you have to jump through to make it work (if you are determined to enable type checking).
First you have to again add this information to the CellAgent as before
 ```python
class SchellingAgent(CellAgent["Schelling", "SchellingAgent"])
```
**and** you have to define the cell and agent class when you define the OrthogonalGrid like so 
```python
            [height, width],
            torus=True,
            capacity=1,
            random=self.random,
            cell_klass=Cell[SchellingAgent],
            agent_class=SchellingAgent,
        )
```

I think this is not really optimal, but so far the only solution I could find. Maybe someone who knows the Python type system better then me can help improve this PR

### DIFF
--- a/benchmarks/Schelling/schelling.py
+++ b/benchmarks/Schelling/schelling.py
@@ -1,9 +1,10 @@
 from mesa import Model
 from mesa.experimental.cell_space import CellAgent, OrthogonalMooreGrid
+from mesa.experimental.cell_space.cell import Cell
 from mesa.time import RandomActivation
 
 
-class SchellingAgent(CellAgent["Schelling"]):
+class SchellingAgent(CellAgent["Schelling", "SchellingAgent"]):
     """
     Schelling segregation agent
     """
@@ -73,6 +74,8 @@ class Schelling(Model):
             torus=True,
             capacity=1,
             random=self.random,
+            cell_klass=Cell[SchellingAgent],
+            agent_class=SchellingAgent,
         )
 
         self.happy = 0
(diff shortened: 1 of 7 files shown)


EwoutH:
Sorry for the late reply. In general, I like the idea of typing in major Mesa components. I don't know if I would like it to be included in user-facing examples (not this PR though). More importantly, I have to read up on typing in Python, especially Generic Types, TypeVar and the square brackets stuff.

I will try to do that in the upcoming week. How quick do you want to move with this? And what's your vision on typing in Mesa in general?

quaquel:
I guess this PR is largely superseded by the recent work on grid spaces, and so can be closed. @Corvince let me know if there is still something here that should be kept. Otherwise, I'll close this somewhere over the course of next week.

quaquel:
we have made substantial progress on typing so this seems largely fixed by now.

PR #2082, opened 2024-03-14
Add tooltip option to Altair chart

FoFFolo:
I added in the altair.py experimental script the option to see a tooltip on each agent in the canvas when hovering on them. The logic is to see every option described by the agent_portrayal function, except for the x, y, color and size properties.
When defining the type of each tooltip, I didn't manage to find an optimal algorithm to check if a string can be considered a date object, in order to change its type from "nominal" to "temporal", without using an external library called [dateutil](https://pypi.org/project/python-dateutil/), which contains the function parser.parse() that returns a boolean if the string is a right date format.

Here is an example:
```Python
import mesa
import solara
from mesa.experimental import JupyterViz
from money_model.agent import MoneyAgent
from money_model.model import MoneyModel


def agent_portrayal(MoneyAgent):
    size = 10
    color = "tab:red"
    if MoneyAgent.wealth > 0:
        size = 50
        color = "tab:blue"
    if MoneyAgent.unique_id == 0:
        size = 60
        color = "yellow"
    return {"size": size, "color": color, "Unique_id": MoneyAgent.unique_id}


def make_histogram(model):
    # ...

model_params = {
    # ...
}

page = JupyterViz(
    model_class=MoneyModel,
    model_params=model_params,
    measures=[make_histogram],
    name="Money Model",
    space_drawer="altair"
)
# This is required to render the visualization in the Jupyter notebook
page
```

In the agent_portrayal function, I set that every agent will have a tooltip to see their unique_id.
If we run this script using solara:

image

If I click on "View Source" in the altair chart, the content is:
```
{
  "config": {"view": {"continuousWidth": 300, "continuousHeight": 300}},
  "data": {
    "values": [
      {"size": 50, "color": "tab:blue", "Unique_id": 3, "x": 0, "y": 1},
      {"size": 50, "color": "tab:blue", "Unique_id": 48, "x": 0, "y": 3},
      {"size": 50, "color": "tab:blue", "Unique_id": 26, "x": 0, "y": 5},
      {"size": 50, "color": "tab:blue", "Unique_id": 12, "x": 1, "y": 0},
     // ...
      {"size": 50, "color": "tab:blue", "Unique_id": 27, "x": 9, "y": 8}
    ]
  },
  "mark": {"type": "point", "filled": true},
  "encoding": {
    "color": {"field": "color", "type": "nominal"},
    "size": {"field": "size", "type": "quantitative"},
    "tooltip": [{"field": "Unique_id", "type": "quantitative"}],
    "x": {"axis": null, "field": "x", "type": "ordinal"},
    "y": {"axis": null, "field": "y", "type": "ordinal"}
  },
  "height": 280,
  "width": 280,
  "$schema": "https://vega.github.io/schema/vega-lite/v5.16.3.json"
}
```
What's new is the tooltip array element and the unique_id value on each agent.

### DIFF
--- a/mesa/experimental/components/altair.py
+++ b/mesa/experimental/components/altair.py
@@ -34,7 +34,6 @@ def _draw_grid(space, agent_portrayal):
                 all_agent_data.append(agent_data)
         return all_agent_data
 
-
     all_agent_data = portray(space)
     invalid_tooltips = ["color", "size", "x", "y"]
 
@@ -44,9 +43,12 @@ def _draw_grid(space, agent_portrayal):
         # no y-axis label
         "y": alt.Y("y", axis=None, type="ordinal"),
         "tooltip": [
-            alt.Tooltip(key, type=alt.utils.infer_vegalite_type([all_agent_data[0][key]])) 
-                for key in all_agent_data[0] if key not in invalid_tooltips
-        ]
+            alt.Tooltip(
+                key, type=alt.utils.infer_vegalite_type([all_agent_data[0][key]])
+            )
+            for key in all_agent_data[0]
+            if key not in invalid_tooltips
+        ],
     }
     has_color = "color" in all_agent_data[0]
     if has_color:
--- a/mesa/experimental/components/altair.py
+++ b/mesa/experimental/components/altair.py
@@ -43,10 +43,8 @@ def _draw_grid(space, agent_portrayal):
         # no y-axis label
         "y": alt.Y("y", axis=None, type="ordinal"),
         "tooltip": [
-            alt.Tooltip(
-                key, type=alt.utils.infer_vegalite_type([all_agent_data[0][key]])
-            )
-            for key in all_agent_data[0]
+            alt.Tooltip(key, type=alt.utils.infer_vegalite_type([value]))
+            for key, value in all_agent_data[0].items()
             if key not in invalid_tooltips
         ],
     }
--- a/mesa/experimental/components/altair.py
+++ b/mesa/experimental/components/altair.py
@@ -1,6 +1,6 @@
 import contextlib
-from typing import Optional
 import datetime
+from typing import Optional
 
 import solara
 
@@ -38,9 +38,9 @@ def _draw_grid(space, agent_portrayal):
     def detect_type(key):
         key_type = type(all_agent_data[0][key])
         tooltip_type = ""
-        if (key_type == int):
+        if key_type == int:
             tooltip_type = "quantitative"
-        elif (key_type == datetime.datetime):
+        elif key_type == datetime.datetime:
             tooltip_type = "temporal"
         else:
             tooltip_type = "nominal"
@@ -49,7 +49,6 @@ def _draw_grid(space, agent_portrayal):
             # change tooltip_type to "temporal"
 
         return tooltip_type
-
 
     all_agent_data = portray(space)
     invalid_tooltips = ["color", "size", "x", "y"]
@@ -65,8 +64,10 @@ def _draw_grid(space, agent_portrayal):
         # no y-axis label
         "y": alt.Y("y", axis=None, type="ordinal"),
         "tooltip": [
-            alt.Tooltip(key, type=tooltip_types[key]) for key in all_agent_data[0].keys() if key not in invalid_tooltips
-        ]
+            alt.Tooltip(key, type=tooltip_types[key])
+            for key in all_agent_data[0].keys()
+            if key not in invalid_tooltips
+        ],
     }
     has_color = "color" in all_agent_data[0]
     if has_color:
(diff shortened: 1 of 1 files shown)


mesa/experimental/components/altair.py
```diff
@@ -34,12 +35,39 @@ def portray(g):
                 all_agent_data.append(agent_data)
         return all_agent_data
 
+    def detect_type(key):
+        key_type = type(all_agent_data[0][key])
+        tooltip_type = ""
+        if key_type == int:
```
rht, line 41: For checking type, it is recommended with `isinstance`, see https://stackoverflow.com/questions/152580/whats-the-canonical-way-to-check-for-type-in-python:
```
first_by_key = all_agent_data[0][key]
if isinstance(first_by_key, int):
    ...
elif isinstance(first_by_key, ...):
    ...
```

mesa/experimental/components/altair.py
```diff
@@ -34,12 +35,39 @@ def portray(g):
                 all_agent_data.append(agent_data)
         return all_agent_data
 
+    def detect_type(key):
```
rht, line 38: The naming is clearer with `detect_tooltip_type`.

mesa/experimental/components/altair.py
```diff
@@ -34,12 +35,39 @@ def portray(g):
...
+            tooltip_types[key] = detect_type(key)
+
     encoding_dict = {
         # no x-axis label
         "x": alt.X("x", axis=None, type="ordinal"),
         # no y-axis label
         "y": alt.Y("y", axis=None, type="ordinal"),
+        "tooltip": [
+            alt.Tooltip(key, type=tooltip_types[key])
+            for key in all_agent_data[0].keys()
```
rht, line 68: The `[key for key in all_agent_data[0].keys() if key not in invalid_tooltips]` here is a repetition from L57-59. They can precomputed once so that there is no repetition.

rht:
There are Ruff errors that can't be autofixed in the pre-commit.ci.

FoFFolo:
I started by changing type checks using `isinstance`, but then I discovered that there is a function inside the altair library that
does the job.
I replaced my implementations with an integrated function on Altair, called 
`infer_vegalite_type()` which detect automatically the type for the tooltip,
which uses the [infer_dtype()](https://pandas.pydata.org/docs/reference/api/pandas.api.types.infer_dtype.html) pandas function.
The altair function can be found in the altair library folder, in the path "altair/utils/core.py".
```
def infer_vegalite_type(
    data: object,
) -> Union[InferredVegaLiteType, Tuple[InferredVegaLiteType, list]]:
    """
    From an array-like input, infer the correct vega typecode
    ('ordinal', 'nominal', 'quantitative', or 'temporal')

    Parameters
    ----------
    data: object
    """
    typ = infer_dtype(data, skipna=False)

    if typ in [
        "floating",
        "mixed-integer-float",
        "integer",
        "mixed-integer",
        "complex",
    ]:
        return "quantitative"
    elif typ == "categorical" and hasattr(data, "cat") and data.cat.ordered:
        return ("ordinal", data.cat.categories.tolist())
    elif typ in ["string", "bytes", "categorical", "boolean", "mixed", "unicode"]:
        return "nominal"
    elif typ in [
        "datetime",
        "datetime64",
        "timedelta",
        "timedelta64",
        "date",
        "time",
        "period",
    ]:
        return "temporal"
    else:
        warnings.warn(
            "I don't know how to infer vegalite type from '{}'.  "
            "Defaulting to nominal.".format(typ),
            stacklevel=1,
        )
        return "nominal"
```
The altair proper function doesn't implement date object detection on strings, so the user has to specify the date object in the agent_portrayal function. E.g:
```
from datetime import datetime

def agent_portrayal(MoneyAgent):
    # ....
    return {"Date": datetime.now()}
```

mesa/experimental/components/altair.py
```diff
@@ -35,11 +35,20 @@ def portray(g):
...
     encoding_dict = {
         # no x-axis label
         "x": alt.X("x", axis=None, type="ordinal"),
         # no y-axis label
         "y": alt.Y("y", axis=None, type="ordinal"),
+        "tooltip": [
+            alt.Tooltip(
+                key, type=alt.utils.infer_vegalite_type([all_agent_data[0][key]])
+            )
+            for key in all_agent_data[0]
```
rht, line 49: One last simplification, you can do
```python
[
    alt.Tooltip(key, type=alst.utils.infer_vegalite_type([value])
    for key, value in all_agent_data[0].items()
    if key not in invalid_tooltips
]
```

rht:
> I started by changing type checks using isinstance, but then I discovered that there is a function inside the altair library that
does the job.

I see, that simplifies the implementation a lot.

mesa/experimental/components/altair.py
FoFFolo: Okay done. Yeah I agree with you.
> One last simplification, you can do

rht (approved):

rht:
Merged, thank you @FoFFolo.

Issue #1522, opened 2022-11-11
Grid: Calling `place_agent` twice in different positions causes an agent to exist in two positions in the grid


rht:
**Describe the bug**

Context: https://github.com/projectmesa/mesa/pull/1508#discussion_r1014769506
Currently, calling `place_agent` twice or more in a row without calling `remove_agent` beforehand causes the agent to exist in several locations.

**Expected behavior**

An error should be raised if an agent's `pos` attribute is not `None`, when calling `place_agent`.
A stronger guarantee would be to check the entire grid for the existence of the agent, but this is very costly to do, unless there is a `dict` attribute for the grid to track the agents existence.

**To Reproduce**

```python
import mesa

a = mesa.Agent(0, None)
grid = mesa.space.SingleGrid(3, 3, False)
grid.place_agent(a, (1, 1))
print("a.pos", a.pos)
grid.place_agent(a, (0, 0))
print("a.pos", a.pos)
for e in grid.coord_iter():
    print(e)
```

Output
```console
a.pos (1, 1)
a.pos (0, 0)
(<mesa.agent.Agent object at 0x7ffa9311a9e0>, 0, 0)
(None, 0, 1)
(None, 0, 2)
(None, 1, 0)
(<mesa.agent.Agent object at 0x7ffa9311a9e0>, 1, 1)
(None, 1, 2)
(None, 2, 0)
(None, 2, 1)
(None, 2, 2)
```

Tortar:
I think this is really linked with discussion #1513 because if we had a dict inside grid it would be more difficult for the user to change it (we can create a method which returns the agent on the position requested so that the user has no access to the dict easily unless he goes to read the source code)

Corvince:
I wouldn't consider this a bug at all. In Agent-Based Modelling we should not assume too much about what an agent is and what a grid represents, it could very well be an abstract space, where an agent can exist in two places at the same time just fine.

But even in a concrete space setting it should be possible. Imagine building a maze in mesa. There is no reason why one should not have a single "Wall" agent that exists on multiple cells. 

What could be nice is a warning that is displayed when an agent already exists. This would help catch some bugs where agents really shouldn't exist in more than one place, but would still allow to do so if needed by the model.

rht:
The wall/maze use case makes sense. But there would still be warning messages. Either these are suppressed with filter warning, or there is a flag parameter in the `Grid` initialization. I'm not sure if I would classify it as something along the line of NaNs in Numpy.

[...]

PR #2083, opened 2024-03-17
warn if placing already placed agent

puer-robustus:
While it might be desired in a specific model to have the same agent be placed in multiple spots simultaneously, the typical use case is that one agent has one position at every given moment. This commit adapts the place_agent() method in a way that it emits a warning when called with an agent which already has a location.

Fixes: #1522

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -71,6 +71,21 @@ def is_integer(x: Real) -> bool:
     return isinstance(x, _types_integer)
 
 
+def warn_if_agent_has_position_already(placement_func):
+    def wrapper(self, agent, *args, **kwargs):
+        if agent.pos is not None:
+            warnings.warn(
+                f"""Agent {agent.unique_id} is being placed with
+place_agent() despite already having the position {agent.pos}. In most
+cases, you'd want to clear the current position with remove_agent()
+before placing the agent again.""",
+                stacklevel=2,
+            )
+        placement_func(self, agent, *args, **kwargs)
+
+    return wrapper
+
+
 class _Grid:
     """Base class for a rectangular grid.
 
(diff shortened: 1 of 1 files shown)


quaquel:
> I could not find an existing logging configuration in the project. Did I miss it somehow?

There currently is no logging configuration. It is very much another item on the todo list.

rht:
I tested and it works:
```
WARNING:You're placing agent 2699 which already has
the position (9, 1). In most cases, you'd want to clear the current
position with remove_agent() before placing the agent again.
```
Though the warning message should have been made clearer to say that it is from a `place_agent` call, otherwise I wouldn't know the source line that causes the warning.

EwoutH:
Thanks for this PR, it would be a nice feature to have.

For everyone, one thing I'm not sure about if we should make this warning optional or not.

EwoutH:
> Maybe it is preferable if I stick with `warnings.warn()` for the time being?

Yes, I think that would be fine and simplify this PR.

> Only adapt the warning message that the root cause is a `place_agent()` call

I think this is good enough.

rht (approved):

rht:
> So, warnings.warn() with the stacklevel=2 parameter actually gives a bit of an indication on the problematic call:

Yes, that's sufficiently informative.

---

Meta: I support your GitHub protest. It's just that there is no bridge to other ethical FOSS tools yet to ensure public record of discussions and to remain inclusive to people who still use GitHub to contribute to Mesa. Maybe there will be such bridge (along the line of the Matrix bridges that have been built with Slack, Discord, etc).

EwoutH:
Hmmm we took quite a big hit on initialization time here. Maybe we should reconsider this approach.

rht:
I fixed the benchmarks performance problem: https://github.com/projectmesa/mesa/pull/2086#issuecomment-2016965795

Issue #1895, opened 2023-12-01
Add Voronoi tessellation space


EwoutH:
Voronoi Tessellations can be useful in ABM scenarios where the concept of influence or control based on proximity is important. For instance, in urban planning, they can be employed to model the service areas of facilities like hospitals or fire stations, where each facility's influence is determined by the closest access for the population. In ecological studies, Voronoi diagrams can simulate animal territories, where each animal's territory is defined by the space closest to it relative to others, aiding in understanding spatial behaviors and interactions. In telecommunications, they can optimize cell tower placements by delineating areas of signal coverage, ensuring efficient network service with minimal overlap. These tessellations are also beneficial in retail and market analysis for determining the catchment areas of stores, helping businesses understand their direct market influence based on customer proximity.

We might want to implement two subtypes:
- [ ] A discrete variant, directly based on the new cell space. Agents are at a cell, and the Vonoroi space is utilized to determine cell area and neighbors.
      PR: #2084 
- [ ] A variant in which the cells are discrete, but agents move in continuous space. In here agents can be anywhere and check in which cell they are (basically what’s the closest centroid). This can be useful for clustering and geographical borders for example.

![Euclidean_Voronoi_diagram](https://github.com/projectmesa/mesa/assets/15776622/5edef57e-3a2b-40d9-a7fd-0b0d2d63d9d0)

EwoutH:
This might now be able to be done elegantly in the experimental cell space:
- #1994

vitorfrois:
I would like to implement this feature, but some questions emerged when reading this and #1994. 
- What should be considered centroid of the cell? The average agent position?
- If not, should I consider only relative distances to create the tesselation?
- Could you give more details about visualization? It is possible to use Scipy?

EwoutH:
> * What should be considered centroid of the cell? The average agent position?

In my view, that's fixed and user definable. So on creation you input a list / datastructure of centroid. Optionally it could be a `CellAgent`.

It would be interesting to have a dynamic Vonoroi space, like with (groups of) agents that represent a cell. But maybe do the fixed one first.

Haven't thought about visualisation yet, that would be something to investigate. Could be in an separate PR.

vitorfrois:
> So on creation you input a list / datastructure of centroid.
This way, the tessellation is defined only by the listed centroids and have nothing to do with Moore/Neumann grid neighborhoods?

EwoutH:
[...]

PR #2084, opened 2024-03-21
Voronoi Tesselation based Discrete Space

vitorfrois:
## First version of Voronoi Tesselation based Discrete Space, 
described on issue #1895 . This feature allows the user to build a discrete space based on a random sample of points, where neighbors are defined by Delaunay Triangulation.

More specifically, Delaunay Triangulation is a dual-graph representation of the Voronoi Tesselation. Using this algorithm, we can easily find nearest neighbors without delimiting cells edges.
image

The library chosen for the triangulation algorithm implementation was [PyHull](https://github.com/materialsvirtuallab/pyhull), a wrapper of qhull C library, which deals with spatial calculations. The choice was made considering performance, size and usability (SciPy has a similar module but much heavier).   

Based on my discussion with @EwoutH on the issue thread, i thought it would be useful to inherit DiscreteSpace class. 
``` python
class VoronoiGrid(DiscreteSpace):
```

### Example
``` python
from mesa import Model
from mesa.experimental.cell_space import VoronoiGrid, CellAgent

class MyAgent(CellAgent):
```
Users can input cells centroids
``` python
points = [
    [0, 0], [0, 1], [0, 2], 
    [1, 0], [1, 1], [1, 2],
    [2, 0], [2, 1], [2, 2]]
grid = VoronoiGrid(centroids_coordinates=points, capacity=1, random=model.random)
```
and cells neighborhoods are defined by Delaunay triangulation

``` python
agent = MyAgent(1, model)
cell = grid.select_random_empty_cell()
agent.move_to(cell)
print(cell)
print(cell.neighborhood())
```
```
Cell([1, 0], [<__main__.MyAgent object at 0x7f2177268450>])
CellCollection({Cell([0, 1], []): [], Cell([1, 1], []): [], Cell([0, 0], []): [], Cell([2, 1], []): [], Cell([2, 0], []): []})
```
### Next Steps
Definitely, next steps involve computing the boundaries of each cell using Voronoi Tesselation. This allow us to define capacity of each cell based on its area/volume.


EwoutH:
Thanks, it's a great start! Using DelaunayTri is a smart approach.

I guess `move_to` currently move an agent to the cells centroid? While that is logical behavior, I would also like to see an method in which agents are moved to A) the closest point in the cell from their current position and B) a random spot in the cell. This could be a separate PR.

Another thing that it's needed in the long term is a method to check in which cell a point (and thus an agent) is.

@wang-boyu I would also appreciate a review from your mesa-geo expertise.

For the currently functionality, I would like some additional docstring and unittests. Then you can decide if you want to add further functionality to this PR, or merge this first and follow up with new PRs.

mesa/experimental/cell_space/voronoi.py
```diff
@@ -0,0 +1,58 @@
...
+from pyhull.delaunay import DelaunayTri
+
+from mesa.experimental.cell_space.cell import Cell
+from mesa.experimental.cell_space.discrete_space import DiscreteSpace
+
+
+class VoronoiGrid(DiscreteSpace):
+    def __init__(
+        self,
+        centroids_coordinates: Sequence[Sequence[float]],
```
EwoutH, line 15: Would this work with Shapely [Points](https://shapely.readthedocs.io/en/stable/reference/shapely.Point.html)? That might be very useful for Mesa geo applications (cc @wang-boyu).

mesa/experimental/cell_space/voronoi.py
```diff
@@ -0,0 +1,58 @@
...
+from typing import Optional
+
+from pyhull.delaunay import DelaunayTri
+
+from mesa.experimental.cell_space.cell import Cell
+from mesa.experimental.cell_space.discrete_space import DiscreteSpace
+
+
+class VoronoiGrid(DiscreteSpace):
+    def __init__(
```
EwoutH, line 13: I think it's important to to initialize the super class here.

mesa/experimental/cell_space/voronoi.py
```diff
@@ -0,0 +1,58 @@
+from collections.abc import Sequence
+from itertools import combinations
+from random import Random
+from typing import Optional
+
+from pyhull.delaunay import DelaunayTri
```
EwoutH, line 6: I'm a bit worried about adding this package as a dependency. It's last update was in 2015.

Maybe we can use SciPy?
- https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.Delaunay.html
- https://docs.scipy.org/doc/scipy/tutorial/spatial.html#delaunay-triangulations

Otherwise maybe `shapely.ops.triangulate` or keeping the code in our own repo.

vitorfrois: Hmmm, I choose PyHull because Mesa is not using Scipy already and it is a very heavy dependency. I'll take a look at Shapely, it looks a good alternative, and as you said, can be integrated with Mesa Geo

EwoutH: Sounds reasonable!

wang-boyu:
Thanks for the PR. I have to say that I'm still a bit confused about why this is based on discrete space, not continuous space (as I mentioned in https://github.com/projectmesa/mesa/issues/1895#issuecomment-2016514829 and https://github.com/projectmesa/mesa/issues/1895#issuecomment-2016521840).

As a concrete example, consider this simple space:

| Cell 1 | Cell 2 | Cell 3 |
|:---:|:---:|:---:|
| O    | ?   | X  |

If it is a Voronoi space, should Cell 2 be split into halves, where left half belongs to agent of type O and right half belongs to agent of type X? How does this work in discrete space?

tpike3:
Apologies, a little late to this discussion, this is really cool @vitorfrois. 

As dependencies and keeping Mesa lightweight is a constant struggle, but form a very superficial look is there a reason we wont use networkx  (https://networkx.org/documentation/stable/reference/algorithms/voronoi.html) and this has a corresponding example that may work for mesa_geo (https://networkx.org/documentation/stable/auto_examples/geospatial/plot_delaunay.html)? 

Let me know what you think/ what I am missing?

vitorfrois:
@wang-boyu 
> I have to say that I'm still a bit confused about why this is based on discrete space, not continuous space. How does this work in discrete space?

I'm considering the discrete space for Voronoi Tesselation as a **static/non changing graph**, where neighbors are obtained by **Delaunay Triangulation**. As @EwoutH said, it is only a special case for HexGrid where cell are not equally spaced. 

I think its more about a choice of implementation where I considered mainly the Cholera example. As I suggested yesterday, 
> since the result of Delaunay Triangulation is a graph, we can use NetworkGrid to represent it, and in the continuous case, implement in the Mesa Geo using Shapely.

Does that make sense?

wang-boyu:
> I'm considering the discrete space for Voronoi Tesselation as a **static/non changing graph**, where neighbors are obtained by **Delaunay Triangulation**.

Thanks for clarifying. Appreciate it.

Question: Is the **Delaunay Triangulation** done in continuous space?

vitorfrois:
> Question: Is the **Delaunay Triangulation** done in continuous space?

Yes, it is

vitorfrois:
**Updates**
- Implemented Voronoi from scratch (basically copied a simple already implemented algorithm)
- Now using the regions resulting from Delaunay/Voronoi Triangulation to draw Voronoi cells in JupyterViz
- Altough, could not cover `visualization/components/matplotlib.py` with tests :(

pyproject.toml
```diff
@@ -43,6 +43,7 @@ dependencies = [
   "pandas",
   "solara",
   "tqdm",
+  "pyhull",
```
EwoutH, line 46: This can now be removed right?

mesa/experimental/cell_space/voronoi.py
```diff
@@ -0,0 +1,264 @@
...
+from itertools import combinations
+from random import Random
+
+import numpy as np
+
+from mesa.experimental.cell_space.cell import Cell
+from mesa.experimental.cell_space.discrete_space import DiscreteSpace
+
+
+class Delaunay:
```
EwoutH, line 11: Should we put this in a separate file? @rht?

EwoutH (approved):
This is really outstanding, Vítor. That space visualization especially, it really offers a lot of value.

Small minor comments. I would also like to see more test coverage, if you have the chance.

I’m going to approve conceptually, especially considering the cell space is still experimental. @rht are you able to do a technical code review and merge if it looks good?

CC @quaquel, you might also find this interesting (and have some comments).

mesa/experimental/cell_space/voronoi.py
rht: In voronoi.py should be fine. We can split it once it is used in more than one context. The docstring should clearly indicate that this is a helper class, and not a space implementation it itself. I'd rename it to `_Delaunay` for now so that we can merge this PR sooner. We can discuss about other use cases later.

mesa/experimental/cell_space/voronoi.py
```diff
@@ -0,0 +1,264 @@
...
+    """
+    Class to compute a Delaunay triangulation in 2D
+    ref: http://github.com/jmespadero/pyDelaunay2D
```
rht, line 14: How much percentage of the code is based on this GitHub repo? Their license is GPLv3, and we actually can't copy their code because there will be a licensing issue.

vitorfrois:
Omg, sorry I did not notice the license. I copied like 70% of it's
implementation, changing variables, function names and other minor things.

rht:
There are 2 ways to proceed:
1. Use ChatGPT to do code laundry, i.e. ask it to do Delaunay triangulation code. It probably is going to be based on that GPLv3 repo, but you can blame ClosedAI for that instead.
2. Use SciPy, but have SciPy to be an optional dependency, only imported when the user needs `VoronoiGrid`

I'm more in favor of option 2, because option 1 is rather shady.

EwoutH:
Yeah the license issue is a big one. Another (far fetched) option is to contact the maintainer to see if he’s willing to publish on another, more permissive license.

EwoutH:
@vitorfrois I would like to try to contact the maintainers. Do you want to write them a message or would you like me to do it?

EwoutH:
I can do it tomorrow, then hopefully we can move this forward next week!

pyproject.toml
EwoutH: @vitorfrois right?

vitorfrois: Right

EwoutH (approved):
You know what, I think this is fine. It's just math in a specific order.

Also, this is in the experimental space. We can break or remove anything from here, and iterate fast on it.

Let's get this moving. If you remove the dependency we're good to go, and I will merge.

**Edit:** There are a few minor open comments. Would be nice to have those resolved.

EwoutH:
Merged! Congratulations @vitorfrois your first PR is in Mesa - and it's a remarkable one.

Thanks for your effort and patience.

It will be included in the Mesa `3.0.0a3` release later today.

PR #2086, opened 2024-03-24
test: Remove place_agent duplicate warnings

rht:
Because of #2083, some faulty test code have been caught. cc: @puer-robustus

### DIFF
--- a/benchmarks/Flocking/flocking.py
+++ b/benchmarks/Flocking/flocking.py
@@ -29,7 +29,6 @@ class Boid(mesa.Agent):
         self,
         unique_id,
         model,
-        pos,
         speed,
         direction,
         vision,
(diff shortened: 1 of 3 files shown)


EwoutH:
Can you split of the ruff update in a separate PR? I like to keep PRs atomic.

rht:
Done.

EwoutH:
Thanks! I guess it isn’t allowed in current Mesa to initialize agents with a `pos` right? I have seen it in other models, maybe we should write that up somewhere.

rht:
> Thanks! I guess it isn’t allowed in current Mesa to initialize agents with a pos right?

Allowed but not recommended.

> Also, which warning is this removing exactly?

#2083

EwoutH (approved):

PR #2087, opened 2024-03-24
docs: improve introductory tutorial

puer-robustus:
This fixes

- some minor grammatical errors,
- an incorrect header indentation level,

and also starts using the standard markdown footnote syntax. Since the latter is a bit of a bigger change (footnotes go to the end of the page), I put it in a separate commit. We could also consider using footnotes for the two scientific quotations, if so desired.

### DIFF
--- a/docs/tutorials/intro_tutorial.ipynb
+++ b/docs/tutorials/intro_tutorial.ipynb
@@ -136,7 +136,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Import Dependencies\n",
+    "### Import Dependencies\n",
     "This includes importing of dependencies needed for the tutorial."
    ]
   },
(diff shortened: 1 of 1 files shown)


EwoutH (approved):
Thanks, good catches!

Can this PR be merged or do you want to make other changes?

EwoutH:
Is this working as expected?

image

References as footnotes that refer to the full source at the bottom of the page sounds good.

EwoutH:
Yes, that's nice. Can we create a solution that also looks nice in the Jupyter notebook?

EwoutH:
Maybe in that case I prefer a direct link to the papers (just `[text](url)`).

EwoutH:
Both ways are fine I think, I slightly prefer the readability of the notebooks over the docs, since they are mainly interactive tutorials.

EwoutH (approved):
Looks good, thanks!

PR #2090, opened 2024-03-27
Set JupyterViz as stable

rht:


EwoutH:
Does this break existing models?

rht:
This is an API breaking change because of the first commit.

EwoutH:
Good to know! Should we do it in two stages:
 - For 2.3, just add a warning we're going to (re)move it
 - For 3.0, actually (re)move it

I might also be a proponent of just removing it in 3.0. You can always keep using the older versions if you want or need to. That keeps the codebase clean and lean.

tpike3:
To be clear on the process: 
1 - Separate PR that has a deprecation warning. 
2 - Merge this in 3.0

I would agree that we want to remove the old visualization code from the codebase entirely. 
 
But will add, don't we want push people to [mesa-viz-tornado](https://github.com/projectmesa/mesa-viz-tornado/)? Interestingly other people updated it in February - [update](https://github.com/projectmesa/mesa-viz-tornado/commit/4001971a3d744e33e15da420d763769c2c974144)

As an aside always thinking about the ABM and Mesa ecosystem this add some intriguing questions on package management.

EwoutH:
2.3 is released, so we can start with 3.0 development work, including moving this PR forward!

Deprecation warning for 2.3.x would still be nice though.

rht:
Should I remove the Tornado viz altogether? This will free up the default dependencies and documentation.

EwoutH:
> Should I remove the Tornado viz altogether?

I'm good with that. Maybe we need to write a (short) migration guide for 3.0, and include a deprecation warning linking to it when trying to import the old tornado viz.

rht:
I already wrote on Matrix several days ago
> can't put warning on 2.3.0 because the old viz is a direct module import of mesa-viz-tornado classes
> you can add to the release note instead

rht:
I'm merging now because if merge conflicts happen again, this is hard to redo.

EwoutH:
Agreed! Let’s continue to test and develop this, we have quite some time before 3.0 stable.

PR #2092, opened 2024-03-28
add `group_by` method to `AgentSet` for attribute-based grouping

ai-naymul:
Discussion(https://github.com/projectmesa/mesa/discussions/2074)

#### In this pr I have added a group_by method in the AgentSet class, allowing agents to grouped based on some specific attributes.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -345,6 +345,22 @@ class AgentSet(MutableSet, Sequence):
         self.model = state["model"]
         self._update(state["agents"])
 
+    def group_by(self, attr_name: str) -> dict:
+        """
+        Group agents in the AgentSet based on the specified attribute.
+
+        Args:
+            attr_name (str): The name of the attribute to group agents by.
+
+        Returns:
+            dict: A dictionary where keys are attribute values and values are lists of agents with those attribute values.
+        """
+        grouped_agents = defaultdict(list)
+        for agent in self:
+            attr_value = getattr(agent, attr_name, None)
+            grouped_agents[attr_value].append(agent)
+        return grouped_agents
+    
     @property
     def random(self) -> Random:
         """
--- a/mesa/experimental/components/altair.py
+++ b/mesa/experimental/components/altair.py
@@ -56,5 +56,10 @@ def _draw_grid(space, agent_portrayal):
         .properties(width=280, height=280)
         # .configure_view(strokeOpacity=0)  # hide grid/chart lines
     )
+    # This is the default value for the marker size, which auto-scales
+    # according to the grid area.
+    if not has_size:
+        length = min(space.width, space.height)
+        chart = chart.mark_point(size=30000 / length**2, filled=True)
 
     return chart
--- a/tests/test_agent.py
+++ b/tests/test_agent.py
@@ -282,3 +282,16 @@ def test_agentset_shuffle():
     agentset = AgentSet(test_agents, model=model)
     agentset.shuffle(inplace=True)
     assert not all(a1 == a2 for a1, a2 in zip(test_agents, agentset))
+
+def test_agentset_group_by():
+    model = Model()
+    agents = [TestAgent(model.next_id(), model) for _ in range(10)]
+    for i, agent in enumerate(agents):
+        agent.category = i % 2  # Assign categories 0 or 1
+    agentset = AgentSet(agents, model)
+
+    grouped = agentset.group_by("category")
+    assert len(grouped[0]) == 5
+    assert len(grouped[1]) == 5
+    assert all(agent.category == 0 for agent in grouped[0])
+    assert all(agent.category == 1 for agent in grouped[1])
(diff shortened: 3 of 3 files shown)


EwoutH:
Thanks for this PR! It looks like it could fill a nice use case, especially for attributes with discrete values.

On thing I'm thinking about is if we can do something smart to select ranges or thresholds for attributes with continuous values. For example divide agents in to bins based on some attribute.

Curious what the performance of this is compared to (multiple) `select()` cases.

ai-naymul:
> I also had a quick look. I like the direction. Some quick points:
> 
> 1. how to handle non-discrete attributes? The current code is not explicit on this. The obvious direction is to have some kind of binning. Alternatively, there could be some check of the type of the value of the attribute. If this is, for example, a float, an exception or warning might be raised. Either way, this issue need to be handled explicitly.
> 2. I would like to see unit tests for this code to ensure we maintain proper coverage.

Addressing non-discrete attributes through binning or type checks is a valid concern. I propose adding an optional `binning_function` parameter to the method to handle continuous attributes. This function can categorize attributes into discrete bins. If not provided for continuous attributes, we'll issue a warning. I'll also add the necessary unit tests.

rht:
I tried pandas' `groupby` on float values:
```ipython
In [4]: data = {"a": np.random.random(4), "b": np.random.random(4)}

In [5]: df = pd.DataFrame(data=data)

In [6]: df
Out[6]:
          a         b
0  0.437550  0.389805
1  0.180405  0.032264
2  0.509254  0.147376
3  0.181636  0.102611
In [9]: for k, v in df.groupby("b"):
   ...:     print(df.groupby("b").get_group(k))
   ...:
          a         b
1  0.180405  0.032264
          a         b
3  0.181636  0.102611
          a         b
2  0.509254  0.147376
         a         b
0  0.43755  0.389805
```
i.e. they seem to just let you do it even if the group has 1 element each.

quaquel:
> I tried pandas' groupby on float values:

I like to get other people's perspective on this as well. Personally, I am in favor of mimicking pandas as much as possible, so just group on the float as normal. This keeps the code nice and simple. If you want to do binning, you can already easily code this up through `AgentSet.do` and a callable. So

```python
my_agents.do(some_binning_function).group_by("my_newly assigned attribute")

```

mesa/agent.py
```diff
@@ -345,6 +345,22 @@ def __setstate__(self, state):
...
+        """
+        Group agents in the AgentSet based on the specified attribute.
+
+        Args:
+            attr_name (str): The name of the attribute to group agents by.
+
+        Returns:
+            dict: A dictionary where keys are attribute values and values are lists of agents with those attribute values.
+        """
+        grouped_agents = defaultdict(list)
```
quaquel, line 358: Should this not be a `defaultdict(AgentSet)`? Similar to the binning discussion, this would mimic pandas in that groupby returns a group identifier and an AgentSet with the agents in that group.

quaquel:
Thanks for this! Just one thought on the binning conversation and a question regarding the appropriate return type. The code itself, and the tests look fine.

ai-naymul:
> Thanks for this! Just one thought on the binning conversation and a question regarding the appropriate return type. The code itself, and the tests look fine.

Hi sorry for the late reply.

I think we don't need to change the It `group_by` method to handle the binning specifically, it will handle continuous attributes directly without enforcing binning.

Here is example usage for binning and grouping :

`
def binning_function(agent):`
    `agent.binned_attribute = agent.continuous_attribute // 10 * 10`

`my_agents.do(binning_function)`

`grouped_agents = my_agents.group_by("binned_attribute") `

EwoutH:
@ai-naymul is this PR ready for (final) review? If so, can you mark it as such (by pressing the "Ready to review") button?

mesa/agent.py
ai-naymul: You raise a valid point regarding the return type of the group_by method. Changing the return type to defaultdict(AgentSet) would indeed align more closely with pandas' groupby functionality.

`grouped_agents = defaultdict(AgentSet)`

when adding agents to the groups:

`for agent in self:
    attr_value = getattr(agent, attr_name, None)
    if attr_value not in grouped_agents:
        grouped_agents[attr_value] = AgentSet(model=self.model)
    grouped_agents[attr_value].add(agent)`

ai-naymul:
> @ai-naymul is this PR ready for (final) review? If so, can you mark it as such (by pressing the "Ready to review") button?
> 
> Please also update the PR message with a bit of context (including which problem this solves and how it's implemented on a high level).

Hi sorry for the inconvenience, I was having some health issues and wasn't able to work on this properly as well.

I think need some improvement like need the feedback regarding the return type of the group_by method from @quaquel .

I think it's final for review although. I will add some description about this pr as well.

EwoutH:
@ai-naymul are you still interested on working on this? If so I would like to have an explanation of the feature and an example in the PR message. A bit like: #1890 or #1916 (but it can be shorter since it's a smaller feature).

Issue #2117, opened 2024-04-23
move_agent_to_one_of is deterministic when pos has multiple closest  choices and selection="closest"


OrenBochman:
**Describe the bug**

move_agent_to_one_of is deterministic when pos has multiple closest  choices and selection="closest"

**Expected behavior**

expected that when using move_agent_to_one_of() in a sugarscape agent's motion will be random within a sugar plateau for low vision agents, but noticed that such agents now exhibit periodic motion.

**To Reproduce**

use the following test 

```Python
def test_move_agent_closest_selection_multiple(self):
    random_locations = []
    agent = self.agents[0]
    agent.pos = (5, 5)
    repetititions = 10

    for _ in range(repetititions):
        possible_positions = [(4, 4), (6, 6), (10, 10), (20, 20)]
        self.space.move_agent_to_one_of(
            agent, possible_positions, selection="closest"
        )
        random_locations.append(agent.pos)
        assert agent.pos in possible_positions
        self.space.move_agent_to_one_of(agent, [(5, 5)], selection="closest")
    non_random_locations = [random_locations[0]] * repetititions
    assert random_locations != non_random_locations
```

**Additional context**

PR #2118, opened 2024-04-23
Fix deterministic behavior in `move_agent_to_one_of` with `selection="closest"`

OrenBochman:
This PR addresses a bug in `move_agent_to_one_of` where agents exhibit deterministic behavior when multiple closest positions are available and `selection="closest"` is used. This issue caused agents to follow periodic motion patterns instead of random selection within sugar plateaus, particularly noticeable in low-vision agents.

**Changes**:
- Modified `move_agent_to_one_of` to ensure randomness when multiple equally distant positions are found by shuffling the positions and randomly selecting from the closest ones.
- Added a new test, `test_move_agent_closest_selection_multiple`, to verify that agents exhibit random behavior in scenarios with multiple closest positions.

**Fixes**: Agents now select randomly from multiple closest positions, preventing deterministic motion.

Closes: https://github.com/projectmesa/mesa/issues/2117


mesa/space.py
```diff
@@ -459,14 +459,19 @@ def move_agent_to_one_of(
             elif selection == "closest":
                 current_pos = agent.pos
                 # Find the closest position without sorting all positions
-                closest_pos = None
+                closest_pos = []
                 min_distance = float("inf")
                 for p in pos:
```
EwoutH, line 464: What if instead of keeping a list, we shuffle the `pos` list here? I think that's both faster and cleaner in implementation.

EwoutH (commented):
Thanks a lot for catching this! You are indeed right, the closest choice is deterministic while it shouldn't be.

I have one suggestion to implement it a bit cleaner.

mesa/space.py
EwoutH, line 464:
```suggestion
                for p in agent.random.shuffle(pos):
```
I think with this single line change we could be done.

OrenBochman: Shuffling the list is a pretty simple solution so I don't mind doing that.

I printed all the pos lists for my simulation and most were pretty big but had a few closest locations. 

It would be faster to shuffle the shorter list like I did. However my idea needs to go over all the points which is slower.

But in terms of complexity the bulk of the work (complexity) is finding the smallest distance. Both ideas are the same. 

Before I found this method I was thinking of adding a distance representation masked layer ndarray so I could let numpy do this via a vectorized op. This would be cached by the agent after each move and reused.

But I did not explore this yet but it might be more efficient to build this for r =1 
check for a match, r=2 check.

Bottom line: I think to find a better answer would require bench-marking. It's probably not that important for this revision. Do as you see fit.

OrenBochman:
Ok I see that the test passes and my example that uses this code is not showing the issue anymore.  SO +1

OrenBochman (commented):
Ok I see that the test passes and my example that uses this code is not showing the issue anymore.  SO +1

tpike3:
@EwoutH Do we want to wait for speed on this one or merge it and do a 2.3.1 to fix the bug?

EwoutH (approved):
Sorry that we did this die in red tape. I'm not sure if our benchmarks use this behavior, but for now correct behavior is more important than performance.

Thanks for the fix!

PR #2119, opened 2024-04-23
Make agent move to actual random closest position

### LINKED ISSUE #2117: move_agent_to_one_of is deterministic when pos has multiple closest  choices and selection="closest"

EwoutH:
This fixes a bug in which `move_agent_to_one_of` is deterministic when pos has multiple closest choices and `selection="closest"`.

Closes #2117.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -461,7 +461,7 @@ class _Grid:
                 # Find the closest position without sorting all positions
                 closest_pos = None
                 min_distance = float("inf")
-                for p in pos:
+                for p in agent.random.shuffle(pos):
                     distance = self._distance_squared(p, current_pos)
                     if distance < min_distance:
                         min_distance = distance


EwoutH:
I don't understand this error:
```Python
        # Only move agent if there are positions given (non-empty list)
        if pos:
            if selection == "random":
                chosen_pos = agent.random.choice(pos)
            elif selection == "closest":
                current_pos = agent.pos
                # Find the closest position without sorting all positions
                closest_pos = None
                min_distance = float("inf")
>               for p in agent.random.shuffle(pos):
E               TypeError: 'NoneType' object is not iterable
```
The `if pos:` should filter out `None` values.

EwoutH:
Right, `random.shuffle` does not return any value; it modifies the list `pos` in-place. Therefore, `agent.random.shuffle(pos)` results in `None`, and trying to iterate over `None` leads to the error you're seeing.

It now works as expected.

rht:
Make sure to delete the branch after the PR has been merged.

PR #2132, opened 2024-05-07
Drop support for Python 3.9, require Python >= 3.10 

EwoutH:
Starting development for Mesa 3.0 and following [SPEC 0](https://scientific-python.org/specs/spec-0000/) we can drop Python 3.9 support and require Python 3.10 or higher for future development. This allows adopting more modern Python features and simplifies the testing and CI configurations.

See the Python 3.10 release notes: https://docs.python.org/3/whatsnew/3.10.html

The 2.3.x release series will keep supporting Python 3.9.

### DIFF
--- a/.github/workflows/build_lint.yml
+++ b/.github/workflows/build_lint.yml
@@ -37,8 +37,6 @@ jobs:
             python-version: "3.11"
           - os: ubuntu
             python-version: "3.10"
-          - os: ubuntu
-            python-version: "3.9"
           # Disabled for now. See https://github.com/projectmesa/mesa/issues/1253
           #- os: ubuntu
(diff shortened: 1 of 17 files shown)


rht:
Duplicate of #2008. Missing update to https://github.com/projectmesa/mesa/pull/2008/files#diff-732a7eb31fbbd25075ee965837dc0092fb53c2cb64c068e60392d9bf69d9f9bb.

EwoutH:
Good catch, added it.

The reason I opened a new PR because the previous one didn't move forward and looked incomplete. This one is a fresh take without conflicts including all the typing changes that can be made automatically.

rht (approved):

PR #2091, opened 2024-03-27
Remove mesa.flat namespace

rht:
This doesn't seem to be used as often as the current simple namespace.

PR #2133, opened 2024-05-08
breaking: Remove visualization_old (mesa-viz-tornado)

rht:
TODO:
- [x] Merge #2091.
- [x] Update cookiecutter template.


rht:
This fixes the tests.

EwoutH (approved):
Personally, I'm going to approve this. Let's cut the cord and focus on the future. The new space needs some work (like support for Cell spaces and PropertyLayers) but it's the future.

I would like more maintainers to pitch in and/or approve (at least a second) before merging though, since this is a big change.

jackiekazil:
To be clear -- Does this do the following? 

- [ ] This fixes the tests mentioned in #2091?
- [ ] Adds deprecation warning?
- [ ] Updates cookie cutter template?

Questions -- 
1. Does #2091 need to be merged first, but this shortly after? 
2. What are the recommended methods for testing this? (I looked it over, but I wan to make sure I am not missing something.)
3. This all is consider 2.3 right, not 3.0?

rht:
> To be clear -- Does this do the following?

It does 2 items except for adding the deprecation warning. The deprecation warning is to be added in the branch https://github.com/projectmesa/mesa/tree/2.3.x-maintenance.

> Does https://github.com/projectmesa/mesa/pull/2091 need to be merged first, but this shortly after?

#2091 needs to be merged first, then I rebase this PR, and then it can be merged.

> What are the recommended methods for testing this? (I looked it over, but I wan to make sure I am not missing something.)

The examples being tested in CI don't have frontend in them. A manual testing would be to test the Solara examples in mesa-examples.

> This all is consider 2.3 right, not 3.0?

This is for 3.0.

rht:
>  A manual testing would be to test the Solara examples in mesa-examples.

Without manual changes, they will fail because they still use the `mesa.experimental` namespace instead of `mesa.visualization`.

wang-boyu:
As I mentioned in our dev meeting, my concern is about what the users need to do for this... So if a user wants to stick to tornado visualization, he/she will not be able to use Mesa's new features? Conversely if a user wants to update Mesa, he/she will need to migrate from tornado to jupyter?

I might be wrong but if this is the case, shall we have some kind of migration guide to assist the users' transition, so that they are not confused or frustrated.

EwoutH:
That would be the idea (and I support it). Old visualisation use Mesa 2.3.x, new visualisation for 3.0.

We can't support everything forever, we have to move forward.

jackiekazil:
@EwoutH, this feels like a clean way to break it. 

@wang-boyu 
RE:
> I might be wrong but if this is the case, shall we have some kind of migration guide to assist the users' transition, so that they are not confused or frustrated. 

We could do a write-up for users and announce it. We need 1 or 2 use cases to create adequate documentation. Maybe we clarify that people can and are welcome to ask for help with conversions?

EwoutH:
I would recommend merging this and then tag a first pre-release (Mesa `3.0.0a0`), so users can start testing with the new visualisation.

EwoutH:
We can use the two examples in the mesa-examples repo:
- https://github.com/projectmesa/mesa-examples/tree/main/examples/boltzmann_wealth_model_experimental
- https://github.com/projectmesa/mesa-examples/tree/main/examples/schelling_experimental

@rht #2091 still need to be merged before this PR?

rht:
I have merged #2091 and rebased this PR so that it only has 1 relevant commit.

docs/mesa.md
```diff
@@ -1,11 +1,5 @@
 # mesa package
 
-## Subpackages
-
-```{toctree}
-mesa.visualization_old
```
EwoutH, line 6: Should we add the new visualisation here? It's nice to have that build in the docs, even if we don't have that much docstring.

rht:
```quote
We can use the two examples in the mesa-examples repo:

https://github.com/projectmesa/mesa-examples/tree/main/examples/boltzmann_wealth_model_experimental
https://github.com/projectmesa/mesa-examples/tree/main/examples/schelling_experimental
```
I think it should be either one of them, and virus on a network or boid flocker, because the 2 examples you listed have some overlap in the usage of space. Virus on a network uses network visualization, and boid flocker uses continuous space.

docs/mesa.md
rht: That can happen in a separate PR. This PR is rather expensive to maintain/rebase.

EwoutH:
Thanks a lot for all the work!

wang-boyu:
> We could do a write-up for users and announce it. We need 1 or 2 use cases to create adequate documentation. Maybe we clarify that people can and are welcome to ask for help with conversions?

Yup for instance NetLogo has transition guide for new releases: https://ccl.northwestern.edu/netlogo/docs/transition.html

PR #1354, opened 2022-06-02
ABM language

GaelLucero:
Per discussion creating additional pull request for ABM improvement as discussed in PR #1278

PR #2149, opened 2024-05-26
Agent spatial methods from GaelLucero

catherinedevlin:
Adapted from https://github.com/projectmesa/mesa/pull/1354

In In-person conversation with @jackiekazil at PyCon, we decided not move the methods to an Agent subclass.  The burden of carrying along sometimes-irrelevant methods seemed smaller than the cognitive load of making end users choose which Agent class to use.

Suggesting several changes to GL's code:

    - Omitted .setxy, .set_pos as unnecessary setters
    - added default arguments 
    - raise exception if model has no attached ContinuousSpace
    - added tests

Removed .die from this PR since it's different in theme, and it deserves a discussion about how necessary it is.

This PR assumes a model with ContinuousSpace, and throws an exception if the methods are called without it.  We might want a follow-up PR that would implement the methods for grids, but we would have to decide whether headings and positions should be limited to those that fit neatly into the grid (only headings of 0, 45, 90, 135... for square grids, for instance), or whether to permit non-aligned headings and positions and then round off positions after each move.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -97,10 +97,10 @@ class Agent:
         except AttributeError as exc:
             raise AttributeError("Agent's model does not define space") from exc
         except Exception as exc:
-            raise UserWarning(
+            warnings.warn(
                 f"agent.py (forward_backwards): could not move agent "
-                f"{self.unique_id} within self.model.space"
-            ) from exc
+                f"{self.unique_id} within self.model.space\n{exc}"
+            )
 
     def move_forward(self, amount=1):
         """Moves the agent forward by the amount given"""


EwoutH:
Thanks a lot for reviving this effort!

> In In-person conversation with @jackiekazil at PyCon, we decided not move the methods to an Agent subclass.

I would love to hear a little bit more about this.

Another question I would have is if this would work with the new experimental Cell Space.

EwoutH:
I see you made some new commits, does everything now work in continuous and discrete space?

EwoutH:
For reference, some further work on this is being done in 
- https://github.com/projectmesa/mesa/pull/2292

quaquel:
Thanks again for this @catherinedevlin, 

Closing this as stale, but linked to it from #2967. I want to pick up these ideas again, but using the new space implementation that was added in January of 2025.

PR #2150, opened 2024-05-26
Agent.die() 

catherinedevlin:
Continued from https://github.com/projectmesa/mesa/pull/1354.  See also https://github.com/projectmesa/mesa/pull/2149 for the remainder of that PR's methods.

I'm less confident that this PR should me merged, since

- It only joins together two method calls that already exist anyway
- It creates a confusing failure mode (if a model is successfully removed from the schedule, but removal from the space fails, should it be added back to the schedule?)
- If it's called for an Agent in a Model without a Space, perhaps the second failure (to remove the model from space) should not be considered a failure at all?  In that case, though, it should really just be a call to self.model.schedule.remove(self)

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -80,12 +80,15 @@ class Agent:
         """Removes the agent from the schedule and the grid """
         try:
             self.model.schedule.remove(self)
-        except:
-            print("agent.py (die): could not remove agent from self.model.schedule")
+        except Exception as exc:
+            warnings.warn(f"agent.py (die): could not remove agent {self.id} "
+                          f"from self.model.schedule.  Not removing from space.\n{exc}")
+            return
         try:
             self.model.space.remove_agent(self)
-        except:
-            print("agent.py (die): could not remove agent from self.model.space")
+        except Exception as exc:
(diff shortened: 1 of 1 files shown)


EwoutH:
Thanks for this PR! A convinient, complete and robust method to remove agents is certainly useful.

I'm only wondering if there aren't redundant elements in this PR. @quaquel if I'm correct, the current weakrefs already do a large part of the heavy lifting, right?

Also, we have to make clear how this differs from `Agent.remove()`:
https://github.com/projectmesa/mesa/blob/6507f510a9f6ec76f26d861896e416e36a427501/mesa/agent.py#L67-L70

And maybe check if it needs to be removed from AgentSets the agent is part of:
https://github.com/projectmesa/mesa/blob/6507f510a9f6ec76f26d861896e416e36a427501/mesa/agent.py#L316

EwoutH:
@catherinedevlin would you like to continue with this PR? If so we can discuss the differences between die and remove.

EwoutH:
Closing for now due to lack of response. If anyone wants to work further on it, feel free to leave a message here!

Issue #2122, opened 2024-04-24
Build docs for experimental modules (including cell space)


EwoutH:
Currently [docs](https://mesa.readthedocs.io/) aren't build for features in [mesa/experimental](https://github.com/projectmesa/mesa/tree/main/mesa/experimental), particularly the [cell_space](https://github.com/projectmesa/mesa/tree/main/mesa/experimental/cell_space) and [devs](https://github.com/projectmesa/mesa/tree/main/mesa/experimental/devs). It would be really usefull to have those docs build in Read the Docs.

It probably requires modifying [docs/conf.py](https://github.com/projectmesa/mesa/blob/main/docs/conf.py).

EwoutH:
Thanks! Our docs are build in CI so just opening a PR should show us if that’s the case.

stephenfmann:
Tried this locally and it seems to work. Pull request is https://github.com/projectmesa/mesa/pull/2154

EwoutH:
Yes, thanks!

PR #2154, opened 2024-06-11
Add experimental features to documentation as per #2122

stephenfmann:
Adding the objects from `experimental.cell_space` and `experimental.devs` to the ReadTheDocs documentation, as per issue #2122.

Tested locally using sphinx and seems to work fine.

### DIFF
--- a/docs/apis/experimental.md
+++ b/docs/apis/experimental.md
@@ -1,0 +1,60 @@
+# Experimental
+
+```{eval-rst}
+.. automodule:: experimental.__init__
+   :members:
+```
+
+## Cell Space
+
+```{eval-rst}
+.. automodule:: experimental.cell_space.__init__
+   :members:
+```
+
+```{eval-rst}
+.. automodule:: experimental.cell_space.cell
+   :members:
+```
+
+```{eval-rst}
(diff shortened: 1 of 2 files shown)


EwoutH (approved):
Thanks! Our CI built looks good! https://mesa--2154.org.readthedocs.build/en/2154/apis/experimental.html

docs/apis/experimental.md
```diff
@@ -0,0 +1,60 @@
+# Experimental
+
+```{eval-rst}
+.. automodule:: experimental.__init__
+   :members:
+```
+
+## Cell Space
```
EwoutH, line 8: Does this whole part need to be hard coded? That means on each change we have to also modify this file right?

stephenfmann: If you create new scripts you need to add them here. At least that's how I've done it in the past and how the other .md files in the directory are doing it e.g. `visualisation.md`. Are you thinking sphinx might have a smarter way of doing it?

EwoutH: Yeah, probably.

That can be done in another PR however, so this is good to go!

PR #2161, opened 2024-07-03
Fix pre-commit issues: Codespels and typecheck

EwoutH:
Fixes two things pre-commit detected:
- Some codespell errors
- A case of E721 Use `is` and `is not` for type comparisons, or `isinstance()` for isinstance checks

After this PR all pre-commit checks are passing again.

### DIFF
--- a/.codespellignore
+++ b/.codespellignore
@@ -6,3 +6,4 @@ inactivate
 ue
 fpr
 falsy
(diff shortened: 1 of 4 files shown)


rht (approved):

EwoutH:
Thanks for merging!

PR #2163, opened 2024-07-03
Jupyter_viz: Allow measures to be None

EwoutH:
Currently `measures` crashed when it's `None` (which is the default). The only way to circumvent this is adding `[]`, but that's non-ideal and not obvious.

This PR fixes that, so that `measures` can actually be the default value of `None`.

This also proves we need to test Jupyter viz in notebooks, not only in console, since both have different code paths.


rht (approved):

EwoutH:
Thanks for merging!

PR #2171, opened 2024-07-06
Visualisation: Allow different shapes agent portrayals#2164 

Aryan1Mahahjan:
Hi, this is Draft Solution to #2164.

I was thinking about using the below, which use the Encode function attribute 

The added code snippet checks for shape and fill keys in agent data. If present, it adds them to the encoding_dict using the corresponding keys shape and fill. This allows you to define visual properties like circle shapes and filled elements for agents in your visualization.

Draft: Addition of 2 keys in enconding_dict: shape(str) and fill(bool)

PR #2172, opened 2024-07-07
Add .venv/ to .gitignore

EwoutH:
Don't track virtual environment files in Git.

See https://github.com/projectmesa/mesa/pull/2171#issuecomment-2211593881.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,3 +90,5 @@ pythonenv*/
 mesa/visualization/templates/external/
 mesa/visualization/templates/js/external/
 
+# Virtual environments
+.venv/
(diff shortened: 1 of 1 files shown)


rht (approved):

.gitignore
```diff
@@ -90,3 +90,5 @@ pythonenv*/
 mesa/visualization/templates/external/
 mesa/visualization/templates/js/external/
 
+# Virtual environments
+.venv/
```
rht, line 94: Looking further, this should be put next to L17 and L18. There is also L87 that they should all be grouped together.

rht: Feel free to merge when this has been addressed.

PR #2173, opened 2024-07-08
CI: Add test job for Python 3.13

EwoutH:
Python 3.13 is now in beta, thus is might be useful to start testing what works and what doesn't.

This PR adds a Python 3.13 job on Ubuntu.

Dependency tracker
- [x] rpds (https://github.com/crate-py/rpds/pull/79)
- [x] watchfiles (https://github.com/samuelcolvin/watchfiles/issues/276, https://github.com/samuelcolvin/watchfiles/pull/287)
- [x] matplotlib (https://github.com/matplotlib/matplotlib/pull/28688, waiting on 3.9.2 release)

### DIFF
--- a/.github/workflows/build_lint.yml
+++ b/.github/workflows/build_lint.yml
@@ -53,9 +53,12 @@ jobs:
         cache: 'pip'
     - name: Install uv
       run: pip install uv
-    - name: Install Mesa
-      # See https://github.com/astral-sh/uv/issues/1945
+    - name: Install Mesa with uv pip
       run: uv pip install --system .[dev]
+      if: matrix.python-version != '3.13'
+    - name: Install Mesa with pip
+      run: pip install .[dev]
+      if: matrix.python-version == '3.13'
     - name: Test with pytest
       run: pytest --durations=10 --cov=mesa tests/ --cov-report=xml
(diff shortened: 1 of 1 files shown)


EwoutH:
Watchfiles has issued a new release that’s compatible with Python 3.13.

Now it looks like it’s stuck, the CI run times out without readable log.

Maybe try an install without `uv`, but with regular `pip`.

rht:
Will it always be the case that `uv` can't be used on the pre-release Python? Or is there an unknown one-time error on `uv` that caused this (should be reported to them?)? The former means we have to live with the CI being as slow as the Python 3.13 build.

EwoutH:
> Will it always be the case that `uv` can't be used on the pre-release Python?

Depends. In this case Pandas doesn't have a Python 3.13 wheel on PyPI yet. uv didn't cache Pandas, so a build takes ~4 minutes. Regular pip does cache stuff, which makes it only 30 seconds.

We could configure uv cache, but I thought it was out of scope for this PR.

We could also wait with this PR until Pandas publishes a Python 3.13 wheel.

rht (approved):

rht:
It's about as slow as the Windows build anyway, so it should be fine.

PR #2178, opened 2024-07-13
Fixed code errors.

ENUMERA8OR:
Fixed code errors in various files based on pylint suggestions.

### DIFF
--- a/mesa/__init__.py
+++ b/mesa/__init__.py
@@ -27,10 +27,14 @@ __license__ = "Apache 2.0"
 _this_year = datetime.datetime.now(tz=datetime.timezone.utc).date().year
 __copyright__ = f"Copyright {_this_year} Project Mesa Team"
 
+
 def get_batch_run():
     from mesa.batchrunner import batch_run
+
     return batch_run
+
 
 def get_Model():
     from mesa import Model
+
     return Model
--- a/mesa/__init__.py
+++ b/mesa/__init__.py
@@ -2,13 +2,12 @@
 Mesa Agent-Based Modeling Framework
 Core Objects: Model, and Agent.
 """
-
 import datetime
-
 from mesa import space, time
 from mesa.agent import Agent
 from mesa.datacollection import DataCollector
 from mesa.model import Model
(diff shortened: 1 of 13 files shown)


rht (approved):

mesa/__init__.py
```diff
@@ -6,8 +6,7 @@
 
 import datetime
 
-import mesa.space as space
-import mesa.time as time
+from mesa import space, time
```
rht, line 9: This is the cause of the circular import in the CI fails.

ENUMERA8OR: Okay

ENUMERA8OR:
Updated code to circumvent CI error please check.

tpike3:
@ENUMERA8OR Thanks!, Could you also please update the tests so the builds are not failing. https://github.com/projectmesa/mesa/actions/runs/9927755923/job/27423103334?pr=2178

ENUMERA8OR:
okay will update them by today itself...

rht:
@ENUMERA8OR I have told you the actual line that causes the CI fails, https://github.com/projectmesa/mesa/pull/2178/files#diff-0a5df34385c06e41f8126b2cf65f4266da23f723edf96250be655f3d16565164R8. You should have listened to it to save time.

tests/test_agent.py
```diff
@@ -282,3 +282,11 @@ def test_agentset_shuffle():
     agentset = AgentSet(test_agents, model=model)
     agentset.shuffle(inplace=True)
     assert not all(a1 == a2 for a1, a2 in zip(test_agents, agentset))
+
+def get_Agent():
+    from mesa.agent import Agent
```
EwoutH, line 288: I don’t think these tests do what you want or expect them to do.

What exactly do you want to test here?

ENUMERA8OR: The tests kept failing so was updating the necessary files. Mainly circular import was the issue.

ENUMERA8OR:
Will close this pull request, will try from a new PR. Thanks for the info @rht, couldn't understand it at the time.

PR #2181, opened 2024-07-14
fix: Render agent marker radius correctly

rht:
Previously, on a 2x20 Schelling grid (extremely tall on the vertical direction, but thin on horizontal direction)
before
With this PR
after


tpike3 (approved):

EwoutH:
Should this be backported to 2.3?

rht:
Yeah, sounds good to me.

PR #2185, opened 2024-07-22
Test batch_run examples in CI

EwoutH:
Add a function to test the batch_run examples, which is also ran in CI.


EwoutH:
Depends on https://github.com/projectmesa/mesa-examples/pull/143.

PR #2187, opened 2024-07-29
Rename JupyterViz to SolaraViz

### LINKED ISSUE #2175: Rename JupyterViz to SolaraViz
I want to propose to rename `JupyterViz` to `SolaraViz` in the [visualization](https://github.com/projectmesa/mesa/tree/main/mesa/visualization). We build heavily onto Solara, and it's also used outside Jupyter notebook environments. I regularly call it the "Solara-based visualisation" right now, and you also enable it from the cli by running `solara run`. So it's a more descriptive and fitting name.

JupyterViz has't been published as a stable release (since https://github.com/projectmesa/mesa/pull/2090 hasn't been included in a stable release), so we can rename it relatively safely, albeit with proper announcements and a alias with a warning.

It's also gives credit where credit is due to the hard work of @maartenbreddels and his team.

rht:
Fixes #2175.

### DIFF
--- a/docs/tutorials/visualization_tutorial.ipynb
+++ b/docs/tutorials/visualization_tutorial.ipynb
@@ -119,9 +119,9 @@
    },
    "outputs": [],
    "source": [
-    "from mesa.visualization import JupyterViz\n",
+    "from mesa.visualization import SolaraViz\n",
     "\n",
-    "page = JupyterViz(\n",
+    "page = SolaraViz(\n",
     "    BoltzmannWealthModel,\n",
     "    model_params,\n",
     "    measures=[\"Gini\"],\n",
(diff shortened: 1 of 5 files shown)


rht:
Note: JupyterViz is still available. It's an alias to SolaraViz. i.e. This is not a breaking change.

rht:
> Just to double check, SolaraViz is the most descriptive and fitting name we could give the current visualisation, right? 

Actually, it could have been called just `Visualization`. But I'm fine either way.

Another backend would be Dash, see. e.g. vizro which has really good looking visualization: https://github.com/mckinsey/vizro.

EwoutH:
But then that would be a new function right? So instead of importing
```Python
```
We add a new, second function called DashViz for visualization, so you could also do:
```python
from mesa.visualization import DashViz
```

rht:
Yes, it would be a new function/class.

EwoutH (approved):

PR #2188, opened 2024-07-29
refactor: Rename jupyter_viz namespace to solara_viz

rht:
This is intentionally a separate PR so that it is easier to review.


EwoutH:
Wait, but if we add a second function/class here (https://github.com/projectmesa/mesa/pull/2187#issuecomment-2256408311), `solara_viz` is not fitting anymore, right?

rht:
A `PyGameViz` would be in pygame_viz.py. Otherwise it would be hard to read 2 different frameworks in 1 file.

EwoutH (approved):
Fair, looks good

PR #2195, opened 2024-08-05
devs/eventlist: Add str and repr methods to print EventList

EwoutH:
Add `__str__` and `__repr__` methods to print `EventList` objects in a useful format.

In models, this allows doing things like `print(self.simulator.event_list)` to validate the events queued in your simulator.

Output of `print(self.simulator.event_list)`:
- Before: `<mesa.experimental.devs.eventlist.EventList object at 0x00000186E42AF830>`
- After: `EventList([Event(time=7.046389752627814, priority=5, id=0), Event(time=8.305480965357855, priority=5, id=9), Event(time=10.904561103085891, priority=5, id=6)])`

### DIFF
--- a/mesa/experimental/devs/eventlist.py
+++ b/mesa/experimental/devs/eventlist.py
@@ -157,7 +157,13 @@ class EventList:
 
     def __str__(self) -> str:
         """Return a string representation of the event list"""
-        events_str = ', '.join([f"Event(time={e.time}, priority={e.priority}, id={e.unique_id})" for e in self._events if not e.CANCELED])
+        events_str = ", ".join(
+            [
+                f"Event(time={e.time}, priority={e.priority}, id={e.unique_id})"
+                for e in self._events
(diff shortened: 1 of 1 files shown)


Corvince:
Tha ks! This should probably be done for a lot more classes of mesa.

Just a note though: it is more common to just define __repr__ and then if __str__ is not defined, but called, it will automatically use the repr definition

EwoutH:
> Use `.__repr__()` for Programmers vs `.__str__()` for Users
> In general, you should use `str` when you want to display a string to the user, and `repr` when you want to represent a string in a way that can be used to recreate the original object, such as in a debugger or when logging the state of an object.

Thanks. I think the current implementation is more `__repr__` than `__str__`, since it's contains all the elements in it in a machine readable way. Will update, and then we can add an (even more) user readable string later.

rht (approved):

PR #2199, opened 2024-08-10
Implement experimental DataCollector API v3

rht:
Given that the fall semester is starting soon, there is not enough time to build a comprehensive new data collection API. So this is what I have cooked this afternoon.
It addresses https://github.com/projectmesa/mesa/discussions/1944#discussioncomment-8364109:
1. You can collect based on agent types using custom groups
2. Also via custom groups
3. Single-use measure is not yet addressed
4. Not yet measured for performance
5. There is `to_df` in this PR, and this feature is actually used in visualization
6. TODO
7. TODO, but this doesn't affect the API
8. I prefer 2-level grouping instead of flat dictionary. This allows for easy export to DF
9. TODO

and https://github.com/projectmesa/mesa/discussions/1944#discussioncomment-8439418:
1. same as 2 above
2. already implemented
3. already implemented in `to_df`

TODO:
- [ ] error handling
- [ ] documentation
- [ ] testing
- [ ] examples

### DIFF
--- a/mesa/experimental/datacollector.py
+++ b/mesa/experimental/datacollector.py
@@ -1,0 +1,42 @@
+from collections import defaultdict
+
+import pandas as pd
+
+
+class DataCollector:
+    def __init__(self, model, group_reporters, groups=None):
+        self.model = model
+        self.group_reporters = group_reporters
+        self.groups = groups
+        self.data = defaultdict(lambda: defaultdict(list))
+
+    def get_group(self, group_name):
+        if group_name == "model":
+            return self.model
+        elif group_name == "agents":
+            return self.model.agents
+        else:
+            try:
+                return getattr(self.model, group_name)
+            except AttributeError as e:
+                raise Exception(f"Unknown group: {group_name}") from e
+
+    def report(self, reporter, group):
+        if group is self.model:
+            return reporter(group)
+        if isinstance(reporter, str):
+            if hasattr(group, "get"):
+                return group.get(reporter)
+            else:
+                raise Exception()
+        return reporter(group)
+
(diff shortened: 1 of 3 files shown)


rht:
For example, see https://github.com/projectmesa/mesa-examples/pull/151.

mesa/experimental/datacollector.py
```diff
@@ -0,0 +1,42 @@
...
+            try:
+                return getattr(self.model, group_name)
+            except AttributeError as e:
+                raise Exception(f"Unknown group: {group_name}") from e
+
+    def report(self, reporter, group):
+        if group is self.model:
+            return reporter(group)
+        if isinstance(reporter, str):
+            if hasattr(group, "get"):
```
rht, line 28: This should be replaced with `isinstance(group, AgentSet)` or isinstance of `Group`.

EwoutH:
Thanks for this effort! It doesn’t look too bad, could you give some example usages of more complicated usage scenarios?

rht:
Which example do you want me to port? Is wolf-sheep (agents by type) sufficient?

EwoutH:
I don't need a full model port, I would like some (more complex) API usage examples. Something in the form of:
- https://github.com/projectmesa/mesa/discussions/1944#discussion-6040717
- https://github.com/projectmesa/mesa/issues/348#issuecomment-1777533710

rht:
@EwoutH this should be a sufficient illustration: https://github.com/projectmesa/mesa-examples/pull/158. It has agents by type.

quaquel:
This PR is stale, and we are at long last picking this up again.

PR #2207, opened 2024-08-12
Use PyData config for RTD colorscheme

rht:
See https://github.com/readthedocs/sphinx_rtd_theme/issues/1496.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -123,7 +123,8 @@ html_theme = "pydata_sphinx_theme"
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
 # documentation.
-# html_theme_options = {}
+# See https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/styling.html#configure-pygments-theme
+html_theme_options = {"pygments_light_style": "gruvbox-light", "pygments_dark_style": "gruvbox-dark"}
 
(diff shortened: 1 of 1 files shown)


rht:
It works: https://mesa--2207.org.readthedocs.build/en/2207/howto.html.

EwoutH:
image

For me the only thing that looks different is the code coloring scheme in dark mode. In light mode it's completely identical.

Is that the intention of this PR?

EwoutH:
To be brutally honest: I like the current color scheme better.

rht:
It's not my opinion, it's that gruvbox is the most popular colorscheme at least in the Vim space (not sure about the other editors): https://vimcolorschemes.com/i/top.

EwoutH:
Let's do a vote!

First, the light color scheme:

- 🎉 = Keep the current code color scheme (left)
- 🚀 = Use the new code color scheme (right)
- 👀 = Don't care / indifferent / both good

Screenshot_475

EwoutH:
Then, the dark color scheme:

- 🎉 = Keep the current code color scheme (left)
- 🚀 = Use the new code color scheme (right)
- 👀 = Don't care / indifferent / both good

Screenshot_474

EwoutH:
It's 5-2 on the light and 4-3 on the dark format, on keeping the current color scheme. I think that concludes that it's preferred to keep the current one?

rht:
What a strange crowd.

Issue #2164, opened 2024-07-03
Visualisation: Allow different shapes agent portrayals


EwoutH:
In the old visualisation you could add values like `"Shape": "circle"` and `"Filled": "true"` to the `agent_portrayal` dict.

With the new Solara based visualisation, that's not yet possible. It should be implemented in [mesa/visualization/components](https://github.com/projectmesa/mesa/tree/74e2df6fdfc2d5172f84eb96469b277b35bfd795/mesa/visualization/components). Somewhere around here:
https://github.com/projectmesa/mesa/blob/74e2df6fdfc2d5172f84eb96469b277b35bfd795/mesa/visualization/components/altair.py#L50-L55
and maybe also in:
https://github.com/projectmesa/mesa/blob/74e2df6fdfc2d5172f84eb96469b277b35bfd795/mesa/visualization/components/matplotlib.py#L40-L47

EwoutH:
In think (in Altair) it's called a Mark: https://altair-viz.github.io/user_guide/marks/index.html

Aryan1Mahahjan:
Hi, was thinking about using the below, which use the Encode function attribute

```py
has_shape = "shape" in all_agent_data[0]
    if has_shape:
        encoding_dict["shape"] = agent_data["shape"]

has_fill = "fill" in all_agent_data[0]
    if has_shape:
        encoding_dict["fill"] = agent_data["fill"]

``` 
The provided code snippet checks for `shape` and `fill` keys in agent data. If present, it adds them to the encoding_dict using the corresponding keys `shape` and `fill`. This allows you to define visual properties like circle shapes and filled elements for agents in your visualization.

EwoutH:
Looks like a good start! I think it does need to be passed to Altair or matplotlib later in the code.

Feel free to open a (draft) PR!

PR #2214, opened 2024-08-15
Visualisation: Allow specifying Agent shapes in agent_portrayal

rmhopkins4:
This PR allows specifying an `"shape"` in the `agent_portrayal` dictionary used by matplotlib component of the Solara visualisation. In short, it allows you represent an Agent in any [matplotlib marker](https://matplotlib.org/stable/api/markers_api.html), by adding a "shape" key-value pair to the `agent_portrayal` dictionary.

This is especially useful when you're using the default shape drawer for grid or continuous space.

For example:
```Python
def agent_portrayal(cell):
    return {
        "color": "blue"
        "size": 5,
        "shape": "h"  # marker is a hexagon!
    }
```
Partial resolution of #2164.

### Examples

In Conway's Game of Life life, you can now use black squares for the agents, as how it's traditionally displayed:

```Python
def agent_portrayal(cell):
    return {
        "color": "white" if cell.alive else "black",
        "size": 45,
        "shape": "s" # marker is a square
    }
```
image

You can also conditionally modify the shape, just like you can with the color and size:

```Python
COP_COLOR = "#000000"
AGENT_QUIET_COLOR = "#648FFF"
AGENT_REBEL_COLOR = "#FE6100"
JAIL_COLOR = "#808080"
JAIL_SHAPE = "s"


def citizen_cop_portrayal(agent):
    portrayal = {
        "Shape": "circle",
        "x": agent.pos[0],
        "y": agent.pos[1],
        "Filled": "true",
    }

    if isinstance(agent, Citizen):
        color = (
            AGENT_QUIET_COLOR if agent.condition == "Quiescent" else AGENT_REBEL_COLOR
        )
        if agent.jail_sentence:
            color = JAIL_COLOR
            shape = JAIL_SHAPE
        else:
            shape = "."
        color = JAIL_COLOR if agent.jail_sentence else color
        portrayal["color"] = color
        portrayal["shape"] = shape
        portrayal["size"] = 15.0

    else:
        assert isinstance(agent, Cop)
        portrayal["color"] = COP_COLOR
        portrayal["shape"] = "."
        portrayal["size"] = 25.0

    return portrayal
```
image

### DIFF
--- a/mesa/visualization/components/matplotlib.py
+++ b/mesa/visualization/components/matplotlib.py
@@ -36,7 +36,12 @@ def _split_and_scatter(portray_data, space_ax):
     c = portray_data["c"]
     m = portray_data["m"]
 
-    assert len(x) == len(y) == len(s) == len(c) == len(m)
+    if not (len(x) == len(y) == len(s) == len(c) == len(m)):
+        raise ValueError(
+            "Length mismatch in portrayal data lists: "
+            f"x: {len(x)}, y: {len(y)}, s: {len(s)}, "
+            f"c: {len(c)}, m: {len(m)}"
+        )
 
     # Group the data by marker
     for i in range(len(x)):
--- a/mesa/visualization/components/matplotlib.py
+++ b/mesa/visualization/components/matplotlib.py
@@ -39,8 +39,8 @@ def _split_and_scatter(portray_data, space_ax):
     if not (len(x) == len(y) == len(s) == len(c) == len(m)):
         raise ValueError(
             "Length mismatch in portrayal data lists: "
-            f"x: {len(x)}, y: {len(y)}, s: {len(s)}, "
-            f"c: {len(c)}, m: {len(m)}"
+            f"x: {len(x)}, y: {len(y)}, size: {len(s)}, "
+            f"color: {len(c)}, marker: {len(m)}"
         )
 
     # Group the data by marker
@@ -53,7 +53,8 @@ def _split_and_scatter(portray_data, space_ax):
 
     # Plot each group with the same marker
     for marker, data in grouped_data.items():
-        space_ax.scatter(data["x"], data["y"], s=data["s"], c=data["c"], marker=marker)
+        space_ax.scatter(data["x"], data["y"], s=data["s"],
+                         c=data["c"], marker=marker)
 
 
 def _draw_grid(space, space_ax, agent_portrayal):
@@ -82,9 +83,9 @@ def _draw_grid(space, space_ax, agent_portrayal):
                     # establishing a default prevents misalignment if some agents are not given size, color, etc.
                     size = data.get("size", default_size)
                     s.append(size)
-                    color = data.get("color", "b")
+                    color = data.get("color", "tab:blue")
                     c.append(color)
-                    mark = data.get("shape", ".")
+                    mark = data.get("shape", "o")
                     m.append(mark)
         out = {"x": x, "y": y, "s": s, "c": c, "m": m}
         return out
@@ -123,9 +124,9 @@ def _draw_continuous_space(space, space_ax, agent_portrayal):
             # establishing a default prevents misalignment if some agents are not given size, color, etc.
             size = data.get("size", default_size)
             s.append(size)
-            color = data.get("color", "b")
+            color = data.get("color", "tab:blue")
             c.append(color)
-            mark = data.get("shape", ".")
+            mark = data.get("shape", "o")
             m.append(mark)
         out = {"x": x, "y": y, "s": s, "c": c, "m": m}
         return out
--- a/mesa/visualization/components/matplotlib.py
+++ b/mesa/visualization/components/matplotlib.py
@@ -2,6 +2,7 @@ import networkx as nx
 import solara
 from matplotlib.figure import Figure
 from matplotlib.ticker import MaxNLocator
+from collections import defaultdict
 
 import mesa
 
@@ -23,12 +24,43 @@ def SpaceMatplotlib(model, agent_portrayal, dependencies: list[any] | None = Non
     solara.FigureMatplotlib(space_fig, format="png", dependencies=dependencies)
 
 
+# matplotlib scatter does not allow for multiple shapes in one call
+def _split_and_scatter(portray_data, space_ax):
+    grouped_data = defaultdict(lambda: {"x": [], "y": [], "s": [], "c": []})
+
+    if "marker" not in portray_data:
+        # standard scatter is fine, default marker
+        space_ax.scatter(**portray_data)
+        return
+
+    # Extract data from the dictionary
+    markers = portray_data["marker"]
+    x = portray_data["x"]
+    y = portray_data["y"]
+    s = portray_data["s"]
+    c = portray_data["c"]
+
+    # Group the data by marker
+    for i in range(len(x)):
+        marker = markers[i]
+        grouped_data[marker]["x"].append(x[i])
+        grouped_data[marker]["y"].append(y[i])
+        grouped_data[marker]["s"].append(s[i])
+        grouped_data[marker]["c"].append(c[i])
+
+    # Plot each group with the same marker
+    for marker, data in grouped_data.items():
+        space_ax.scatter(data["x"], data["y"], s=data["s"],
+                         c=data["c"], marker=marker)
+
+
 def _draw_grid(space, space_ax, agent_portrayal):
     def portray(g):
         x = []
         y = []
         s = []  # size
         c = []  # color
+        marker = []  # shape
         for i in range(g.width):
             for j in range(g.height):
                 content = g._grid[i][j]
@@ -45,6 +77,8 @@ def _draw_grid(space, space_ax, agent_portrayal):
                         s.append(data["size"])
                     if "color" in data:
                         c.append(data["color"])
+                    if "shape" in data:
+                        marker.append(data["shape"])
         out = {"x": x, "y": y}
         # This is the default value for the marker size, which auto-scales
         # according to the grid area.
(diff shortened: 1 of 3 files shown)


mesa/visualization/components/matplotlib.py
```diff
@@ -23,12 +25,42 @@ def SpaceMatplotlib(model, agent_portrayal, dependencies: list[any] | None = Non
...
+    # Group the data by marker
+    for i in range(len(x)):
+        marker = markers[i]
+        grouped_data[marker]["x"].append(x[i])
+        grouped_data[marker]["y"].append(y[i])
+        grouped_data[marker]["s"].append(s[i])
+        grouped_data[marker]["c"].append(c[i])
+
+    # Plot each group with the same marker
+    for marker, data in grouped_data.items():
```
rht, line 53: I'm not sure of the performance implication of plotting each points separately. What about https://stackoverflow.com/a/43528356?

mesa/visualization/components/matplotlib.py
```diff
@@ -23,12 +25,42 @@ def SpaceMatplotlib(model, agent_portrayal, dependencies: list[any] | None = Non
     solara.FigureMatplotlib(space_fig, format="png", dependencies=dependencies)
 
 
+# matplotlib scatter does not allow for multiple shapes in one call
```
rht, line 28: It seems that Altair natively supports this: https://github.com/vega/altair/issues/1181.

mesa/visualization/components/matplotlib.py
rmhopkins4: Points are grouped by marker (shape) before plotting, like in the link provided.

mesa/visualization/components/matplotlib.py
rmhopkins4: Altair space_drawer does not support continuous space. To have continuous space w/ multiple shapes altair must either support continuous space or matplotlib must support multiple shapes.

mesa/visualization/components/matplotlib.py
rmhopkins4: one way to potentially speed the grouping up is to use numpy.
if the x, y, s, c, markers arrays were np arrays, masks could be used on each np.unique() marker

rmhopkins4: But I understand not wanting to introduce dependencies if not strictly necessary

rht:
Implementation LGTM, thank you for implementing this feature. I could test it on the Epstein civil violence model if I have the time. But if you don't mind, showing that the jailed citizens as square in that example would be helpful. You can used the code from https://github.com/projectmesa/mesa-examples/pull/158 and undo the experimental data collector part.

mesa/visualization/components/matplotlib.py
rht: I think we should do both, with the Altair one in a separate PR. In your code, it falls back to the default code branch when shape is not present, so it's fine.

rmhopkins4:
image
Here is an example of jailed citizens as squares.
While working on this I made matplotlib.py provide default visualization values so that users only need to provide overrides

rmhopkins4:
> image Here is an example of jailed citizens as squares. While working on this I made matplotlib.py provide default visualization values so that users only need to provide overrides

Matplotlib's 's' and 'c' (size, color) accept no values (all default), one value (all the same), and lists exactly as long as the number of points provided. 
When the list lengths were not aligned problems arose.

mesa/visualization/components/matplotlib.py
```diff
@@ -23,12 +25,39 @@ def SpaceMatplotlib(model, agent_portrayal, dependencies: list[any] | None = Non
...
+    grouped_data = defaultdict(lambda: {"x": [], "y": [], "s": [], "c": []})
+
+    # Extract data from the dictionary
+    x = portray_data["x"]
+    y = portray_data["y"]
+    s = portray_data["s"]
+    c = portray_data["c"]
+    m = portray_data["m"]
+
+    assert len(x) == len(y) == len(s) == len(c) == len(m)
```
rht, line 39: Can't use assertion here. Needs to raise an exception because this would be disabled by the Python `-O` flag (to disable assertions).

EwoutH:
Thanks a lot for this PR! I will review it somewhere tomorrow.

mesa/visualization/components/matplotlib.py
```diff
@@ -23,12 +25,44 @@ def SpaceMatplotlib(model, agent_portrayal, dependencies: list[any] | None = Non
...
+    # Extract data from the dictionary
+    x = portray_data["x"]
+    y = portray_data["y"]
+    s = portray_data["s"]
+    c = portray_data["c"]
+    m = portray_data["m"]
+
+    if not (len(x) == len(y) == len(s) == len(c) == len(m)):
+        raise ValueError(
+            "Length mismatch in portrayal data lists: "
```
rht, line 41: The error needs to show the public name of these attributes: x, y, size, color, marker instead of their Matplotlib argument names.

mesa/visualization/components/matplotlib.py
```diff
@@ -41,23 +75,23 @@ def portray(g):
...
-        if len(c) > 0:
-            out["c"] = c
+
+                    # This is the default value for the marker size, which auto-scales
+                    # according to the grid area.
+                    default_size = (180 / max(g.width, g.height)) ** 2
+                    # establishing a default prevents misalignment if some agents are not given size, color, etc.
+                    size = data.get("size", default_size)
+                    s.append(size)
+                    color = data.get("color", "b")
```
rht, line 85: Matplotlib's `tab:blue` used in their default plot looks nicer than the default `b` (`blue`).

rht: See https://matplotlib.org/stable/gallery/color/named_colors.html#tableau-palette.

mesa/visualization/components/matplotlib.py
```diff
@@ -41,23 +75,23 @@ def portray(g):
...
+                    c.append(color)
+                    mark = data.get("shape", ".")
```
rht, line 87: According to https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.scatter.html, the default is actually `"o"`.

EwoutH:
@rht could you give a short summary what you already have sufficiently reviewed and what could use a second pair of eyes?

rht:
Actually, this PR is already LGTM, and I didn't merge because you still wanted to review. If any, maybe the perf consideration of an array of defaults instead of a single default value scalar, but this is minor because any loop done without calling Matplotlib many times is going to be fast.

rht (approved):

EwoutH:
Thanks. I would have expected this feature taking less complexity / lines of code, let me take a closer look tomorrow morning.

EwoutH:
Okay, I'm largely good with this PR. It's a bit weird https://github.com/matplotlib/matplotlib/issues/11155 was never implemented, but since the only thing changes for our users is that they can now _optionally_ pass a matplotlib shape in a `portrayal` dict (right @rht?), we can always implement the back-end later in a more (performance) efficient way.

I would like the documentation around this PR to improve a little:
- Update the docstring to note that a `"shape"` can be passed, and add a link to the list with allowed values (I think it's [this one](https://matplotlib.org/stable/api/markers_api.html))
- Update the first comment of this PR with the screenshot, one or two more examples, and a link to 
- Mention that you can pass "shape" in the [visualization_tutorial](https://mesa.readthedocs.io/en/latest/tutorials/visualization_tutorial.html) ([source](https://github.com/projectmesa/mesa/blob/main/docs/tutorials/visualization_tutorial.ipynb)).

rht:
> Okay, I'm largely good with this PR. It's a bit weird https://github.com/matplotlib/matplotlib/issues/11155 was never implemented, but since the only thing changes for our users is that they can now optionally pass a matplotlib shape in a portrayal dict (right @rht?), we can always implement the back-end later in a more (performance) efficient way.

Yes, it's optional and is not a breaking change.

While it's good to update the docstring now, I think the tutorial change can happen in a separate PR. Otherwise it makes Mesa too bureaucratic to contribute to (that every new feature needs to be in the tutorial in the same PR).

EwoutH:
Yeah I see your perspective and also considered that. But otherwise I'm afraid it doesn't get done. Especially with our user base changes need to be (very) visible.

@rmhopkins4 let me know if you want to do these things yourself, or otherwise allow us to do it within this PR.

rmhopkins4:
> Yeah I see your perspective and also considered that. But otherwise I'm afraid it doesn't get done. Especially with our user base changes need to be (very) visible.
> 
> @rmhopkins4 let me know if you want to do these things yourself, or otherwise allow us to do it within this PR.

You should have the ability to edit the PR already, no?

EwoutH:
Yes I do, but I don't prefer to edit other peoples PRs without their explicit permission :)

EwoutH:
@rmhopkins4 Could you let me know if you want to take up (any of) the remaining points yourself, or if you want us as maintainers to do it?

rmhopkins4:
I can take up the remaining points. If I understand correctly it's just documentation-related.

rmhopkins4:
Is there a docstring that currently mentions the available options for the `agent_portrayal`? They seem to just state that `agent_portrayal` lets you specify agents' display options but not how.

EwoutH:
I think you’re right, it’s weird we don’t have that documented.

It might have gone lost with the transition from the old to the new visualization.

I think this would be the most logical place to add it:
https://github.com/projectmesa/mesa/blob/3cf1b76ca3a157c8a064b60130755a86e73718be/mesa/visualization/solara_viz.py#L106
But is also shows how important PR, release and tutorial documentation is - that’s how users discover new features!

EwoutH:
@rmhopkins4 are you done working on it? It looks good!

mesa/visualization/components/matplotlib.py
rht: I want to add that NumPy has a lot of overhead. It is faster only for large arrays.

EwoutH:
> Here is an example of jailed citizens as squares.

@rmhopkins4 I'm ready to merge. Could you add the portrayal code you used to the first PR post?

EwoutH:
Congratulations of getting your first PR into Mesa @rmhopkins4!

EwoutH:
@rmhopkins4 would you like to update a few example models to use the shapes?

If so, this issue includes a list of models already using the new visualization:
- https://github.com/projectmesa/mesa-examples/issues/154

PR #1928, opened 2023-12-26
Work around for initializing model._agent

quaquel:
In response to the discussion in #1917, this PR contains a simple workaround if model._agent has not been set up because `super().__init__` was not called when subclassing `Model`. 

It also fixes the bug mentioned in #1917 regarding `model.get_agents_of_type`.

PR #2218, opened 2024-08-17
Require Mesa models to be initialized with `super().__init__()`

EwoutH:
Remove the workaround for when the model super has not been called, and thereby require Mesa models to be initialized with `super().__init__()`.

```Python
class MyModel(mesa.Model):
    def __init__(self):
        super().__init__()  # This is now required!
```
This is needed to correctly add Agents to the model and setup other Mesa model variables.

This warning was present since Mesa 2.2.0 (https://github.com/projectmesa/mesa/pull/1928), so there has been some time to update for existing users.

This PR contains a second commit that names experimental warning in PropertyLayer correctly in space.py.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -11,9 +11,7 @@ from __future__ import annotations
 import contextlib
 import copy
 import operator
-import warnings
 import weakref
-from collections import defaultdict
 from collections.abc import Callable, Iterable, Iterator, MutableSet, Sequence
 from random import Random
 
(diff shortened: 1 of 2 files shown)


mesa/agent.py
```diff
@@ -50,19 +48,7 @@ def __init__(self, unique_id: int, model: Model) -> None:
...
-            self.model.agents_ = defaultdict(dict)
-            self.model.agents_[type(self)][self] = None
-            self.model.agentset_experimental_warning_given = False
-
-            warnings.warn(
-                "The Mesa Model class was not initialized. In the future, you need to explicitly initialize the Model by calling super().__init__() on initialization.",
-                FutureWarning,
-                stacklevel=2,
-            )
+        self.model.agents_[type(self)][self] = None
```
rht, line 51: What about `try-except` then raise error, instead of a warning, for now? I'm not sure if some people do read the release note for such a subtle change in the way they have to do `__init__()`.

EwoutH: I don't think it's a bad idea, but with Mesa 3.0 I would like to clean the codebase and remove it altogether.

In the 3.0 release notes we can put a note to first update to the latest Mesa 2.x, which also contains this warning (and it now has been in there for all 2.2.x and 2.3.x releases).

EwoutH: Thought about it some more, and I think you're actually right. It's better to throw an useful error than let it crash without any guidance.

rht (approved):

PR #2219, opened 2024-08-17
Allow AgentSet.do() to take Callable function

quaquel:
This PR enhances `AgentSet.do` to take a callable or str. Currently, AgentSet.do takes a `str` which maps to a method on the agents in the set. This PR makes it possible to also use a `Callable` instead.

This callable will be called with the `agent` as the first argument.

### ⚠️ Breaking change ⚠️ 
A small breaking change is introduced here: the `method_name` parameter is renamed to `method`. For models that use this as a keyword argument this is a breaking change, and need to replace `do(method_name="something")` with `do(method="something")`.

### Usage examples
Here are a few examples of how to use `do()` with `Callable` functions.

Suppose we want to move each agent by a fixed offset:

```python
def move_agent(agent, dx, dy):
    # Assume each agent has a 'pos' attribute which is a tuple (x, y)
    if agent.pos:
        agent.pos = (agent.pos[0] + dx, agent.pos[1] + dy)

# Usage with do()
agentset.do(move_agent, 1, -1)
```

You might want to increase the energy level of all agents by a fixed amount:

```python
agentset.do(lambda agent: setattr(agent, 'energy', agent.energy + 10))
```

You can apply a function conditionally to agents that meet a certain criterion:

```python
def heal_agent(agent):
    if agent.health < 50:
        agent.health += 20

# Apply the healing only to agents with health below 50
agentset.do(lambda agent: heal_agent(agent) if agent.health < 50 else None)
```

You may want to gather results from applying a function to each agent:

```python
def get_agent_status(agent):
    return f"Agent {agent.unique_id}: {agent.status}"

statuses = agentset.do(get_agent_status, return_results=True)
```

If your callable needs additional arguments, you can pass them directly via `do()`:

```python
def update_wealth(agent, amount, tax_rate):
    agent.wealth += amount * (1 - tax_rate)

agentset.do(update_wealth, 100, 0.2)
```

Suppose you want to update each agent's position based on their velocity:

```python
def update_position(agent):
    if agent.pos and hasattr(agent, 'velocity'):
        agent.pos = (agent.pos[0] + agent.velocity[0], agent.pos[1] + agent.velocity[1])

agentset.do(update_position)
```

And of course the current behavior of passing a string as method is still supported:

```python
agentset.do("step")
```

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -227,7 +227,7 @@ class AgentSet(MutableSet, Sequence):
         return self
 
     def do(
-        self, callable: str|Callable, *args, return_results: bool = False, **kwargs
+        self, callable: str | Callable, *args, return_results: bool = False, **kwargs
     ) -> AgentSet | list[Any]:
         """
         Invoke a method or function on each agent in the AgentSet.
@@ -258,7 +258,6 @@ class AgentSet(MutableSet, Sequence):
                 for agentref in self._agents.keyrefs()
                 if (agent := agentref()) is not None
             ]
-
 
         return res if return_results else self
 
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -389,7 +389,7 @@ class AgentSet(MutableSet, Sequence):
         else:
             raise ValueError(f"axis should be `agent` or `agentset` not {axis}")
 
-    def group_by(self, by: Callable | str) -> dict[str, AgentSet]:
+    def group_by(self, by: Callable | str, return_agentset=False) -> dict[str, list] | dict[str, AgentSet]:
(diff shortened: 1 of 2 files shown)


EwoutH (commented):
Awesome! Code wise and docstring wise it looks really good to me. Only a few tests are needed.

EwoutH (approved):

EwoutH:
I added tests and updated the Readme with a note about the breaking change and some usage examples.

@quaquel please check the tests and examples and @projectmesa/maintainers I would love one additional review.

rht (approved):

PR #2220, opened 2024-08-17
Add AgentSet.groupby

quaquel:
This PR adds pandas style groupby to AgentSet (a first draft for this was #2092, but this PR has stalled). It also adds, as in pandas, a GroupBy helper class to enable method chaining. The resulting API is illustrated below.

**Key Changes**

* adds a `group_by` method to AgentSet; this method enables pandas style group_by operations on the agents in the set. 
* adds a `GroupBy` helper class; this helper class is returned by the new `group_by` method and, as in pandas, has a few methods to enable method chaining in a [split, apply, combine](https://pandas.pydata.org/docs/user_guide/groupby.html) style syntax. 

**usage examples**

```python

# eppstein civil violence
# will give you a dict with the number of agents for each condition
# the result_type optional kwarg allows finer control over the group data structure
# which enhances performance if you don't need the fancy stuff in agentset
# map returns a dict with group_name and the return of whatever function is applied to each group
some_agentset.groupby("condition", result_type="list").map(len)

# random activation by type
groups = some_agentset.groupby(type)
for group, agents in groups: # we can iterate over the groups
    agents.shuffle(inplace=True).do("step")

# we can easily get the number of groups
len(groups)

#  we can get a dict with the group_name and the groups
groups.groups

# one liner random activation by type
# do returns the underlying agent set
# note how do and appply can take either a callable or a string analogous to AgentSet.do
some_agentset.groupby(type).do("shuffle", inplace=True).do("step")

```


EwoutH:
> specifically enable user choice between agentsets and lists. Lists are faster but should be used with care

It seems that this is a recurring issue.

I have no idea if something like the following is possible in Python, but ideally you would have a fast construct that some function can accept and some functions can output aside from the agentset. As soon as you match or chain two of such functions, they recognize they can take the "fast path" and avoid compiling an recompiling an agentset.

But that's on a high level, no idea if it's possible/feasible to implement in our case in Python.

EwoutH:
Apparently this is a concept called "Deferred Execution". Basically, you can do something like this:
- Instead if directly performing each operation, you chain all operations that need to be performed in a list
- Once you have gathered all operations, you check for each out if they support the fast data construct as both input and output
- If there are any matches, flip a boolean for those operations
- They you execute them as the original chain

You need some wrapper and/or decorator magic to make this work, and it adds a lot of code complexity. But then it could prevent unnecessary conversions.

It's kind of related to the lazy execution concept, but now you're able to look forward instead of only backwards.

quaquel:
> It seems that this is a recurring issue.

Yes it is

> Apparently this is a concept called "Deferred Execution". Basically, you can do something like this:

I had a look but this seems a lot of trouble to get working. You need to use reflection to figure out all calls and then complex logic to figure out what do to. The short term and easier solution is to allow users to control this explicitly. With the `inplace` argument on for example `shuffle` we are already doing so. It is just a matter of documenting it carefully and making clear that it can be beneficial to profile this carefully.

quaquel:
This is ready for review. All tests are passing and the docstring is in place. The list/agentset stuff has been raised by Ewout and I believe giving users control is the least worst solution at the moment. 

What remains is the split apply combine stuff. Split/groupby is there. Apply is available on the `GroupBy` helper class. I am less sure on the need and desirability of an explicit control for combine. In pandas this is needed because everything needs to be transformed back to a DataFrame/Series. We don't have that need in MESA.

EwoutH:
Thanks a lot for this. Will take a close look tomorrow. If you have the chance, could you maybe add a few more user API examples in the PR start?

mesa/agent.py
```diff
@@ -357,7 +358,69 @@ def random(self) -> Random:
...
+        """
+
+        # return group for specified name
+        return self.groups[name]
+
+    def apply(self, callable: Callable, *args, **kwargs):
+        """Apply the specified callable to each group
+
+        Args:
+            callable (Callable): The callable to apply to each group, it will be called with the group as first argument
```
EwoutH, line 416: Would it be useful to also pass the group key here?

EwoutH:
On initial look this is a clean and solid start, especially API wise. I don't expect much work to be needed, we can extend what we find missing in later PRs.

Tomorrow I will play with some models to get a real feel for this function.

quaquel:
> On initial look this is a clean and solid start, especially API wise. I don't expect much work to be needed, we can extend what we find missing in later PRs.

I came up with a couple of simple additions to `GroupBy.apply`. First, it now only takes a callable, this should be changed to a callable or string so you can call methods on the underlying agentset if desired. Next, it might be nice to control the return type to be either the return of the callable/str or the GroupBy instance itself (as we do in AgentSet in places). This allows for chaining like

```python

# random activation by type
some_agentset.groupby(type).apply("shuffle", returntype="groupby", inplace=true).apply("do")

```

mesa/agent.py
quaquel: No. Apply in pandas goes over all groups and if you want to do it on a single group you can just do

```python
groupes = some_agentset.groupby("condition")
callable(groupes.get_group["quiet"])
```

EwoutH:
Thanks. I think the best to keep group_by properly scoped is to have a few very clear use cases. Data collection (with aggregation) is clearly one, RandomActivationByType could be another one.

What I find weird about that scheduler by the way, is that is shuffles the order of the agent groups and the agents within the group _both_ by _default_:

https://github.com/projectmesa/mesa/blob/3cf1b76ca3a157c8a064b60130755a86e73718be/mesa/time.py#L349

Which means the same group can go twice in a row, etc. Not a weird option to have, but strange by default.

Luckily the AgentSet is way more flexible. A `shuffle_groups()` method might be useful in `GroupBy` if you want to replicate this behavior.

But all based on clear use cases.

mesa/agent.py
```diff
@@ -357,7 +358,69 @@ def random(self) -> Random:
...
+        Group agents by the specified attribute
+
+        Args:
+            by (Callable, str): used to determine what to group agents by
+
+                                * if ``by`` is a callable, it will be called for each agent and the return is used
+                                  for grouping
+                                * if ``by`` is a str, it should refer to an attribute on the agent and the value
+                                  of this attribute will be used for grouping
+            result_type (str, optional): The datatype for the resulting groups {"agentset", "list"}
```
rht, line 372: A documentation saying that using list is faster than AgentSet would be informative.

Corvince:
I think this PR adds a nice feature, but I would advice to carefully reconsider the return types. Currently the flow is like the following, if we consider all possible return types (including the ones proposed for apply)

```python
agentset.group_by("foo").apply(bar)
AgentSet -> GroupBy[str, AgentSet | list] -> Dict[str, Any] | GroupBy[str, AgentSet | list]
```

It took me quite some time to actually figure this out, because you really have to think through what can be returned by each function. Compare this to Pandas
```python
dataframe.groupby("foo").apply(bar)
DataFrame -> DataFrameGroupBy[str, DataFrame] -> DataFrame
```

I would recommend following a similar path and staying as close to our AgentSet as possible. I think defining return types is also a bit unpythonic. Instead of `agentset.groupby(foo, response_type="list")` it is more pythonic to just use `list(agentset.groupby(foo)`. 

Also please note the different spelling of `group_by` (this PR) and `groupby` (pandas)

quaquel:
> I think this PR adds a nice feature, but I would advice to carefully reconsider the return types. Currently the flow is like the following, if we consider all possible return types (including the ones proposed for apply)

I have been somewhat concerned about this as well, but you drive home the importance of sorting it out. Its even more confusing then I thought. 

> I think defining return types is also a bit unpythonic. Instead of agentset.groupby(foo, response_type="list") it is more pythonic to just use list(agentset.groupby(foo).

Here I disagree somewhat. Pandas has control for returntype in various places E.g. DataFrame.apply has a way of specifying the result_type. We also do some of it in AgentSet already (i.e., AgentSet.do). The main problem is that in pandas, groupby will allways go back to a dataframe. I don't think there is such a obvious default in our case. 

One possible solution is to only add `group_by` (renamed as @Corvince  points out to groupby) and don't add the GroupBy class with the apply function.

rht:
> One possible solution is to only add group_by (renamed as @Corvince points out to groupby) and don't add the GroupBy class with the apply function.

I thought this was because Polars had decided to use `group_by` instead of `groupby`. In dplyr, it is [group_by](https://dplyr.tidyverse.org/reference/group_by.html). I'm not sure of Polars' reasoning, but it seems to be for consistency with dplyr and that Polars is a cross-language library. I'm fine with either way.

mesa/agent.py
```diff
@@ -357,7 +358,74 @@ def random(self) -> Random:
...
+        Returns:
+            GroupBy
+
+
+        Nptes:
```
rht, line 377: Typo.

rht:
Other than the typo, everything else LGTM.

EwoutH:
@quaquel ready for final review, or were you still working on stuff?

quaquel:
> @quaquel ready for final review, or were you still working on stuff?

There is still [this idea](https://github.com/projectmesa/mesa/pull/2220#issuecomment-2298042278). The first one (i.e. callable or string on `GroupBy.apply`), I most definately will this morning. Feedback on second point is appreciated and interacts with the point of @Corvince.

EwoutH:
Thanks

> changed to a callable or string so you can call methods on the underlying agentset

That might be useful indeed.

> return of the callable/str

How would this work exactly? Can you give some usage examples?

quaquel:
> How would this work exactly? Can you give some usage examples?

SImilar to `AgentSet.do` and `return_results=False`. 

```python

# one liner random activation by type
some_agentset.groupby(type).apply("shuffle", inplace=True, return_results=False).apply("step", return_results=False)

```

Corvince:
I like how this evolves and I think this will be a nice building block.

However, I think providing do-like functionality under the apply name will be confusing. 

I don't know if we had this discussion already for AgentSet, but what if we split into a `apply` function and a `do` function? Where apply applies a function and returns the result. And do applies a function, but returns the AgentSet. I think this would be clearer and easier to understand in terms of data flow. 
Maybe this could also apply to the distinction of inplace? Where "do" would apply inplace and "apply" would return a new AgentSet (for functions that return an AgentSet´). Right now this would make a lot of sense in my mind, but I haven't fully thought this through yet, so I might overlook something.

quaquel:
>I don't know if we had this discussion already for AgentSet, but what if we split into a apply function and a do function?  

I have to think about this a bit. There is tradeoff between using different methods and just having a keyword to control dataflow.  Having several methods expands the API that a user has to remember. But indeed, it might be clearer because you don't have to remember the various kwargs for return types. @EwoutH @rht what are your thoughts?

EwoutH:
Could you add three methods to GroupBy allow more dict-like behavior?

```Python
    def __getitem__(self, key):
        """Allow direct access to groups using the dictionary syntax"""
        return self.groups[key]

    def __len__(self):
        """Allow len() to return the number of groups"""
        return len(self.groups)

    def __getattr__(self, item):
        """
        Delegate attribute access to the internal `groups` dict, 
        allowing for all dictionary methods to be used directly on the GroupBy object.
        """
        return getattr(self.groups, item)
```
`__getattr__` is especially magic, it allows you to do things like `GroupBy.keys()`, `.update()`, `.values()` and any other [dict](https://docs.python.org/3/library/stdtypes.html#dict) method. 

This will allow users to also be able to just threat a groupby object like a dict, which is a know construct to many modellers and doesn't require to learn our API.

EwoutH:
> Where apply applies a function and returns the result. And do applies a function, but returns the AgentSet.

In general, I like methods to have different names if they have different output types or structures.

Let me play with it some more.

(also maybe that's just me, but I really need usage examples to visualise this properly)

quaquel:
> Could you add three methods to GroupBy allow more dict-like behavior?
> 
> ```python
>     def __getitem__(self, key):
>         """Allow direct access to groups using the dictionary syntax"""
>         return self.groups[key]
> 
>     def __len__(self):
>         """Allow len() to return the number of groups"""
>         return len(self.groups)
> 
>     def __getattr__(self, item):
>         """
>         Delegate attribute access to the internal `groups` dict, 
>         allowing for all dictionary methods to be used directly on the GroupBy object.
>         """
>         return getattr(self.groups, item)
> ```
> 
> `__getattr__` is especially magic, it allows you to do things like `GroupBy.keys()`, `.update()`, `.values()` and any other [dict](https://docs.python.org/3/library/stdtypes.html#dict) method.
> 
> This will allow users to also be able to just threat a groupby object like a dict, which is a know construct to many modellers and doesn't require to learn our API.
> 
> **Edit:** Maybe we also want `list` and maybe `del`. It would be nice if we someway just also could allow any function that calls a dict.

I deliberately did not do this. I tried to mimic pandas.AgentSetGroupBy rather than have our GroupBy be effectivley a dict. For example, if we want GroupBy to be a proper dict we would have to write

```python

for group_name, group in some_gouby_instance.items():
    # do something here

```

At the moment, because GroupBy is not a full dict, you can instead do

```python

for group_name, group in some_gouby_instance:
    # do something here

```

Which is exactly how it works in pandas.

EwoutH:
Interesting, thanks for sharing.
> ```python
> for group_name, group in some_gouby_instance:
>     # do something here
> ```

I think as long as we keep `__iter__` we can still do that right, since it overwrites the default dict behavior?

Direct key access with `__getitem__` seems very useful anyways.

quaquel:
> Direct key access with __getitem__ seems very useful anyways.

In line with pandas, this is possible via `get_group`. Checking [DataFrameGroupBy](https://pandas.pydata.org/docs/reference/groupby.html), I noticed that it has a `GroupBy.groups` which returns the underlying dict. would that solve your problem?

EwoutH:
Let me explain my train of thought:
- We're adding more and more stuff to our API, it's getting quite convoluted
- Users have to learn more and more specific details about our inputs, arguments and returns
- How can we simplify that?
- What if it behaved more like standard structures that people know?
- GroupBy is basically just a dict with some polish around it, right?
- What if we could let it behave as much as a dict, a beloved built-in construct basically everyone knows
- Could it be just like a superset of a dict? With a very similar API, with some additional functionality?
- That way, people could start using it like a dict and use more advanced functionality when they get to the limitations of our current approach (in line with scaffolded learning principle @tpike3 mentioned)

So from that perspective, my preferred return type would actually be a dict. But you can't call apply on a dict, so you just need a little magic around it to make a few mesa-specific function work. But personally, I was thinking of staying as close to a dict as possible.

quaquel:
I follow that train of thought, but rather than let it behave like a dict, I want it to behave like pandas DataFrameGroupBy.

Corvince:
> (also maybe that's just me, but I really need usage examples to visualise this properly)

Adopting the example from @quaquel 

> ```python
> # one liner random activation by type
> some_agentset.groupby(type).apply("shuffle", inplace=True, return_results=False).apply("step", return_results=False)
> ```

```python
some_agentset.groupby(type).do("shuffle").do("step").apply(len)
```

This would randomly activate agents and then collect the length of each agent group, e.g. to see how many are left in each group. It would be equal to 

```python
some_agentset.do("shuffle").do("step").groupby(type).apply(len)
```

Another example
```python
some_agentset.do("select", lambda a: a.wealth > 5) # <- modifies some_agentset inline
other_agentset = some_agentset.apply("select", lambda a: a.wealth > 5) # <- returns new agentset
```

quaquel:
> ```python some_agentset.do("select", lambda a: a.wealth > 5) # <- modifies some_agentset inline
>  other_agentset = some_agentset.apply("select", lambda a: a.wealth > 5) # <- returns new agentset```

This won't work. Select is a method on the AgentSet itself, while `do` applies the method/function to each agent in the set.

Corvince:
I just realized that, but you beat me to editing it!

EwoutH:
> but rather than let it behave like a dict, I want it to behave like pandas DataFrameGroupBy.

I'm going to do a little bit of reading up on DataFrameGroupBy.

(but the fact that I need to do reading up on DataFrameGroupBy worries me for our general users)

What are some things you can do when it behaves like a DataFrameGroupBy object that are possible or more elegant compared that it behaves mostly like a dict with an apply function?

EwoutH:
Thanks for the examples @Corvince! The aggregation use-case I'm fully behind.

Here's maybe an interesting thought exercise: What would the least complex / unfamiliar implementation (especially in user API) be to properly facilitate aggregation? Maybe we're making this way too complex.

EwoutH:
For now I have found the following use cases:
- Aggregation into different groups is useful. Group, aggerate, output in some nice format (`{group: agg_value}` dict?)
- Transformation
- Doing something to a specific group: That would use `select().do()`
- Doing something to all groups: Yes, but then you would need to pass the `value` of the groupby also, instead of only the group. Otherwise you can just apply it on all agents at once.
- Doing something to all groups in a specific order of groups (random activation by type). That would indeed need a full `agents.groupby().sort_groups().do()`
  - If you have a predetermined order, you could do
    ```Python
    groups = agents.groupby()
    for group in predetermined_order:
        groups[group].do("step")
    ```
- You might want to take it further, if you want to dynamically want to order your groups on either:
  - The group key.
  - A aggregated value

<hr>

But taking a step back, and looking at the code, I think it's quite clean and a good first step. There are cases to be made for both an AgentSet or a List to be returned as each group value, but I think AgentSet is the best default, because it offers flexibility. When you need more performance you can explore list.

I would really like some syntactic sugar to allow for clean, intuitive dict-like behavior. Personally, I think `__len__` and `__getitem__` are obvious: You want to know how many groups you have or you want to select a specific group. `__getattr__` also sounds very useful, to quickly get the keys or items. How would you do that else?

quaquel:
What about the following next steps?
1. I'll seperate `apply` and `do` along the line suggested by @Corvince but only in the GroupBy helper class. I do think it is a good idea also for AgentSet, but it is a breaking change (can be partially handled with a temporary warning) and might better be done in a seperate PR.
2. I still don't agree with @EwoutH on the dict behavior. First, the iteration as currently implemented is not dict like because it returns group_name and group rather than only group_name as would happen in a normal dict. Second, GroupBy in pandas is really a helper class in chaining calls (groupby, apply, combine). I'll happily add the `groups` attribute that returns the underlying dict. Third, adding dict behavior to this class will come with performance penalties and are really just passed on to the underlying dict anyway. Why not make it easy to return this dict and work on that directly?

EwoutH:
What about an optional return type that returns a dict instead of a group?

quaquel:
> What about an optional return type that returns a dict instead of a group?
> 
> Edit or just a `group_to_dict()` function

We just agree to seperate do and apply to get rid of a kwarg for return type. My solution is just a groups attribute (rather than a function). So

```python

some_dict: dict[Any, Agentset] = some_agentset.groupby(type).groups

```

benchmarks/WolfSheep/__init__.py
EwoutH, line 1: Why is this needed by the way? And why only for wolf-sheep?

mesa/agent.py
```diff
@@ -357,7 +358,83 @@ def random(self) -> Random:
...
+    def __init__(self, groups: dict[Any, list | AgentSet]):
+        self.groups: dict[Any, list | AgentSet] = groups
+
+    def get_group(self, name: Any):
+        """
+        Retrieve the specified group by name.
+        """
+
+        # return group for specified name
+        return self.groups[name]
```
EwoutH, line 415: I think these are the only lines we really disagree on. Let's remove them, you can just cal `GroupBy.groups["some_group"]` if you want to access it right now.

Then we can discuss which convenience method to add later.

EwoutH (commented):
I feel like I'm getting dug in a little and I don't like it. We agree on most things, let's merge those and leave the rest for later.

benchmarks/WolfSheep/__init__.py
quaquel: it should not be there and no idea why it is there. Changes should be confined to agent.py and test_agent.py

EwoutH:
> I would really like some syntactic sugar to allow for clean, intuitive dict-like behavior. Personally, I think `__len__` and `__getitem__` are obvious: You want to know how many groups you have or you want to select a specific group. `get_group()` is something new you have to learn. Just accessing an item like in a dict is obvious, especially since it already iters like a dict.
> 
> `__getattr__` also sounds very useful, to quickly get the keys or items. How would you do that else?

I'm okay with leaving `__getattr__` out of this PR. `.values()` would be weird anyway, you just get back all the agents in some weird lists. `.items()` is covered by `__iter__` and if you want to know the group name, just do `GroupBy.groups.keys()`.

How to retrieve a group we disagree on. You want to do `get_group()`, I think `__getitem__` is more elegant. Let's leave both out and discuss this separately.

I still think `__len__` would be very useful. I just grouped, but how many groups do I now have? Can be useful for statistics, for determining things like bins for plotting, etc.

quaquel:
> I still think __len__ would be very useful. I just grouped, but how many groups do I now have? Can be useful for statistics, for determining things like bins for plotting, etc.

Fair enough

quaquel:
I have added `__len__`, seperated `apply` and `do`, and removed `get_groups`. Unittests have also been updated. This is ready for a last review.

Corvince (approved):

EwoutH:
Thanks for the work, both this PR and the group split ended up great!

PR #2223, opened 2024-08-18
model: Automatically increase `step` counter

### LINKED ISSUE #2222: Automatic time and step incrementing
_This issue is a spin-off of the discussion in https://github.com/projectmesa/mesa/pull/1942#issuecomment-1903386624_

### Background

Currently, Mesa models require users to explicitly increase the time and step counters using `_advance_time()`. This approach can be unintuitive and prone to errors, especially for new users. It's also weird to have to use a private method to advance the time.

### Motivation

We want to simplify the model creation process and reduce the potential for errors by automatically incrementing time and step counters. This change would make Mesa more user-friendly.

Simultaneously, the time-increase should be able to be overwritten if needed.

### API proposal
I would propose the following user interface:
```Python
class Model:
    ...
    def step(self, time=1, step=1):
        ...
```
The default is `time=1` and `step=1`. Users can change these numbers if they want, but they don't need to. If you want to turn both off, you can just say `step(self, time=False, step=False)`, because in Python, the `False` is interpreted as `0`.

Then we need to have one giant final battle if the time and step increases at the beginning of the step or the end of the step.

Curious what everybody thinks. If anyone has the golden bullet on an implementation idea, please share!

EwoutH:
This pull request adds automatically increasing of `steps` counter in the Mesa model, even if the users overwrites the `Model.step()` method. It also removes the `_time` variable.

#### Background
Previously, Mesa required users to manually increment the step counters using the `_advance_time()` method, or by modifying the `model._time` and `model._step` values directly. This use of an private method was non-ideal.

- Closes #2222
- Initial discussion: https://github.com/projectmesa/mesa/pull/1942#issuecomment-1903386624

#### Changes
`steps` is now incremented automatically within the `Model` class itself. Each call to the `step()` method will increment the `steps` counter by 1.

The `_time` counter is removed (for motivation, read the discussion below).

#### Implementation details
The core change involves wrapping the user-defined `step` method with `_wrapped_step` which handles the incrementing process **before** executing the user's step logic. This ensures that all increment operations are centrally managed and transparent to the user.

I choose to increment the time before the user step, in the spirit of "a new step" (or a new day) has begun. Then the model does things during the step, and.
```Python
def step():
    # Time is increased automatically at the beginning of the step
    self.agents.shuffle().do("step")  # Do things during the step
    self.datacollector.collect()      # Collect data at the end of the step
```

Here's the updated version of your examples with the initial whitespaces removed from the code blocks:

#### Example Usage
You don't need to do anything to use this features. You can access steps at any moment.
```python
class MyModel(Model):
    def step(self):
        # User logic here

my_model = MyModel()
my_model.step()
```

#### Todo:
- [x] Tests pass
- [x] Check and if needed, update examples
  - [x] https://github.com/projectmesa/mesa-examples/pull/162
- [x] Add tests
- ~Check how this interacts with the experimental DEVS simulator~

#### Resolve:
- https://github.com/projectmesa/mesa/discussions/2227
  - [x] Resolution: `steps` is automatically increased by exactly `1`.
- https://github.com/projectmesa/mesa/discussions/2228
  - [x] Resolution: `_time` is removed.
- https://github.com/projectmesa/mesa/discussions/2229
  - [x] Resolution: 1) increase counter, 2) perform step, 3) collect data
- https://github.com/projectmesa/mesa/discussions/2230
  - [x] Resolution: We claim `steps`. Further discussion can continue.
- https://github.com/projectmesa/mesa/discussions/2231
  - [x] Resolution: Needle is moving towards removing them. Other PR.

### DIFF
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -17,8 +17,6 @@ from typing import Any
 from mesa.agent import Agent, AgentSet
 from mesa.datacollection import DataCollector
 
-TimeT = float | int
-
 
 class Model:
     """Base class for models in the Mesa ABM library.
@@ -35,6 +33,8 @@ class Model:
     Properties:
         agents: An AgentSet containing all agents in the model
         agent_types: A list of different agent types present in the model.
+        steps: An integer representing the number of steps the model has taken.
+               It increases automatically at the start of each step() call.
 
     Methods:
         get_agents_of_type: Returns an AgentSet of agents of the specified type.
@@ -62,10 +62,6 @@ class Model:
             # advance.
             obj._seed = random.random()
         obj.random = random.Random(obj._seed)
-
-        # TODO: Remove these 2 lines just before Mesa 3.0
-        obj._steps = 0
-        obj._time = 0
         return obj
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:
@@ -77,25 +73,20 @@ class Model:
         self.running = True
         self.schedule = None
         self.current_id = 0
+        self.steps: int = 0
 
         self._setup_agent_registration()
 
-        self._steps: int = 0
-        self._time: TimeT = 0  # the model's clock
-
         # Wrap the user-defined step method
-        if hasattr(self, "step"):
-            self._user_step = self.step
-            self.step = self._wrapped_step
+        self._user_step = self.step
+        self.step = self._wrapped_step
 
     def _wrapped_step(self, *args: Any, **kwargs: Any) -> None:
         """Automatically increments time and steps after calling the user's step method."""
         # Automatically increment time and step counters
-        self._time += kwargs.get("time", 1)
-        self._steps += kwargs.get("step", 1)
+        self.steps += 1
         # Call the original user-defined step method
-        if self._user_step:
-            self._user_step(*args, **kwargs)
+        self._user_step(*args, **kwargs)
 
     @property
     def agents(self) -> AgentSet:
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -83,6 +83,20 @@ class Model:
         self._steps: int = 0
         self._time: TimeT = 0  # the model's clock
 
+        # Wrap the user-defined step method
+        if hasattr(self, "step"):
+            self._user_step = self.step
+            self.step = self._wrapped_step
+
+    def _wrapped_step(self, *args: Any, **kwargs: Any) -> None:
+        """Automatically increments time and steps after calling the user's step method."""
+        # Automatically increment time and step counters
+        self._time += kwargs.get("time", 1)
+        self._steps += kwargs.get("step", 1)
+        # Call the original user-defined step method
+        if self._user_step:
+            self._user_step(*args, **kwargs)
+
     @property
     def agents(self) -> AgentSet:
         """Provides an AgentSet of all agents in the model, combining agents from all types."""
@@ -179,11 +193,6 @@ class Model:
 
     def step(self) -> None:
         """A single step. Fill in here."""
-
-    def _advance_time(self, deltat: TimeT = 1):
-        """Increment the model's steps counter and clock."""
-        self._steps += 1
-        self._time += deltat
 
     def next_id(self) -> int:
         """Return the next unique ID for agents, increment current_id"""
--- a/tests/test_batch_run.py
+++ b/tests/test_batch_run.py
@@ -83,8 +83,8 @@ class MockModel(Model):
         return 42
 
     def step(self):
-        self.datacollector.collect(self)
         self.schedule.step()
+        self.datacollector.collect(self)
 
 
 def test_batch_run():
--- a/benchmarks/BoltzmannWealth/boltzmann_wealth.py
+++ b/benchmarks/BoltzmannWealth/boltzmann_wealth.py
@@ -38,7 +38,6 @@ class BoltzmannWealth(mesa.Model):
         self.datacollector.collect(self)
 
     def step(self):
-        self._advance_time()
         self.agents.shuffle().do("step")
         # collect data
         self.datacollector.collect(self)
--- a/mesa/batchrunner.py
+++ b/mesa/batchrunner.py
@@ -132,14 +132,14 @@ def _model_run_func(
     """
     run_id, iteration, kwargs = run
     model = model_cls(**kwargs)
-    while model.running and model._steps <= max_steps:
+    while model.running and model.steps <= max_steps:
         model.step()
 
     data = []
 
-    steps = list(range(0, model._steps, data_collection_period))
-    if not steps or steps[-1] != model._steps - 1:
-        steps.append(model._steps - 1)
+    steps = list(range(0, model.steps, data_collection_period))
+    if not steps or steps[-1] != model.steps - 1:
+        steps.append(model.steps - 1)
 
     for step in steps:
         model_data, all_agents_data = _collect_data(model, step)
--- a/mesa/datacollection.py
+++ b/mesa/datacollection.py
@@ -180,7 +180,7 @@ class DataCollector:
         rep_funcs = self.agent_reporters.values()
 
         def get_reports(agent):
-            _prefix = (agent.model._steps, agent.unique_id)
+            _prefix = (agent.model.steps, agent.unique_id)
             reports = tuple(rep(agent) for rep in rep_funcs)
             return _prefix + reports
 
@@ -216,7 +216,7 @@ class DataCollector:
 
         if self.agent_reporters:
             agent_records = self._record_agents(model)
-            self._agent_records[model._steps] = list(agent_records)
+            self._agent_records[model.steps] = list(agent_records)
 
     def add_table_row(self, table_name, row, ignore_missing=False):
         """Add a row dictionary to a specific table.
--- a/mesa/time.py
+++ b/mesa/time.py
@@ -69,8 +69,6 @@ class BaseScheduler:
         self.model = model
         self.steps = 0
         self.time: TimeT = 0
-        self._original_step = self.step
-        self.step = self._wrapped_step
 
         if agents is None:
             agents = []
@@ -112,11 +110,6 @@ class BaseScheduler:
         self.do_each("step")
         self.steps += 1
         self.time += 1
-
-    def _wrapped_step(self):
-        """Wrapper for the step method to include time and step updating."""
-        self._original_step()
-        self.model._advance_time()
 
     def get_agent_count(self) -> int:
         """Returns the current number of agents in the queue."""
--- a/mesa/visualization/solara_viz.py
+++ b/mesa/visualization/solara_viz.py
@@ -244,7 +244,7 @@ def ModelController(model, play_interval, current_step, reset_counter):
         """Advance the model by one step."""
         model.step()
         previous_step.value = current_step.value
-        current_step.value = model._steps
+        current_step.value = model.steps
 
     def do_play():
         """Run the model continuously."""
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -66,3 +66,4 @@ class TestExamples(unittest.TestCase):
                 model = model_class()
                 for _ in range(10):
                     model.step()
+                self.assertEqual(model.steps, 10)
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -9,7 +9,9 @@ def test_model_set_up():
     assert model.current_id == 0
     assert model.current_id + 1 == model.next_id()
     assert model.current_id == 1
+    assert model.steps == 0
     model.step()
+    assert model.steps == 1
 
 
 def test_running():
@@ -18,12 +20,12 @@ def test_running():
 
         def step(self):
             """Increase steps until 10."""
-            self.steps += 1
             if self.steps == 10:
                 self.running = False
 
     model = TestModel()
     model.run_model()
+    assert model.steps == 10
 
 
 def test_seed(seed=23):


mesa/model.py
```diff
@@ -77,6 +77,20 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         self._steps: int = 0
         self._time: TimeT = 0  # the model's clock
 
+        # Wrap the user-defined step method
+        if hasattr(self, "step"):
```
rht, line 81: What if the user defines their step with a different name, e.g. in the simultaneous activation?

EwoutH: Good edge case! I think when using a scheduler, you're also mainly using the scheduler's time, so `model.schedule.time`. And as long as one of the functions is a `step()`, the model time will also be increased.

Personally I'm in favor of moving away from the schedulers, but this PR won't break that.

quaquel:
I'll try to look at the code later today, but just some quick first questions/thoughts/concerns.

1. Do we want to maintain the distinction between step and time? Is this distinction not confusing to a novice user?
2. Should time not be handled by the simulator?

EwoutH:
> * Do we want to maintain the distinction between step and time? Is this distinction not confusing to a novice user?

I think both can be useful in some cases. Steps are just the number of times `model.step()` is called (by default). Time can be anything you want.

For example, if I have a model that runs from 6 o'clock in the morning to 23 o'clock in the evening, with 5 minute steps, I might do:

```Python
class MyModel(Model):
    def __init__(self):
        ...
        self._time = 8
    def step(self, time=1/12):
        ...
```
Then I have a human-readable value of the time of day, while the steps can be useful for plotting, datacollection and debugging. So one is human readable and one is an uninterrupted sequence of integers.

Of course, we could aim our documentation and tutorials to mainly use one. 

> * Should time not be handled by the simulator?

In my opinion, a model should be runnable without too many additional fluff. A simulator or scheduler should not be required for a simple ABM (and https://github.com/projectmesa/mesa-examples/pull/161 proves those aren't needed for the vast majority of models).

mesa/model.py
EwoutH: > If the model doesn't have a step, it is more informative to raise an error with an informative message.

If you want to use your own process without using Mesa's internal clock, it's perfectly valid. I think we can cover this with documentation, and this will be less and less a problem as we move away from schedulers.

I added an example of staged time increase to the first post.

> Alternatively, bumping the step & time whenever model.agents.do is called, is another alternative.

The problem is, how do you handle multiple agents.do calls? Or agents.do to a partial agentset? Or with a select in between?

quaquel: > > Alternatively, bumping the step & time whenever model.agents.do is called, is another alternative.

> The problem is, how do you handle multiple agents.do calls? Or agents.do to a partial agentset? Or with a select in between?

If we want to track the number of steps, it should be tied to `model.step`. It's a bad idea to tie it to the AgentSet. Conceptually, this would mix up two completely different things. Which, as @EwoutH indicates, gives rise to all kinds of knock-on problems.

quaquel:
> In my opinion, a model should be runnable without too many additional fluff. A simulator or scheduler should not be required for a simple ABM

Fair enough, but then I would keep support for tracking time / the number of ticks also minimal. What about having full time support in the simulator and only have number of ticks/steps in the model class?

EwoutH:
To do stages / simultaneous activation it can be nice to separate steps and time (see example in PR start). Also steps is just a counter and time can be used for a lot more.

Maybe the experimental `ABMSimulator` is redundant with this PR, and we only need to keep the `DEVSSimulator`

quaquel:
> Maybe the experimental ABMSimulator is redundant with this PR, and we only need to keep the DEVSSimulator

No, `ABMSimulator` supports event scheduling while also having fixed time advancement (i.e., ABM style ticks).

quaquel:
I have been thinking about this PR a bit more. I think having an automatic counter for the number of times `model.step` is called is useful. 

However, I would keep the other stuff like time, running automatically for a given number of steps, etc. for the `ABMSimulator` which already supports most of this via `run_for` and `run_until`. Being able to increment steps with more than one, even though `step` is called only once, as is possible with this API, really makes no sense to me and is confusing. Likewise, if a model needs to translate between steps and wall clock time, why not simply multiply with a fixed constant? Making models where two ticks represent different periods of wall clock time is conceptually confusing and invites bad modeling. 

Splitting the functionality between tracking the number of times steps is called in the model and having a separate class for more advanced time management is my preferred way forward. This makes it possible for a user to keep building simple models as currently done (via a `for` loop as is current practice), but if you want more advanced control over time and running, we have an optional class that can be used.

EwoutH:
Thanks for your insights. I will be thinking about them a bit more, but for now, I see it this way:
- Steps is the most basic variant. It's like ticks in NetLogo. Just the number. Ideal for simple models that loop model.step().
  - I agree that maybe we need to leave it fixed to 1 (otherwise, what harm does it do if people find a good reason to change is?)
- Time is the intermediate variant. You can adjust it, conditionally skip it or increase it, let it depend on model or agent variables. It offers a bit of flexibility when you need just a bit more control, or want a human readable value (like in the example).
- The ABMSimulator is where you do things between ticks, but still want to have ticks and a model step as construct. Just to give some certainty.
- The DEVS scheduler you can do whatever you want. Go wild.

The `ABMSimulator` sits (for me) now in a weird place. I accidentally started using the DEVSimulator, before I knew I also wanted a model.step, and getting a model.step back was one or two lines of code. So I don't know if we really need it.

Looking at code complexity, adding and tracking a `time` aside from `step` is exactly _one_ line of code. The ABMSimulator is about 90, with additional code you need to understand, document, write tutorial for, etc.

So, my current standpoint:
- Maybe fix steps to always be 1. But what's the harm from letting it be a nice little hook to mount in to?
- Keep the time. It adds very little complexity, and I think the examples above show it can sometimes be useful.
- Consider removing the `ABMSimulator`, and see if we can just document how to use the `DEVSimulator` with an Model step. In advanced use cases, implicit is better than explicit.
- Document that if you start to use a simulator, you should use the simulator time.

rht:
`steps` is bread and butter: used in the batch runner and the visualization (to detect if user has reset the simulation), so it can't be fixed to 1 by default.

EwoutH:
I didn't want to run too far ahead, but I was thinking of allowing the index of the DataCollector to be changeable. Default could be `steps` or `time`, but you could input anything you did like. 

And maybe the same for visualisation.

(with fixed to 1, I meant to fix it to increase always by 1)

quaquel:
> Time is the intermediate variant. You can adjust it, conditionally skip it or increase it, let it depend on model or agent variables. It offers a bit of flexibility when you need just a bit more control, or want a human readable value (like in the example)

Why skip it or increase it in a non-fixed way? from a modeling point of view, this makes no sense to me. If "for readability", why should this be supported by MESA?

> Consider removing the ABMSimulator, and see if we can just document how to use the DEVSimulator with an Model step. In advanced use cases, implicit is better than explicit.

The ABMSimulator can indeed be done by DEVSSimulator as well. The main use case for me is fixed time advancement (so increment by 1 and automagically calls model.step each tick) while having integer-based event scheduling. Conceptually, it is a real devs-ABM hybrid.

EwoutH:
You can do cute stuff like conditionally modifying your time resolution, to perform more ticks at certain moments:
```Python
    def step(self, time=False):
        # Increase volatility and decrease time increment during opening and closing hours
        if 9 <= self._time < 10 or 15 <= self._time < 16:
            self.volatility = 0.03
            self._time += timedelta(minutes=5)
        else:
            self.volatility = 0.01
            self._time += timedelta(minutes=30)

        # Update stock prices
        for stock in self.agents:
            stock.update_price(self.volatility)
```
If you allow datetime, you can even go a bit further:
```Python
        # If we've passed 4:00 PM, move to 9:00 AM the next day
        if self._time >= 16:
            next_day = self._time.date() + timedelta(days=1)
            self._time = datetime.combine(next_day, datetime.min.time()) + timedelta(hours=9)
```
And imagine you have external data lookups that use the time. Or in your data collector you can aggerate by hour, while still having more resolution available.

It's basically just an additional built-in counter you can use before having to jump immediately to discrete even scheduling.

quaquel:
> You can do cute stuff like conditionally modifying your time resolution, to perform more ticks at certain moments:

That's not cute stuff, but bad modeling. It breaks discrete time advancement, which is a core principle behind ABMs. If a model requires variable time steps, you should use DEVS and carefully check all code to ensure time units are handled correctly.

Also, and separately, is this use case so common that it should be supported in a core class of MESA?

rht:
> Also, and separately, is this use case so common that it should be supported in a core class of MESA?

If you looked at the PR's timeline, the time attribute was added months before the DEVS PR https://github.com/projectmesa/mesa/pulls?q=is%3Apr+discrete+event+is%3Aclosed. At the time, it was the only official way of having time being tracked. But again, even if the DEVS did not exist, the time attribute doesn't necessarily have to be there by default: for users who need it, they can always define their time attribute as needed, and it would need only a few lines of code for the feature.

EwoutH:
> That's not cute stuff, but bad modeling. It breaks discrete time advancement, which is a core principle behind ABMs.

And that's why it's not the default. I agree that it's not bad practice for most cases. But if an users finds a novel case or use, why deny the flexibility?

<hr>

I'm going to summarize my points:
- By default, model.time is increased with a fixed amount of 1 each step.
- model.time can be used as a human-readable values and potentially Datetime object
  - This allows data lookups in time-based lookup tables, decision logic based on a human-readable value, etc.
- model.time can be used to acknowledge that a step has different sub-steps. It can be increased by a custom (but fixed) amount if you want divide you steps in different substeps
  - This is exactly how [StagedActivation](https://mesa.readthedocs.io/en/stable/apis/time.html#mesa.time.StagedActivation) now works:
     > The scheduler also tracks steps and time separately, allowing fractional time increments based on the number of stages. Time advances in fractional increments of 1 / (# of stages), meaning that 1 step = 1 unit of time.
  ```Python
  class MyModel(Model):
      def step(self, time=0.25):
          model.agents.do("morning")
          model._time += 0.25
          model.agents.do("afternoon")
          model._time += 0.25
          model.agents.do("evening")
          model._time += 0.25
          model.agents.do("night")
  
  my_model = MyModel()
  my_model.step()
  ```
   - Which means we can even keep track of both the full steps and the sub-steps, without having to do any rounding or flooring or sorts.
- It gives users an additional built-in variable to play with, offering a nice hook to tie into, like a proper library would.
- It adds very little complexity (a _single_ line of code).

quaquel:
I think my main objection is not to time attribute perse but to the use of time and step as optional keyword arguments to model.step. This is a new feature, might break existing models, and gives rise to a lot of the problems I have been alluding to. If users want to do something like this in their custom models, they can do so. But why support it within MESA itself given the conceptual issues I have been highlighting? 

Staged activation has its own internal time attribute. Which is conceptually strange. As argued before, there should be a single truth for time. I have to think about substeps with staged activation. I am unsure about how to handle that properly but seperating step and time might indeed be defendable for this.

EwoutH:
Would removing the `step=1` argument from the Model.step be an acceptable compromise for you, and then it will always be increased by `1`?

<hr>

Fun fact: I have a DEVS model in development right now where a steps counter would be very useful, since I still have a model step and want to do something once every N steps, while the remainder of the model is DEVS controlled.

EwoutH:
Wait there is one use case we haven't discussed: Some people might want to step at another place in the step, than only at the beginning of the step. Thereful it would be useful to turn the automatic step increase of.

We could make step a bool, allowing it either to be `False` (interpreted as `0`) or `True` (interpreted as `1`).

But in general, I'm for a flexible library with sensible defaults. `time=1, step=1` are very sensible defaults, while also being very flexible.

quaquel:
> Fun fact: I have a DEVS model in development right now where a steps counter would be very useful, since I still have a model step and want to do something once every N steps, while the remainder of the model is DEVS controlled.

use devs_simulator.time? or swicht to the ABMSimulator and use its time attribute should fix this for you.

> Wait there is one use case we haven't discussed: Some people might want to step at another place in the step, than only at the beginning of the step. 

why? I think with clear documentation (so allways at the start or allways at the end) any use case can easily be accomodated.

EwoutH:
I think at this point we disagree on a philosophical level what a simulation library should offer.

@projectmesa/maintainers, I would love some fresh perspectives.

wang-boyu:
As an alternative example, in our agents-and-networks gis example, we keep a clock manually : https://github.com/projectmesa/mesa-examples/blob/0ebc4d11711eb59480942d24dd13e9a744744704/gis/agents_and_networks/src/model/model.py#L215

> - By default, model.time is increased with a fixed amount of 1 each step.

This default behavior seems a bit redundant to me. TBH I wasn't really aware of the `_time` attribute and always used `_steps`. Unitless `time` is essentially the same thing as `step` (or am I missing something?)

If a clock is really really needed, we probably need something more dedicated, like what AnyLogic offers: https://www.anylogic.com/upload/books/new-big-book/16-model-time-date-and-calendar.pdf which is also mentioned above:

> We might allow time to be a full Datetime object. Can imagine some scenarios in which that would be very handy.

But this may not be needed by all models, especially those simple models. So it may be better to be optional.

EwoutH:
> So it may be better to be optional.

Let me be perfectly clear: It's 100% optional. You can just do `model.step()`.

quaquel:
> I think at this point we disagree on a philosophical level what a simulation library should offer.

My last question was really about trying to understand the issue. So not philosopical at al. 

In line with @wang-boyu, however, I am in favor of simply counting steps and leave detailed time management to an optional more advanced feature (possibly to be integrated into the simulator classes which offer much of what is described in the book chapter).

tests/test_batch_run.py
```diff
@@ -83,8 +83,8 @@ def get_local_model_param(self):
         return 42
 
     def step(self):
-        self.datacollector.collect(self)
         self.schedule.step()
+        self.datacollector.collect(self)
```
rht, line 87: Why is this moved after the schedule step?

mesa/model.py
rht: > If you want to use your own process without using Mesa's internal clock, it's perfectly valid. I think we can cover this with documentation, and this will be less and less a problem as we move away from schedulers.

`steps`'s need is hardcoded it the batch runner and visualization. Currently, they are not optional.

tests/test_batch_run.py
EwoutH: Great question. That's because this way data is collected after agents have done something, and thus most importantly, the last step after agents have done something.

If you want to collect before the first step (at time = 0), you can also add `self.datacollector.collect(self)` once in the init.

All example models already do this, this was the only test case that needed to be adjusted.

See for further details the _Implementation details_ section of the PR start.

rht:
> I choose to increment the time before the user step, in the spirit of "a new step" (or a new day) has begun. Then the model does things during the step, and.

But that means the `steps` start from 1 instead of 0. There is no action happening at `steps` 0, the zeroth-day. Wouldn't this introduce a breaking change in other area, e.g. [batchrunner](https://github.com/projectmesa/mesa/blob/3cf1b76ca3a157c8a064b60130755a86e73718be/mesa/batchrunner.py#L135)?

EwoutH:
> There is no action happening at steps 0.

The whole __init__ is happening at step 0. If you have N steps, step 0 is the init and step 1 to N are the steps. I think this would be a good convention for Mesa, also something what flexible and easily explainable is.

quaquel:
> Let me be perfectly clear: It's 100% optional.

No it is not. `step` and `time` are now reserved keyword arguments used by `Model.step`. Sorry to be annoying about this. But this PR does more than just automatically increase time and step. And this additional stuff I have issues with. As well as that this PR reopens some earlier discussions about time/clock management in MESA that have never been fully resolved.

quaquel:
>What I meant with philosophical differences, is that I think [convention over configuration](https://en.wikipedia.org/wiki/Convention_over_configuration) is in general a good principle for Mesa to follow. Why? Because it offers two things that perfectly fits our mixed target audience.

Ok, clear and I don't disagree with you on the principle. I think the disagreement is rather how to go about it in this particular case. A lot is handled here through keyword arguments. My intuition, in contrast, is to have basic step counting as default and have classes for more advanced time management.

rht:
> > There is no action happening at steps 0.
> 
> The whole `__init__` is happening at step 0. If you have N steps, step 0 is the init and step 1 to N are the steps. I think this would be a good convention for Mesa, also something what flexible and easily explainable is.

Regardless of my opinion on this, this means it is a breaking change to the existing convention. And would require documentation in the release note. The `max_steps` in the batch runner stopping condition needs to be augmented by 1, and so on. This is a lot of change for a PR meant to only automatically increase the steps count.

EwoutH:
Okay, this PR is indeed opening a lot of old, partially unresolved issues and discussions
1. Increasing the step counter automatically
2. Does we want an time/clock in Mesa
3. The order of which things happen in a step and what counts as the first step
4. How to handle (private) / reserved variable in Mesa, considering we get subclassed all the time
5. What do we want with schedulers in future (if at all)

Shall we put this PR on pause and open a separate discussion topics on each of them?

On a bit of a personal note: This issue was open for half a year, and I was so proud and excited I finally found a solution and everything clicked in a very nice and elegant way yesterday. Just a few lines of code, very pretty API, sensible defaults, yet very flexible. Finally properly standardized step order and counting, complexity removed from user flexibility, it all just clicked together.

Never celebrate your victory early.

EwoutH:
Okay, here we go: In a brand new category [🧭Future of Mesa](https://github.com/projectmesa/mesa/discussions/categories/future-of-mesa?discussions_q=category%3A%22Future+of+Mesa%22+) I opened 5 beautiful new, totally hand-written discussions where we can explore these topics and hopefully come to solutions.

- https://github.com/projectmesa/mesa/discussions/2227
- https://github.com/projectmesa/mesa/discussions/2228
- https://github.com/projectmesa/mesa/discussions/2229
- https://github.com/projectmesa/mesa/discussions/2230
- https://github.com/projectmesa/mesa/discussions/2231

tpike3:
@EwoutH thanks for pointing this out. Below are my thoughts for what they are worth. 

If I try and boil this discussion down in my own understanding. I would say it revolves around what features are obviously present and what features are more advanced. I think in general this is also very relevant to `datacollector` and is critical for how Mesa matures over time.   

My inclination then would be to have "simple Mesa" up front. This is the step function as the basic method. Then as users get more familiar they can start digging deeper and add `clock time`

I would then generalize this approach as a core dynamic of Mesa. The goal being one is able to rapidly build models with "beginner knowledge", but Mesa then has the capability for advanced models as users become more proficient. 

Generalizing to the  `datacollector` discussion. Maybe by default `datacollector` collects everything or only the changes as I interpreted @quaquel references to [psygnal](https://psygnal.readthedocs.io/en/latest/). Then as users get more advanced models `datacollector` is setup so users can begin to customize collection features. I know this was a big focus at George Mason for awhile (maybe still is) as they were doing 20 million agent models and so only wanted specific agent/agent group or other types of data. Similarly, as we build out Mesa and put some more performant backend written in other languages like Rust or have an HPC or AWS variants then users can grow to these more advanced options. 

My core argument for this approach would be learning theory.  Short version, humans can't learn everything all at once so you do scaffolded learning. Scaffolded learning is like nesting dolls, where you go from simple to more advanced as the initial ideas become the scaffolding on which one can interpret and internalize more advanced ideas.

EwoutH:
> My core argument for this approach would be learning theory. Short version, humans can't learn everything all at once so you do scaffolded learning.

I like this concept and I agree it fits Mesa well. I think it's also in line with [convention over configuration](https://en.wikipedia.org/wiki/Convention_over_configuration): At first, you can use the sensible defaults without having to think about them, and then as you grow as a modeller, you can start playing changing these defaults in situations where it's useful.

I do some bouldering, and the nicest feeling is if you think "I think they placed a grip there", reach blindly, and it's exactly there. That's what Mesa should ideally be: "I think I would like to change something here", and then there's a nice way to do it already in place.

DrEntropy:
As a newcomer to mesa, this seems to provide less surprising behavior to me than using the wrapped step in scheduler, especially for models that don't use a wrapped step.

EwoutH:
Thanks for the feedback @DrEntropy! Do you have a preference at which point during the step the step counter is increased?
- https://github.com/projectmesa/mesa/discussions/2229

Corvince:
I agree that this PR is doing everything and possibly even more. I strongly disagree that "step" size should be set to everything but 1, just by definition. But lets leave this discussion. I think we should merge as is

Corvince:
Thanks for all the work and the lengthy discussions @EwoutH !!

Corvince (approved):

EwoutH:
Thanks. I'm okay with making `step` a Boolean, so it's either 1 or 0. Would you think that's better than the current int?

EwoutH:
I'm going to wait for at least a second maintainer to approve, because I don't want a disputed PR to be merged.

EwoutH:
Another option could be to just throw a general `advance_time` flag in, instead of the seperate `step` and `time` flags, which disables both time increases.

At the very minimum `time` should be able to be disactivated, otherwise you can't do [StagedActivation](https://mesa.readthedocs.io/en/stable/apis/time.html#mesa.time.StagedActivation).

rht (approved):

rht:
The examples being broken is urgent, so I approved this PR.

EwoutH:
Okay there isn't consensus on the [order of time stepping](https://github.com/projectmesa/mesa/discussions/2229#discussioncomment-10418845), which is the one fundamental assumption this PR makes, so I'm closing this PR.

I will be reverting the example models back to the schedulers so that they work anyway.

No blame to anyone, everyone has made good points and advocated their thoughts fairly. We were just not ready for this PR, and I failed to recognize that.

I have said everything about time that I wanted to say, you all are smart enough to reach good consensus together.

Corvince:
I don't completely follow you on closing this or after me and @rht already approved it. The ordering topic seems to be very much drifting towards a consensus, so I think we are moving in a good direction. 

As I read this pr it doesn't change anything at all, so why not merge this and then decide and make further prs?

EwoutH:
@Corvince: rht mentioned that he was fine with increasing the step count at the beginning of the step, but not the time. This PR increases the time also at the beginning of the step. That's what I meant with:

> there isn't consensus on the order of the step, which is the one fundamental assumption this PR makes

Because this PR assumes that both step and time are increased at the beginning of the step.

<hr>

Reverting the examples back to the schedulers, so they work again:

 - https://github.com/projectmesa/mesa-examples/pull/170

rht:
> @Corvince: rht mentioned that he was fine with increasing the step count at the beginning of the step, but not the time. This PR increases the time also at the beginning of the step.

That does not affect the examples at all, because they don't use the time attribute. This PR is fine as it is.

EwoutH:
This PR was closed due to a few open point. Those now have been resolved in the 5 discussions above and in the call at https://github.com/projectmesa/mesa/discussions/2247. The conclusions of that call can be found in the discussion post.

@quaquel will attempt to modify this PR so that it uses a decorator with keyword arguments to wrap stuff around the user step, instead of copying the inherited step with keyword arguments.

quaquel:
I have made some progress on this, but it still needs further testing. I have done
* moved `_wrapped_step` into a separate function that can be used as a decorator
* Made it possible to parameterize the decorator so you can control stepsize and time_delta/timestep
* moved the automatic decorating into `Model.__new__` because it needs to be done at class creation but before class instantiation
* Add a check to see if `Model.step` is already wrapped/decorated to avoid doing it more than once. 
* this is all done in a way that should ensure that the docstring still works

More testing is required, and naming needs work. But the following API works in initial testing

```python

class myModel(mesa.model.Model):
    def step(self):
        # my step method, not wrapped, so wrapping happens automagically


class myModel(mesa.model.Model):
    @mesa.model.parameterized_step_decorator(stepsize=5)
    def step(self):
        # my step method, not wrapped, so wrapping happens automagically


class myModel(mesa.model.Model):
    @mesa.model.parameterized_step_decorator(time_delta=0.25)
    def step(self):
        # my step method, not wrapped, so wrapping happens automagically

```

rht:
For the decorator name, I find `tick` (as suggested by @DrEntropy) to be short and descriptive. Maybe it should be `@mesa.tick`. An alternative would be `@mesa.clock`.

EwoutH:
Thanks for working on this. It's great that in this way we can change step parameters and can move forward with automatic timekeeping.

Personally, I still find the keyword arguments (as originally proposed in this PR) that turned things in the wrapped step on and off way cleaner.

```Python
    @mesa.model.parameterized_step_decorator(stepsize=5)
    def step(self):
```
```Python
    def step(self, stepsize=5):
```

But I can live with this, functionally it's the same.

DrEntropy:
I wonder if there could be an additional keyword argument that would disable the stepping altogether?  I can imagine edge cases where people want to increment the step and time manual (it might change from step to step).  
Sending in `stepsize=0, time_delta=0` would be equivalent, but it does seem unclean to keep adding 0's each time step.  If I could pass in `disable = True` or something, then you could just skip the wrapping,  and perhaps use @functools.wraps()  so that `step.___wrapped__` still gets set.

quaquel:
> For the decorator name, I find tick (as suggested by @DrEntropy) to be short and descriptive. Maybe it should be @mesa.tick. An alternative would be @mesa.clock.

Thanks for the suggestions. I am thinking of something like `advance_clock`, because the decorator does something so it should have a verb in its name.  However, if in the future we want to expand this decorator to also do `datacollecter.collect` by default as has been suggested in the conversation on the future of datacollection, just  `advance_clock` might not cover everything that is triggered by this decorator. Since I hate the NetLogo/logo language with a vengeance, I don't like the smell of `tick`.

> I wonder if there could be an additional keyword argument that would disable the stepping altogether? I can imagine edge cases where people want to increment the step and time manual (it might change from step to step).

A few thoughts:
1. Anyone who wants to have variable time steps is breaking the fundamental principle of incremental time advancement, which is at the core of classic ABMs. The resulting model is nonsensical unless great care is taken to ensure any time-dependent variables and parameters account for the changing timestep. If you want to do something like that, why not use DEVS?
2. The use case is possible, as you indicated already, although you have the minor overhead of incrementing by 0.
3. the code becomes convoluted quickly. What if you set stepsize to 0 but time_delta to 0.25, or vice versa?
4. Currently, this decorator only interacts with private model attributes that are not publicly exposed. Users who want to mess with internal/protected attributes can do so at their own peril.

DrEntropy:
>  I am thinking of something like `advance_clock`, because the decorator does something so it should have a verb in its name. 

Maybe `@mesa.manage_step`  ? Meh...   hopefully something awesome will occur to someone :) 

>  Since I hate the NetLogo/logo language with a vengeance, I don't like the smell of `tick`.

I am with you on that one.     

> > I wonder if there could be an additional keyword argument that would disable the stepping altogether? I can imagine edge cases where people want to increment the step and time manual (it might change from step to step).
> 
> A few thoughts:
> 
> 1. Anyone who wants to have variable time steps is breaking the fundamental principle of incremental time advancement, which is at the core of classic ABMs. 

I don't understand why one even has a time variable if it is just going to be fixed.  Just track steps, and save $\Delta t$ somewhere (or leave it to the user) and compute elapsed time on the fly when needed.  I would opine this extra variable should just dropped and handled by the user.

Further, a *fixed* 'step' size can be handled the same way.  As @Corvince  suggested, steps can be defined as the number of times 'model.step' was called, in which case `stepsise` should just also be 1.    If the user wants a different $\Delta \text{step}$ they can multiple the number of steps by $\Delta \text{step}$ at their leisure.   

My point is, if all you are doing is adding a fixed number to two variables, you really only need one variable and saved 'deltas'.

EwoutH:
You know what, it's fine. I'm convinced.

I'm finishing this PR, I will create a backup branch of the current state.

rht (approved):

Corvince (approved):

EwoutH:
There we go. Mesa is now incrementing time automatically.

After these two weeks I can conclude this effort with the honor of writing the 65st message in this PR (of course not counting all the discussions outside). Like a good friend of mine once said: "De rijst is de bestelling" 🍋

Issue #2212, opened 2024-08-15
Clean-up private variables (`_agents`, `_step`)


EwoutH:
To keep backwards compatibility with Mesa 2.x, we added some attributes as private ones, like `_agents` and `_step`. We should clean that up for 3.0.

See:
- https://github.com/projectmesa/mesa/pull/2026

EwoutH:
Partly done in https://github.com/projectmesa/mesa/pull/2251 (`agents_` removal) and https://github.com/projectmesa/mesa/pull/2223 (`_time` removal and `_steps` renamed to `steps`).

quaquel:
What's still remaining?

EwoutH:
It’s mentioned in the migration guide, but I think we want to enforce this somehow.

EwoutH:
That would actually be
- https://github.com/projectmesa/mesa/discussions/2230

So closing this one.

PR #2225, opened 2024-08-19
Change warning when setting model.agents to error

EwoutH:
Replaced the warning with an `AttributeError` when attempting to set `model.agents`. This change enforces that the `agents` attribute will be reserved for internal use by Mesa in future releases. Users must update their code to use a different attribute name for custom agent storage.

This allows us to start using `model.agents` in Mesa 3.0 for ourselves, which helps with https://github.com/projectmesa/mesa/issues/2212.

### DIFF
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -10,7 +10,6 @@ from __future__ import annotations
 
 import itertools
 import random
-import warnings
 from collections import defaultdict
 
 # mypy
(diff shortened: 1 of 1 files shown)


EwoutH:
@projectmesa/maintainers I would like this in the next pre-release. Would somebody be so kind to review?

quaquel (approved):

PR #2226, opened 2024-08-20
Automatically assign `unique_id` to Agents

### LINKED ISSUE #2213: Automate and enforce unique `unique_id` in Agents
Currently, Agent's `unique_id` values are assigned manually on Agent creation, and not guaranteed to be unique. This issue:
- Automatically assign agents an `unique_id` value on creation.
- Guarantee that all `unique_id` values are actually unique.

GittyHarsha:
Automatically assign `unique_id` to Agents. This has two goals:

- Automatically assign agents an `unique_id` value on creation.
- Guarantee that all `unique_id` values are actually unique.

A few design considerations:
- The `current_id` counter starts now at `0`, instead of `1`.
  - This is consistent with `for i in range(n)` loops, which also assign unique_ids from 1.
- Warnings are thrown for the most common cases of agent initialization. Other cases get an informative error.
- The counter is kept inside the model. Testing of unique ids becomes almost impossible without that. It can be moved in another PR if desired.
- `next_id()` was commonly use, so that also gets a deprecation warning.
- The `current_id` counter is kept public. It can be useful to see how many agents are created in total, and is used here and there. Again, can be modified in another PR.

### Usage
Instead of:
```Python
Agent(unique_id, model, ...)
```
You can (and should) now initialize your Agent with:
```Python
Agent(model, ...)
```

Resolves #2213

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -11,6 +11,7 @@ from __future__ import annotations
 import contextlib
 import copy
 import operator
+import warnings
 import weakref
 from collections.abc import Callable, Iterable, Iterator, MutableSet, Sequence
 from random import Random
@@ -35,16 +36,43 @@ class Agent:
         self.pos: Position | None = None
     """
 
-    def __init__(self, unique_id: int, model: Model) -> None:
+    def __init__(
+        self,
+        model: Model = None,
+        fallback: None = None,
+    ) -> None:
         """
         Create a new agent.
 
         Args:
-            unique_id (int): A unique identifier for this agent.
             model (Model): The model instance in which the agent exists.
+            fallback (None): A fallback parameter for the old way of initializing agents. Deprecated.
+
+        Note:
+            Previously, agents were initialized with a unique_id and a model instance. The unique_id is now
+            automatically assigned, so only the model instance is now required (as the first argument).
         """
-        self.unique_id = unique_id
-        self.model = model
+        from .model import Model  # avoid circular import
+
+        # If the first argument is a model, that's the correct new way to initialize the agent.
+        if isinstance(model, Model):
+            self.model = model
+        else:
+            # If the fallback is a Mesa model, assign that to the model attribute
+            if isinstance(fallback, Model):
+                self.model = fallback
+                warnings.warn(
+                    "The unique_id parameter is now generated automatically and doesn't need to be passed in Agent() anymore.\n"
+                    "Please initialize agents with the model instance only: Agent(model, ...), instead of Agent(unique_id, model, ...)",
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
+            else:
+                raise ValueError(
+                    "The model parameter is required to initialize an Agent object. Initialize the agent with Agent(model, ...)."
+                )
+
+        self.unique_id = self.model._next_id
         self.pos: Position | None = None
 
         # register agent
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -123,10 +123,23 @@ class Model:
         self._steps += 1
         self._time += deltat
 
-    def next_id(self) -> int:
-        """Return the next unique ID for agents, increment current_id"""
+    @property
+    def _next_id(self) -> int:
+        """Return the next unique ID for agents. Starts at 0."""
+        return_id = self.current_id
         self.current_id += 1
-        return self.current_id
+        return return_id
+
+    def next_id(self) -> int:
+        """Old version of next_id, kept for compatibility."""
+        warnings.warn(
+            "Model.next_id() is deprecated and will be removed in a future version.\n"
+            "The Agent.unique_id parameter is now generated automatically and doesn't need to be passed in Agent() anymore.\n"
+            "Please initialize agents with the model instance only: Agent(model, ...), instead of Agent(unique_id, model, ...)",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return 0  # This is a dummy return value, but some functions might expect an int. It isn't used in agent creation anyway.
 
     def reset_randomizer(self, seed: int | None = None) -> None:
         """Reset the model random number generator.
--- a/benchmarks/BoltzmannWealth/boltzmann_wealth.py
+++ b/benchmarks/BoltzmannWealth/boltzmann_wealth.py
@@ -28,7 +28,7 @@ class BoltzmannWealth(mesa.Model):
         )
         # Create agents
         for i in range(self.num_agents):
-            a = MoneyAgent(i, self)
+            a = MoneyAgent(self)
             # Add the agent to a random grid cell
             x = self.random.randrange(self.grid.width)
             y = self.random.randrange(self.grid.height)
@@ -50,8 +50,8 @@ class BoltzmannWealth(mesa.Model):
 class MoneyAgent(mesa.Agent):
     """An agent with fixed initial wealth."""
 
-    def __init__(self, unique_id, model):
-        super().__init__(unique_id, model)
+    def __init__(self, model):
+        super().__init__(model)
         self.wealth = 1
 
     def move(self):
--- a/benchmarks/Flocking/flocking.py
+++ b/benchmarks/Flocking/flocking.py
@@ -27,7 +27,6 @@ class Boid(mesa.Agent):
 
     def __init__(
         self,
-        unique_id,
         model,
         speed,
         direction,
@@ -42,6 +41,7 @@ class Boid(mesa.Agent):
 
         Args:
             unique_id: Unique agent identifier.
+            model: Model instance in which the agent is located.
             speed: Distance to move per step.
             direction: numpy vector for the Boid's direction of movement.
             vision: Radius to look around for nearby Boids.
@@ -51,7 +51,7 @@ class Boid(mesa.Agent):
             match: the relative importance of matching neighbors' directions
 
         """
-        super().__init__(unique_id, model)
+        super().__init__(model)
         self.speed = speed
         self.direction = direction
         self.vision = vision
@@ -137,7 +137,6 @@ class BoidFlockers(mesa.Model):
             pos = np.array((x, y))
             direction = np.random.random(2) * 2 - 1
             boid = Boid(
-                unique_id=i,
                 model=self,
                 speed=speed,
                 direction=direction,
--- a/benchmarks/Schelling/schelling.py
+++ b/benchmarks/Schelling/schelling.py
@@ -8,15 +8,16 @@ class SchellingAgent(CellAgent):
     Schelling segregation agent
     """
 
-    def __init__(self, unique_id, model, agent_type, radius, homophily):
+    def __init__(self, model, agent_type, radius, homophily):
         """
         Create a new Schelling agent.
         Args:
            unique_id: Unique identifier for the agent.
+           model: Model instance in which the agent is located.
            x, y: Agent initial location.
            agent_type: Indicator for the agent's type (minority=1, majority=0)
         """
-        super().__init__(unique_id, model)
+        super().__init__(model)
         self.type = agent_type
         self.radius = radius
         self.homophily = homophily
@@ -81,9 +82,7 @@ class Schelling(Model):
         for cell in self.grid:
             if self.random.random() < density:
                 agent_type = 1 if self.random.random() < self.minority_pc else 0
-                agent = SchellingAgent(
-                    self.next_id(), self, agent_type, radius, homophily
-                )
+                agent = SchellingAgent(self, agent_type, radius, homophily)
                 agent.move_to(cell)
                 self.schedule.add(agent)
 
--- a/benchmarks/WolfSheep/wolf_sheep.py
+++ b/benchmarks/WolfSheep/wolf_sheep.py
@@ -17,8 +17,8 @@ from mesa.experimental.devs import ABMSimulator
 
 
 class Animal(CellAgent):
-    def __init__(self, unique_id, model, energy, p_reproduce, energy_from_food):
-        super().__init__(unique_id, model)
+    def __init__(self, model, energy, p_reproduce, energy_from_food):
+        super().__init__(model)
         self.energy = energy
         self.p_reproduce = p_reproduce
         self.energy_from_food = energy_from_food
@@ -29,7 +29,6 @@ class Animal(CellAgent):
     def spawn_offspring(self):
         self.energy /= 2
         offspring = self.__class__(
-            self.model.next_id(),
             self.model,
             self.energy,
             self.p_reproduce,
@@ -107,7 +106,7 @@ class GrassPatch(CellAgent):
                 function_args=[self, "fully_grown", True],
             )
 
-    def __init__(self, unique_id, model, fully_grown, countdown, grass_regrowth_time):
+    def __init__(self, model, fully_grown, countdown, grass_regrowth_time):
         """
         TODO:: fully grown can just be an int --> so one less param (i.e. countdown)
 
@@ -119,7 +118,7 @@ class GrassPatch(CellAgent):
             grass_regrowth_time : time to fully regrow grass
             countdown : Time for the patch of grass to be fully regrown if fully grown is False
         """
-        super().__init__(unique_id, model)
+        super().__init__(model)
         self._fully_grown = fully_grown
         self.grass_regrowth_time = grass_regrowth_time
 
@@ -189,7 +188,6 @@ class WolfSheep(Model):
             )
             energy = self.random.randrange(2 * sheep_gain_from_food)
             sheep = Sheep(
-                self.next_id(),
                 self,
                 energy,
                 sheep_reproduce,
@@ -205,7 +203,6 @@ class WolfSheep(Model):
             )
             energy = self.random.randrange(2 * wolf_gain_from_food)
             wolf = Wolf(
-                self.next_id(),
                 self,
                 energy,
                 wolf_reproduce,
@@ -221,9 +218,7 @@ class WolfSheep(Model):
                 countdown = grass_regrowth_time
             else:
                 countdown = self.random.randrange(grass_regrowth_time)
-            patch = GrassPatch(
-                self.next_id(), self, fully_grown, countdown, grass_regrowth_time
-            )
+            patch = GrassPatch(self, fully_grown, countdown, grass_regrowth_time)
             patch.move_to(cell)
 
     def step(self):
--- a/mesa/experimental/cell_space/cell_agent.py
+++ b/mesa/experimental/cell_space/cell_agent.py
@@ -19,7 +19,7 @@ class CellAgent(Agent):
         cell: (Cell | None): the cell which the agent occupies
     """
 
-    def __init__(self, unique_id: int, model: Model) -> None:
+    def __init__(self, model: Model) -> None:
         """
         Create a new agent.
 
@@ -27,7 +27,7 @@ class CellAgent(Agent):
             unique_id (int): A unique identifier for this agent.
             model (Model): The model instance in which the agent exists.
         """
-        super().__init__(unique_id, model)
+        super().__init__(model)
         self.cell: Cell | None = None
 
     def move_to(self, cell) -> None:
--- a/mesa/experimental/devs/examples/epstein_civil_violence.py
+++ b/mesa/experimental/devs/examples/epstein_civil_violence.py
@@ -7,8 +7,8 @@ from mesa.space import SingleGrid
 
 
 class EpsteinAgent(Agent):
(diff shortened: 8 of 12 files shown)


EwoutH:
This is an interesting approach, thanks!

Do you mind if I write some test cases and add them to this PR? Then we can see what edge cases are covered and for which we still have to come up with a solution.

quaquel:
Good stuf. A few comments that might be useful to move this PR forward

1. please make sure the original message is self contained. So summarize what the PR does. I now have to follow the link to figure it out. Please still include the link for those who want to know more.
2. As asked in the original issue (q1), the counting methods should move from the model class to the agent class and be a class method.
3. I will try to look closely at the code later today and think about MESA 2 versus 3 and how to handle this cleanly.

GittyHarsha:
> Do you mind if I write some test cases and add them to this PR? Then we can see what edge cases are covered and for which we still have to come up with a solution.

sure!

EwoutH:
Done! I added 3 test cases, as you can see, currently one of them fails.

What helps is that most people just pass `super().__init__(unique_id, model)`, without the keywords. That all our example models pass is also a good sign.

The test that currently fails is the one that uses all keyword arguments:
```Python
super().__init__(unique_id=unique_id, model=model)
```
Maybe `unique_id` could be another optional parameter that's `None` by default, but get's checked if it's passed.

Curious if you can get all three test cases simultaneously working!

GittyHarsha:
@EwoutH here is the code changes I am making:
```
    def __init__(
        self, unique_id_or_model: int | Model = None, model: Model | None = None, unique_id: int | None = None
    ) -> None:
        """
        Create a new agent.

        Args:
            unique_id_or_model (int | Model, optional): A unique identifier for this agent or the model instance in which the agent exists.
            Defaults to None, meaning `model` and `unique_id` parameters are given as keyword arguments.
            model (Model, optional): The model instance in which the agent exists given the argument `unique_id_or_model` is a unique_id for this agent.
            Defaults to None, meaning only the model instance is given as an argument to `unique_id_or_model`.
            unique_id (int, optional): A unique identifier for this agent. Defaults to None, meaning it is not given a keyword parameter.
        """
        if unique_id is not None and model is not None:
            self.unique_id = unique_id
            self.model = model
        else:
            if isinstance(unique_id_or_model, int):
                if model is None:
                    raise RuntimeError('The second argument must be a model instance')
                else:
                    self.unique_id = unique_id_or_model 
                    self.model = model 
            elif unique_id is not None:
                self.unique_id = unique_id
                self.model = unique_id_or_model
            else:
                self.unique_id = unique_id_or_model.next_id()
                self.model = unique_id_or_model
                

```

All tests are passing expect the tests in `tests\test_time.py`. because here the parameter `unique_id_or_model` is being passed a string 
```
self = <tests.test_time.MockAgent object at 0x000001F31840B4D0>, unique_id_or_model = 'A', model = <tests.test_time.MockModel object at 0x000001F31840B440>, unique_id = None
```

My code changes checks only for instance `int`, thus it then tries to execute `self.unique_id = unique_id_or_model.next_id()` and thus gives error: `AttributeError: 'str' object has no attribute 'next_id' `


So, can strings be passed as arguments for a unique_id? 
So, making changes : `if isinstance(unique_id_or_model, (int, str))`: should resolve it? or are there any other data types that can be passed to be assigned as unique_id which will be implicitly converted to int?


To solve this I also tried to do `if not isinstance(unique_id_or_model, Model)` but this doesn't work as `Model` cannot be imported during run time and if imported causes cyclic dependency

I can refactor the code to not include any type checking for `unique_id_or_model` but this will miss many cases like when both the parameters are of type `Model`.

GittyHarsha:
Also Instead of adding another optional paramter `unique_id`, just changing the name `unique_id_or_model` to `unique_id` passes all the tests, but is semantically incorrect because only a single argument of type `Model` will be referenced by `unique_id` paramter,

quaquel:
I don't see a simple solution for supporting both current style and new style Agents. Currently, `unique_id` and `model` are arguments, not keyword arguments. It might be possible to switch to `Agent.__init__(first_arg, *args)` and just check whether there is anything in `*args`.  Personally, I am more included to just have a hard break with MESA 3. It is easy to document and easy to fix in existing models. It would also make the code here much simpler. 

For my point 2 above, see [the SimulationEvent code](https://github.com/projectmesa/mesa/blob/50b2964e80897777e92935cb1271cce1c3314845/mesa/experimental/devs/eventlist.py#L18-L62) on how to handle the counting nicely. The code there is the recommended way of counting instances of a class on [StackOverflow](https://stackoverflow.com/questions/8628123/counting-instances-of-a-class). This can be implemented in a backward compatibly way and then just let `Model.next_id` raise a `DeprecationWarning`.

EwoutH:
I thought it would break a lot more models, but since the name of the variable in your subclass are defined by yourself and the variables names only matter when calling `super()`, the impact is way smaller.

The fact that all regular example models passed makes it a lot less impactful to make a breaking change here.

mesa/agent.py
```diff
@@ -35,16 +35,26 @@ class Agent:
         self.pos: Position | None = None
     """
 
-    def __init__(self, unique_id: int, model: Model) -> None:
+    def __init__(
+        self, unique_id_or_model: int | Model, model: Model | None = None
```
EwoutH, line 39: I would make it this:
```suggestion
        self, model: Model, fallback: None = None
```
And then warn if either `model` is not a Model instance or if `fallback` is not None.

EwoutH: @GittyHarsha what do you think of this?

GittyHarsha: Is this meant to support existing codes? If so then keywordAgent cannot be supported?

Sorry, I didn't understand the purpose of fallback

EwoutH: Yes, that's the idea.

The purpose of the fallback is to not let models that use two positional arguments (which is very common) crash hard.

So:
We support and recommend:
```Python
super().__init__(my_model)
```
We allow
```Python
super().__init__(unique_id, my_model)
```
which we convert internally to to let my_model be the model and throw unique_id away, with a warning to convert to `super().__init__(my_model)`.

I think this will be a good balance to moving towards the future of Mesa and not breaking existing models too hard.

GittyHarsha: > self, model: Model, fallback: None = None

according to this, the order of arguments should be (model, unique_id) but existing codes use the order (unique_id, model).
Also is there a need to support keyword arguments? In that case, this doesn't work?

EwoutH: Yeah so we have to handle that internally

GittyHarsha: I implemented your suggestions:

    _ids = itertools.count()

```
    def __init__(
       self, model: Model, fallback: None = None, **kwargs
    ) -> None:

        if "unique_id" in kwargs:
            self.unique_id = kwargs["unique_id"]
            self.model = model 
        elif fallback is None:
            self.unique_id = next(self._ids)
            self.model = model 
        else:
            self.unique_id = model 
            self.model = fallback
```
only one test is failing for this type of input
```
     super().__init__(unique_id, model=model)
E       TypeError: Agent.__init__() got multiple values for argument 'model'
```

EwoutH: Thanks. I don't care too much about that test case, it won't happen too often.

I do want to use the `next(self._ids)` always, so just ignore the unique_id inputted by the user.

GittyHarsha:
```
    _ids = itertools.count()

    def __init__(
       self, unique_id_or_model: int | Model | None = None, fallback: Model | None = None, **kwargs
    ) -> None:
        """
        Create a new agent.

        Args:
            model (Model):
            fallback (optional):
            kwargs
        """
        if "unique_id" in kwargs:
            if "model" in kwargs:
                self.unique_id = kwargs["unique_id"]
                self.model = kwargs["model"]
            else:
                self.unique_id =kwargs["unique_id"]
                self.model = unique_id_or_model
        elif "model" in kwargs:
            self.unique_id = unique_id_or_model
            self.model = kwargs["model"]
        elif fallback is None:
            self.unique_id = next(self._ids)
            self.model = unique_id_or_model
        else:
            self.unique_id = unique_id_or_model 
            self.model = fallback

```


this passes all the tests.

EwoutH:
Do you mind if I make some modifications on your branch? You made a great start but I would like to tidy it up myself.

GittyHarsha:
Sure! No problem

EwoutH:
@projectmesa/maintainers Ready for review!

mesa/model.py
```diff
@@ -185,10 +185,23 @@ def run_model(self) -> None:
...
-    def next_id(self) -> int:
-        """Return the next unique ID for agents, increment current_id"""
+    @property
+    def _next_id(self) -> int:
+        """Return the next unique ID for agents. Starts at 0."""
+        return_id = self.current_id
         self.current_id += 1
-        return self.current_id
+        return return_id
+
```
quaquel, line 194: in my opinion, this is part of the behavior of the Agent class so it is strange that it is located in the model class.

EwoutH: Agreed, but let's move it in a separate PR. It breaks practically all tests with more than one test class or function in a file.

mesa/agent.py
```diff
@@ -37,16 +37,43 @@ class Agent:
...
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
+            else:
+                raise ValueError(
+                    "The model parameter is required to initialize an Agent object. Initialize the agent with Agent(model, ...)."
+                )
+
+        self.unique_id = self.model._next_id
         self.pos: Position | None = None
```
quaquel, line 77: I'll try to have a closer look tomorrow. I think it might be possible to ensure backward compatibility in a more elegant way.

mesa/model.py
quaquel: Not sure I fully understand your point. Why not move the counting code into Agent in this PR? 

I haven't looked at how bad this would be in terms of tests, but it seems logical to include it here?

EwoutH:
The benchmarks are slower because they still use the old structure which creates a huge amount of warnings.

wang-boyu:
A few comments from my side (open for discussions):

1. In the past I used `uuid` module to create unique agent ids, such as [here](https://github.com/projectmesa/mesa-examples/blob/3e5ebd12c8034dc6a61a3c809cc4fe23014927f3/gis/geo_schelling_points/geo_schelling_points/model.py#L37). In this way we don't need a counter in `Model`, so we can get rid of `model.next_id()` as well as `model.current_id`.

2. What if users want to explicitly set agent ids by themselves? For instance, I may set agent ids to be their unique names. It might be better to keep the `unique_id` parameter of the `Agent` class optional.
   - If `unique_id` is provided, `agent.unique_id` is set accordingly. This is current behavior.
   - Otherwise (i.e., `unique_id = None`), as a default behavior, Mesa creates a unique agent id automatically.

3. It may be useful to check for uniqueness of agent ids. For example, we could have a `model._agent_id_set` attribute that is a `set`. While creating agents, we check whether new agent id is already taken and raise exception.

quaquel:
> A few comments from my side (open for discussions):
> 
> 1. In the past I used `uuid` module to create unique agent ids, such as [here](https://github.com/projectmesa/mesa-examples/blob/3e5ebd12c8034dc6a61a3c809cc4fe23014927f3/gis/geo_schelling_points/geo_schelling_points/model.py#L37). In this way we don't need a counter in `Model`, so we can get rid of `model.next_id()` as well as `model.current_id`.

Regardless of how it is done, it should move the Agent class. 

I am not familiar with `uuid` but is that not overkill for the function it has in MESA? The basic function of `unique_id` is to serve as a unique identifier for indexing agents when collecting agent level data. Why would we need a 128-bit identifier for this? A possibly convenient side-effect of counting is that you know how many agents have been created over the runtime of the model. 

> 2. What if users want to explicitly set agent ids by themselves? For instance, I may set agent ids to be their unique names. It might be better to keep the `unique_id` parameter of the `Agent` class optional.
>    * If `unique_id` is provided, `agent.unique_id` is set accordingly. This is current behavior.
>    * Otherwise (i.e., `unique_id = None`), as a default behavior, Mesa creates a unique agent id automatically.

`unique_id` is necessary for the functioning of MESA (e.g., data collection). So, analogous to step, we should do it automatically. If the user wants her own id, they can do so with their own attribute. 

> 3. It may be useful to check for uniqueness of agent ids. For example, we could have a `model._agent_id_set` attribute that is a `set`. While creating agents, we check whether new agent id is already taken and raise exception.

If you make it part of MESA, uniqueness is guaranteed by construction. Messing with internal attributes of classes is at your own peril. So I don't see a need to be so defensive in how we code it up. If we want to be defensive, you could move `unique_id` into `_unique_id` and have a property for `unique_id` and a setter that raises some appropriate exception.

mesa/model.py
EwoutH: Copy my branch and move it. You will see what happens, and maybe you have an elegant solution for it.

Corvince:
First of all lets not again discuss this PR to its death. Great work by @GittyHarsha and @EwoutH. I think making it automatic by default is very good and the main motivation for this PR. Everything else can be discussed separately. 



> 1. In the past I used `uuid` module to create unique agent ids, such as [here](https://github.com/projectmesa/mesa-examples/blob/3e5ebd12c8034dc6a61a3c809cc4fe23014927f3/gis/geo_schelling_points/geo_schelling_points/model.py#L37). In this way we don't need a counter in `Model`, so we can get rid of `model.next_id()` as well as `model.current_id`.

I did the same, but I think its overkill. UUID is often used to make the ids universally unique and to prevent guessing the next id. We don't need either of those. As a side effect a counter even conveys additional information, for example the order in which agents were created. Its also much easier for humans to track a number than a uuid. 

> 2. What if users want to explicitly set agent ids by themselves? For instance, I may set agent ids to be their unique names. It might be better to keep the `unique_id` parameter of the `Agent` class optional.

I think its unnecessary. For us maintainers we now know for certain that the unique_id is truly unique. Before this PR I was always reluctant to fully rely on this.
For users I don't see the benefit of overwriting unique_id. E.g. If you already have the name just use it as name, not id. And more importantly if users want to use their own id, they can still simply overwrite the unique_id attribute. 

> 3. It may be useful to check for uniqueness of agent ids. For example, we could have a `model._agent_id_set` attribute that is a `set`. While creating agents, we check whether new agent id is already taken and raise exception.

Again, we don't need this if unique_id is not, but automatically assigned. 





> Regardless of how it is done, it should move the Agent class.

Partly agree. At first I thought we should move it to the `mesa` or `mesa.utils` namespace and make it generally available. But its nice that it only counts agents. But still I don't think it should be part of the agent class, but the agent namespace. And we can just expose an instance of [`itertools.count` ](https://docs.python.org/3/library/itertools.html#itertools.count)

quaquel:
> Partly agree. At first I thought we should move it to the mesa or mesa.utils namespace and make it generally available. But its nice that it only counts agents. But still I don't think it should be part of the agent class, but the agent namespace. And we can just expose an instance of [itertools.count](https://docs.python.org/3/library/itertools.html#itertools.count)

I would be fine with that as well. In [SimulationEvent](https://github.com/projectmesa/mesa/blob/95e1cd26c78c1fce5426db8d25e5d92b61b8e827/mesa/experimental/devs/eventlist.py#L18), I used a class based approach based on the most upvoted StackOverflow answer, but I am fine either way. The main point for me is that it is not part of the behavior of `Model` so it should be removed from there (here or in a next PR).

EwoutH:
I just want to add: I try to keep my PRs small in scope and if possible even [atomic](https://www.pauline-vos.nl/atomic-commits/).

I know there's benefit into doing things right once, but sometimes that makes PRs unmanageable big, either code wise or, more often, conceptually (which is why we don't have a new datacollector yet).

This PR modifies one things: Automatically increasing the `unique_id`. While I _agree_ that should be ideally done in the Agent, it would make this PR way too big. In two ways:
- It introduces a large conceptual change that `unique_id`s are not per Model, but per Agent superclass.
- It breaks a lot of tests that rely on the Agent count starting at 0 when a new model is initialized. So if you define multiple models (like you do in test classes) is messes a lot of stuff up.

image

So again, I'm not fundamentally against any of these changes. It just are more changes that are better fitted in a new PR.

wang-boyu:
Thanks for your responses @quaquel @Corvince! Some further comments -

**Regarding implementation details of `unique_id`**

I don't really care if it's using `uuid` or a counter. I used `uuid` because it's less work for me. In this PR, if we're using a counter, is it possible to remove `model.next_id()` and `model.current_id`, or make them private? There appear to be implementation details of `unique_id` and shouldn't be exposed to users. In addition, they may easily cause breaking changes in the future if we'd like to change anything.

> A possibly convenient side-effect of counting is that you know how many agents have been created over the runtime of the model.

I wouldn't count on these because agents can be removed/destroyed anytime. We then need to differentiate how many agents were created and how many agents are still alive.

> As a side effect a counter even conveys additional information, for example the order in which agents were created.

`unique_id` are supposed to be unique identifiers, nothing more, nothing less. I don't really like the ideas of having side effects that rely on implementation details. Sorry : )

> Regardless of how it is done, it should move the Agent class.

I agree. But if it's using a counter, then the counter has to be stored somewhere outside individual agents. Keeping it inside a model makes sense. However, this is also why I think using `uuid` is better - it eliminates the need of maintaining a counter by ourselves.

**Regarding user-settable `unique_id`**

The central question is: why are users not allowed to use their own unique ids?

My suggestions are:

- By default, we (Mesa) create unique ids so users don't have to worry about them.
- Advanced users can use their own unique ids. Mesa validates uniqueness to make sure they don't screw up.

As an example, Mesa-Geo sets `unique_id` of agents when batch creating geo-agents from geodataframe: https://github.com/projectmesa/mesa-geo/blob/9c3196c932be94fdc6ead4a599740445c56d63bd/mesa_geo/geoagent.py#L198-L200

quaquel:
I ran a quick performance test on just 1000 agents, UUID vs. itertools.count

Screenshot 2024-08-30 at 13 44 55

I just don't see the need for UUID in this context, and it comes with a real performance difference. 

> In this PR, if we're using a counter, is it possible to remove model.next_id() and model.current_id, or make them private? There appear to be implementation details of unique_id and shouldn't be exposed to users. 

I agree but there is @EwoutH point on how this causes all tests to fail. I want to take a closer look myself at this later today. I think it is possible to move everything into `Agent`, while raising deprecation warnings from the old structure. 

> I agree. But if it's using a counter, then the counter has to be stored somewhere outside individual agents. 

In my screenshot, `_ids` is a class attribute, so it is not stored inside the individual agents. Given that it is part of the behavior of the Agent class, it makes sense it is either done at the class level (my preference) or at least part of the Agent namespace (as suggested by @Corvince).

> The central question is: why are users not allowed to use their own unique ids?

This is perfectly analogous to the discussion on automatically counting steps. Users can still do this, but just not by using the unique_id attribute. MESA relies on unique_id for agent data collection, so it makes sense that MESA internally handles this correctly. As a minor point, it was deeply confusing to many of my students why they had to supply the unique_id themselves.

quaquel:
@EwoutH, you might want to check #2260. This is how I would go about solving the issue. I don't have any existing tests failing, while unique_id counting is done inside the Agent class. #2260 is, of course, not complete. For example, tests are still needed, and I would add a deprecation warning inside `Model`. Am I missing something important? I fail to understand what is so complicated about any of this.

Corvince:
What I don't understand, but unfortunately can't test right now: doesn't this keep increasing the counter indefinitely, even across models? So if I run the same model twice without restarting my kernel, won't the ids be different?

EwoutH:
@Corvince you’re spot on. That’s what I discovered when I tried to move it. That’s why I didn’t move it.

And I explained this I think 5 times now.

Sorry guys, but I’m getting a bit done with the “whataboutism” this way we never get anything done.

We’re all experienced developers. We have to trust each other and have to give each other a bit of room.

wang-boyu:
Just for the record, my comments are not blocking. I remembered we discussed this before (correct me if I'm wrong), if at least two maintainers approve a PR then it's considered mergable.

EwoutH:
I’m resigning as owner of this specific effort. Feel free to work further on this branch.

Corvince:
> unique_id are supposed to be unique identifiers, nothing more, nothing less. I don't really like the ideas of having side effects that rely on implementation details. Sorry : )

Isn't this contradictory to the rest of your argument? ;)

If this is the case, and I basically agree with this, then why make so much fuss about being able to change them? If it's just for identification, one can be happy that it happens automatically, no need to assign custom names or anything else to them.

What I meant with positive side effects that you can eyeball some information from them. Nothing to rely on, but a good estimate about the age of the agent and if you are tracking the same agent in multiple places

Corvince (approved):

quaquel:
closing as solved via #2260

PR #2240, opened 2024-08-23
SolaraViz: Reset components when params are changed

### LINKED ISSUE #2239: SolaraViz: visualization should update on parameter changes
This is on branch main (or 3.0.0a2 (2024-08-21)) 

**Describe the bug**
Currently, using SolaraViz,  manipulating the controls to change `model_params` doesn't change seem to cause a redraw for the `space_drawer`.  (Changing the seed, or advancing one step does redraw with the current parameters).  This is most noticeable for network space models, for example `virus_on__network`.     I also note that pressing reset redraws but not with the updated model parameters.

**Expected Behavior**
Ideally (perhaps optionally) the model parameters should be reactive. This way the user can experiment with the network structures in real time.    

**To Reproduce**
Use main branch mesa-examples `virus_on_network`,  run `solara run app.py`, note that model steps fine, however changing parameters has no visible effect until you press `step` or change the seed.

EDIT: Clarified what happens with the 'reset' button.

rht:
This fixes #2239 (summary can be found in the issue). cc: @DrEntropy.

> Currently, using SolaraViz,  manipulating the controls to change `model_params` doesn't change seem to cause a redraw for the `space_drawer`.  (Changing the seed, or advancing one step does redraw with the current parameters).  This is most noticeable for network space models, for example `virus_on__network`.     I also note that pressing reset redraws but not with the updated model parameters.

### DIFF
--- a/mesa/visualization/solara_viz.py
+++ b/mesa/visualization/solara_viz.py
@@ -158,7 +158,11 @@ def SolaraViz(
         """Update the random seed for the model."""
         reactive_seed.value = model.random.random()
 
-    dependencies = [current_step.value, reactive_seed.value]
+    dependencies = [
+        *list(model_parameters.values()),
+        current_step.value,
+        reactive_seed.value,
+    ]
 
     # if space drawer is disabled, do not include it
     layout_types = [{"Space": "default"}] if space_drawer else []


DrEntropy:
I verified for my use case this solves the problem.  I had not expected it to be a one liner ! Nice.

EwoutH (approved):

PR #2241, opened 2024-08-23
SolaraViz: Remove card title

rht:
They are soon going to be part of one umbrella of "measures", instead of distinctly named ones. If a user wants to add a title, they can do it inside Matplotlib/Altair.

### DIFF
--- a/mesa/visualization/solara_viz.py
+++ b/mesa/visualization/solara_viz.py
@@ -58,7 +58,6 @@ def Card(
         style_=f"background-color: {color}; width: 100%; height: 100%"
     ) as main:
         if "Space" in layout_type:
-            rv.CardTitle(children=["Space"])
             if space_drawer == "default":
                 # draw with the default implementation
                 components_matplotlib.SpaceMatplotlib(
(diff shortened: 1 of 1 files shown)


tpike3:
@rht Thanks! -- Could you elaborate on "They are soon going to be part of one umbrella of measures". Is this going to be a change to Mesa, Solara or both? 

Could you add to the visualization tutorial adding names to the plots? I think this is a small but critical feature for users.

EwoutH:
How can I help move this forward?

rht:
> Could you elaborate on "They are soon going to be part of one umbrella of measures"

https://github.com/projectmesa/mesa/discussions/2236, and in https://github.com/projectmesa/mesa/pull/2263, where measures and space are merged into "components" (what I earlier called measures).

> Could you add to the visualization tutorial adding names to the plots? I think this is a small but critical feature for users.

Currently there is no public API for that, and this PR doesn't add/remove any feature in this area.

> How can I help move this forward?

This PR just needs a clarification answer to @tpike3 's questions. Unless there is other missing aspect.

EwoutH:
@rht @Corvince what's the current state / trajectory of this PR?

quaquel:
This PR is outdated and in the revised solara_viz this code is not used anymore.

PR #2242, opened 2024-08-23
More parameter functionality in matplotlib (default) drawer visualization

rmhopkins4:
(screenshots of changes w/ example `agent_portrayal` functions coming)

Stuff like this is now possible:
```Python
def agent_portrayal(cell):

    return {
        "c": "white" if cell.isAlive else "black",
        "s": 45,
        "marker": "*",
        "edgecolors": "purple",
        "linewidths": 0.5,
        "alpha": 0.5,
    }
```
image

- updates necessary docstring
- supports all params in matplotlib's [scatter](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.scatter.html) with the exception of `plotnonfinite`
- manual colormapping to get around restrictions similar to those with 'marker', where a unique colormap cannot usually be provided for each point plotted
- streamlined code for `_split_and_scatter` and `portray`

### DIFF
--- a/mesa/visualization/components/matplotlib.py
+++ b/mesa/visualization/components/matplotlib.py
@@ -4,6 +4,7 @@ import networkx as nx
 import solara
 from matplotlib.figure import Figure
 from matplotlib.ticker import MaxNLocator
+from matplotlib.pyplot import get_cmap
 
 import mesa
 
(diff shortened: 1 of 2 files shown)


EwoutH:
Thanks for this PR, sorry none of us got back to you (@projectmesa/maintainers we should be better in this - a full week is quite long).

It looks really useful. We indeed need to discuss how we handle variable names.

I might support going full `matplotlib` convention. So `marker` instead of `shape`, etc.

rht:
There is a problem with using Matplotlib nomenclature because Altair's API naming might be different. But it's better to pick either Matplotlib/Altair than to create a separate new naming scheme that nobody is familiar with. I haven't looked at Altair's but I assume the naming would be more modern.

EwoutH (commented):
I really like the capabilities this brings. Thanks for working on this, good visualization is important.

This does need to be documented really well. Try if you can add some more examples and edge cases to your PR post.

EwoutH:
@rmhopkins4 are you still interested in working on this? If so, how can I help?

jackiekazil:
@EwoutH, while in a perfect world, we hope that @rmhopkins4 hasn't moved on, but if they have - I wonder if we should create a tag to recognize incomplete but valuable PRs that might be picked up in the future for completion.

EwoutH:
Yeah I agree this one is worth getting in. Maybe @Corvince or @rht wants to finish it?

For now we can just add it to the 3.0 milestone.

EwoutH:
As soon as #2336 is merged I'm going to try to revive this.

quaquel:
superseded by #2467

PR #2245, opened 2024-08-24
Update solara_viz.py

tpike3:
- adjust sizes to make more visually appealing

Adjust background grid to encompass displays; interestingly I tried at a bunch of other options like using `ResponsiveColumn`, tried several different number combinations, but it seemed the simplest was the best. 

Boltzmann Wealth Current:
Mesa current
Boltzmann Wealth Adjusted
Mesa after adjust

Geo-Sir Current
Screenshot 2024-08-24 074812
Geo-Sir Adjusted
geo_mesa current

### DIFF
--- a/mesa/visualization/solara_viz.py
+++ b/mesa/visualization/solara_viz.py
@@ -449,7 +449,7 @@ def make_initial_grid_layout(layout_types):
         {
             "i": i,
             "w": 6,
-            "h": 10,
+            "h": 14,
             "moved": False,
             "x": 6 * (i % 2),
             "y": 16 * (i - i % 2),


mesa/visualization/solara_viz.py
```diff
@@ -449,7 +449,7 @@ def make_initial_grid_layout(layout_types):
         {
             "i": i,
             "w": 6,
-            "h": 10,
+            "h": 14,
```
rht, line 452: I tested. This is too tall for my screen (12.5"). The long term solution is probably too use `ResponsiveColumn`.

rht: There is a solution at https://github.com/widgetti/solara/discussions/158.

rht:
Yes, as said earlier, the solution is https://github.com/widgetti/solara/discussions/158.

tpike3:
Makes sense i was going to use responsive column but seemed like too much refactoring... I will close this PR.

PR #2246, opened 2024-08-25
Resolve multiprocessing warning, state Python 3.13 support

rht:
This PR addresses a DeprecationWarning related to using `fork()` in multi-threaded processes, which can lead to deadlocks in the child processes when using Python's `multiprocessing` module.  The 'spawn' method avoids the issues that arise from forking in a multi-threaded environment by starting a new Python interpreter process for each worker, which is safer and more compatible with modern systems.

It also adds the Python 3.13 qualifier to the pyproject.toml.

### DIFF
--- a/mesa/batchrunner.py
+++ b/mesa/batchrunner.py
@@ -1,4 +1,5 @@
 import itertools
+import multiprocessing
 from collections.abc import Iterable, Mapping
 from functools import partial
 from multiprocessing import Pool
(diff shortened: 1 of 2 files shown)


EwoutH:
Thanks! I notice there's a warnings that source directly from Python 3.13's multiprocessing, it might be nice to resolve that first.
```
tests/test_batch_run.py::test_batch_run_no_agent_reporters
  /opt/hostedtoolcache/Python/3.13.0-rc.1/x64/lib/python3.13/multiprocessing/popen_fork.py:67: DeprecationWarning: This process (pid=2042) is multi-threaded, use of fork() may lead to deadlocks in the child.
    self.pid = os.fork()
```

EwoutH (approved):
Warning resolved, good to go!

PR #2248, opened 2024-08-26
DEVS/simulator: Use Model time

EwoutH:
This PR is not intended to (ever) be merged, but to spark discussion about how the simulator could integrate with timekeeping in the model.

The main point that I want to discuss is that now that we are moving away from schedulers, and thus moving away from `schedulers.time`, I don't to move time away from the model _again_ (in this case to `simulator.time`), where a datacollector or visualisation module is unable to find it.

I think it would be logical if the simulator would update the model time. If one is used, the model time should also be increased on a step or otherwise, unless explicitly stated.

The reason this is a PR and not a discussion is that this way we can also see and discuss potential code issues.

### DIFF
--- a/mesa/experimental/devs/simulator.py
+++ b/mesa/experimental/devs/simulator.py
@@ -33,7 +33,6 @@ class Simulator:
         self.start_time = start_time
         self.time_unit = time_unit
 
-        self.time = self.start_time
         self.model = None
 
     def check_time_unit(self, time: int | float) -> bool: ...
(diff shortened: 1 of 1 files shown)


mesa/experimental/devs/simulator.py
```diff
@@ -164,7 +157,7 @@ def schedule_event_relative(
 
         """
         event = SimulationEvent(
-            self.time + time_delta,
+            self.model._time + time_delta,
```
quaquel, line 160: perhaps betraying my java background, but changing attributes of other objects directly like this makes me uncomfortable.

EwoutH: I understand that. I think in this case it allows the model to be the ground truth of time. We should document it as "A DEVS simulator is an alternative way to manage agent activation and progress time".

EwoutH:
Closing this, since it was just a conceptual PR to start thinking about this.

@quaquel I would support a property that copies simulator time, if that makes implementation significantly more elegant. I'm also good with the simulator not keeping it's time at all (or keep it private in like `_time`), to keep the focus on `model.time`.

Issue #2224, opened 2024-08-18
Boltzman performance issue


quaquel:
In mesa-frames, there has been a [bit of discussion](https://github.com/adamamer20/mesa-frames/issues/25) on the performance of the Boltzman Wealth model. I have investigated the issue and conclude that the problem is **not** with the AgentSet as used in the scheduler. Instead, the problem stems from how a user can access all agents in the model. 

Below, we see the performance of 3 versions of the model. Original is the current example version of the model. List is a version of the model where the scheduler is replaced with a list, and in the Agent, this list is also used to select another agent with whom to swap wealth. Test is an intermediate version. The model uses a regular scheduler but the Agent uses the list. 

Below are the key code snippets 

```python

# original
other_agent = self.random.choice(self.model.schedule.agents)

# test & list
other_agent = self.random.choice(self.model.my_agents)

# in MoneyModel for test and list

self.my_agents = []

# Create agents
for i in range(self.num_agents):
    a = MoneyAgent(i, self)

    # Add the agent to the scheduler for test
    self.schedule.add(a)
    
    self.my_agents.append(a)

```

![readme_plot](https://github.com/user-attachments/assets/d8428c23-e4ce-4e1b-b7dc-9f442ca2f432)

As seen in the first plot, the original version scales poorly, but list and test seem very similar. So, below, we compare these in more detail up to 10k agents. As can be seen, using a list within the Agent solves virtually all performance loss. The small remaining difference in performance is the overhead of using AgentSet inside the scheduler instead of using a simple list. This is entirely due to the reliance on weak references in AgentSet.

![readme_plot3](https://github.com/user-attachments/assets/83d1560d-201e-4203-8564-1ee421ae0870)

So what does this mean? The way in which a user can get hold of all agents in the model needs to be refined. Currently, the model has a dictionary with classes as keys, and lists with hard references as values. Every time you do `model.agents` this is copied into a new AgentSet class. This creates a lot of overhead. For Mesa 3, I suggest having both a list with the hard references and actively maintaining an AgentSet with all agents. Model.agents can then always return this AgentSet, while the list with hard references can be a private attribute on the model. 

I also want to know if we need to maintain the dict with Agent classes as key and lists of agents as values. With [AgentSet.groupby](https://github.com/projectmesa/mesa/pull/2220#issue-2471564909), this is redundant unless a user needs frequent and fast access to agents by class and the agents in the model change a lot from one tick to the next.

EwoutH:
[...]

PR #2251, opened 2024-08-27
Performance enhancements for Model.agents

quaquel:
This PR is a performance enhancement for Model.agents. It emerged out of [a discussion](#2224) on the weird scaling performance of the Boltzman wealth model. 

# Key changes
* `model.agents` now returns the agentset as maintained by the model, rather than a new copy based on the hard references
* agent registration and deregistration have been moved from the Agent into the model. The agent now calls `model.register` and `model.deregister`. This encapsulates everything cleanly inside the model class and makes Agent less dependent on the inner details of how Model manages the hard references to agents
* the setup of the relevant datastructures is moved into its own helper method, again this cleans up code.


mesa/model.py
```diff
@@ -94,16 +99,51 @@ def agents(self, agents: Any) -> None:
...
 
-        self._agents = agents
-
     @property
     def agent_types(self) -> list[type]:
         """Return a list of different agent types."""
-        return list(self.agents_.keys())
+        return list(self.agents_by_type.keys())
 
     def get_agents_of_type(self, agenttype: type[Agent]) -> AgentSet:
```
EwoutH, line 107: I think we can deprecate this function and just let users do `self.agents_by_type[agenttype]`.

mesa/model.py
```diff
@@ -71,20 +72,24 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
 
         self._steps: int = 0
         self._time: TimeT = 0  # the model's clock
 
+    def _setup_agent_registration(self):
+        def hack():
+            return AgentSet([], self)
+
+        self.agents_by_type: defaultdict[type, AgentSet] = defaultdict(hack)
+        self._agents = {}
```
EwoutH, line 86: How often are we using this dict still? Does it remain useful?

EwoutH (commented):
Thanks, looks interesting!

EwoutH:
@quaquel can you reproduce the speedup of WolfSheep locally?

mesa/model.py
quaquel: its essential because it is the only place that contains the hard reference to agents. If that line is removed, agents are garbage collected and will magically disappear.

mesa/model.py
quaquel: I have been going back and forth on this. It all depends on which attributes of the agent registration stuff become part of the public api and which ones we want to keep protected.

quaquel:
> Performance benchmarks:
> 
> Model	Size	Init time [95% CI]	Run time [95% CI]
> Schelling	small	🔴 +77.0% [+76.3%, +77.6%]	🔵 -0.4% [-0.7%, -0.1%]
> Schelling	large	🔴 +94.2% [+92.9%, +95.3%]	🔵 -0.8% [-2.9%, +0.7%]
> WolfSheep	small	🔴 +130.1% [+128.1%, +131.9%]	🟢 -97.5% [-97.5%, -97.4%]
> WolfSheep	large	🔴 +130.8% [+129.1%, +132.4%]	🟢 -99.9% [-99.9%, -99.9%]
> BoidFlockers	small	🔴 +83.6% [+82.2%, +84.7%]	🔵 -0.3% [-1.0%, +0.6%]
> BoidFlockers	large	🔴 +83.1% [+82.0%, +84.1%]	🔵 -0.6% [-1.2%, +0.0%]

I'll try to figure out what is happening with the init times here. Its probably overhead from setting up the additional datastructures, which we get back when running the model (at least for wolf sheep?).

mesa/model.py
EwoutH: Another option would be to make a fast path in `self.agents.select(type=agenttype)`

Corvince:
Thanks for this PR, I like the way it encapsulates the logic much better. 

I think it is a good default to return the actual agentset and not just a copy. I think this is actually more intuitive.

Regarding the copy function, doesn't `agentset.select()` already do that? I'm not arguing against adding an additional method but I just want to mention that it's already possible and it should rightfully be discussed separately from this PR

quaquel:
> Regarding the copy function, doesn't agentset.select() already do that? I'm not arguing against adding an additional method but I just want to mention that it's already possible and it should rightfully be discussed separately from this PR

Yes, select does this as a corner case. But fair enought to seperate `copy` into a seperate PR

mesa/model.py
quaquel: That would be tricky because it would involve messing with the internal structure of AgentSet.

quaquel:
This is ready for a final review.

quaquel:
\trigger-benchmark

EwoutH:
Thanks for this. I think the regressions in init time are perfectably acceptable. What I'm curious about is why wolf_sheep has become slower. Because a lot of agents get added and removed?

I would like to get #2252 in first to get the full picture.

mesa/model.py
```diff
@@ -71,39 +77,95 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+        Notes:
+            This method is called automatically by ``Agent.__init__``, so there is no need to use this
+            if you are subclassing Agent and calling its super in the ``__init__`` method.
+
+        """
+        if not hasattr(self, "_agents"):
+            self._setup_agent_registration()
+
+            warnings.warn(
+                "The Mesa Model class was not initialized. In the future, you need to explicitly initialize "
```
rht, line 137: This should raise an error because in Mesa 3.0, the `super().__init__()` is no longer optional.

quaquel: I agree, but this change was not yet made in Model itself and it might be good to keep that in a separate PR

mesa/model.py
```diff
@@ -71,39 +77,95 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
             "Please adjust your code to use a different attribute name for custom agent storage."
         )
 
-        self._agents = agents
-
     @property
     def agent_types(self) -> list[type]:
         """Return a list of different agent types."""
-        return list(self.agents_.keys())
+        return list(self._agents_by_type.keys())
```
rht, line 102: The previous and the new change have different meaning altogether (list of all type of each agents vs list(set(agent types))). I suppose the docstring is ambiguous even though it is closer to the former meaning.

quaquel: That's not correct. In the old version `self.agents_` was the exact same datastructure as what is now `_agents_by_type`. In the old version: `self.agents_: defaultdict[type, dict] = defaultdict(dict)`. In the new version: `self._agents_by_type: dict[type, AgentSet] = {}`. So, in either case, keys is a list of types.

EwoutH:
Let me review it in depth tonight (I'm trying to stay in the model-building flow)

EwoutH:
Do we need to modernize `BoltzmannWealth` first to see the performance changes? Because 20% speedup in the large case with 10.000 agents doesn't seem that much.

quaquel:
This version of the model is on a grid. #2224 is with all agents in the model. So, I am not sure what to expect in the first place.

EwoutH:
Yeah so the Grid operations might be the bottleneck. Sounds logical.

I will try to update the benchmarks later tonight or tomorrow, and then we can bring this one in.

I'm also curious if we can speed up `shuffle()` at some point. With a model with 60.000 agents it's getting slow (also with `inplace=True`), but most notably, it's about the only AgentSet operation that's getting slow.

quaquel:
> I'm also curious if we can speed up shuffle() at some point. With a model with 60.000 agents it's getting slow (also with inplace=True), but most notably, it's about the only AgentSet operation that's getting slow.

In the case of `inplace=True`, the slowdown happens either when creating the list from the keys, when applying `random.shuffle`, or rebuilding the dict from the shuffled keys. None of these is easy to speedup any further.

mesa/model.py
rht: That means it still stands that the docstring is not clear enough.

mesa/model.py
EwoutH: Agreed, but separate PR.

mesa/model.py
```diff
@@ -71,39 +77,95 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+            KeyError: If agenttype does not exist
+
+
+        """
+        return self._agents_by_type[agenttype]
+
+    def _setup_agent_registration(self):
+        self._agents = {}
+        self._agents_by_type: dict[type, AgentSet] = {}
+        self._all_agents = AgentSet([], self)
```
EwoutH, line 120: Okay, so now we're keeping 4 records (if we count `self.agents`)? Could you explain a bit why each is needed?

I think this can be a place where some performance is lost, in models that frequently kill and create agents (like wolf-sheep)?

EwoutH (commented):
- I really like the model.register / deregister. It's very clear and cleanly separated, and the right way around.
- `model.agents` now is the ground truth right?
- I would really appreciate an overview of all the variables and private variables we now keep around agents, agent types and AgentSets.

mesa/model.py
quaquel: This docstring has not been changed by me in this PR, so it is already in the existing code. If you suggest changing it, that would be fine, but why would that hold up this PR?

Also, what do you suggest changing? The return is a list, and this list contains the types of the agents in the model. The only, minor, ambiguity is that it is a list of the unique types of agents, not the type for each agent in the model.

mesa/model.py
quaquel: self._agents contains the hard refs, so this is essential.

self._all_agents is the agentset with all agents. The main performance motivation for this PR. So again essential.

The only one you could debate is agents_by_type. This you could also do via `self._all_agents.groupby(types).groups` and thus as a property. What is better from a performance standpoint is hard to say. adding and removing stuff from dicts is relatively cheap to do.

quaquel:
> * `model.agents` now is the ground truth right?

no, model._agents is the hardref dict and the ground truth. 

> * I would really appreciate an overview of all the variables and private variables we now keep around agents, agent types and AgentSets.

can we wait with that until we remove all the old 2.x stuff from the model?

mesa/model.py
EwoutH: Thanks, appreciated. Could you add this in a comment or docstring behind/above each?

`self.agents` is the only thing we formally expose, all others are private and use-at-you-own risk.

For now I see the use case of a separate `_agents_by_type`. We can dive into the performance later.

And at least we got rid of `self.agents_`!

mesa/model.py
```diff
@@ -71,39 +77,95 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+
+        Raises:
+            KeyError: If agenttype does not exist
+
+
+        """
+        return self._agents_by_type[agenttype]
+
+    def _setup_agent_registration(self):
+        self._agents = {}
```
EwoutH, line 118: If you add a few lines of docstring here I'm good with it as is.

I trust on performance level you have tested extensively (and if we find regressions performance stuff can be fixed under the hood).

Thanks for the deep dive into this and fix for slow single/random agent selection!

EwoutH (approved):

mesa/model.py
quaquel: > self.agents is the only thing we formally expose, all others are private and use-at-you-own risk.

Which is exactly why I moved everything else into private. It keeps the public side the same as before. 

I'll add a few comments on the datastructures later today

EwoutH:
Should `register_agent()` and `deregister_agent()` be private methods?

quaquel:
If you want to get technical, I guess in a java sense they would be protected rather than private. They are not private because that would mean that agent could not call model.register_agent. They are not public in the sense of user facing.

Corvince:
In Python speak, however, it is either a public or an internal method. Public is everything user-faced, documented and guaranteed to be stable. So if there isn't any user facing use case for these functions I think they should indeed be made non-public by a leading underscore. This means its meant to be used internally, which is exactly whats happening here.

EwoutH:
Thanks for the explanation Jan, and I agree with Corvince.

PR #2252, opened 2024-08-28
benchmarks: Add BoltzmannWealth model

EwoutH:
BoltzmannWealth model to the benchmarks. It's probably the "best" test of Mesa that we have, since it does a lot of Mesa things and not a lot of other things.

A separate PR will update it to use AgentSet functionality, but I want to see the performance difference between that. This PR just adds the example model [from mesa-examples](https://github.com/projectmesa/mesa-examples/blob/1d67252a9f2abad051470a07a1607a2a3aa02b29/examples/boltzmann_wealth_model_experimental/model.py).

<hr>

@quaquel I noticed in that with the addition of discrete event scheduling in #2066 _all_ models were converted to DEVS. While I understand it was useful to test, and I think it's great to have that in the benchmark for one or two models, I think for all is a bit overkill as long as it's experimental.

So the novelty is clearly in WolfSheep, I would propose keeping that on DEVS and reverting the other two back to our regular activation.

### DIFF
--- a/benchmarks/BoltzmannWealth/boltzmann_wealth.py
+++ b/benchmarks/BoltzmannWealth/boltzmann_wealth.py
@@ -1,0 +1,78 @@
+# https://github.com/projectmesa/mesa-examples/blob/main/examples/boltzmann_wealth_model_experimental/model.py
+import mesa
+
+
+def compute_gini(model):
+    agent_wealths = [agent.wealth for agent in model.agents]
+    x = sorted(agent_wealths)
+    N = model.num_agents
+    B = sum(xi * (N - i) for i, xi in enumerate(x)) / (N * sum(x))
+    return 1 + (1 / N) - 2 * B
+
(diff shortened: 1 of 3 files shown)


quaquel:
Good idea. 

 > @quaquel I noticed in that with the addition of discrete event scheduling in https://github.com/projectmesa/mesa/pull/2066 all models were converted to DEVS. While I understand it was useful to test, and I think it's great to have that in the benchmark for one or two models, I think for all is a bit overkill as long as it's experimental.

>  So the novelty is clearly in WolfSheep, I would propose keeping that on DEVS and reverting the other two back to our regular activation.

Fine with me. Do we also want to have a mix of scheduler use and agentset use?

EwoutH:
Thanks, and yes.

Can you approve? Merging is blocked until.

quaquel (approved):

PR #2253, opened 2024-08-28
AgentSet: Allow selecting a fraction of agents in the AgentSet

EwoutH:
This PR updates the `select` method in the `AgentSet` class by replacing the `n` parameter with a more versatile `at_most` parameter. The `at_most` parameter allows for selecting either a specific number of agents or a fraction of the total agents when provided as an integer or a float, respectively. Additionally, backward compatibility is maintained by supporting the deprecated `n` parameter, which will trigger a warning when used.

### Motive
Previously, the `select` method only allowed users to specify a fixed number of agents (`n`) to be selected. The new `at_most` parameter extends this functionality by enabling the selection of agents based on a proportion of the total set, which is particularly useful in scenarios where relative selection is desired over absolute selection.

### Implementation
- **`at_most` Parameter:** 
  - Accepts either an integer (to select a fixed number of agents) or a float between 0.0 and 1.0 (to select a fraction of the total agents).
  - `at_most=1` selects one agent, while `at_most=1.0` selects all agents.
  - If a float is provided, it determines the maximum fraction of agents to be selected from the total set. It rounds down to the nearest number of whole agents.
- **Backward Compatibility:**
  - The deprecated `n` parameter is still supported, but it now serves as a fallback for `at_most` and triggers a deprecation warning.
- **Behavior Notes:**
  - `at_most` serves as an upper limit on the number of selected agents. If additional filtering criteria are provided, the final selection may include fewer agents.
  - For random sampling, users should shuffle the `AgentSet` before applying `at_most`.

### Usage Examples
```python
# Select the first 5 agents from the AgentSet
selected_agents = agents.select(at_most=5)
selected_agents = agents.select(n=5)  # Still works but throws a deprecation warning
```
```python
# Select the first 20% of agents from the AgentSet (as currently sorted, rounded down)
selected_agents = agents.select(at_most=0.2)
```
To randomly select a fraction, add a `shuffle()`:
```python
# Select 20% of agents randomly from the AgentSet
random_agents = agents.shuffle().select(at_most=0.2)
```
Combining with sorting:
```python
# Select the 20% of agents with the lowest wealth
selected_agents = agents.sort("wealth", ascending=True).select(at_most=0.2)
```
The most powerful feature is that you can combine `at_most` with additional criteria:
```python
# Select agents with "wealth" less than 5, and at most 20% of the total
selected_agents = agents.select(lambda agent: agent.wealth < 5, at_most=0.2)

# First filter agents, then select 20% of those remaining
filtered_agents = agents.select(lambda agent: agent.wealth < 5).select(at_most=0.2)
```
You can also use it with chaining:
```Python
# Randomly select 40% of the agents from the AgentSet and set a value
model.agents.shuffle().select(at_most=0.4).set('has_license', True)
```

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -121,10 +121,10 @@ class AgentSet(MutableSet, Sequence):
     def select(
         self,
         filter_func: Callable[[Agent], bool] | None = None,
-        n: int = 0,
-        fraction: float = 1.0,
+        max: int | float = float("inf"),
         inplace: bool = False,
         agent_type: type[Agent] | None = None,
+        n: int | None = None,
     ) -> AgentSet:
         """
         Select a subset of agents from the AgentSet based on a filter function and/or quantity limit.
@@ -132,7 +132,9 @@ class AgentSet(MutableSet, Sequence):
         Args:
             filter_func (Callable[[Agent], bool], optional): A function that takes an Agent and returns True if the
                 agent should be included in the result. Defaults to None, meaning no filtering is applied.
-            n (int, optional): The number of agents to select. If 0, all matching agents are selected. Defaults to 0.
+            max (int | float, optional): The maximum amount of agents to select. Defaults to infinity.
+              - If an integer of 1 or larger, the first n matching agents are selected.
+              - If a float between 0 and 1, at most that fraction of original the agents are selected.
             fraction (float, optional): The fraction of agents to select. If 0.2, it selects 20% of Agents in the AgentSet. Defaults to 1.0.
             inplace (bool, optional): If True, modifies the current AgentSet; otherwise, returns a new AgentSet. Defaults to False.
             agent_type (type[Agent], optional): The class type of the agents to select. Defaults to None, meaning no type filtering is applied.
@@ -140,38 +142,38 @@ class AgentSet(MutableSet, Sequence):
         Returns:
             AgentSet: A new AgentSet containing the selected agents, unless inplace is True, in which case the current AgentSet is updated.
 
-        Raises:
-            ValueError: If both n and fraction are set.
-
         Notes:
-            - n and fraction just return the first n or fraction of agents. To take a random sample, shuffle() beforehand.
-            - n and fraction are upper limits. When specificing other criteria it can be smaller.
-            - A fraction is taken of the original AgentSet length, not the finally selected one.
+            - max just return the first n or fraction of agents. To take a random sample, shuffle() beforehand.
+            - max is an upper limit. When specifying other criteria it can be smaller.
         """
+        if n is not None:
+            warnings.warn(
+                "The parameter 'n' is deprecated. Use 'max' instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            max = n
 
-        if filter_func is None and agent_type is None and n == 0 and fraction == 1.0:
+        inf = float("inf")
+        if filter_func is None and agent_type is None and max == inf:
             return self if inplace else copy.copy(self)
 
-        if n != 0 and fraction != 1.0:
-            raise ValueError(
-                "Cannot set both n and fraction. Please set only one of these parameters."
-            )
+        # Check if max is of type float
+        if max <= 1.0 and isinstance(max, float):
+            max = round(len(self) * max)
 
-        if fraction < 1.0:
-            n = round(len(self) * fraction)
-
-        def agent_generator(filter_func=None, agent_type=None, n=0):
+        def agent_generator(filter_func, agent_type, max):
             count = 0
             for agent in self:
+                if count >= max:
+                    break
                 if (not filter_func or filter_func(agent)) and (
                     not agent_type or isinstance(agent, agent_type)
                 ):
                     yield agent
                     count += 1
-                    if 0 < n <= count:
-                        break
 
-        agents = agent_generator(filter_func, agent_type, n)
+        agents = agent_generator(filter_func, agent_type, max)
 
         return AgentSet(agents, self.model) if not inplace else self._update(agents)
 
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -122,38 +122,41 @@ class AgentSet(MutableSet, Sequence):
         self,
         filter_func: Callable[[Agent], bool] | None = None,
         n: int = 0,
-        p: float = 1.0,
+        fraction: float = 1.0,
         inplace: bool = False,
         agent_type: type[Agent] | None = None,
     ) -> AgentSet:
         """
         Select a subset of agents from the AgentSet based on a filter function and/or quantity limit.
-
+    
         Args:
             filter_func (Callable[[Agent], bool], optional): A function that takes an Agent and returns True if the
                 agent should be included in the result. Defaults to None, meaning no filtering is applied.
             n (int, optional): The number of agents to select. If 0, all matching agents are selected. Defaults to 0.
-            p (float, optional): The fraction of agents to select. If 0.2, it selects 20% of Agents in the AgentSet. Defaults to 1.0.
+            fraction (float, optional): The fraction of agents to select. If 0.2, it selects 20% of Agents in the AgentSet. Defaults to 1.0.
             inplace (bool, optional): If True, modifies the current AgentSet; otherwise, returns a new AgentSet. Defaults to False.
             agent_type (type[Agent], optional): The class type of the agents to select. Defaults to None, meaning no type filtering is applied.
-
+    
         Returns:
             AgentSet: A new AgentSet containing the selected agents, unless inplace is True, in which case the current AgentSet is updated.
+    
+        Raises:
+            ValueError: If both n and fraction are set.
 
         Note:
-            n and p can't both be set at the same time, set one at most.
+            n and fraction just return the first n or fraction of agents. To take a random sample, shuffle() beforehand.
         """
 
         if filter_func is None and agent_type is None and n == 0 and p == 1.0:
             return self if inplace else copy.copy(self)
 
-        if p != 1.0 and n != 0:
+        if n != 0 and fraction != 1.0:
             raise ValueError(
-                "Cannot set both n and p. Please set only one of these parameters."
+                "Cannot set both n and fraction. Please set only one of these parameters."
             )
 
-        if p < 1.0:
-            n = round(len(self) * p)
+        if fraction < 1.0:
+            n = round(len(self) * fraction)
 
         def agent_generator(filter_func=None, agent_type=None, n=0):
             count = 0
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -139,10 +139,16 @@ class AgentSet(MutableSet, Sequence):
 
         Returns:
             AgentSet: A new AgentSet containing the selected agents, unless inplace is True, in which case the current AgentSet is updated.
+
+        Note:
+            n and p can't both be set at the same time, set one at most.
         """
 
         if filter_func is None and agent_type is None and n == 0 and p == 1.0:
             return self if inplace else copy.copy(self)
+
+        if p != 1.0 and n != 0:
+            raise ValueError("Cannot set both n and p. Please set only one of these parameters.")
 
         if p < 1.0:
             n = round(len(self) * p)
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -147,7 +147,7 @@ class AgentSet(MutableSet, Sequence):
             n and fraction just return the first n or fraction of agents. To take a random sample, shuffle() beforehand.
         """
 
-        if filter_func is None and agent_type is None and n == 0 and p == 1.0:
+        if filter_func is None and agent_type is None and n == 0 and fraction == 1.0:
             return self if inplace else copy.copy(self)
 
         if n != 0 and fraction != 1.0:
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -148,7 +148,9 @@ class AgentSet(MutableSet, Sequence):
             return self if inplace else copy.copy(self)
 
         if p != 1.0 and n != 0:
-            raise ValueError("Cannot set both n and p. Please set only one of these parameters.")
+            raise ValueError(
+                "Cannot set both n and p. Please set only one of these parameters."
+            )
 
         if p < 1.0:
             n = round(len(self) * p)
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -133,9 +133,8 @@ class AgentSet(MutableSet, Sequence):
             filter_func (Callable[[Agent], bool], optional): A function that takes an Agent and returns True if the
                 agent should be included in the result. Defaults to None, meaning no filtering is applied.
             at_most (int | float, optional): The maximum amount of agents to select. Defaults to infinity.
-              - If an integer of 1 or larger, the first n matching agents are selected.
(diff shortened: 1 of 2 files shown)


quaquel:
what is the motivation for adding this to the agentset?

Corvince:
That seems useful, thanks! 

The only worry I have is how this behaves if a user specifies both n and p. That probably should raise an error?

Or maybe there is a good name that could incorporate both p and n? So if it is between 0 and 1 use a fraction and if it is a whole number above 1 use that number?

EwoutH:
> what is the motivation for adding this to the agentset?

Sorry, was still working on other features (and my actual model), wrote it up.

> That seems useful, thanks!

> The only worry I have is how this behaves if a user specifies both n and p. That probably should raise an error?

Yeah I was thinking about that. Maybe just don't do that (and we mention it in the docstring)?

If you just want to select a fraction of `n`, you can do `n=round(n*p)`, so having both doesn't make sense.

> Or maybe there is a good name that could incorporate both p and n? So if it is between 0 and 1 use a fraction and if it is a whole number above 1 use that number?

Very interesting idea, but maybe in this case explicit is better than implicit. Except if you can come up with a killer name.

quaquel:
I like the clarity of p. So my suggestion would be to raise a value error if both n and p are passed

mesa/agent.py
```diff
@@ -132,16 +133,28 @@ def select(
             filter_func (Callable[[Agent], bool], optional): A function that takes an Agent and returns True if the
                 agent should be included in the result. Defaults to None, meaning no filtering is applied.
             n (int, optional): The number of agents to select. If 0, all matching agents are selected. Defaults to 0.
+            p (float, optional): The fraction of agents to select. If 0.2, it selects 20% of Agents in the AgentSet. Defaults to 1.0.
```
quaquel, line 136: p suggests random, which it not really is because it is based on the order in the underlying dict. 

My suggestion is to add something here stating that if you want to ensure randomness, first shuffle in place. Could also be covered by a note.

mesa/agent.py
```diff
@@ -132,16 +133,28 @@ def select(
...
             inplace (bool, optional): If True, modifies the current AgentSet; otherwise, returns a new AgentSet. Defaults to False.
             agent_type (type[Agent], optional): The class type of the agents to select. Defaults to None, meaning no type filtering is applied.
 
         Returns:
             AgentSet: A new AgentSet containing the selected agents, unless inplace is True, in which case the current AgentSet is updated.
+
+        Note:
```
quaquel, line 143: would turn this into Raises: value error if n and p are passed at the same time.

mesa/agent.py
```diff
@@ -132,16 +133,28 @@ def select(
...
+            n and p can't both be set at the same time, set one at most.
         """
 
-        if filter_func is None and agent_type is None and n == 0:
+        if filter_func is None and agent_type is None and n == 0 and p == 1.0:
             return self if inplace else copy.copy(self)
 
+        if p != 1.0 and n != 0:
+            raise ValueError(
+                "Cannot set both n and p. Please set only one of these parameters."
```
quaquel, line 152: suggest changing to cannot **pass** both n and p.

quaquel:
see the few minor comments and once unit tests are added, this is good to go.

mesa/agent.py
EwoutH: Yeah I'm constantly (almost) forgetting to shuffle before select.

With `n` and `p`, it almost seems like shuffle should be the default. Except of course when you sort first, then you want a sorted number/percentage.

Corvince: Based on this discussion I would suggest renaming it to frac or fraction, based on how its called in pandas or polars, respectively. I would prefer fraction

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.sample.html
https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.sample.html

EwoutH: Agreed!

EwoutH:
Okay, I:
- Changes `p` to `fraction`
- Used the ValueError
- Updated the other docstring, including notes
- Added tests
- Updated the examples

<hr>

_However_, I noticed that there's an important difference between `n` and `fraction`. `n` is always fixed, it's just an upper limiter. `fraction` does matter _when_ you apply it, before or after the rest of the selection.

Currently `fraction` is interpreted as a `fraction` of the _input_ AgentSet. When writing the usage examples that felt really counter intuitive. It would be more logical if you could apply it afterwards, such that a fraction of the _selected_ AgentSet is returned.

Why? Because if you take these two use cases:
- Select the agents with "wealth" less than 5 but at most 20% of total agents
- Select the agents with "wealth" less than 5, and then 20% of those agents

The latter is used way more than the former. And it will be _way_ more logical if you select by type.

So I would suggest applying fraction afterwards, on the selected AgentSet after all other operations are done. Then you could still do both:
```Python
# Select the agents with "wealth" less than 5, and at most 20% of total agents
agents.select(fraction=0.2).select(lambda agent: agent.wealth < 5)

# Select the agents with "wealth" less than 5, and then 20% of those agents
agents.select(lambda agent: agent.wealth < 5, fraction=0.2)
# or, equivalently:
agents.select(lambda agent: agent.wealth < 5).select(fraction=0.2)
```
But now the one that's more used and more intuitive will go well by default.

<hr>

Totally other options could be:
- Don't allow `fraction` and/or `n` with other functions, but enforce chaining
- Introduce a new method, like `sample`, that give a sample of `n` or a sample of `fraction`.

mesa/agent.py
EwoutH: Done

rht:
> what is the motivation for adding this to the agentset?

@EwoutH  I'm also wondering about this. Not saying that this shouldn't be in the library, but a concrete example could give some illustration. Is this used in your project?

EwoutH:
This was the thing I wanted to do:
```Python
# Randomly select 40% of the agents from the AgentSet and give them a license
model.agents.shuffle().select(fraction=0.4).set('has_license', True)
```
I needed to do this:
```Python
n_license = round(model.agents * license_chance)
model.agents.shuffle().select(n=n_license).do(lambda agent: setattr(agent, 'has_license', True))
```
With #2254 it got simplified to:
```Python
n_license = round(model.agents * license_chance)
model.agents.shuffle().select(n=n_license).set('has_license', True)
```
It's not a huge use case, but it's nice. Especially that you don't need to break the chain.

Combine it with a function and it get's really powerful though. Assume I want to distribute some cars around (I know a certain percentage of all people has a car), but only to agents with licenses.

```Python
agents.select(lambda a: a.has_license, fraction=car_chance).set('has_car', True)
```
Without the fraction, this would have been:
```Python
n_car = round(model.agents * car_chance)
model.agents.shuffle().select(n=n_car ).set('has_license', True)
```
So yeah, it's not a huge use case. Maybe it adds some complexity.

<hr>

There's an unique application for fraction as upper limit (cap), as currently implemented, and a unique application for doing it afterwards. I need to think about this a bit longer.

EwoutH:
Right, `n=0` has a special status. With a small `fraction` or small agentset, `n` can become `0`, returning all agents.

Corvince:
> Right, `n=0` has a special status. With a small `fraction` or small agentset, `n` can become `0`, returning all agents.

Good catch! 

I see two possibilities now. Either just change the special meaning from 0 to -1. I don't know if there was a good use case for 0, but it's rather strange for 0 to indicate all agents.

The more holistic approach would be to split select into a filter function and a sample function. This would also simplify the logic and solve the "before or after" question (which was present but unconsidered before fraction was introduced)

EwoutH:
The brain is so interesting that after a nights sleep you look at it again and you think _oh_, and it all clicks together.

Now I just have to write it up, rewrite the codes, tests and examples.

Can’t wait for 2026/2027 where with a voice message a bit does that automatically.

Long story short: There’s a special use case for when filtering, you want a certain number or fraction _at most_. Especially the fraction should happen right there in the function, because after the function is done, you don’t know how large the 

For all other cases (before, after) a `sample` method would be perfect (and can be implemented pretty fast I think). `sample` could also draw a random sample, where `select` selects the first n/fraction.

> Or maybe there is a good name that could incorporate both p and n? So if it is between 0 and 1 use a fraction and if it is a whole number above 1 use that number?

Obviously the way to go. I was thinking `max`, `limit`, `ceiling` or `at_most`.

quaquel:
Looking forward at what @EwoutH comes up with. But I like @Corvince suggestion of having `filter` and `sample`. From a pure performance standpoint, minimizing the looping required for the use cases described here would be really beneficial. Chaining multiple `select` or `shuffle` and `select`, for something like "give me a random sample of 40% of the agents that have a particular attribute" is not ideal. It requires multiple loops where 1 should be sufficient. In particular, in the case of large numbers of agents, this becomes very inefficient. 

Of course, a user could do most of this stuff with a well designed custom function passed to e.g. select so performance is not the only design concern and their are already clean ways of handling those.

quaquel:
> Agreed on the performance aspect. One way to solve this but keep the chainable approach would be to use generator functions to return iterators instead of the complete AgentSet. But maybe as you said this is all mainly catered towards nice semantics and there are other ways already available for performance critical operations.

That's an interesting idea worth exploring at some point (but not this PR). Basically, what if we have a generator interface to an AgenSet? And can we make a chainable API work with generators?

EwoutH:
It seems we keep coming back to this (https://github.com/projectmesa/mesa/pull/2220#issuecomment-2297117745), so it’s certainly worth exploring at some point.

Corvince:
> > Agreed on the performance aspect. One way to solve this but keep the chainable approach would be to use generator functions to return iterators instead of the complete AgentSet. But maybe as you said this is all mainly catered towards nice semantics and there are other ways already available for performance critical operations.
> 
> That's an interesting idea worth exploring at some point (but not this PR). Basically, what if we have a generator interface to an AgenSet? And can we make a chainable API work with generators?
> 

I think having an __iter__ method is kind of enough, so

```python
(agent for agent in agentset)
```

should already give you an iterator over the agentset. Definitely worth exploring that more, but certainly way out of scope for this PR

EwoutH:
I updated this PR to replace `n` with `max`.

> max (int | float, optional): The maximum amount of agents to select. Defaults to infinity.
>   - If an integer of 1 or larger, the first n matching agents are selected.
>   - If a float between 0 and 1, at most that fraction of original the agents are selected.

Some details:
- `max=1` will give one agent, `max=1.0` gives all agents.
- A fallback for `n` was added, which does `max = n` and throws a warning.

Tests are updated. Please double check the internal `agent_generator` function.

If we decide this is the way to go, I will update the PR description.

quaquel:
I am unsure about using a single keyword for both the number and the percentage, but I won't object to it either. I would change the name, however. `max` shadows the name of a build-in. 

It would be nice to see a quick overview of what the API is now becoming just for clarity.

> sample(n, shuffle=True) will be equivalent to NetLogo's [up-to-n-of](https://ccl.northwestern.edu/netlogo/docs/dict/up-to-n-of.html). @quaquel I know you hate NetLogo with all your hearth, but sometimes you can learn a lot from them ;).

I hate the language, but, yes, we can pick up useful ideas and give them a better name. `sample` is much better than that weird construct with hyphens in the name 😉.

EwoutH:
> I was thinking `max`, `limit`, `ceiling` or `at_most`.

Any suggestions (either these or another)?

Corvince:
I like `at_most` the best. It conveys that "n" can be arbitrary large, but must the number of returned agents must not match. It also sort of implies that you first apply a filter and then take a sample. And it also makes the rounding clear for fractions. So 1/3 of 5 (1.67) will be 1 agent, otherwise it would be more than 1/3.

EwoutH:
> So 1/3 of 5 (1.67) will be 1 agent, otherwise it would be more than 1/3.

Currently it does round, do you think it shouldn't?

Corvince:
If its an upper limit I think it should always round down/floor

EwoutH:
Difficult one. Because if you describe it as "selecting a fraction" I would expect it to select the closest match.

I think in many practical scenarios the closest selection to the fraction you wanted is most logical.

quaquel:
If we go with `at_most`, it should round down in the case of fractions. Otherwise, the name and behavior don't match.

Corvince:
Valid argument for "selecting a fraction", but for selecting "at most" 33% I would not expect it to select 40%

Corvince:
> If we go with `at_most`, it should round down in the case of fractions. Otherwise, the name and behavior don't match.

Exactly. Thats why I think its a good name (if we floor), because people will always have different expectations for "selecting a fraction" with respect to rounding.

EwoutH:
I renamed `max` to `at_most`, made sure it rounded down, and updated the tests.

EwoutH:
PR description is updated, including the usage examples

mesa/agent.py
```diff
@@ -121,39 +121,59 @@ def __contains__(self, agent: Agent) -> bool:
...
     ) -> AgentSet:
         """
         Select a subset of agents from the AgentSet based on a filter function and/or quantity limit.
 
         Args:
             filter_func (Callable[[Agent], bool], optional): A function that takes an Agent and returns True if the
                 agent should be included in the result. Defaults to None, meaning no filtering is applied.
-            n (int, optional): The number of agents to select. If 0, all matching agents are selected. Defaults to 0.
+            at_most (int | float, optional): The maximum amount of agents to select. Defaults to infinity.
+              - If an integer of 1 or larger, the first n matching agents are selected.
```
Corvince, line 136:
```suggestion
- If an integer, at most the first number of matching agents are selected.

File: mesa/agent.py [hunk 5]
```diff
@@ -121,39 +121,59 @@ def __contains__(self, agent: Agent) -> bool:
...
+              - If a float between 0 and 1, at most that fraction of original the agents are selected.
+            fraction (float, optional): The fraction of agents to select. If 0.2, it selects 20% of Agents in the AgentSet. Defaults to 1.0.
```
Corvince, line 138:
```suggestion

```

mesa/agent.py
EwoutH: Thanks, done!

mesa/agent.py
EwoutH: Good catch, done

EwoutH:
@projectmesa/maintainers ready to go?

mesa/agent.py
```diff
@@ -121,39 +121,58 @@ def __contains__(self, agent: Agent) -> bool:
...
+              - If a float between 0 and 1, at most that fraction of original the agents are selected.
             inplace (bool, optional): If True, modifies the current AgentSet; otherwise, returns a new AgentSet. Defaults to False.
             agent_type (type[Agent], optional): The class type of the agents to select. Defaults to None, meaning no type filtering is applied.
 
         Returns:
             AgentSet: A new AgentSet containing the selected agents, unless inplace is True, in which case the current AgentSet is updated.
+
+        Notes:
+            - at_most just return the first n or fraction of agents. To take a random sample, shuffle() beforehand.
+            - at_most is an upper limit. When specifying other criteria it can be smaller.
```
quaquel, line 146: Probably add here how `at_most` interacts with other criteria as per your last usage example in the PR text. t

Corvince (approved):

quaquel (approved):

EwoutH:
(keeping the branch in case of regressions)

PR #2254, opened 2024-08-28
AgentSet: Add `set` method

EwoutH:
This PR adds a `set` method to the `AgentSet` class that allows setting a specified attribute of all agents within the set to a given value. This method provides a streamlined way to update agent attributes in bulk, without having to resort to `lambda` or other callable functions.

### Motive
The motivation behind this feature is to simplify and expedite the process of modifying attributes across all agents in an `AgentSet`. Previously, this required iterating manually over the agents or using a `lambda` or other callable function. This new method makes the operation more elegant.

### Implementation
The `set` method was added to the `AgentSet` class. It iterates through all agents in the set and applies the `setattr` function to assign the specified value to the desired attribute. The method operates in place, modifying the existing agents directly.

The (now updated) AgentSet itself is returned, which allows for chaining.

You can both set existing Agent variables or set new ones.

### Usage Examples
```python
# Set the 'has_license' attribute to True for all agents in the AgentSet
agents.set('has_license', True)
```
Note that it doesn't matter if `has_license` already exists in the Agent, both work.

Together with #2253, you can now set a value for a fraction of your AgentSet:
```Python
# Select 40% of the agents from the AgentSet
model.agents.select(p=0.4).set('has_license', True)
```
It's also possible to set multiple things by chaining the `set()` commands:
```python
agents.set('has_license', True).set('age', 45).set('continent', 'Europe')
```
### Additional Notes
It feels logical that if you have a `get` method, you also should have a `set` method.

**Edit:**
- Modified to return an AgentSet, so you can keep the output or keep chaining.
- It modified the Agents itself, like with `do()`, so inplace isn't needed.
- We don't allow dicts, those can be added later if needed.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -323,7 +323,7 @@ class AgentSet(MutableSet, Sequence):
         Args:
             attr_name (str): The name of the attribute to set.
             value (Any): The value to set the attribute to.
-    
+
         Returns:
             AgentSet: The AgentSet instance itself, after setting the attribute.
         """
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -316,16 +316,20 @@ class AgentSet(MutableSet, Sequence):
                 for agent in self._agents
             ]
 
-    def set(self, attr_name: str, value: Any) -> None:
+    def set(self, attr_name: str, value: Any) -> AgentSet:
         """
         Set a specified attribute to a given value for all agents in the AgentSet.
 
         Args:
             attr_name (str): The name of the attribute to set.
             value (Any): The value to set the attribute to.
+    
+        Returns:
+            AgentSet: The AgentSet instance itself, after setting the attribute.
         """
         for agent in self:
             setattr(agent, attr_name, value)
+        return self
 
     def __getitem__(self, item: int | slice) -> Agent:
         """
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -316,6 +316,17 @@ class AgentSet(MutableSet, Sequence):
                 for agent in self._agents
             ]
 
+    def set(self, attr_name: str, value: Any) -> None:
+        """
+        Set a specified attribute to a given value for all agents in the AgentSet.
+
+        Args:
+            attr_name (str): The name of the attribute to set.
+            value (Any): The value to set the attribute to.
+        """
+        for agent in self:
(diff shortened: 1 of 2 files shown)


quaquel:
good idea, and probably an inplace keyword should be included preferably in this PR. It keeps the API clear and consistent across methods.

EwoutH:
Can we think of cases where you want to continue the chain and return an AgentSet? In that case, should a default `return=False` which you can switch to `True` be also included?

Maybe I'm now mixing things up with in place.

EwoutH:
I see three scenarios here:
1. I want to modify my agentset inplace
2. I want to modify my agentset inplace and return that updated agentset (to assign or continue chaining)
3. I want to create a new agentset (inplace) and return that new agentset (to assign or continue chaining)

Of course you can just ignore the return for 1. So return is not really needed, just return it and the user can decide if the y use it or not. Okay, I get it.

So we probably need both

Now we're in a bit of a pickle, because with `.set()` `inplace=True` seems like a good default. But most other functions have it `False` by default.

I think the most similar function is `.do()`. That also only does it only in place. This is basically a `.do()` with less steps.

To summarize:
- Always return the new AgentSet
- We need to pick one of two options:
  1. Like `do()`, and just always do it "in place"
  2. Add an `inplace` defaulting to `True`

quaquel:
good analysis. There is no point to return a new agent set, just as with do.

Corvince (approved):
Nice and simple 🔝

EwoutH:
Thanks, I will add a few unittest tonight and merge.

EwoutH:
What I just now realized and what's quite powerful is that you also can set completely new variables this way.

Punching above it's weight!

quaquel:
let me check if I understand this correctly, you want to set an agent attribute on the basis of another agent attribute?

EwoutH:
No on the basis of it's own agent attribute. I updated the `available_modes` attribute for all agents, and I just want to copy that to `currently_available_modes` for each agent, before starting all trips.

It would be nice to be able to refer to the agent itself without the whole lambda thing. Not only in `set()`, but also in some other methods.

(not high priority)

quaquel:
Sorry, that is what I was trying to say, and, indeed, I agree. Would be a good issue to open and a feature to add at some point.

Issue #2233, opened 2024-08-21
Write Mesa 3.0 migration guide


EwoutH:
_**Update:** The [current version](https://github.com/projectmesa/mesa/blob/main/docs/migration_guide.md) of the migration guide is now [available on Read the docs](https://mesa.readthedocs.io/en/latest/migration_guide.html)!_

It would be very useful have a Mesa 3.0 migration guide. It will also help to start writing it soon, so users have time to read it and update while we're still in alpha/beta.

The _migration guide_ is purely for current functionatity that's removed or changed in an backwards-incompatible way. The _release notes_ will be for new features that don't break older models.

### Topics

- [x] Reserved variables (https://github.com/projectmesa/mesa/discussions/2230)
  - https://github.com/projectmesa/mesa/pull/2225
- [x] Model initialization requirement (https://github.com/projectmesa/mesa/pull/2218)
- [x] Flat namespace removal (https://github.com/projectmesa/mesa/pull/2091)
- [x] Automatic assignment of `unique_id` to Agents
  - https://github.com/projectmesa/mesa/pull/2226
- [x] AgentSet (and model.agents)
  - https://github.com/projectmesa/mesa/pull/2225
  - https://github.com/projectmesa/mesa/pull/2219
  - https://github.com/projectmesa/mesa/pull/2237
  - https://github.com/projectmesa/mesa/pull/2251
- [x] Time and schedulers:
  - https://github.com/projectmesa/mesa/pull/2223
  - https://github.com/projectmesa/mesa-examples/pull/183
- [x] Visualisation (https://github.com/projectmesa/mesa/pull/2133 and https://github.com/projectmesa/mesa/pull/2180

_This list may be incomplete, feel free to extend_

A lot of information is already in the PR descriptions themselves. There are also some Mesa-examples updates that could help:
- https://github.com/projectmesa/mesa-examples/pull/153
- https://github.com/projectmesa/mesa-examples/pull/155
- https://github.com/projectmesa/mesa-examples/pull/161

### PRs
- https://github.com/projectmesa/mesa/pull/2257

EwoutH:
I think the main thing is written, so closing as completed!

EwoutH:
@quaquel suggested that the seed need to be passed to model when calling super().

PR #2257, opened 2024-08-30
Write initial Mesa Migration guide

EwoutH:
A rough outline for a Mesa Migration guide, including Mesa 3.0.

Already wrote something on:
- Reserved and private variables
- Removal of `mesa.flat` namespace
- Automatic increase of the `steps` counter.

Added headers and TODOs for other topics. This allows updating it continuously as we further development.

I think I correctly added it to the Readthedocs index, but please check.

Tracking issue for the Migration guide:
- https://github.com/projectmesa/mesa/issues/2233

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -81,6 +81,7 @@ ABM features users have shared that you may want to use in your model
 Mesa Overview <overview>
 tutorials/intro_tutorial
 tutorials/visualization_tutorial
+Migration guide <migration_guide>
 Best Practices <best-practices>
 How-to Guide <howto>
 API Documentation <apis/api_main>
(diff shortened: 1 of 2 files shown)


docs/index.md
```diff
@@ -81,6 +81,7 @@ ABM features users have shared that you may want to use in your model
 Mesa Overview <overview>
 tutorials/intro_tutorial
 tutorials/visualization_tutorial
+Migration guide <migration_guide>
```
tpike3, line 84: This is great! My overarching reaction is to make it really stand out and make it explicit. 

Keep it in the index but also like around line 57 add it as a hyperlink--- maybe add it up top right under the header to make it easy for users and to stand out -- I would also recommend we are more explicit (e.g. Mesa 2 -> 3 Migration Guide)

EwoutH: Maybe start here, and then push it wide when it's (nearly) finished?

quaquel (approved):

PR #2260, opened 2024-08-30
track unique_id automatically

quaquel:
This PR ensures that the `unique_id` in Agent is assigned automatically. It draws on work done in #2226. In short, this PR makes 3 changes: 

- `Agent.unique_id` is assigned automatically and no longer used as an argument to Agent. A deprecation warning is issued if the Agent is instantiated with 2 arguments (`unique_id` and `model`). The value passed for `unique_id` is ignored in favor of the new system. This last point is in line with #2226.
- `Model.next_id()` raises a deprecation warning and always return 0. 
-  `unique_id` is unique relative to a Model instance. Internally, Agent has a class-level attribute `_ids`. This attribute is a dictionary with the Model instance as key and the "generator" for `unique_id` as value.

All unit tests and benchmarks work with the new structure. They raise no deprecation warnings. 

examples, tutorials, etc., still need to be updated to be consistent with this change.


mesa/agent.py
```diff
@@ -37,14 +38,28 @@ class Agent:
...
+        if len(args) == 1:
+            model = args[0]
+            unique_id = next(self._ids)
+        else:
+            warnings.warn(
+                "unique ids are assigned automatically in MESA 3",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            unique_id, model, *args = args
```
EwoutH, line 60: Now you’re allowing users to assign their own unique_id right, isn’t that exactly what we don’t want?

It does make the “break” smaller though. But if we’re going to replace the warning before 3.0, what’s the point?

EwoutH: Like normally I would like this: first a PR that introduces a warning, and then a few minor releases later we remove the deprecated behavior.

But now we’re jumping to 3.0, and PR this allows two different things at once. We can’t trust on the unique id being unique.

This might be useful as a 2.4 backport to warn for new behavior.

quaquel: > Like normally I would like this: first a PR that introduces a warning, and then a few minor releases later we remove the deprecated behavior.

You equate a single PR with a release. My suggestion would be first to do this PR, then fix all deprecation warnings in the tests as a PR, and then remove this `else` statement in a PR. Once these 3 atomic PRs are done, this entire feature is ready for a next release. 

> But now we’re jumping to 3.0, and PR this allows two different things at once. We can’t trust on the unique id being unique.

I am sorry, but I just don't follow. Yes, this code allows the old code to still function but with a deprecation warning that allows us to clean up the rest of the code base. Once the rest of the code base is cleaned up, this can also be removed. 

I don't understand what any off this has to do with trusting unique_id to be truly unique. The current 2.x code also cannot guarantee that.

EwoutH: Most of the errors are not in the structure of the arguments, but in the values themselves. A lot of our tests demanded very specific `unique_id` values. I fixed those tests and that’s #2226.

Corvince:
What I don't understand, but unfortunately can't test right now: doesn't this keep increasing the counter indefinitely, even across models? So if I run the same model twice without restarting my kernel, won't the ids be different?

mesa/agent.py
```diff
@@ -36,15 +37,26 @@ class Agent:
         model (Model): A reference to the model instance.
         self.pos: Position | None = None
     """
+    _ids = itertools.count()
```
wang-boyu, line 41: I'm a bit confused - are agent ids unique at the model level, or at the level of each Agent type?

As a concrete example, suppose a model have agents of type A and type B. Are we allowing Type A agent with id 0 and Type B agent with id 0 at the same time?

EwoutH:
I would like all maintainers to check our Matrix chat before moving further.

mesa/agent.py
quaquel: Just to clarify the code, and I am not hung up on the implementation details at all, `_ids` is a class attribute that is incremented using `next`. So, if this is used to get a unique_attribute within an instance as in `self.unique_id = next(self._ids)`, what happens is that `self._ids` via the MRO is resolved into the class level attribute and so returns the next number. We assign this to a instance attribute. Net result is that all subclasses of Agent will be counted and have unique numbers. 

So. 
1. With this implementation, there will be only one Agent Zero. 
2. Yes, as pointed out by @Corvince,  if you instantiate multiple models in a row, the count just keeps increasing and  is not reset with each newly instantiated model. In my view, that is actually desirable behavior (But can indeed be tricky for testing).

wang-boyu: Thanks for your clarification.

It is indeed very tricky. Not very certain about the differences between 2) and 3), so I might be wrong about this - from the perspective of reproducibility, is 2) better?

Basically when we re-run a model, the agent ids are unique but stay the same as previous runs (such as a uuid that can be reproduced with a random seed). Or is it too much to ask?

EwoutH: I’m really fine with what everyone decides, personally

> Should it be unique to 1 Model instance?

would be enough for me.

quaquel:
I continued some work on this, mainly to understand how breaking automatic counting actually is. And it indeed creates a massive mess throughout the various tests, including very old test code that probably has not been touched in a long while (e.g., the test code of the schedulers). 

I did 3 things here
* `_ids` has become essentially a defaultdict with the model as key and the intertools.count(1) as value. This means that (a) `unique_id` is unique relative to the model (so option 2 above) and (b) that unique_id starts from 1 just as the old code did
* removed model.next_id 
* made the various tests work again. 

Like in #2226, The code here still accepts a unique_id and raises a warning. If you remove that, the number of tests that fails explodes again. There are many tests that use custom unique ids (so not even model.next_id. 

Not sure where this leaves #2226 or this one. It seems there is broad agreement on having unique_id assigned automatically in MESA 3, but it is equally clear that this involves quite some more work depending on how breaking this new feature is implemented.

tests/test_datacollector.py
```diff
@@ -61,8 +53,8 @@ def __init__(self):
         self.model_val = 100
 
         self.n = 10
-        for i in range(self.n):
-            self.schedule.add(MockAgent(i, self, val=i))
+        for i in range(1, self.n + 1):
```
EwoutH, line 56: This is indeed the challenge by starting from 1. But I support it.

EwoutH (approved):
Overall looks good, few minor things.

mesa/agent.py
```diff
@@ -32,19 +34,38 @@ class Agent:
...
+        else:
+            warnings.warn(
+                "unique ids are assigned automatically in MESA 3",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            unique_id, model, *args = args
+            # if *args is not empty, what should we do?
+            # raise ValueError
+            unique_id = next(self._ids[model])
```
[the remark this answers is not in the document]
quaquel: Nice!

mesa/model.py
```diff
@@ -189,11 +189,6 @@ def run_model(self) -> None:
     def step(self) -> None:
         """A single step. Fill in here."""
 
-    def next_id(self) -> int:
```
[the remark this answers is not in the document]
quaquel: added a deprecation warning and return 0

mesa/agent.py
```diff
@@ -32,19 +34,38 @@ class Agent:
...
-        self.pos: Position | None = None
+        unique_id (int): A unique identifier for this agent.
+        pos (Position): A reference to the position where this agent is located.
     """
 
-    def __init__(self, unique_id: int, model: Model) -> None:
+    # this is a class level attribute
+    # it is a dictionary, indexed by model instance
+    # so, unique_id is unique relative to a model, and counting starts from 1
+    _ids = defaultdict(functools.partial(itertools.count, 1))
```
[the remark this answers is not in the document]
quaquel: added a notes to the docstring of Agents

EwoutH:
Great work, thanks!

divilian:
I like the idea of `unique_id` being automatically set for an Agent, instead of having to pass it to the constructor explicitly. But the sad thing for me is that all my agents are also nodes in a networkx graph, and nodes in networkx are numbered starting from 0, not 1. So my code that worked before upgrading to Mesa 3.0 now breaks, because before I was setting all the agent IDs from 0, and now they are "helpfully" being automatically set starting from 1.

Any obvious clean way around this? Making my networkx graph "one node larger" and ignoring its node #0 to accommodate Mesa is a hack I'm not prepared to consider at this point. I'm wondering if there's any easy option to putting in a bunch of logic in my code to subtract 1 and add 1 to IDs at various points where I switch back and forth between working with "agent k" vs. "graph node k-1".

quaquel:
The easiest thing would be to add a `@property` to your Agent:

```python
@property
def note_id(self):
    return self.unique_id - 1


```

PR #2261, opened 2024-08-30
CI: Add optional dependency for examples

EwoutH:
The Game of Life fast example (https://github.com/projectmesa/mesa-examples/pull/182) needs SciPy. Instead of adding it to our dev dependencies, I created a new, lighter examples optional dependency category (with SciPy).


EwoutH:
And everything is green again!

✅ We like green.

Issue #2176, opened 2024-07-11
Access and modify model from Solara visualisation.


EwoutH:
Let's say you're running the Solara visualisation in a Jupyter notebook. Currently you can initialize the visualisation, and then it initializes a model instance from a model class.

![image](https://github.com/projectmesa/mesa/assets/15776622/d6d32211-a349-4ca6-91f9-62c2d0eade89)

However, you can currently not:
- [ ] Start Solara with an existing model
- [ ] Directly access the model instance, to get variables or data (like `model.datacollectors.get_model_vars_dataframe()`)
- [ ] Perform model methods, like `model.step()`.

Ideally you want full control of the model, and Solara just updating whatever the current state of the model is. All the buttons in Solara should access methods that are also accessible by Python. 

I think one non-ideal design choice we made is initializing the model in the JupyterViz class itself ([here](https://github.com/projectmesa/mesa/blob/313469604ba1eeddf89e93c9e8cc18dfd11ee235/mesa/visualization/jupyter_viz.py#L127-L135)). If it's initialized outside there, you can just pass the model _instance_ instead of the _class_. 

One way of approaching is being able to directly access `page.model` and do things from there.

There might be some overlap with [Mesa-interactive](https://github.com/Corvince/mesa-interactive) / #1805 here (cc @Corvince).

cc @rht and @maarten

EwoutH:
If we can get something like the NetLogo [Command Center](https://ccl.northwestern.edu/netlogo/docs/interfacetab.html#command-center) in Solara that would also be amazing.

![image](https://github.com/projectmesa/mesa/assets/15776622/53839d21-8fef-4e67-a620-82cd2a3b0477)

Just execute Python commands from a bar in your browser, in the visualisation. Things like:

```Python
model.step()
print(model.agents[0])
```

Corvince:
Well that was one of my design goals for mesa interactive, so yes there are overlaps :)

Two issues are somewhat solved there. You can access the model at least from within the solar frontend, so you can easily write your own components for interactions. And those components only depend on the model - so you can "apply" those components to any model, not just the one created with mesa interactive. 

But running code outside the frontend, e.g. in a different Jupiter cell doesn't yet automatically update the component plots - that's one thing I was looking into last (several months ago)


I am currently on vacation until the end of the month and then starting a new job. So time is very limited, but hopefully I can eventually port some of the ideas of mi back to mesa

maartenbreddels:
[...]

PR #2263, opened 2024-08-30
Solaraviz api

Corvince:
This is the start of a new API for the Solara frontend. See https://github.com/projectmesa/mesa/discussions/2236 for more background information. I will eventually update this intro to accommodate more information, but I wanted to start this draft PR now and not wait longer just because I am too lazy right now to write more sentences.

So far everything worked out nicely. The interesting changes are in the `myviz.py`  file. This is where the new entrypoint can be found. The other components ModelController and UserParams only received very minor changes and were just moved to other files to not interefere with the current SolaraViz (during development).

MyViz also includes a minimal example, if someone wants to try this out just run `solara run myviz.py`

image

### DIFF
--- a/mesa/visualization/utils.py
+++ b/mesa/visualization/utils.py
@@ -1,0 +1,7 @@
+import solara
+
+update_counter = solara.reactive(0)
+
+
+def force_update():
+    update_counter.value += 1
--- a/mesa/__init__.py
+++ b/mesa/__init__.py
@@ -8,7 +8,6 @@ import datetime
 
 import mesa.space as space
 import mesa.time as time
-import mesa.visualization as visualization
 from mesa.agent import Agent
 from mesa.batchrunner import batch_run
 from mesa.datacollection import DataCollector
@@ -22,7 +21,6 @@ __all__ = [
     "DataCollector",
     "batch_run",
     "experimental",
-    "visualization",
 ]
 
 __title__ = "mesa"
--- a/mesa/__init__.py
+++ b/mesa/__init__.py
@@ -8,6 +8,7 @@ import datetime
 
 import mesa.space as space
 import mesa.time as time
+import mesa.visualization as visualization
 from mesa.agent import Agent
 from mesa.batchrunner import batch_run
 from mesa.datacollection import DataCollector
@@ -21,6 +22,7 @@ __all__ = [
     "DataCollector",
     "batch_run",
     "experimental",
+    "visualization",
 ]
 
 __title__ = "mesa"
--- a/mesa/visualization/UserParam.py
+++ b/mesa/visualization/UserParam.py
@@ -1,0 +1,56 @@
+class UserParam:
+    _ERROR_MESSAGE = "Missing or malformed inputs for '{}' Option '{}'"
+
+    def maybe_raise_error(self, param_type, valid):
+        if valid:
+            return
+        msg = self._ERROR_MESSAGE.format(param_type, self.label)
+        raise ValueError(msg)
+
+
+class Slider(UserParam):
+    """
+    A number-based slider input with settable increment.
+
+    Example:
+
+    slider_option = Slider("My Slider", value=123, min=10, max=200, step=0.1)
+
+    Args:
+        label: The displayed label in the UI
+        value: The initial value of the slider
+        min: The minimum possible value of the slider
+        max: The maximum possible value of the slider
+        step: The step between min and max for a range of possible values
+        dtype: either int or float
+    """
+
+    def __init__(
+        self,
+        label="",
+        value=None,
+        min=None,
+        max=None,
+        step=1,
+        dtype=None,
+    ):
+        self.label = label
+        self.value = value
+        self.min = min
+        self.max = max
+        self.step = step
+
+        # Validate option type to make sure values are supplied properly
+        valid = not (self.value is None or self.min is None or self.max is None)
+        self.maybe_raise_error("slider", valid)
+
+        if dtype is None:
+            self.is_float_slider = self._check_values_are_float(value, min, max, step)
+        else:
+            self.is_float_slider = dtype is float
+
+    def _check_values_are_float(self, value, min, max, step):
+        return any(isinstance(n, float) for n in (value, min, max, step))
+
+    def get(self, attr):
+        return getattr(self, attr)
--- a/mesa/visualization/__init__.py
+++ b/mesa/visualization/__init__.py
@@ -1,6 +1,14 @@
-
-from .solara_viz import JupyterViz, Slider, SolaraViz, make_text
 from .components.altair import make_space_altair
-from .components.matplotlib import make_space_matplotlib, make_plot_measure
+from .components.matplotlib import make_plot_measure, make_space_matplotlib
+from .solara_viz import JupyterViz, SolaraViz, make_text
+from .UserParam import Slider
 
-__all__ = ["JupyterViz", "SolaraViz", "make_text", "Slider", "make_space_altair", "make_space_matplotlib", "make_plot_measure"]
+__all__ = [
+    "JupyterViz",
+    "SolaraViz",
+    "make_text",
+    "Slider",
+    "make_space_altair",
+    "make_space_matplotlib",
+    "make_plot_measure",
+]
--- a/mesa/visualization/__init__.py
+++ b/mesa/visualization/__init__.py
@@ -1,0 +1,6 @@
+
+from .solara_viz import JupyterViz, Slider, SolaraViz, make_text
+from .components.altair import make_space_altair
+from .components.matplotlib import make_space_matplotlib, make_plot_measure
+
+__all__ = ["JupyterViz", "SolaraViz", "make_text", "Slider", "make_space_altair", "make_space_matplotlib", "make_plot_measure"]
--- a/mesa/visualization/components/altair.py
+++ b/mesa/visualization/components/altair.py
@@ -1,0 +1,86 @@
+import contextlib
+
+import solara
+
+with contextlib.suppress(ImportError):
+    import altair as alt
+
+from mesa.visualization.utils import update_counter
+
+
+def make_space_altair(agent_portrayal=None):
+    if agent_portrayal is None:
+
+        def agent_portrayal(a):
+            return {"id": a.unique_id}
+
+    def MakeSpaceAltair(model):
+        return SpaceAltair(model, agent_portrayal)
+
+    return MakeSpaceAltair
+
+
+@solara.component
+def SpaceAltair(model, agent_portrayal, dependencies: list[any] | None = None):
+    update_counter.get()
+    space = getattr(model, "grid", None)
+    if space is None:
+        # Sometimes the space is defined as model.space instead of model.grid
+        space = model.space
+    chart = _draw_grid(space, agent_portrayal)
+    solara.FigureAltair(chart)
+
+
+def _draw_grid(space, agent_portrayal):
+    def portray(g):
+        all_agent_data = []
+        for content, (x, y) in g.coord_iter():
+            if not content:
+                continue
+            if not hasattr(content, "__iter__"):
+                # Is a single grid
+                content = [content]  # noqa: PLW2901
+            for agent in content:
+                # use all data from agent portrayal, and add x,y coordinates
+                agent_data = agent_portrayal(agent)
+                agent_data["x"] = x
+                agent_data["y"] = y
+                all_agent_data.append(agent_data)
+        return all_agent_data
+
+    all_agent_data = portray(space)
+    invalid_tooltips = ["color", "size", "x", "y"]
+
+    encoding_dict = {
+        # no x-axis label
+        "x": alt.X("x", axis=None, type="ordinal"),
+        # no y-axis label
+        "y": alt.Y("y", axis=None, type="ordinal"),
+        # "tooltip": [
+        #     alt.Tooltip(key, type=alt.utils.infer_vegalite_type([value]))
+        #     for key, value in all_agent_data[0].items()
+        #     if key not in invalid_tooltips
+        # ],
+    }
+    has_color = "color" in all_agent_data[0]
+    if has_color:
+        encoding_dict["color"] = alt.Color("color", type="nominal")
+    has_size = "size" in all_agent_data[0]
+    if has_size:
+        encoding_dict["size"] = alt.Size("size", type="quantitative")
+
+    chart = (
+        alt.Chart(
+            alt.Data(values=all_agent_data), encoding=alt.Encoding(**encoding_dict)
+        )
+        .mark_point(filled=True)
+        .properties(width=280, height=280)
+        # .configure_view(strokeOpacity=0)  # hide grid/chart lines
+    )
+    # This is the default value for the marker size, which auto-scales
+    # according to the grid area.
+    if not has_size:
+        length = min(space.width, space.height)
+        chart = chart.mark_point(size=30000 / length**2, filled=True)
+
+    return chart
--- a/mesa/visualization/components/altair.py
+++ b/mesa/visualization/components/altair.py
@@ -56,11 +56,11 @@ def _draw_grid(space, agent_portrayal):
         "x": alt.X("x", axis=None, type="ordinal"),
         # no y-axis label
         "y": alt.Y("y", axis=None, type="ordinal"),
-        # "tooltip": [
-        #     alt.Tooltip(key, type=alt.utils.infer_vegalite_type([value]))
-        #     for key, value in all_agent_data[0].items()
-        #     if key not in invalid_tooltips
-        # ],
+        "tooltip": [
+            alt.Tooltip(key, type=alt.utils.infer_vegalite_type([value]))
+            for key, value in all_agent_data[0].items()
+            if key not in invalid_tooltips
+        ],
     }
     has_color = "color" in all_agent_data[0]
     if has_color:
--- a/mesa/visualization/components/matplotlib.py
+++ b/mesa/visualization/components/matplotlib.py
@@ -1,0 +1,246 @@
+from collections import defaultdict
+
+import networkx as nx
+import solara
+from matplotlib.figure import Figure
+from matplotlib.ticker import MaxNLocator
+
+import mesa
+from mesa.experimental.cell_space import VoronoiGrid
+from mesa.visualization.utils import update_counter
+
+
+def make_space_matplotlib(agent_portrayal=None):
+    if agent_portrayal is None:
+
+        def agent_portrayal(a):
+            return {"id": a.unique_id}
+
+    def MakeSpaceMatplotlib(model):
+        return SpaceMatplotlib(model, agent_portrayal)
+
+    return MakeSpaceMatplotlib
+
+
+@solara.component
+def SpaceMatplotlib(model, agent_portrayal, dependencies: list[any] | None = None):
+    update_counter.get()
+    space_fig = Figure()
+    space_ax = space_fig.subplots()
+    space = getattr(model, "grid", None)
+    if space is None:
+        # Sometimes the space is defined as model.space instead of model.grid
+        space = model.space
+    if isinstance(space, mesa.space.NetworkGrid):
+        _draw_network_grid(space, space_ax, agent_portrayal)
+    elif isinstance(space, mesa.space.ContinuousSpace):
+        _draw_continuous_space(space, space_ax, agent_portrayal)
+    elif isinstance(space, VoronoiGrid):
+        _draw_voronoi(space, space_ax, agent_portrayal)
+    else:
+        _draw_grid(space, space_ax, agent_portrayal)
+    solara.FigureMatplotlib(space_fig, format="png", dependencies=dependencies)
+
+
+# matplotlib scatter does not allow for multiple shapes in one call
+def _split_and_scatter(portray_data, space_ax):
+    grouped_data = defaultdict(lambda: {"x": [], "y": [], "s": [], "c": []})
+
+    # Extract data from the dictionary
+    x = portray_data["x"]
+    y = portray_data["y"]
+    s = portray_data["s"]
+    c = portray_data["c"]
+    m = portray_data["m"]
+
+    if not (len(x) == len(y) == len(s) == len(c) == len(m)):
+        raise ValueError(
+            "Length mismatch in portrayal data lists: "
+            f"x: {len(x)}, y: {len(y)}, size: {len(s)}, "
+            f"color: {len(c)}, marker: {len(m)}"
+        )
+
+    # Group the data by marker
+    for i in range(len(x)):
+        marker = m[i]
+        grouped_data[marker]["x"].append(x[i])
+        grouped_data[marker]["y"].append(y[i])
+        grouped_data[marker]["s"].append(s[i])
+        grouped_data[marker]["c"].append(c[i])
+
+    # Plot each group with the same marker
+    for marker, data in grouped_data.items():
+        space_ax.scatter(data["x"], data["y"], s=data["s"], c=data["c"], marker=marker)
+
+
+def _draw_grid(space, space_ax, agent_portrayal):
+    def portray(g):
+        x = []
+        y = []
+        s = []  # size
+        c = []  # color
+        m = []  # shape
+        for i in range(g.width):
+            for j in range(g.height):
+                content = g._grid[i][j]
+                if not content:
+                    continue
+                if not hasattr(content, "__iter__"):
+                    # Is a single grid
+                    content = [content]
+                for agent in content:
+                    data = agent_portrayal(agent)
+                    x.append(i)
+                    y.append(j)
+
+                    # This is the default value for the marker size, which auto-scales
+                    # according to the grid area.
+                    default_size = (180 / max(g.width, g.height)) ** 2
+                    # establishing a default prevents misalignment if some agents are not given size, color, etc.
+                    size = data.get("size", default_size)
+                    s.append(size)
+                    color = data.get("color", "tab:blue")
+                    c.append(color)
+                    mark = data.get("shape", "o")
+                    m.append(mark)
+        out = {"x": x, "y": y, "s": s, "c": c, "m": m}
+        return out
+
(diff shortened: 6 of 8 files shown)


EwoutH:
Exciting!

I tried both
```bash
mesa>solara run mesa/visualization/myviz.py
```
and 
```bash
mesa\mesa\visualization>solara run myviz.py
```
But for some reason I can't get the imports working. Will take a better look tomorrow.

rht:
I tried it and it worked. Will add inline comments.

mesa/visualization/myviz.py
```diff
@@ -0,0 +1,97 @@
...
+            self.grid.place_agent(
+                a, (self.random.randrange(10), self.random.randrange(10))
+            )
+
+    def step(self):
+        self.agents.do("step")
+
+
+model = MyModel(10)
+Page = MyViz(model, [ShowSteps, ShowNumAgents, make_space_altair()])
```
rht, line 97: New API looks more modular. This means user can add any existing Solara components from Solara library, including Plotly if they want to.

I think it would be convenient to have a string "default" that automatically expands into `[ShowSteps, make_space_altair()]`. Not sure about num agents being shown by default, but this is a minor point.

EwoutH:
@rht from where did you run it? I would also like to try it.

mesa/visualization/utils.py
```diff
@@ -0,0 +1,7 @@
+import solara
+
+_update_counter = solara.reactive(0)
```
rht, line 3: Is this global variable a temporary artifact? Because we want to support showing multiple SolaraViz in one go: https://simrisk.pulcloud.io/.

rht:
I ran it In the visualization directory.

EwoutH:
Maybe something with working directories? Hate this kind of stuff, and imports in general. So many inexplainable issues.

image

EwoutH:
It was a namespace conflict between the installed Mesa package and the GitHub repo. Removing mesa and adding

```Python
import sys
import os

sys.path.insert(0, os.path.abspath('C:/Users/Ewout/Documents/GitHub/mesa'))
```
solved it. Let's take a look!

EwoutH:
Looks clean, feels fast and responsive!

What's the difference between these two buttons? When running they both seem to pauze the simulation in the same way.

image

mesa/visualization/ModelCreator.py
```diff
@@ -0,0 +1,92 @@
+import solara
+
+from mesa.visualization.solara_viz import split_model_params
+from mesa.visualization.UserParam import Slider
+
+
+@solara.component
+def ModelCreator(model, model_cls, model_params):
```
EwoutH, line 8: So you can input a model instance, model class and params? I get either the first or the second two are required?

If you input an instance, will it be updated in place, so you can access it from your interpreter later?

mesa/visualization/ModelCreator.py
```diff
@@ -0,0 +1,92 @@
...
+                min=options.min,
+                max=options.max,
+                step=options.step,
+            )
+            continue
+
+        # label for the input is "label" from options or name
+        label = options.get("label", name)
+        input_type = options.get("type")
+        if input_type == "SliderInt":
```
EwoutH, line 60: What does this structure add to the user? Isn't it easier to just let them just the Solara components directly?

rht:
The 2 buttons having the same functionality is out of scope for the PR. It is due to ipywidget displaying both and there is no way to turn either off.

EwoutH (commented):
I really love that you can just input an model instance, and if I'm correct, acces that model instance from the interpreter.

This would solve a huge part of https://github.com/projectmesa/mesa/issues/2176, right?

mesa/visualization/ModelCreator.py
rht: This is already present in existing SolaraViz, and so is out of scope. The point was so that user can declaratively specify the param in a JSON, and so is serializable.

rht: If the scope of the PR is small, the API change can be merged sooner.

EwoutH: Thanks for the context, I’m not 100% up to speed with the current visualization.

Corvince:
I have updated the code so it now updates solara_viz.py instead of creating a new myviz. Also moved back the modified ModelController.

All in all the diff should be much more reflective of the actual code changes. 

Things that are still missing (please add if you notice something):

- [ ] Layout stuff (draggable grid layout)
- [ ] reactive seed
- [ ] Easy way to plot simple measures just by their attribute name

mesa/visualization/myviz.py
Corvince: Yes I think thats a good idea. Num agents are currently just shown for testing purpose - I don't think it should be default.

mesa/visualization/utils.py
Corvince: I know that has been an issue in older versionis of solara, but solara might handle this gracefully. The docs now include this sentence

> Reactive objects in Solara are also context-aware, meaning that they can maintain separate values for each browser tab or user session. This enables each user to have their own independent state, allowing them to interact with the web application without affecting the state of other users.
https://solara.dev/documentation/api/utilities/reactive

But I haven't tested his yet.

Corvince:
> I really love that you can just input an model instance, and if I'm correct, acces that model instance from the interpreter.
> 
> This would solve a huge part of https://github.com/projectmesa/mesa/issues/2176, right? And would [this](https://github.com/projectmesa/mesa/issues/2176#issuecomment-2282808460) be possible?

Yes, exactly! Although for it to fully work, even if you reset the model or change parameters, you should pass in the model as a reactive model, i.e. call `solara.reactive(model)` on it. Inside the SolaraViz it also attaches a call to `force_update()` after each step, which triggers a viz update.

mesa/visualization/ModelCreator.py
Corvince: Actually it expects a reactive version of a model, so that the model from the outside still receives the recreated versions. 

I now removed the model_cls, as it can be resolved from the model instance. 

So now it receives an actual model (as a reactive variable) and user params. So I think ModelCreator isn't a good name

EwoutH:
If you need any input let me know!

mesa/visualization/utils.py
rht: What about the case when there are multiple `page` objects in a single Jupyter notebook? As can be seen in the [visualization tutorial](https://mesa.readthedocs.io/en/latest/tutorials/visualization_tutorial.html). They seem to share the same context.

Corvince: That shouldn't be a big problem luckily. It only forces a rerender, no data change. So at worst you have a few unnecessary rerenders in your notebook. Let's wait and see if this causes serious performance issues and if yes reconsider. But for now it should be good

Corvince:
Now that you say it, what's your opinion on the layout? I thought at first to implement how it's currently done. But maybe we can be more unopinionated. In my implementation all components are just displayed with no surrounding layout. But the components themselves can of course include any layout. We could keep it that way and provide some layout components. So users could just pass `solaraviz(model, components=GridDraggable(componentA, componentB)` or something like that.

Also, regarding naming. Is Components a good enough name?

EwoutH:
I think the current layout is quite good. I like controls on the left, visualisation and output in the center/right.
- Maybe it would be nice if you could specify is you want things below or right of previous components, but maybe keeping it adaptive is more flexible.
- It might be nice to be able to specify an aspect ratio per component. Some graphs you might want wide, some tall etc.
  - For the `space_drawer` this might be done automatically based on the grid width and height.

> Also, regarding naming. Is Components a good enough name?

Many names could work here, but Components is quite good.

rht:
I actually find the naming to be quite restrictive in order to encompass other Solara components, because it needs to be general, e.g. if you want to use the `DataFrame`/`PivotTable` component in https://solara.dev/documentation/components. "Measures" doesn't cover this. Meanwhile, "components" is what is already used by Solara.

mesa/visualization/utils.py
rht: I also notice that all the components take `model` as an argument, and so, would it be possible to smuggle in the update counter as an attribute of the model?

Corvince: Interesting idea. Not sure this will work. I am rather sceptical, but will definitely try it out.

quaquel:
This is just a clarifying question from my end. In the various conversations on data collection, we converged on building this on top of some pub-sub or signal-based API. How might that choice affect this effort, or vice versa, what would be convenient to use inside MESA such that both data collection and model visualization can utilize it?

In #2176, @maartenbreddels suggested using either Solara reactive variables or psygnal. I have looked at the latter, and it is very elegant and does a lot of what we want for data collection. Importantly, it also allows for subscribing to updates in collections. However, it would add an additional dependency to MESA. I also started exploring Solara's reactive variables. It appears this covers a subset of what we would need for data collection, but I am not sure until I actually play around with it more. 

In my vision, both the visualization side and the data collection side of MESA use the same mechanism for being informed/getting state information from a MESA model.

Corvince:
> This is just a clarifying question from my end. In the various conversations on data collection, we converged on building this on top of some pub-sub or signal-based API. How might that choice affect this effort, or vice versa, what would be convenient to use inside MESA such that both data collection and model visualization can utilize it?
> 
> In #2176, @maartenbreddels suggested using either Solara reactive variables or psygnal. I have looked at the latter, and it is very elegant and does a lot of what we want for data collection. Importantly, it also allows for subscribing to updates in collections. However, it would add an additional dependency to MESA. I also started exploring Solara's reactive variables. It appears this covers a subset of what we would need for data collection, but I am not sure until I actually play around with it more.
> 
> In my vision, both the visualization side and the data collection side of MESA use the same mechanism for being informed/getting state information from a MESA model.

The top level view on how it currently works (in this PR) is that each component registers a "update_counter" as a render dependency. Whenever this counter increases the component rerenders. Currently the counter is increased after each mode..step.

Of course it would be much more elegant to auto-update whenever an attribute changes, for example not through a call to step, but maybe through some form of user interactions. If the variable would be a reactive variable this would already happen automatically. 

So from a visualization point of view the use of solaras reactive values would be very charming. So I would appreciate it if you further investigate solaras reactive suitability. 

But overall this PR is mostly neutral to how we handle this - it mostly deals with the surroundings and establishing a cleaner API. The use of signals/reactive values/pub-sub will mostly play some role for individual components (e.g. a plot of a specifc variable)

Corvince:
I just marked this as ready for review! Even though the tests are still failing, this should probably also be fixed as part of this PR? 

Otherwise I think this is good to review as a first PR just for the API changes.

Next todos are:

- [ ] Update example models
- [ ] Write migration guide from current API to this PR. You should be able to have the same viz as before with small adjustemnts, but those adjustments need to be written out.
- [ ] Bring back GridDraggable layout. This was used before, but I found the code a bit confusing so I didn't manage to adapt just yet. Per default the layout now is unopinionated, but you can bring back the old style with `SolaraViz(model, components=[GridDraggable(...components)]`. But the GridDraggable component needs to be written in a further PR
- [ ] Decide which subcomponents will be part of the public API for Mesa v3. 

Not so immediate, but also a nice todo: Rewrite ModelController. There is already code for a much simpler threaded version and from a first test this finally seems to work. Maybe it doesn't even need to be threaded and just works. In any case it can be vastly simplified (and no more unnecessary pause and stop button)

But first lets get this PR through!

EwoutH:
@Corvince looks awesome! Any specific things you would like me to test or review?

Corvince:
Yes! You can try to use it 🙈 

Heres a quick migration guide. Where previous you could have something like this for a minimal frontend
```python
SolaraViz(model_cls, model_params, agent_portrayal=agent_portrayal)
```

you would now do

```python
SolaraViz(model, components=[make_space_altair(agent_portrayal)])
```

you can test if that works for `make_space_altair` and `make_space_matplotlib` and then additionally add `model_params` to SolaraViz. And then test if everything works (start, stop, reset, reseed, change parameters).

That would be awesome!

EwoutH:
Awesome, I'm going to 
```
pip install -U -e git+https://github.com/projectmesa/mesa@solaraviz-api#egg=mesa
```
and test some examples.

I just talked to Jan, he is going to try to dust of hist pub-sub efforts to see if that will integrate nicely with this.

EwoutH:
Playing a bit around! Got `boltzmann_wealth_model_experimental` working:

image

> * Easy way to plot simple measures just by their attribute name

Yeah this would be really useful. We could think of another/more robust API to parse measures other than just a string name (@quaquel?), but I think this is important to get right.

I like that's way more flexible: You don't need to input `model_params` for values you don't need as sliders and such. You can just use the default values or override them.

Another this that's interesting is that the slider overrides default/other values. I think that's good to document.

image

It does show that something like 
- https://github.com/projectmesa/mesa/issues/2268

can be useful.

EwoutH:
I also tried [`virus_on_network`](https://github.com/projectmesa/mesa-examples/tree/main/examples/virus_on_network). However, it has a custom [`make_plot`](https://github.com/projectmesa/mesa-examples/blob/330b4d9c8e44127b3c4b89603972a46dd5ec39b9/examples/virus_on_network/app.py#L47C1-L60C33) function, that seems difficult to adapt, and I didn't get it to work.

```Python
def make_plot(model):
    # This is for the case when we want to plot multiple measures in 1 figure.
    fig = Figure()
    ax = fig.subplots()
    measures = ["Infected", "Susceptible", "Resistant"]
    colors = ["tab:red", "tab:green", "tab:gray"]
    for i, m in enumerate(measures):
        color = colors[i]
        df = model.datacollector.get_model_vars_dataframe()
    fig.legend()
    # Set integer x axis
    ax.xaxis.set_major_locator(MaxNLocator(integer=True))
    solara.FigureMatplotlib(fig)
```

```Python
page = SolaraViz(
    VirusOnNetwork,
    model_params,
    measures=[
        make_plot,
        make_text(get_resistant_susceptible_ratio),
    ],
    name="Virus Model",
)
page  # noqa
```

Corvince:
Awesome, thank you so much for testing this out!

> > * Easy way to plot simple measures just by their attribute name
> 
> Yeah this would be really useful. We could think of another/more robust API to parse measures other than just a string name (@quaquel?), but I think this is important to get right.

Yes, this needs some rework. Currently it takes its data from the data collector, but I think it would make more sense to get the values directly. But I already implemented the current approach with `make_plot_measure("foo")`


> I also tried [`virus_on_network`](https://github.com/projectmesa/mesa-examples/tree/main/examples/virus_on_network). However, it has a custom [`make_plot`](https://github.com/projectmesa/mesa-examples/blob/330b4d9c8e44127b3c4b89603972a46dd5ec39b9/examples/virus_on_network/app.py#L47C1-L60C33) function, that seems difficult to adapt, and I didn't get it to work.

It should be as easy as turning it into a solara component and adding the update_counter:

```python
@solara.component  # first change
def make_plot(model):
    update_counter.get() # second change
    # This is for the case when we want to plot multiple measures in 1 figure.
    fig = Figure()
    ax = fig.subplots()
    measures = ["Infected", "Susceptible", "Resistant"]
    colors = ["tab:red", "tab:green", "tab:gray"]
    for i, m in enumerate(measures):
        color = colors[i]
        df = model.datacollector.get_model_vars_dataframe()
    fig.legend()
    # Set integer x axis
    ax.xaxis.set_major_locator(MaxNLocator(integer=True))
    solara.FigureMatplotlib(fig)
```

But yes, documenting how to add author custom components should als be pretty high on the follow up todo list

EwoutH:
The only question I now have, is do we want to directly replace the current viz or do we want to have some deprecation scheme and have them parallel for a bit?

maartenbreddels:
> this could be done using Solara's `Reactive`, but this is closely wrapped up with all kinds of GUI stuff that we don't need.

Why do you think that is the case? They are, I think, quite separated from the UI part (except being in the same package).

Something to also consider is 'scope'. In solara server (if you plan to use that), you can have multiple browser pages (often seen as 'users') connected to the same runtime. Reactive variables understand that, similar to threading.local for thread local storage, Solara reactive variables use kernel-local storage.

quaquel:
> Why do you think that is the case? They are, I think, quite separated from the UI part (except being in the same package).

Let me clarify. I have been trying to figure out which parts from Solara I can use to reimplement my existing pubsub idea and envisioned API. The documentation for Solara focusses on "GUI stuff". So, I have been diving into the source code. This has taken me to parts of the code base that are not documented (e.g.,`Reactive`), so I have been trying to figure out what is going on and how to utilize it for MESA purposes.

 In short, my impression is based perhaps more on the existing Solara documentation and my struggle to parse from the entire code base the parts I would need. As you said yourself in #2176, ideally the reactive part should be its own library.

Corvince:
> The only question I now have, is do we want to directly replace the current viz or do we want to have some deprecation scheme and have them parallel for a bit?

Thats a good question and one I have been pondering for a while. Technically the solaraviz has always been experimental and so we don't break any contract with the changes. However, there hasn't been an alternative built-in frontend for a while now and we have been pointing towards the new viz for some time. It will be a bit annoying for everyone to rebuild there modesl yet again. But then again we have several breaking changes with mesa 3.0, so people will probably have to rebuild their models anyway. 

One difficulty with deprecation is how to handle it gracefully. I think the code for that will be a bit ugly and then is it really necessary to deprecate experimental features? The most elegant deprecation would of course be to find a new name for "SolaraViz". I think its not perfect at all, because its neither specific to mesa, nor generic enough. Sure, currently it only supports solara components, but maybe this could change in the future (e.g. you can already use vue components, but you need to wrap them with `@solara.vue_component`). 

So maybe there is a better name to be discovered? Then we could leave the old code intact (deprecated) and write nice migration guides to the new viz.

EwoutH:
Since we haven't clear policy, versioning systems and integration testing in our examples, I would propose placing this in the `.experimental` space for now. That way we can continue to develop this at a faster pace without being afraid that we have to update everything continuously.

Furthermore, I really want to be able to test the visualisation in CI. I think this will really help with that:

> I want to directly trigger run the Solara visualisation from the Python script. So doing something like `page.run()` or whatever.

Would something like this be possible?

EwoutH:
> One difficulty with deprecation is how to handle it gracefully.

1. Add new stuff in new location
2. Develop it further
3. Declare it stable
4. Start pointing to it, write docs, etc.
5. Start deprecating the old one (throw a warning)
6. Wait a few versions
7. Throw error for the old one
8. Wait a few versions
9. Remove the old one
10. If preferred, move the new one. Point from the initial new location to the new initial location not to break things.

Best for it would be to have a new name to make it clearly distinct.

Corvince:
> > I want to directly trigger run the Solara visualisation from the Python script. So doing something like `page.run()` or whatever.

I am not sure I understand this correctly. Whats the workflow here? Normally you would execute `solara run file.py`. Then you can edit the file and solara will reload and reflect all changes. 
Or you can use the solaraviz in a notebook. Then (with this PR) if you pass in a reactive model you can modify the model outside the solaraviz display, and it will update anytime you call `model.value.force_update()`. This is automatically done for `model.step`

Or do you mean in an interactive python session? Not sure if this is possible, but since solara automatically reloads, I think its easier to just start it on a file.

EwoutH:
I just want to press run here (which does `python app.py`) and have the visualisation start.

image

Corvince:
I think you just need some custom settings in your editor. Have a look at

https://solara.dev/documentation/advanced/howto/debugging

EwoutH:
I tried a lot of things but unfortunately none worked.

The point it, I just want to run a Python script when I'm normally do, and then, when the SolaraViz instance is called (`page()`, `page.run()`, whatever), Solara starts. Then, I do my stuff in Solara, at some point I'm done, press something like "Close", and now I'm back in the Python script and if there's any code left, it continues running.

(we can have this discussion in a separate thread if you want, since it's not directly linked to this PR)

Corvince:
> Since we haven't clear policy, versioning systems and integration testing in our examples, I would propose placing this in the `.experimental` space for now. That way we can continue to develop this at a faster pace without being afraid that we have to update everything continuously.

Disagree here. We currently don't have a frontend currently. The latest stable release (2.3) still has the "old" SolaraViz under experimantal. The next release (3.0) will have the "old" solaraviz marked as stable. But the whole purpose of this PR was to have a good, stable API for 3.0. Dont like the idea of moving this to experimental and then soon deprecating the "old" SolaraViz, when its not out of experimental yet.

EwoutH:
Yeah that's a good point. I'm inclined to agree.

But then the docs need to be on-par before we get into Mesa 3.0 beta territory. Do we have the capacity to do that (@projectmesa/maintainers)?

rht:
> But then the docs, tutorial and examples need to be on-par before we get into Mesa 3.0 beta territory. Do we have the capacity to do that (@projectmesa/maintainers)?

It may seem to be a lot of work, but given that in Mesa 2.x it is still an experimental API, users of the JupyterViz (now SolaraViz) were already expecting API changes. It's less work overall to include the API changes altogether now, than to delay it to Mesa 4.0.

EwoutH (approved):
I'm good with it. Just saying it has to be done.

Corvince:
Exactly. I think it may make sense to actually move the JupyterViz/SolaraViz back into experimental. This way 2.x users who already use the experimental SolaraViz module don't *need* to change anything. They can still `import SolaraViz from mesa.experimental`  But at the same time we have the new, stable API `import SolaraViz from mesa.visualization`.  How does that sound? This way we only need changes from people on the 3.x pre release channel, but changes are expected there. 

Still we should consider a new name, but this can come later (before 3.0 stable of course. But if we have a new name we can alias it to SolaraViz, like we currently do for JupyterViz)

Corvince:
I rebased this PR. Hopefully tonight or tomorrow at the latest I will update the tests, so ci will be happy again

EwoutH:
Leftover from #2278, but could you fix this import?

https://github.com/projectmesa/mesa/blob/cfa26359b3aecef12636031e08230d38768ae4b9/tests/test_solara_viz.py#L8

(I already approved)

Corvince:
Yes, see my last comment

wang-boyu:
> A new name for the new viz would help a lot. Maybe just `MesaViz`?

How about `ModelViz` or `ModelView`?

Corvince:
I would like to extend the functionality to allow for more model interactions, adding plots interactively and having an "agent observer" for single agents. So not just a pure visualization. After a chat with chatgpt my best idea so far might be "ModelExplorer"

EwoutH:
`ModelExplorer` sounds good. Also fresh, and new.

I would everything that's strictly in Solara (like Solara-components) to keep a Solara-like name. Then if we add non-Solara stuff, we can give them non-Solara names.

wang-boyu:
Hi @tpike3, we probably need to rethink the jupyter visualization for mesa-geo, considering the new API design from this PR.

Corvince:
Can you point me to that? I can take a look tomorrow

wang-boyu:
Thanks! The current implementation is here: https://github.com/projectmesa/mesa-geo/tree/main/mesa_geo/visualization, from PR https://github.com/projectmesa/mesa-geo/pull/212.

There was some comments from @rht that needs to be addressed: https://github.com/projectmesa/mesa-geo/pull/199#issuecomment-2041764285 but we decided to merge and have something working first.

As I mentioned above in https://github.com/projectmesa/mesa/pull/2263#discussion_r1743872440, we probably don't need to copy over anything. Instead, we could try to refactor existing implementations into some solara components for geospace.

Let me know if you need anything else : ) also cc @tpike3

Corvince:
Thanks for the extensive reply! Sorry, I didn't catch your previous comment, but you are spot on! 

It will just require a function `make_geospace_leaflet()` that returns what I think is the "map" component currently. So something like this

```python
from mesa.visualization.utils import update_counter

def make_geospace_leaflet(agent_portrayal, view, zoom, tiles,scroll_wheel_zoom):

    map_drawer = leaflet_viz.MapModule(
        portrayal_method=agent_portrayal,
        view=view,
        zoom=zoom,
        tiles=tiles,
        scroll_wheel_zoom=scroll_wheel_zoom,
    )

    center_default= ...


    @solara.component
    def map(model):
        update_counter.get() # <- Important line to notify the frontend whenever model updates (e..g step)
        # render map in browser
        zoom_map = solara.reactive(zoom)
        center = solara.reactive(center_default)

        base_map = map_drawer.tiles
        layers = map_drawer.render(model)

        ipyleaflet.Map.element(
            zoom=zoom_map.value,
            center=center.value,
            scroll_wheel_zoom=scroll_wheel_zoom,
            layers=[
                ipyleaflet.TileLayer.element(url=base_map["url"]),
                ipyleaflet.GeoJSON.element(data=layers["agents"][0]),
                *layers["agents"][1],
            ],
        )

    return map
```

Then in a notebook you should already be able to do 
```python
GeoSpaceLeaflet = make_geospace_leaflet(...)

GeoSpaceLeaflet(model)
```

And you should see your visualization! And then you can of course pass it to SolaraViz 

Would you like to give it a shot? Unfortunately I don't have the mesa_geo stack currently installed and last time it was a bit of a hassle to get the geo tools to work under windows. Just ask if you need any help

wang-boyu:
Thanks! This is precisely what I thought. I will find some time to try it out!

EwoutH:
I would prefer to have a new name before a potential (alpha) release.

EwoutH:
> I would prefer to have a new name before a potential (alpha) release.

> `ModelExplorer` sounds good. Also fresh, and new.

What do we think of this?

<hr>

On another note, te [Visualization tutorial](https://mesa.readthedocs.io/en/latest/tutorials/visualization_tutorial.html), since it’s still based on the previous version of SolaraViz. I see two options:
1. Use the previous SolaraViz version for now, which we moved to experimental.
2. Update to the new API.

Corvince:
> On another note, te [Visualization tutorial](https://mesa.readthedocs.io/en/latest/tutorials/visualization_tutorial.html), since it’s still based on the previous version of SolaraViz. I see two options:
> 
> 1. Use the previous SolaraViz version for now, which we moved to experimental.
> 2. Update to the new API.

I can quickly do a minimal update to the new API and then do a bit of an overhaul later on, if thats fine?

EwoutH:
Sounds perfect!

jackiekazil:
I am late, but I like ModelExplorer

PR #2265, opened 2024-09-01
Reduce core dependencies, split in optional dependencies

### LINKED ISSUE #626: Remove all library requirements
Mesa currently ships as a "batteries included" package with requirements for all possible functionality provided. While this is nice for new users, concerns have been raised that mesa is unsuitable for integration into larger projects due to its (possibly unused) dependencies. Upon further inspection all currently required packages are optional.

 - `click`, `cookiecutter` only used for `mesa` command
 - `jupyter` not used at all
 - `networkx` only implicitly used by NetworkGrid
 - `numpy` only used for ContinuousSpace
 - `pandas` only used for DataFrame output of the DataCollector
 - `tornado` only used for Visualization
 - `tqdm` only used for batch runner

Those requirements could be put under some extra_requires in the `setup.py` file and then users could still install mesa with `pip install mesa[all]` to have everything included. 

Steps for this issue would include:
 - [ ] Update setup.py (decide on name for "all")
 - [ ] Lazy load all dependencies with helpful error messages if it fails
 - [ ] Update documentation 

But before working on this issue we should decide whether this is a good idea or not

Credits to @rht for starting this discussion in #614

EwoutH:
This PR introduces a more flexible dependency management system for Mesa, allowing users to choose which additional components they need while maintaining core functionality.

### Motivation
This change allows for more tailored installations, reducing unnecessary dependencies and potential conflicts, while ensuring core functionality remains intact.

### Key Changes
1. Updated `pyproject.toml` to define new optional dependency groups:
   - `[network]`: Network-related dependencies
   - `[viz]`: Visualization dependencies
   - `[rec]`: Recommended dependencies (includes network and viz)
   - `[all]`: All dependencies, including developer dependencies
   - Additional groups for development, examples, and documentation
2. Updated `README.md` with new installation instructions for optional dependencies.
3. Core dependencies now only include essential packages (numpy, pandas, tqdm).

### Installation Options
Users can now customize their Mesa installation:

```bash
# Basic installation with core dependencies
pip install -U --pre mesa

# Install with recommended dependencies
pip install -U --pre mesa[rec]

# Install with specific additional dependencies
pip install -U --pre mesa[network,viz]

# Install all dependencies (including dev dependencies)
pip install -U --pre mesa[all]
```

### Impact
- Existing projects using only core Mesa functionality should not be affected.
- Projects using network or visualization features may need to update their installation to include the appropriate optional dependencies.

### Points to Review
1. Are the chosen dependency groups (`[network]`, `[viz]`, `[rec]`, `[all]`) appropriate and sufficient?
2. Is the README clear about the new installation options?
3. Are there any parts of the documentation or examples that need updating to reflect these changes?

Closes #626

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,17 @@ To install our latest pre-release (3.0.0 alpha), run:
 ``` bash
 pip install -U --pre mesa
 ```
+With Mesa 3.0, we don't install all our dependencies anymore by default.
+```bash
+# You can customize the additional dependencies you need, if you want. Available are:
+pip install -U --pre mesa[network,viz]
+
+# This is equivalent to our recommended dependencies:
+pip install -U --pre mesa[rec]
+
+# To install all, including developer, dependencies:
+pip install -U --pre mesa[all]
+```
 
 You can also use `pip` to install the latest GitHub version:
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,28 +36,37 @@ classifiers = [
 ]
 readme = "README.md"
 dependencies = [
-  "click",
-  "cookiecutter",
-  "matplotlib",
-  "networkx",
   "numpy",
   "pandas",
-  "solara",
   "tqdm",
 ]
 dynamic = ["version"]
 
 [project.optional-dependencies]
+# User collections
+rec = ["mesa[network,viz]"]
+all = ["mesa[network,viz,dev,examples,docs]"]
+# Core dependencies
+network = [
+  "networkx",
+]
+viz = [
+  "matplotlib",
+  "solara",
+]
+# Dev and CI stuff
 dev = [
-  "ruff~=0.1.1",   # Update periodically
-  "pytest >= 4.6",
+  "mesa[rec]",
+  "ruff",
+  "pytest",
   "pytest-cov",
   "sphinx",
   "pytest-mock",
 ]
 examples = [
-  "pytest >= 4.6",
-  "scipy"
+  "mesa[rec]",
+  "pytest",
+  "scipy",
 ]
 docs = [
   "sphinx",
@@ -65,7 +74,7 @@ docs = [
   "pydata_sphinx_theme",
   "seaborn",
   "myst-nb",
-  "myst-parser",         # Markdown in Sphinx
+  "myst-parser",  # Markdown in Sphinx
 ]
 
 [project.urls]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,6 +62,7 @@ dev = [
   "pytest-cov",
   "sphinx",
   "pytest-mock",
+  "cookiecutter",
 ]
 examples = [
   "mesa[rec]",


quaquel:
> 1. Do we still want to allow direct imports for model and agents (like from mesa import Agent, Model)?

Yes, I am strongly in favor of a simple high-level API for the core functionality. Why should a novice user bother with the internal module structure of MESA?

> 2. Do we still want to allow direct imports like for space and time (like from mesa import RandomActivation, SingleGrid)?

Again, yes. The main issue is what is part of the default set of things that users might use for 80% of the models?

> 3 Do we want to allow direct imports for DataCollector and batch_run, if the required dependencies are installed? If so, what > do we do if that fails, throw a warning? What if it's the intended behavior for the user?

As far as I know, during coding dependencies are not checked so while running an import error will automatically be raised. 

> 4. Any suggestions on how the dependency groups in  pyproject.toml are structured?

I'll try to look at this asap.

> 5. Did I miss any documentation, tutorials, etc.

This PR is getting a bit big in terms of affected files. Might it be possible to split it into two: update toml, update docs?

EwoutH:
Thanks for the review!

On point 2 and 3, the biggest issue is in the init:

```diff
 __all__ = [
     "Agent",
     "time",
     "space",
-    "DataCollector",
-    "batch_run",
-    "experimental",
 ]
```
DataCollector, batch_run and experimental are removed because they contain dependencies that are note installed by default anymore. Not everyone might need a DataCollector, batch_run or experimental feature, so that's the idea.

The thing is, ideally you don't import those three if they are not used. As far as I have found out, you have two options:
1. Always attempt to import them
   a. Warning if they can't be. But it's weird behavior, because why get a warning from a module you might not even need?
   b. Do it silently. This may be confusing, why does it work sometime and other times not?
2. Try some lazy import evaluation. But that's difficult, since it's running in `__init__.py`. Haven't found a satisficing solution.

quaquel:
I quickly checked the toml file. It seems sensible not to always install, e.g., networks or Solara. I am less sure about pandas. In my view this is such a standard default library, why not always include this. TQDM is very lightweight as far as I know, so why not always also install this? My personal view on this, in general, is that this fine sliceing of dependencies is a bit odd given Python's overall batteries included philosophy. In particular, if dependencies are stable and readily available (so not requiring conda forge or so), I rarely see the point of not simply installing everything.

EwoutH:
I think there is a middle way feasible here. Don't underestimate how heavy pandas is though. One example is that they currently don't have Python 3.13 wheels (months after the betas, and two weeks before stable release), and building (in CI) takes a full 4 minutes. But I agree there not that many scenarios you don't need it.

Including `tqdm` and `pandas` in the core could be a way to go, to enable `batch_run` and `DataCollector` by default. Curious what the rest thinks about that.

rht:
I'm not sure of the performance impact, but one could import `pandas` inside of the `DataCollector` class instead of globally within the datacollection.py file. This way, if an import error happens, you could provide a helpful error message saying that pandas needs to be installed. Needs to see how Solara does it with their `FigureMatplotlib` component (Matplotlib is not a dependency of Solara). Also, Solara has recently separated the heavy server code from their core package. I read it somewhere in a thread, that it is possibly solara-ui and solara-server, but couldn't find an evidence of this after reading their pyproject.toml.

EwoutH:
> I'm not sure of the performance impact, but one could import `pandas` inside of the `DataCollector` class instead of globally within the datacollection.py file.

I think we can do this quite elegantly:

```Python
class DataCollector:
    ....

    @cached_property
    def _pd(self):
        """Lazy-loaded pandas module using cached_property."""
        try:
            import pandas as pd
            return pd
        except ImportError as e:
            raise ImportError(
                "Pandas is required for this operation. Please install it using 'pip install pandas'."
            ) from e
            
    def get_model_vars_dataframe(self):
        ...
        return self._pd.DataFrame(self.model_vars)
```
Also, in the datacollector, Pandas is generally only used on methods that are ran after the model is done, like `get_model_vars_dataframe`. The lazy import allows to not install in on - let's say - a super computer, and do the analysis afterwards.

However, I'm also still considering making `tqdm` and `pandas` required dependencies again.

rht:
I was stating `pandas` just as an illustration, as it applies to other optional dependencies such as `matplotlib`. This is how Solara does it for Altair: https://github.com/widgetti/solara/blob/2cf59d9bb40ed7d10d976c9c44c292d61e01ec89/solara/components/figure_altair.py#L24.

docs/tutorials/intro_tutorial.ipynb
```diff
@@ -60,7 +60,13 @@
...
     "Create and activate a [virtual environment](http://docs.python-guide.org/en/latest/dev/virtualenvs/). *Python version 3.10 or higher is required*.\n",
     "\n",
-    "Install Mesa:\n",
+    "Install Mesa with the recommended dependencies.\n",
+    "\n",
+    "```bash\n",
+    "pip install --upgrade mesa[rec]\n",
+    "```\n",
+    "\n",
+    "If you already have all the dependencies installed or are just updating, you can install Mesa without the recommended dependencies:\n",
```
tpike3, line 69: How does this deconflict from the stable version (e.g. 2.3.4) and the pre-release? 

As one has [stable in the url](https://mesa.readthedocs.io/en/stable/tutorials/intro_tutorial.html) and one has [latest in the url](https://mesa.readthedocs.io/en/latest/tutorials/intro_tutorial.html). The challenge at this point is there so many changes model implementation will be pretty different.  

I think it would be too painful to try and keep the stable and the latest tutorials, but maybe explicitly call out of the user to check which version they are in???

I am not sure the right answer this is an interesting problem.

tpike3 (changes requested):
@EwoutH This is awesome!

I had one question on the import and thinking through latest and stable readthedocs pages. Not sure the best answer, except may be calling explicitly that there is a difference.

My thought on the discussion is and to @quaquel's comments of "batteries included" can you set it up so `all` is installed by default (e.g. if no specification install all or rec) this is good for new users but for more advanced users they can specify specific dependencies they want. (e.g. which sub categories they want)

docs/tutorials/intro_tutorial.ipynb
EwoutH: It’s quite elegant: the stable one is using the latest stable release, so now 2.3.4. We generally point users to that.

The latest release is built from `main`, which would be updated if this PR got merged.

Once we then release a 3.0.0 stable release, the stable docs get switched over also.

EwoutH:
Thanks for reviewing! I answered your comment on the docs.

I will try to update the introduction tutorial, but we probably have to rewrite large parts of it. I did a small update in https://github.com/projectmesa/mesa/pull/2315, but removing the schedulers requires completely rewriting the Agent activation parts.

tpike3:
> Thanks for reviewing! I answered your comment on the docs.
> 
> I will try to update the introduction tutorial, but we probably have to rewrite large parts of it. I did a small update in #2315, but removing the schedulers requires completely rewriting the Agent activation parts.

Ha just saw #2315 and went right into the code so didn't see everyone else comments, as such I updated my comment. 

Thanks for all this work, this is such great progress!

docs/tutorials/intro_tutorial.ipynb
tpike3: I understand that, but with `a5` intro tutorial switching to new `AgentSet` set up I think we need to explicitly call out there is a difference as that subtlety can be lost on users. For example, if I send out a link to latest to a new user but then they search Mesa later on they will likely get stable and the differences I think will only confuse them. 

Make sense?

tpike3:
Just want to make sure you know the readthedocs build is failing due to datacollecter; it also seems there is a bunch of collisions on the names 

image

docs/tutorials/intro_tutorial.ipynb
EwoutH: Yeah we could add notes to both that one is for Mesa 2.3.x and the other is for Mesa 3.0 pre-releases. We did that already with the Viz tutorial.

EwoutH:
Okay, I vastly simplified this PR, by keeping Pandas and tqdm core dependencies.

Now only the pyproject.toml and Readme are updated. Please review.

pyproject.toml
```diff
@@ -36,36 +36,46 @@ classifiers = [
...
 dev = [
-  "ruff~=0.1.1",   # Update periodically
-  "pytest >= 4.6",
+  "mesa[rec]",
+  "ruff",
+  "pytest",
   "pytest-cov",
   "sphinx",
   "pytest-mock",
+  "cookiecutter",
```
quaquel, line 65: Not sure whether this belongs in dev. Dev for me means further development of mesa itself. However, if I understand this correctly, it is used to setup a default project which uses mesa. So then it should be a core dependency.

quaquel (changes requested):
Sorry one last question about cookiecutter. But yes this looks fine and clear now.

pyproject.toml
```diff
@@ -36,36 +36,46 @@ classifiers = [
...
 ]
 examples = [
-  "pytest >= 4.6",
-  "scipy"
+  "mesa[rec]",
+  "pytest",
```
Corvince, line 69: why is pytest a part of "examples"?

EwoutH: Pytest is also used to test the examples. The CI examples workflow uses this for example.

pyproject.toml
EwoutH: Yeah bit in doubt if we want to ship this by default. It has quite a lot of required dependencies:

> [arrow](https://pypistats.org/packages/arrow) | [binaryornot](https://pypistats.org/packages/binaryornot) | [click](https://pypistats.org/packages/click) | [jinja2](https://pypistats.org/packages/jinja2) | [python-slugify](https://pypistats.org/packages/python-slugify) | [pyyaml](https://pypistats.org/packages/pyyaml) | [requests](https://pypistats.org/packages/requests) | [rich](https://pypistats.org/packages/rich)

Of which some have their own dependencies, etc.

Related to https://github.com/projectmesa/mesa/issues/2331.

Corvince: Yes, cookiecutter and also click should be part of core mesa, since they provide the "mesa startproject / mesa run" command. While I think we should deprecate those in favor of GitHub templates for the scope of this PR they should belong to core

Corvince:
> Okay, I vastly simplified this PR, by keeping Pandas and tqdm core dependencies.
> 
> Now only the pyproject.toml and Readme are updated. Please review.

Its indeed much better now, although pandas is really a mixed bag. Previously I would have said everyone who uses mesa should probably also use pandas anyway, but nowadays with polars, ibis and others this is much less true. And, as far as I remember, we only  use pandas in one place, to convert the data collected data into a dataframe. I don't think that is really worth the dependency. But lets maybe do that in a separate PR, to move this one forward.

pyproject.toml
quaquel: I agree and would be in favor of just removing the cookie-cutter stuff for mesa 3.

EwoutH: I'm going to investigate this before the next alpha.

quaquel: ok, then let's get this merged

README.md
```diff
@@ -39,6 +39,17 @@ To install our latest pre-release (3.0.0 alpha), run:
 ``` bash
 pip install -U --pre mesa
 ```
+With Mesa 3.0, we don't install all our dependencies anymore by default.
+```bash
+# You can customize the additional dependencies you need, if you want. Available are:
+pip install -U --pre mesa[network,viz]
+
+# This is equivalent to our recommended dependencies:
+pip install -U --pre mesa[rec]
```
rht, line 48: Hold on. This will trip people over. Visualization is what a lot of people coming to Mesa want to use it for. The default needs to include visualization. I can do `pip install solara` and have the common functionality working out of the box. Why do I have to do `pip install mesa[rec]` specifically for just Mesa?

rht: Doing a minimal install of Mesa is less common than doing an install of Mesa for visualization in a classroom setting. Can the split be `mesa-minimal`/`mesa-core` and `mesa` instead?

EwoutH: I have multiple complex models that don't touch Solara, the batch_run or the DataCollector at all.

Also the initial error message when you can't install Solara is quite clear.

Also it's literally in our front-page Readme.

quaquel: You also typically want to avoid installing stuff like solara on an HPC. As long as it is well documented, which it is, I am fine with the current structure.

rht: I have similar situations (initially wanted to embed Mesa inside other simulation frameworks), but why make it more inconvenient to students and first time learners, who have to read the fine print of the README.md?

EwoutH: We can put it in a few other places, like the visualization tutorial, and potentially also add a helpful warning on import somewhere.

Corvince: Sorry I also didn't saw this properly before, but I side with @rht on this one. Viz should be in core imho. Mesa should be "batteries included". For hpc or similar environments you can always install with "--no-deps". And in the future it might be more convinient to have a "mesa-core" package with minimal dependencies. But for now it should be included

EwoutH: You know you where the person who opened https://github.com/projectmesa/mesa/issues/626 ;)

Let’s discuss at the next dev meeting.

PR #2267, opened 2024-09-02
Model: Replace `get_agents_of_type` method with `agents_by_type` property

EwoutH:
This PR replaces the Model method `get_agents_of_type()` with an `agents_by_type` property, which directly returns the dict.

Instead of using:
```Python
model.get_agents_of_type(Sheep)
```
You should now use:
```Python
model.agents_of_type[Sheep]
```
Since we also use `agents` and not `get_agents`, it's more intuitive to directly have access to the object itself (in this case the `agents_by_type` dict). It's also is more concise and since you have full dict access, more flexible.

Examples and tests are updated and an deprecation warning is added for `get_agents_of_type()`.

### DIFF
--- a/benchmarks/WolfSheep/wolf_sheep.py
+++ b/benchmarks/WolfSheep/wolf_sheep.py
@@ -227,8 +227,8 @@ class WolfSheep(Model):
             patch.move_to(cell)
 
     def step(self):
-        self.get_agents_of_type(Sheep).shuffle(inplace=True).do("step")
-        self.get_agents_of_type(Wolf).shuffle(inplace=True).do("step")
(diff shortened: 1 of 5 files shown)


EwoutH:
Fair point. At this point we can just expose the `agents_by_type` dict though a property, right?

EwoutH:
Implemented as suggested!

quaquel:
If you add a test for this in test_model, its good to go as far as I am concerned.

rht (approved):

EwoutH:
Tests are added and the PR description and title are updated with some explanation and examples.

Thanks for reviewing!

PR #2275, opened 2024-09-04
Add pytest script and CI job to test tutorials

EwoutH:


EwoutH:
Maybe #2276 is enough, let's see.

EwoutH:
Fun idea, I will leave the branch open for a while, but #2276 is far more elegant than this PR and also solves the problem, using a build we already have set up and use.

So closing in favor of #2276.

PR #2277, opened 2024-09-04
Update issue template to point directly at discussions

EwoutH:
Update issue template to point directly at discussions, instead of doing that in the "ask-help" template.

- Add a `config.yml` file with `contact_links` section that points directly to our discussion forum
- Remove the `ask-help.md` template
- Update the Bug report template to ask for a Include a minimal reproducible example.

Based on the GitHub [docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser).

### DIFF
--- a/.github/ISSUE_TEMPLATE/asking-help.md
+++ b/.github/ISSUE_TEMPLATE/asking-help.md
@@ -1,9 +1,0 @@
----
-name: Asking for help
-about: If you need help using Mesa, you should post in https://github.com/projectmesa/mesa/discussions
----
-
-<!--
-    ATTENTION: Don't raise an issue here!
-    If you need help, ask in https://github.com/projectmesa/mesa/discussions
--->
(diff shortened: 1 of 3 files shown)


jackiekazil:
LGTM

jackiekazil (approved):

EwoutH:
Thanks for merging, links work, and now looks like this:

image

No idea where the "security vulnerability" section came from or how to turn it off.

jackiekazil:
I thought that was enabled by default.

EwoutH:
Yeah since we’re pure Python vulnerabilities aren’t really relevant for us. I would propose disabling it to prevent confusion and false positives.

jackiekazil:
https://github.com/orgs/community/discussions/45567#discussioncomment-10550883

jackiekazil:
It also covers other vulnerabilities - such as vulnerabilities in our setup or procedures if someone observes them. I am not against disabling; it is a rare exception, and having the option is visually noisy. 
If someone observed something - how do you suggest they report it without that option? That option keeps reports private with the idea that you wouldn't want them published world until you were ready to do so and mitigate them.

PR #2278, opened 2024-09-04
move solara_viz back to experimental

Corvince:
Move the current SolaraViz back into experimentals. This way users of Mesa 2.x that already adapted their frontend can continue to use it. But we will provide a different, stable API for Mesa 3 in https://github.com/projectmesa/mesa/pull/2263 

Follow-up PRs should include a deprecation warning and link to a migration guide from the experimental SolaraViz to the new SolaraViz (to be written)


EwoutH (approved):
Right, I like this.

Warning can be added once new-new viz is semi-stable.

EwoutH:
Maybe we can also revert it back to JupyterViz. And then we can simply revert https://github.com/projectmesa/mesa-examples/pull/155.

Corvince:
SolaraViz and JupyterViz both work. But the examples need to be updated anyway, so not sure. 

By the way I can take care of updating the examples. This way I get a better feeling on how the frontend is used and directly fix anything that's missing

Corvince:
@rht do you agree with this approach? Would like to have your approval before merging

rht (approved):

rht:
I agree.

EwoutH:
This PR did break the CI, unfortunately:

```
__________________ ERROR collecting tests/test_solara_viz.py ___________________
ImportError while importing test module '/home/runner/work/mesa/mesa/tests/test_solara_viz.py'.
Traceback:
/opt/hostedtoolcache/Python/3.12.5/x64/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_solara_viz.py:8: in <module>
    from mesa.visualization.solara_viz import Slider, SolaraViz, UserInputs
E   ModuleNotFoundError: No module named 'mesa.visualization.solara_viz'
```

PR #2280, opened 2024-09-05
add default SolaraViz

Corvince:
add default visualization, if no components are provided (or the string "default")


EwoutH (approved):
Nice, I think it helps that people using the viz directly notice that it's there for them to use and modify.

PR #2282, opened 2024-09-05
Simplify ModelController

Corvince:
Not the most urgent of things to do, but I was in the mood to play around with this. Overall, should be much simpler now.

### DIFF
--- a/mesa/visualization/solara_viz.py
+++ b/mesa/visualization/solara_viz.py
@@ -24,10 +24,9 @@ See the Visualization Tutorial and example models for more details.
 """
 
 import copy
-import threading
+import time
 from typing import TYPE_CHECKING, Literal
 
-import reacton.ipywidgets as widgets
 import solara
 from solara.alias import rv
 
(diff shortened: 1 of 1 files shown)


EwoutH (approved):
I have no idea what any of this code does, but I always like simpler!

mesa/visualization/solara_viz.py
```diff
@@ -154,110 +153,49 @@ def ModelController(model: solara.Reactive["Model"], play_interval):
...
-            playing.value = False
-        elif model.value.running:
+    def step():
+        while playing.value:
+            time.sleep(play_interval / 1000)
             do_step()
-        else:
-            playing.value = False
+
+    solara.use_thread(step, [playing.value])
```
rht, line 175: Needs to test if this works on Google Colab.

rht:
It is indeed simpler, more readable, and hence can be extended by users if they need to fork the component. The only concern is if it works on Google Colab.

Corvince:
Confirmed that it works in Google Colab!

PR #2289, opened 2024-09-09
Update viz tutorial to the new API

Corvince:
Quick update to use the new SolaraViz Update. More updates to highlight more features should go into a separate PR, this is just to have it working again

### DIFF
--- a/docs/tutorials/visualization_tutorial.ipynb
+++ b/docs/tutorials/visualization_tutorial.ipynb
@@ -11,7 +11,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "*This version of the visualisation tutorial is updated for Mesa 3.0, and works with Mesa `3.0.0a1` and above. If you are using Mesa 2.3.x, check out the [stable version](https://mesa.readthedocs.io/en/stable/tutorials/visualization_tutorial.html) of this tutorial on Readthedocs.*\n",
+    "*This version of the visualisation tutorial is updated for Mesa 3.0, and works with Mesa `3.0.0a4` and above. If you are using Mesa 2.3.x, check out the [stable version](https://mesa.readthedocs.io/en/stable/tutorials/visualization_tutorial.html) of this tutorial on Readthedocs.*\n",
     "\n",
     "**Important:** \n",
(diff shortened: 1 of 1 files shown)


EwoutH:
Thanks!

wang-boyu:
Quick question: did the model in the viz tutorial run very slowly with the new Solara viz api, as compared with the previous implementation? @Corvince 

Same thing happened for the mesa-geo intro tutorial model, when I try to update the tutorial to use the new api (https://github.com/projectmesa/mesa-geo/pull/246).

Corvince:
Hmm I haven't noticed any significant performance degradations. Under some circumstances that I do not fully understand yet I experience it to be quite slow, but this has also happened to me with the old API. If there are any performance problems it will probably be due to the updates of the ModelController. 

But I also think the frontend is too slow. But this seems to be mainly due to the used components. I tested displaying only the number of steps and  this runs with several dozen steps/sec, so the solara stack doesn't seem to be the problem, mainly the space drawing is. Maybe this could somehow be optimized

rht:
I tested the viz tutorial on Colab with an intent to reproduce the slowness. Encountered another issue: it produced the Matplotlib output just fine, but the control buttons (play/pause/reset/reseed/...) are not rendered, and so I couldn't play the simulation to see whether it is the evolution that is slow or something else.

Corvince:
That's bad! I sometimes had the local issue that the ModelController had appeared after the main components. Maybe that's related. I will also investigate this on colab

wang-boyu:
Thanks for your reply @Corvince. I see that in the new Model controller, the entire `model` is converted to reactive, whereas previously a `current_step` counter is used. Could this potentially cause slowness, since we need to track the entire model object now?

Would you like to try this on your machine @rht ? Thanks for looking into this!

rht:
> Would you like to try this on your machine @rht ? Thanks for looking into this!

Running locally, I was able to click the hamburger button, which then revealed the control buttons. And I confirm that each steps took 2-3 s, i.e. very slow for a simple Boltzmann wealth model.

Corvince:
Strange, i'll have to look at it from another laptop. Does this happen for both step and play/pause?

rht:
The step is much more immediate, sub 1 s response time.

PR #2291, opened 2024-09-11
Add support for Observables to MESA

quaquel:
see #2281 

This PR adds various new classes to MESA. The key public-facing ones are, at the moment, `Observable,` `ObservableList,` and `HasObservables`. `Observable` is a descriptor that is used to declare a given attribute to be observable. Being observable means changing the attribute using the assignment operator `=` willl result in the sending of a signal. Callbacks can subscribe to this signal. `ObservableList` is a proof of principle on how the idea of an observable can be expanded to collections. Future extensions could include, e.g., `ObservableDict`. `HasObservables` is a mixin that can be used with e.g., `Model` or `Agent`.  It is required if you want to use `Observable` or `ObservableList`. `HasObservable` contains the key logic for subscribing to signals and the emitting of signals.

`Observable` emits a `change` signal when its attribute value is changed via the assignment operator (i.e., `=`). `ObservableList` at the moment defines five signal types: `"replace", "remove", "insert", "append", "change"`. `"change"` is emmited when the entire list is changed via the assignment operator (i.e., `=`). The other signals are tied to the respective list operations. Currently, signals are following the approach used in traitlets. The main reason for making the signals identical to traitlets is that it should make integration with solara easier, which already supports traitlets.

The emitted signals are instances of an `AttributeDict`. This is in line with Traitlets. It is just a dict with attribute-style access to the fields. By default, a signal has four fields: "owner", "type", "old", and "new". It is possible to add additiona fields if the user wants. For example, the `ObservableList ` signals (accept `"change"`) all also include `"index"`.

I have also added a WIP `Computed` and `Computable` approach. In javascript-style signals, but, e.g., also in [solara](https://solara.dev/documentation/api/utilities/computed), a Computed is a callable that is dependent on one or more signals. When it receives a signal that any of its dependencies has changed, it will mark itself as being outdated. It will only update its value if it is called to do so. This can be used to set up a hybrid push/pull structure where Obervables push signals, but a Computed is only recalculated on a pull if it knows its underlying data has changed. A Computed is also Observable, so it will emit a signal if it knows it is outdated. This allows the chaining of observables and computed's. 

# API
Below, there is a quick sketch of the resulting API. We define Obervable attributes at the class level. We can use the attribute normally in the rest of the code. 
```python

class MyAgent(Agent, HasObservables):
    wealth = Observable()
    income = Observable() 

   def __init__(self, model)
       super().__init__(model)
       self.wealth = 5  # we can use wealth as a normal attribute, Observable hides all the magic

class MyModel(Model, HasObservables):
    gini = Computable()
    agent_wealth = ObservableList()

    def __init__(self, random=random):
        super().__init__(random=random)
        
        agent_wealth = [agent.wealth for agent in self.agents]  # ObservableList can be used as a normal list
        self.gini = Computed(calculate_gini, agent_wealth)

model = MyModel()
agent = MyAgent(model)

# whenever wealth is changed some_callback_function will be called
agent.observe("wealth", "change", some_callback_function)

# Whenever either income or wealth is changed, generic_callback_function
# is called. All() is a new helper class
agent.observe(All(), "change", generic_callback_function)

# observables is a class-level attribute with all observables defined for the class
print(agent.observables)

```

# Implementation Details
After a long discussion in #2281, I decided to implement everything from scratch. Existing libraries like traitlets or psygnal add too much overhead because they are not tailored to the specific use case of ABMs with potentially many updates to an attribute. 

# future extensions
This PR puts in place the basic machinery, but there are ample options for further improvements:

- At the moment, anything that wants to emit a signal needs to be registered as being observable. Ideally, something like `model.agents` should emit `agent_added` and `agent_removed` signals. At the moment, there is no machinery in place to make it easy to add this functionality. I don't think it is overly difficult, but I want to take a closer look soon.
- I currently include only an ObservableList as an example of how we can use observables to also observe changes in collections. Other collections can easily be added if and when desired.
- The current implementation of `Computed` is heavily inspired by signals and Solara. It involves the autodiscovery of dependencies. The params library, instead, forces users to pass the list of dependencies explicitly. I am not sure which approach is better. 
- I don't like the current API for `Computed`. It requires the user to do 2 things: declare an attribute as being computable at the class level and assign a computed instance to this attribute during object initialization. If I fix the first bullet, it might be possible to improve this API and eliminate the `Computable` declaration.  
- The way in which the signals that a given object can emit is declared is less than ideal. `Observable` and `ObservableList` both have a list of strings that enumerate the signals that can be sent. However, there is no explicit declaration anywhere of the fields that are part of any signal. A signal is an instance of an `AttributeDict`, following traitlets, but it is completely up to the user to decide what fields will be in it. This means that the user is dependend on the docstring, or some test code to discover what is in a signal, rather than that it can be declared and checked explicitly via e.g., a type checker. 
- At the moment, the `notify` method, which is responsible for emitting the signals requires the name of the observable (`name`), the old value (`old`), the new value (`new`), and the signal type (`type`). Further elements can be added as desired. For example, most signals from `ObservableList` include `index`. However, for e.g., an `append` signal, having `old` in there makes little sense.


EwoutH:
Could you update or rebase this branch on `main`?

quaquel:
The most basic version of `Computed` now works. See [this simple Boltzman example for the draft API](https://github.com/projectmesa/mesa/blob/43b2a7bad937f898d2c42fe094cd7d9d65974f29/mesa/experimental/signals/test.py). 

What works is a simple chain of observable --> computed --> computed. So if the observable is updated, this emits a signal to the first computed, which then relays it to its children. In the example, this is `agent.wealth` --> `model.agent_wealth` --> `model.gini`. Now, when the data collector wants to retrieve `model.gini`, the Computed/Computable checks whether it is dirty (i.e., is outdated) and, if so, will recalculate. This requires, in turn, calling `model.agent_wealth`, which does the same check. 

What does not work at the moment is the correct detection of cycles. So Computable A depends on Computable B, which depends on C which in turn depends on A. At the moment, I guess this will trigger some kind of infinite recursion error. I hope to fix this soonish.

benchmarks/BoltzmannWealth/boltzmann_wealth.py
```diff
@@ -74,9 +75,11 @@ def run_model(self, n):
             self.step()
 
 
-class MoneyAgent(mesa.Agent):
+class MoneyAgent(mesa.Agent, HasObservables):
```
rht, line 78: How expensive would it be for `mesa.Agent` to automatically inherit `HasObservables`? The less boilerplate, the better.

quaquel: It would not be expensive at all. The same applies to Model, and possible in the future the spaces. However, I have atm designed everything to be in experimental so as not to interfere with the stable code.

Corvince:
Awesome stuff! Just by looking at the code I can tell you are having a lot of fun with this one, am I right?

I am just slightly worried about the code complexity this is getting to. When you are done with everything else you want to do with this, could you maybe add a very simple data collector prototype to the example? I would like to then replicate the example using the param library, to have a more holistic comparison about the performance difference. So don't think too much about a possible api for the data collection, just some object where data is collected/updated/persisted once per step

quaquel:
> Awesome stuff! Just by looking at the code I can tell you are having a lot of fun with this one, am I right?
> 
> I am just slightly worried about the code complexity this is getting to. When you are done with everything else you want to do with this, could you maybe add a very simple data collector prototype to the example? I would like to then replicate the example using the param library, to have a more holistic comparison about the performance difference. So don't think too much about a possible api for the data collection, just some object where data is collected/updated/persisted once per step

It's been fun indeed. The basic Observable stuff is quite straightforward and not that complicated. Complications arise due to 2 things: adding in Computable, and avoiding memory leaks. 

I have followed the draft signals library in the current setup of Computable. This means that dependencies are automatically and on the fly discovered and can also dynamically change over the course of a running program. I am not sure this is actually needed for something like mesa. The alternative I see (and just noticed is used by `params.bind`) is to explicitly indicate the dependencies. This might make certain parts of the code quite a bit shorter and cleaner. But you freeze the dependencies explicitly. 

Memory leaks are another issue. Subscribing to signals involves references between objects. Ideally, you want all of this to be weak references to ensure that, for example, an agent who is listening to some other agent can still be properly removed from the simulation. Part of the tricky thing here is that weakreffed bound methods differ from normal weakrefs, and the off-the-shelve weakref collections (dict/set) don't play nice with bound methods. I hope to move some stuff into a separate utils package to keep the main code cleaner.

Regardless, a comparison to params would indeed be very nice.

quaquel:
> I am just slightly worried about the code complexity this is getting to. When you are done with everything else you want to do with this, could you maybe add a very simple data collector prototype to the example? I would like to then replicate the example using the param library, to have a more holistic comparison about the performance difference.

The code is basically ready for such a comparison. I have started doing some timings myself to see what the overhead is. It seems that the autodiscovery of Computable can become quite expensive. So @Corvince, do you think we need this or might be get by by the approach used by [`param.bind`](https://param.holoviz.org/reference/param/generated/param.bind.html#param.bind) to explicitly indicated what the arguments and keyword arguments to the callable are that must be observed?

Corvince:
I started to prepare such comparison, and from the documentation it seemed straightforward to do in `param`, but it actually isn't. I also experienced some errors, that seemed to be caused by some initialization clash between param and agents. What I want to say is, it is not that easy and seemed quite advanced and different than "typical" python. I hope to still finish the comparison, but I am even less convinced that `param` is the way to go (Except for the param specification itself, that seems to be super useful). 

Regarding the necessity of "autodiscovery": I would say it depends on the API implications. I always prefer to stay simple and expand later, so even without "computables" this PR would be a nice start. Or do you want to keep computables, but make the dependency explicit?

quaquel:
> Regarding the necessity of "autodiscovery": I would say it depends on the API implications. I always prefer to stay simple and expand later, so even without "computables" this PR would be a nice start. Or do you want to keep computables, but make the dependency explicit?

I don't know yet. At a minimum, I want to give the  `param.bind` approach to computables a go to see how difficult it would be to implement and to see the performance implications. 

The rest of this PR is indeed ready for feedback and discussion.

Corvince:
I thought about this whole issue a lot more and the distinction between event listeners, signals, computables, etc. Here is how I would like this to work from the point of a data collector. Please let me know how much this aligns with the current state, because I kind of lost the current state due to the rising complexity. This by no means should sound authorative, this is just how I currently imagine things. 

So lets stay with the boltzmann example and asume we want to gather two kinds of data: The individual wealth of each agent and the gini coefficient. 

Now in an observable world the data collector observes each agents "wealth" and is notified each time that changes. This is good and simple if the attribute doesnt change very often. But when the wealth changes multiple times during each "step" the datacollector gets flooded with information that it doesnt need, because it wants to collect the data only once after each step. 

So in that case I think it would be more desirable to only notify the data collector once to let it know the wealth changed. After that the data collector marks the wealth as "dirty" and stops listening to events until it needs the current value. That means at the end of the step the datacollector retrieves the current value from the agent and starts to listen again for any changes. 
That is at least how I understand the signals API, is that correct so far?

Now for the gini coefficient the simplest solution for the data collector would be to just store it as a @property on the model. Then the data collector could just collect it as any other attribute and its lazily evaluated only when needed. However in a signal world it would probably be a computed value, that itself listens to each agents wealth. With observables this would become expensive again, but with computed it should only listen for the first change event and then stop listening. 

Is this somewhat along the lines you implemented it? Because circling back to the question if "computed" properties are needed, I don't know how else derived values like the gini coefficient should work?

quaquel:
@Corvince yes this tracks largely with how I look at it. A few remarks

1. Currently, I don't unsubscribe upon marking a computable as dirty. I'll test the performance implications of this. The issue is that you would have to unsubscribe from many agents. It might be cheaper to receive all messages because the handling is cheap (I am already dirty, so ignore the signal). The draft python signals library did not ussubscribe after the first signal and from looking at Solara, I think that it also does not unsubscribe. 
2. Most signal libraries can pause or stop the emissions of signals. This might be a complementary mechanism, currently not implemented, that could be used to avoid flooding observers with signals. For example, if you have some agent method that repeatedly updates an observable attribute, observers should only receive a signal for the last and final update. In such a case, having a way of pausing signals is useful. Various Python libraries use a context manager for this.
3. Yes `Gini` is a `Computed` and would take, ideally, the data collected agent wealth as its input. The data collected agent wealth would then also be a `Computed`, and signals would propagate. So, upon the first `agent.wealth` update, the data collector for this would receive a signal, mark itself as dirty, and broadcast this to its dependents. Gini, as a dependent, would then also mark itself as dirty. This is in essence up and working already for the Boltzman benchmark model.

Corvince:
Thanks for the additional insights! 

1. Yes that makes sense I guess. Much simpler to just ingore messages
2. At some stage that might make sense. I will come back to that below
3. Interesting order. That makes a lot more sense. In my head it was agent.wealth -> gini -> data collector. So both data collector and gini would observer agent.wealth. But yours makes more sense and is less expensive.


I am still a bit skeptical about the performance implications. Overall seems like a lot of additional operations in case of "volatile" attributes that change often. Also with lots of agents its a lot of subscriptions. I am wondering if we can somewhat use `agentset` as a middle man. So data collector would subscribe to agentset. Agentset already is connected to all agents, so maybe we can use that connections to pass agent messages to agentset. And Agentset than relays a single message with all the agents/attributes that should be marked as dirty (or the latest value). There, supressing messages while executing a single "do" or "apply" method might make sense.

quaquel:
To give you a sense of the overhead, Here is a quick benchmark based on the current version of the code for the Boltzman wealth model. This benchmark is not based on something that is feature-complete. I have added a Table to track agent wealth. This table is then used in the gini calculation. So this table at all times reflects the value of `wealth` for all agents. 

In this version, gini is not a computed because making that work with this table is a bit tricky (which suggests that the computed stuff needs a bit more thought to make it easy to extend and use). 

```python
class Table:
    """This table tracks the wealth of agents."""

    def __init__(self, model: BoltzmannWealth):
        self.data = {}
        for agent in model.agents:
            agent.observe("wealth", "on_change", self.update)
            self.data[agent.unique_id] = agent.wealth

    def update(self, signal):
        """handler for signal that updates the wealth for the agent in the table"""
        self.data[signal.owner.unique_id] = signal.new_value

    def get(self):
        """return the wealth of agents"""
        return self.data.values()

```

So, the startup is much slower (and I have not tried to improve this at all yet). Overhead while running is present but not bad. And remember, Boltzmann is a worst-case example, and none of the code here has been profiled. I have analyzed some parts a little bit for performance, but not in any great depth. 

| Model           | Size   | Init time [95% CI]            | Run time [95% CI]          |
|-----------------|--------|-------------------------------|----------------------------|
| BoltzmannWealth | small  | 🔴 +302.2% [+295.2%, +309.0%] | 🔴 +8.6% [+8.2%, +9.1%]    |
| BoltzmannWealth | large  | 🔴 +407.5% [+375.9%, +467.4%] | 🔴 +24.8% [+23.9%, +25.8%] |


So why is startup so slow? There is quite some overhead from initializing the agents with the additional data structures for emitting signals etc. Moreover, at the moment of object instantiation, the code figures out what observables have been defined and adds these to the instance. This is silly, and I have to figure out a way to do this once during class creation (instead of 100 times for each agent).  Also, the percentages mislead for startup time. We go from roughly 0.00026 seconds to 0.001 seconds.

Corvince:
I am positively surprised by the benchmark! It is really not bad for a worst case scenario. I agree that startup times are not so relevant and can probably be improved. @EwoutH maybe it makes sense to also include absolute times in the comment? I think its different for different models, but in case where startup time makes up less than 1% of total runtime its totally negligible.

But can you maybe run these benchmarks also for the "large" model? I want to see how it scales

quaquel:
> But can you maybe run these benchmarks also for the "large" model? I want to see how it scales

I am not sure what you mean. The benchmark results include both the small and the large version of the Boltzman wealth model.

EwoutH:
I like the API!

mesa/experimental/mesa_signals/__init__.py
```diff
@@ -0,0 +1,13 @@
+"""Functionality for Observables."""
+
+from .mesa_signal import All, Computable, Computed, HasObservables, Observable
+from .observable_collections import ObservableList
+
+__all__ = [
+    "Observable",
+    "ObservableList",
```
EwoutH, line 8: Would an `ObservableDict` (also) be useful, to be able to track which Agent the value originated from?

EwoutH (approved):
I didn't review every line of code, but I would like to see this move forward. The user API is clean enough, considering what's possible in Python, and it's all in the experimental space.

Thanks for working on this, curious where it will go!

mesa/experimental/mesa_signals/__init__.py
quaquel: An `ObservableDict` would indeed also be useful. Not sure what you mean with 

> to be able to track which Agent the value originated from

tpike3: I think I see what @EwoutH means here where it would be  something like `{agent.unique_id: <observables>}

mesa/experimental/mesa_signals/test.py
tpike3, line 1: Just curious why this is here? Did we maybe want to add this to the docs, so users have an example to follow? Maybe subset of the getting started and experimental how to??

tpike3 (commented):
This is awesome! I think as the develops this and the datacollection update will inextricably intertwined. 

I know @EwoutH has approved and I am good merging it, but is there anything else you wanted to complete before we merge?

mesa/experimental/mesa_signals/test.py
quaquel: This will be removed before merging, was just there during development. Good catch!

quaquel:
> Let us know what you want to complete before we merge.

I want to do a bit of further code cleaning (like the tests.py file) and improve code coverage a bit further. Beyond that, any feedback or ideas are welcome here or in #2281.

Corvince:
Not directly related to this PR, but I just watched [a video](https://www.youtube.com/watch?v=RHTP_NZcnl0) about [alien-signals](https://github.com/stackblitz/alien-signals/blob/master/src/system.ts), a new signals library for JavaScript. Its supposed to be very fast and will probably be used by Vue.js. Vue.js is interesting because it previously switched from a push-based system to a pull-based system and alien-signals is now push-pull-based. So maybe that is interesting for you @quaquel.

quaquel:
Thanks, I'll take a look. There might be useful ideas that we can borrow.

EwoutH:
Congratulations on merging this!

PR #2293, opened 2024-09-13
make typing behavior of AgentSet.get explicit

quaquel:
This is just a small update using typing.overload to make it explicit how arguments and the 2 possibly returns of AgentSet.get are related.


Corvince (approved):
Looks good! Somehow my language server Pylance doesn't pick this up correctly, but I think it should, I see no error in this code

PR #2294, opened 2024-09-13
Enforce google docstrings

quaquel:
This PR enforces Google docstrings in ruff, enables parsing of Google docstrings in building the documentation for readthedocs, and (the real work) fixes all documentation to be in line with the Google docstring. While going through all the docstrings, I have made various minor modifications to improve wording. 

I have used `noqa D103` in quite a few places (e.g., outdated parts of the code, various tests, solara parts that I don't fully know). This means that some parts don't have a docstring yet.


Corvince:
Awesome effort! That is something that has bugged me for some time and finally some effort to fix this!

I think instead of mixing and matching our own way of writing docstrings we should strive to fully adapt a style guide and I think the [numpy](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard) style guide would be most appropriate and I see this most used in the scientific community. 

Note that with a clear structure of the docstring format the visual representation on for example readthedocs can become a separate thought. Take for example psygnal: It uses numpy docstring format at least for the parameters ([link](https://github.com/pyapp-kit/psygnal/blob/824346778632dcdf48573d5b6b4b3fbd504434b6/src/psygnal/_signal.py#L252-L285)), but renders completely different on [readthedocs ](https://psygnal.readthedocs.io/en/latest/API/#psygnal.Signal)(with bulleted lists, links to type definitions, etc.)

quaquel:
> I think instead of mixing and matching our own way of writing docstrings we should strive to fully adapt a style guide and I think the [numpy](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard) style guide would be most appropriate and I see this most used in the scientific community.

Judging the current docstrings, it seems closests to the google docstring standard. So I sort of implicitly assumed that as the standard while working on MESA. I personally also prefer numpy. It would take a few hours to rewrite everything to numpy if all maintainers agree to use that.

EwoutH (approved):
Awesome, always love documentation improvements!

Also agree on the style guide. Especially to get things as bulleted lists correct consistently.

Corvince:
> It would take a few hours to rewrite everything to numpy if all maintainers agree to use that.

There are actually several docstring converters you can find on pypi. Not sure how good they work

EwoutH:
I would support using the NumPy [Style guide](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard) for docstring formatting and [numpydoc](https://github.com/numpy/numpydoc) for Sphinx / Readthedocs rendering. Especially if we can enforce it in CI (preferably pre-commit).

quaquel:
That makes 3 in favor of numpy. I am curious to get the perspective of @jackiekazil, @tpike3, and @rht. If there is a majority in favour, I'll happily work on moving everything over to numpy doc standard.

EwoutH:
I think we can enforce this with ruff, by adding this to our `pyproject.toml` config:

```pyproject.toml
[tool.ruff.lint.pydocstyle]
convention = "numpy"
```
See the [ruff faq](https://docs.astral.sh/ruff/faq/#does-ruff-support-numpy-or-google-style-docstrings).

<hr>

[Looking](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html#google-vs-numpy) at [some](https://betterprogramming.pub/3-different-docstring-formats-for-python-d27be81e0d68) [examples](https://stackoverflow.com/questions/3898572/what-are-the-most-common-python-docstring-formats), I might prefer the the Google docstring format over the NumPy one.

quaquel:
>  I might prefer the the Google docstring format over the NumPy one.

I guess this is a taste kind of thing, but what is about google that you like? What I like with numpy is that I find it easy to read because it cleanly seperates the name and dtype information from the explanation. Moreover, the underlying help me to quickly find the section I am looking for. Its drawback is that it requires more horizontal space.

EwoutH:
> what is about google that you like

I like the “cleanness” of the docstring, al the information is there, but nothing more boilerplate than needed. In a single screen I can view the docstring and code for most medium-complexity functions.

> What I like with numpy is that I find it easy to read because it cleanly seperates the name and dtype information from the explanation

Agreed, that are advantages of NumPy.

I guess for complicated classes and methods NumPy might be better suited, for more single / frequent the Google one might be more suitable.

Either way, since this is a preference thing (I think), we might just want to go with the majority one.

jackiekazil:
I do not have a strong opinion - so if others are passionate, about the opposite, I am okay going that direction. 

As long as it can be enforced, I lean towards Google because it is a little cleaner and easier to write. I think I have seen it more of Google style, but this is conjecture.

quaquel:
Ok, I'll go with Google for now because the current docstrings seem closest to that.

Corvince:
Yes the current format is closest to Google, i think we even decided that a long time ago. I also think it reads a bit easier in the code, but I would still prefer numpydoc, since it's so much more used in the scientific community.
Numpy, pandas, polars, matplotlib, altair, scipy, psygnals, networkx 

They all use numpydoc. I actually couldn't find a single package with Google (although I am sure in other contexts it is used much more)

Also, converting to another format seems like a very easy task for ChatGPT et al. 

But to me personally it is not important, so if we decide for google, then be it.

quaquel:
I am currently at 607 errors if I enforce Google.....  With numpydoc, which I also prefer, it is even worse. It seems easy to convert everything from one to the other, assuming the convention is used properly. So, for now, I make sure Google is implemented properly. We can then convert a few docstrings and have a discussion/vote about moving to numpydoc.

quaquel:
down to 423...

quaquel:
291 remaining, 265 of which are in the tests. The others are in the latest solara code. I'll have a closer look at the solara stuff tomorrow, but it likely that I can only provide the most basic docstrings here since I don't know a lot of the internal details.

Corvince:
I'll see what I can do about the solara stuff, most docstrings there are also way outdated since my refactoring!

Corvince:
> I'll see what I can do about the solara stuff, most docstrings there are also way outdated since my refactoring!

Since this PR here is basically unreviewable due to its size, I guess I will do that in a separate PR. So for the time being you just add some noqa D to solara, is this your understanding as well?

quaquel:
> Since this PR here is basically unreviewable due to its size, I guess I will do that in a separate PR. So for the time being you just add some noqa D to solara, is this your understanding as well?

Yes that makes the most sense and is what I have done now.

quaquel:
If there are no objections, I plan to merge this PR at the close of business today. All tests and ruff stuff passes. I have checked readthedocs also and this looks much cleaner now.

EwoutH (approved):
Skimmed through it, looks great in general!

docs/conf.py
```diff
@@ -1,3 +1,4 @@
+# noqa
```
rht, line 1: This noqa disables all the Ruff checks for this file, doesn't it?

maintenance/fetch_unlabeled_prs.py
```diff
@@ -66,7 +67,7 @@ def get_closed_pull_requests_since_latest_release(
     return pull_requests
 
 
-def main() -> None:
+def main() -> None:  # noqa
```
rht, line 70: Shouldn't the noqa here be specifically for docstring?

maintenance/fetch_unlabeled_prs.py
```diff
@@ -1,3 +1,4 @@
+# noqa
```
rht, line 1: Same here.

rht:
I have a class of comments related to the non-specific `noqa` comments. Shouldn't they be `noqa: D103`? But this is not blocking the merge of this PR, and can be done in a subsequent PR, since this PR is already too big.

rht (approved):

docs/conf.py
quaquel: as far as I know, no it only disables it for this first line which is a module level docstring.

To disable the entire file you need # ruff: noqa at the top of file (see [here](https://github.com/astral-sh/ruff/pull/2978))

quaquel:
> I have a class of comments related to the non-specific `noqa` comments. Shouldn't they be `noqa: D103`? But this is not blocking the merge of this PR, and can be done in a subsequent PR, since this PR is already too big.

I'll make a pass to make the noqa's specific. Will also be easier in the future to find them.

docs/conf.py
```diff
@@ -1,3 +1,4 @@
+# noqa D103
```
rht, line 1: I think the ':' is necessary, `# noqa: D103`, but haven't tested.

quaquel: Thanks for pointing this out. I have fixed it everywhere. This noqa stuff is quite new to me.

Corvince:
Over 1000 changed lines! Awesome work @quaquel , thanks for doing this

PR #2296, opened 2024-09-15
Make cell connections public and named

Corvince:
This PR changes the formerly private attribute `_connections` from the `Cell` class to be public and to be no longer a list, but a dictionary that includes a name for each connections. For Grids this defaults to the relative direction ( `(0, 1)`, `(-1, -1)`, etc.). For Network this is the node name of the "other" cell. 

This allows more powerful and direct agent movements, because you no longer need to query the connection to see where it leads or be dependent on the order in which directions were created. +

This PR also includes some type improvements

### DIFF
--- a/mesa/experimental/cell_space/cell.py
+++ b/mesa/experimental/cell_space/cell.py
@@ -69,16 +69,17 @@ class Cell:
         self.properties: dict[Coordinate, object] = {}
         self.random = random
 
-    def connect(self, other: Cell, name: Coordinate | None = None) -> None:
+    def connect(self, other: Cell, key: Coordinate | None = None) -> None:
         """Connects this cell to another cell.
 
         Args:
             other (Cell): other cell to connect to
+            key (Tuple[int, ...]): key for the connection. Should resemble a relative coordinate
 
         """
-        if name is None:
-            name = other.coordinate
-        self.connections[name] = other
+        if key is None:
+            key = other.coordinate
+        self.connections[key] = other
 
     def disconnect(self, other: Cell) -> None:
         """Disconnects this cell from another cell.
--- a/mesa/experimental/cell_space/cell.py
+++ b/mesa/experimental/cell_space/cell.py
@@ -76,7 +76,7 @@ class Cell:
         """
         if name is None:
             name = str(other.coordinate)
-        self.connections.update({name: other})
+        self.connections[name] = other
 
     def disconnect(self, other: Cell) -> None:
         """Disconnects this cell from another cell.
@@ -85,7 +85,9 @@ class Cell:
             other (Cell): other cell to remove from connections
 
         """
-        self.connections = {k: v for k, v in self.connections.items() if v != other}
+        keys_to_remove = [k for k, v in self.connections.items() if v == other]
+        for key in keys_to_remove:
+            del self.connections[key]
 
     def add_agent(self, agent: CellAgent) -> None:
         """Adds an agent to the cell.
--- a/mesa/experimental/cell_space/cell.py
+++ b/mesa/experimental/cell_space/cell.py
@@ -9,6 +9,7 @@ from typing import TYPE_CHECKING
 from mesa.experimental.cell_space.cell_collection import CellCollection
 
 if TYPE_CHECKING:
+    from mesa.agent import Agent
     from mesa.experimental.cell_space.cell_agent import CellAgent
 
 
@@ -26,7 +27,7 @@ class Cell:
 
     __slots__ = [
         "coordinate",
-        "_connections",
+        "connections",
         "agents",
         "capacity",
         "properties",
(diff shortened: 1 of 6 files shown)


mesa/experimental/cell_space/cell.py
```diff
@@ -56,20 +57,24 @@ def __init__(
         """
         super().__init__()
         self.coordinate = coordinate
-        self._connections: list[Cell] = []  # TODO: change to CellCollection?
-        self.agents = []  # TODO:: change to AgentSet or weakrefs? (neither is very performant, )
+        self.connections: dict[str, Cell] = {}
```
quaquel, line 60: From my understanding of the code, in case of orthogonal grids, the key is a tuple not a str? So is the typing here correct or should it be Any?

mesa/experimental/cell_space/voronoi.py
```diff
@@ -224,8 +224,8 @@ def _connect_cells(self) -> None:
 
         for point in self.triangulation.export_triangles():
             for i, j in combinations(point, 2):
-                self._cells[i].connect(self._cells[j])
-                self._cells[j].connect(self._cells[i])
+                self._cells[i].connect(self._cells[j], (i, j))
```
quaquel, line 227: Here you are using a tuple rather than a string as key for connecting

mesa/experimental/cell_space/grid.py
```diff
@@ -89,7 +89,7 @@ def _connect_single_cell_nd(self, cell: T, offsets: list[tuple[int, ...]]) -> No
             if self.torus:
                 n_coord = tuple(nc % d for nc, d in zip(n_coord, self.dimensions))
             if all(0 <= nc < d for nc, d in zip(n_coord, self.dimensions)):
-                cell.connect(self._cells[n_coord])
+                cell.connect(self._cells[n_coord], f"{d_coord}")
```
quaquel, line 92: here you indeed cast the coordinates from a tuple to a string.

quaquel:
I like this PR. I am just wondering whether using a string as key in connections is the way to go (and in voronoi it is not done correctly as far as I can tell). 

Any idea about the performance benchmarks? Slower inits make sense because the data structure is a bit more involved to built up. I fail to see why runs are slower?

Corvince:
The initial idea was to make the connection names completely arbitrary, so users could call them for example "up", "north", "top", etc., whatever seems most appropriate in the specific context. 

But probably it makes sense to have that abstraction somewhere else. Relative directions seem to make the most sense. That's unambiguous for grids, for network it kind of makes sense (the direction of node xy) and for voronoi I have to think about it. 

So I am probably going to change it to type tuple

EwoutH:
Interesting PR, coordinates seem more scalable then names.

Probably related:
- https://github.com/projectmesa/mesa/issues/1900

Something like a `Neighbours` class/object might be useful.

quaquel:
> But probably it makes sense to have that abstraction somewhere else. Relative directions seem to make the most sense. That's unambiguous for grids, for network it kind of makes sense (the direction of node xy) and for voronoi I have to think about it.

I guess Voronoi is a network with some interesting properties (e.g., you can lay it out in 2d without any crossing edges).

Corvince:
rebased and ready to be reviewed again

mesa/experimental/cell_space/cell.py
```diff
@@ -58,20 +61,24 @@ def __init__(
...
-        self.properties: dict[str, object] = {}
+        self.properties: dict[Coordinate, object] = {}
         self.random = random
 
-    def connect(self, other: Cell) -> None:
+    def connect(self, other: Cell, name: Coordinate | None = None) -> None:
         """Connects this cell to another cell.
 
         Args:
             other (Cell): other cell to connect to
```
quaquel, line 77: please complete the docstring. Name is not defined as one of the arguments

mesa/experimental/cell_space/cell.py
```diff
@@ -80,7 +87,9 @@ def disconnect(self, other: Cell) -> None:
             other (Cell): other cell to remove from connections
 
         """
-        self._connections.remove(other)
+        keys_to_remove = [k for k, v in self.connections.items() if v == other]
+        for key in keys_to_remove:
+            del self.connections[key]
```
quaquel, line 92: Can't this be simplified. If the name is just the coordinate of the cell, can't you do something like

self.connections.pop(other.coordinate)

Since removal is likely to be rare, not something to be overly worried about.

mesa/experimental/cell_space/cell.py
```diff
@@ -123,30 +132,34 @@ def __repr__(self):  # noqa
 
     # FIXME: Revisit caching strategy on methods
     @cache  # noqa: B019
-    def neighborhood(self, radius=1, include_center=False):
+    def neighborhood(self, radius: int = 1, include_center: bool = False):
```
quaquel, line 135: would it be possible to add a docstring. 

Also, and not perse necessary for this PR, but the name of this method has been tripping me up. The name suggests that it is an attribute, so perhaps at some point we should change this to something like `get_neighborhood`

mesa/experimental/cell_space/cell.py
Corvince: Interesting, because I rebased the PR and not merged main into it, my precommit hooks did not fire locally, so I didn't catch this. Next time I am going to merge again.

Also note that this is indeed an error and happily catched by our CI :) [pre-commit.ci](https://results.pre-commit.ci/run/github/24245737/1726651287.yuS9Ahq6QX602H5xdQnDvg)

mesa/experimental/cell_space/cell.py
Corvince: No, the name is not the absolute coordinate of other, but a relative coordinate of self, so unfortunately this doesn't work

mesa/experimental/cell_space/cell.py
Corvince: agreed, there is some revisiting to be done here. I think at one point the idea was that this basically is an attribute/property for the default case, which is just a CellCollection of the the connections. Maybe we could actually have something like 

```python
def get_neighborhood(self, radius: int = 1, include_center: bool = False):
     ...

@property
def neighborhood(self):
    return self.get_neighborhood()
``` 

So for the "normal" neighborhood you could do `cell.neighborhood` and for everything more complex `cell.get_neighborhood(...)`

quaquel: Yes, this makes good sense. The most common case is indeed getting the direct neighborhood and it would be good if that is just a property. Naming-wise, what about `direct_neighborhood` and `get_neighborhood`? It would be nice if the names were a bit more distinctive.

Corvince: Yes more distinctive names sounds good. I'll have to think about `direct_neighborhood` cause I think its a bit verbose. But lets leave that and adding a docstring to a different PR

PR #2298, opened 2024-09-18
Api docs

quaquel:
This PR reorganizes the API documentation. 

It is a follow-up of #2294. The current structure of the API overview in readthedocs has become more messy partly due to explicitly using and parsing Google docstrings. Moreover, the API docs reflect a slightly outdated view of the core classes in MESA. This PR reorganizes the API docs in such a way that it is just a convenient way to get to the docs of all relevant classes. Tutorials and other docs are available to highlight how to use MESA. The API docs should be a straightforward reference manual and be easy to navigate.


EwoutH (approved):
Diff looks good, new rendered docs also. Good to have the AgentSet in the API docs!

EwoutH:
With hindsight, I liked the "Base Classes" concept. It separate the classes we expect users to subclass from other ones. Maybe we can structure it like this:

image

PR #2299, opened 2024-09-18
SolaraViz Updates

Corvince:
Some more work on the SolaraViz Frontend. Highlights

1. Much better docstrings
2. Removed `make_text` function. It was used like `make_text(renderer)`, but now you can use `renderer` directly as a component
3. Merged Play/Pause button and disabled step button while model is running
4. Reorganised the Sidebar to look like this
image

### DIFF
--- a/mesa/visualization/__init__.py
+++ b/mesa/visualization/__init__.py
@@ -2,13 +2,12 @@
 
 from .components.altair import make_space_altair
 from .components.matplotlib import make_plot_measure, make_space_matplotlib
-from .solara_viz import JupyterViz, SolaraViz, make_text
+from .solara_viz import JupyterViz, SolaraViz
 from .UserParam import Slider
 
 __all__ = [
     "JupyterViz",
     "SolaraViz",
-    "make_text",
     "Slider",
     "make_space_altair",
     "make_space_matplotlib",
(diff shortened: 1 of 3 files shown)


EwoutH (approved):
Looks great!

Not directly for this PR, but are you still supporting a name change (we were considering `ModelExplorer` right)?

rht:
> Removed make_text function. It was used like make_text(renderer), but now you can use renderer directly as a component

Which objects are allowed/not allowed? `make_text` was specific to string objects. Would float work?

rht:
I didn't voice this concern in the past: but before this PR, the reset button was dangerously close to the play button, that people could accidentally fat-fingered it. But now it seems safer.

rht:
One of the tests in test_solara_viz.py failed.

Corvince:
> Which objects are allowed/not allowed? `make_text` was specific to string objects. Would float work?

Good question, I don't know. I would guess only strings work, but I will test this out

rht:
In this case, it's fine since I haven't provided any change requests, so the agreement is implicit, but in the future, I'd appreciate if the visualization PR gets merged when I explicitly approve, since @EwoutH hadn't dev-ed in the viz area yet.

Corvince:
Sounds fine to me!  But yes, since you already replied without any change requests I thought you are fine with the changes. And appreciated the heads-up to the failing test

PR #2303, opened 2024-09-19
Use python threading instead of solara thread in model controller

wang-boyu:
This seems to speed up model execution as a fix to what was reported in https://github.com/projectmesa/mesa/pull/2289#issuecomment-2356707386 and following comments. But I don't really know why.

### DIFF
--- a/mesa/visualization/solara_viz.py
+++ b/mesa/visualization/solara_viz.py
@@ -24,6 +24,7 @@ See the Visualization Tutorial and example models for more details.
 from __future__ import annotations
 
 import copy
+import threading
 import time
 from collections.abc import Callable
 from typing import TYPE_CHECKING, Literal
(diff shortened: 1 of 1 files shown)


rht:
I assume this solution also works on Colab?

wang-boyu:
Btw it's probably worth mentioning that it's good that play and pause buttons are merged into one (in PR https://github.com/projectmesa/mesa/pull/2299). Otherwise the solara app runs into trouble if users keep clicking on the play button.

Corvince:
Wow, this runs indeed a lot faster. But I was worried about the implications of using threads, since they potentially keep running forever in the background. `use_thread` does a lot to prevent this, but this seems to be quite slow for that reason. I now saw that its also deprecated and there is `use_task`, which also allows running in the python event loop. Its equally fast compared to this PR, but potentially saves us some headaches in the future. See #2304

Corvince:
Again, thanks a lot for showing whats possible. Its way faster now, would not have happened without your effort/noticing of bad performance

wang-boyu:
Great! Thanks @Corvince

PR #2304, opened 2024-09-20
Solara viz: use_task for non-threaded continous play

Corvince:
This is an alternative to https://github.com/projectmesa/mesa/pull/2303 

While investigating #2303 I realized that `use_thread` is deprecated and recommended to replace with `use_task`, which also allows not using a thread at all, but a coroutine. This allows this version to also work in environments where threading can be problematic (colab) or isn't available at all (Pyodide, WASM, etc.). It also avoids possible issues with threads running in the background forever. 

It also runs blazingly fast!

[screen-capture (1).webm](https://github.com/user-attachments/assets/857e4ea7-519a-491c-bf5e-52fc87a7c40c)

### DIFF
--- a/mesa/visualization/solara_viz.py
+++ b/mesa/visualization/solara_viz.py
@@ -23,8 +23,8 @@ See the Visualization Tutorial and example models for more details.
 
 from __future__ import annotations
 
+import asyncio
 import copy
-import time
 from collections.abc import Callable
 from typing import TYPE_CHECKING, Literal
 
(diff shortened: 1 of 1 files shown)


mesa/visualization/solara_viz.py
```diff
@@ -235,12 +235,12 @@ def save_initial_model():
 
     solara.use_effect(save_initial_model, [model.value])
 
-    def step():
+    async def step():
         while playing.value:
-            time.sleep(play_interval / 1000)
+            await asyncio.sleep(1 / 1000)
```
rht, line 240: Shouldn't `1` be `play_interval`?

Corvince: oops, of course. Thats why we have code reviews I guess :)

rht (approved):

EwoutH:
Awesome stuff!

As a reference, `use_task` is currently in the [Solara Lab](https://solara.dev/documentation/getting_started/what-is-lab), which contains parts of Solara that are slightly experimental. They are currently considering stabilizing it, but if any changes in the meantime, this is probably the issue to track:

- https://github.com/widgetti/solara/issues/784

Issue #201, opened 2016-04-11
How to import external data and define as model attributes at each step?


lovedavidsilva:
Thanks for developing this great ABM framework in Python. I am trying to transfer my model from NetLogo to Mesa. It is a model that simulates trading in stock market. I have historical stock prices and other information as input (model level attributes), and based on the information, calculate some derived model level attributes. Each step after the model level attributes are imported and calculated, then each agent start trading according to the model level attributes and its own preference.

To start with, I only tried to import model level information and computed one of the derived attributes. I used data collector and chart visualization to check if my codes worked. However, it seemed that I did not import the data successfully. So my question for now is that how to import external data and define as model attributes? I also upload my code here. Thanks.
[bac_model1.txt](https://github.com/projectmesa/mesa/files/212435/bac_model1.txt)

dmasad:
We don't have that functionality built in yet (though it would be a nice thing to have).

Looking at your code, it looks like you're missing a couple ``self``s in the agent step method. (In Python, every method and variable that belongs to an object needs to be prefaced with a ``self`). It should read:

```.python
def step(self, model):
    self.market_update(model)
    self.rsi_update(model)
    self.indi_trade()
```

I can't test your code without the data, but that might be enough to solve the problem.

Two more quick suggestions: If you're working with tabular data, you may also want to look at the [pandas](http://pandas.pydata.org/) library, which lets you load data into an R-style data frame. 

And for future code uploading, you may want to check out the [GitHub gist feature](https://gist.github.com/), which lets you quickly copy-paste files to upload them for sharing. 

Hope that helps!

lovedavidsilva:
Thanks David. I have added the prefix to each function now. However, the data still cannot be read.

Here is how I test the data import now:
At each step, 1)read a line from data and give the values to model attributes ("bac" for example), then 2)collect the value of the attribute using datacollector, finally 3)use chart module to visualize the data, thus checking if data was correctly imported.

It turns out that the output chart is still a line at 0. Since I went through the code so many times, it is hard to tell where is the problem. Is it possible for you to run the test and see what's wrong with it? I have uploaded several lines of the external [data](https://gist.github.com/lovedavidsilva/e68ed944092e0f124835c2eb9131d844).

Thanks!

dmasad:
One other bug is that you aren't actually adding agents to the schedule, so when you run the model, no agents are actually running. That section of the init method should be:

```.python
'''create agents'''
for i in range(self.num_agent):
[...]

PR #183, opened 2016-01-30
Add Conway's Game of Life simulation to the examples.

VictorNorman:
Fixes issue #182.

PR #2306, opened 2024-09-20
Deprecate Time module and all its Schedulers

EwoutH:
As concluded in https://github.com/projectmesa/mesa/discussions/2231, this PR deprecates the whole Time module and all its Schedulers. This change aims to provide more flexibility and control in agent activation and time management, inspired by the NetLogo framework.

### Changes
1. Deprecated the entire `mesa.time` module.
2. Added a module-level deprecation warning to `mesa.time`.
3. Updated the migration guide to include instructions for replacing each scheduler with AgentSet functionality.
4. Removed the `tests/test_time.py` file as it's no longer relevant.

### Migration Guide
The migration guide has been updated to include detailed instructions on how to replace each scheduler with AgentSet functionality. This includes examples for:

- BaseScheduler
- RandomActivation
- SimultaneousActivation
- StagedActivation
- RandomActivationByType

The guide also includes general notes on changes to related functionality, such as the automatic incrementing of `Model.steps` and the replacement of scheduler-specific methods with AgentSet methods.

For this PR, a rendered version of the migration guide can be [viewed here](https://mesa--2306.org.readthedocs.build/en/2306/migration_guide.html#time-and-schedulers).

### Benefits
As discussed extensively in [#1912](https://github.com/projectmesa/mesa/discussions/1912) and [#2231](https://github.com/projectmesa/mesa/discussions/2231), replacing the time module with AgentSet functionality has several benefits. A short summary:

1. **Flexibility**: Users can now create custom activation patterns without being restricted to predefined schedulers.
2. **Explicitness**: The behavior of agent activation is now explicitly defined in the model, improving readability and maintainability.
3. **Simplicity**: The codebase is simplified by removing the need for multiple scheduler classes.
4. **Consistency**: This change aligns Mesa more closely with other agent-based modeling frameworks like NetLogo.

### References
- Original discussion: [#1912](https://github.com/projectmesa/mesa/discussions/1912)
- Decision discussion: [#2231](https://github.com/projectmesa/mesa/discussions/2231)
- Example updates: [#183](https://github.com/projectmesa/mesa-examples/pull/183) and [#201](https://github.com/projectmesa/mesa-examples/pull/201)
- This PR: [#2306](https://github.com/projectmesa/mesa/pull/2306)
- Faster alternative for `.shuffle().do()`: [#2283](https://github.com/projectmesa/mesa/pull/2283)

### PR checklist
- [x] Write migration guide
- [x] Update this PR message
- [x] Find a faster alternative for `.shuffle().do()` (see [#2283](https://github.com/projectmesa/mesa/pull/2283))

### Next Steps
- Update all example models to use the new AgentSet functionality instead of schedulers.
- Review and update any remaining documentation that references the deprecated schedulers.


mesa/time.py
```diff
@@ -31,6 +31,15 @@
 from mesa.agent import Agent, AgentSet
 from mesa.model import Model
 
+# Throw module-level deprecation warning
+warnings.warn(
+    "The time module and all its Schedulers are deprecated and will be removed in a future version. "
```
rht, line 36: Does this mean 3.1 or 4.0?

quaquel: I would not put a timeline on its removal. 

Just deprecation is enough. We don't need to worry about maintaining and testing anymore, but it might be convenient to keep it around for a while to make it easier for people to migrate models.

EwoutH: Yeah good question, I'm not sure about that. I'm nudging towards 3.1 or 3.2.

Another option is leaving it as a warning for now, and flipping it to an Error at some point.

EwoutH:
Updated the migration guide, so this PR is ready for review.

A rendered version of the migration guide can be [viewed here](https://mesa--2306.org.readthedocs.build/en/2306/migration_guide.html#time-and-schedulers).

mesa/time.py
EwoutH: I wrote the comment before I read Jan's comment, but I agree with him. We don't support it anymore and it will get removed.

quaquel (approved):

mesa/time.py
```diff
@@ -31,6 +31,15 @@
 from mesa.agent import Agent, AgentSet
 from mesa.model import Model
 
+# Throw module-level deprecation warning
+warnings.warn(
```
rht, line 35: Since time.py is imported into `__init__.py`, this will throw a warning all the time `mesa` is imported. This will be too noisy.

quaquel: That's a good catch, and I agree. the fix is to remove it from the overall init then?

rht: One option is to add the warning to every classes `__init__`.

EwoutH: Would the warning in the Basescheduler work for now? Since all other schedulers use it.

EwoutH: Moved.

EwoutH: Doesn’t really seem to solve it.

quaquel: It should go into the init itself. A class is parsed on import so currently the error is thrown at that moment already. In the init the error gets thrown when initializing the class.

EwoutH: Done, and I checked that all other schedulers call the Baseclass init, which is the case.

EwoutH:
Moved the DepreciationWarning, and also added a nice visible warning to the docstring.

IMG_1466

Merging, will work on updating the examples next.

PR #2307, opened 2024-09-21
Update to CellCollection.select

quaquel:
This PR updates CellCollect.select to reflect the updated API of AgentSet.select. So, we switch to `at_most` instead of using `n`. Since `n` was not used anyway before and this is still experimental, I have just removed `n` altogether.


quaquel:
these performance benchmarks are strange. Neither one uses `CellCollection.select`.

EwoutH:
Nice to align them. Reminds me I wanted to add a `sample` method.

We merged some stuff, I think that's causing this PR branch to have more differences with main. Maybe rebase it.

EwoutH (approved):

PR #2308, opened 2024-09-21
Remove schedulers from benchmark models.

quaquel:
this removes the schedulers from the benchmark models (in one case it was in the init but not used in step anyway).


rht (approved):

EwoutH (approved):

EwoutH:
I don’t understand how BoltzmannWealth gets slower while doing less (being already on `shuffle_ do()`).

Probably just a fluke.

Can you rebase?

quaquel:
I think it has to do with the data collection in the Boltzman wealth model. I need to dig deeper to confirm.

quaquel:
My hunch is correct. The old boltzman wealth model contained a bug. It used a schedule but no agents were added to it. The datacollecter then collected data from the agents in the schedule (which were None). So, the new benchmark is correct.

Also, this means that the deprecation of schedules should result in an update to the datacollector. In fact, I think the entire agent_collector can be deprecated in favor of the new agent type stuff.

EwoutH:
Yeah, that's correct, you can also see it in https://github.com/projectmesa/mesa/pull/2300. These two are completely equivalent:

```Python
     agent_reporters={"energy": "energy"},
     agenttype_reporters={
         Agent: {"energy": "energy"}
     }
```We will redesign this whole thing any way. Let's leave the API alone for now.

Implementation wise, these parts could be updated as soon as the schedulers are actually removed.
https://github.com/projectmesa/mesa/blob/e6874ad60db388c8209555b281a120393a50f5d1/mesa/datacollection.py#L231-L237

In between we could add a warning. But I don't think that's easy or worth it.

This PR is good to go!

PR #2309, opened 2024-09-21
Have a dedicated neighborhood property and a get_neighborhood method on Cell

quaquel:
This picks up on a discussion in #2296. In the current code, we have method named `Cell.neighborhood`. This is an unfortunate name. So we agreed to split this into a property for the direct neighborhood (i.e. radius = 1) and a `get_neighborhood` method for larger radii. We had a bit of discussion on the names, but I decided to keep it simple: `neighborhood` and `get_neighborhood`. I have updated the docstrings to clearly explain when to use which one.


mesa/experimental/cell_space/cell.py
```diff
@@ -131,10 +131,34 @@ def is_full(self) -> bool:
...
 
+    @property
+    @cache
+    def neighborhood(self) -> CellCollection:
+        """Returns the direct neigborhood of the cell
+
+        This is equivalent to cell.get_neighborhood(radius=1)
+
+        """
+        return self.get_neighborhood(radius=1)
```
EwoutH, line 142: The `radius=1` is redundant right? I think it's more powerful if it uses the fully default get_neighborhood() method and is thus equivalent at calling that.
```suggestion
        return self.get_neighborhood()
```

EwoutH (commented):
I like the general design!

Any performance implication on caching? Mostly good probably, right?

quaquel:
just a quick observation about cellspaces. A random walk is now a one liner: `self.cell.neighborhood.select_random_cell()`.

EwoutH:
That’s really nice!

I would appreciate a deprecationwarning for `neighborhood()`

mesa/experimental/cell_space/cell.py
```diff
@@ -131,10 +131,34 @@ def is_full(self) -> bool:
     def __repr__(self):  # noqa
         return f"Cell({self.coordinate}, {self.agents})"
 
+    @property
+    @cache
```
Corvince, line 135: You can use cachedproperty here (I think from functools)

Corvince (approved):

quaquel:
> I would appreciate a deprecationwarning for neighborhood()

AFAIK that is not possible because the name clashes with the property.

EwoutH:
It looks like this does introduce some errors.

EwoutH:
These benchmark results are exactly opposite of https://github.com/projectmesa/mesa/pull/2308.

So seems to cancel out and are just based on an outdated branch or whatever.

EwoutH (approved):
Was starting to type something about API compatibility and deprecation warnings, but then remembered this is all in the experimental space.

Love the experimental space!

EwoutH:
Thanks, great to have it in!

Please make sure that pre-commit passes before merging. Always a bitch to clean up (done in 7037a32).

PR #2311, opened 2024-09-21
Readthedocs: Add release notes, move up API docs

EwoutH:
Readthedocs:
- Add `HISTORY.MD` release notes
- Move up API docs (very useful, should be a bit more visible)


EwoutH:
This is not as easy unfortunately.

PR #2312, opened 2024-09-21
LICENSE: Add full Apache 2.0 text

EwoutH:
Changes nothing to the license, but makes it more visible for external projects and in the GitHub interface.

Currently:
image
After this PR:
image

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,13 +1,201 @@
-Copyright 2023 Core Mesa Team and contributors
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
 
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
 
-    http://www.apache.org/licenses/LICENSE-2.0
+   1. Definitions.
 
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -1,13 +1,201 @@
-Copyright 2023 Core Mesa Team and contributors
```
rht, line 1: The original text is still necessary, except that it should be renamed to `NOTICE`, while having the `LICENSE` text with the new content in this PR. See Apache Spark's [NOTICE](https://github.com/apache/spark/blob/master/NOTICE) and [LICENSE](https://github.com/apache/spark/blob/master/LICENSE).

rht: cc: @wang-boyu for Mesa-Geo, @adamamer20 for mesa-frames to be consistent with the outcome of this PR.

EwoutH: In the official license that’s moved to line 189.

Do you think it should be at the top?

rht: [L189](https://github.com/apache/spark/blob/067f8f188eb22f9abe39eee0d70ad1ef73f4f644/LICENSE#L189) is in the appendix of the LICENSE text, and is just a guide on how to apply it.

jackiekazil:
The logic behind having it link to the text and not be the full text is that if there were updates to the Apache license -- we wouldn't have to keep up with those because the spirit would still be the same. I am not necessarily for or against the full license in the repo, but I wanted to share the reason for not including it.

LICENSE
jackiekazil: I am okay with NOTICE if LICENSE points to the NOTICE.

rht: According to https://www.apache.org/foundation/license-faq.html#Apply-My-Software, it should be the other way around.
> Include a copy of the Apache License, typically in a file called LICENSE, in your work, and consider also including a NOTICE file.

rht:
> The logic behind having it link to the text and not be the full text is that if there were updates to the Apache license

If there are updates to the license, the version will be bumped to be >2.0. If we want mutable reference to the Apache license, it needs to be Apache-2.0-or-later, which itself is very rarely used. Though in the case of GPL-3.0, it is quite common (GPL-3.0-or-later).

LICENSE
```diff
@@ -1,13 +1,201 @@
...
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
 
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
 
-    http://www.apache.org/licenses/LICENSE-2.0
```
rht, line 7: Also, today is 2024, so this needs to be `https://`, not `http://`, see https://www.apache.org/foundation/license-faq.html#Apply-My-Software.

EwoutH:
Interesting stuff, but more complicated than I like. I think we're fine for now.

If anyone wants to pick this up, of course go ahead.

rht:
Created #2318 so that my research here did not go to waste.

PR #2314, opened 2024-09-21
Migration guide: Intro, upgrade strategy, model.agents, headers

EwoutH:
For the migration guide:
- Write a short introduction
- Add an recommended upgrade strategy
- Add some notes on model.agents
- Make whitespaces and headers more consistent

### DIFF
--- a/docs/migration_guide.md
+++ b/docs/migration_guide.md
@@ -5,7 +5,7 @@ Non-breaking changes aren't included, for those see our [Release history](https:
 
 
 ## Mesa 3.0
-Mesa 3.0 introduces significant changes to core functionalities, such as agent and model initialization, scheduling, and visualization. The guide below outlines these changes and provides instructions for migrating your existing Mesa projects to version 3.0.
+Mesa 3.0 introduces significant changes to core functionalities, including agent and model initialization, scheduling, and visualization. The guide below outlines these changes and provides instructions for migrating your existing Mesa projects to version 3.0.
 
 _This guide is a work in progress. The development of it is tracked in [Issue #2233](https://github.com/projectmesa/mesa/issues/2233)._
 
--- a/docs/migration_guide.md
+++ b/docs/migration_guide.md
@@ -3,10 +3,21 @@ This guide contains breaking changes between major Mesa versions and how to reso
 
 Non-breaking changes aren't included, for those see our [Release history](https://github.com/projectmesa/mesa/releases).
 
+
 ## Mesa 3.0
-<!-- TODO small introduction-->
+Mesa 3.0 introduces significant changes to core functionalities, such as agent and model initialization, scheduling, and visualization. The guide below outlines these changes and provides instructions for migrating your existing Mesa projects to version 3.0.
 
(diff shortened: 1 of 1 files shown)


docs/migration_guide.md
```diff
@@ -3,11 +3,22 @@ This guide contains breaking changes between major Mesa versions and how to reso
 
 Non-breaking changes aren't included, for those see our [Release history](https://github.com/projectmesa/mesa/releases).
 
+
 ## Mesa 3.0
-<!-- TODO small introduction-->
+Mesa 3.0 introduces significant changes to core functionalities, such as agent and model initialization, scheduling, and visualization. The guide below outlines these changes and provides instructions for migrating your existing Mesa projects to version 3.0.
```
rht, line 8: "encompassing" is more apt than "such as". Otherwise, the sentence implies as if visualization was newly for the first time in Mesa 3.0.

docs/migration_guide.md
```diff
@@ -75,11 +87,27 @@ In Mesa 3.0, `unique_id` for agents is now automatically assigned, simplifying a
...
 ### AgentSet and `Model.agents`
-#### AgentSet
-<!-- TODO  -->
+In Mesa 3.0, the Model class internally manages agents using several data structures:
+
+- `self._agents`: A dictionary containing hard references to all agents, indexed by their `unique_id`.
+- `self._agents_by_type`: A dictionary of AgentSets, organizing agents by their type.
+- `self._all_agents`: An AgentSet containing all agents in the model.
+
+These internal structures are used to efficiently manage and access agents. Users should interact with agents through the public `model.agents` property, which returns the `self._all_agents` AgentSet.
```
rht, line 96: If I want to scrutinize an agent from `model.agents`, could be for data collection, but more often for debugging purpose, how would I do this efficiently? Currently, it is doable with `select(at_most=1)`, but this returns yet another `AgentSet`. In Polars, I can do `df.row(2)`.

Corvince: You can just do list-like `agentset[2]`

rht: I didn't see there is `__getitem__` documented in the public API: https://mesa.readthedocs.io/en/latest/apis/agent.html. This feature is documented in the code, but doesn't show up there.

rht: What if I want to access by a unique ID instead of the index? (iloc vs loc in pandas)

EwoutH: All good questions, but I don’t think they are relevant to this PR.

Maybe open an issue or discussion about this?

docs/migration_guide.md
EwoutH: Made it “including”

docs/migration_guide.md
rht: It's relevant in the sense that the user may have to do such thing in their existing code.

EwoutH: I don’t exactly understand what you want added / modified, so could you open an PR to add it?

PR #2315, opened 2024-09-21
intro_tutorial: Don't initialize agents with an unique_id

EwoutH:
This resolves about half of the warnings in the intro_tutorial.

The other half is the just replace scheduler, but that's harder to replace, if we also want to rewrite all the explanations.

Maybe we should completely rewrite it.

For now this PR just resolves some warnings.


EwoutH:
Build passed, the updated rendered tutorial can be [viewed here](https://mesa--2315.org.readthedocs.build/en/2315/tutorials/intro_tutorial.html#) (and compared to the [current one](https://mesa.readthedocs.io/en/latest/tutorials/intro_tutorial.html)). It halves the number of warnings from 14 to 7.

rht (approved):

PR #2318, opened 2024-09-22
Add NOTICE and full copy of Apache 2.0 license text

rht:
Supersedes #2312.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,13 +1,202 @@
-Copyright 2023 Core Mesa Team and contributors
 
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
 
-    http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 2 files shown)


jackiekazil:
RE: This ticket and #2312 -- Why are we doing this? What are we trying to fix?

rht:
This is how it is done according to https://www.apache.org/foundation/license-faq.html#Apply-My-Software
> Include a copy of the Apache License, typically in a file called LICENSE, in your work, and consider also including a NOTICE file.

> It is also valuable to tag each of your source-code files in case they become detached from the LICENSE file. To apply the Apache License to your source-code files, one approach is to attach the following notice to as a comment at the top of each file. Replace the copyright templates with your own identifying information:

EwoutH:
My initial intention was to adopt the "standard" way GitHub suggests it.

Changes nothing to the license, but makes it more visible for external projects and in the GitHub interface.

Currently:
image
After this PR:
image

Corvince:
If we use the Apache license we should do it how they suggest -> approved

Corvince (approved):

rht:
@jackiekazil can I merge the PR or do you have more reservations?

jackiekazil:
Thank you for checking! I am good. Please merge

EwoutH:
Awsome! GitHub seems to like it:

IMG_1477

PR #2319, opened 2024-09-23
experimental: Integrate PropertyLayers into cell space

EwoutH:
Integrate the PropertyLayers into the cell space. Initially only in the DiscreteSpace, and thereby the Grids.

Since we're in the experimental space, this PR is ready for review as-is.

In future PRs:
- Add a more comprehensive select function
https://github.com/projectmesa/mesa/blob/4c829896975f57ceffaf6b403e89fac25490b04d/mesa/space.py#L872
- Integrate PropertyLayer into CellCollection
  - Thinking about how to do this fast. For larger collections, keeping a mask is probably most performant, but for smaller collections, maybe just iterate though the cells.

### DIFF
--- a/mesa/experimental/cell_space/cell.py
+++ b/mesa/experimental/cell_space/cell.py
@@ -36,6 +36,7 @@ class Cell:
         "capacity",
         "properties",
         "random",
+        "_mesa_property_layers",
         "__dict__",
     ]
 
@@ -71,7 +72,7 @@ class Cell:
         self.capacity: int = capacity
         self.properties: dict[Coordinate, object] = {}
         self.random = random
-        self.property_layers: dict[str, PropertyLayer] = {}
+        self._mesa_property_layers: dict[str, PropertyLayer] = {}
 
     def connect(self, other: Cell, key: Coordinate | None = None) -> None:
         """Connects this cell to another cell.
@@ -198,16 +199,16 @@ class Cell:
     # PropertyLayer methods
     def get_property(self, property_name: str) -> Any:
         """Get the value of a property."""
-        return self.property_layers[property_name].data[self.coordinate]
+        return self._mesa_property_layers[property_name].data[self.coordinate]
 
     def set_property(self, property_name: str, value: Any):
         """Set the value of a property."""
-        self.property_layers[property_name].set_cell(self.coordinate, value)
+        self._mesa_property_layers[property_name].set_cell(self.coordinate, value)
 
     def modify_property(
         self, property_name: str, operation: Callable, value: Any = None
     ):
         """Modify the value of a property."""
-        self.property_layers[property_name].modify_cell(
+        self._mesa_property_layers[property_name].modify_cell(
             self.coordinate, operation, value
         )
--- a/mesa/experimental/cell_space/cell.py
+++ b/mesa/experimental/cell_space/cell.py
@@ -2,11 +2,13 @@
 
 from __future__ import annotations
 
+from collections.abc import Callable
 from functools import cache, cached_property
 from random import Random
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 from mesa.experimental.cell_space.cell_collection import CellCollection
+from mesa.space import PropertyLayer
 
 if TYPE_CHECKING:
     from mesa.agent import Agent
@@ -69,6 +71,7 @@ class Cell:
         self.capacity: int = capacity
         self.properties: dict[Coordinate, object] = {}
         self.random = random
+        self.property_layers: dict[str, PropertyLayer] = {}
 
     def connect(self, other: Cell, key: Coordinate | None = None) -> None:
         """Connects this cell to another cell.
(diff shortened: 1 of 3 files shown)


mesa/experimental/cell_space/cell.py
```diff
@@ -191,3 +194,30 @@ def _neighborhood(
             if not include_center:
                 neighborhood.pop(self, None)
             return neighborhood
+
+    # PropertyLayer methods
+    def add_property_layer(self, property_layer: PropertyLayer):
```
quaquel, line 199: do add and remove an entire layer belong in a cell?

In my view a cell would just do a get and a set on its coordinates in a specified layer.

mesa/experimental/cell_space/discrete_space.py
```diff
@@ -73,3 +83,38 @@ def empties(self) -> CellCollection[T]:
...
+        del self.property_layers[property_name]
+        if remove_from_cells:
+            for cell in self._cells.values():
+                cell.remove_property_layer(property_name)
+
+    def set_property(
+        self, property_name: str, value, condition: Callable[[T], bool] | None = None
+    ):
+        """Set the value of a property for all cells in the grid."""
+        self.property_layers[property_name].set_cells(value, condition)
```
quaquel, line 110: please (for all) specify the arguments as well as passing a docstring.

in pycharm, you can specify which docstring standard you want to use to make this easy (https://www.jetbrains.com/help/pycharm/settings-tools-python-integrated-tools.html)

quaquel:
It's remarkably straightforward. Good to see.

mesa/experimental/cell_space/discrete_space.py
EwoutH: First of all, very useful tool. And you know I love good tooling, so thanks for linking to it.

> (for all) specify the arguments

I don't know if I necessarily agree with this. Every function, method and class should have at least the oneliner and type-hinted input and output. But I also think that sometimes can be enough.

Comments/docstring are there to:
- Explain conceptual or higher level workings
- Explain unintuitive or complex code.

If there are no conceptual workings and the code is simple and intuitive, it's readable on its own. So I don't think it needs docstring in those cases.

I will reflect on these specific cases.

mesa/experimental/cell_space/cell.py
EwoutH: Having direct access to the PropertyLayer is what makes it performant. You don't want to copy all the values everything. So normally you directly access the layer at your coordinate.

mesa/experimental/cell_space/discrete_space.py
EwoutH: Added quite some docstring

mesa/experimental/cell_space/cell.py
EwoutH: Removed those two functions, you can directly access the dicts.

EwoutH:
If there are any other remarks please let me know.

Corvince:
> It's remarkably straightforward. Good to see.

I can second that. When I first skimmed this PR I thought its not ready yet, because its not a lot of code. Would have expected more work to be needed. Great to see! One thing I need more time to think about is how the individual cell interacts with the layer. Right now a reference to the whole Property layer ist passed. But Cells also have a "properties" attribute. I would have imagined that inside this properties dictionary there is just a "view" to the propertylayers individual coordinate. Does that make sense?

quaquel:
One last small request: 4 lines are currently not covered by the tests. Can you make sure we have 100% coverage on this? Its a small thing to fix now, but it adds up towards the future.

EwoutH:
Agreed, feel free to add them, otherwise I will pick it up later this week.

quaquel (approved):

EwoutH:
Updated the tests, and removed the possibility to initialize grids with propertylayers, because that doesn't give you the control if you want those layers added to the cells or not.

I'm struggeling a bit with how - on a conceptual level - we want Cell `poperties` and `property_layers` to interact / coexist / complement each other. Functionally it's clear, `poperties` are for individual cell properties (also other than numbers) and `property_layers` are for high-performance properties that are shared though a whole space.

Input is welcome.

quaquel:
> I'm struggeling a bit with how - on a conceptual level - we want Cell poperties and property_layers to interact / coexist / complement each other. Functionally it's clear, poperties are for individual cell properties (also other than numbers) and property_layers are for high-performance properties that are shared though a whole space.

I don't understand the point, but also in the code, I am a bit lost. In my understanding, property layers are define at the space level and individual cells can only access their values in those same property layers. However it seems cell now has its own property layer dict?

mesa/experimental/cell_space/cell.py
```diff
@@ -69,6 +71,7 @@ def __init__(
         self.capacity: int = capacity
         self.properties: dict[Coordinate, object] = {}
         self.random = random
+        self.property_layers: dict[str, PropertyLayer] = {}
```
quaquel, line 74: sorry I don't understanand why this is here? should this not be linked to the property layers defined at the grid level?

EwoutH: It also is, but it might be useful to be able to access your value in a PropertyLayer directly from the cell.

quaquel: I understand that and agree with it. I now also looked at `add_property_layer` and start to understand what is going on. 

1. I suggest making this `self._property_layers` so it's not declared as part of the public API. My concern is that one might add a layer via cell directly which would then not exist on the grid. 
2. I think overall, a more elegant solution is possible here (but that can be a separate PR). My idea would be to add on the fly Descriptors to cells for each property layer. This would make it possible to de `cell.property_layer_name`, which would return the value for this cell in the named property layer. Another benefit would be that we don't have code by default in cell that is only functional in OrthogonalGrids.

Corvince: On 2) my high-level idea would be that we use the existing property dictionary of the cell and add to it an entry with the name of the property layer, so `cell.properties[property_layer_name]`. *getting* the value should be easy, but I don't know how that would work for setting the value. Any ideas?

quaquel: I can't think of an immediate solution other than to subclass dict and have a dedicated PropertiesDict for all this.

EwoutH: Was thinking about that. It might be great as an API, but can also create naming conflicts.

Unfortunately, I'm really making a sprint on my thesis now, so not much conceptual-thought capacity to work on this.

quaquel: With the one minor change (to a private attribute), I think we can merge this for now. I might have some time over the weekend to play around with more descriptor fun 😄 and see what I can do.

EwoutH: Go ahead in making the change and merging!

PR #2324, opened 2024-09-24
Readthedocs: Add version switch and update URL

EwoutH:
This does two things:
- Add version switch to choose between the stable and latest docs
- And update the URLs to the shorter `mesa.readthedocs.io/stable` (instead of `mesa.readthedocs.io/en/stable`)

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,5 +1,5 @@
 # Read the Docs configuration file
-# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+# See https://docs.readthedocs.io/stable/config-file/v2.html for details
 
 # Required
 version: 2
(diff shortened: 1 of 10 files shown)


EwoutH:
Only way to test this is to merge and deploy on the main branch, since the switch URL points to that. Let's go!

Corvince:
Wait, did you just nuke all existing links out there? 😅 If you want to do that they must at least redirect. If that's not possible please revert this :)

EwoutH:
No, I would never do that! I think I figured it out, can you check if it works on your side? https://mesa.readthedocs.io/en/latest/tutorials/visualization_tutorial.html

Corvince:
Haha, good one, I fell for that...

You just got me worried because I just tested mesa.readthedocs.io/en/stable and that gives a 404. But the other links seem to redirect correctly.

EwoutH:
Couldn't not do it ;)

That one now also works, I was redirecting `/en/stable/*`, now `/en/stable*`

PR #2325, opened 2024-09-25
CI: Use Python 3.13

EwoutH:
Use Python 3.13 in all CI jobs. It’s faster and gives better error messages.


EwoutH:
This is exactly why you have these kind of PRs before a new stable Python release:
- https://github.com/widgetti/solara/issues/797

Benchmarks numbers look remarkably similar. Where Python 3.11 and 3.12 gave a noticeable speedup, that doesn't look to be the case with 3.13.

rht (approved):

PR #2326, opened 2024-09-25
Migration guide: Add note about removing Agents from the model

EwoutH:
Add note about removing Agents from the model to the Migration guide.

It now reads:

> 5. Instead of `self.schedule.add()` and `self.schedule.remove()`, agents are now automatically added to and removed from the model's AgentSet (`model.agents`) when they are created or removed.
>    - You still need to remove the Agent itself. Use `Agent.remove()` for that. Often you would:
>      - Replace `self.schedule.remove(agent)` with `agent.remove()` (in the Model)
>      - Replace `self.model.schedule.remove(self)` with `self.remove()` (in the Agent)

### DIFF
--- a/docs/migration_guide.md
+++ b/docs/migration_guide.md
@@ -221,10 +221,10 @@ self.agents_by_type[AgentType].shuffle_do("step")
 2. If you were using `self.schedule.agents`, replace it with `self.agents`.
 3. If you were using `self.schedule.get_agent_count()`, replace it with `len(self.agents)`.
 4. If you were using `self.schedule.agents_by_type`, replace it with `self.agents_by_type`.
-5. Instead of `self.schedule.add()` and `self.schedule.remove()`, agents are now automatically added to and removed from the model's AgentSet (`model.agents`) when they are created or removed.
-   - You still need to remove the Agent itself. Use `Agent.remove()` for that. Often you would:
-     - Replace `self.schedule.remove(agent)` with `agent.remove()` (in the Model)
-     - Replace `self.model.schedule.remove(self)` with `self.remove()` (in the Agent)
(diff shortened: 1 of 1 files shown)


docs/migration_guide.md
```diff
@@ -221,7 +221,10 @@ self.agents_by_type[AgentType].shuffle_do("step")
 2. If you were using `self.schedule.agents`, replace it with `self.agents`.
 3. If you were using `self.schedule.get_agent_count()`, replace it with `len(self.agents)`.
 4. If you were using `self.schedule.agents_by_type`, replace it with `self.agents_by_type`.
-5. Instead of `self.schedule.add()` and `self.schedule.remove()`, agents are now automatically added to and removed from the model's AgentSet when they are created or removed.
+5. Instead of `self.schedule.add()` and `self.schedule.remove()`, agents are now automatically added to and removed from the model's AgentSet (`model.agents`) when they are created or removed.
```
rht, line 224: This first sentence about "automatically removed" is misleading. Better just delete it.

rht: That's clearer.

EwoutH: Done!

Corvince:
Not related to the PR itself, but I find the semantics here a bit confusing. Why should I have to call agent.remove if I want to remove it from model.agents? And why can't I do agent.add()? I think we should actually remove `agent.remove` because of its unclear semantics. Thus, removing an agent can mean two things

1. `del agent`, in which case it is automatically removed from `model.agents` (right?)
2. `model.deregister_agent(agent)`. In case I want to remove it from the model, but keep it around, maybe to add it back later via `model.register_agent`. This would mean we should keep those methods public (https://github.com/projectmesa/mesa/pull/2251#issuecomment-2371043880)

quaquel:
> del agent, in which case it is automatically removed from model.agents (right?)

I don't think that will work. Del removes a reference from a namespace nothing more and there AFAIK (and just checked) no magic method that can used to change the behavior of del. `__del__` and `__delete__` are not tied to del and do subtle different things.

Issue #1221, opened 2022-03-30
bug: batch_run assumes model datacollector is called datacollector


tpike3:
**Describe the bug**
batch_run assumes user will call `DataCollector` `self.datacollector`. This is not annotated in the documentation, nor is it necessarily a requirement  

**Expected behavior**
users can name their model `DataCollector` whatever they want (e.g. `self.dc`)

**To Reproduce**
create model and name `DataCollector` anything else and then execute model with batch_run and try to retrieve the data. 

**Additional context**
Bug identified via students from the University of Mary Washington. They also had some very positive comments from their Professor
"They keep saying "wow, that's so easy! That makes it so fun!" Thanks to you and your crew for all you do."

rht:
https://github.com/projectmesa/mesa/pull/1213#issuecomment-1080158845= will solve this issue. Although I personally prefer https://github.com/projectmesa/mesa/pull/1213#issuecomment-1080517695=.

> "They keep saying "wow, that's so easy! That makes it so fun!" Thanks to you and your crew for all you do."

This needs to be judged relative to NetLogo, because I think it's easy to do equivalent stuff in NetLogo as well.

PR #2327, opened 2024-09-25
Deprecate `initialize_data_collector`

EwoutH:
Deprecate the Model's `initialize_data_collector`, which was added in #1287 to solve #1221.

Back when we had a schedulers it was logical, but now you can't forget to add Agents to a scheduler since that happens automatically.

The other reason was the `batch_run` assuming a Model attribute called `datacollector` to be present. An error in `batch_run` itself is far more robust to enforce that behavior, so that's added..

It also removes the behavior of directly collecting after init. Now you can do it when you want. It makes using the DataCollector explicit.

The migration guide was updated. Basically, replace:
```python
self.initialize_data_collector(...)
```

With:
```python
self.datacollector = DataCollector(...)
```

### DIFF
--- a/docs/migration_guide.md
+++ b/docs/migration_guide.md
@@ -309,3 +309,5 @@ With:
 ```python
 self.datacollector = DataCollector(...)
 ```
+
+- Ref: [PR #2327](https://github.com/projectmesa/mesa/pull/2327), Mesa-examples [PR #208](https://github.com/projectmesa/mesa-examples/pull/208))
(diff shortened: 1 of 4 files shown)


rht:
> The other reason was the batch_run assuming a Model attribute called datacollector to be present. An error in batch_run itself is far more robust to enforce that behavior, so that's added..

Another reason is that there could be multiple data collector objects. As of today, I don't think there are any reasonable use cases that necessitate multiple data collectors (each with own scheduler/agent containers) anymore.

I'm fine with reducing the API surface, i.e. deleting this method.

PR #2332, opened 2024-09-28
viz: stop running and disable buttons when model.running is False

wang-boyu:
This PR stops calling `model.step()` from the frontend when `model.running` is `False`. Play and step buttons are also disabled.

### DIFF
--- a/mesa/visualization/solara_viz.py
+++ b/mesa/visualization/solara_viz.py
@@ -225,6 +225,7 @@ def ModelController(model: solara.Reactive[Model], play_interval=100):
         play_interval (int, optional): Interval for playing the model steps in milliseconds.
     """
     playing = solara.use_reactive(False)
+    running = solara.use_reactive(True)
     original_model = solara.use_reactive(None)
 
     def save_initial_model():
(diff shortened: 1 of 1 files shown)


Corvince (approved):

Corvince:
Looks good to me!

EwoutH:
How/when should/can I use this?

wang-boyu:
Oh sorry I should've provided more context.

The current solara viz keeps calling model.step() and updating plots, even when the model stops itself, in models such as the Schelling example:

```python
def step(self):
    # model logic
    ...
    if self.happy == len(self.agents):
        self.running = False
```

This PR stops the frontend when model.running becomes False, and disables the play and step buttons.

EwoutH (approved):
Awesome, sounds like an useful feature!

EwoutH:
@quaquel want do you think about the whole `self.running` concept?

quaquel:
> @quaquel want do you think about the whole self.running concept?

A currently underdeveloped feature that needs to be implemented better. It is used with `self.run_model`, which is also strange and creates a typing mess. I. have looked at it a few times without coming up with a clean solution. My rough idea is to have some `stopping_condition` callable, which can be parameterized by arguments to `self.run_model`, and when the stopping condition becomes true, `self.running` is automatically set to false. By default this `stopping_condition` might simply be tied to `steps` and be parameterized with `n` for the number of steps. If this `stopping_condition` is a method on `Model`, you would have a meaningful default while making it user-extendable,

rht:
Minor question:

I forgot about the old behavior of SolaraViz. Which one is it:
1.has it been this way but changed in #2299, then this PR fixes it
2. the bug has always been there even before #2299

This only affects whether it should be communicated as a bug fix or a new feature in the release note.

rht (approved):

wang-boyu:
It's probably included in https://github.com/projectmesa/mesa/pull/2263/? I didn't try it but there were some usages of `model.value.running`. It was replaced with `playing.value` in https://github.com/projectmesa/mesa/pull/2282.

wang-boyu:
Merged! Pls feel free to change the label if anyone thinks there's a better one.

rht:
It's a bug fix. The old JupyterViz does check the model.running: https://github.com/projectmesa/mesa/blob/7addec3e008d7c5d66aba131c50b55cd31ba18ab/mesa/experimental/jupyter_viz.py#L231-L234.

PR #2337, opened 2024-10-03
CI: Add GitHub Action to append installation instructions to PRs

EwoutH:
This PR adds a simple workflow that adds the installation instructions to each PR. It helps contributors, testers and reviews quickly get an installation command to test a PR, especially if they want to install it to use with other packages such as Mesa examples.

For example, for this PR the following message would be added:

You can install this PR branch with:
```
pip install -U -e git+https://github.com/EwoutH/mesa@pr_message#egg=mesa
```

### DIFF
--- a/.github/workflows/pr_branch_message.yml
+++ b/.github/workflows/pr_branch_message.yml
@@ -1,8 +1,7 @@
-name: PR install note
-
+name: PR Install Note
 on:
   pull_request:
-    types: [opened]
+    types: [opened, reopened, ready_for_review]
 
 jobs:
   add-install-note:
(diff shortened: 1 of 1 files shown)


rht:
I'd say that a new message instead of an amend is less intrusive (the bot is editing people's PR message body without their consent here).

rht:
> How is this different?

Contributor can opt to make their PR commits non-editable. Does this auto-edit does the edit bypassing this condition?

At the very least, needs to make it clear with a header that it is edited by a GH Actions bot.

EwoutH:
Doesn’t feel worth it pushing this, maybe too niche.

PR #2344, opened 2024-10-10
use GridDraggable instead of Column in SolaraViz

wang-boyu:
This PR brings back the draggable and resizable Solara layout that was in the old Solara viz (now experimental).

### DIFF
--- a/mesa/visualization/solara_viz.py
+++ b/mesa/visualization/solara_viz.py
@@ -30,64 +30,13 @@ from typing import TYPE_CHECKING, Literal
 
 import reacton.core
 import solara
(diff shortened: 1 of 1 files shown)


EwoutH:
If I understand this correctly, I really like this conceptually!

I will test it tomorrow. Thanks for working on this, the visualization deserves more love!

EwoutH (approved):
Tested, looks goods and works well. Thanks!

One small remark: It looks like it only rescales on width, and not on height. Maybe we want to modify it that it also follows the maximum height:

image

But this isn't blocking, and can be done in another PR.

PR #2346, opened 2024-10-10
update legend, xlabel & format of matplotlib plots

wang-boyu:
- change matplotlib plot format from "png" to "svg"
- automatically set legend location through `ax.legend(loc="best")`
- add text "Step" to x axis label

Before:

before

After:

after

### DIFF
--- a/mesa/visualization/components/matplotlib.py
+++ b/mesa/visualization/components/matplotlib.py
@@ -63,7 +63,9 @@ def SpaceMatplotlib(
     elif space is None and propertylayer_portrayal:
         draw_property_layers(space_ax, space, propertylayer_portrayal, model)
 
-    solara.FigureMatplotlib(space_fig, format="png", dependencies=dependencies)
+    solara.FigureMatplotlib(
+        space_fig, format="png", bbox_inches="tight", dependencies=dependencies
+    )
 
 
 def draw_property_layers(ax, space, propertylayer_portrayal, model):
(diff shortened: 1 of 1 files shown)


EwoutH (approved):
All obviously good changes!

One question, are there any implications for performance? Is exporting to SVG slower (or faster) than to PNG?

Otherwise, go ahead and merge.

mesa/visualization/components/matplotlib.py
```diff
@@ -63,7 +63,9 @@ def SpaceMatplotlib(
     elif space is None and propertylayer_portrayal:
         draw_property_layers(space_ax, space, propertylayer_portrayal, model)
 
-    solara.FigureMatplotlib(space_fig, format="png", dependencies=dependencies)
+    solara.FigureMatplotlib(
+        space_fig, format="svg", bbox_inches="tight", dependencies=dependencies
```
rht, line 67: Using png is intentional: https://github.com/projectmesa/mesa/issues/1806#issue-1899325310. At the very least, a comment should be added to say it is for performance reason.

mesa/visualization/components/matplotlib.py
```diff
@@ -327,11 +329,14 @@ def PlotMatplotlib(model, measure, dependencies: list[any] | None = None):
     elif isinstance(measure, dict):
         for m, color in measure.items():
             ax.plot(df.loc[:, m], label=m, color=color)
-        fig.legend()
+        ax.legend(loc="best")
     elif isinstance(measure, list | tuple):
         for m in measure:
             ax.plot(df.loc[:, m], label=m)
-        fig.legend()
+        ax.legend(loc="best")
```
rht, line 336: I'm not sure if "best" is already the default when not specified, but a problem is that the legend location might change over time depending on the lines being plotted.

wang-boyu: Yes it does change over time depending on how lines are plotted. If we want the legend to be at a fixed location, I tried to make it at top right corner outside the main axis. Then the legend will never intersect with any plot. However, the issue is that the plotting area becomes smaller.

EwoutH: I think it's fine if it moves, that just means it's more out of the way again.

mesa/visualization/components/matplotlib.py
wang-boyu: Yes svg is indeed very low as compare to png: [solara-perf.ipynb](https://github.com/wang-boyu/mesa-examples/blob/solara-perf/examples/schelling_experimental/solara-perf.ipynb) (863 ms for png vs. 3.89 s  for svg, for a 200x200 schelling model). I also tried increasing png dpi from default 100 to 300, and it becomes slower with no noticeable improvement in terms of plot quality. I have changed back to png format in this PR.

rht (approved):

PR #2347, opened 2024-10-12
Update migration_guide.md

quaquel:
This is a minor expansion to using `super` in `Model`. I explicate how to handle the seed keyword argument, and briefly state something about how to handle arguments and keyword arguments via super.


docs/migration_guide.md
```diff
@@ -42,15 +42,16 @@ The `mesa.flat` namespace is removed. Use the full namespace for your imports.
...
+In Mesa 3.0, it is now mandatory to call `super().__init__()` when initializing your model class. This ensures that all necessary Mesa model variables are correctly set up and agents are properly added to the model. The convention in Python is to extract the arguments and keyword arguments you need in your `__init__` and pass on the rest via `super`. To control the random number generation, `mesa.Model` relies on the `seed` keyword argument.
 
 Make sure all your model classes explicitly call `super().__init__()` in their `__init__` method:
 
 ```python
 class MyModel(mesa.Model):
-    def __init__(self, *args, **kwargs):
-        super().__init__()  # This is now required!
+    def __init__(self, some_arg_I_need, *args, seed=None, my_init_kwarg=True, **kwargs):
+        super().__init__(*args, seed=seed, **kwargs)  # This is now required!
```
EwoutH, line 52: Is the `*args, **kwargs` here essential for the Model class?

quaquel: not for the class itself but for proper use of super. See my remark above the code on this.

EwoutH:
I found the `super().__init__(*args, seed=seed, **kwargs)` so ugly I started to find some creative ways around it.

One problem, why we have to call `super().__init__()` at all, is that we advise users to override the Model `__init__` method, and then fix that by calling `super().__init__()`.

Maybe that doesn't have to be that way. What is if we recommend users to put their setup code into a _differently named_ method, so the `Model.__init__` does get overridden?

In our model, we now call a new `setup` (or `initialize`) function in the init by default:

```Python
class Model:
    def __init__(self, seed: Optional[float] = None, **kwargs: Any) -> None:
        self.running = True
        self.steps: int = 0
        self._seed = seed if seed is not None else random.random()
        self.random = random.Random(self._seed)
        self._setup_agent_registration()
        
        # Remove 'seed' from kwargs before passing to setup
        kwargs.pop('seed', None)
        # Call user-defined setup
        self.setup(**kwargs)

    def setup(self, **kwargs: Any) -> None:
        """
        User-defined setup method.
        Subclasses should override this method instead of __init__.
        """
        pass
```
We request users to override `setup` instead of `__init__`. Their model then looks something like this:
```Python
class MyModel(Model):
    def setup(self, param1: int, param2: str):
        self.param1 = param1
        self.param2 = param2
        # Additional setup logic...

    def step(self):
        # Model step logic...
        pass

# Usage
model = MyModel(seed=42, param1=10, param2="example")
```
Which I find to look extremely clean.

As an additional advantage, we can now more easily modify the order of Model arguments, rename ones or add new ones without breaking user models, because we can handle all that (and throw warnings if necessary, etc.

We never have to be afraid an user doesn't call the `super().__init__()` incorrectly, because we simply don't allow overriding it anymore.

This is the most breaking change of changes, but I think it allows for _so much_ flexibility in the future that it might be worth it.

quaquel:
I am not against shifting to having a model setup method, but that probably deserves its own discussion rather than discuss it here in this PR. This PR really is about making sure seed is handled correctly in the migration guide.

quaquel:
@EwoutH in line with our conversation, I simplified the migration guide and not bother with the *args and *kwargs and just explicitly only pass `seed` to super. This is what 95% of users needs. The *args, **kwargs stuff is only relevant when having a class hierarchy that includes other super classes next to `Model`. Users that do this should be well equipped to understand how to make that work themselves.

EwoutH (approved):
Thanks for cleaning it up!

PR #2348, opened 2024-10-12
Add convenience properties for grid width and height

quaquel:
90% of all use cases of grid will be 2D so it is convenient to have easy access to the width and the height of the grid. 

While porting the examples over to the experimental grid spaces, I frequently need to write `self.grid.dimensions[0]` etc. This makes it a bit more readable because I can just do `self.grid.width` etc.


EwoutH (approved):
Nice, we discussed the API before, it clean and convenient.

Thanks!

Issue #2343, opened 2024-10-09
Fix `__all__` imports in `__init__.py`s


EwoutH:
On a few places in Mesa uses `__all__` imports in different `__init__.py`s files. 

https://github.com/projectmesa/mesa/blob/2cdbcb6289c7ed0a7b086768bf5930f05c30568b/mesa/__init__.py#L8-L23
https://github.com/projectmesa/mesa/blob/2cdbcb6289c7ed0a7b086768bf5930f05c30568b/mesa/experimental/__init__.py#L3-L7

This allows easier importing of modules, but does import modules that are not even necessary or used at all. This created some problems after #2265, because for some of those imports require dependencies that are not included by default.

Follow-up on #2265, fixes https://github.com/projectmesa/mesa-examples/pull/198.

invcble:
@EwoutH is this still open? I would like to work on it as a good starting point for contribution.

DashamiJituri:
Hi, I’ve opened a PR addressing this issue by replacing eager imports in `mesa/__init__.py` with lazy loading using `__getattr__`.

This avoids importing optional dependencies at package import time while preserving the existing API.

PR: https://github.com/mesa/mesa/pull/3560

Would appreciate any feedback!

EwoutH:
Let's see how [PEP 810](https://peps.python.org/pep-0810/) (Explicit lazy imports) develops. It might offer some new best-practices starting with Python 3.15.

In that light I can imagine we're jumping early to 3.15, if it offers significant benefits (somebody could already research if this is the case).

By the way, we already require Python 3.12+, so I believe we could fully utilize [PEP 690](https://peps.python.org/pep-0690/). I don't know if we do that already.

quaquel:
690 was rejected, but I agree that mimicking scipy seems sensible. Note, however, that the import times for mesa are very small compared to e.g., scipy. 

I am not in favour of moving quickly to 3.15. It will render Mesa unusable on many HPC systems for several years. HPC systems generally rely on conda. To my knowledge, the default conda distribution currently uses 3.13. Most HPC systems update their libraries once or twice a year and prefer very stable distributions. So, many are likely to still be on 3.12 or even 3.10.

dietcoke-17:
Consolidating the history on this while working in discrete_space, since the pattern keeps recurring:

- #2357 (Oct 2024): quick fix — wrapped the eager Solara import in `mesa/experimental/__init__.py` in a try/except. Called out at the time as a stopgap for this issue.
[...]

Issue #963, opened 2020-11-28
Discuss: Moving examples to examples repo


jackiekazil:
A long long long time ago we talked about moving examples over to the examples repo. 
I feel like we should go ahead and do that. The more the examples grow the bigger the main project will get. 

Examples repo: https://github.com/projectmesa/mesa-examples/ 

People have been submitting issues and PRs against this repo.
So either we kill or officially move. I feel like it makes sense to kill and I would like to prioritize doing this. 

What moving looks like... 
- Remove from this repo
- Overwrite all date here: https://github.com/projectmesa/mesa-examples/
- Update Readme for mesa-examples
- Review and update any references to examples in the docs of this repo

Thoughts?

rht:
Previous discussion at #134.

rht:
Another alternative is to create a separate `examples` repo, but link it as a submodule inside `projectmesa/mesa`. Known precedent: NetLogo does it, see https://github.com/NetLogo/NetLogo and look at `models` folder. But if the examples are going to be updated only for each major release, it doesn't make sense for `projectmesa/mesa` to track `projectmesa/examples`. Not sure what the rationale of the NetLogo developers is.

dmasad:
My main concern with putting examples in a submodule is that submodules are hard to use even for moderately-experienced Git / GitHub users.  Since the examples serve as de-facto tests, a contributor making an API change or adding a feature will probably also update the examples. Requiring the contributor to figure out submodules and keep their example branch and core branches in sync introduces a significant barrier to entry and opportunity for confusion and error.

jackiekazil:
RE: #735 - We need to get this move prioritized -- Unless someone else wants to take on, I  will take on POC before next dev session. Let me know if there are any concerns/considerations related to how this is done.

tpike3:
@jackiekazil No issues here I will add it to the dev notes

jackiekazil:
RE: #1451 - Time to dust this old ticket off! 
> Mesa Examples
> Discussion centered on accessibility to new users so they can easily run examples. To this end, it seemed to be the best way forward was to put current examples in the example repo and make more user friendy. TODO: Move examples to examples repo and make more user friendly.

* What needs to be considered with doing this before I just pick and move? 
* We were moving everything right? 
* Permissions need to be granted to all maintainers of Mesa & Mesa-Geo

jackiekazil:
RE: #1451 - Time to dust this old ticket off! 
> Mesa Examples
[...]

PR #2349, opened 2024-10-12
Move core example model back

EwoutH:
This PR moves four core examples back to Mesa:

- Boltzmann wealth
- Boid Flockers
- Conway's Game of Life
- Wolf Sheep

It uses the approach listed in https://github.com/projectmesa/mesa/discussions/2330#discussioncomment-10922651. It keeps the git history, with the only side effect that the PR numbers now link to the wrong repo (there's no easy way to fix this).

Moving these four core examples back into the main Mesa repository:
- Allows to open PRs that show how it changes both code and models
- Removes the need for two PRs for each feature (but still allows it)
- Simplifies version control, each example works with that exact commit/release of Mesa, always.
- Allow for more comprehensive testing of these examples within our CI pipeline, including batch runs and visualizations
- These diverse, high-quality examples will directly allow each users to test a model after installing Mesa.

After this PR is merged, these examples can be updated further to become fully exemplary.

<hr>

@projectmesa/maintainers I would request reviews of highest priority, since this git magic is very time intensive and I would like to know ASAP if I need to add other models.

To review:
1. Do you agree each of these four models adds unique value to keep as core example in Mesa?
2. Do we want more models included? If so which ones? (see https://github.com/projectmesa/mesa/discussions/2330#discussioncomment-10922651 for some suggestions)
3. Is this way of porting back the examples up to everybody standards?

Not merging before 3 green checkmarks, I want consensus on this.


quaquel:
I agree with this move. 

However, I also agree with the current description of mesa-examples: Seminal agent-based models developed using Mesa. So, what are seminal ABMs?

- Boltzmann wealth
- Boid Flockers
- Conway's Game of Life
- Wolf Sheep with grass regrowth
- Schelling
- Epstein civil violence
- Some version of Axelrod emergence of collaboration ( the most elaborate would be on a grid with evolution. I have a working version of this [here](https://github.com/quaquel/EPA123a/tree/367e5120ded3b49aef91140a1aae58b21b8ab808/src)
- el farol
- sugarscape
- Virus on a network (not really seminal, but a good example for networks); the alternative would be a traveling salesman. 

In my view, we should have a single emblematic version of each model. For some, we can go with the basic version (e.g., Schelling, Boltzmann); for others, we might go a bit beyond the standard. For example, we could to wolf sheep with event scheduling (as the benchmark version) and do sugarscape with traders, and use propery layers for sugar and spice regrowth.

I appreciate the massive effort to retain the git history. However, if we keep mesa-examples and turn this into a show and tell gallery that is not actively maintained (perhaps add an environment file to each example?) I am not sure we need to keep this full history here.

EwoutH:
> In my view, we should have a single emblematic version of each model.

Agreed. The big question is, where do we keep them?

My intention is to have not duplicates, or at least the least amount possible.

> For some, we can go with the basic version (e.g., Schelling, Boltzmann); for others, we might go a bit beyond the standard.

This might be an interesting distinction. We could have:
- Basic models, which are clean, simple and only use stabilized Mesa features
- Advanced models, which are more complex, and in which we allow experimental features (currently DEVS, PropertyLayer and Cell Space).

This way we don’t have duplicate models, but do have a nice split between stable and experimental.

So maybe all models that we maintain to our highest standards should be in Mesa itself. Others, which aren’t maintained (by us), and thus more user showcases, can be in Mesa-examples.

We might “degrade” Mesa models to Mesa-examples from time to time if we feel we have too much examples to actively maintain.

quaquel:
> This way we don’t have duplicate models, but do have a nice split between stable and experimental.

Yes exactly 

> So maybe all models that we maintain to our highest standards should be in Mesa itself. Others, which aren’t maintained (by us), and thus more user showcases, can be in Mesa-examples.

Yes, this is exactly what I am advocating with one addition: our examples in the main repo should be emblematic ABMs that everyone familiar with ABMs knows and have been covered extensively in e.g., popular scientific books on complexity science.

rht:
Overall, I disagree.

> Allows to open PRs that show how it changes both code and models
> Removes the need for two PRs for each feature (but still allows it)

Breaking changes are not meant to be frequent. As such, this is not a basic need. Even the [NumPy 2.0 migration guide](https://numpy.org/devdocs/numpy_2_0_migration_guide.html) is not that drastic. People need stability, as this is not web frameworks. Otherwise, tutorials, courses, LLM knowledge all are being made obsolete too often.

> Simplifies version control, each example works with that exact commit/release of Mesa, always.

This is monorepo vs polyrepo again. It's not as clear-cut as you have painted so far, i.e. monorepo good polyrepo bad.

> Allow for more comprehensive testing of these examples within our CI pipeline, including batch runs and visualizations

Testing the examples should be localized to these examples, and they already have a weekly CI to catch bugs from projectmesa/mesa changes. While unit testing the visualization elements in general should be automated according to Solara's way of testing.

> These diverse, high-quality examples will directly allow each users to test a model after installing Mesa.

This is an argument to improve pedagogy, which I could support. This does sound more convenient than `pip install mesa_models` or `pip install mesa[models]`. If the non-GUI Schelling model could become <=28 LOC (this is the shortest example in Agents.jl's benchmark), while still being reasonably coded, then maybe it's fine.

I haven't had the time to summarize the pros and cons in https://github.com/projectmesa/mesa/issues/134 and https://github.com/projectmesa/mesa/issues/963.

jackiekazil:
This is more complex than one might think for the reasons already stated. 
I do appreciate the requirement of 3 check marks. 

My hesitation is that people couldn't find the examples. I think we need to have a wider discussion about what our users need from examples and how to make them usable.

EwoutH:
Thanks for both your insights!

> Breaking changes are not meant to be frequent. As such, this is not a basic need.

This has nothing to do with if a change is breaking or not, having the examples in the main repo is useful for _any_ change. It allows you to directly show how a new feature can be used in an example, even if it also would have worked the old way.

This makes the PRs which we link to in the release notes also way more useful, since they now contain the code changes _and_ how it can be used in the examples.

> It's not as clear-cut as you have painted so far

Okay, but could you than argue for the polyrepo?

> This is an argument to improve pedagogy, which I could support.

Thanks!

> I haven't had the time to summarize the pros and cons in https://github.com/projectmesa/mesa/issues/134 and https://github.com/projectmesa/mesa/issues/963.

Looking forward to it!

<hr>

> This is more complex than one might think for the reasons already stated.

Could you expand a bit on which reasons or point to something?

> My hesitation is that people needed help finding the examples. 

I think this will make it easier for people to find the examples, especially if we document it well. Instead of installing something extra, and making sure its in the right place, etc., you can now just directly import them.
```Python
from mesa.examples import BoltzmanWealth
```
And they show up directly in your IDE.

<hr>

@quaquel so that we should make three lists:
1. Emblematic examples we keep basic & stable (for Mesa)
2. Emblematic examples that can be more complex and use experimental features (for Mesa)
3. Showcase examples that are not the two above (for Mesa-examples)

quaquel:
Why am I in favor of having examples in the main mesa repo?

1. It allows for more comprehensive integration testing then we currently can do. In fact mesa-example tests currently are breaking and have been braking for several weeks because of #2343. If they were in the main repo, we would have caught this problem much earlier and it would already have been fixed.
2. When developing new features, one of the most common questions all maintainers ask is to give examples of how the feature can be used. Because of this, devs (and gridspace before as well) have a dedicated example folder. For the signal stuff, I am doing the same. It would be so much more convenient to be able to illustrate features with one or more of the canonical examples. 
3. The current mesa-examples is unwieldy and not well maintained. As part of moving [examples over to grid spaces](https://github.com/projectmesa/mesa-examples/pull/198), I have also made many other fixes that were well over due such as proper use of a Python class hierarchy (instead of a horrible breed attribute or somethign similar), removal of all CamelCase notation etc. If we have a small, carefully selected set of examples in the main repo, we can much more easily commit to actively maintain these.
4. Building on 3, having a small set of examples in the main repo frees up mesa-examples to be a show and tell where we can be much more open to pull requests. See for example [the currenlty outstanding PRs](https://github.com/projectmesa/mesa-examples/pulls): many of these are either additional examples or variations of existing examples. We have been hesitant to merge these PRs because of concerns about their maintainability etc. This problem would be resolved if we don't have to actively maintain mesa-examples. We can be much more open to PRs with new examples. All that would be required is to agree amongst ourselves on a template for such a show and tell example. In my view, a minimum would be to have short readme explaining the example and an environment file to ensure future reproducibility. 
5. With the ema workbench, I have been including examples as part of the main repo for a very long time. This has been working great for teaching and explaining. Many questions about how to use a given feature can simply be answered by pointing to an example that is already installed. 

> My initial division would be:

Can you give some short motivation for this breakdown? 

I personally would include epstein (and el farol) also in basic. The main reason for me for keeping them in the mesa repo is that these are canonical ABMs. I would also include a much expanded version of pd_grid (which is just Axelrod emergence of collaboration on a grid) in advanced. Again, this is is a canonical ABM and enables showing off some nice Python / MESA features. You currently have virus on a network under advanced, but the code is very basic ATM.

tpike3:
I agree with putting canonicals model in the main repo. 

My main reason for this view is slightly altering Kurt Vonnegut's writing quote is "have mercy on the user," which falls under the pedagogical justification. From our current understanding our main user base is coming from academia. The more we can make Mesa an easier tool for them to use the better. Generalizing, however, the user is the target and this same dynamic goes for them. Being able to rapidly reference canonical works which serve as a examples or "How Tos" is critical to enabling education per scaffolded learning. As we expand Mesa and get into higher end models different backends (rust etc), use of HPCs etc, then we can optimize for a more diverse user community. Ideally after Mesa 3.0 is released we will do a push on our docs, best case we end up with something like Solara https://solara.dev/ 

Regarding the monorepo vs polyrepo, I am intrigued but cant see the path yet. When I can see the specifics as [described here](https://github.com/projectmesa/mesa/discussions/2330#discussioncomment-10923080), then I can make an informed recommendation. However, I put this in here to emphasize the hard problem for ABMs is how do we build a sustainable ecosystem so ABMs are as easy to build as say the MLs in scikitlearn. If we figure this out then we can get past boutique models and ABMs will become ubiquitous.  

Either way this is a great discussion! If you are not having hard conversations, you are not talking substantive change, and the great news is we can always change again! :)

EwoutH:
Thanks for your extensive reply.

Let me clarify one thing: This PR doesn’t make this a “monorepo” or whatsoever. Monorepo have multiple projects/packages in a single repository, this is still one project and one package in one repo. So the discussion about Mesa (mono)repositories is largely unrelated. I might have not communicated that clearly earlier, sorry.

If we discuss/decide to move Mesa-geo, Mesa-frames, etc into this repo, that becomes relevant again.

quaquel:
My breakdown would be:

#### Basic
- [boltzmann_wealth_model](https://github.com/projectmesa/mesa-examples/tree/main/examples/boltzmann_wealth_model). This covers grid spaces, simple data collection, and basic mesa stuff
- [boid_flockers](https://github.com/projectmesa/mesa-examples/tree/main/examples/boid_flockers). This covers continous spaces.
- [conways_game_of_life](https://github.com/projectmesa/mesa-examples/tree/main/examples/conways_game_of_life)
- [schelling](https://github.com/projectmesa/mesa-examples/tree/main/examples/schelling)
- [virus_on_network](https://github.com/projectmesa/mesa-examples/tree/main/examples/virus_on_network). This covers network spaces.

#### Complex
- [sugarscape_g1mt](https://github.com/projectmesa/mesa-examples/tree/main/examples/sugarscape_g1mt). Can be done with property layers for sugar and spices. 
- [wolf_sheep](https://github.com/projectmesa/mesa-examples/tree/main/examples/wolf_sheep). Can showcase ABM with scheduling
- [axelrod on a grid](https://github.com/projectmesa/mesa-examples/tree/main/examples/pd_grid). Ideally, I would go with an evolutionary version of this. It can be done to showcase a class hierarchy for agents, has a nice grid and can make for some neat visualizations as well. 
- [epstein_civil_violence](https://github.com/projectmesa/mesa-examples/tree/main/examples/epstein_civil_violence). Can showcase ABM with scheduling and some clever data collection

#### In stand alone show and tell repo
- [aco_tsp](https://github.com/projectmesa/mesa-examples/tree/main/examples/aco_tsp). This might be an alternative to virus on a network for showcasing network spaces. 
- [bank_reserves](https://github.com/projectmesa/mesa-examples/tree/main/examples/bank_reserves)
- [caching_and_replay](https://github.com/projectmesa/mesa-examples/tree/main/examples/caching_and_replay)
- [charts](https://github.com/projectmesa/mesa-examples/tree/main/examples/charts)
- [color_patches](https://github.com/projectmesa/mesa-examples/tree/main/examples/color_patches)
- [el_farol](https://github.com/projectmesa/mesa-examples/tree/main/examples/el_farol)
- [forest_fire](https://github.com/projectmesa/mesa-examples/tree/main/examples/forest_fire)
- [hex_snowflake](https://github.com/projectmesa/mesa-examples/tree/main/examples/hex_snowflake)
- [hotelling_law](https://github.com/projectmesa/mesa-examples/tree/main/examples/hotelling_law)
- [shape_example](https://github.com/projectmesa/mesa-examples/tree/main/examples/shape_example)

rht:
> > It's not as clear-cut as you have painted so far

> Okay, but could you than argue for the polyrepo?

It's bikeshedding, and I shouldn't rehash the existing general arguments in this topic. But IIRC the motivation for Mesa-Geo and mesa-frames being a separate repo is analogous to why the scikit-* (as @tpike3 had raised) are a separate repo from SciPy. It's to grow an ABM ecosystem.

> It allows for more comprehensive integration testing then we currently can do. In fact mesa-example tests currently are breaking and have been braking for several weeks because of https://github.com/projectmesa/mesa/issues/2343.

Should have been caught by the weekly CI. Also, it's understandable given that Mesa is undergoing lots of breaking changes recently, and shouldn't happen all the time.

> Let me clarify one thing: This PR doesn’t make this a “monorepo” or whatsoever. Monorepo have multiple projects/packages in a single repository, this is still one project and one package in one repo. So the discussion about Mesa (mono)repositories is largely unrelated. I might have not communicated that clearly earlier, sorry.

I initially thought this PR as a Trojan horse to later justify putting the examples back here (it came from #2330 after all). But it seems to be a fair bargaining session instead.

I wonder if we have a common ground that we want to deduplicate effort in maintaining the core/canonical/classic/basic examples (not sure). There are already performance benchmark examples, which sometimes uses experimental code instead of stable code. What is going to happen to them? How would the maintainers decide whether the core examples should use stable or experimental code? In any case, pedagogy should be prioritized first, which means to recommend the stable code. Which means we can't avoid having more than one Boltzmann/Schelling code as long as experimental API exist.

> When developing new features, one of the most common questions all maintainers ask is to give examples of how the feature can be used. 

I thought the performance benchmark examples have been used for this purpose so far.

wang-boyu:
Thanks for all the discussions so far. In general I'm hesitate about moving some examples back here. The main reason is that I share @jackiekazil's concern about the confusion this may cause to the users regarding where to find examples. This was also one of the reasons why examples were separated to its own repo previously.

To elaborate further, I don't generally expect users to look into Mesa source code. Instead, they should be redirected to example code and readthedocs site. Having examples in two different places makes this more difficult.

I understand that it would be easier to develop and test features with some examples back in. Hence to me, this is a trade-off between user experience vs. maintainer experience. In this case I would choose users over us maintainers.

I also agree with @quaquel's concern that currently examples are not well maintained. But, moving some of them back in would still leave rest of the examples at the risk of under-maintaince. Instead I would suggest to look for ways of properly maintain all examples.

But I'm not strongly against this PR. Similar to @jackiekazil, I'm not rejecting this PR, but I do not approve it either.

quaquel:
> I also agree with @quaquel's concern that currently examples are not well maintained. But, moving some of them back in would still leave rest of the examples at the risk of under-maintaince. Instead I would suggest to look for ways of properly maintain all examples.

Which is why I advocate for turning mesa-examples into more of a show-and-tell/gallery with environment files for each example. This has 2 benefits: we can accept examples more liberally, and, because of the environment file, guarantee a minimum level of reproducibilty. As developers, we are already spread thin, so I don't see a simple solution to improve maintenance of the current examples (let alone if they continue to grow as evidenced by the various outstanding PRs).

quaquel:
> I wonder if we have a common ground that we want to deduplicate effort in maintaining the core/canonical/classic/basic examples (not sure). There are already performance benchmark examples, which sometimes uses experimental code instead of stable code. What is going to happen to them? How would the maintainers decide whether the core examples should use stable or experimental code? 

In the EMA workbench, all examples are included in [the documentation](https://emaworkbench.readthedocs.io/en/latest/examples.html). This makes the examples very easy to find for all users. To avoid duplication, I have done the following there: when building the docs, the example folder is copied to the docs folder and the relevant files for read the docs are automagically generatored (all handled in conf.py). This implies that whenever an example is updated in the code base, or a new example is added, these changes or additions are automatically included in the documentation. 

For the benchmarks, we can easily do the same. Just copy the relevant folders from mesa.examples to the benchmark folder and run the benchmarks. This avoids duplication.

EwoutH:
> Should have been caught by the weekly CI.

And it might have, but that would always be after the actual change is made. Then a follow-up PR is needed to fix it, or revert it, while it could have been catched directly at the beginning in the PR itself.

> There are already performance benchmark examples, which sometimes uses experimental code instead of stable code. What is going to happen to them?

In my vision those won't be separate, so we merge the example and benchmark models, so there aren't any duplicate ones.

We probably benchmark some of the stable and some of the experimental models.

> The main reason is that I share @jackiekazil's concern about the confusion this may cause to the users regarding where to find examples.

I think we can make a clear split here, like @quaquel also suggested:
- Exemplary and classic examples in Mesa (which we maintain to high standards)
- Fun and creative projects in Mesa-examples

A huge advantage from having them in the main repo, is we can (also like Jan suggested) directly in the docs. Even more, we can use them in tutorials and in (future) interactive demos. And of course we will link to the example folder in the Readme and docs. It just makes a ton of use cases possible if we can directly refer to them.

> Hence to me, this is a trade-off between user experience vs. maintainer experience. 

I don't think this is actually the case, and we can optimize what's best for both.

There's also a transition area of users that are becoming contributors - which is a hugely important group for us. I think this will also help them, because the git-mess becomes much easier when working in a single repo.

@quaquel, I like your division of models.

So here's my updated proposal:
1. We use the division of models as suggested by @quaquel
2. We create two folders in Mesa examples: `examples/basic` and `examples/advanced`
3. We move over those 9 examples from mesa-examples, retaining git history.
4. We update those 9 to be exemplary.
5. We include them in Read the Docs (like currently on the [EMAworkbench](https://emaworkbench.readthedocs.io/en/latest/examples.html))
6. We update all other docs and readmes to point towards the new location
7. We integrate the benchmarks models in them and run benchmarks on them.
8. We update the tests to run continuous testing
9. We add visualisation testing
10. We add batch_run testing

quaquel:
> 9. We integrate the benchmarks models in them and run benchmarks on them.

Clarifying question: benchmark all 9 or just the same as we currently do?

EwoutH:
In my vision same as currently, due to run time. Maybe we could add an "full-benchmarks" method at some point that allows running all models as benchmarks (but takes longer).

Corvince:
Great discussion. I'll give my approval on this PR, regardless of how the final folder structure or example selection will look like.

Although bringing some examples back into the main repo might at first look like it might have been a mistake to move them outside in the first place. But I think we made everything right there and I think the mesa examples repo is a success story. We now have a lot more examples using mesa and more are coming through open pull requests. But it also is apparent now that we can no longer fully maintain all of them. 

So moving some examples back now and providing full support for them seems like a sensible step and in case of the benchmarks even reduces example duplication. And I also agree for the examples repo to switch to fixed requirement files so we always have a known state of dependencies for which they work regardless of how mesa and other libraries evolve. 

I think we can do a lot more in terms of example discoverability and retrieval but this is a separate topic

Corvince (approved):

EwoutH:
Thanks for your perspective Corvince!

<hr>

One interesting note is that we got multiple issues / questions filed (https://github.com/projectmesa/mesa-examples/issues/217, https://github.com/projectmesa/mesa-examples/issues/216, https://github.com/projectmesa/mesa/discussions/2356) about examples not working. All those came from people trying to use Mesa 3.0 examples with Mesa 2.x. I think we can also improve this experience if examples are always bundled with the working Mesa version. And included in the docs, in which we now have a version selector (https://github.com/projectmesa/mesa/pull/2324)

<hr>

If I'm tracking this correctly, it looks like Tom, Jan, Vincent and I are now in favor of moving the seminal / core examples back into the main Mesa repo. rht is against. Jackie and Wang, in your last replies you were hesitant/undecided, is this still the case?

wang-boyu:
Yup!

rht:
> One interesting note is that we got multiple issues / questions filed (https://github.com/projectmesa/mesa-examples/issues/217, https://github.com/projectmesa/mesa-examples/issues/216, https://github.com/projectmesa/mesa/discussions/2356) about examples not working. All those came from people trying to use Mesa 3.0 examples with Mesa 2.x. I think we can also improve this experience if examples are always bundled with the working Mesa version. And included in the docs, in which we now have a version selector (https://github.com/projectmesa/mesa/pull/2324).

Once again, this is not happening all the time. This could solved by checking `mesa.__version__` when running the code. Actually, since in e.g. a docker-compose.yml file, there is a version field, why not add that enforcement in the examples code. The "environment file" requirements.txt doesn't constrain enough, as this is not `npm`.

EwoutH:
Superseded by https://github.com/projectmesa/mesa/pull/2358, thanks everyone!

Issue #2323, opened 2024-09-24
Remove "model" from agentset initialization


Corvince:
**What's the problem this feature will solve?**
I just noticed that `AgentSet` is still initialized with a ´model` instance. I think this is unnecessary, since the only use case is retrieving `model.random`. I think I remember some discussion about `random` by @quaquel regarding its use in agentset. But before we declare AgentSet fully stable for 3.0 we should probably revisit this issue. Without model AgentSet will be completely independent, simplifying testing et al.

quaquel:
I think the way we handle this in the grid spaces is the way forward for now. Instead of passing model, there you need to pass random explicitly. The main drawback is that if you run multiple experiments without re-initializing the model class, you can't explicitly control random. In #2021, I played with other possible solutions but got a bit stuck (see e.g., [this comment](https://github.com/projectmesa/mesa/pull/2021#issuecomment-1919813029) and them mainly point 3.).

EwoutH:
Might be interesting how other scientific libraries approach this. Have you read SPEC 7? https://scientific-python.org/specs/spec-0007/

rht:
I see access to `model` as a catch-all object where you can retrieve any globally accessible content (clock, steps, global average temperature, total money supply, etc). If you localize access to just the RNG, then:
1. More RAM on the mental model: User will have to remember that `AgentSet` only has access to the RNG instead of a less specific `model`, and another set of rules for other Mesa components
2. In case in the future you want `AgentSet` to access another global content (could be the system clock), you have to modify AgentSet's list of public attributes to include this object.

Corvince:
Interesting viewpoint. I think there is some truth in that, but I am still in favor of keeping the API surface minimal. I don't understand your first point of user RAM. Users should not have to think about what agentset has access to and what not, because why would you want to access the model through agentset (as a user).

I agree that "model" is a catch-all object, but for that reason it shouldn't be exposed completely to every subcomponent. This only creates circular dependencies, which isn't a problem per se, but can lead to nasty bugs and surprising behavior. 

Also I would immediately make random optional with the hope to deprecate it later and only be dependent on agents. So instead of adding more dependencies later I would strive to remove all dependencies, this moving in the opposite direction

quaquel:
There are a few additional reasons to explicitly pass random.

[...]

PR #2350, opened 2024-10-13
replace model with random in AgentSet init

quaquel:
This PR replaces the model argument in `AgentSet.__init__` with `random` as a keyword argument as argued in #2323.

**Motivation:**
* the only thing retrieved from model is the random number generator. Passing `model` serves no other function and there is no other use case for having model in the agentset. 
* For advanced use cases, explicitly passing random allows users to control which random number generator is being used. The most common reason for wanting this is for variance reduction through common random numbers. See #2323 and links provided there for a further elaboration on this argument.
* I cannot currently change `CellCollection.agents` to returning an agentset because CellCollection does not have a reference to model even though it does have access to its random number generator. 

Some have argued that this change creates extra load on the user. I disagree. First, In none of the examples in mesa-examples is the user creating an AgentSet object themselves. This highlights that it will be rare for users to create AgentSet themselves directly in all but the most advanced use cases. It also implies that this change is almost excusively confined to the inner workings of mesa. Second, this API is identical to the new gridspaces were we pass random explicitly instead of passing model only to get access to the central random number generator. So, this PR in fact makes the code base internally more coherent and therefore easier to use.

Another argument against this change is that we might potentially expand in the future AgentSet to use other model attributes. Again, I am not convinced by this unless clear examples of useful AgentSet functionality are given that require access to Model. We have had AgentSet for almost a year. No example has come up, I cannot come up with a convincing example, and None have not been offered.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -99,14 +99,16 @@ class AgentSet(MutableSet, Sequence):
         which means that agents not referenced elsewhere in the program may be automatically removed from the AgentSet.
     """
 
-    def __init__(self, agents: Iterable[Agent], model: Model):
+    def __init__(self, agents: Iterable[Agent], random: Random | None = None):
         """Initializes the AgentSet with a collection of agents and a reference to the model.
 
         Args:
             agents (Iterable[Agent]): An iterable of Agent objects to be included in the set.
-            model (Model): The ABM model instance to which this AgentSet belongs.
+            random (Random): the random number generator
         """
-        self.model = model
+        if random is None:
+            random = Random()  # FIXME see issue 1981, how to get the central rng from model
+        self.random = random
         self._agents = weakref.WeakKeyDictionary({agent: None for agent in agents})
 
     def __len__(self) -> int:
@@ -177,7 +179,7 @@ class AgentSet(MutableSet, Sequence):
 
         agents = agent_generator(filter_func, agent_type, at_most)
 
-        return AgentSet(agents, self.model) if not inplace else self._update(agents)
+        return AgentSet(agents, self.random) if not inplace else self._update(agents)
 
     def shuffle(self, inplace: bool = False) -> AgentSet:
         """Randomly shuffle the order of agents in the AgentSet.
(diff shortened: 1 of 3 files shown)


EwoutH (approved):
I agree with this change. Thanks especially for the extensive motivation, that’s really appreciated.

> Another argument against this change is that we might potentially expand in the future AgentSet to use other model attributes.

This was the first thing I was thinking about. However, the AgentSet has been with us since 2.2, is very actively integrated and developed and so far we never needed it. So a few months back I might give this argument a lot of weight, but now I think we will perfectly manage without it.

Since this is technically a breaking change, I would like a second maintainer to approve, but from my perspective, good to go!

rht:
> I cannot currently change CellCollection.agents to returning an agentset because CellCollection does not have a reference to model even though it does have access to its random number generator.
> So, this PR in fact makes the code base internally more coherent and therefore easier to use.

Having access to the model everywhere is arguably also coherent and easy to use. So it could go either way.

> We have had AgentSet for almost a year. No example has come up, I cannot come up with a convincing example, and None have not been offered.

> This was the first thing I was thinking about. However, the AgentSet has been with us since 2.2, is very actively integrated and developed and so far we never needed it. So a few months back I might give this argument a lot of weight, but now I think we will perfectly manage without it.

It's because the use cases (the "experimental evidences") are limited to the existing mesa-examples, and we haven't received feedback from advanced users yet. This is in contrast to the datacollector where we received lots of requests for multi-agent data collection.

But I'm fine with this PR. It can always be changed back in the future when needed. The added bonus with this PR is that I have observed that AgentSet's utility may extend beyond ABM, as an arbitrary object DataFrame, and so tying it to an ABM model might actually be limiting it.

rht (approved):

EwoutH:
Apparently this broke our docs build: https://readthedocs.org/projects/mesa/builds/25927550/

EwoutH:
@quaquel could you check what's going wrong in the tutorial?

quaquel:
The tutorial is still using a scheduler, so the tutorial needs a bigger overhaul.

I have no idea what causes the error and I can't seem to replicate it locally.

EwoutH:
Okay, @tpike3 is working on a tutorial update, let's first wait if he already has fixed the issue (feel free to open a draft PR with your current progress, Tom).

quaquel:
should I split the fix in time.py into a seperate PR then?

PR #2352, opened 2024-10-14
Add Model.rng for SPEC-7 compliant numpy random number generation

quaquel:
This PR adds a [SPEC 7](https://scientific-python.org/specs/spec-0007/) compliant numpy random number generator to the Model class. 

**Implementation**
I have chosen a straightforward approach. I have added an additional optional keyword argument `rng` that can be used to seed the `numpy.random.Generator` instance when instantiating a model. I have also added 2 new attributes to `Model`: `rng` and `_rng`. The first is the generator which can be used analogously to model.random. The second is `_rng` which is the dictionary describing the state of the numpy.random.Generator instance. Knowing this dict allows for reproducibility.

The code itself and e.g., the typing behavior is lifted straight from SPEC 7.  

At the moment, a user has to pass either seed or rng, but not both which will raise a value error. The value passed for the one will be used for the other as well. If this argument is invalid, a valid integer based seed/rng number will be generated instead. What does this mean practically: if you pass seed we start with seeding `random.Random`. We try to use seed also for `numpy.random.default_rng`. If this raises a type error because seed is not valid here (e.g., seed is a float). Instead, we draw a valid integer via the seeded `random.Random()`. It works the same way if you only pass rng. So we create a Generator, try to use rgn for `random.Random`. If this failes because it is e.g. an ArrayLike sequence of ints, we draw a random int to use for `random.Random(()`

**What's next?**
At present, I have not touched the existing random throughout the code base. SPEC 7 is only about numpy and does not imply anything for python.random. It is for me an open question whether we should completely deprecate python.random throughout mesa in favor of numpy.random. I am inclined to be in favor of making this change because the quality of the random number generators in numpy is superior (long story about mersene twister's "bad" behavior when seeded with non 32 bit numbers or with adjacent numbers in case of seed analysis). However, this requires a substantial effort throughout the code base and might come with a performance penalty. What do others think and should we do this before MESA 3.0 becomes stable?

### DIFF
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -45,11 +45,11 @@ class Model:
     """
 
     def __init__(
-            self,
-            *args: Any,
-            seed: float | None = None,
-            rng: RNGLike | SeedLike | None = None,
-            **kwargs: Any,
+        self,
+        *args: Any,
+        seed: float | None = None,
+        rng: RNGLike | SeedLike | None = None,
+        **kwargs: Any,
     ) -> None:
         """Create a new model.
 
@@ -65,7 +65,7 @@ class Model:
             kwargs: keyword arguments to pass onto super
 
         Notes:
-            you have to pass either seed or rng, but not both. 
+            you have to pass either seed or rng, but not both.
 
         """
         super().__init__(*args, **kwargs)
@@ -76,17 +76,18 @@ class Model:
             raise ValueError("you have to pass either rng or seed, not both")
         elif seed is None and rng is None:
             pass
-        elif seed is None:
+        elif seed is None or rng is None:
             seed = rng
-        elif rng is None:
-            seed = rng
-
 
         self.rng: np.random.Generator = np.random.default_rng(rng)
-        self._rng: dict = self.rng.bit_generator.state  # this allows for reproducing the rng
+        self._rng: dict = (
+            self.rng.bit_generator.state
+        )  # this allows for reproducing the rng
 
         self.random = random.Random(seed)
-        self._random = self.random.getstate()[1] # this allows for reproducing stdlib.random
+        self._random = self.random.getstate()[
+            1
+        ]  # this allows for reproducing stdlib.random
 
         # Wrap the user-defined step method
         self._user_step = self.step
@@ -235,11 +236,11 @@ class Model:
         self.rng = np.random.default_rng(rng)
 
     def initialize_data_collector(
-            self,
-            model_reporters=None,
-            agent_reporters=None,
-            agenttype_reporters=None,
-            tables=None,
+        self,
+        model_reporters=None,
+        agent_reporters=None,
+        agenttype_reporters=None,
+        tables=None,
     ) -> None:
         """Initialize the data collector for the model.
 
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -82,9 +82,7 @@ class Model:
             rng = seed
 
         self.rng: np.random.Generator = np.random.default_rng(rng)
-        self._rng: dict = (
-            self.rng.bit_generator.state
-        )  # this allows for reproducing the rng
+        self._rng: dict = self.rng.bit_generator.state # this allows for reproducing the rng
 
         if seed is None:
             seed = int(self.rng.integers(np.iinfo(np.int32).max))
@@ -237,6 +235,7 @@ class Model:
             seed: A new seed for the RNG; if None, reset using the current seed
         """
         self.rng = np.random.default_rng(rng)
+        self._rng = self.rng.bit_generator.state
 
     def initialize_data_collector(
         self,
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -8,6 +8,7 @@ Core Objects: Model
 from __future__ import annotations
 
 import random
+import sys
 import warnings
 from collections.abc import Sequence
 
@@ -74,23 +75,21 @@ class Model:
 
         if (seed is not None) and (rng is not None):
             raise ValueError("you have to pass either rng or seed, not both")
-        elif seed is None and rng is None:
-            pass
         elif seed is None:
-            seed = rng
-        elif rng is None:
-            rng = seed
+            self.rng: np.random.Generator = np.random.default_rng(rng)
+            self._rng = self.rng.bit_generator.state  # this allows for reproducing the rng
 
-        self.rng: np.random.Generator = np.random.default_rng(rng)
-        self._rng: dict = (
-            self.rng.bit_generator.state
-        )  # this allows for reproducing the rng
-
-        if seed is None:
             seed = int(self.rng.integers(np.iinfo(np.int32).max))
+            self.random = random.Random(seed)
+            self._seed = seed  # this allows for reproducing stdlib.random
+        elif rng is None:
+            self.random = random.Random(seed)
+            self._seed = seed  # this allows for reproducing stdlib.random
 
-        self.random = random.Random(seed)
-        self._seed = seed  # this allows for reproducing stdlib.random
+            rng = self.random.randint(0, sys.maxsize)
+            self.rng: np.random.Generator = np.random.default_rng(rng)
+            self._rng = self.rng.bit_generator.state
+
 
         # Wrap the user-defined step method
         self._user_step = self.step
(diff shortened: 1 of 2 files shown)


EwoutH:
Implementation looks clean and good!

> SPEC 7 is only about numpy and does not imply anything for python.random. It is for me an open question whether we should completely deprecate python.random throughout mesa in favor of numpy.random.

I think we should have one consistent recommended way to pass seeds. In my view, the random number generator should be reset directly in Model init with a certain seed, to make model runs also reproducable when run in (for example) a Jupyter notebook multiple times, or in (custom) batch scripts.

How it's implemented internally doesn't matter that much in my opinion. If that seed number is passed to both python.random and numpy.random and that provides consistently reproducable results, that perfectly fine. If it uses numpy.random for everything, that's also great.

It's something we should test on a very minimal model I guess.

> What do others think and should we do this before Mesa 3.0 becomes stable?

I'm willing to advocate and enable an aggressive timeline for this.

quaquel:
> I think we should have one consistent recommended way to pass seeds. In my view, the random number generator should be reset directly in Model init with a certain seed, to make model runs also reproducable when run in (for example) a Jupyter notebook multiple times, or in (custom) batch scripts.

I started out this way but its not trivial. For example, `python.random.Random` does accept a float, `numpy.random.default_rng` does not accept a float. Also what defines the state of the rng is different accross both. In case of `python.random.Random`, all you need is the initial value for seed. In case of `numpy.random.default_rng` state is a dict with several fields (the class of the generator, another state dict with state and inc as fields, and some other fields). 

If there is a broader concensus on moving aggresively on this, I would probably approach it a bit differently. Now, I just added `rng` next to `random`. In case of an aggressive timeline, I would switch to `rng` as the basis and use this to seed `python.random.Random`. It would also imply raising at least a deprecation warning if both `seed` and `rng` are passed to `Model.__init__` as avocated by SPEC 7, and deprecation warnings from Model,random and Agent.random.

EwoutH:
> For example, python.random.Random does accept a float, numpy.random.default_rng does not accept a float. Also what defines the state of the rng is different accross both.

Can we solve this by requiring the user to input the seed in a certain dtype, and doing the rest behind the scenes? In this case, require inputting a int and constructing the remainder of the dict in the Model init?

rht:
In general, there is no need to rush, as SciPy (one of the endorsers of the spec and the one behind the spec) itself has yet to make a move https://github.com/scipy/scipy/issues/14322 (because their user-facing surface with random state is so huge). No project hosted on GH has completely adopted the spec yet (https://github.com/topics/spec-7).

The restriction in not being able to use a float for the seed will be a huge one. What should be the recommended conversion from users's existing seeds? If we move aggressively with the `rng` today, then existing users results can no longer be reproducible in Mesa 3.0. If we still have both random and rng, then they can migrate to Mesa 3.0 while still having their results intact.

rht:
> There are various other changes in MESA 3.0 that make me wonder whether perfect numerical reproducability is not already impossible when switching from MESA 2.x to MESA 3.

If it is already at the point of no return, then it is fine to do the breaking change. (https://github.com/scikit-learn/scikit-learn/issues/16988#issuecomment-1518037853 (from the creator of https://github.com/pyutils/line_profiler) has an extensive read on why a generator construct eliminates a lot of headaches in scikit-learn. I haven't read the lengthy explanation version.)

I haven't said elsewhere, but `[rng.random() for i in range(100)]` is a lot slower with NumPy than with stdlib random. The performance benchmark might show this outcome later.

quaquel:
> If it is already at the point of no return, then it is fine to do the breaking change.

Ok, I'll update the PR based on some of the other points discussed but make sure we keep random in there while adding spec-7 compliant rng. 

> I haven't said elsewhere, but [rng.random() for i in range(100)] is a lot slower with NumPy than with stdlib random. The performance benchmark might show this outcome later.

This is an important point and a possible argument for keeping both stdlib random and rng if the benchmarks further done the line indicate that the overhead is massive.

quaquel:
Ok, I have updated the PR. Users can now not pass both seed and rng. I currently raise a value error. This is in line with SPEC 7. If both seed and rng are None, I create `numpy.random.default_rng` and use this to create a random integer with which to seed `random.Random()`. If either is defined, the other takes the same value. The only drawback is that some of the argument that are valid for `random.Random` are not valid for `numpy.random.default_rng` and vice versa. There is no simple solution for this. 

The current implementation limits seed and rng to values that are valid in both. Based on the documentation, the only valid argument for both is `int`. I am open for ideas on how to approach this better.

EwoutH (approved):
I don’t really understand all the little details and nuances of this implementation and how it will interact with everything, but in general it looks clean and doesn’t break anything, and I agree on a conceptual level a SPEC 7 implementation in Mesa is useful.

So I’m approving this, but requesting for the PR to remain open for about a day or so other maintainers have a chance to take a look at it.

EwoutH:
@projectmesa/maintainers any of you additional insights / an opinion on this?

EwoutH:
@quaquel Unfortunately, this PR broke pre-commit.

quaquel:
so, what does that mean? I don't know what is covered by pre-commit?

EwoutH:
The [`.pre-commit-config.yaml`](https://github.com/projectmesa/mesa/blob/main/.pre-commit-config.yaml) defines what's checked with each commit. Ruff is part of that, and in this case the one that [fails](https://results.pre-commit.ci/run/github/24245737/1728985211.-4BYKg7CSGKJUH7_sYVJ2g):

image

jackiekazil:
Was this not noticed before the merge? 
I wonder how this accidentally got through to ensure it doesn't happen in the future.

EwoutH:
A while ago we un-required some checks, because some times they error either erroneously or because of unrelated issues to that specific PR, which than blocked it unnecessarily. Yesterday I re-required pre-commit and one of the build jobs again.

It’s a bit about finding a balance, but in general I’m a fan of educating ourselves about the systems and checks we use and interpreting them as guidelines, then making everything strict rules and being held hostage to automated processes.

These things are really minor and easily fixed, so also really not a big deal. As we say in Dutch: “Waar gehakt wordt, vallen spaanders”.

Corvince:
> Yesterday I re-required pre-commit and one of the build jobs again.

The problem was that when you have a PR that doesn't change code, for example a pure documentation update, then the build runs are not triggered at all. But they are still required so you run into a situation where you have to use the big red button, which isn't available to all maintainers. 

I just wish there would be an extra step involved when you want to merge something, but some checks are failing. Right now it's way too easy to miss failed checks. But making everything required would be too strict, sometimes it's okay when you know you have a dependency that will fix the error soon. 

But let's see how it works out now with the required workflows

quaquel:
In addition to @Corvince's point, I was just assuming pre-commit etc. was required and did not check (I always check unit tests). So @EwoutH, did you post somewhere that you changed the required checks, and did I miss it? If not, a simple post in Element would be useful as a heads up.

EwoutH:
> The problem was that when you have a PR that doesn't change code, for example a pure documentation update, then the build runs are not triggered at all.

You’re right, that’s why we removed that one. I removed it again as a required check.

@quaquel Tom did it a while back before I had admin permissions. I will communicate clearly if I change anything in the future.

Currently only pre-commit is a required check, since that runs always. It’s good habit as a maintainer to always check if all checks in GitHub are green before merging, and if something isn’t to check the root cause. If it’s due to the PR, fix it within the PR, otherwise post it somewhere relevant.

EwoutH:
Found this FIXME in the Agent code, with this merged, could this be resolved?

https://github.com/projectmesa/mesa/blob/22784dfec9c8436f18260ec6db0c344774075ca2/mesa/agent.py#L125

quaquel:
The issue still exists for both `AgentSet` and the various new-style discrete spaces. 

At the moment, they all create a new `Random` instance if `random` is `None`, which is convenient but also makes it easy to create models that are not reproducible. In particular, for the cell spaces, this has bitten me already a few times.

Given spec-7, we cannot really go the global route, and as discussed in #1980, there are additional reasons why that is undesirable. We might consider raising a user warning?

EwoutH:
I love sensible user warnings, sounds like a good idea!

PR #2353, opened 2024-10-14
refactor: Simplify Schelling code

rht:
Ported from https://github.com/projectmesa/mesa-examples/pull/222.


quaquel (approved):

benchmarks/Schelling/schelling.py
```diff
@@ -34,11 +34,8 @@ def __init__(
 
     def step(self):
         """Run one step of the agent."""
-        similar = 0
-        neighborhood = self.cell.get_neighborhood(radius=self.radius)
-        for neighbor in neighborhood.agents:
-            if neighbor.type == self.type:
-                similar += 1
+        neighbors = self.cell.get_neighborhood(radius=self.radius).agents
+        similar = sum(1 for neighbor in neighbors if neighbor.type == self.type)
```
EwoutH, line 38:
```suggestion
        similar = len([neighbor for neighbor in neighbors if neighbor.type == self.type])
```
I benchmarked it, and it seems using list comprehension and `len()` is faster in this case.

- Generator sum: 0.251
- Len list comprehension: 0.196
- Manual loop: 0.212

Theoretically it could use a bit more memory, but the list of neighbours won't be very long so in this case the increased performance matters far more.

rht: Maybe just revert back to manual loop? Reasonably fast and also doesn't need extra memory. Did you benchmark on 3.13? What if no-GIL/JIT (will be the norm in the future) is enabled?

EwoutH: Benchmark was on 3.12. I don’t see how GIL is relevant here. JIT we can benchmark when it’s enabled by default (in 3.14).

I would go with the `len` approach. Memory is completely negligible on these kinds of sample sizes, and the list isn’t stored anywhere.

rht: > Benchmark was on 3.12.

I see.

> I don’t see how GIL is relevant here.

You never know. In Cython, nogil code is usually blazing fast, even on list operations.

> JIT we can benchmark when it’s enabled by default (in 3.14).

There is a quick flag to enable this. I will benchmark when I have time.

EwoutH (approved):

PR #2359, opened 2024-10-15
fix: pass `model.random` to schedulers

quaquel:
As pionted out in by @EwoutH, #2350 broke the tutorial. This is due to not updating the schedulers to pass model.random. This fixes that bug (and does a few minor updates to the tutorial).

The tutorial needs a complete update to remove schedulers and use (in my view) the new grid spaces. I am not going to do that in this PR.


EwoutH (commented):
Thanks for the quick fix, changes to time.py look good. It's nice to give it a proper deprecation period this way.

Few comments about the files.

> I am not going to do that in this PR.

So you say ;)

Luckily Tom is already on it!

quaquel:
I reverted all notebook stuff to make @tpike3 life easier.

EwoutH (approved):
Awesome!

PR #2361, opened 2024-10-15
Code coverage: ignore experimental and visualization

Corvince:
As raised by @EwoutH frequent failing codecov tests can easily lead to missing other test failures. 

This PR fixes the "experimental" ignore, because we can be more lenient with experimental features and introduces a "visualization" ignore. Code coverage is mostly meaningful for unit tests, but unit testing visualization code is not always helpful. We still should do something about testing our visualization, but this should primarily aim at integration tests, not unit tests.

### DIFF
--- a/codecov.yaml
+++ b/codecov.yaml
@@ -6,8 +6,8 @@ coverage:
         threshold: 1%
 
 ignore:
-  - "experimental/**"
   - "benchmarks/**"
-  - "visualization/**"
+  - "mesa/experimental/**"
+  - "mesa/visualization/**"
 
 comment: off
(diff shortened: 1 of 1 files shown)


EwoutH (approved):
Thanks for fixing this!

When the examples get merged I should probably also add `"examples/**"`, right?

Corvince:
> Thanks for fixing this!
> 
> When the examples get merged I should probably also add `"examples/**"`, right?

No, I think once the examples are merged we should include running them in the CI as soon as possible. And once they are run they should also be covered. If some lines are not covered than we probably need to fix the examples.

PR #2366, opened 2024-10-15
add codecov token, fixes #2363

### LINKED ISSUE #2363: Codecov upload broken - needs token from someone with admin access
**Describe the bug**
Working on #2361 I now discovered that our latest commit uploaded to codecov is 2 months old. Regularly posting coverage reports stopped working in January

**Expected behavior**
Codecov reports should be uploaded by github actions everytime they are run 

**To Reproduce**
```Error: Codecov token not found. Please provide Codecov token with -t flag.``` 
can be seen for every GA action under "codecov", for example
https://github.com/projectmesa/mesa/actions/runs/11346031665/job/31554205137?pr=2362

**Additional context**
Here are the required steps to add a codecov token to our CI
https://docs.codecov.com/docs/github-2-getting-a-codecov-account-and-uploading-coverage#add-the-codecov-token-to-your-repo-or-your-org
This can only be done by someone with admin access to mesa.

Corvince:
Lets see if it works now

### DIFF
--- a/.github/workflows/build_lint.yml
+++ b/.github/workflows/build_lint.yml
@@ -60,6 +60,8 @@ jobs:
     - if: matrix.os == 'ubuntu'
       name: Codecov
       uses: codecov/codecov-action@v4
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }} 
 
   examples:
     runs-on: ubuntu-latest
--- a/.github/workflows/build_lint.yml
+++ b/.github/workflows/build_lint.yml
@@ -61,7 +61,7 @@ jobs:
       name: Codecov
       uses: codecov/codecov-action@v4
       with:
-        token: ${{ secrets.CODECOV_TOKEN }} 
+        token: ${{ secrets.CODECOV_TOKEN }}
 
   examples:
     runs-on: ubuntu-latest
--- a/.github/workflows/build_lint.yml
+++ b/.github/workflows/build_lint.yml
@@ -61,6 +61,7 @@ jobs:
       name: Codecov
       uses: codecov/codecov-action@v4
       with:
+        fail_ci_if_error: true
         token: ${{ secrets.CODECOV_TOKEN }}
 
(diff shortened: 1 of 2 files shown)


quaquel:
seems to be working again!

Corvince:
Yes! But we still miss the 90% target, so I lowered it temporarily to 80%. This is mostly caused by the deprecated time.py file, which is untested. Were the tests already removed or did they never exist? Anyways it would be unfair to have every PR fail because our project is currently undercovered, thats why i lowered it to 80%.

I also enabled a "fail_ci_if_error" flag, so hopefully we catch errors with codecov earlier next time. Or maybe it will cause too many errors, in which we can later disable it again. Lets see

quaquel:
> Yes! But we still miss the 90% target, so I lowered it temporarily to 80%. This is mostly caused by the deprecated time.py file, which is untested. Were the tests already removed or did they never exist? Anyways it would be unfair to have every PR fail because our project is currently undercovered, thats why i lowered it to 80%.

I agree. Not perse for this pr, but could we not exclude time.py somehow?

Corvince:
Not sure how I would feel about this. Even when deprecated, I think its good to highlight gaps in our code coverage. Of course no one is going to write tests for time.py now, but since the 80% or 90% mark is arbitrary anyways I feel like including it is less cheating. But I don't have a strong opinion about that, so if someone else wants to open a PR to exclude it I wont reject it.

quaquel:
There was a test_time.py that I think got removed when we deprecated it. This was unfortunate because although deprecated it is still being used in placed. Today I fixed #2359 which would not have been needed if test_time.py had not been removed. I'll try to remember adding it back in asap.

EwoutH:
> There was a test_time.py that I think got removed when we deprecated it.

Yeah sorry that's a mistake from my side, dully noted for future reference.

Thanks a lot for picking this up so quickly!

PR #2368, opened 2024-10-15
fix: register agent after creating unique_id and pos attributes

wang-boyu:
One gis example uses agent's `unique_id` to compare and hash agents: https://github.com/projectmesa/mesa-examples/blob/5758b6504fb26a2af05d861f29d32a81a07200f6/gis/agents_and_networks/src/agent/building.py#L34-L40

It was working until https://github.com/projectmesa/mesa/pull/2328, in which the call to `self.model.register_agent(self)` in `Agent.__init__()` was moved from _below_ `self.unique_id = next(self._ids[model])` to _above_ it.

As a result, in the gis example we now have the following error message:

```bash
File "/Users/boyu/GitHubProjects/mesa-examples/gis/agents_and_networks/src/agent/building.py", line 23, in __init__
    super().__init__(model=model, geometry=geometry, crs=crs)
  File "/Users/boyu/GitHubProjects/mesa-examples/venv/lib/python3.12/site-packages/mesa_geo/geoagent.py", line 36, in __init__
    Agent.__init__(self, model)
  File "/Users/boyu/GitHubProjects/mesa-examples/venv/lib/python3.12/site-packages/mesa/agent.py", line 65, in __init__
    self.model.register_agent(self)
  File "/Users/boyu/GitHubProjects/mesa-examples/venv/lib/python3.12/site-packages/mesa/model.py", line 145, in register_agent
    self._agents[agent] = None
    ~~~~~~~~~~~~^^^^^^^
  File "/Users/boyu/GitHubProjects/mesa-examples/gis/agents_and_networks/src/agent/building.py", line 40, in __hash__
    return hash(self.unique_id)
                ^^^^^^^^^^^^^^
AttributeError: 'Building' object has no attribute 'unique_id'
```

This PR should fix it. But I'm not very sure whether this will cause any trouble.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -62,9 +62,9 @@ class Agent:
         super().__init__(*args, **kwargs)
 
         self.model: Model = model
-        self.model.register_agent(self)
         self.unique_id: int = next(self._ids[model])
         self.pos: Position | None = None
+        self.model.register_agent(self)
 
     def remove(self) -> None:
         """Remove and delete the agent from the model."""


quaquel:
This code change seems fine to me. There is no particular reason for the ordering in `Agent.__init__` and the code also does not imply such ordering.

Why do you use a custom hash function in the example?

EwoutH (approved):
Sounds smart to first assign all the variables to the agent, and then register with the model. Especially if it helps the hash problem!

quaquel:
> Why do you use a custom hash function in the example?

Just to clarify, I am fine with the fix, but I think you don't need the hash function in your example anyway, avoiding the need for the fix. Hashing on custom classes is available by default already via `id(self)`, which returns the memory address used by the object.

wang-boyu:
Initially there was only a `__eq__()` method so that I could do `building_a == building_b` instead of `building_a.unique_id == building_b.unique_id`. But having just the `__eq__()` method without `__hash__()` was causing troubles, and it was subsequently fixed in https://github.com/projectmesa/mesa-examples/commit/f24626e4dac359fd3d3c23e3db8e154497140cd8 as part of https://github.com/projectmesa/mesa-examples/pull/187.

However now I looked at the example, `building_a == building_b` was not really used anywhere anymore, which was probably removed during development. So removing `__eq__()` as well as `__hash__()` from the example will most likely avoid the issue. But I still think that this fix is necessary in case someone else wants a custom `__eq__()` and `__hash__()`.

Thanks @quaquel and @EwoutH for the review btw.

wang-boyu:
Merging now. Thanks!

PR #2373, opened 2024-10-16
examples: Clean-up Epstein Civil Violence and PD grid

EwoutH:
Clean up the epstein civil voilence models.

One issue is that a deepcopy currently isn't possible, and Solara currently uses that.

```Python
model1 = EpsteinCivilViolence()
import copy
# make deepcopy
model2 = copy.deepcopy(model1)
```

@quaquel would you like to pick this further up?


quaquel:
I have committed a possible fix in line with my previous comment. In short, In `Cell.__getstate__`, I set "connections" to an empty dict. In `Grid.__setstate__`, I rebuild the connections via `Grid._connect_cells()`. It needs further testing and additional tests, but it seems to work.

EwoutH:
Thanks! I will continue updating the examples, if this works for all of them than we can include this in a separate PR.

quaquel:
>Thanks! I will continue updating the examples, if this works for all of them than we can include this in a separate PR. 

Ok, I'll move the fix into a separate PR and add additional unit tests. Let me know how you get on with the other examples.

EwoutH:
Awesome!

Updated `pd_grid` (Demographic Prisoner's Dilemma on a Grid). One issue is that we can't visualize any of the cell spaces currently. Without that the visualisations are quite useless. Tracking issue here:
- https://github.com/projectmesa/mesa/issues/2377

quaquel:
What's there is good. Would it make sense to remove the notebooks?

quaquel:
Yep the merge conflict is massive (as with the docs stuff). Might be better to start a new branch and copy these changes into it (As you are doing with docs atm).

quaquel:
Superseded by #2408

PR #2382, opened 2024-10-17
Include examples in readthedocs

### LINKED ISSUE #2360: Experimental documentation
The documentation for Cell Space and DEVS in Experimental is duplicated. 

[page](https://mesa.readthedocs.io/stable/apis/experimental.html#experimental.__init__.Slider) 

Example of priority being duplicated. 
image

Happy to help if needs be!

quaquel:
This updates conf.py to create markdown files for including the examples in the documentation. It also cleans up some warning on RTD, which happened while debugging the code. In this process, this also fixes #2360.

It currently only covers the basic examples because the advanced examples are not fully cleaned yet.


quaquel:
Ok, its up an running: https://mesa--2382.org.readthedocs.build/2382/examples.html 

Refinement is off course possible. I have only included basic examples because those have been cleaned already.

EwoutH:
Starting to look interesting!

image

Let me know when it's ready for a first round of review.

quaquel:
Any feedback is welcome already. For example, should I keep the toc for the examples at 2 levels, or only 1 level (only the example names). I only used readme.md, agents.py, and model.py. What else should I include?

EwoutH:
Really awesome, this already helps a lot. Few minor points:

I would either put the Readme headers under a "Description" section, and let the Agent and Model sections jump in, and maybe also not display the Readme chapters at all (since they will be very similar).

image

Secondly, maybe add `app.py` at the end, which contains the visualisation code.

quaquel:
I have made several updates
1. I now only include the name of the example in the overview TOC. There is, in my view, little value in having level 2 headings (e.g., summary, Agents, Model, etc.) here. 
2. I added app.py
3. I haven't included a generic part in the overarching TOC. This is actually quite tricky because I cannot currently parse the generic readme for all examples. This readme contains links that are valid on GitHub but these are I think not valid on the docs (I can test this just to see what happens...)

The main things that I still want to do

1. Only recreate the `docs/example/<example>.md` file if any of the input files has a timestamp that is newer than the timestamp on the existing `docs/example/<example>.md` file. This ensures automatic updating of the example docs if the underlying example is updated.
2. Have some more elegant error messaging, rather than just failing to build the docs. 
3. Clean up and slightly reorganize the code in conf.py.

docs/index.md
```diff
@@ -88,7 +88,7 @@ ABM features users have shared that you may want to use in your model
 
 Mesa Overview <overview>
 tutorials/intro_tutorial
-tutorials/visualization_tutorial
+Examples <examples>
```
tpike3, line 91: Don't we still need the visualization on the index as well? (I am adding an seminal examples link to the intro

mesa/examples/basic/conways_game_of_life/app.py
```diff
@@ -0,0 +1,71 @@
+import time
+
+import altair as alt
+import numpy as np
+import pandas as pd
+import streamlit as st
```
tpike3, line 6: Do we still want to use streamlit or update to Solara?

mesa/examples/basic/virus_on_network/Readme.md
```diff
@@ -0,0 +1,61 @@
...
+```
+    $ pip install -r requirements.txt
+```
+
+## How to Run
+
+To run the model interactively, run ``mesa runserver`` in this directory. e.g.
+
+```
+    $ mesa runserver
```
tpike3, line 30: Update to Solara?

docs/examples/virus_on_network.md
```diff
@@ -0,0 +1,384 @@
...
+
+The SIR model is one of the simplest compartmental models, and many models are derivatives of this basic form. The model consists of three compartments:
+
+S: The number of susceptible individuals. When a susceptible and an infectious individual come into "infectious contact", the susceptible individual contracts the disease and transitions to the infectious compartment.
+I: The number of infectious individuals. These are individuals who have been infected and are capable of infecting susceptible individuals.
+R for the number of removed (and immune) or deceased individuals. These are individuals who have been infected and have either recovered from the disease and entered the removed compartment, or died. It is assumed that the number of deaths is negligible with respect to the total population. This compartment may also be called "recovered" or "resistant".
+
+For more information about this model, read the NetLogo's web page: http://ccl.northwestern.edu/netlogo/models/VirusonaNetwork.
+
+JavaScript library used in this example to render the network: [d3.js](https://d3js.org/).
```
tpike3, line 16: This could be removed

mesa/examples/basic/virus_on_network/Readme.md
```diff
@@ -0,0 +1,61 @@
+# Virus on a Network
```
tpike3, line 1: Why are these in two places?

tpike3 (changes requested):
This is awesome!

This also epitomizes to me why we want those seminal models in the main repo so we can ensure they are current with whatever Mesa version we are on.

docs/index.md
quaquel: This is a left over from the problems building the docs with the visualization tutorial errorring out. So yes, you are right.

mesa/examples/basic/virus_on_network/Readme.md
quaquel: not sure what you mean?

quaquel:
@tpike3, I agree with most of your comments regarding the content of the examples, but can we leave that for a separate PR and focus this one on just including the examples in the docs in the first place? The updated conf.py ensures that any updates in the readme, agent.py, model.py, and app.py result in an updated page in the docs.

EwoutH:
One question, why does it add so much additional files? Aren't all/most of those files in the repo already?

quaquel:
1. It updates conf.py, 6 associated autogenerated example markdown files in docs, and 2 templates to create these markdown files (9 files in total)
2. It renames the readme files to all have the same format (case sensitive) and makes similar name changes to some other files. 
3. I updated some files here and there in the docs because there are many warnings on RTD. These changes removed most of those allowing me to better debug the updates to conf.py.

EwoutH:
Thanks for the explanation. I will be honest, I don't like these monster PRs that do multiple things and are quite far away from an atomic change. There's a big chance something slips up in there. I also do recognize real-world development flows and practical limitations.

> 6 associated autogenerated example markdown files in docs

Is there anyway we can not keep these files in our history, since they are autogenerated from existing content and thus 100% duplicate. I suspect this will trigger (huge) CI issues and changesets when updating examples or working on other stuff.

I see two options:
1. Build then in Sphinx / RTD and don't generated them in git at all
2. (automatically) add them to the `.gitignore`.

For me 1 is preferred, but 2 is also acceptable. 

The other files are fine, renames and cleanups look good. Ideally the warnings cleanup would be a separate PR, but I do recognize how this goes so for now its fine from my part.

quaquel:
> Is there anyway we can not keep these files in our history, since they are autogenerated from existing content and thus 100% duplicate. I suspect this will trigger (huge) CI issues and changesets when updating examples or working on other stuff.

conf.py handles all this so there should not be a problem. I prefer to have them in the history because otherwise rtd has to first regenerate the md files and then regenerate the html pages. This speeds up the building of docs because they are only recreated by RTD if the associated example python files are updated.

EwoutH:
But that means every time I update an example, the all the changes will be duplicate into the PR diff. I find that to be a major downside.

Can't we cache them elsewhere, and just hash the source files and templates use to check for cache hits?

Sorry to make this PR more complicated, but this duplication is something I really really would like to avoid.

quaquel:
> But that means every time I update an example, the all the changes will be duplicate into the PR diff. I find that to be a major downside.

This only happen if you build the docs yourself locally and commit those changes. Also there is a single md file per example, so I am not sure what the problem is.

quaquel:
> Can't we cache them elsewhere, and just hash the source files and templates use to check for cache hits?

The simpler solution would be to just always rebuild the md files and ignore them via gitignore, but I fail to see what the problem is with the current solution as explained above.

EwoutH:
> The simpler solution would be to just always rebuild the md files and ignore them via gitignore

That's a totally acceptable solution for me right now. We can add caching later if that seems to be needed.

> I fail to see what the problem is with the current solution as explained above.

Basically, each change to the examples would have a completely duplicate diff, as both the example itself and generated readme is changed in exactly the same way.

quaquel:
> Basically, each change to the examples would have a completely duplicate diff, as both the example itself and generated readme is changed in exactly the same way.

I guess you mean markdown rather than readme. As indicated, this only happens if you build the docs locally. Who does that other than when developing and testing doc-related things? Note that this approach has been used for the workbench for several years. Have you ever run into this potential diff problem there?

I am happy to change it but I still fail to see the problem here.

EwoutH:
But if these files are tracked/included in git, but not updated in a PR, when are they updated? And why does this PR have them?

quaquel:
> But if these files are tracked/included in git, but not updated in a PR, when are they updated? And why does this PR have them?

If not updated in the PR, `conf.py` during building the docs on RTD will overwrite the existing files (depending on timestamps, basically if agent.py, model.py, or readme.md is newer than the exising <example_name>.md file, the md file will be overwritten). They are here because I tested conf.py locally.

mesa/examples/basic/virus_on_network/Readme.md
EwoutH: The old one wasn’t removed, so there are now two Readme files.

EwoutH:
> They are here because I tested conf.py locally.

Okay, so they are only generated within Sphinx / Readthedocs, and not tracked in git by default. So it’s already like the preferred solution I proposed.

When committing, please check which files you want to include in the PR and which files are just local because you tested / experimented with some things.

Two remaining points, then we can properly review:
- There are two Readme files that are now duplicated. I think instead of rename you did copy and rename.
- The same happened to the `st_app.py` I think. I think it would be good to keep the different name for the Streamlit app, since then users recognize it’s different. We might add a SolaraViz in the future called `app.py`, and then it would be nice to include both. I would say: Check for both app.py and st_app.py if they exist, and if they do, include them.

quaquel:
It's a bit more complicated than you make it out to be. But I don't want to continue arguing the point. 

I don't see two readme files that are identical, neither in my IDE nor in my file explorer. Please can you be explicit which files according to you are duplicated?

Including both app and st_app if they both exist and only either one, if they exist, is tricky with string templates. I actually experimented with trying to get that to work but it did not render nicely. So happy to take yet another stab at it but it's not entirely trivial as far as I can tell. 

If there is any file duplication, it happened after rebasing because of the exampels --> mesa.examples move. This rebase broke somehow the GitHub Desktop and I had to resort to the command line.

quaquel:
go ahead

mesa/examples/basic/conways_game_of_life/app.py
quaquel: in my view, we standardize all examples on solara

EwoutH:
I tried doing a proper rebase, but the git threw a few errors I never have seem before. So I ported the changes over manually in https://github.com/projectmesa/mesa/pull/2392.

PR #2383, opened 2024-10-18
viz_tutorial: Fix BoltzmannWealthModel import

EwoutH:
The currently working way to import examples seems to be:

```Python
from examples.basic import BoltzmannWealthModel
```

Not sure if we want to have it this way on the long term, since it could cause merge conflicts etc.

Rendered docs: https://mesa--2383.org.readthedocs.build/2383/tutorials/visualization_tutorial.html

### DIFF
--- a/docs/tutorials/visualization_tutorial.ipynb
+++ b/docs/tutorials/visualization_tutorial.ipynb
@@ -48,7 +48,7 @@
     "\n",
     "# You can either define the BoltzmannWealthModel (aka MoneyModel) or install it\n",
     "\n",
-    "from mesa.examples.basic import BoltzmannWealthModel"
+    "from examples.basic import BoltzmannWealthModel"
    ]
   },
   {
--- a/docs/tutorials/visualization_tutorial.ipynb
+++ b/docs/tutorials/visualization_tutorial.ipynb
@@ -87,7 +87,7 @@
    "outputs": [],
    "source": [
     "model_params = {\n",
-    "    \"N\": {\n",
+    "    \"n\": {\n",
     "        \"type\": \"SliderInt\",\n",
     "        \"value\": 50,\n",
     "        \"label\": \"Number of agents:\",\n",


quaquel:
Strange. I tried exactly this import in my branch and it errorred out. I have to see what's going on here.

EwoutH:
I suspect local imports when cloned from git behave differently then when it's installed using pip (like in the tutorial).

Likely due to this addition in the `pyproject.toml`:
https://github.com/projectmesa/mesa/blob/d2df9d47a15d0e73d3fccc9ffbbde12cd5fdfb1f/pyproject.toml#L89-L90

CC @Corvince

Corvince:
Strange indeed, but I can confirm that when installed from pypi `import mesa.examples.basic` does not work. 

I don't have the time the coming days to look into this closely, although there is probably something to learn here. The easiest solution would be to move the examples into the mesa folder. Would there be a downside to that?

Corvince:
> Likely due to this addition in the `pyproject.toml`:
> 
> https://github.com/projectmesa/mesa/blob/d2df9d47a15d0e73d3fccc9ffbbde12cd5fdfb1f/pyproject.toml#L89-L90
> 
> CC @Corvince

Yes, that works different then what I expected. Through pypi this installs now both a package "mesa" and "examples". I thought it would just make both folders available inside "Mesa".

quaquel:
> I don't have the time the coming days to look into this closely, although there is probably something to learn here. The easiest solution would be to move the examples into the mesa folder. Would there be a downside to that?

I don't think there is any major downside to it. I am not sure why @EwoutH did not move them there initially. I recall a brief exchange on this somewhere but I can't find it atm.

EwoutH:
Honestly didn't thought that much about it. Also nobody noted it in the discussions/PRs before merging.

I'm fine with them being moved if that makes things easier.

quaquel:
Either I misremember or I might have made a remark about it while discussing the move in person. It does not really matter. 

I think that moving them into mesa.examples is the best option for now. It resolves the weird input stuff and makes our life easier. I can do a PR if you want. It would be good to address this before merging some of the other PRs building on the examples being in the main repo.

quaquel:
I guess this is now superseded by #2387 and so can be closed?

Corvince:
Yes, although I have no idea why and how it works now on RTD, since the changes in #2387 are not yet released. But it does, so we have a working CI again! 🎉

quaquel:
As far as I can tell RTD will install the branch. Next, we also install via pip the latest from pip (so 3.0b2). My hunch is that the branch appears earlier in sys.path and so that gets used.

EwoutH:
```Python
%pip install --quiet --upgrade --pre mesa
```
Interesting, without the `--upgrade` I would have expected the first version to get used, but with the `--upgrade` it should overwrite it.

Enfin, it works.

quaquel:
> Interesting, without the --upgrade I would have expected the first version to get used, but with the --upgrade it should overwrite it.

Depends on what RTD exactly does. My hunch is that RTD simply clones the repo rather than using pip to install the repo. In that case you can have 2 versions of mesa: the cloned repo and the installed library in site-packages.

PR #2386, opened 2024-10-18
Add support for drawing discrete grids

### LINKED ISSUE #2377: Visualize Cell Space
Now that the experimental Cell Space is developing further, and all the [advanced examples](https://github.com/projectmesa/mesa/tree/main/examples/advanced) are using it, we should start visualizing it in SolaraViz.

quaquel:
Add support for drawing experimental Grid Spaces (i.e. HexGrid, OrthogonalMooreGrid, OrthogonalVonNeumannGrid). 

Closes #2377. This code needs further refinement (e.g., duplication of agent potrayal code from voroinoi grid, offsets for hexgrids, and adding experiment NetworkGrids).


EwoutH:
Uninstall your current mesa, `pip install . -e` and then `solara run app.py` is what I found out working.

quaquel:
> Uninstall your current mesa, pip install . -e and then solara run app.py is what I found out working.

Thanks! This gave me the idea for a slightly different approach: insert the correct version of mesa into sys.path at the top. Based on this, I can confirm that this code works as intended. 

Let me know what else would be needed to complete this PR since its my first contribution to the visualization side of things.

EwoutH:
> insert the correct version of mesa into sys.path at the top. Based on this, I can confirm that this code works as intended.

That's the other obvious way to go, both have their advantages based on what your working on.

@Corvince I'm also quite new to this stack, could you please review this one?

mesa/visualization/components/matplotlib.py
```diff
@@ -11,7 +11,7 @@
 from matplotlib.figure import Figure
 
 import mesa
-from mesa.experimental.cell_space import VoronoiGrid
+from mesa.experimental.cell_space import OrthogonalMooreGrid, VoronoiGrid
```
Corvince, line 14: This should also work for OrthogonalVonNeumannGrid

Corvince (approved):
I think this PR is very fine and a good first step. I pre-approve this, just please also add OrthogonalVonNeumann

mesa/visualization/components/matplotlib.py
quaquel: yes, I'll fix that! Just need to figure out how to capture this in the match case statement. 

I guess it even will work for hexgrids (although we might want to visualize these a bit differently with offsets for every odd row.

mesa/visualization/components/matplotlib.py
```diff
@@ -52,16 +52,20 @@ def SpaceMatplotlib(
     if space is None:
         space = getattr(model, "space", None)
 
-    if isinstance(space, mesa.space._Grid):
-        _draw_grid(space, space_ax, agent_portrayal, propertylayer_portrayal, model)
```
wang-boyu, line 56: Hi @quaquel, I understand that this is already merged. May I ask whether it is intentional to remove the call to `_draw_grid()` here? Thanks!

quaquel: I guess I made a mistake. I'll put in a PR to fix this asap.

Thanks for the post-merge review!

quaquel: and fixed via #2398

wang-boyu: Great! Thanks!

EwoutH:
One improvement for this function would be to use the `_get_agent_data` when drawing the new grids. I got a weird error, and using `_get_agent_data` would have prevented and handled that. I wrote it up here:
- #2407

quaquel:
> One improvement for this function would be to use the _get_agent_data when drawing the new grids. I got a weird error, and using _get_agent_data would have prevented and handled that

I looked at that and I could not use it because that method is hard coded to only work with `_Grid`: `for agents, pos in space.coord_iter()`.

PR #2387, opened 2024-10-18
Move examples into mesa

Corvince:
Based on previous discussion and to rectify some errors made in #2381 

See commits for atomic changes:
1. move examples into mesa folder - the main purpose of this PR
2. remove the "examples" package which was a mistake of #2381 
3. Update paths
4. Add an empty `__init__.py` file to each basic example. While I find this a bit annoying, I think its actually the best way to make the examples truly portable, combined with relative imports. Each example becomes its own "package" in python speak. Also updated the imports so now you can do `from mesa.examples import BoltzmannWealthModel` instead of `from mesa.examples.basic import BoltzmannWealthModel`


quaquel (approved):
The move makes sense and should resolve the various issues we have been running into. Thanks!

EwoutH (commented):
Good stuff, thanks.

Git history is maintained remarkable well!

image

Corvince:
Yes, luckily for git its just a rename operation, since git doesn't know about the concept of folders.

PR #2390, opened 2024-10-19
Explicitly test basic examples

quaquel:
This is a quick update to test_examples.py for testing the basic examples.

Presently, it only tests if the models run, no further assertions are made about e.g., results. So the main purpose of this type of test is integration testing.


EwoutH (approved):
Huge code reduction, same functionality. Love it!

PR #2397, opened 2024-10-21
Altair spaces

quaquel:
This adds altair support for new-style grids (moore, von neumann and hex), as well as old-style continuous spaces


EwoutH (approved):
Looks more robust than we currently do it, and no user API breaking changes

PR #2399, opened 2024-10-22
fixes weakref bug in shuffe_do

quaquel:
if one agent removes another agent, it could still be activated by shuffle_do. This adds a unit test to spot the bug and contains the fix.

# longer explanation
shuffle_do used `agents = list(self._agents.keys())`. This returns a list of hardrefs to all agents. So if `agents[0]` removes `agents[1]`, it will be removed from the model, but it is still activated in `shuffle_do`. we had the same bug before in shuffle in january.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -294,15 +294,17 @@ class AgentSet(MutableSet, Sequence):
 
         It's a fast, optimized version of calling shuffle() followed by do().
         """
-        agents = list(self._agents.keys())
-        self.random.shuffle(agents)
+        weakrefs = list(self._agents.keyrefs())
+        self.random.shuffle(weakrefs)
 
         if isinstance(method, str):
-            for agent in agents:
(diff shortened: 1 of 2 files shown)


EwoutH:
Right, this happens because an Agent can get removed within a step.

Have you done any benchmarks by any chance?

quaquel:
> Right, this happens because an Agent can get removed within a step.

Yes exactly. I ran into it with a wolf sheep version where I simply used self.agents.shuffle_do, rather than activate by class. The test no captures the problem.

> Have you done any benchmarks by any chance?

Not beyond the default mesa benchmarks. Feel free to rerun your original tests. My hunch is that the performance after the fix will be marginally slower but not much than before the bugfix. The delta with chaining should however still persist because  you iterate over the agents only once instead of twice.

EwoutH (approved):
I’m going to approve, slow is better than buggy. Performance can be checked and optimized later.

Thanks!

PR #2400, opened 2024-10-23
Pass through model.rgn in agent analogous to model.random

quaquel:
This is a follow up PR from #2352. It adds a new property to Agent: `Agent.rng`. Implementation wise it is identical to `Agent.random.` This PR also explicitly adds tests for both properties.


EwoutH (approved):
Very clean and consistent with the Model implementation

PR #2402, opened 2024-10-23
Make example import absolute

quaquel:
As discussed in #2381, the current relative imports with the examples create a problem. This at least makes things run again by using absolute imports.


Corvince (approved):
Thanks for quickly doing this

EwoutH:
Awesome, let's validate that this works in all our required cases and I will get a new pre-release out tonight.

PR #2406, opened 2024-10-23
Update Schelling Readme.md

quaquel:
This updates the Schelling readme. It has the correct file names for all files. It also removes the text on requirements.txt since this file is not included.

### DIFF
--- a/mesa/examples/basic/schelling/Readme.md
+++ b/mesa/examples/basic/schelling/Readme.md
@@ -25,7 +25,7 @@ To run the model with the grid displayed as an ASCII text, run `python run_ascii
 ## Files
 
 * ``model.py``: Contains the Schelling model class
-*  ``agent.py``: Contains the Schelling agent class
+*  ``agents.py``: Contains the Schelling agent class
 * ``app.py``: Code for the interactive visualization.
 * ``analysis.ipynb``: Notebook demonstrating how to run experiments and parameter sweeps on the model.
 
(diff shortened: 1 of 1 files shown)


jackiekazil:
This works. We have a pattern of having an installation section in the readme though -- I think we should keep that pattern. 
Someone might not know they need to add mesa to their environment -- they might try to pip install from pypi, which wouldn't work with the example necessarily.  I think we should have this in all the examples. Thoughts?

quaquel:
> This works. We have a pattern of having an installation section in the readme though -- I think we should keep that pattern.
Someone might not know they need to add mesa to their environment -- they might try to pip install from pypi, which wouldn't work with the example necessarily. I think we should have this in all the examples. Thoughts?

My concern with adding it to the readme's for all examples is the same as my concern with requirements.txt in all examples. It results in duplication of text, which is easy to miss and annoying to maintain. I agree, however, that it would be good to be clear somewhere that these examples ship with mesa and require mesa to be installed. 

Looking at the current docs, I noticed that [Using Mesa](https://mesa.readthedocs.io/latest/#using-mesa) on the landing page covers installing mesa, but also showcases how to run an example. First, this example is outdated and needs to be removed. Might it be a solution to update this text using one of the basic examples? We could then add a link here also to the TOC for all examples that ship with mesa. 

Alternatively, what about including some additional text in the [TOC](https://mesa.readthedocs.io/latest/examples.html) for examples? It would be easy to add a paragraph at the top with installation instructions and run instructions that is valid for all examples.

mesa/examples/basic/schelling/Readme.md
```diff
@@ -32,8 +24,9 @@ To run the model with the grid displayed as an ASCII text, run `python run_ascii
 
 ## Files
 
+* ``model.py``: Contains the Schelling model class
+*  ``agent.py``: Contains the Schelling agent class
```
EwoutH, line 28: I started the convention of `agents.py` instead of `agent.py`:
```suggestion
*  ``agents.py``: Contains the Schelling agent class
```

EwoutH (approved):
One minor nitpick, otherwise good!

quaquel:
@jackiekazil I am merging these PRs and will add your point on installation instructions to the tracking issue. My current inclination is to address it at the level of the TOC for all examples. This page needs work anyway because it should be tied to the readme at the root of the examples folder.

PR #2413, opened 2024-10-24
add advanced examples to rtd

quaquel:
This updates conf.py so the now cleaned examples are included in read the docs


EwoutH (approved):
Can we add subsections for the Basic and Advanced examples? Maybe with single-line explanations that basic examples only contain stable features, while the advanced ones also include experimental features.

image

quaquel:
> Can we add subsections for the Basic and Advanced examples? Maybe with single-line explanations that basic examples only contain stable features, while the advanced ones also include experimental features.

I want to get on that next, ideally by included the root readme file as well but that will be its own pr

EwoutH:
You do get the gist of atomic PRs!

PR #2417, opened 2024-10-25
Update docstring of SimEvent

quaquel:
In response to [this feedback on the beta](https://github.com/projectmesa/mesa/discussions/2338#discussioncomment-11044515), this PR adds a notes to SimEvent to explicitly point out that using lambda's in `SimEvent` won't work.

### DIFF
--- a/mesa/experimental/devs/eventlist.py
+++ b/mesa/experimental/devs/eventlist.py
@@ -33,6 +33,12 @@ class SimulationEvent:
         function_args (list[Any]): Argument for the function
         function_kwargs (Dict[str, Any]): Keyword arguments for the function
 
+
+    Notes:
+        simulation events use a weak reference to the callable. Therefore, you cannot pass a lamda function in fn.
+        A simulation event where the callable no longer exists (e.g., because the agent has been removed from the model)
+        will fail silently.
+
     """
 
     _ids = itertools.count()


EwoutH:
Thanks! @jtcooper10, would this solve your point raised?

EwoutH (approved):

Corvince:
Just out of curiosity, why does it use a weak reference at all?

jtcooper10:
> Thanks! @jtcooper10, would this solve your point raised?

I think so! The note being here will help to save from any future head-scratching about lambdas.

quaquel:
> Just out of curiosity, why does it use a weak reference at all?

Imagine an event is scheduled to invoke `some_agent.method`, but this agent is, for some other reason, removed from the simulation. If we don't use a weakref, the agent cannot be garbage collected. There is a SimEvent.cancel method that could be used if `some_agent.method` was scheduled by some_agent itself. Basically, you could track all events some_agent has scheduled and cancel all of those. However, `some_agent.method` can also be scheduled by other objects. How are they to be informed of the fact that some_agent is being removed from the simulation and that therefore the events has to be canceled? All these problems are avoided by using a weakref in SimEvent.

Corvince:
Ah, yes of course! Thanks for detailing it out, made it very clear

PR #2420, opened 2024-10-25
space: Add note that Grids are maintenance only

EwoutH:
Add a note to the space module level docstring that the Grids are maintenance only.

Also add the PropertyLayer to the classes and remove the Grid, since that's now private (and has been for a while).

Now looks like this:
image

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -3,12 +3,12 @@
 Objects used to add a spatial component to a model.
 
 .. note::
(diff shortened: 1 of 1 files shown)


tpike3 (approved):
LGTM

quaquel:
indeed go ahead and merge it. 

One thing I noted but can be done in a separate: remove HexGrid from mesa.space. It has been deprecated already so it can be safely removed.

quaquel (approved):

Issue #2423, opened 2024-10-26
example: Update EpsteinCivilViolence to use OrthogonalVonNeumann


EwoutH:
The OrthogonalMooreGrid make the behavior remarkably stable, @quaquel found out updating it to OrthogonalVonNeumann helps get dynamic/interesting behavior back.

See https://github.com/projectmesa/mesa/issues/2408.

PR #2429, opened 2024-10-27
Updates to Epstein example

quaquel:
This is an update to the Epstein model to address #2423. Next to shifting to a von Neumann grid, there are several other changes. 

1. A minor change in the arrest_probability calculation by using `round`. This is based on an extensive literature that shows that without it, the original Epstein results cannot be reproduced.
2. Shifting to using an enum for the state of the citizen (this has various knock-on consequences elsewhere in the code that make the code cleaner and shorter).
3. Rework of app.py to only use stuff that actually works when drawing new-style discrete grids. 

The net result is shown below.

Screenshot 2024-10-27 at 10 36 26

### DIFF
--- a/mesa/examples/advanced/epstein_civil_violence/app.py
+++ b/mesa/examples/advanced/epstein_civil_violence/app.py
@@ -16,8 +16,9 @@ COP_COLOR = "#000000"
 agent_colors = {
     CitizenState.ACTIVE: "#FE6100",
     CitizenState.QUIET: "#648FFF",
-    CitizenState.ARRESTED: "#808080"
+    CitizenState.ARRESTED: "#808080",
 }
+
 
 def citizen_cop_portrayal(agent):
     if agent is None:
--- a/mesa/examples/advanced/epstein_civil_violence/app.py
+++ b/mesa/examples/advanced/epstein_civil_violence/app.py
@@ -12,10 +12,12 @@ from mesa.visualization import (
 )
 
 COP_COLOR = "#000000"
-QUIET_COLOR = "#648FFF"
-ACTIVE_COLOR = "#FE6100"
-ARRESTED_COLOR = "#808080"
 
+agent_colors = {
+    CitizenState.ACTIVE: "#FE6100",
+    CitizenState.QUIET: "#648FFF",
+    CitizenState.ARRESTED: "#808080"
+}
 
 def citizen_cop_portrayal(agent):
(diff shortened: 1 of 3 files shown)


mesa/examples/advanced/epstein_civil_violence/app.py
```diff
@@ -8,10 +12,9 @@
 )
 
 COP_COLOR = "#000000"
-AGENT_QUIET_COLOR = "#648FFF"
-AGENT_REBEL_COLOR = "#FE6100"
-JAIL_COLOR = "#808080"
-JAIL_SHAPE = "rect"
+QUIET_COLOR = "#648FFF"
+ACTIVE_COLOR = "#FE6100"
+ARRESTED_COLOR = "#808080"
```
EwoutH, line 17: I would make this a dict
```Python
agent_colors = {
    CitizenState.ACTIVE: "#FE6100",
    CitizenState.QUIET: "#648FFF",
    CitizenState.ARRESTED: "#808080"
}
```
then you can do fast lookups instead of matching (which is quite slow for every agent for every step).

mesa/examples/advanced/epstein_civil_violence/model.py
```diff
@@ -45,102 +50,65 @@ def __init__(
...
+        if cop_density + citizen_density > 1:
             raise ValueError("Cop density + citizen density must be less than 1")
 
         for cell in self.grid.all_cells:
-            if self.random.random() < self.cop_density:
-                cop = Cop(self, vision=self.cop_vision)
-                cop.move_to(cell)
+            klass = self.random.choices(
+                [Citizen, Cop, None],
+                cum_weights=[citizen_density, citizen_density + cop_density, 1],
```
EwoutH, line 78: Seems a bit complicated. Can't we just generate cop_density * cells cops and citizen_density * cells citizens, and place them randomly?

Or aren't they allowed to start at the same cell?

EwoutH (approved):
Few minor comments, generally seems like a nice improvement.

Have you any idea about performance?

mesa/examples/advanced/epstein_civil_violence/model.py
quaquel: There can be at most one agent per cell.

quaquel:
> Have you any idea about performance?

I haven't done any testing, this was really focussed on getting the proper behavior back.

Issue #2401, opened 2024-10-23
Refactoring mesa.visualization.components.matplotlib


quaquel:
I have taken a closer look at the solara matplotlib visualization support in MESA and tried to refactor it a bit because of various redundancies. However, I ran into a few issues. Basically, how the data is gathered before plotting is dependent on the type of space that is to be visualized. For example, there is no relevant position data when visualizing networks. Offsets to x and y are added for grids. Moreover, for discrete spaces, we have multiple implementations of both old-style and new-style (e.g., HexSingleGrid, HexMultiGrid, HexGrid). I believe it is quite easy to abstract away whether you are plotting an old-style or a new-style space. They mainly differ with respect to how you get the agents to plot and their (x, y) coordinates, which is easily wrapped into custom generators (these generators are probably also useful for the altair side of things, so I'll probably add a components.util module). 

In an attempt to make sense of this all, I have come to the following breakdown:

```python
OrthogonalGrid = SingleGrid | MultiGrid | OrthogonalMooreGrid | OrthogonalVonNeumannGrid
HexGrid = HexSingleGrid | HexMultiGrid | HexGrid
ContinousSpace
Network = NetworkGrid | mesa.experimental.grid_space.Network 
VoronoiGrid

draw_orthogonal_grid(space: OrthogonalGrid, agent_portrayal: Callable, propertylayer_portrayal: Callable)
    ...

draw_hex_grid(space: HexGrid, agent_portrayal: Callable, propertylayer_portrayal: Callable):
    ...

draw_network(space: Network, agent_portrayal: Callable, propertylayer_portrayal: Callable):
    ...

draw_continuous_space(space: ContinousSpace, agent_portrayal: Callable):
    ...


draw_voroinoi_grid(space: VoronoiGrid, agent_portrayal: Callable, propertylayer_portrayal: Callable):
    ...

```

I have several questions:
1. Does the breakdown into types of spaces make sense?
2. Did I miss anything?
3. Property layers are currently implemented in `DiscreteSpace` for new-style spaces. So they are available in all new style spaces, including `Network` and `VoronoiGrid`.  Should property grids be moved from `DiscreteSpace` to `Grid`? This would mimic old-style spaces. I also question whether property layers are presently meaningful for networks or Voronoi meshes. (I want to focus the conversation on improving the current visualization code, so please don't bring up ideas on how to expand property layers to also work in e.g., VoronoiGrid or continuous spaces). 

[...]

Issue #2389, opened 2024-10-19
Giving the user more control in solara visualization of spaces


quaquel:
The current signature for displaying a space is 

```Python
def make_space_matplotlib(agent_portrayal=None, propertylayer_portrayal=None):
```

This gives the user explicit control over the `agent_portraya`l and the `propertylayer_portrayal`. However, it also means that the mesa code is making implicit assumptions regarding the attribute to which the space is assigned in the model, and how (depending on the used space class) it is to be visualized. Moreover, it makes space visualization not easily extendable by the user. So, I suggest changing the signature to

```Python
def make_space_matplotlib(agent_portrayal=None, propertylayer_portrayal=None, space_portrayal=None, space:str =None):
```

If `space_portrayal ` is None, we can use the existing` if elif` structure in `SpaceMatplotlib` to identify the correct space_portrayal function (probably moved into a separate helper function). Likewise, if `space` is None, we can fall back on checking `model.grid` and `model.space`. This same more fine grained API can also be applied to `make_space_altair`. @Corvince, @EwoutH, any thoughts?

EwoutH:
Definitely a step in the right direction, but can we go further? All this passing of strings and dicts doesn't seem proper OOP.

Otherwise I guess you just pass configuration data and a space to apply it to, so this is the proper way to do things?

quaquel:
I agree that this is not yet the desirable API. Ideally, it becomes declarative in some sense, but some further development is needed under the hood to get to that point. For now, passing dicts and callables is a step in the right direction.

Corvince:
Yes this is something we need to address before the 3.0 release I think. I think the API for SolaraViz itself is good (although we still want to rename it), the API for the components and for creating the components is not. Either we find a good solution or we should mark that part experimental/unstable. 

Actually that's also a point for going monorepo, because I can imagine the frontend developing at a different pace than core mesa. So having a dedicated viz package would let us iterate faster without breaking mesa version semantica. But this is another topic.

Corvince:
I like a lot of things about this API. However, I don't think we should go done this path. While the API is very altair-like, its probably unfamiliar for most python users. I would like to have an API that makes 2 things easy:

1. Have a basic space plot
2. Have full control over the plotting mechanism 

1. Should in the extreme case be as simple as `make_space_plot(model)`. That should be enough to have a simple idea of how the space looks like. Additionally we can provide some whats (what should determine the color, etc.), but  no hows.

[...]

Issue #2243, opened 2024-08-23
Customize network layout, currently only spring layout


rmhopkins4:
Currently, without implementing a custom drawer, networks are always drawn with the spring layout. A way to customize this would be well appreciated, whether it be in the `agent_portrayal` or in the model itself (which does somewhat violate the separation of model & visualization).

```
## matplotlib.py

# draws using networkx's matplotlib integration
def _draw_network_grid(space, space_ax, agent_portrayal):
    graph = space.G
    pos = nx.spring_layout(graph, seed=0)
    nx.draw(
        graph,
        ax=space_ax,
        pos=pos,
        **agent_portrayal(graph),
    )
```

rht:
What if there is an arbitrary `kwargs` that you can pass to `nx.draw`, `ax.plot`, and so on, to expose their full API?

rht:
Oh, the layout has to be initialized separately from `nx.draw`, so this needs a separate kwarg.

zjost:
I got around this by just creating a custom viz: https://github.com/projectmesa/mesa-examples/blob/main/examples/aco_tsp/app.py#L39

PR #2430, opened 2024-10-27
Matplotlib refactor

quaquel:
This refactors the existing matplotlib plotting code in line with #2401. Note that this refactor is useful regardless of how #2389 is resolved. This makes what is there better, cleaner, and complete for all spaces without changing the user-facing API.  The return of `agent_potrayal` is now the same for all spaces because they all plot agents with matplotlib's scatter. 

In short, it is built around dedicated plotting functions for different types of spaces while abstracting away the difference between old-style and new-style discrete spaces. These plotting functions return an axes instance, making it easier to test and use each plotting function separately from Solara.

Any field returned by agent_portrayal that is not used results in an a user warning. 

I also changed the behavior of orthogonal grids a bit. The current code adds an offset to the x and y coordinates, while the grid lines are drawn on integer values. Instead, I change the view limits and draw the grid lines on 0.5, 1.5, etc. I believe this better reflects the desired behavior (i.e., the agent on the coordinate (1,1) is shown centered (1, 1) while the grid lines nicely go around it. 

Hexgrids were not supported before. This is now added although property layers are not yet shown correctly. All other stuff works.

Network plotting is refactored substantially. Users can specify which layout algorithm to use. The network edges are optional and drawn analogously to the grid in other discrete spaces. 

Tests for the individual plotting functions have been added. These only check if the code runs, not if the plots are correct. This has been checked visually.


mesa/visualization/components/matplotlib.py
```diff
@@ -46,31 +64,53 @@ def SpaceMatplotlib(
...
+            fig, ax = draw_orthogonal_grid(
+                space, agent_portrayal, propertylayer_portrayal, model
+            )
+        case OrthogonalMooreGrid():
+            fig, ax = draw_orthogonal_grid(
+                space, agent_portrayal, propertylayer_portrayal, model
+            )
+        case OrthogonalVonNeumannGrid():
+            fig, ax = draw_orthogonal_grid(
+                space, agent_portrayal, propertylayer_portrayal, model
```
EwoutH, line 84: Can't these all be grouped?  `case OrthogonalGrid():`, `case HexGrid():`, etc?

mesa/visualization/components/matplotlib.py
```diff
@@ -146,94 +186,180 @@ def draw_property_layers(ax, space, propertylayer_portrayal, model):
...
 
     # Draw grid lines
-    for x in range(space.width + 1):
-        space_ax.axvline(x, color="gray", linestyle=":")
-    for y in range(space.height + 1):
-        space_ax.axhline(y, color="gray", linestyle=":")
+    for x in np.arange(-0.5, space.width - 0.5, 1):
+        ax.axvline(x, color="gray", linestyle=":")
+    for y in np.arange(-0.5, space.height - 0.5, 1):
+        ax.axhline(y, color="gray", linestyle=":")
```
EwoutH, line 257: Not necessarily this PR, but I think we should add a boolean `draw_gridlines` or something like that to enable/disable this behavior.

Could also be `gridline_dict`, which if empty doesn't draw the gridlines, and otherwise takes a color, linestyle and width for example.

EwoutH (commented):
Lot's of interesting stuff in here!

> This makes what is there better, cleaner, and complete for all spaces without changing the user-facing API.

Nice!

> A debatable choice made here is that agent_portrayal must return keys that are valid for ax.scatter. That is 'c', 's' and 'marker'. I am not sure about this.

I would for now support the "old" values with a simple translation dict `{"color": "c, "Layer", "zorder", ...}`. Ideally we shouldn't have to modify our examples.

> I also changed the behavior of orthogonal grids a bit. The current code adds an offset to the x and y coordinates, while the grid lines are drawn on integer values. Instead, I change the view limits and draw the grid lines on 0.5, 1.5, etc. I believe this better reflects the desired behavior (i.e., the agent on the coordinate (1,1) is shown centered (1, 1) while the grid lines nicely go around it.

This is a difficult one. It will help maintaining smaller differences between continuous and discrete spaces (since 1, 1 is now identical to 1.0, 1.0), but it might be a bit inconvenient with grids, cell agents, PropertyLayers, etc. Have you checked those now correctly align?

> Hexgrids were not supported before. There is a draft implementation

Really awesome!

mesa/visualization/components/util.py
```diff
@@ -0,0 +1,122 @@
...
+    for _ in range(10):
+        agent = Agent(model)
+        pos = (
+            agent.random.random() * space.width,
+            agent.random.random() * space.height,
+        )
+        space.place_agent(agent, pos)
+
+    agents = list(get_agents_olstyle_space(space))
+    assert len(agents) == 10
```
EwoutH, line 122: Is this just test code or should this be in the PR?

mesa/visualization/components/matplotlib.py
quaquel: As also indicated in #2389, yes, we need to find a clean solution for both view limits, grid lines, and possible user-specific post plotting operations. I'll leave that out of this PR.

mesa/visualization/components/util.py
quaquel: the entire util.py became redundant with #2418. I'll remove the file.

mesa/visualization/components/matplotlib.py
quaquel: I don't know. I want to do some testing here to make it a bit cleaner.

quaquel:
Thanks for the feedback!

>  would for now support the "old" values with a simple translation dict {"color": "c, "Layer", "zorder", ...}. Ideally we shouldn't have to modify our examples.

So "Layer" and "Zorder" are already not supported in the current code, and so also not supported here. This is in fact one of the reasons why I am included to raise a value error for any element in the agent portrayal dict that is not supported. Errors should not be passed over in silence. 

Not sure about the dict. With 3.0, we can break backward compatability. I prefer a clear choice over maintaining 2 different labels for each valid element in the dict. I personally like the clear one-to-one relation between arguments for `ax.scatter` and what can go into the dict. 

> This is a difficult one. It will help maintaining smaller differences between continuous and discrete spaces (since 1, 1 is now identical to 1.0, 1.0), but it might be a bit inconvenient with grids, cell agents, PropertyLayers, etc. Have you checked those now correctly align?

I have tested this but not for property layers. I want to take a closer look at those anyway also because of hexgrids.

quaquel:
I added some first tests. These only make sure that the plotting code (so the `draw_x` functions) runs, not that the code is correct. I have used a notebook to check the resulting figures. Most of it looks good. 

Some quick updates:

1. I added width and height properties to voroinoi grid. This makes it possible to use the same scatter code as continuous and orthogonal spaces. I noticed various other issues with vorionoi grid that need their own PR. For example, there is a `cell_coloring_property` that has no business being there. Moreover, these grids are similar to continuous spaces in that they should have an xmin, xmax, ymin, ymax, next to width and height. I'll try to open a separate PR that only cleans up VoronoiGrid.  
2. The current network plotting code is a mess. Check the [agent_portrayal function in the virus_on_a_network example](https://mesa.readthedocs.io/latest/examples/virus_on_network.html#app). I will clean this up so that agent_portrayal behaves the same across all spaces. I already alluded to this idea in #2389.
3. Hexgrid plotting runs, but I still need to confirm it's actually correct. 
4. The current `mesa.space.NetworkGrid `is just horrible. The API is completely different from the other old-style spaces and very slow. I can't wait to deprecate this stuff in favor of `mesa.experimenta.cell_space.Network`.

quaquel:
Some further updates:

1. All mesa spaces are now being visualized. I have also visually confirmed that stuff is plotted correctly
2. Hexgrids come with a hex mesh
3. `agent_portrayal` now behaves consistently across all spaces
4. I reverted the changes in voronoi and solved the problem in a different way
5. `collect_agent_data` now allows for explicit control over the defaults
6. Network plotting now uses `ax.scatter` for the agents, and `nx.draw_edges` for the edges. The benefit is that agent_portroyal now works identically to other spaces. Moreover, it resolves various bugs we were unaware of (e.g., multiple agentse old visualization code) on a node would have crashed th. The drawback is that it's no longer possible to control how edges are visualized (as done in the virus_on_a_network example). 

Still to do:
1. Check property layer visualization and interaction with orthogonal grids and hex grids
2. Add some grid style drawing to networks
3. Figure out a clean solution to re-enable edge drawing for networks


Future extensions
1. Control over whether grids are drawn or not. This is relevant for orthogonal grids, hexgrids (and possibly networks)
2. Figure out a clean solution for user post processing of the figure and axes (e.g., to add a title, add a legend).

EwoutH:
Awesome work! Do you need any input at this point?

Corvince:
After a first glance and from reading your description this seems to be an excellent PR! I started modifying things in these functions here and there and was always too frustrated with the current state and quit. Hope this makes it more enjoyable to work with. 

Tomorrow I will have some time to do a code review. I'll focus on code, but is there something specific you want me to actually try and run?

--- 

But otherwise, please refrain from statements such as 

> 4\. The current `mesa.space.NetworkGrid `is just horrible

This is just insulting for the authors of the code. We value contributions from everyone, expert and novice programmers and the code specifically partly stems from a time where it was better to have something rather than nothing. I haven't looked at it now, and it probably indeed would need some refactoring to meet modern standards. But I don't recall a lot of issues with the NetworkGrid, so I can't imagine it being truly "horrible".

mesa/visualization/components/matplotlib.py
quaquel: So you can use `|` as an `or` operator, which cleans up the code considerably:

```python
    match space:
        case mesa.space._Grid() | OrthogonalMooreGrid() | OrthogonalVonNeumannGrid():
            draw_orthogonal_grid(space, agent_portrayal, ax)
        case HexSingleGrid() | HexMultiGrid() | mesa.experimental.cell_space.HexGrid():
            draw_hex_grid(space, agent_portrayal, ax)
        case mesa.space.NetworkGrid() | mesa.experimental.cell_space.Network():
            draw_network(space, agent_portrayal, ax)
        case mesa.space.ContinuousSpace():
            draw_continuous_space(space, agent_portrayal, ax)
        case VoronoiGrid():
            draw_voroinoi_grid(space, agent_portrayal, ax
```

quaquel:
@EwoutH 

In `propertylayer_portrayal` there is now the following snippet

```python
        if space and data.shape != (width, height):
            warnings.warn(
                f"Layer {layer_name} dimensions ({data.shape}) do not match space dimensions ({width}, {height}).",
                UserWarning,
                stacklevel=2,
            )
```

In my view, this does not belong in visualization code, because if this is not allowed, you have already created an invalid model. If you agree, we should open an issue about this so it can be added in the right place in `PropertyLayer`.

quaquel:
This is a quick figure showcasing how all spaces are now visualized. Note that all drawing functions take `ax` as an argument. This makes them also useful outside of solara components (It would be trivial for example to create an mp4 of a a simulation in this way). 

Some other minor updates: I added a draw_grid keyword argument to `draw_hex_grid`, `draw_orthogonal_grid`, and `draw_network` (treating the network itself as the grid seems sensible to me, but feedback is appreciated). I also made it possible to control the layouting algorithm used by `draw_network` via 2 optional keyword arguments. Neither the `draw_grid` or layouting algorithm kwargs are supported yet in `make_space_matplotlib`. I have to figure out a clean way of doing this. I am currently thinking of borrowing from matplotlib via a `draw_space_kwargs`. This would be a dict and you can use it to pass any additional keyword arguments to whichver `draw_x` function will be used. 

Unknown

quaquel:
This is ready for a proper review.

**open questions**
1. `get_agent_data` only uses `c`, `s`, and `marker`. This is different from the current code but consistent with `matplotlib.scatter`. @EwoutH  suggested keeping both the current labels and these new labels. I am not in favor. With MESA 3, we can start with a clean slate, so let's use it.
What do others think? 
2. What to do if fields are remaining in the `agent_portrayal` dict that are not used? Errors should not be passed over in silence. I see two options: explicitly raise an exception or issue a warning that field `x` is ignored.
3. How to handle the `draw_grid` and other `draw_x` related keyword arguments (e.g., for network layouting)?


**Future extensions**
1. drawing property layers for hex grids. It will be done already, but it will look weird because the coordinate systems don't align. The inner function `setup_hexmesh` can be adapted to solve this problem. This is a good issue to solve for 3.1.
2. Figure out a clean solution for user post-processing of the figure and axes (e.g., to add a title, add a legend). I am thinking of adding an optional `post_ploting: Callable|None = None` to `make_space_matplotlib` that will be called with the figure and axes. Users can then use this for any further tuning they want to do, like adding a legend, setting the aspect ratio of the figure, etc. 
3. Clean up VoroinoiGrid, remove the `cell_coloring_property` kwarg, and move support for it into here. In a way, coloring the mesh is like adding a property layer so this might interact with point 1 and be solvable analogously. 
4. Add support for `zorder` and possible other arguments in `get_agent_data` and `_scatter`. See #2341 and #2242 for further ideas. The challenge here is that some fields can be passed as a list (one entry per agent), such as color, while other arguments are valid for the entire collection of points (e.g., marker, zorder).

EwoutH:
Looks really amazing. And that's an understatement.

I'm still a bit focussed on my thesis paper, so I won't dive in the code, but I can share my thoughts on the high-level questions:

**Open questions**
1. I'm fine with replacing them, but 
   - with a proper warning for the current labels
   - well documented in the migration guide and in the tutorial
  (yes, realise this is extra work, but that's the obligation of making breaking changes in my perspective. breaking changes does not set us free from proper deprecation procedures. documentation can be a separate PR though).
2. I would go with an issue a warning that field x is ignored. Those are irritating and people often won't ignore them, but it also won't immediately break the whole model with a weird error. Feels like a good middle way.
   - Maybe the warnings could give specific suggestions if old labels are used, solving the first part of 1.
4. No immediate thoughts, except very maybe `make_space_matplotlib` shouldn't be a single function for all spaces. Having different `make_grid_space_matplotlib`, `make_network_space_matplotlib`, etc. with different (keyword) args could be an option.
5. Fine with follow-up PRs, what you prefer.

Thanks again for not only the extensive work but also the extensive communication around it. Appreciated!

mesa/visualization/components/matplotlib.py
```diff
@@ -146,132 +180,269 @@ def draw_property_layers(ax, space, propertylayer_portrayal, model):
             )
 
 
-def _draw_grid(space, space_ax, agent_portrayal, propertylayer_portrayal, model):
-    if propertylayer_portrayal:
-        draw_property_layers(space_ax, space, propertylayer_portrayal, model)
+def collect_agent_data(
+    space: OrthogonalGrid | HexGrid | Network | ContinuousSpace,
+    agent_portrayal: Callable,
+    c_default="tab:blue",
```
Corvince, line 186: Two things here: 
For API stability I think it would be better if this were a dictionary named "defaults". Maybe we will in the future add more things that can be collected and then it wont grow into an endless list and also we don't have to worry about the order. Now its c, marker, size, whereas matplotlibs order is s, c, marker.

Secondly I would much prefer if we don't use the abbreviated names c and s, even if that is used by matplotlibs api. But we don't want our collect function be tight to matplotlib API, it should be usable for different drawing backends as well.

mesa/visualization/components/matplotlib.py
```diff
@@ -146,132 +180,269 @@ def draw_property_layers(ax, space, propertylayer_portrayal, model):
...
+    arguments["loc"] = loc
+
+    # plot the agents
+    _scatter(ax, arguments)
+
+    # further styling and adding of grid
+    ax.set_xlim(-1, space.width + 0.5)
+    ax.set_ylim(-offset, space.height * offset)
+
+    def setup_hexmesh(
```
Corvince, line 299: No idea how it works, but can we maybe use matplotlibs hexbin? https://matplotlib.org/stable/plot_types/stats/hexbin.html#sphx-glr-plot-types-stats-hexbin-py

Corvince (commented):
This really is a fantastic PR. I think how some plots look is debatable and I haven't' looked into the specific implementations for each space. But this is definitely stuff for follow up PRs and this PR makes it just so much easier to iterate on.

I like the general API of  
```python
draw_space(space, agent_portrayal, ax, **kwargs)
```
I am not sure about the best way to describe the ominous agent_portrayal dictionary, but this now nicely factored out. Just a suggestion regarding the `ax` keyword. Can we make that optional? And if its not passed, it creates a new ax element? What I would like to have is the ability to just call `draw_orthogonal_grid(mymodel.space, portrayal)` and have a visualization, without handling matplotlib figure creation myself.

Or maybe this can be further abstracted away into a `draw_space` function a la 
```python
def draw_space(space, agent_portrayal=None, ax=None, **draw_kwargs, **kwargs):
    pass
```
That is, only space is required, everything else is optional. Additional drawing kwargs to into draw_kwargs. Maybe even have `backend` as a keyword argument and decide there between "matplotlib" and "altair" (or more depending on contributions). In any way a dedicated function for drawing a mesa space and return a working figure. 

This gives us a nice way to plot our spaces in general. Preparing/wrapping those figures than in another function that returns a solara component becomes trivial.

Corvince:
> Unknown

Just some feedback on those, so its not lost. As I said earlier, this is completely independent for merging this PR and should be handled in follow ups (some of these things are not even changed in this PR).
HexBin: Nice hexagons, but the fact that inner ones are drawn with a thicker border because of overlaps is visible and somewhat irritating. 
OrthogonalGrid: I find it irritating that the cells are not square, even if their visualization is arbitrary since they are dimensionless themselves. Also, not seen here, I noticed that the axis ticks are labeled automatically and thus sometimes show as `[0, 2.5, 5, 7,5, ...]` They should be just integers
ContinuousSpace: I find the bold, dashed borders irritating and ugly. Also it sort of implies a torus space, independently of its actual toroidity 
Network: Dont know how much this depends on the used agent_portrayal, but right now its just visually unpleasing.

mesa/visualization/components/matplotlib.py
quaquel: > For API stability I think it would be better if this were a dictionary named "defaults". 
I was actually thinking of using a named tuple, rather than a dict, but I would need to take a closer look what works best. Note that since these are keyword arguments, order does not matter.

> Secondly I would much prefer if we don't use the abbreviated names c and s, even if that is used by matplotlibs api

I find this a bit tricky. The only other "backend" at the moment is altair, and none of this will work for altair plotting given how radically different altair is (for the record, I really like altair). I am, however, fine with reverting to the current labels (i.e., color, size, marker).

mesa/visualization/components/matplotlib.py
quaquel: Looking at the source of hexbin, it uses a polycollection instead of patchcollection as I do here. I am fine with taking a closer look to see if polycollection creates a cleaner grid. I need to look at that anyway also for plotting propertylayers. 

However, I would prefer to leave that for a future PR.

quaquel:
Some quick reactions

> Just a suggestion regarding the ax keyword. Can we make that optional? And if its not passed, it creates a new ax element? What I would like to have is the ability to just call draw_orthogonal_grid(mymodel.space, portrayal) and have a visualization, without handling matplotlib figure creation myself.

Yes that should be easy to do. 

> Or maybe this can be further abstracted away into a draw_space function. That is, only space is required, everything else is optional. 

I think this can be done for agent_portrayal quite easily. I am less sure about the backend idea. 


> Just some feedback on those, so its not lost. As I said earlier, this is completely independent for merging this PR and should be handled in follow ups (some of these things are not even changed in this PR).

I agree, I am also not always happy with the styling of the figures. I agree with most of your comments. For me, the first one to check is voroinoi. Agents are not shown in the "center" of the tile. I find this strange. I haven't modified any of the original code, other than the refactoring to a generic collect_agent_data function. I agree on hexbins and the overlap. Perhaps polycollection as used in matplotlib's `hexbin` resolves this. The orthogonal grid is unavoidable. At the moment, grid cells are centered on integers, and the grid lines are therefore drawn at halves. If you want squares, you need to do `ax.set_aspect('equal')`. Another possibility is to force the figure holding the ax to also be square (e.g., `fig.set_figsize(6,6)`). It's because of these kinds of styling issues that I want to add a user specifiable function that is called after plotting is complete.

mesa/visualization/components/matplotlib.py
EwoutH: If we really want to support multiple backends (which is definitely debatable), I think we need two sets of terms:
- Shared arguments, like "color" and "size" for example, that work with multiple (preferably all) backends. We translate them to the proper input value for that backend on the fly.
- Backend-specific arguments, that work only with that specific backend.

For shared argument, ideally those would be actual arguments, which can be type hinted and document, instead of dict keys.

EwoutH:
On thought I had is that it would be really nice to allow figures with multiple axes. For example, I have two PropertyLayers (partly transparent) and on top two Agent layers. That can be two to four axes (depending how you model/visualize it), which could be overlayed into a single feature.

Ideally you would have some checkboxes that you can turn layers on and off.

All not needed in this PR, but It would be nice if the base would be extendable that this sort of multiple axes in a figure behavior can be supported in the future.

mesa/visualization/components/matplotlib.py
Corvince: Maybe the term "backend" was a bit misleading in this context, given my other suggestion about a generic draw_space function with only matplotlib and altair possibilites. But in the context of this `collect_agent_data` function it just refers to the idea of collecting data specific to the plotting requirements. Its independent on how you use that data for plotting, but you need that data for any plotting library. 

Also note that its the exact requirement we have for (a new, possible) data collector: Iterating over an agentset and collecting some variables (here: c, s and marker), possibly transforming them on the way. Not sure what to make with that information 😆

Corvince:
Here is a proposed API for using all this
```python
# Outside solara
mySpace = MesaSpace()
fig, ax = draw_space(mySpace. **other_args)
ax.modify(somehow) # Basically do anything you want to make it look like you want

# Preparing for SolaraViz
SpaceComponent = make_space_component(**other_args)  
SpaceComponent(myModel) # <= usage as a solara component, standalone or inside SolaraViz. Calls draw_space of the models space

def postprocess(fig, ax):
    ax.modify(somehow)

SpaceComponentModified = make_space_component(**other_args, post_processing=postprocess)
SpaceComponentModified(myModel) # calls postprocess before wrapping in a solara component
``` 

Just spitting out ideas here (again, mostly for follow up PRs)

wang-boyu:
Thanks @Corvince for the ideas on the APIs. I was thinking about the same thing but couldn't figure it out.

> I like the general API of `draw_space(space, agent_portrayal, ax, **kwargs)`

This was exactly what I thought about. But if we put it with existing API:

- `make_space_matplotlib(agent_portrayal, propertylayer_portrayal)`: returns a Solara component, that gets updated when model runs
- `draw_space(space, agent_portrayal, ax, **kwargs)`: returns a matplotlib figure, that is static

The naming of the APIs is a bit confusing to me.

To this end, `make_space_component` appears to be a better API. Questions is, shall we change `make_plot_measure` to something like `make_measure_component` as well?

To elaborate further, I think the overall design is a bit mixed up, with OOP-style classes for Model, Agent, etc, and some functional-style APIs for visualization, e.g., `draw_space`, etc. I raised https://github.com/projectmesa/mesa/issues/1441 a few years ago but couldn't sort it out.

Sorry about the digression. Thanks for the great work on refactoring the code.

Corvince:
> To elaborate further, I think the overall design is a bit mixed up, with OOP-style classes for Model, Agent, etc, and some functional-style APIs for visualization, e.g., draw_space, etc. I raised #1441 a few years ago but couldn't sort it out.

This is something I've thought about as well. Ideally one could just write 
`myspace` in an interactive session/notebook cell and it would just show a visualization (through the ipython "display" magic). And this could be built much more easily on top of this PR.

Also agree on the proposed naming. make_space_component sounds good

quaquel:
I see many interesting ideas, but at this point, I would like to wrap up this PR. There is a lot here already upon which to built further. 

* I have added the possibility of calling any of the draw_x functions with `ax=None` in which case a figure and ax are created using `plt.subplots`. This means that in a notebook, you can get the figure/ax with `plt.gcf()` / `plt.gca()`. 
* I moved the match case part from `SpaceMatplotlib` into its own `draw_space` function. This draw_space function takes an optional `post_process` callable to do any further refinement on the plot. This argument is also available in `make_space_matplotlib` but not in any of the draw_x plotting functions (this seems redundant to me). If you are using these, you are likely in jupyter lab or so and thus can do that easily already.
* I made it possible to call `make_space_matplotlib` with `**space_drawing_kwargs`. Basically, any additional keyword arguments to `make_space_matplotlib` will be passed on to the relevant draw_x function. This means it's now possible to turn off the grid drawing or control the Networx layout algorithm (so closing #2243). 
* I updated the docs in many places to be a lot clearer on what is what and what is expected
*  have added an explicit warning to `collect_agent_data` to emit a warning for all fields in the agent portrayal dict that are ignored. 

What I haven't done, and find a bit tricky, is to make it possible to call any of the draw_x functions without agent_portrayal. There is, of course, a simple solution by falling back on a lambda function that returns an empty dict (as is done in `make_space_matplotlib`). I am happy to add that, but it's not the cleanest solution. Resolving this interacts with getting rid of the defaults in collect_agent_data. I think there is a clean solution possible here using a named tuple or a data class. However, I am not intimately familiar with data classes, and it would change the expected return of `agent_portrayal.` Having an `AgentPortrayalStyle` class and having that as the expected return of `agent_potrayal` might also allow for abstracting away many of the differences between Altair and matplotlib and would make it impossible to return invalid fields (making the new warning redundant).

quaquel:
> On thought I had is that it would be really nice to allow figures with multiple axes. For example, I have two PropertyLayers (partly transparent) and on top two Agent layers. That can be two to four axes (depending how you model/visualize it), which could be overlayed into a single feature.

> Ideally you would have some checkboxes that you can turn layers on and off.

In matplotlib, I doubt you can stack axes to be drawn visually on top of each other. In altair, this is possible. Regardless, all draw_x functions take an optional ax kwarg so you can specify 4 axes and call draw_x for each of them. So what you want is already supported.

quaquel:
I have updated the wolf sheep example to showcase what is now possible. It can look like this:

Screenshot 2024-10-29 at 19 41 57

This example uses both `draw_grid=False` and a simple `post_process` function. 

Note that wolf and sheep agents are not visible because zorder is not controllable (this problem was already there before this refactor).

Corvince (approved):
I think this is ready to merge as-is, but two additional things (that can be part of this PR or outside) I would like to discuss.

1. I would have expected `draw_space` to return fig, ax as well and thus not need the additional `post_process` argument, since you can just work with the axes afterwards (or use gca/gcf as you said). I think post_process is just needed for the solara stuff, where we can't really control the axis after its creation

2. What do you think of the proposed name change to `make_space_component`?

quaquel:
> I would have expected draw_space to return fig, ax as well and thus not need the additional post_process argument, since you can just work with the axes afterwards (or use gca/gcf as you said). I think post_process is just needed for the solara stuff, where we can't really control the axis after its creation

It's a bit complicated. In the current setup, you can call it with an optional `ax` argument. If `ax` does not exist, a new figure and axes are created and could be returned. However, if you pass ax yourself, the code does not know the figure instance and thus cannot return it. One option, which is what seaborn does, is to just always return the axes instance. You can get the figure from this if you want. 

Since `SpaceMatplotlibComponent` also uses draw_space, we need to have post_process here, although I agree that it is redundant in all other use cases.

> What do you think of the proposed name change to make_space_component?

I like the suggested name and will implement it.

EwoutH:
Awesome progress. I would like to do an extensive review tomorrow, if everyone is okay with that.

quaquel:
> Awesome progress. I would like to do an extensive review tomorrow, if everyone is okay with that.

Fine, but please separate interesting ideas and follow-ups from what is required to merge this.

mesa/examples/basic/conways_game_of_life/app.py
```diff
@@ -1,12 +1,20 @@
+import os.path
+import sys
+
+sys.path.insert(
+    0, os.path.abspath(os.path.join(os.path.dirname(__file__), "../../../../"))
+)
+
+
```
EwoutH, line 8: This section was only used for testing, right? Or do we want it in every example? (might help, I also add it often)

Don't mind what as long as it's consistent.

If you want to remove it:
```suggestion
```

mesa/visualization/components/matplotlib.py
```diff
@@ -1,38 +1,80 @@
...
+    NetworkGrid,
+    PropertyLayer,
+    SingleGrid,
+)
 from mesa.visualization.utils import update_counter
 
+# For typing
+OrthogonalGrid = SingleGrid | MultiGrid | OrthogonalMooreGrid | OrthogonalVonNeumannGrid
+HexGrid = HexSingleGrid | HexMultiGrid | mesa.experimental.cell_space.HexGrid
+Network = NetworkGrid | mesa.experimental.cell_space.Network
```
EwoutH, line 40: This seems something we want to maintain throughout Mesa (like a space hierarchy). Can we put this on a more central place?

mesa/visualization/components/matplotlib.py
```diff
@@ -43,48 +85,157 @@ def SpaceMatplotlib(
...
+        draw_property_layers(space, propertylayer_portrayal, ax=ax)
+
+    if post_process is not None:
+        post_process(ax=ax)
 
+    return ax
 
-def draw_property_layers(ax, space, propertylayer_portrayal, model):
+
+def draw_property_layers(
```
EwoutH, line 214: Note to self: Make singular `draw_property_layer` function, which returns a single axes. Multiple axes with one (semi-transparent) propertylayer each, and optionally a space, can than be combined into a single figure.

EwoutH (approved):
This PR is remarkably clean for it's size and scope, excellent work. Few minor comments, but they are really nitpicks.

I especially like how the arguments in the dict are now handled and how everything is documented. Thanks!

mesa/visualization/components/matplotlib.py
quaquel: This is just typing information, not sure how that could be centralized.

mesa/visualization/components/matplotlib.py
quaquel: Not sure I understand the idea

Issue #2431, opened 2024-10-28
Move propertylayer from DiscreteSpace to Grid


quaquel:
In #2319, support for propertylayers was added to the new style grids. This was done in the root space class: `DiscreteSpace`. However, property layers are meaningful only, as far as I can tell, for orthogonal grids and hexgrids. I don't see how they would meaningfully work for `Network` or `VoroinoiGrid`. If I am correct, then the relevant code should be moved from `DiscreteSpace` into `Grid`.

EwoutH:
The distinction here needs to be made if PropertyLayers have their own (independent) coordinate system, or inherit it from a Grid or other object.

I’m nudging towards the former, so you can have (for example) a Network on top of a PropertyLayer. But the PropertyLayer should probably be slightly extended to allow that.

CC @wang-boyu

wang-boyu:
Thanks for tagging me here. Unfortunately I'm not very familiar with the new style grids, so my comments below may not apply.

> if PropertyLayers have their own (independent) coordinate system

In Mesa-Geo a "PropertyLayer" or rather, a raster layer can have its own coordinate reference system (crs), although it will be converted to the GeoSpace's common crs once it's added into the GeoSpace. For Mesa, I'd imagine that all PropertyLayers will share the same default crs, in which the bottom left corner is the origin? Otherwise we'll need to provide ways to convert crs (and possibly resolution) and it's not very straightforward.

> a Network on top of a PropertyLayer

I guess a more common way of doing this would be through [node attributes](https://networkx.org/documentation/stable/reference/generated/networkx.classes.function.set_node_attributes.html)...? But I could be wrong about this.

quaquel:
> In Mesa-Geo a "PropertyLayer" or rather, a raster layer can have its own coordinate reference system (crs), although it will be converted to the GeoSpace's common crs once it's added into the GeoSpace. For Mesa, I'd imagine that all PropertyLayers will share the same default crs, in which the bottom left corner is the origin? Otherwise we'll need to provide ways to convert crs (and possibly resolution) and it's not very straightforward.

This would be my preferred implementation for property layers in mesa itself. As far as I know, conversions are not supported at the moment. It would, therefore, be best to explicitly enforce that property layers have the same reference system by having property layers "inherit" the width and height from their owning space.

We can always revisit that choice down the line and provide conversion systems, but at the moment, we are stuck between both. We don't enforce a common reference system explicitly but don't provide any conversion. In my view, that's the worst of both worlds.

[...]

PR #2336, opened 2024-10-03
Visualize PropertyLayers

EwoutH:
This PR adds support for visualizing PropertyLayers in the Matplotlib-based space visualization component. It allows users to overlay PropertyLayer data on top of the existing grid and agent visualizations, or on its own.

It introduces a new `propertylayer_portrayal` parameter to customize the appearance of PropertyLayers and refactors the existing space visualization code for better modularity and flexibility.

### Motive
The PropertyLayer is becoming a key feature in Mesa, so it should be visualized easily. This feature enables users to easily visualize multiple layers of data simultaneously, such as environmental factors or agent properties, alongside the agents themselves.

### Implementation
- Added a new `propertylayer_portrayal` parameter to the `make_space_matplotlib` function.
- Implemented a `draw_property_layers` function to render PropertyLayer data using Matplotlib's `imshow`.
- Modified the `_draw_grid` function to incorporate PropertyLayer visualization before drawing agents.
- Added support for both color-based and colormap-based PropertyLayer rendering.
- Implemented alpha blending for PropertyLayer visualization to allow overlaying multiple layers.
- Added colorbar support for PropertyLayer visualization.

### Usage Examples
You can visualize multiple PropertyLayers ("grass" and "temperature" in this case) along with agents in a grid space:
```python
def agent_portrayal(agent):
    return {"color": "red", "shape": "o", "size": 20}

propertylayer_portrayal = {
    "grass": {
        "colormap": "Greens",
        "alpha": 1,
    },
    "temperature": {
        "colormap": "coolwarm",
        "alpha": 0.333,
        "vmin": 0,
        "vmax": 100
    },
}

model = ExampleModel(10, 10)
page = SolaraViz(
    model,
    [make_space_matplotlib(agent_portrayal, propertylayer_portrayal)],
)
```
image

You can also use a single color for a PropertyLayer. Here we render the "cell_layer" PropertyLayer using shades of black, with the highest value having 0% opacity and the lowest value having 100% opacity. For Game of Life that means alive cells are fully black, and dead cells transparent.

```python
propertylayer_portrayal = {
    "cell_layer": {
        "color": "Black",
        "alpha": 1,
        "colorbar": False,
    },
}
```
image

Any [matplotlib color](https://matplotlib.org/stable/api/colors_api.html#module-matplotlib.colors) is valid. The highest value will have that color with 0% opacity, the lowest with 100% opacity (which can again be scaled with modifying `vmin` and `vmax`).

## Additional Notes
[...]

PR #2440, opened 2024-10-30
Stabilize PropertyLayer

EwoutH:
Removes the warning that the PropertyLayer is experimental, and thus stabilize it.

It has been used in example models (https://github.com/projectmesa/mesa-examples/pull/214), integrated in the cell space (https://github.com/projectmesa/mesa/pull/2319) and the visualisation (both standalone (https://github.com/projectmesa/mesa/pull/2336) and as part of cell spaces (https://github.com/projectmesa/mesa/pull/2430)) and in Mesa for almost a year (https://github.com/projectmesa/mesa/pull/1898) without needing API changes.

Note that while the PropertyLayer itself is stable, not all it's integrations might be, like the integration into the still experimental cell space.


quaquel:
We need to resolve the coordinate system discussion raised in #2431 as part of the stabilization.

EwoutH:
What would minimally be done to declare the current functionality stable?

(CC @wang-boyu)

quaquel:
Resolve #2431.

EwoutH:
Yeah I was also thinking of splitting it out. We could do `mesa.spatial` maybe for if we want to add more spatial / CRS / geo stuff

quaquel:
I> Yeah I was also thinking of splitting it out. We could do mesa.spatial maybe for if we want to add more spatial / CRS / geo stuff

I suggest underpromising by calling it `property_layer`and keeping the idea of integrating where possible stuff from mesa-geo as a possible over-delivery in the longer run. Remember, we have to life with the API choices we make on this for at least MESA 3.

EwoutH:
That's why I was considering a more general name instead of a specific file for one single class.

quaquel:
In my view it would be PropertyLayer and HasPropertyLayers (a refactor of _PropertyGrid that can also be used with newstyle discrete spaces).

quaquel:
Just a quick remark: at the moment, property layers throw user warnings based on the dtype stuff. Its not possible to get rid of those, because of the hardcoded `np.float64`, this should be changed to `float` or the test before throwing the user warning should be changed. I prefer the former since for numpy np.float64 and the python built-in float are equivalent.

EwoutH:
@quaquel What's your current stance?

quaquel:
I think in 3.2 we can stabilize it

EwoutH:
We have branched 3.1 maintenance, is this good to go for 3.2?

EwoutH:
@quaquel with the discrete space stabilized, does it make sense to also stabilize the PropertyLayer?

quaquel:
yes

EwoutH:
@projectmesa/maintainers if no objections, I would like a green checkmark on this :)

quaquel (approved):

PR #2441, opened 2024-10-30
remove deprecated HexGrid class

### LINKED ISSUE #1847: Clean up HexSingleGrid(HexGrid) andHexGrid
### Discussed in https://github.com/projectmesa/mesa/discussions/1827

<div type='discussions-op-text'>

<sup>Originally posted by **jackiekazil** October  2, 2023</sup>
Why is the HexSingleGrid subclassed from HexGrid and HexGrid has nothing in it? 
I was reading through the code and confused by this -- because it is also visible in the docs. Wondering if there was logical reason or if someone remembered why? I was thinking we should get rid of HexGrid or make it private. 

See the following two lines -https://github.com/projectmesa/mesa/blob/b2f642fd1a986f3114ba6f6f6e9bc119c037062f/mesa/space.py#L742
https://github.com/projectmesa/mesa/blob/b2f642fd1a986f3114ba6f6f6e9bc119c037062f/mesa/space.py#L768</div>

quaquel:
This removes the long deprecated HexGrid class from mesa.space.

closes #1847

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -1286,6 +1286,7 @@ class HexMultiGrid(_HexGrid, MultiGrid):
         empties: Returns a set of hexagonal coordinates for all empty cells.
     """
 
+
 class ContinuousSpace:
     """Continuous space where each agent can have an arbitrary position.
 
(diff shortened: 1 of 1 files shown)


EwoutH (approved):
Looks good. Are there tests that can/need to be removed?

quaquel:
> Looks good. Are there tests that can/need to be removed?

those would have broken and all tests pass, so no.

PR #2445, opened 2024-10-30
expand ax.scatter kwargs that can be used

### LINKED ISSUE #1183: Color gradient feature in agents portrayal is not documented properly
See e.g. https://github.com/projectmesa/mesa/blob/main/examples/wolf_sheep/wolf_sheep/server.py#L31

This feature was added via https://github.com/projectmesa/mesa/commit/da194e3556e7a8e61af05da3f726ef42ec7c7c0c, but is lacking documentation.

quaquel:
#2430 added the possibility to pass keyword arguments on to the various draw functions. This builds on that by passing any keyword arguments not used explicitly in any of the draw functions on to `ax.scatter`. This can, for example, be used to set a custom colormap, control vmin, vmax etc., as shown in the updated Boltzmann wealth example (see below for a screenshot)

Screenshot 2024-10-30 at 20 02 09

Closes #1183.

### DIFF
--- a/mesa/examples/basic/boltzmann_wealth_model/app.py
+++ b/mesa/examples/basic/boltzmann_wealth_model/app.py
@@ -2,6 +2,7 @@ from mesa.examples.basic.boltzmann_wealth_model.model import BoltzmannWealthMode
 from mesa.visualization import (
     SolaraViz,
     make_space_component,
+    make_plot_component
 )
 
 
@@ -40,7 +41,7 @@ model1 = BoltzmannWealthModel(50, 10, 10)
 SpaceGraph = make_space_component(
     agent_portrayal, cmap="viridis", vmin=0, vmax=10, post_process=post_process
 )
-GiniPlot = make_plot_measure("Gini")
+GiniPlot = make_plot_component("Gini")
 
 # Create the SolaraViz page. This will automatically create a server and display the
 # visualization elements in a web browser.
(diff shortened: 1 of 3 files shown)


mesa/examples/basic/boltzmann_wealth_model/app.py
```diff
@@ -7,12 +7,8 @@
 
 
 def agent_portrayal(agent):
-    size = 10
-    color = "tab:red"
-    if agent.wealth > 0:
-        size = 50
-        color = "tab:blue"
-    return {"size": size, "color": color}
+    color = agent.wealth
```
EwoutH, line 10: Maybe add a comment here what’s happening for those not familiar with colormapping.

EwoutH (approved):
It’s so incredibly nice to see examples updated together with code. And thanks for the screenshot, always appreciated!

Looks clean, let’s give others also a bit of time to review.

At some point we should add a note or section about handeling axes and their arguments (including things like colormapping) to the tutorial. Can be a follow up.

quaquel:
If linewith and edgecolor are handled analogously to color and size, basically everything in ax.scatter is controllable and we can also close #2341. I might just do that in a seperate PR to keep this one clear and focussed.

EwoutH:
I would say go ahead and merge tomorrow morning, our US-timezone friends can check tonight and otherwise it's good to go.

Issue #1823, opened 2023-09-26
JupyterViz: Allow user to specify space figure output format in svg/png


rht:
Currently, it is hardcoded to be png, because png is much faster. According to https://github.com/projectmesa/mesa/pull/1819#issuecomment-1733624072, we should allow user to pass in extra arguments to `solara.FigureMatplotlib`. According to https://solara.dev/api/matplotlib, the `kwargs` is for extra arguments to Matplotlib's `savefig`, i.e. Solara renders a figure by doing a `savefig` operation.

cc: @ankitk50. This doesn't require agreement/consensus on the design of how this is going to implemented, and so there shouldn't be any blocker.

Issue #795, opened 2020-02-09
Optional title in ChartVisualization modules 


rebeccadavidsson:
#### Add an attribute to the chart visualization modules that gives the ability to add a title to the plots. 
Hereby changing the module input and js side of the charts. This will help multiple different charts to be clearer.

tpike3:
Closing --- overcome by updates to visualization modules

Issue #2434, opened 2024-10-30
rename altair.py and matplotlib.py


quaquel:
The naming of `mesa.visualization.components.altair` and `mesa.visualization.components.matplotlib` can cause namespace clashes. They thus need to be renamed to e.g., `altair_components` and `matplotlib_components`.

Corvince:
Also note that I would suggest the naming "component" to refer to solara components specifically. But we now also provide general drawing functions that return matplotlib figures, so the file name is outdated and should be reconsidered more generally

quaquel:
It might then make sense to move the draw_space and draw_x functions into their own module (e.g., space_drawing) and leave the solara specific stuff in matplotlib_components

PR #2446, opened 2024-10-30
rename make_plot_measure to make_plot_component and add some kwargs

### LINKED ISSUE #2443: port API ideas from make_space_component to make_plot_measure
#2430 adds more fine-grained control for the user to fine tune a plot via post_process. This would also be very convenient for make_plot_measure.

quaquel:
This renames make_plot_measure to make_plot_component and adds 2 new keyword arguments. `post_process` is identical to the one used in `make_space_component`. The save_format closes another open issue. See virus_on_a_network for using post_process.

closes #2443, #1823, #1537, #795

Screenshot 2024-10-30 at 20 25 09

### DIFF
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -172,27 +172,27 @@ from mesa.visualization import SolaraViz, make_space_component, make_plot_compon
 
 
 def agent_portrayal(agent):
-   return {"color": "blue", "size": 50}
+    return {"color": "blue", "size": 50}
 
 
 model_params = {
-   "N": {
-      "type": "SliderInt",
-      "value": 50,
-      "label": "Number of agents:",
-      "min": 10,
-      "max": 100,
-      "step": 1,
+    "N": {
+       "type": "SliderInt",
+       "value": 50,
+       "label": "Number of agents:",
+       "min": 10,
+       "max": 100,
+       "step": 1,
    }
 }
 
 page = SolaraViz(
-   MyModel,
-   [
-      make_space_component(agent_portrayal),
-      make_plot_component("mean_age")
-   ],
-   model_params=model_params
+    MyModel,
+    [
+       make_space_component(agent_portrayal),
+       make_plot_component("mean_age")
+    ],
+    model_params=model_params
 )
 page
 ```
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -168,31 +168,31 @@ The results are returned as a list of dictionaries, which can be easily converte
 Mesa now uses a new browser-based visualization system called SolaraViz. This allows for interactive, customizable visualizations of your models. Here's a basic example of how to set up a visualization:
 
 ```python
-from mesa.visualization import SolaraViz, make_space_component, make_plot_measure
+from mesa.visualization import SolaraViz, make_space_component, make_plot_component
 
 
 def agent_portrayal(agent):
-    return {"color": "blue", "size": 50}
+   return {"color": "blue", "size": 50}
 
 
 model_params = {
-    "N": {
-        "type": "SliderInt",
-        "value": 50,
-        "label": "Number of agents:",
-        "min": 10,
-        "max": 100,
-        "step": 1,
-    }
+   "N": {
+      "type": "SliderInt",
+      "value": 50,
+      "label": "Number of agents:",
+      "min": 10,
+      "max": 100,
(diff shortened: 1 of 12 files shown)


EwoutH (commented):
Can we let `make_plot_measure` throw a warning that it is replaced with `make_plot_component`? That way people using the betas have some clue how to fix their models.

The migration guide might also need an update.

docs/overview.md
```diff
@@ -168,31 +168,31 @@ The results are returned as a list of dictionaries, which can be easily converte
...
-    "N": {
-        "type": "SliderInt",
-        "value": 50,
-        "label": "Number of agents:",
-        "min": 10,
-        "max": 100,
-        "step": 1,
-    }
+   "N": {
+      "type": "SliderInt",
```
EwoutH, line 180: What’s happening here? Moving to 3 space intends?

quaquel: I guess an artifact of the automatic renaming?

EwoutH: Looks almost like some changed IDE setting. Strange that ruff likes it?

quaquel: It makes no sense and I did not touch that file directly but only indirectly by using pycharms refactor --> rename

mesa/visualization/components/matplotlib.py
```diff
@@ -618,30 +618,46 @@ def _scatter(ax: Axes, arguments):
             )
 
 
-def make_plot_measure(measure: str | dict[str, str] | list[str] | tuple[str]):
```
EwoutH, line 621: Letting this function call a warning / informative error is an oneliner right?

quaquel: and more lines to move around in some of the other refactoring still to come to address other open issues.

quaquel: I just checked, None of this is in 2.4 only in the mesa 3 alpha's and beta's. So I really don't see the point of bothering with warnings. I am going to move more stuff around for example for #2434. I agree that warnings etc. are in order for official releases and when deprecating stable stuff. This has never been officially stable yet anyway.

EwoutH: I don’t like such breaking changes without proper error handling so late in the beta release cycle, but since we never released this as stable before I find it acceptable if it’s your strong preference.

Don’t expect me to make an habit out of it ;)

<hr>

If I knew we were going to do such extensive overhauls I would have suggested staying longer in alpha, maybe something to more explicitly discuss if/when 4.0 is targeted.

quaquel: I did not know that either until I started looking at this stuff. 

Do you have any sense of how many downloads there have been for some of the pre releases?

quaquel: Also, we were more or less in agreement that the components part of solara would be experimental in 3.0 anyway 😄

docs/overview.md
quaquel: I have added the spaces back in to ensure the code is rendered correctly.

Corvince:
I agree with @EwoutH to keep the old function names as deprecated aliases for the new functions. Apart from the name change it should still just work, right? No need to cause more change than needed. Yes it was always experimental but we recommended it for quite some time now, since mesa 2.2 or something? 

Yes it's a bit more code to move around, but I think it's a good trade off. And then immediately delete it for 3.1

quaquel:
I added the deprecation warning (also for make_space_component / make_space_matplotlib )

Corvince:
Thanks!

Corvince (approved):

EwoutH:
Well handled, appreciated!

One minor note, I think the renames in the `__init__.py` might brake some import ways, since the old names are now not present in there.

docs/overview.md
```diff
@@ -177,20 +177,20 @@ def agent_portrayal(agent):
...
-        "max": 100,
-        "step": 1,
-    }
+       "type": "SliderInt",
+       "value": 50,
+       "label": "Number of agents:",
+       "min": 10,
+       "max": 100,
+       "step": 1,
+   }
```
EwoutH, line 186: We still went from 8 to 7 whitespaces here somehow.

quaquel:
exactly my plan

PR #2450, opened 2024-11-02
Docs reorganization

quaquel:
Quick test based on a discussion in #2448. This reorganizes the docs by moving items from the top menu bar into the left-hand section navigation column. 

I have gone for a rather radical approach and this is really proof of concept only at the moment. The top menu now only has overview, getting started, and API. All other stuff have been added to the getting started section. The getting_started.md file needs to have text in addition to the toctree to serve as a proper landing page. 

I also think the migration guide might be better of in the top menu bar so its easy to find for existing users.

@EwoutH let me know your thoughts.


EwoutH:
Thanks for opening the PR. Looking at the rendered docs, I’m a bit conflicted.

> I also think the migration guide might be better of in the top menu bar so its easy to find for existing users.

The challenge is that you can make this argument for a lot of the resources. Let me think about this a bit and maybe look at a few other libraries how they do it.

One idea is to (almost) abolish the top bar and have a semi permanent left side bar, since that fits more (but also takes more space).

quaquel:
> One idea is to (almost) abolish the top bar and have a semi permanent left side bar, since that fits more (but also takes more space).

That might require changing to a completely different rtd style.

EwoutH:
Rome isn’t built in a day (unfortunately).

I must say I find these docs very clean and clear in their layout, I wonder which brilliant maintainer designed those 😇

IMG_0074

tpike3:
This is a good discussion thanks @quaquel 

I like the pydata theme but am not necessarily opposed to other approaches. 

Some thoughts would be 
- Agree to a simpler lay out with groupings
- I would argue we can get rid of mesa packages, that really never went anywhere; maybe work with other groups like [rbb4abm](https://www.rbb4abm.com/) ??
- Maybe instead of examples, we should call it gallery like [Network](https://networkx.org/documentation/stable/auto_examples/index.html) and in this case we could overtime merge How to and Examples in one page. 

Just some off the top of the head thoughts. I am not tied to anything.

quaquel:
I made a few changes. First, I moved examples and the migration guide out of the getting started page. They are now back in the top bar menu.

Screenshot 2024-11-04 at 16 03 41

Second, I figured out how to get stuff in the left hand column. This is populated based on a toctree directive, which can also be declared hidden. See here what it now looks for the examples:

Screenshot 2024-11-04 at 16 03 52

Personally I think this is moving in the right direction. I want to add some text to the getting started page. Probably have a subsection called tutorials, with 2 tutorials for now. This creates space for more tutorials in the future. Again, feedback is welcome.

quaquel:
[Mesa Overview](https://mesa--2450.org.readthedocs.build/2450/overview.html) might be combined with the getting started part. This would remove one more item from the top row that is only a single page and thus is not really a section with several associated pages. Some of the material might be used to fill out the getting started landing page, while the rest might be a good follow up in between tutorials and the full gory details of the API docs.

EwoutH:
That's looking great!

> Second, I figured out how to get stuff in the left hand column. This is populated based on a toctree directive, which can also be declared hidden.

That's good to know. Looking through it, I think the Section Navigation (left column) can be hidden in most cases (by default). Navigation could go through the top bar and the main Getting Started, Examples, etc. pages themselves.

quaquel:
I have done some further reorganization as discussed. So, I combined the old overview page and the new getting started idea. I reorganized the material a bit. So, the overview is now: tutorials, examples that ship with mesa, mesa packages overview, further resources. 

I also updated the best practices a little bit wrt random. 

This is ready for a proper review

docs/getting_started.md
```diff
@@ -1,9 +1,20 @@
...
-## Mesa Modules
+
+## Tutorials
+If you want to get a quick start on how to build agent based models with MESA, check the tutorials:
+
+- [Introductory Tutorial](tutorials/intro_tutorial): Learn how to create your first Mesa model.
+- [Visualization Tutorial](tutorials/visualization_tutorial.html): Learn how to create interactive visualizations for your models.
+
+## Examples
+Mesa ships with a collection of example models. These are classic ABMs, so if you are familiar with ABMs and want to get a quick sense of how MESA works, these examples are great place to start. You can find them [here](examples)
```
EwoutH, line 14: Missing dot
```suggestion
Mesa ships with a collection of example models. These are classic ABMs, so if you are familiar with ABMs and want to get a quick sense of how MESA works, these examples are great place to start. You can find them [here](examples).
```

docs/getting_started.md
```diff
@@ -165,7 +176,11 @@ results = mesa.batch_run(
...
 
 ### Visualization
-Mesa now uses a new browser-based visualization system called SolaraViz. This allows for interactive, customizable visualizations of your models. Here's a basic example of how to set up a visualization:
+Mesa now uses a new browser-based visualization system called SolaraViz. This allows for interactive, customizable visualizations of your models.
+
+The core functionality for building your own visualizations resides in the [`mesa.visualization`](apis/visualization) namespace
+
+Here's a basic example of how to set up a visualization:
 
 > **Note:** SolaraViz is experimental and still in active development for Mesa 3.0. While we attempt to minimize them, there might be API breaking changes between Mesa 3.0 and 3.1. There won't be breaking changes between Mesa 3.0.x patch releases.
```
EwoutH, line 185: This note is now placed a bit unfortunate. I would move it a few lines up, maybe directly after the first paragraph.

docs/getting_started.md
EwoutH, line 1: With the rename, we should do a vigilant search for old links to overview.

I will also setup a redirect in Readthedocs.

EwoutH (approved):
Thanks for your patience. Overal looks really good, clear improvement over what we had. Few minor comments.

docs/getting_started.md
EwoutH: image

docs/getting_started.md
quaquel: I am fine with moving it, but I prefer notes at the end of a section (so multiple paragraphs). Also, this is how its rendered, so the note is directly there when looking at the page.

Screenshot 2024-11-08 at 12 49 04

docs/getting_started.md
quaquel: I already did a find in files and fixed all of them. As far as I can tell, there are none left in the docs unless they are in the docstrings of modules which would highly surprise me.

docs/conf.py
tpike3, line 1: This is a bit nitpicky so I am not tied to it but for the Migration Guide can we get rid of the empty section navigation bar on the left hand side. It would be added somewhere in the html options (start line 123) . Should be something like 

```html_sidebars = {
    'migration_guide': [],  # No sidebar migration
   }
```

tpike3 (approved):
Thanks @quaquel!

docs/getting_started.md
EwoutH: It’s more that it now says: here’s an example and then you get a note, not the example.

docs/getting_started.md
EwoutH: Perfect!

docs/conf.py
EwoutH: Oh I like this

docs/index.md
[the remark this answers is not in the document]
quaquel: its there now

Issue #885, opened 2020-08-19
Export the simulation as a video


nivedkrishnans:
**What's the problem this feature will solve?**
A readily obtainable video output of our simulation would make it easier for collaboration between people working on a model as well as for presentations, publications, etc. A normal screen recording will not do well, as the time required for each step may fluctuate during a simulation. NetLogo offers an option to record the simulation region or even the whole interface (with the sliders, graphs, etc).

**Describe the solution you'd like**
Just a thought: It would be better if we can obtain the exact same thing as the browser visualization as a video. Maybe we can screen capture the active elements of the screen during every step and store it temporarily, and at the end of the simulation, stitch them together (say 25 steps per second) and output it into a predefined location/filename. 

EDIT: I came across a few python packages that can 'program' svgs. We can assign some shape for every agent and then put them together at the appropriate coordinates to get a vector image for each step. Maybe also display the parameters of the model, etc.

rht:
Temporary solution is to record the browser screen using e.g. https://chrome.google.com/webstore/detail/screen-recorder/hniebljpgcogalllopnjokppmgbhaden.

But given that we use Solara now, it should be easier to capture the Matplotlib space state as pictures that can be turned to animated GIF.

vagechirkov:
> Temporary solution is to record the browser screen using e.g. https://chrome.google.com/webstore/detail/screen-recorder/hniebljpgcogalllopnjokppmgbhaden.
> 
> But given that we use Solara now, it should be easier to capture the Matplotlib space state as pictures that can be turned to animated GIF.

The second option (to capture the matplotlib space state as images) sounds good to me. Any hints on where to start with this? I'm happy to work on this feature if it might be useful to others.

rht:
A quick hack to test the idea is to add a line after https://github.com/projectmesa/mesa/blob/9c9a02e25df7c83c866ae9a23416c5d32e76a48d/mesa/experimental/jupyter_viz.py#L262 to save the plot to a file, which filename is an increasing number. And then combine the pictures into 1 file via `imagemagick` CLI.

rht:
https://askubuntu.com/questions/648244/how-do-i-create-an-animated-gif-from-still-images-preferably-with-the-command-l

Corvince:
You can also have a look at https://matplotlib.org/stable/api/animation_api.html

Especially the examples on the bottom are helpful to create an animation. No need to use solara here, you can also just do your plots normally and call model.step() inside the Animation function.

[...]

PR #2453, opened 2024-11-03
implement VideoViz to record model runs in a video

wang-boyu:
## Summary

This PR consists of three commits:

1. Various `draw_space()` functions are moved and refactored into a separate `matplotlib_renderer.py` file.
   Previously:
   ```python
   draw_space(grid, agent_portrayal, ax)
   ```
   Now:
   ```python
   space_renderer = SpaceRenderMatplotlib(agent_portrayal=agent_portrayal)
   space_renderer.draw(model, ax)
   ```
   and for measures:
   ```python
   measure_renderer = MeasureRendererMatplotlib("measure_name")
   measure_renderer.draw(model, ax)
   ```
   This is mainly to extract matplotlib plotting functions/classes out of Solara components, so that they can be re-used by other visualization methods, such as the VideoViz in the third commit. Besides, I don't think it changes how SolaraViz is used.
2. `make_space_component` and `make_plot_component` are moved from `visualization.components.matplotlib` into `visualization.solara_viz`, with a `backend` option to choose from `matplotlib` or `altair`. Default space drawing is changed from `altair` to `matplotlib`.
3. Implement a `VideoViz` that has similar APIs to `SolaraViz` with an example for the basic Schelling model. Essentially, running
   ```bash
   cd mesa/examples/basic/schelling
   python3 video.py
   ```
   will produce a `video.mp4` file:
   
   https://github.com/user-attachments/assets/f8612ec4-bda0-496f-bc0e-2fa65ca939b9
   
   As mentioned in 1), VideoViz uses common matplotlib renderers extracted from SolaraViz.

## Usage Examples

In file `mesa/examples/basic/schelling/video.py`.

## Additional Notes


EwoutH:
I will look at the code tomorrow, but the video looks lit! 🔥

quaquel:
I'll take a detailed look tomorrow. I like the video! 

This does, however, interact with #2447, and I am, at face value, not yet convinced by the OO style over the function style API for now (which also interacts with some of the ideas in #2389). @wang-boyu, how do you view the interdependencies between this pr and #2447 and #2389?

wang-boyu:
Oh sorry I didn't mean to replace https://github.com/projectmesa/mesa/pull/2447. I wanted to have VideoViz for a project I'm working on and ended up with this PR. To me the OO-style vs. functional style is really just a design choice -

1. It doesn't really affect how users use `SolaraViz` and `make_space_component`.
2. I chose OO-style to have a common protocol `draw(model, ax) -> plt.Axes` between space and measure renderers, so that VideoViz can have `components: Sequence[MatplotlibRenderer]`. But this should also be possible with the functional-style API.

So if we end up with funcitonal-style API instead, VideoViz can be built based on that as well. I think my main point here is to isolate matplotlib rendering out of SolaraViz so they it can be re-used by VideoViz.

Regarding https://github.com/projectmesa/mesa/issues/2389, I really like the ideas of having the more powerful APIs, instead of defining dictionaries in a `agent_portrayal` function. It reminds of me of Hadley Wickham's [A Layered Grammar of Graphics](https://doi.org/10.1198/jcgs.2009.07098) and ggplot2 in R (which is also funtional style btw). I think what we want is something very similar but for ABM visualization? My concerns are that 1) designing such APIs takes lots of efforts, and 2) implementing the APIs also takes lots of efforts. But it would be truly great to have.

quaquel:
> Oh sorry I didn't mean to replace #2447. I wanted to have VideoViz for a project I'm working on and ended up with this PR.

No worries. Having VideoViz seems like a good idea, would close #885, and it's just very useful. I only want to make sure that we maintain some coherence in the development. 

So, in my understanding, this pr adds:

1. Various draw_space() functions are moved into a separate matplotlib_renderer.py file.

This directly overlaps with #2447. It seems we agree that separating the `draw_x` functions into their own namespace makes sense. @wang-boyu, would you have a chance to review #2447 and let us know what you think? We either move forward there, or we move forward here with this code reorganization.

2. Refactoring the draw_x functions into an OO-style API.

I saw your comment in #2389 and hope to return to that asap. For MESA 3.0, I would counsel against this refactoring because it is breaking the established API. In my view, we should hash out an OO API for space and plot drawing, have this as experimental in 3.1, and see where it goes. I suggest we have the conversation on this in #2389. A major challenge, I expect, will be to abstract away the backend, given how radically different Altair and Matplotlib are. 

3. `make_space_component` and `make_plot_component` are moved from `visualization.components.matplotlib` into `visualization.solara_viz`, with a backend option to choose from matplotlib or altair. 

This directly overlaps with #2447, although I moved these functions into the `__init__` of the components namespace. To me, it makes more sense to keep everything component related in the component namespace. They would, however, still be importable from `mesa.visualization`, so `from mesa.visualization import make_space_component`.

4. Default space drawing is changed from altair to matplotlib.

Yes, I agree, and this was also on my to-do list for after #2447. 

5. Implement a VideoViz that has similar APIs to SolaraViz

The core of this PR, in my understanding, and, yes, a very good idea. 


So, I propose the following next steps. 
1. @wang-boyu could you share your thoughts on $2447? I would like to come to a decision on whether to do the API reorganization via that PR or this PR. 
2. Open a separate small PR, only changing the backend of solara_vis from Altair to matplotlib. This is a good idea and can be its own separate atomic PR. 
3. how hard would it be to modify VideoViz to use the `draw_x` / `make_x_component`? That would be, for now, my preferred implementation that could even be included in MESA 3.0 if others agree.

mesa/examples/basic/schelling/video.py
quaquel, line 1: great to see the examples used for this as well.

One concern, not for this pr, is how to add these additional files to the docs.

mesa/visualization/video_viz.py
```diff
@@ -0,0 +1,184 @@
...
+
+    return SpaceRenderMatplotlib(
+        agent_portrayal,
+        propertylayer_portrayal,
+        post_process=post_process,
+        **space_drawing_kwargs,
+    )
+
+
+def make_measure_component(
```
quaquel, line 55: In the other parts of the visualization, we shifted to just calling it a plot component; what are the reasons for using the, in my view, ambiguous label of measure here?

mesa/visualization/video_viz.py
```diff
@@ -0,0 +1,184 @@
...
+        """Initialize video visualization configuration.
+
+        Args:
+            model: The model to simulate and record
+            components: Sequence of component objects defining what to visualize
+            title: Optional title for the video
+            figsize: Optional figure size in inches (width, height)
+            grid: Optional (rows, cols) for custom layout. Auto-calculated if None.
+        """
+        # Check if FFmpeg is available
```
quaquel, line 95: This is perhaps for users the most tricky part. I agree on the error handling here.  I would suggest also adding this to the module level docstring.

mesa/visualization/video_viz.py
```diff
@@ -0,0 +1,184 @@
+"""Video recording components for Mesa model visualization."""
```
quaquel, line 1: I suggest making clear that this only works with the  matplotlib backend

tests/test_matplotlib_renderer.py
quaquel, line 1: Great to see tests!

EwoutH:
I will give you two a bit of time to align on this topic, let me know if a fresh perspective is appreciated!

wang-boyu:
Thanks for reviewing this PR @quaquel. Appreciate it. I'll respond in next few days and also take a look at https://github.com/projectmesa/mesa/pull/2447. Lots of changes are indeed similar despite in different styles.

mesa/visualization/video_viz.py
wang-boyu: Thanks for asking. I kept the impression from the old (now experimental) SolaraViz: https://github.com/projectmesa/mesa/blob/69571b401d97af68d85aa49205b0f53f494ac721/mesa/experimental/solara_viz.py#L58 and https://github.com/projectmesa/mesa/blob/69571b401d97af68d85aa49205b0f53f494ac721/mesa/experimental/solara_viz.py#L72
where there's a card titled "Space" and another titled "Measure". "Plot" seems a bit amgiguous for me because we can have a plot for space, a plot for measure, and possible plots for other stuff. I might have missed some discussions when we made the change.

quaquel: I just followed a suggestion from @Corvince in #2443. 

I agree on the ambiguity, but I contend that space is different in kind from, say, line plots, histograms, or other visualizations of data created by a model. So, for now, I am fine with separeting space from other kinds of plots. 
Since we are still hashing out and refining the visualization style, I am fine with this distinction at the moment.

Corvince:
I thought I already commented on this PR, but it seems I only wanted to do so. Nice work with the output, the video looks very clearn.

On the topic of OO vs FP: There are use-cased for both approaches, but I don't think OO is very beneficial in this case. The only available method is "record", so why not use it directly?

```python
#Instead of 
viz = VideoViz(
    model,
    components,
    title="Schelling's Segregation Model",
)
viz.record(steps=50, filepath="schelling.mp4")

# Just do
record(
    model, 
    components, 
    title="Schelling's Segregation Model",
    steps=50,
    filepath="schelling,mp4
)

# This leaves more room for users to compose how they want. To replicate your interface
def my_record(steps, filepath):
    return record(
        model, 
        components, 
        title="Schelling's Segregation Model",
        steps,
        filepath
    )

my_record(50, "schelling.mp4")

# Or maybe you only want to change the model
def my_record(model):
    return record(
        model=model, 
        components=components, 
        title=f"${model.name} Segregation Model",
        steps=50,
        filepath=f"${model.name}.mp4"
    )
```

wang-boyu:
Now that https://github.com/projectmesa/mesa/pull/2447 is merged, I'll update this PR accordingly.

wang-boyu:
Thanks @Corvince for your feedback. I mainly wanted to have similar APIs in creating the different Viz "frontends":

```python
page = SolaraViz(
    model,
    [make_space_component(agent_portrayal=agent_portrayal),
     make_measure_component("happy")],
)

viz = VideoViz(
    model,
    [make_space_component(agent_portrayal=agent_portrayal),
     make_measure_component("happy")],
)
```

where the usages might differ:

```python
page   # simply use the page to render Solara webpage

viz.record(steps=50, filepath="schelling.mp4")
```

wang-boyu:
Rebased to main branch and updated for https://github.com/projectmesa/mesa/pull/2447:

1. Extract a `draw_plot` method similar to `draw_space` for plotting measures
2. Rename `mpl_space_drawing.py` to `mpl_drawing.py` since it now contains both `draw_space` and `draw_plot`
3. Update the implementation of `VideoViz` to use `draw_space` and `draw_plot`

quaquel:
I looked through the code and have at the moment a few high-level reactions

> 1. Extract a draw_plot method similar to draw_space for plotting measures

I don't mind this change. I am not convinced it is fully needed either. 

> 2. Rename mpl_space_drawing.py to mpl_drawing.py since it now contains both draw_space and draw_plot

Given 1. this makes perfect sense. 

> 3. Update the implementation of VideoViz to use draw_space and draw_plot

Here I have some questions. On the Solara side, we call things components because that is what they are in Solara. You borrow that terminology here, even though this code, as far as I can tell, will only work with matplotlib. So, does this component naming make sense, given that `VideoViz` only works with matplotlib? 

In my understanding of the API based on the typing information, a component is a callable that is called with the model and the ax as arguments. But do you need the complexity of videoviz specific `make_plot_component` and `make_space_component`? Looking at your code, there is some convenience in both functions, so there might be a good argument for keeping them. However, is calling it `component` the best choice then? Effectively these make_ functions wrap a draw function that takes the model and the ax as input.  

I understand your motivation for going OO to keep the similarity with SolaraViz. I don't mind doing it that way, but again, I wonder whether just accepting that it will be matplotlib specific might not open up a more generic and powerful API. 

For example, you hard-code the grid layout at the moment. What if the user wants to change this? For example, in a 16:9 aspect ratio slide, I might like to have a 3x2 grid, while on a web page, I might prefer 2x3. What about having a gridspec optional keyword argument on the `__init__`

To be clear, I needed to make a short mp4 of the mesa model today, and having this in would have saved me quite some time. I am just not yet fully convinced about some of the design choices made here.

wang-boyu:
Thanks for the comments!

> > 1. Extract a draw_plot method similar to draw_space for plotting measures
> 
> I don't mind this change, but I am not fully convinced it is needed.

If we want the same plot for measures in SolaraViz and VideoViz, without extracting a common method `draw_plot`, then there'll be two plotting functions that are essentially the same...?

> > 3. Update the implementation of VideoViz to use draw_space and draw_plot
> 
> Here I have some questions. On the Solara side, we call things components because that is what they are in Solara. You borrow that terminology here, even though this code, as far as I can tell, will only work with matplotlib. So, does this component naming make sense, given that `VideoViz` only works with matplotlib?

I'm hoping there could be more backend options for VideoViz in the future, other than matplotlib. It'll be easier to develop and integrate new backend into both SolaraViz and VideoViz with the current API design, hopefully.

> In my understanding of the API based on the typing information, a component is a callable that is called with the model and the ax as arguments. But do you need the complexity of videoviz specific `make_plot_component` and `make_space_component`? Looking at your code, there is some convenience in both functions, so there might be a good argument for keeping them. However, is calling it `component` the best choice then? Effectively these make_ functions wrap a draw function that takes the model and the ax as input.

Yes, these `make_` functions took the ideas from `make_mpl_space_component` and `make_mpl_plot_component` that are also wrappers of functions that take a model and return a Solara component. They differ in how they are created - `space` needs portrayal functions, and `plot` needs measure names.

> For example, you hard-code the grid layout at the moment. What if the user wants to change this? For example, in a 16:9 aspect ratio slide, I might like to have a 3x2 grid, while on a web page, I might prefer 2x3. What about having a gridspec optional keyword argument on the `__init__`?

Yeah it'll be good to allow users to configure grid layout, something I also wondered here: https://github.com/projectmesa/mesa/discussions/2236#discussioncomment-10873425. I would prefer some common ways of addressing this, and a common API between SolaraViz and VideoViz would benefit from it, e.g.:

```python
# `make_` functions are imported from mesa.visualization
layout = SomeLayoutCustomization(
        make_space_component(agent_portrayal=agent_portrayal),
        make_measure_component("happy")
)

page = SolaraViz(model, layout)

# here `make_` functions are imported from mesa.visualization.video_viz
layout = SomeLayoutCustomization(
        make_space_component(agent_portrayal=agent_portrayal),
        make_measure_component("happy")
)
viz = VideoViz(model, layout)
```

so that when users replace `SolaraViz` with `VideoViz`, they can expect the same layout to be exported in a video.

I'm also wondering whether it'll be possible to make the `make_` functions even more generic - return Solara components when they are used in SolaraViz, and return video components when used in VideoViz. This way users don't need to change where they are imported.

Or maybe something even further:

```python
model_explorer = ModelExplorer(
    model,
    SomeLayoutCustomization(
        make_space_component(agent_portrayal=agent_portrayal),
        make_measure_component("happy")
    ),
)

# renders Solara app if used in Jupyter notebook or called in `solara run app.py`
model_explorer

# exports a video
model_explorer.record(steps=10, filepath="my_amazing_model.mp4")
```

But I guess this is beyond the scope of this PR? Apologies for being off topic. But this is also close to what @Corvince proposed above, to have just a single `record()` function.

quaquel:
Thanks for the extensive reaction. 

> If we want the same plot for measures in SolaraViz and VideoViz

This is the core of the discussion. Do we want to make it trivial to swap between the UI and recording a movie? I see great value in this for ease of use and API simplicity. So, ideally, all you would have to change is `VideoViz` to `SolaraViz`, while the rest stays the same. Or indeed, as @wang-boyu describes at the end, you might even collapse it all into a single `ModelExplorer class`. While this is an ideal to aim at, I am not convinced that it is achievable in part because of the second point.

> I'm hoping there could be more backend options for VideoViz in the future, other than matplotlib.

Do we have any description of creating mp4's with altair or plotly (a possible other backend supported in solara)? If we have, then I am less sceptical about the potential of having `VideoViz` and `SolaraViz` working similarly. However, a quick search did not turn up anything for Altair. 

If we don't have any clear direction for supporting both Altair and matplotlib in VideoViz, it might actually be confusing to a user to have the same API for both `VideoViz` and `SolaraViz` even though the former only supports a single backend. Also, for those familiar with matplotlib, which are most scientific users, I assume the entire API might become too restrictive. 

Playing devil's advocate, another separate reason for not bothering with the same API for VideoViz` and `SolaraViz` is that you can do a simple screen recording of the UI. So, what is the value?

wang-boyu:
> you can do a simple screen recording of the UI

One motivation for me is that I will need to run the same model with many different parameter settings, and I don't want to do manual recordings for all of them.

Back on the APIs, I think the benefit of current design is that it is open for future backends, without making breaking changes.

What are the proposed changes to the APIs? I'm a bit lost in these discussions : ) Is it to change to a single `record()` function (still with the `make_x_component` functions) as suggested above?

quaquel:
> One motivation for me is that I will need to run the same model with many different parameter settings, and I don't want to do manual recordings for all of them.

Fair point and good motivation

> Back on the APIs, I think the benefit of current design is that it is open for future backends, without making breaking changes.

Ok, but I still would like to understand the plausibility of this being realized in the future. It is great to design for future extensions, but if those future extensions are not plausible or the API becomes convoluted, why bother?

> What are the proposed changes to the APIs? I'm a bit lost in these discussions : ) Is it to change to a single record() function (still with the make_x_component functions) as suggested above?

I haven't discussed or commented on the actual API in much detail yet. For me, the second point here is the key. If there is a plausible path to mp4 with Altair, I would comment on this API differently compared to a situation where the focus is on matplotlib only. 

(and sorry for being perhaps a bit annoying, I really like the functionality provided here, but since it is new functionality, I believe it is worth it to have a slightly longer back and forth on how to best provide this functionality).

EwoutH:
@wang-boyu what would you like to do with this PR? Can I help you move this forward?

wang-boyu:
Yes! @EwoutH Thank you for looking into this matter.

Most of the implementation should be ready. The main remaining question is about the API. So far my favourite is (other than what was in this PR):

```python
page = SolaraViz(
    model,
    SomeLayoutCustomization(
        make_space_component(agent_portrayal=agent_portrayal),
        make_measure_component("happy")
    ),
)

# renders Solara app if used in Jupyter notebook or called in `solara run app.py`
page

# exports a video
page.record(steps=10, filepath="my_amazing_model.mp4")
```

In this way users don't have to repeatedly define `components` and use them in SolaraViz and video recordings. The downside is that (as @quaquel mentioned) currently videos only support matplotlib backend. This may confuse users if they have SolaraViz with altair backend, while recordings are saved differently using matplotlib backend.

@Corvince also suggested APIs above.

Please feel free to work on this if you like!

EwoutH:
Now that @Sahil-Chhoker’s GSoC project is finished, should we revisit this?

I think `page.record()` makes a lot of sense. Is it possible to view and record at the same time?

quaquel:
I still like the idea of built-in support for making movies, but this code is no longer aligned with the main, will be non-trivial to align again, and we seem to have different ideas on how to best add this feature. So, I am closing this PR.

PR #2454, opened 2024-11-04
Fix #2452 - handle solara viz model params better

### LINKED ISSUE #2452: Bugs in SolaraViz?
I started working on #2427 and have run into three issues:

1. If the model takes one or more arguments and those arguments are not part of the user-controllable parameters that are set from the UI, stuff will crash immediately 
2. Strangely, it crashes immediately even though we started SolaraViz with a valid model instance, so this points to a second bug: at the moment, the model shown after starting the UI is not the same as the original model with which the UI was initialized. Again, this is easy to reproduce. In the boltzman wealth model, in app.py, start the model with say a width and height of 20. Next, start Solara, and in the GUI, you'll see that the actual model has reverted back to 10 by 10 (as specified in `model_params`). 
3. Seed is currently handled differently from other parameters. With #1940 merged, this is no longer needed, and seed can just be integrated into the user-specifiable parameters.

Corvince:
## Summary
Bug description is in #2452, here is a copy
> 1. If the model takes one or more arguments and those arguments are not part of the user-controllable parameters that are set from the UI, stuff will crash immediately
> 2. Strangely, it crashes immediately even though we started SolaraViz with a valid model instance, so this points to a second bug: at the moment, the model shown after starting the UI is not the same as the original model with which the UI was initialized. Again, this is easy to reproduce. In the boltzman wealth model, in app.py, start the model with say a width and height of 20. Next, start Solara, and in the GUI, you'll see that the actual model has reverted back to 10 by 10 (as specified in `model_params`).
> 3. Seed is currently handled differently from other parameters. With [model: Move random seed and random to __init__ #1940](https://github.com/projectmesa/mesa/pull/1940) merged, this is no longer needed, and seed can just be integrated into the user-specifiable parameters.

## Implementation
1. On startup solara viz will check the user_params and compare to `model.__class__.__init__`. If there are are parameters that don't have a default value but are not in user_params solara viz will raise a ValueError. Also raises ValueError if user provided model_params include parameters not used by the model.
2. Currently the ModelCreator was creating a new instance whenever model parameters changed. However, the initial call to ModelCreator "changed" the parameters, triggering an update and overwriting the provided model. With this PR, only actual parameter changes trigger a model re-creation
3. "Reactive" Seed handling was removed completely. No its treated just like any other model parameter.

### DIFF
--- a/mesa/examples/basic/boltzmann_wealth_model/model.py
+++ b/mesa/examples/basic/boltzmann_wealth_model/model.py
@@ -10,7 +10,7 @@ class BoltzmannWealthModel(mesa.Model):
     highly skewed distribution of wealth.
     """
 
-    def __init__(self, hans=213, n=100, width=10, height=10, seed=None):
+    def __init__(self, n=100, width=10, height=10, seed=None):
         super().__init__(seed=seed)
         self.num_agents = n
         self.grid = mesa.space.MultiGrid(width, height, True)
--- a/mesa/examples/basic/boltzmann_wealth_model/model.py
+++ b/mesa/examples/basic/boltzmann_wealth_model/model.py
@@ -10,7 +10,7 @@ class BoltzmannWealthModel(mesa.Model):
     highly skewed distribution of wealth.
     """
 
-    def __init__(self, seed, n=100, width=10, height=10):
+    def __init__(self, hans=213, n=100, width=10, height=10, seed=None):
         super().__init__(seed=seed)
         self.num_agents = n
         self.grid = mesa.space.MultiGrid(width, height, True)
--- a/mesa/examples/basic/boltzmann_wealth_model/model.py
+++ b/mesa/examples/basic/boltzmann_wealth_model/model.py
@@ -10,7 +10,7 @@ class BoltzmannWealthModel(mesa.Model):
     highly skewed distribution of wealth.
     """
 
-    def __init__(self, n=100, width=10, height=10, seed=None):
+    def __init__(self, seed, n=100, width=10, height=10):
         super().__init__(seed=seed)
         self.num_agents = n
         self.grid = mesa.space.MultiGrid(width, height, True)
--- a/mesa/visualization/solara_viz.py
+++ b/mesa/visualization/solara_viz.py
@@ -94,13 +94,6 @@ def SolaraViz(
     if not isinstance(model, solara.Reactive):
         model = solara.use_reactive(model)  # noqa: SH102, RUF100
 
-    model_class_arguments = inspect.signature(model.value.__class__).parameters
-    for k in model_class_arguments:
-        if k not in model_params:
-            if model_class_arguments[k].default == inspect.Parameter.empty:
-                print(f"Missing parameter: {k}")
-                return solara.Text(f"Missing parameter: {k}")
-
     def connect_to_model():
         # Patch the step function to force updates
         original_step = model.value.step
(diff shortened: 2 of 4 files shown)


quaquel:
> Potentially allows us to remove deepcopy for reinitializing a model.

I had a look at this yesterday and my approach would be to change `do_reset` from using the original model to create a new model instance based on the current values of model_parameters. This requires some moving around of code. Basically the stuff before `on_change` in `ModelCreator` would move up to `SolaraViz` such that the reactive version of `model_parameters` is available to `ModelController`. Hope this makes sense, happy to open a PR with my attempted fix if you want a closer look at the code.

mesa/examples/basic/boltzmann_wealth_model/model.py
```diff
@@ -10,7 +10,7 @@ class BoltzmannWealthModel(mesa.Model):
     highly skewed distribution of wealth.
     """
 
-    def __init__(self, n=100, width=10, height=10, seed=None):
+    def __init__(self, seed, n=100, width=10, height=10):
```
quaquel, line 13:
```suggestion
    def __init__(self, seed=None, n=100, width=10, height=10):
```

quaquel:
A view minor comments, but great to see you working on this.

mesa/examples/basic/boltzmann_wealth_model/model.py
Corvince: This change was just for testing, removed now

mesa/visualization/solara_viz.py
```diff
@@ -93,6 +94,13 @@ def SolaraViz(
     if not isinstance(model, solara.Reactive):
         model = solara.use_reactive(model)  # noqa: SH102, RUF100
 
+    model_class_arguments = inspect.signature(model.value.__class__).parameters
+    for k in model_class_arguments:
+        if k not in model_params:
+            if model_class_arguments[k].default == inspect.Parameter.empty:
+                print(f"Missing parameter: {k}")
+                return solara.Text(f"Missing parameter: {k}")
+
```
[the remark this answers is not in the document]
Corvince: I updated the checking logic. A ValueError is now raised if either a non-default value argument is missing or a parameters is provided that is not used by the model.

mesa/visualization/solara_viz.py
```diff
@@ -301,36 +309,18 @@ def ModelCreator(model, model_params, seed=1):
...
-
-    with solara.Row(justify="space-between"):
-        solara.InputText(
-            label="Seed",
-            value=reactive_seed,
-            continuous_update=True,
-        )
-
-        solara.Button(label="Reseed", color="primary", on_click=do_reseed)
+        print(f"Setting {name} to {value}")
```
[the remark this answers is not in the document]
Corvince: Added by copilot, but I liked the bit of information added to the console. However, you are right and logging should be approached separately and more holistically. Thus, removed it for now

Corvince:
Updated the logic, added tests and reverted temporary example changes. Also moved the checking to `ModelCreator` where it is actually needed. Ready for review

mesa/examples/basic/boltzmann_wealth_model/model.py
```diff
@@ -10,7 +10,7 @@ class BoltzmannWealthModel(mesa.Model):
     highly skewed distribution of wealth.
     """
 
-    def __init__(self, n=100, width=10, height=10, seed=None):
+    def __init__(self, hans=213, n=100, width=10, height=10, seed=None):
```
quaquel, line 13:
```suggestion
    def __init__(self, n=100, width=10, height=10, seed=None):
```

quaquel (approved):
One suggestions (probably a left over from testing) and then this is good to be merged.

Issue #2457, opened 2024-11-05
#2452 broke the visualization tutorial


quaquel:
The visualization tutorial is broken ATM:  https://mesa.readthedocs.io/latest/tutorials/visualization_tutorial.html

quaquel:
closed via #2458

PR #2458, opened 2024-11-05
Update MoneyModel.py

quaquel:
attempted fix for #2457


EwoutH:
Does it work like expected?

quaquel:
yes

EwoutH (approved):

PR #2459, opened 2024-11-05
Mark SolaraViz as experimental for Mesa 3.0

EwoutH:
As discussed in today's meeting, this PR adds a note at four places that marks SolaraViz as experimental in Mesa 3.0.

### DIFF
--- a/docs/migration_guide.md
+++ b/docs/migration_guide.md
@@ -253,6 +253,8 @@ Ref: Original discussion [#1912](https://github.com/projectmesa/mesa/discussions
 Mesa has adopted a new API for our frontend. If you already migrated to the experimental new SolaraViz you can still use
 the import from mesa.experimental. Otherwise here is a list of things you need to change.
 
+> **Note:** SolaraViz is experimental and still in active development for Mesa 3.0. While we attempt to minimize them, there might be API breaking changes between Mesa 3.0 and 3.1. There won't be breaking changes between Mesa 3.0.x patch releases.
+
 #### Model Initialization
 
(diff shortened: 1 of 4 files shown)


wang-boyu:
Sorry that I missed many meetings recently due to my work and class schedule. Here we're marking SolaraViz as experimental without moving it into the experimental namespace right?

EwoutH:
Yes, we won’t want to break namespaces again.

We will attempt not to break API, but since it’s such active development we want to give ourselves a bit of room.

To be honest, I also don’t like it, because this means shipping Mesa 3.0 without formally stable visualization, but I do feel this is the better long term option.

Corvince:
I share this concern. "Experimental" sounds quite unstable to me. What about the idea of just declaring the components experimental? That would mean we ship a stable SolaraViz (which API I think is indeed stable, I don't know that we should change about it in a breaking way). We just say that creating the components is a bit up to the user, but we provide these experimental functions users can use and will likely not break, but might. (Again, I don't think the API of make_space_component needs to change in a backwards-incompatible way, but I am fine with declaring it experimental for now). So, to be super safe users could just code up their own components and have a stable frontend.

But yeah, also sorry that I wasn't at the meeting. 


One last point: we always wanted to rename SolaraViz to ???. Is this still a thing (by now we should at least alias it to SolaraViz, but we could still do a rename)

EwoutH:
I’ve a really conflicted opinion here, so I’m not the one to argue for this. I’m okay with any consensus that’s reached.

The only thing I find really important that we release Mesa 3.0 this week.

Corvince:
Yes, this is the most important thing. Agree, if there already is an agreement we should probably just go with it.

quaquel:
I suggest we just call it experimental and revisit that decision for 3.1. This gives us ample time to pick up various outstanding issues as well as ideas we individually might have about future development. If we can solve the open issues and implement some of these ideas without braking the existing API by 3.1. We can declare it stable. However, for it to be called stable, I really want to have proper test code coverage again for all of it.

quaquel (approved):

EwoutH:
If there are no further objections I will merge in an hour or so, an then release the Mesa 3.0 Release candidate.

PR #2460, opened 2024-11-05
Bugfix for Solara deepcopy bug

### LINKED ISSUE #2427: unable to start solara app when model cannot be pickled
**Describe the bug**

When a model or agents have attributes that cannot be pickled, Solara app won't start due to

```bash
TypeError: cannot pickle '_thread.lock' object
```

**Expected behavior**

Should be able to run model with Solara regardless it can be deep copied or not

**To Reproduce**

Take the Schelling model as an example, add the following into the `__init__()` method in `model.py`:

```python
import threading
self.t = threading.Thread(target=print, args=("Hello"))
self.t.start()
```

Then run `solara run app.py` in the command line.

**Additional context**

It appears to be related to https://github.com/projectmesa/mesa/blob/1d985608db1abe346a6e817bf5bb12c18cad556d/mesa/visualization/solara_viz.py#L188 and https://github.com/projectmesa/mesa/blob/1d985608db1abe346a6e817bf5bb12c18cad556d/mesa/visualization/solara_viz.py#L212

where we try to create deep copies of the model object.

If we remove all usages related to deepcopy:

```diff
diff --git a/mesa/visualization/solara_viz.py b/mesa/visualization/solara_viz.py
index 4bca98d..c2d4960 100644
--- a/mesa/visualization/solara_viz.py
+++ b/mesa/visualization/solara_viz.py
@@ -183,14 +183,6 @@ def ModelController(model: solara.Reactive[Model], play_interval=100):
     running = solara.use_reactive(True)
     original_model = solara.use_reactive(None)
 
-    def save_initial_model():
-        """Save the initial model for comparison."""
-        original_model.set(copy.deepcopy(model.value))
-        playing.value = False
-        force_update()
-
-    solara.use_effect(save_initial_model, [model.value])
-
     async def step():
         while playing.value and running.value:
             await asyncio.sleep(play_interval / 1000)
@@ -205,18 +197,11 @@ def ModelController(model: solara.Reactive[Model], play_interval=100):
         model.value.step()
         running.value = model.value.running
 
-    def do_reset():
-        """Reset the model to its initial state."""
-        playing.value = False
-        running.value = True
-        model.value = copy.deepcopy(original_model.value)
-
     def do_play_pause():
         """Toggle play/pause."""
         playing.value = not playing.value
 
     with solara.Row(justify="space-between"):
-        solara.Button(label="Reset", color="primary", on_click=do_reset)
         solara.Button(
             label="▶" if not playing.value else "❚❚",
             color="primary",
```

Then the Solara app runs but without the `Reset` button.

Probably need a better way to create or save the initial model object?

quaquel:
This removes the deepcopy operation from SolaraViz. This changes the behavior. The current behavior is that when reset is hit, the model is reset to the original model passed when initializing SolaraViz. Getting rid of deepcopy implies that this behavior cannot be maintained. So, I replace it here by resetting the model to whatever the current values are for the model parameters (i.e., the user controllable settings, combined with the passed fixed values). 

I also added (but this can be its own PR) an extra user input field for a TextInput. This allows for recreating the seed input that we had before #2454.

If we agree on this PR, we might revert #2378. 

In passing, I fixed another bug: if model_params was not specified SolaraViz would error in _check_params. 

closes #2427

### DIFF
--- a/mesa/examples/basic/boltzmann_wealth_model/app.py
+++ b/mesa/examples/basic/boltzmann_wealth_model/app.py
@@ -42,7 +42,7 @@ def post_process(ax):
 
 
 # Create initial model instance
-model1 = BoltzmannWealthModel(50, 10, 10)
+model = BoltzmannWealthModel(50, 10, 10)
 
 # Create visualization elements. The visualization elements are solara components
 # that receive the model instance as a "prop" and display it in a certain way.
@@ -61,7 +61,7 @@ GiniPlot = make_plot_component("Gini")
 # solara run app.py
 # It will automatically update and display any changes made to this file
 page = SolaraViz(
-    model1,
+    model,
     components=[SpaceGraph, GiniPlot],
     model_params=model_params,
     name="Boltzmann Wealth Model",
--- a/mesa/examples/basic/boltzmann_wealth_model/app.py
+++ b/mesa/examples/basic/boltzmann_wealth_model/app.py
@@ -1,3 +1,7 @@
+import sys
+import os.path
+sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), '../../../../')))
+
 from mesa.examples.basic.boltzmann_wealth_model.model import BoltzmannWealthModel
 from mesa.visualization import (
     SolaraViz,
@@ -19,6 +23,11 @@ model_params = {
         "min": 10,
         "max": 100,
         "step": 1,
+    },
+    "seed": {
+        "type": "InputText",
+        "value": 42,
+        "label": "Random Seed",
     },
     "width": 10,
     "height": 10,
--- a/mesa/examples/basic/boltzmann_wealth_model/app.py
+++ b/mesa/examples/basic/boltzmann_wealth_model/app.py
@@ -1,10 +1,3 @@
-import os.path
-import sys
-
-sys.path.insert(
-    0, os.path.abspath(os.path.join(os.path.dirname(__file__), "../../../../"))
-)
-
 from mesa.examples.basic.boltzmann_wealth_model.model import BoltzmannWealthModel
 from mesa.visualization import (
     SolaraViz,
--- a/mesa/examples/basic/boltzmann_wealth_model/app.py
+++ b/mesa/examples/basic/boltzmann_wealth_model/app.py
@@ -1,6 +1,9 @@
-import sys
 import os.path
-sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), '../../../../')))
+import sys
+
+sys.path.insert(
+    0, os.path.abspath(os.path.join(os.path.dirname(__file__), "../../../../"))
+)
 
 from mesa.examples.basic.boltzmann_wealth_model.model import BoltzmannWealthModel
 from mesa.visualization import (
--- a/mesa/visualization/solara_viz.py
+++ b/mesa/visualization/solara_viz.py
@@ -298,6 +298,7 @@ def ModelCreator(model, user_params, model_parameters):
         - The component provides an interface for adjusting user-defined parameters and reseeding the model.
 
     """
+
     def on_change(name, value):
         new_model_parameters = {**model_parameters.value, name: value}
         model.value = model.value.__class__(**new_model_parameters)
--- a/mesa/visualization/solara_viz.py
+++ b/mesa/visualization/solara_viz.py
@@ -84,7 +84,7 @@ def SolaraViz(
           value results in faster stepping, while a higher value results in slower stepping.
     """
     if components == "default":
-        components = [components_altair.make_space_altair()]
+        components = [components_altair.make_altair_space()]
     if model_params is None:
         model_params = {}
 
@@ -108,26 +108,16 @@ def SolaraViz(
     solara.use_effect(connect_to_model, [model.value])
 
     # set up reactive model_parameters shared by ModelCreator and ModelController
-    solara.use_effect(
-        lambda: _check_model_params(model.value.__class__.__init__, fixed_params),
-        [model.value],
-    )
-    user_params, fixed_params = split_model_params(model_params)
-    model_parameters = solara.use_reactive(
-        {
-            **fixed_params,
-            **{k: v.get("value") for k, v in user_params.items()},
-        }
-    )
+    reactive_model_parameters = solara.use_reactive({})
 
     with solara.AppBar():
         solara.AppBarTitle(name if name else model.value.__class__.__name__)
 
     with solara.Sidebar(), solara.Column():
         with solara.Card("Controls"):
-            ModelController(model, model_parameters, play_interval=play_interval)
+            ModelController(model, model_parameters=reactive_model_parameters, play_interval=play_interval)
         with solara.Card("Model Parameters"):
-            ModelCreator(model, user_params, model_parameters)
+            ModelCreator(model, model_params, model_parameters=reactive_model_parameters)
         with solara.Card("Information"):
             ShowSteps(model.value)
 
@@ -273,7 +263,7 @@ def check_param_is_fixed(param):
 
 
 @solara.component
-def ModelCreator(model, user_params, model_parameters):
+def ModelCreator(model: solara.Reactive[Model], user_params: dict, model_parameters: dict | solara.Reactive[dict] = None):
     """Solara component for creating and managing a model instance with user-defined parameters.
 
     This component allows users to create a model instance with specified parameters and seed.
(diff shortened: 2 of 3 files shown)


Corvince:
Thanks, please let me review/comment before merging this (hopefully later tonight I'll have some time)

mesa/examples/basic/boltzmann_wealth_model/app.py
```diff
@@ -1,3 +1,10 @@
+import os.path
+import sys
+
+sys.path.insert(
+    0, os.path.abspath(os.path.join(os.path.dirname(__file__), "../../../../"))
```
EwoutH, line 5: We need some policy if we want to include this in every example, or handle it differently.

quaquel: it simply should not be there and we need to find a cleaner way to run these examples from the command line. Not critical for 3.0 but a prioirty afterwards

quaquel:
> Thanks, please let me review/comment before merging this (hopefully later tonight I'll have some time)

Of course.

Corvince:
Okay, so I am still a bit undecided on what to think about this. This is now pretty much how [the experimental solara_viz](https://github.com/projectmesa/mesa/blob/main/mesa/experimental/solara_viz.py) approached this problem. But one of the reasons for the rewrite was to have more separated control, thus moving the reinstantiating logic into "ModelCreator". But there is also an argument to be made about "SolaraViz" being the smart component handling all the logic and the other components being dumb. But I lean towards not liking a monolithic approach that much. I don't find the removal of "ModelCreator" very bad, but passing model_params to the ModelController (which purpose is to mainly run the model)... I don't like that very much and I think it hurts reusability. Without this PR you could just pass it a model and control your model interactively (e.g. in a notebook). Now you also have to think about how to instantiate it. 

Another reason to use deepcopy, which I unfortunately forgot to mention before, was that at some soon point (hopefully for 3.1) I wanted to introduce the "timeline control" from mesa-viz, which allows users to not just step forward in time, but also to get back to earlier steps. But this requires serializing the model some way or another and just deepcopying is a very fast and simple solution (and if you can't deepcopy an object its also very hard to serialize it another way). Which is a convoluted way of saying: SolaraViz not working with non-deepcopyable objects is maybe not a bug, but just a limitation. And I don't think its a big one. For example #2427 mentions binding *and starting* a thread in a models `__init__` function. I don't think thats a good idea in the first place and very likely solvable in another way.

Corvince:
Maybe I shouldn't write comments late at night. Here is a more productive comment: 

What about keeping ModelCreator, but just declaring model_parameters as a solara.reactive variable inside SolaraViz. Then we can keep the logic inside ModelCreator, but SolaraViz also has the current state. And my comment about not wanting to pass model_params to ModelController was a bit nonsensical. It can be an optional parameter and if it is not provided just fall back to an empty dict. That's enough to restart the model if it has default values.

And we really want to reset the model by recreating it from scratch. Otherwise we keep the bug where the model just resets to the initial state, which might not be step 0.

With these changes I would be in approval with this PR

quaquel:
Thanks for both comments @Corvince.

> But one of the reasons for the rewrite was to have more separated control, thus moving the reinstantiating logic into "ModelCreator". 

I was indeed wondering about that, but I did not see the point of `ModelCreator` after having moved over all the code preceding the creation of UserParam. 


> What about keeping ModelCreator, but just declaring model_parameters as a solara.reactive variable inside SolaraViz.

This is indeed the cleaner solution and I'll implement that here. 


> But I lean towards not liking a monolithic approach that much. I don't find the removal of "ModelCreator" very bad, but passing model_params to the ModelController (which purpose is to mainly run the model)... I don't like that very much and I think it hurts reusability.

I agree. This is, in part, why I felt uncomfortable with removing `ModelCreator.` 


> Which is a convoluted way of saying: SolaraViz not working with non-deepcopyable objects is maybe not a bug, but just a limitation. And I don't think its a big one.

If we can avoid making that decision now, that would be my preference. Many other things can leave an object in an unpickleable state. For example, generators, nested functions, lambda's, various descriptors, etc. If we need a way of serializing the model at some future point, we might look into using `dill` instead of `pickle`, although personally, I would prefer finding a cleaner solution than either one. This issue is directly related to caching a mesa run, and I believe that the observable/signal stuff might be a cleaner route for this. We should discuss this separately. 

A separate reason for changing the behavior of SolaraViz is that, from a UI point of view, it makes more sense to me that reset is in line with the user-settable parameters rather than whatever state the model was in when starting the UI. For example, you start with a 20 x 20 grid, but the model_params uses 10 x 10. Everytime you hit reset, you now would go back to 20 x 20, but there is no way to interact with that version of the model because the moment you change any slider or so, it reverts to 10 x 10. This seems strange behavior to me.

Corvince:
Okay, than I think we are in line with how to update this PR! Great

> A separate reason for changing the behavior of SolaraViz is that, from a UI point of view, it makes more sense to me that reset is in line with the user-settable parameters rather than whatever state the model was in when starting the UI.

Exactly, this is what I was refering to with 
>And we really want to reset the model by recreating it from scratch. Otherwise we keep the bug where the model just resets to the initial state, which might not be step 0.

So, yeah, independent of the deepcopy/pickle subject this should be changed in line with how this PR does it (just making model_params an optional argument). (On the discussion itself, yes, should be done somewhere else, but note that deepcopy and pickle are different - pickle is much more restrictive in what can be pickled)

quaquel:
I think I figured it out, but @Corvince please check the changes in `solara_vis` to make sure I don't do anything stupid. I have tested it with Boltzman and it works as intended. I'll fix the rest including docs later. 

In short:
1. ModelCreator is back.
2. I create a reactive variable based on `model_parameters`. This is done in SolaraViz and then shared between `ModelCreator` and `ModelController`.
3. However, this does require moving the parameter check (i.e., `_check_model_params`) and splitting (i.e., split_model_params) to SolaraViz. It might be possible to keep these in `ModelCreator` but that would make creating the reactive variable in SolaraViz a bit more complicated. @Corvince let me know your thoughts.
4. ModelCreator takes the reactive version of `model_parameters` and the user settable parameters.

ps. `check_param_is_fixed` might be a source for bugs. Imagine a model that takes a dict as a keyword argument. It is currently not possible to pass a default fixed value for this dict to SolaraViz because `check_param_is_fixed ` assumes that all dicts are specifications for user interaction items. I think the longer-term objective would be to move away from purely dict-based specifications of interaction elements (and agent_portrayal, property_layer portrayal, etc.).

mesa/visualization/solara_viz.py
```diff
@@ -173,24 +178,17 @@ def ComponentsView(
 
 
 @solara.component
-def ModelController(model: solara.Reactive[Model], play_interval=100):
+def ModelController(model: solara.Reactive[Model], model_parameters, play_interval=100):
```
Corvince, line 181: Can you insert a `*` here to make model_parameters and play_interval keyword only? Less arguing about the order of arguments then. Also: can we default this to None and then create an empty (reactive) dict if not provided?

mesa/visualization/solara_viz.py
```diff
@@ -109,20 +107,27 @@ def step():
 
     solara.use_effect(connect_to_model, [model.value])
 
+    # set up reactive model_parameters shared by ModelCreator and ModelController
+    solara.use_effect(
```
Corvince, line 111: On your question about the "need" to move this here: Can't you just initialize model_parameters as an empty dict here? 
`model_parameters = solara.reactive({})` And then the rest of the logic can stay in ModelCreator

Corvince:
Looking good, 2 more comments. Also like the additional small fixes you added  (cleanup of seed,)

quaquel:
I made the requested changes. I also added some additional type hinting and changed a call to a deprecated function (from the API reorganization) with the updated correct function call. I again tested various examples and they seem to be working correctly.

mesa/visualization/solara_viz.py
```diff
@@ -173,24 +174,23 @@ def ComponentsView(
 
 
 @solara.component
-def ModelController(model: solara.Reactive[Model], play_interval=100):
+def ModelController(
```
Corvince, line 177: Awesome work! Maybe more of a clarification question: I previously suggested adding a `*` here to have keyword-only arguments. Maybe I am wrong here and thats why you ignored it, but my understanding was that this is how it works
```python
def foo(bar, baz=2):
    ...
# all of this works:
foo(1)
foo(3, baz=4)
foo(5, 6)
# Compared to
def foo2(bar, *, baz):
    pass

foo2(1) # works
foo2(3, baz=4) # works
foo2(3, 4) # doesnt work
```
My thinking behind this was to have API stability. If we later change the signature to
```python
def ModelController(model, awesome_new_arg, model_parameters=None)
```
we don't have a breaking change. right now it would be breaking if model_parameters is only provided as a positional argument.

Does that make sense? Is that how it works?

Corvince (approved):

mesa/visualization/solara_viz.py
quaquel: I was wondering what you ment, but this makes good sense. I'll update it accordingly.

Corvince: Great, feel free to merge afterwards

PR #2473, opened 2024-11-09
Integrate benchmarks and example models

EwoutH:
This PR integrates the benchmark models into the example models and runs the benchmarks on the example models.

The first commit moves over the benchmark scripts to run on the example models, the following commits merge the benchmark models into the example models one-by-one.


mesa/examples/advanced/wolf_sheep/model.py
```diff
@@ -9,129 +9,131 @@
...
+            energy = self.random.randrange(2 * wolf_gain_from_food)
+            Wolf(self, energy, wolf_reproduce, wolf_gain_from_food, self.grid[pos])
+
+        # Create grass patches if enabled
+        if grass:
+            possibly_fully_grown = [True, False]
+            for cell in self.grid:
+                fully_grown = self.random.choice(possibly_fully_grown)
+                countdown = (
+                    0 if fully_grown else self.random.randrange(grass_regrowth_time)
```
quaquel, line 115:
```suggestion
                    0 if fully_grown else self.random.randrange(0, grass_regrowth_time)
```

see my point on this in #2470

mesa/examples/advanced/wolf_sheep/model.py
```diff
@@ -9,129 +9,131 @@
...
         self.datacollector.collect(self)
 
     def run_model(self, step_count=200):
+        """Run the model for a specified number of steps.
+
+        Args:
+            step_count: Number of steps to run the model
+        """
         for _ in range(step_count):
             self.step()
```
quaquel, line 139:
```suggestion
```

since running is done via the simulator, this won't work anymore and can be removed.

EwoutH:
Ready for review. Note that wolf_sheep uses experimental features like a ABMSimulator and the Cell space, while the other models are basic models only using stable features.

I added a lot of docstring in the process (maybe too much, let me know). Rendered docs: https://mesa--2473.org.readthedocs.build/2473/examples.html

quaquel:
just a quick clarifying question: this will break app.py for wolf-sheep until #2470 is merged. So, what is your envisioned timeline on both?

EwoutH:
Fair, didn't think about that yet. We can do a few things:

1. Split off wolf_sheep from this PR
2. Wait for #2470
3. Merge after 3.0-branch cutoff
4. Accept one of our advanced examples has temporarily broken visualisation

It depends I think if #2470 will be API breaking of not and when its ready. If it's non-breaking, it can go into 3.0.1.

quaquel:
As indicated in #2470, it won't be API breaking. It is also complete accept for the startup problem. I just have to decide on how to address that. I might be able to put a monkey patch in place. Once #2291 is merged, we can then go back in an do it cleaner for 3.1. All of this would still be implementatin changes, no API changes.

EwoutH:
I propose merging this as is, and following up the visualisation with the SolaraViz PR. Then that PR can directly demonstrate how an model or app.py needs to be updated to work with such a visualisation (if at all).

quaquel:
i'll try to take a look at the other examples later today.

Corvince:
> We will tag 3.0.1 when both are merged, so our stable docs keep working the whole time.

At first I thought we shouldn't be so lenient with semantic versioning, since #2470 actually changes the API and thus should be 3.1. But it think both changes of 2470 can also be considered as bug fixes. Removing the monkey patch is good since we then no longer change the underlying model. And supporting DEVS is something that should have also been possible in 3.0, so I am fine with releasing as 3.0.1.

Corvince:
And It would be great to get it through this weekend. But after that let's be a bit more stringent with semantic versioning. No problem if we already reach something like 3.4 before the end of the year.

EwoutH:
Agree on both points!

quaquel:
> At first I thought we shouldn't be so lenient with semantic versioning, since https://github.com/projectmesa/mesa/pull/2470 actually changes the API and thus should be 3.1.

Just a clarifying question: when is something an API change? #2470 adds a keyword argument to `SolaraViz`, but past calls are still valid. Likewise, I don't change anything on the outside of devs, just slight changes in behavior inside the methods.

EwoutH:
[Official spec](https://semver.org/) says:

> Patch version Z (x.y.Z | x > 0) MUST be incremented if only backward compatible bug fixes are introduced. A bug fix is defined as an internal change that fixes incorrect behavior.
> 
> Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backward compatible functionality is introduced to the public API. It MUST be incremented if any public API functionality is marked as deprecated. It MAY be incremented if substantial new functionality or improvements are introduced within the private code. It MAY include patch level changes. Patch version MUST be reset to 0 when minor version is incremented.
> 
> Major version X (X.y.z | X > 0) MUST be incremented if any backward incompatible changes are introduced to the public API. It MAY also include minor and patch level changes. Patch and minor versions MUST be reset to 0 when major version is incremented.

But breaking things is the thing the main thing you don't want to do. Considering:
- We declared SolaraViz experimental
- The Cell Space is experimental
- We break nothing
- 3.0 has been out for an extremely short time
 
We're fine doing it in a patch instead of an minor release.

But yeah, something like a minor release a month, there's nothing wrong with that.

Corvince:
> Just a clarifying question: when is something an API change? #2470 adds a keyword argument to SolaraViz, but past calls are still valid. Likewise, I don't change anything on the outside of devs, just slight changes in behavior inside the methods.

To be a bit more specific in this example. Adding an additional keyword is changing the public API. But it's fully backwards compatible, so a x.Y.z. change and not a X.y.z. change (which would be breaking). Patch version changes (x.y.Z.) are when the public API stays the same but the internal mechanism is changed in a way to achieve the actually documented behavior.

mesa/examples/basic/boid_flockers/agents.py
```diff
@@ -47,25 +54,49 @@ def __init__(
         self.cohere_factor = cohere
         self.separate_factor = separate
         self.match_factor = match
-        self.neighbors = None
 
     def step(self):
         """Get the Boid's neighbors, compute the new vector, and move accordingly."""
```
quaquel, line 59: why these extensive changes in this step method?

quaquel (approved):
looks fine to me

Will be a bit annoying depending on whether this one or #2470 is merged first given that they both change wolf-sheep

mesa/examples/basic/boid_flockers/agents.py
EwoutH: Mainly performance, a lot of the boid might not have any neighbors at some time, which is where the early exit might add a lot of performance.

EwoutH:
Thanks for the review!

If you need any help with a rebase of #2470 let me know.

quaquel:
it was not very difficult, but thanks for the offer!

PR #2474, opened 2024-11-09
Drop support for Python 3.10, require Python >= 3.11

EwoutH:
Starting development for Mesa 3.1 and following [SPEC 0](https://scientific-python.org/specs/spec-0000/) we can drop Python 3.10 support and require Python 3.11 or higher for future development. This allows adopting more modern Python features and simplifies the testing and CI configurations.

See the Python 3.11 release notes: https://docs.python.org/3/whatsnew/3.11.html

The 3.0.x release series will keep supporting Python 3.10.

image

> ### Summary – Release highlights
> Python 3.11 is between 10-60% faster than Python 3.10. On average, we measured a 1.25x speedup on the standard benchmark suite. See [Faster CPython](https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-faster-cpython) for details.
> 
> New syntax features:
> - [PEP 654: Exception Groups and except*](https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-pep654)
> 
> New built-in features:
> - [PEP 678: Exceptions can be enriched with notes](https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-pep678)
> 
> New standard library modules:
> - [PEP 680](https://peps.python.org/pep-0680/): [tomllib](https://docs.python.org/3/library/tomllib.html#module-tomllib) — Support for parsing [TOML](https://toml.io/) in the Standard Library
> 
> Interpreter improvements:
> - [PEP 657: Fine-grained error locations in tracebacks](https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-pep657)
> - New [-P](https://docs.python.org/3/using/cmdline.html#cmdoption-P) command line option and [PYTHONSAFEPATH](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONSAFEPATH) environment variable to [disable automatically prepending potentially unsafe paths](https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-pythonsafepath) to [sys.path](https://docs.python.org/3/library/sys.html#sys.path)
> 
> New typing features:
> - [PEP 646: Variadic generics](https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-pep646)
> - [PEP 655: Marking individual TypedDict items as required or not-required](https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-pep655)
> - [PEP 673: Self type](https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-pep673)
> - [PEP 675: Arbitrary literal string type](https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-pep675)
> - [PEP 681: Data class transforms](https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-pep681)
> 
> Important deprecations, removals and restrictions:
> 
> - [PEP 594](https://peps.python.org/pep-0594/): [Many legacy standard library modules have been deprecated](https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-pep594) and will be removed in Python 3.13
> - [PEP 624](https://peps.python.org/pep-0624/): [Py_UNICODE encoder APIs have been removed](https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-pep624)
> - [PEP 670](https://peps.python.org/pep-0670/): [Macros converted to static inline functions](https://docs.python.org/3/whatsnew/3.11.html#whatsnew311-pep670)


EwoutH:
@quaquel two things you might find cool we can start using when we require Python 3.11:

> 1. **Exception Notes (PEP 678)**
> We could use `add_note()` to provide additional context for errors in Mesa. This would be particularly useful for:
> - Adding model state information when errors occur
> - Providing hints about common configuration mistakes
> - Adding contextual information about agent states
> 
> For example:
> 
> ```python
> try:
>     # Some model operation
> except ValueError as e:
>     e.add_note(f"Current model step: {self.steps}")
>     e.add_note(f"Number of active agents: {len(self.agents)}")
>     raise
> ```
> 
> 2. **Self Type (PEP 673)**
> We can use the new `Self` type annotation to better annotate methods that return instances of their own class, particularly useful in Agent and Model classes:
> 
> ```python
> from typing import Self
> 
> class Agent:
>     def clone(self) -> Self:
>         return copy.deepcopy(self)
> ```

Those notes really seem handy!

quaquel (approved):

EwoutH:
Not that this is not a breaking change by the way, because the `requires-python = ">=3.11"` prevents anyone from installing Mesa 3.1+ with Python 3.10 or lower. They still can install Mesa 3.0.x, which is fully compatible.

PR #2478, opened 2024-11-09
Bugfixes to DEVS

quaquel:
This isolates the changes to DEVS from #2470 into their own stand-alone PR.

While working on #2479, I discovered a few subtle bugs in `experimental.devs` and the `wolf-sheep` benchmark model
1. setup cleared the event list, but a model (e.g., wolf sheep) might have already scheduled events. Simulators now raises exceptions when (a) setup is called while either time != starttime or when events have already been scheduled; (b) any of the run methods will raise an exception if setup has not yet been called. 
2. A knock-on consequence of point 1 is that `simulator.setup` should now be called inside the `model.__init__` if the `__init__` triggers any event scheduling. For example, for wolf-sheep this is the case because all grass patches that are not fully grown will schedule their moment of being fully regrown. 
3. with #2223 and the associated discussion, there was now a bug in `ABMSimulator`, which scheduled the first step on $t=0$, while this should be $t=1$.


EwoutH:
Thanks, really appreciated. Generally, it looks good. Let's wait for the benchmarks to confirm performance.

Unrelated, but to confirm, the best practice is currently to define the `simulator.setup(model)` insode the init of the Model class, so `self.simulator.setup(self)`?

EwoutH:
Unfortunately, the performance degradation looks reproducible and significant.

Any idea what causing it?

quaquel:
> Unrelated, but to confirm, the best practice is currently to define the simulator.setup(model) insode the init of the Model class, so self.simulator.setup(self)? 

Yes. If setup is called after events have been scheduled, you get an exception. So if `__init__` results in events being scheduled, you have to do it in `model.__init__`. I actually think I like this over the old situation.

quaquel:
> Unfortunately, the performance degradation looks reproducible and significant.
>
> Any idea what causing it?

yes: inccorrect behavior before. What happend was the following

1. grass schedules regrowth events if not fully grown
2. simulator.setup is called, canceling all events
3. Only full grown grass is available to eat and to reschedule itself
4. sheep die very quickly (first 10 turns or so)
5. wolfs die shortly after

Once I had the animation in #2470, the error was obvious.

EwoutH (approved):
Good to know the root cause!

Also shows how valuable visual debugging can be.

quaquel:
> Also shows how valuable visual debugging can be.

absolutely

PR #2485, opened 2024-11-11
examples: Add required components keyword

EwoutH:
SolaraViz now requires `components` to be a keyword argument, this adds it to the two example that didn't have that.

It also shows we really need to start testing our examples in CI.


quaquel (approved):

Corvince:
I think a better solution would be to just make it non-required keyword anymore!? I think its more natural to have solaraviz(model, components). These are always required (without components it doesnt make a lot of sense)

EwoutH:
I'm good with that, I mainly just want working examples 😅

PR #2488, opened 2024-11-11
allow components as a positional argument again

Corvince:
Followup on #2485 

I missed this in the code review. I think "components" should also be allowed as a positional argument. Thats at least how I usually use it when playing around. My mental model is that "model" and "components" are required arguments, which should always be provided and the other arguments can be provided, but in no specific order. The counter argument is that components also have a default value ("default"), so my mental model is not fully correct since its not strictly required. @quaquel  whats your stance on this?


EwoutH (approved):
Thanks, I'm good with it, let's see what Jan thinks.

PR #2489, opened 2024-11-11
examples/gol: Add initial fraction alive, add sliders to viz

EwoutH:
Improve the Game of Life examples:
- Add a initial_fraction_alive keyword argument for the how many cells are alive at the start
- Add sliders for width, height and the initial fraction alive.

image

### DIFF
--- a/mesa/examples/basic/conways_game_of_life/app.py
+++ b/mesa/examples/basic/conways_game_of_life/app.py
@@ -20,12 +20,34 @@ def post_process(ax):
 
 
 model_params = {
-    "width": 50,
-    "height": 50,
+    "width": {
+        "type": "SliderInt",
+        "value": 50,
+        "label": "Width",
+        "min": 5,
+        "max": 60,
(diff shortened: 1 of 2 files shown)


quaquel:
I would also add the seed (here and it all example models)

```python
    "seed": {
        "type": "InputText",
        "value": 42,
        "label": "Random Seed",
    },

```

EwoutH:
Agree, but let's do that in a separate model for all models at once.

Maybe we can even add it pragmatically in SolaraViz for every visualisation. If `"seed"` is not specified in `model_params`, add your dict entry automatically at the end of the dict.

quaquel:
> Maybe we can even add it pragmatically in SolaraViz for every visualisation. If "seed" is not specified in model_params, add your dict entry automatically at the end of the dict.

I looked into that when we decided to collapse the seed into the model arguments, and it's not easy to do allways. For now, let's just add it to all examples and fine to do it via a separate PR.

PR #2491, opened 2024-11-11
test app init of examples

Corvince:
First stab at testing the exmamples app. 

This already tests if there are any errors in app.py itself and if solara starts without any exceptions by calling app.page. We still don't know if anything is displayed and if the buttons etc. do anything, but initial errors are caught.

### DIFF
--- a/mesa/examples/advanced/wolf_sheep/app.py
+++ b/mesa/examples/advanced/wolf_sheep/app.py
@@ -83,7 +83,7 @@ lineplot_component = make_plot_component(
 )
 
 simulator = ABMSimulator()
-model = WolfSheep(simulator, grass=True)
+model = WolfSheep(simulator=simulator, grass=True)
 
 page = SolaraViz(
     model,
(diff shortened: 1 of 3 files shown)


Corvince:
@quaquel  this seems to already catch an error in WolfSheep related to the ABMSimulator?

EwoutH:
I did so way more complex things to try to accomplish this.

Thanks, we really need it.

EwoutH (approved):

quaquel:
> @quaquel this seems to already catch an error in WolfSheep related to the ABMSimulator?

in app.py, change the line to `model = WolfSheep(simulator=simulator, grass=True)`. I don't understand how this has slipped through. My hunch is the split of #2470 into two separate PRs and the resulting rebase mess. I though I tested everything afterwards, but I most somehow have overlookted this.

Corvince:
> I did so way more complex things to try to accomplish this.

Same! Luckily, this is a simple solution



> in app.py, change the line to `model = WolfSheep(simulator=simulator, grass=True)`

I thought I am crazy, because I wanted to do it on GitHub directly and it was already updated. But it was just you being faster. Thanks!

PR #2496, opened 2024-11-11
Add seed control to all examples

quaquel:
This makes sure all example models allow the user to control the seed from the UI.


EwoutH:
Like I suggested, I would like to explore if we can add this programmatically to SolaraViz itself. That way, user models also have it built-in without additional boilerplate.

quaquel:
> Like I suggested, I would like to explore if we can add this programmatically to SolaraViz itself. That way, user models also have it built-in without additional boilerplate.

Don't let the perfect be the enemy of the good. I looked at it and its not easy to do in an elegant way. I prefer to have this in while we have an open issue to explore it further.

EwoutH:
Let me take a quick look tomorrow, otherwise we’ll go this way.

quaquel:
> Let me take a quick look tomorrow, otherwise we’ll go this way.

Feel free to take a look, but let me explain why it is not easy. First, some models might not need the seed control. So now we need the add an option to the API to control whether the seed is included or not. Second, some users might want to use something different from the text box we use here. I just did this because we used that before and have not thought about it extensively. If we want to offer control over this to the user, we have to then offer this flexiblity in the API. One option would be to check what is in `model_params` passed to SoloraViz. If `seed` is not there, we add it in. Otherwise, we use what is there. But how does this interact with option 1? And, what if seed is set to a constant in `model_params`? Should we overwrite this? How do we document all these different code paths and make sure that the behavior of SolaraViz is intuitive to a user? 

We can avoid all this mess by just explicitly asking users to add seed to `model_params` if they want to control it from the GUI.

If, at some future point, we add support for e.g., [Param](https://param.holoviz.org), we might revisit all these questions and add meaningful defaults for for `Parameter`.

Corvince:
In practice I don't think it is *that* complicated. 

1. Check if "seed" is in `model_params`. If it is, just use that with the existing logic (static value => no control, UserParam => UserParam control). Done
2. If "seed" is not in model_params, we need to check if the Model accepts "seed" at all. We already have the basics for that in the `_check_params` function. If it accepts "seed", add the default control. Done

The larger question I have regarding this and also this PR is how to "deactivate" the seed. With this PR (or default controls) I would expect "reset" to always yield the same starting conditions (since we fixed the seed). But for exploring models often times you want to see how it plays out with different starting conditions. Then you want "reset" to mean: Give me a new (random) model. Right now we would need to manually change the seed number. So it would good to also be able to set it to `None` and probably also have this as default - if you want to replicate the same starting conditions you enable it.

Corvince:
I don't think there is, but an easy solution would be to combine a checkbox with an InputFloat. If the checkbbox ("random") is checked, the value is set to None, otherwise to the value of the InputFloat.

quaquel:
I suggest we approve this and open an issue about creating a dedicated widget for seed control unless someone has time to pick that up shortly.

Corvince:
Maybe I'll give it a shot tonight, but this shouldn't hinder this PR anyway

PR #2500, opened 2024-11-13
add a `_mesa_weakref` attribute to Agent

quaquel:
One of the performance bottlenecks in mesa is the use of weakrefs in AgentSet. Weakrefs are great to reduce the potential for memory leaks, but the constant creating of weakrefs, and evaluating whether the weakreffed object is still alive creates quite some overhead. So, what if we can avoid the overhead of continiously creating new weakrefs to agents?

In this test, I added a new attribute, `_mesa_weakref`, to `Agent.` I also updated `AgentSet` to use this attribute. There are still some fixme's, and I doubt all tests pass, but the benchmark models run locally. This is a test and might be a stupid idea born out of jetlag, but I am curious to see the performance impacts.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -69,7 +69,6 @@ class Agent:
         self.pos: Position | None = None
         self.model.register_agent(self)
 
-
     def remove(self) -> None:
         """Remove and delete the agent from the model.
 
@@ -146,9 +145,10 @@ class AgentSet(MutableSet, Sequence):
 
     def __iter__(self) -> Iterator[Agent]:
         """Provide an iterator over the agents in the AgentSet."""
-
         # fixme should this not also be used to cleanout the weakrefs that resolve to None?
-        return (agent for entry in self._agents.keys() if (agent := entry()) is not None)
+        return (
+            agent for entry in self._agents.keys() if (agent := entry()) is not None
+        )
 
     def __contains__(self, agent: Agent) -> bool:
         """Check if an agent is in the AgentSet. Can be used like `agent in agentset`."""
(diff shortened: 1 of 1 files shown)


quaquel:
These benchmark results align with what I see locally, so that is good news. 
1. the reduction in startup time is likely largely due to the fact that on main, we create 2 weakrefs for each agent (one for `model._all_agents` and one for `model._agents_by_type`). In this PR, we only create 1 (inside `Agent`). Attribute lookup is simply faster than creating new objects
2. The runtime differences for WolfSheep are most likely explained by the same mechanism. The main method on `AgentSet` that benefits in this PR is `__contains__` because we don't have the resolve the underlying agent object to check for identity but can do this directly for the weakref.
3. the lack of runtime benefits in the other models is still a bit surprising to me, but it shows that we have optimized e.g., `shuffle_do` already a lot.

EwoutH:
Great work! Some significant gains on WolfSheep, without regressions based on the global benchmarks.

I think the WolfSheep model is the only model that adds and removed agents during run time. So maybe that explains why we only see benefits there?

quaquel:
> I think the WolfSheep model is the only model that adds and removed agents during run time. So maybe that explains why we only see benefits there?


Yes, that is my hunch as well. I hope to do some more tests and profiling on this in the coming days and make sure all existing tests pass again before even considering merging this, but its a small but nice boost allready.

quaquel:
I have done a bit more work on this. It is possible to get this working, but the complexity of AgentSet increases a lot. I am therefore closing this. 

I hope to get back to this basic idea, but with a slight twist at some point in the near future. Most of the increase in code complexity is due to switching from a `WeakrefDict` to a normal dict. However, `WeakrefDict` has some interesting features that we need. Most notably ensuring that dead weakrefs are automagicallly removed from the inner dict. By subclassing dict, we might contain all the new complexity in a single datastructure, while still keeping the AgentSet itself clean and readable.

PR #2502, opened 2024-11-13
cell_space: Allow CellCollection to be empty

EwoutH:
Sometimes a CellCollection needs to be empty, especially after filtering with `.select()`. However, the code would raise a StopIteration error trying to get `capacity` from the first cell. This PR handles empty collections gracefully.

### Summary
Fix StopIteration error when creating empty CellCollections by allowing `capacity` to be None when there are no cells.

### Bug / Issue
Creating a CellCollection with no cells (after filtering) raises StopIteration:
```python
# This raises StopIteration if no cells meet the condition
cells = neighbors.select(lambda cell: not any(isinstance(obj, Wolf) for obj in cell.agents))
```

### Implementation
Check for empty collection before accessing first cell's capacity:
```python
self._capacity: int | None = (
    next(iter(self._cells.keys())).capacity if self._cells else None
)
```

### Testing
- Verify empty CellCollection creation
- Test select() with no matching cells
- Check behavior of empty collection methods

### DIFF
--- a/mesa/experimental/cell_space/cell_collection.py
+++ b/mesa/experimental/cell_space/cell_collection.py
@@ -48,8 +48,10 @@ class CellCollection(Generic[T]):
         else:
             self._cells = {cell: cell.agents for cell in cells}
 
-        #
-        self._capacity: int = next(iter(self._cells.keys())).capacity
+        # Get capacity from first cell if collection is not empty
+        self._capacity: int | None = (
+            next(iter(self._cells.keys())).capacity if self._cells else None
+        )
 
         if random is None:
             warnings.warn(
--- a/tests/test_cell_space.py
+++ b/tests/test_cell_space.py
@@ -583,6 +583,40 @@ def test_cell_collection():
     assert len(cells) == len(collection)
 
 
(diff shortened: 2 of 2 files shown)


mesa/experimental/cell_space/cell_collection.py
```diff
@@ -48,8 +48,10 @@ def __init__(
         else:
             self._cells = {cell: cell.agents for cell in cells}
 
-        #
-        self._capacity: int = next(iter(self._cells.keys())).capacity
+        # Get capacity from first cell if collection is not empty
+        self._capacity: int | None = (
+            next(iter(self._cells.keys())).capacity if self._cells else None
```
quaquel, line 53: I am fine with doing it this way, but the pythonic way is to do a try catch instead (its beter to ask forgiveness then permission).

quaquel (approved):

mesa/experimental/cell_space/cell_collection.py
EwoutH: I think a try-catch is overkill here, since there are two clear, predictable scenarios.

PR #2503, opened 2024-11-13
examples/wolf_sheep: Don't allow dumb moves

EwoutH:
Agents now move completely random in WolfSheep. That makes no sense whatsoever, so this PR makes them move a little bit less dumb.

- Wolf moves to random cell with sheep on them, if available (otherwise completely random)
- Sheep move to a random cell without a wolf (if available), and preferably with grass.

This enables sheep to actually "search" for grass, wolfs to search for sheep and sheep to not move to a cell with wolves.

More importantly, it shows of some nice selection mechanics with the cell space.

I expect the WolfSheep model to be slower, but curious by how much.

Depends on https://github.com/projectmesa/mesa/pull/2502.

**Edit:** Dynamics are quite interesting, I had this epic run where one single sheep survived endlessly until all the wolves where finally dead.

image

### DIFF
--- a/mesa/examples/advanced/wolf_sheep/agents.py
+++ b/mesa/examples/advanced/wolf_sheep/agents.py
@@ -37,7 +37,8 @@ class Animal(CellAgent):
     def step(self):
         """Execute one step of the animal's behavior."""
         # Move to random neighboring cell
-        self.cell = self.cell.neighborhood.select_random_cell()
+        self.move()
+
         self.energy -= 1
 
         # Try to feed
@@ -62,6 +63,27 @@ class Sheep(Animal):
             self.energy += self.energy_from_food
             grass_patch.fully_grown = False
 
+    def move(self):
+        """Move towards a cell where there isn't a wolf, and preferably with grown grass."""
+        cells_without_wolves = self.cell.neighborhood.select(
+            lambda cell: not any(isinstance(obj, Wolf) for obj in cell.agents)
+        )
+        # If all surrounding cells have wolves, stay put
+        if len(cells_without_wolves) == 0:
+            return
+
+        # Among safe cells, prefer those with grown grass
+        cells_with_grass = cells_without_wolves.select(
+            lambda cell: any(
+                isinstance(obj, GrassPatch) and obj.fully_grown for obj in cell.agents
+            )
+        )
+        # Move to a cell with grass if available, otherwise move to any safe cell
+        target_cells = (
+            cells_with_grass if len(cells_with_grass) > 0 else cells_without_wolves
+        )
+        self.cell = target_cells.select_random_cell()
+
 
 class Wolf(Animal):
     """A wolf that walks around, reproduces (asexually) and eats sheep."""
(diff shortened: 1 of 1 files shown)


EwoutH:
Double the runtime is quite an increase. But that's good, that means we now have a meaningful step which can be profiled and optimized.

Corvince:
Oof, yes that's quite an increase and invites to further investigate! And this seems to be a way more interesting model dynamic

quaquel:
I have an overarching question. This model is used in the [ABM Framework comparison](https://github.com/JuliaDynamics/ABMFrameworksComparison). We, therefore, use it in our own internal benchmarks. Clearly, with the examples back in the main repo, we should not duplicate code between benchmarks and examples. This change, however, breaks the link with the ABM Framework comparison, so do we want that, and what does this imply for our own benchmark models?

> Oof, yes that's quite an increase and invites to further investigate! And this seems to be a way more interesting model dynamic

It highlights a point we both have made before, no matter how fast we would make the core of mesa (e.g., via rust or cython), often the real bottleneck is user code.

As an aside, and more for my own understanding, how does the benchmarking now work with examples back in the main repo?

EwoutH:
> This change, however, breaks the link with the ABM Framework comparison, so do we want that, and what does this imply for our own benchmark models?

I think breaking this link is fine. The main use case of the examples is showcasing Mesa features. I think this change showcases what's possible with the Mesa Cell Space well.

> As an aside, and more for my own understanding, how does the benchmarking now work with examples back in the main repo?

Their goal is to see how fast Mesa is, and to detect performance regressions or improvements when we do feature development. Most operation are still Mesa internals (the spaces), so we're still benchmarking Mesa, and not arbitrary Python code.

mesa/examples/advanced/wolf_sheep/agents.py
```diff
@@ -62,6 +63,27 @@ def feed(self):
             self.energy += self.energy_from_food
             grass_patch.fully_grown = False
 
+    def move(self):
+        """Move towards a cell where there isn't a wolf, and preferably with grown grass."""
+        cells_without_wolves = self.cell.neighborhood.select(
+            lambda cell: not any(isinstance(obj, Wolf) for obj in cell.agents)
```
EwoutH, line 69: @quaquel I think this eats some performance. Ideas on how to do it faster?

quaquel: Tricky. 

1. you could make it a normal for loop and jump out the moment you see a wolf. This will shave of a little bit of performance
2. Use a property layer of booleans and modify the wolf to set this property to True/False in their move method. Now you can express the entire movement in a vectorized manner.

EwoutH:
I'm merging this, because:

1. It's an obvious improvement in behavior
2. It demonstrates Cell Space capability not shown earlier in other examples
3. I tested it extensively
4. Performance can be optimized later
     (opened https://github.com/projectmesa/mesa/issues/2565)
5. Keeping it open increases potential conflicts with other changes to this example model.

mesa/examples/advanced/wolf_sheep/agents.py
EwoutH: Agree it's tricky, wrote it up for now in #2565.

tpike3:
@EwoutH next time can you ping one of us just to make sure we get the approve on the PR. Thank you

EwoutH:
You're right, will do

EwoutH:
This would be a implicit change of policy though. Previously, in Mesa-examples, we didn't require explicit reviews, while we could request them.

I'm not against requiring reviews for example updates per se, but I do think this warrants a bit of a bigger discussion.

Personally I might have the stance that only core Mesa would require explicit reviews. Other stuff is non-critical, and we can trust each others judgement to request a review when they are nessesary.

We might have to accept the occasional revert. I don't think that is a bad thing, it means we're finding our path and learning things.

PR #2511, opened 2024-11-17
Refactored docs for Introductory Tutorial

Spartan-71:
## Summary

Fixed minor logical error in `give_money` function of `MoneyAgent` class.
Fixed code inconsistency.

## Bug / Issue

Earlier `MoneyAgent` was able to transfer the money to itself.

## Implementation

Now MoneyAgent `self` has been removed from the `cellmates` list and the condition to transfer the money is modified to `if len(cellmates) > 0`

## Additional Notes

### DIFF
--- a/docs/tutorials/intro_tutorial.ipynb
+++ b/docs/tutorials/intro_tutorial.ipynb
@@ -682,17 +682,6 @@
     "            self.wealth -= 1\n",
     "```\n",
     "\n",
-    "And with those two methods, the agent's ``agent_act`` method becomes:\n",
-    "\n",
-    "```python\n",
-    "class MoneyAgent(mesa.Agent):\n",
-    "    # ...\n",
-    "    def agent_act(self):\n",
-    "        self.move()\n",
-    "        if self.wealth > 0:\n",
-    "            self.give_money()\n",
-    "```\n",
-    "\n",
     "Now, putting that all together should look like this:"
    ]
   },
@@ -725,11 +714,6 @@
     "            other_agent.wealth += 1\n",
     "            self.wealth -= 1\n",
     "\n",
-    "    def agent_act(self):\n",
-    "        self.move()\n",
-    "        if self.wealth > 0:\n",
-    "            self.give_money()\n",
-    "\n",
     "\n",
     "class MoneyModel(mesa.Model):\n",
     "    \"\"\"A model with some number of agents.\"\"\"\n",
@@ -748,7 +732,8 @@
     "            self.grid.place_agent(a, (x, y))\n",
     "\n",
     "    def step(self):\n",
-    "        self.agents.shuffle_do(\"agent_act\")"
+    "        self.agents.shuffle_do(\"move\")\n",
+    "        self.agents.do(\"give_money\")"
    ]
   },
   {
@@ -867,13 +852,7 @@
     "            other = self.random.choice(cellmates)\n",
     "            other.wealth += 1\n",
     "            self.wealth -= 1\n",
-    "\n",
-    "    # There are several ways in which one can combine functions to execute the model\n",
-    "    def agent_act(self):\n",
-    "        self.move()\n",
-    "        if self.wealth > 0:\n",
-    "            self.give_money()\n",
-    "\n",
+    "            \n",
     "\n",
     "class MoneyModel(mesa.Model):\n",
     "    \"\"\"A model with some number of agents.\"\"\"\n",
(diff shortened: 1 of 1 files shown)


EwoutH:
Thanks for your PR!

> Ensure agent is not giving money to itself

Great catch! I remember fixing this bug somewhere sometime, interesting how it keeps appearing.

@quaquel can we add a method/keyword to exclude yourself from the “get cell contents” and other similar functions?

I don’t agree with the step being merged into a single `agent_act`. It changes the activation order. Previously, all agents first moved, and then all gave money. Now each agent moves, gives money, and then the next agent etc. Can you motivate why you made that change?

quaquel:
> @quaquel can we add a method/keyword to exclude yourself from the “get cell contents” and other similar functions?

This is still using old style spaces, not new style spaces, so yes you can do it. However, we want to deprecate these for 3.1, so I don't see the point. In new-style spaces, it is a lot harder to fix this, since `agents` is an attribute not a method on both `Cell` and `CellCollection`.

Spartan-71:
> I don’t agree with the step being merged into a single `agent_act`. It changes the activation order. Previously, all agents first moved, and then all gave money. Now each agent moves, gives money, and then the next agent etc. Can you motivate why you made that change?

In the "**Adding Space**" -> "**Moving Mesa**" section, there is already a function named `step` in the `MoneyAgent` class. However, in the subsequent code snippet, this function seems to disappear. Instead, the `move` and `give_money` functions are directly called within the `step` function of the `MoneyModel` class.

image

image

While in "**Collecting Data**" Section the `agent_act` function is mentioned !

image

Ref: https://mesa.readthedocs.io/stable/tutorials/intro_tutorial.html

EwoutH:
@tpike3 you last looked into the tutorial, do you know what the intended behavior was?

Spartan-71:
Hey @EwoutH and @tpike3! If there’s anything you’d like to change, just let me know. 😊

EwoutH:
> In the "**Adding Space**" -> "**Moving Mesa**" section, there is already a function named `step` in the `MoneyAgent` class. However, in the subsequent code snippet, this function seems to disappear. Instead, the `move` and `give_money` functions are directly called within the `step` function of the `MoneyModel` class.

I think the old `step` method can be removed, in favor of the new move and give_money implementation.

docs/tutorials/intro_tutorial.ipynb
```diff
@@ -509,8 +514,10 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "model = MoneyModel(10) # Tels the model to create 10 agents\n",
-    "for _ in range(30): #Runs the model for 10 steps; an underscore is common convention for a variable that is not used\n",
+    "model = MoneyModel(10)  # Tels the model to create 10 agents\n",
```
tpike3, line 517: Fix spelling of Tels

docs/tutorials/intro_tutorial.ipynb
```diff
@@ -706,22 +717,28 @@
...
-    "                other_agent.wealth += 1\n",
-    "                self.wealth -= 1\n",
+    "        cellmates = self.model.grid.get_cell_list_contents([self.pos])\n",
+    "        # Ensure agent is not giving money to itself\n",
+    "        cellmates.pop(cellmates.index(self))\n",
+    "        if len(cellmates) > 0:\n",
+    "            other_agent = self.random.choice(cellmates)\n",
+    "            other_agent.wealth += 1\n",
+    "            self.wealth -= 1\n",
+    "\n",
```
tpike3, line 727: Per the comment, I think this could either be left as "step" common convention with an explicit comment on options or

make it consistent where it would be 
`
def step(self):
        self.agents.shuffle_do("move")
        self.agents.do("give_money")
`

tpike3:
@Spartan-71 and @EwoutH Sorry for the delay, if I remember correctly I believe the intent was to show users that there is multiple ways to code it. So I think we have two options

1 - Explicitly call out why they are different in the write up and comments
2 - Make it consistent where the agent function care called form the model step

I dont think we need to belabor this, just need to pick one

Thank you @Spartan-71 for doing this!

Spartan-71:
@tpike3 I chose to go ahead with the 2nd option.

tpike3 (approved):

EwoutH:
Thanks for your effort, and congratulations on getting the PR in!

Is this worth backporting to the 3.0.x maintenance branch?

EwoutH:
@tpike3 minor thing, but could you squash such PRs when merging them?

IMG_0162

PR #2512, opened 2024-11-18
Add full support for property layers to cell spaces

quaquel:
This is a somewhat opinionated take on how to offer full support for property layers in new-style cell spaces. It draws on ideas from #2431 and #2440. In short, I have taken the property layer code from `mesa.space`, and created a new `property_layers` module in `mesa.experimental.cell_space`. This ensures that the existing code in `mesa.space` remains untouched, and I can tune the behavior to align with the design ideas of cell_space.

All relevant code for adding property layer behavior to grids is contained in a new `HasPropertyLayer` class. This is primarily for easy development and allowing me to understand better what is going on. 

I have moved property layer functionality from the generic `DiscreteSpace` to `Grid` as discussed in #2431. Because Grid can be n-dimensional, I modified PropertyLayer and all relevant code to work for n-dimensional grids. 

I enforce that the dimensionality of the property layer and the dimensionality of the grid are the same. This was left implicit before. Now, it raises errors if you try to have a layer with a different coordinate system.

Cells have attribute-like access to their properties. So `cell.grass` or `cell.elevation`. Likewise, any modification done at the cell level can just use assignment (i.e., `=`). This is all implemented via a `PropertyDescriptor`, which offers a view into the underlying numpy array defined in the property layer.

What currently does not work yet is:
1. ~~Masks (this is a trivial problem, just a bit more tricky because it needs to work for n-dimensional grids).~~ 
2. ~~The `empty` property layer is not yet working. I have to take a closer look at this. It involves two separate problems: (1) Where should the layer be added? It cannot be done in the `__init__` of `HasPropertyLayers` because when this code is executed, the `__init__` of `Grid` has not been completed. So e.g., `dimensions` is not yet defined. (2) It requires updating `Cell` to set a boolean to true or false. Since I am relying on adding descriptors dynamically, but only for grids and not for e.g., `Network`, I have to figure out how to make this all work correctly independently of the exact subclass of `DiscreteSpace` to which `Cell` belongs. Ideally, the Cell class should be unaware of the presence or absence of an `empty` property layer~~
3. ~~How the dynamic adding of descriptors interacts with deepcopy and pickle operations. Again, I have no idea here, but it should be fixable.~~


quaquel:
> The empty property layer is not yet working.

This is now working again. It required some Python tricks involving dynamically creating subclasses.

Corvince:
Sorry for not having commented on this earlier, but this is again fantastic work! I haven't looked into the implementation yet, but how it interacts is very nice and very close to what I imagined how Cells and PropertyLayers should interact.

The only thing I find worthy of a discussion is how to access the properties. You now have `cell.elevation`. I had thought properties would be access via `cell.properties.elevation`. In my original thought this would separate a bit the attributes of the `Cell` class itself and the specific properties a cell or the underlying property layer *has*. But I am now unsure which one is better. Since I don't have a strong preference, what are some other opinions on this? Also @quaquel  would this simplify or complicate the implementation?

Corvince:
Thanks for the extensive reply @quaquel . I see the difficulties. Would it work if we store it like `cell._elevation` and then just add a reference to that private variable to the properties dict? So `cell.properties["elevation"] = cell._elevation`? No, that probably won't work right? By setting it to a new value we are just updating the dict I think. Maybe there is a way to make this work? Probably not through an ordinary dictionary. 

I guess it also depends on if we find additional benefits of a dedicated "Property" class. If there is anything else that might enhance this class its worth a shot, but if its really only for adding the descriptor I don't think it makes a lot of sense. One advantage though might be the ability to easily detect a property. If we go through the list of cell attributes we can detect its a property and for example use that for visualization stuff. But we probably also can detect if it uses the Property descriptor?

quaquel:
> Probably not through an ordinary dictionary.

Indeed, not through an ordinary dictionary. It might be possible to let `Properties` subclass a dict (instead of using `__getatttribute__`) but I personally prefer attribute like access over field access as in dicts. 

> One advantage though might be the ability to easily detect a property. 

This is quite easy to do also with descriptors. See also the mesa signals PR, where I have a simple generator that yields all Observables on a class. It might be even simpler in this case, since I am dynamically modifying the `cell_klass`, we could simply at a `_mesa_properties: Set[str]` field and add the name of each property to `_mesa_properties`. Now, e.g., visualization can just read this field, and it knows all defined properties at the cell level.

tpike3 (approved):
@quaquel This is awesome! I know you have some more work work to so but I hit the approve as it is experimental and I do not want to hold ups its integration. Hopefully I get a chance to play around with it next week.

quaquel:
> Hopefully I get a chance to play around with it next week.

I'll probably merge this tomorrow. I want to add a few more tests to finish this off.

After that, I plan to update the sugarscape example to use this functionality.

What are you planning to use this for?

mesa/experimental/cell_space/property_layer.py
```diff
@@ -0,0 +1,392 @@
...
+class HasPropertyLayers:
+    """Mixin-like class to add property layer functionality to Grids."""
+
+    # fixme is there a way to indicate that a mixin only works with specific classes?
+    def __init__(self, *args, **kwargs):
+        """Initialize a HasPropertyLayers instance."""
+        super().__init__(*args, **kwargs)
+        self._mesa_property_layers = {}
+
+    def create_property_layer(
```
EwoutH, line 189: Can we let `create_property_layer` return the PropertyLayer also? That would allow users to directly assign it and manipulate, if they want to do so.

It doesn't have to be a keyword, because `self.grid.create_property_layer()` will also still work if the return is not used.

quaquel:
> 3. How the dynamic adding of descriptors interacts with deepcopy and pickle operations. Again, I have no idea here, but it should be fixable.

As expected, `pickle` and `deepcopy` don't like dynamical subclassing and adding descriptors dynamically. I am figuring out how to resolve this, which is most definitely possible. It requires, however, learning more about the inner workings of `__reduce__`, `__setstate__`, and `__getstate__`, which are the magic methods used by both `pickle` and `deepcopy`. I have been making steady progress on this, but I need a few more hours to fully resolve these issues.

EwoutH:
I asked an LLM to create an overview of all API features available to users. I will discuss each, but also please let me know if any of these are actually incorrect.

### Creating Property Layers
```python
from mesa import Model
from mesa.experimental.cell_space import OrthogonalMooreGrid, PropertyLayer

class CompleteExample(Model):
    def __init__(self):
        super().__init__()
        self.grid = OrthogonalMooreGrid((10, 10))
        
        # 1. Direct creation on grid
        self.grid.create_property_layer(
            name="elevation",
            default_value=0.0,
            dtype=float
        )
        
        # 2. Create first, then add
        layer = PropertyLayer(
            name="temperature",
            dimensions=(10, 10),
            default_value=20.0,
            dtype=float
        )
        self.grid.add_property_layer(layer)
        
        # 3. Create from existing data
        import numpy as np
        data = np.random.random((10, 10))
        moisture = PropertyLayer.from_data("moisture", data)
        self.grid.add_property_layer(moisture)
```
I really like this flexibility here. It allows adding PropertyLayers to multiple grids, etc. `from_data` is new right? Also useful!

@wang-boyu @SongshGeo do you see anything that would be useful for for Mesa-geo?

### Cell-Level Operations

```python
# Get cell reference
cell = self.grid._cells[(0, 0)]

# Get property value
height = cell.elevation

# Set property value
cell.elevation = 100

# Increment/decrement
cell.elevation += 10
cell.elevation -= 5

# Multiply/divide
cell.elevation *= 2
cell.elevation /= 4

# Use numpy operations
cell.elevation = np.sqrt(cell.elevation)
```
Okay I really like that you can now do `.property` directly. That is very powerful, without a dict lookup, strings or anything. That's using Python's descriptor protocol right? Impressive, especially if it's fast!

How are naming conflicts handled? For example if I define a PropertyLayer named `capacity`?

### Grid-Level Operations
#### Basic Property Setting

```python
# Set single value for all cells
self.grid.set_property("elevation", 50)

# Set using array
import numpy as np
new_values = np.random.random((10, 10))
self.grid.set_property("elevation", new_values)

# Set conditionally
self.grid.set_property(
    "elevation",
    100,
    condition=lambda x: x < 50  # Only set cells where elevation < 50
)
```
Nice, extends well! Was inputting an array always supported?

#### Property Modification
```python
# Using lambda functions
self.grid.modify_properties(
    "temperature",
    lambda x: x + 10
)

# Using numpy operations
self.grid.modify_properties("temperature", np.sqrt)
self.grid.modify_properties("temperature", np.add, value=5)
self.grid.modify_properties("temperature", np.multiply, value=2)

# With conditions
self.grid.modify_properties(
    "temperature",
    np.add,
    value=10,
    condition=lambda x: x < 30
)
```
Very consistent with PropertyLayer itself, like it!

### Cell Selection
#### Basic Selection

```python
# Select cells based on conditions
high_ground = self.grid.select_cells(
    conditions={"elevation": lambda x: x > 100}
)

# Select only empty cells
empty_high_ground = self.grid.select_cells(
    conditions={"elevation": lambda x: x > 100},
    only_empty=True
)

# Get boolean mask instead of coordinates
mask = self.grid.select_cells(
    conditions={"elevation": lambda x: x > 100},
    return_list=False
)
```

#### Advanced Selection

```python
# Select based on extreme values
highest_points = self.grid.select_cells(
    extreme_values={"elevation": "highest"}
)

lowest_points = self.grid.select_cells(
    extreme_values={"elevation": "lowest"}
)

# Combine conditions and extreme values
wet_peaks = self.grid.select_cells(
    conditions={"moisture": lambda x: x > 0.7},
    extreme_values={"elevation": "highest"}
)

# Using pre-computed masks
neighborhood = self.grid.get_neighborhood_mask((5, 5))
selected = self.grid.select_cells(
    conditions={"temperature": lambda x: x > 25},
    masks=neighborhood
)

# Multiple masks
mask1 = self.grid.get_neighborhood_mask((0, 0))
mask2 = self.grid.get_neighborhood_mask((9, 9))
selected = self.grid.select_cells(masks=[mask1, mask2])
```
Selection stuff also looks great, that `HasPropertyLayers` Mixin really pulls it weight. Smart construct!

### Neighborhood Operations

```python
# Get neighborhood mask
center = (5, 5)
# With center cell
mask = self.grid.get_neighborhood_mask(
    coordinate=center,
    include_center=True,
    radius=1
)

# Without center cell, Larger radius
mask = self.grid.get_neighborhood_mask(
    coordinate=center,
    include_center=False,
    radius=2
)

# Use mask in property operations
self.grid.modify_properties(
    "temperature",
    np.add,
    value=5,
    condition=lambda x: mask
)
```
Powerful. Can we also allow directly inputting a mask (with a mask keyword), instead of using `condition=lambda x: mask`?

### Layer Management

```python
# Remove a property layer
self.grid.remove_property_layer("temperature")

# Access underlying numpy array
elevation_data = self.grid._mesa_property_layers["elevation"].data

# Get dimensions of a layer
dimensions = self.grid._mesa_property_layers["elevation"].dimensions

# Check if layer exists
has_temperature = "temperature" in self.grid._mesa_property_layers
```
Maybe it would be nice to have attributes like the dimensions and data a bit more accessible. On the other hand, in most case you want that you probably are using PropertyLayers directly and not through the grid.

### The "empty" Property Layer

```python
# Every grid automatically has an "empty" property layer
cell = self.grid._cells[(0, 0)]
is_empty = cell.empty  # True if no agents in cell

# When adding/removing agents, empty is automatically updated
agent = MyAgent(self)
self.grid.place_agent(agent, (0, 0))
print(cell.empty)  # False

agent.remove()
print(cell.empty)  # True

# Select all empty cells
empty_cells = self.grid.select_cells(only_empty=True)
```
Very useful. Do we also have direct access to empty cells? Maybe we can add `grid.empty_cells` and `grid.empty_mask` properties.

### PropertyLayer Class Methods

```python
# Create layer from array
data = np.random.random((10, 10))
layer = PropertyLayer.from_data("values", data)

# Set all cells
layer.set_cells(5.0)

# Set cells conditionally
layer.set_cells(10.0, condition=lambda x: x < 5)

# Modify cells
layer.modify_cells(np.add, value=1)
layer.modify_cells(lambda x: x * 2)

# Select cells
high_values = layer.select_cells(lambda x: x > 0.5)

# Aggregate values
total = layer.aggregate(np.sum)
average = layer.aggregate(np.mean)
maximum = layer.aggregate(np.max)
```
Not new, but left it here for completion of we ever want to reference it.

### Type Handling

```python
# Different data types
self.grid.create_property_layer("height", default_value=0, dtype=int)
self.grid.create_property_layer("active", default_value=False, dtype=bool)
self.grid.create_property_layer("name", default_value="empty", dtype=str)
self.grid.create_property_layer("complex", default_value=1+0j, dtype=complex)

# Warning is issued if default_value type doesn't match dtype
self.grid.create_property_layer("number", default_value=1.0, dtype=int)  # Warning
```
Idem dito.

### Additional Features
```python
# Access through grid's property layers dict
layers = self.grid._mesa_property_layers

# Property layer name
name = self.grid._mesa_property_layers["elevation"].name

# Check layer dimensions
dims = self.grid._mesa_property_layers["elevation"].dimensions

# N-dimensional grids supported
grid3d = OrthogonalMooreGrid((10, 10, 10))
grid3d.create_property_layer("density", default_value=0.0)

# Direct data access/modification
layer = self.grid._mesa_property_layers["elevation"]
layer.data = np.zeros_like(layer.data)  # Replace all data
```
Some of this is a bit cumbersome, but I also don't really see frequent use cases currently for most of these. Especially if you can also directly interact with the PropertyLayer without going through the grid or _mesa_property_layers. So fine.

<hr>

This is really excellent work. I don't always agree with your abstractions, and the complexity it sometimes brings, but this is brilliant. I really like the flexibility that `HasPropertyLayers` offers, smart chocie to make that a Mixin. `PropertyDescriptor` giving direct access is also a nice touch.

Feel free to address any comments you like, non are blocking (from my side). I hope you find this kind of review (from a "user" standpoint) useful, otherwise also let me know.

quaquel:
> I asked an LLM to create an overview of all API features available to users. I will discuss each, but also please let me know if any of these are actually incorrect.

Thanks a lot for finding time to do this. I'll try to respond in more detail somewhere later this week.

EwoutH (approved):

quaquel:
found some time on the train to The Hague



**Creating Property Layers**

> I really like this flexibility here. It allows adding PropertyLayers to multiple grids, etc. from_data is new right? Also useful!

Thanks! `from_data` is indeed new, the rest follows your original code in `mesa.space`.


**Cell-Level Operations**

> That's using Python's descriptor protocol right? Impressive, especially if it's fast!

Yes, this relies on the descriptor proptocol, dynamic subclassing, and dynamicly adding descriptors. I don't know of any other way to make this work, but the code complexity is worth the user friendlyness of the API in my view. Descriptors are fast (they sit high in the method resolution order). 

> How are naming conflicts handled? For example if I define a PropertyLayer named `capacity`?

Good question. I'll add a test for this because I don't know. I expect it will raise an AttributeError if you do this.


**Grid-Level Operations**

> Nice, extends well! Was inputting an array always supported?

`set_property` is a new method I added. It just made sense to have it.


> Very consistent with PropertyLayer itself, like it!

I think this largely follows your code from `_PropertyGrid`.

> Selection stuff also looks great, that HasPropertyLayers Mixin really pulls it weight. Smart construct!

Again, most of this selection stuff was in `_PropertyGrid`. I just encapsulated it in `HasPropertyLayers` and made some minor tweaks to align it with the rest of the API of cell-spaces.

> Powerful. Can we also allow directly inputting a mask (with a mask keyword), instead of using `condition=lambda x: mask`?

At the moment, no. I was actuallly thinking about adding this while working on the code, but decided to leave it for a future PR because I tried stay as close as possible to the current code in `mesa.space`. 

> Maybe it would be nice to have attributes like the dimensions and data a bit more accessible. On the other hand, in most case you want that you probably are using PropertyLayers directly and not through the grid.

`dimensions` is not needed since by design this will be the same as `grid.dimenions`. The names etc., might indeed be refined a bit. It is not ideal that at the moment users have to interact with a protected attribute (`_mesa_properties`). I'll think about this a bit more but it should be possible to add some more methods to `HasPropertyLayers` to make this easier and more user friendly. Something like a `get_layer(name:str)-> PropertyLayer` method would already help a lot. The other option is to make `grid.elevation` work which can be done via the `__getattr__` magic.

> Very useful. Do we also have direct access to empty cells? Maybe we can add grid.empty_cells and grid.empty_mask properties.

`DiscreteSpace` already has an `empties` attribute, so we have to think a bit about how this would interact with the `empty` property layer. Again, I decided to leave this for a future PR.

> This is really excellent work. I don't always agree with your abstractions, and the complexity it sometimes brings, but this is brilliant. I really like the flexibility that HasPropertyLayers offers, smart chocie to make that a Mixin. PropertyDescriptor giving direct access is also a nice touch.

Thanks. Curious to hear what abstractions you don't agree with. Most of the complexity that is currently there is unavoidable to make the descriptor stuff work.

EwoutH:
For completion, I think technically there's a fourth option, like Pandas does it to access series:

4. `grid["elevation"]`

However, 1 feels really great once you know it, and there is precedence for this kind of attribute acces in the Python framework: 

```Python
        arr = np.array([[1, 2], [3, 4]])
        arr.shape  # (2, 2)

        df = pd.DataFrame({'a': [1, 2], 'b': [3, 4]})
        df.a  # Series([1, 2])
```
There should be some checks on the variable names, at least they should be valid Python identifiers (similar to what we now enforce in the EMA workbench). Maybe we should also check against some other things.

Best to see is how NumPy and Pandas do it, they probably already figured this out fully. Maybe packages like [nameutils](https://pypi.org/project/nameutils/) could also help out here.

quaquel:
> Best to see is how NumPy and Pandas do it, they probably already figured this out fully. Maybe packages like [nameutils](https://pypi.org/project/nameutils/) could also help out here.

pandas does not do a check, but if it is not a valid identifier, the attribute like access is simply not supported. So you can have column names with a space in them and pandas accepts that. I am inclined to just follow that route and document clearly that for attribute access, you need to ensure that the property layer name needs to be a valid python identifier.

Note that the package you link to does something interesting but quite different (namely handling family names and capitalization).

EwoutH:
@quaquel Thanks for continuing working on this. How far do you expect it from being ready to be merged?

quaquel:
>How far do you expect it from being ready to be merged?

Its basically done, but I like to give it one last look over before merging.

EwoutH:
I updated the PR description based on the final implementation, could you check it?

kiefernforst:
@quaquel Dear Kwakkel,

I'm trying to use the last commit compatible with Python 3.10

Is it true that https://github.com/mesa/mesa/blob/6f71fcdf770621befc704ecc4cf7b1cc10818164/mesa/discrete_space/property_layer.py#L324 is not 3.10-compatible despite being older than e21a5b5282149d5fa38d3c55aa3cf51d87ff1b0e ?

```
    mask[*indices] = True
         ^
SyntaxError: invalid syntax
```

EwoutH:
Something is weird here:
- Our latest version to support Python 3.10 is Mesa 3.0.3.
- Mesa 3.0.3 [doesn't](https://github.com/mesa/mesa/tree/3.0.x-maintenance) have a discrete space with property layer support.

Which exact Mesa and Python versions are you using? And where is the error originating exactly from (both source code and test)?

PR #2518, opened 2024-11-25
Fix: Schelling Model Neighbor Similarity Calculation

### LINKED ISSUE #2515: homophily parameter in basic schelling example
**Describe the bug**

The agents in our basic schelling example compare number of similar agents in their neighborhood, against a `homophily` parameter set by users: https://github.com/projectmesa/mesa/blob/a1da4345ad5b8a9f8853145c049fbccc5cd1334e/mesa/examples/basic/schelling/agents.py#L24-L27

where `homophily` is default to 3: https://github.com/projectmesa/mesa/blob/a1da4345ad5b8a9f8853145c049fbccc5cd1334e/mesa/examples/basic/schelling/app.py#L29

Now consider the following example agent at the center cell:

```
X |   |
---------
X | X |
---------
  |   |
```

It has two surrounding agents with the same type, which is smaller than 3, so the agent at the center is considered as "unhappy" in our model.

**Expected behavior**

[Wikipedia](https://en.wikipedia.org/wiki/Schelling%27s_model_of_segregation) uses fraction of neighbors, instead of absolute number of similar neighbors:

 > Each round consists of agents checking their neighborhood to see if the fraction of neighbors $B$ that matches their group - ignoring empty spaces - is greater than or equal $B_a$.

In the example above, the agent has a fraction of 100% neighbors with the same type, so it is very happy. This leads to different results from our model.

Essentially our model counts empty spaces as valid neighbors, while Wikipedia says otherwise. I think NetLogo's Segregation example uses fraction too, following Wikipedia.

Is our segregation model intentionally designed this way?

Sahil-Chhoker:
## Summary
Fixed #2515 Schelling segregation model to calculate agent happiness using neighbor similarity fraction, aligning with Wikipedia and NetLogo standards.

## Bug / Issue
Current model incorrectly counts empty spaces as neighbors, leading to inaccurate agent happiness determination.

## Implementation
Modified `SchellingAgent.step()` to:
- Filter out empty cells
- Calculate similar neighbor fraction
- Scale homophily threshold

```python
similar_neighbors = [
    neighbor for neighbor in neighbors 
    if hasattr(neighbor, 'type') and neighbor.type == self.type
]
total_neighbors = [
    neighbor for neighbor in neighbors 
    if hasattr(neighbor, 'type')
]

if len(total_neighbors) > 0:
    similarity_fraction = len(similar_neighbors) / len(total_neighbors)
    
    if similarity_fraction < self.model.homophily / 8.0:
        self.model.grid.move_to_empty(self)
    else:
        self.model.happy += 1
```

### DIFF
--- a/mesa/examples/basic/schelling/agents.py
+++ b/mesa/examples/basic/schelling/agents.py
@@ -1,12 +1,9 @@
 from mesa import Agent
 
-
 class SchellingAgent(Agent):
     """Schelling segregation agent."""
-
     def __init__(self, model, agent_type: int) -> None:
         """Create a new Schelling agent.
-
         Args:
             model: The model instance the agent belongs to
             agent_type: Indicator for the agent's type (minority=1, majority=0)
@@ -19,12 +16,23 @@ class SchellingAgent(Agent):
         neighbors = self.model.grid.iter_neighbors(
             self.pos, moore=True, radius=self.model.radius
         )
-
-        # Count similar neighbors
-        similar = sum(neighbor.type == self.type for neighbor in neighbors)
-
-        # If unhappy, move to a random empty cell:
-        if similar < self.model.homophily:
-            self.model.grid.move_to_empty(self)
-        else:
-            self.model.happy += 1
+        
+        # Filter out empty cells
+        similar_neighbors = [
+            neighbor for neighbor in neighbors 
+            if hasattr(neighbor, 'type') and neighbor.type == self.type
+        ]
+        total_neighbors = [
+            neighbor for neighbor in neighbors 
+            if hasattr(neighbor, 'type')
+        ]
+        
+        # Calculate fraction of similar neighbors
+        if len(total_neighbors) > 0:
+            similarity_fraction = len(similar_neighbors) / len(total_neighbors)
+            
+            # If unhappy, move to a random empty cell
+            if similarity_fraction < self.model.homophily / 8.0:
+                self.model.grid.move_to_empty(self)
+            else:
+                self.model.happy += 1
--- a/mesa/examples/basic/schelling/agents.py
+++ b/mesa/examples/basic/schelling/agents.py
@@ -1,7 +1,9 @@
 from mesa import Agent
 
+
 class SchellingAgent(Agent):
     """Schelling segregation agent."""
+
     def __init__(self, model, agent_type: int) -> None:
         """Create a new Schelling agent.
         Args:
@@ -16,21 +18,21 @@ class SchellingAgent(Agent):
         neighbors = self.model.grid.iter_neighbors(
             self.pos, moore=True, radius=self.model.radius
         )
-        
+
         # Filter out empty cells
         similar_neighbors = [
-            neighbor for neighbor in neighbors 
-            if hasattr(neighbor, 'type') and neighbor.type == self.type
+            neighbor
+            for neighbor in neighbors
+            if hasattr(neighbor, "type") and neighbor.type == self.type
         ]
         total_neighbors = [
-            neighbor for neighbor in neighbors 
-            if hasattr(neighbor, 'type')
+            neighbor for neighbor in neighbors if hasattr(neighbor, "type")
         ]
-        
+
         # Calculate fraction of similar neighbors
         if len(total_neighbors) > 0:
             similarity_fraction = len(similar_neighbors) / len(total_neighbors)
-            
+
             # If unhappy, move to a random empty cell
             if similarity_fraction < self.model.homophily / 8.0:
                 self.model.grid.move_to_empty(self)
--- a/mesa/examples/basic/schelling/agents.py
+++ b/mesa/examples/basic/schelling/agents.py
@@ -23,7 +23,7 @@ class SchellingAgent(Agent):
         similar_neighbors = len([n for n in neighbors if n.type == self.type])
 
         # Calculate the fraction of similar neighbors
-        if (valid_neighbors := len(neighbors) )> 0:
+        if (valid_neighbors := len(neighbors)) > 0:
             similarity_fraction = similar_neighbors / valid_neighbors
 
             # If unhappy, move to a random empty cell
--- a/mesa/examples/basic/schelling/agents.py
+++ b/mesa/examples/basic/schelling/agents.py
@@ -15,21 +15,15 @@ class SchellingAgent(Agent):
 
     def step(self) -> None:
         """Determine if agent is happy and move if necessary."""
-        neighbors = self.model.grid.iter_neighbors(
+        neighbors = self.model.grid.get_neighbors(
             self.pos, moore=True, radius=self.model.radius
         )
 
-        valid_neighbors = 0
-        similar_neighbors = 0
-
-        for neighbor in neighbors:
-            if hasattr(neighbor, "type"):  # Exclude empty cells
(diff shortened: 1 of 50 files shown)


mesa/examples/basic/schelling/agents.py
```diff
@@ -20,11 +19,22 @@ def step(self) -> None:
...
+        if len(total_neighbors) > 0:
+            similarity_fraction = len(similar_neighbors) / len(total_neighbors)
 
-        # If unhappy, move to a random empty cell:
-        if similar < self.model.homophily:
-            self.model.grid.move_to_empty(self)
-        else:
-            self.model.happy += 1
+            # If unhappy, move to a random empty cell
+            if similarity_fraction < self.model.homophily / 8.0:
```
quaquel, line 37: why not change homopily at the model level to be a fraction? That makes the model independent of the neighborhood size.

mesa/examples/basic/schelling/agents.py
```diff
@@ -20,11 +19,22 @@ def step(self) -> None:
...
-        # Count similar neighbors
-        similar = sum(neighbor.type == self.type for neighbor in neighbors)
+        # Filter out empty cells
+        similar_neighbors = [
+            neighbor
+            for neighbor in neighbors
+            if hasattr(neighbor, "type") and neighbor.type == self.type
+        ]
+        total_neighbors = [
+            neighbor for neighbor in neighbors if hasattr(neighbor, "type")
```
quaquel, line 29: you iterate twice, with one check being done in both. You can make this more efficient by looping only once.

EwoutH:
@Sahil-Chhoker Thanks for your PR. When do you expect to be able to incorporate @quaquel's feedback?

If you have any question about it feel free to ask!

Sahil-Chhoker:
Thank you, @quaquel, for your review! Could you please provide more details about the model and how can I can make changes to it?

quaquel:
My feedback is quite clear, and there is a link to a more detailed description of the model in the original issue. I requested 2 well-defined changes to this PR: change `homophily` to a fraction and rewrite so you only loop once.

From your reaction, I therefore deduce that you are not particularly familiar with agent-based modeling.

mesa/examples/basic/schelling/agents.py
```diff
@@ -19,22 +19,23 @@ def step(self) -> None:
...
-        ]
-
-        # Calculate fraction of similar neighbors
-        if len(total_neighbors) > 0:
-            similarity_fraction = len(similar_neighbors) / len(total_neighbors)
+        valid_neighbors = 0
+        similar_neighbors = 0
+
+        for neighbor in neighbors:
+            if hasattr(neighbor, "type"):  # Exclude empty cells
```
quaquel, line 26: this is not needed, `iter_neighbors` returns a list of agents. In this model all agents have the `type` attribute.

mesa/examples/basic/schelling/agents.py
```diff
@@ -19,22 +19,23 @@ def step(self) -> None:
...
+        # Calculate the fraction of similar neighbors
+        if valid_neighbors > 0:
+            similarity_fraction = similar_neighbors / valid_neighbors
 
             # If unhappy, move to a random empty cell
-            if similarity_fraction < self.model.homophily / 8.0:
+            if similarity_fraction < self.model.homophily:
                 self.model.grid.move_to_empty(self)
             else:
                 self.model.happy += 1
```
quaquel, line 39:
```
        neighbors = self.model.grid.get_neighbors(
            self.pos, moore=True, radius=self.model.radius
        )

        # Count similar neighbors
        similar_neighbors = len([n for n in neighbors if n.type == self.type])

        # Calculate the fraction of similar neighbors
        if (valid_neighbors := len(neighbors) )> 0:
            similarity_fraction = similar_neighbors / valid_neighbors

            # If unhappy, move to a random empty cell
            if similarity_fraction < self.model.homophily:
                self.model.grid.move_to_empty(self)
            else:
                self.model.happy += 1
```

quaquel:
Please switch from `iter_neighbors` to `get_neighbors`.

```
        neighbors = self.model.grid.get_neighbors(
            self.pos, moore=True, radius=self.model.radius
        )
```

currently, as you can see the tests fail because the model won't run. The list expression exhausts the iterator, so getting the number of valid neighbors fails.

mesa/examples/basic/schelling/agents.py
```diff
@@ -16,15 +15,19 @@ def __init__(self, model, agent_type: int) -> None:
...
     def step(self) -> None:
         """Determine if agent is happy and move if necessary."""
-        neighbors = self.model.grid.iter_neighbors(
+        neighbors = self.model.grid.get_neighbors(
             self.pos, moore=True, radius=self.model.radius
         )
 
         # Count similar neighbors
-        similar = sum(neighbor.type == self.type for neighbor in neighbors)
+        similar_neighbors = len([n for n in neighbors if n.type == self.type])
```
EwoutH, line 23: The model looks significantly slower. I’m curious if we could minimize that performance degradation, I suspect the main problem is with this line doing additional work.

quaquel: The slowdown is not due to this line of code but because the behavior changes a lot. Prior to this PR, the model would reach steady state in a few ticks. Now, it can take many ticks and frequently the model never reaches stable state.

mesa/examples/basic/schelling/agents.py
```diff
@@ -16,15 +15,19 @@ def __init__(self, model, agent_type: int) -> None:
...
-        # If unhappy, move to a random empty cell:
-        if similar < self.model.homophily:
-            self.model.grid.move_to_empty(self)
-        else:
-            self.model.happy += 1
+            # If unhappy, move to a random empty cell
+            if similarity_fraction < self.model.homophily:
+                self.model.grid.move_to_empty(self)
+            else:
+                self.model.happy += 1
```
quaquel, line 33: what happens if `len(neighbors)` is 0? Currently, nothing happens, which most definitely is not correct.

Sahil-Chhoker: Please correct me if I’m wrong, but if there are no neighbors surrounding the agent, it should be considered unhappy as it does not meet the criteria for happiness. Should I proceed with this approach?

quaquel: I have checked some literature, including the original article. It's not entirely clear. However, your reading is defendable, so I am fine with it.

quaquel (approved):

quaquel:
I am not surprised as indicated before, but I still want to check before merging.

quaquel:
I have checked the model, and something is not quite right. Within a few steps, all agents are supposedly happy but if you check the grid manually, you see that this cannot be true. I am not yet sure what's causing this. It might be merely the order of agent activation (so an agent goes first and is happy, but some of its neighbors move afterwards, meaning it is unhappy if checked again). I need time to look more closely at this and reason true what is going on.

quaquel:
I figured out what was going on, an the mistake was on my side. I am merging this. Thanks @Sahil-Chhoker .

EwoutH:
Awesome, thanks both for your work!

Sahil-Chhoker:
Hey @quaquel thanks for merging this PR, can you please explain what was the problem and how did you fix that because the model was behaving the same on my machine as well.

PR #2520, opened 2024-11-26
Added homophily ratio in basic schelling example

### LINKED ISSUE #2515: homophily parameter in basic schelling example

vbv-shm:
Changed homophily comparing logic as per discussion in #2515.
Now homophily_ratio is checked for agents to be happy. homophily_ratio is similar neighboring agents divided by the total number of neighbors.

### DIFF
--- a/benchmarks/configurations.py
+++ b/benchmarks/configurations.py
@@ -35,7 +35,7 @@ configurations = {
             "parameters": {
                 "height": 40,
                 "width": 40,
-                "homophily": 3,
+                "homophily": 0.4,
                 "radius": 1,
                 "density": 0.625,
             },
@@ -47,7 +47,7 @@ configurations = {
             "parameters": {
                 "height": 100,
                 "width": 100,
-                "homophily": 8,
+                "homophily": 0.4,
                 "radius": 2,
                 "density": 0.8,
             },
--- a/benchmarks/configurations.py
+++ b/benchmarks/configurations.py
@@ -47,7 +47,7 @@ configurations = {
             "parameters": {
                 "height": 100,
                 "width": 100,
-                "homophily": 0.4,
+                "homophily": 1,
                 "radius": 2,
                 "density": 0.8,
             },
--- a/mesa/examples/basic/schelling/model.py
+++ b/mesa/examples/basic/schelling/model.py
@@ -9,11 +9,11 @@ class Schelling(Model):
 
     def __init__(
         self,
-        height: int = 40,
-        width: int = 40,
+        height: int = 20,
+        width: int = 20,
         density: float = 0.8,
         minority_pc: float = 0.5,
-        homophily: int = 3,
+        homophily: float = 0.4,
         radius: int = 1,
         seed=None,
     ):
--- a/mesa/examples/basic/schelling/app.py
+++ b/mesa/examples/basic/schelling/app.py
@@ -26,7 +26,7 @@ model_params = {
     },
     "density": Slider("Agent density", 0.8, 0.1, 1.0, 0.1),
     "minority_pc": Slider("Fraction minority", 0.2, 0.0, 1.0, 0.05),
-    "homophily": Slider("Homophily", 0.3, 0.0, 0.8, 0.1),
+    "homophily": Slider("Homophily", 0.4, 0.0, 1.0, 0.1),
     "width": 20,
     "height": 20,
 }
--- a/mesa/examples/basic/schelling/app.py
+++ b/mesa/examples/basic/schelling/app.py
@@ -26,7 +26,7 @@ model_params = {
     },
     "density": Slider("Agent density", 0.8, 0.1, 1.0, 0.1),
     "minority_pc": Slider("Fraction minority", 0.2, 0.0, 1.0, 0.05),
-    "homophily": Slider("Homophily", 0.4, 0.0, 1.0, 0.1),
+    "homophily": Slider("Homophily", 0.4, 0.0, 1.0, 0.125),
     "width": 20,
     "height": 20,
 }


mesa/examples/basic/schelling/agents.py
```diff
@@ -20,11 +20,19 @@ def step(self) -> None:
...
         )
 
+        type_of_neighbors = [neighbor.type == self.type for neighbor in neighbors]
+
         # Count similar neighbors
-        similar = sum(neighbor.type == self.type for neighbor in neighbors)
+        similar = sum(type_of_neighbors)
+
+        # Count total neighbors
+        total_neighbors = len(type_of_neighbors)
```
quaquel, line 29: these numbers will always be the same so this PR does not solve the problem.

mesa/examples/basic/schelling/model.py
```diff
@@ -13,7 +13,7 @@ def __init__(
         width: int = 40,
         density: float = 0.8,
         minority_pc: float = 0.5,
-        homophily: int = 3,
+        homophily_ratio: float = 0.3,
```
quaquel, line 16: why 0.3? It would make more sense to go to 3/8 (8 is the neighborhood size of a Moore grid with a radius of 1 and all fields occupied.

vbv-shm: Thank you for reviewing. I will change 0.3 to 3/8 (0.375).

mesa/examples/basic/schelling/agents.py
vbv-shm: image

I am getting these values while printing.

EwoutH:
Thanks for the PR! Could you update your PR description to contain all the necessary information to understand this change on its own? You can use [this template](https://github.com/projectmesa/mesa/blob/main/.github/PULL_REQUEST_TEMPLATE/feature.md?plain=1) if you want.

vbv-shm:
> Thanks for the PR! Could you update your PR description to contain all the necessary information to understand this change on its own? You can use [this template](https://github.com/projectmesa/mesa/blob/main/.github/PULL_REQUEST_TEMPLATE/feature.md?plain=1) if you want.
> 
> **Edit:** I saw there's also this PR open, which also modifies schelling:
> 
> * [Fix: Schelling Model Neighbor Similarity Calculation #2518](https://github.com/projectmesa/mesa/pull/2518)
> 
> Are there any conflicts between them? I would tend to let #2518 go first, since it was opened earlier. (or did we assign someone to this specific issue?)

vbv-shm:
@EwoutH Thank you for the reply.

vbv-shm:
@EwoutH Thank you for the reply.
I have understood that #2815 will be given priority because it was opened first. Still, can I work on this PR just in case #2815 doesn't work out? Or is this PR required to be closed?

quaquel:
I have provided feedback in #2518, if that is not addressed in the coming days and there is no further response, I am fine with merging this instead.

mesa/examples/basic/schelling/model.py
vbv-shm: @quaquel The step size of the slider is 0.1. So the ratio can either be 0.3 or 0.4. 
If the ratio is 0.3, agents with 3 similar neighbors and 8 total neighbors will be happy, as 3/8 = 0.375. 
And if the ratio is 0.4, agents with 3 similar neighbors and 8 total neighbors won't be happy.
So should I change 0.3 to 0.4?

vbv-shm:
@quaquel Thank you. i understand.

mesa/examples/basic/schelling/model.py
EwoutH: Maybe we can change the step size to 0.125 as well?

vbv-shm:
@EwoutH May you please run the benchmark tests?

vbv-shm:
@quaquel @EwoutH Previously the benchmark tests were failing because I changed `homophily` to `homophily_ratio` in the Schelling example. `homophily` was provided by the configuration file while the model now takes `homophily_ratio` as input. In order to fix this, I changed `homophily` to `homophily_ratio` inside the configuration. 
I am curious if my fix is right or not. Could you please run the benchmark tests so that  changes can be checked?

vbv-shm:
Looks like benchmarks ran, but performance is not up to the mark. So changing `homophily` to `homophily_ratio` inside the configuration worked.

EwoutH:
Thanks for your work. We’re currently waiting on #2518 to be performance validated, one that’s done we can continue with rebasing this PR and checking it’s performance properly.

It’s great it now runs without problem!

vbv-shm:
Thank you

quaquel:
I have merged #2518, so let us know if you want to rebase this and further refine the example.

vbv-shm:
@quaquel Yes, I would like to refine this. A few things that I think can be updated are: 
1. Update configuration.py file for correct values for running benchmarks
2. Change default value of homophily from 0.3
3. Update slider step size in visualization
As I add these changes, if something else is encountered, I will add that as well. 
Please provide your thoughts on this and anything else required to be added.

EwoutH:
Great, make sure to [git rebase](https://docs.github.com/en/get-started/using-git/about-git-rebase) first and [resolve merge conflicts](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/addressing-merge-conflicts/resolving-a-merge-conflict-using-the-command-line).

vbv-shm:
I was running scheduling example locally. 
image
No agents were getting happy, and the happy line is constant, as can be seen in the picture above.


But after I pressed reset's grid shape changed, and it behaved normally, and agents are getting happy according to condition. Please see below image:
image


I remember 2 weeks back when I ran it; it was running fine even without pressing reset, although the grid shape changed back then also.

mesa/examples/basic/schelling/model.py
```diff
@@ -9,11 +9,11 @@ class Schelling(Model):
 
     def __init__(
         self,
-        height: int = 40,
-        width: int = 40,
+        height: int = 20,
+        width: int = 20,
```
EwoutH, line 13: I'm curious, what's the motivation for making the grid smaller?

vbv-shm: @EwoutH It was just to fix the problem I mentioned above. The problem was the grid size was changing from 40 by 40 to 20 by 20 after pressing RESET button. Before reset, grid size is what is provided by default in the definition of Schelling model(40 by 40), and after pressing RESET , grid size is what is passed into SolaraViz class for making visualization ( 20 by 20 ).  
So changing grid size from 40 by 40 to 20 by 20 inside the model makes grid size uniform. However, 40 by 40 grid size could also be passed inside SolaraViz class instead of 20 by 20.

benchmarks/configurations.py
```diff
@@ -47,7 +47,7 @@
             "parameters": {
                 "height": 100,
                 "width": 100,
-                "homophily": 8,
+                "homophily": 0.4,
```
quaquel, line 50: this was 8 (the max for moore neighborhood, so it should be 1 now.

quaquel (commented):
one minor point remaining

benchmarks/configurations.py
vbv-shm: @quaquel I will change it.

vbv-shm:
@quaquel I am assuming you are referring to the step size of the slider. Should I change slider step size from 0.1 to 0.125 ? Then it also makes sense to change the required homophily ratio from 0.4 to 0.375.
If these changes are made or not, 4 out of 8 similar agents are required in neighbour in order for agent to be happy. That is why I was doubtful to add these or not.

vbv-shm:
@quaquel I have made the changes. Please check.

quaquel (approved):

vbv-shm:
@quaquel thank you for checking and merging. I learned a lot working on it. :)

PR #2524, opened 2024-12-01
Fix: running Mesa in Docker with Schelling model

AdamZh0u:
## Summary
This pull request addresses issues with the existing Docker configuration, including missing dependencies and untested references to the mesa-examples repository. The updates ensure that the Mesa Schelling model runs correctly with Solara visualization in Docker on port 8765.

## Bug / Issue
1. Missing Dependencies: The original Dockerfile did not install the mesa[all] package, which led to a solara not found error when attempting to run the docker compose.
2. Outdated Example Repository: The mesa-examples repository referenced in the original Dockerfile contains broken visualization modules and no longer includes the Schelling model.
3. Impact: Users were unable to run the Schelling model or any visualizations successfully using the provided Docker configuration.

## Implementation
1. Dependencies Fix: Updated the Dockerfile to ensure mesa[all] is installed, which includes all necessary dependencies for Solara-based visualization.
2. Updated Model Reference:
Replaced the mesa-examples repository reference with the mesa.examples module, specifically pointing to the built-in Schelling model:
```
ENV MODEL_DIR=/opt/mesa/mesa/examples/basic/schelling
```

## Testing
1. Started the container using docker compose:
```
docker compose up -d
```
2. Verified that the Schelling model runs correctly with a working visualization.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,7 +31,7 @@ COPY . /opt/mesa
 
 EXPOSE 8765/tcp
 
-RUN pip3 install -e /opt/mesa[all]
+RUN pip3 install -e /opt/mesa[rec]
 
 CMD ["sh", "-c", "cd $MODEL_DIR && solara run app.py --host=0.0.0.0"]
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,8 +7,8 @@ FROM python:bookworm
 LABEL maintainer="rht <rhtbot@protonmail.com>"
 
 # To use this Dockerfile:
-# 1. `docker build . -t mymesa_image`
-# 2. `docker run --name mymesa_instance -p 8765:8765 -it mymesa_image`
+# 1. `docker build . -t mesa_image`
+# 2. `docker run --name mesa_instance -p 8765:8765 -it mesa_image`
 # 3. In your browser, visit http://127.0.0.1:8765
 #
 # Currently, this Dockerfile defaults to running the Schelling model, as an
@@ -19,7 +19,7 @@ LABEL maintainer="rht <rhtbot@protonmail.com>"
 # `docker run --name mymesa_instance -p 8765:8765 -e MODEL_DIR=/mesa-examples/examples/sugarscape_cg -it mymesa_image`
 # Note: the model directory MUST contain an app.py file.
 
-ENV MODEL_DIR=/mesa-examples/examples/schelling_experimental
+ENV MODEL_DIR=/opt/mesa/mesa/examples/basic/schelling
 
 # Don't buffer output:
 # https://docs.python.org/3.10/using/cmdline.html?highlight=pythonunbuffered#envvar-PYTHONUNBUFFERED
@@ -29,10 +29,11 @@ WORKDIR /opt/mesa
 
 COPY . /opt/mesa
 
-RUN cd / && git clone https://github.com/projectmesa/mesa-examples.git
-
 EXPOSE 8765/tcp
 
-RUN pip3 install -e /opt/mesa
+RUN pip3 install -e /opt/mesa[all]
 
 CMD ["sh", "-c", "cd $MODEL_DIR && solara run app.py --host=0.0.0.0"]
+
+# To check file system:
+# docker exec -it mesa_instance bash
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,8 @@
 # We can't use slim because we need either git/wget/curl to
 # download mesa-examples, and so installing them requires
 # updating the system anyway.
-# We can't use alpine because NumPy doesn't support musllinux yet.
-# But it's in the RC version https://github.com/numpy/numpy/issues/20089.
 FROM python:bookworm
-LABEL maintainer="rht <rhtbot@protonmail.com>"
+LABEL maintainer="projectmesa maintainers@projectmesa.dev"
 
 # To use this Dockerfile:
 # 1. `docker build . -t mesa_image`
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,6 +16,6 @@ services:
       # within the current directory.
       # E.g. if it is at my-model, then you specify it as
       # /opt/mesa/my-model.
-      MODEL_DIR: /mesa-examples/examples/schelling_experimental
+      MODEL_DIR: /opt/mesa/mesa/examples/basic/schelling
     ports:
       - 8765:8765


EwoutH:
Thanks for the PR. @projectmesa/maintainers or @AdamZh0u, can someone remind me what the use case for running Mesa in docker is exactly?

wang-boyu:
Well, I guess the motivation is probably similar for many projects that use docker. You can get something up and running locally with one single `docker run xxx` or `docker compose up`, instead of having to 1) clone the repo, 2) set up a virtual environment, 3) install dependencies, and 4) finally run some models.

I do not have docker installed on my mac, because it's not as convenient as it is on ubuntu. There might be others who prefer to use docker.

jackiekazil:
+1 to what @wang-boyu said. 
Would love to hear from @AdamZh0u though as well.

AdamZh0u:
Hi all,

Docker allows seamless deployment of Mesa applications across Windows, macOS, and Linux without any platform-specific differences. It ensures that all dependencies and environments are consistent, making it easier for users to get started quickly.

In fact, on macOS, you can run an example model with just two simple commands:
```bash
brew install docker  
#cd mesa  
docker compose up -d  
```
This simplicity is particularly useful for users who want a hassle-free setup experience.

AdamZh0u:
No worries! I’ve actually started promoting Docker because I’m leading a Python course with over 100 students, and their systems range from macOS 10 to Windows Home version. We’ve been using Docker, and it’s been working smoothly on almost all computers, making things easier for everyone.

Dockerfile
```diff
@@ -29,10 +29,11 @@ WORKDIR /opt/mesa
 
 COPY . /opt/mesa
 
-RUN cd / && git clone https://github.com/projectmesa/mesa-examples.git
-
 EXPOSE 8765/tcp
 
-RUN pip3 install -e /opt/mesa
+RUN pip3 install -e /opt/mesa[all]
```
EwoutH, line 34: I think `rec` is enough here, since `all` also installs dev dependencies and such.

```suggestion
RUN pip3 install -e /opt/mesa[rec]
```

EwoutH (approved):
I don’t see anything weird in the changeset, so no objections from my side. I will leave final review and merge to a maintainer more familiar with (our) Docker.

EwoutH:
> I’ve actually started promoting Docker because I’m in a Python course with over 100 students

I would love to hear more about the course you’re teaching!

Dockerfile
AdamZh0u: I tested rec and it works.

Dockerfile
```diff
@@ -7,8 +7,8 @@ FROM python:bookworm
 LABEL maintainer="rht <rhtbot@protonmail.com>"
```
tpike3, line 7: `LABEL maintainer="projectmesa <maintainers@projectmesa.dev>"

tpike3 (approved):
Thanks @AdamZh0u I just asked for some comment updates

But like @EwoutH I would be curious to here about your class

Dockerfile
AdamZh0u: Modified accroding to the suggestions

AdamZh0u:
Hi @EwoutH  and @tpike3

Thanks for your interest!  I’m involved in teaching a basic Python course this term, and Docker has been really helpful in making everything run smoothly across different systems. Next term, we’ll focus partly on ABM, so I’m exploring Mesa for this, trying to avoid the need to learn another language like NetLogo.

wang-boyu (approved):

wang-boyu:
LGTM. Merging now.

Thanks @AdamZh0u for your first contribution to Mesa!

EwoutH:
@AdamZh0u Cool, thanks for sharing!

SongshGeo:
> No worries! I’ve actually started promoting Docker because I’m in a Python course with over 100 students, and their systems range from macOS 10 to Windows Home version. We’ve been using Docker, and it’s been working smoothly on almost all computers, making things easier for everyone.

Thanks for your contribution! @AdamZh0u 

I just tested the docker, and it works well. One small comment on your message: I think it's also necessary to install `docker-compose` (maybe `brew install docker-compose` as well to run the demo.)

SongshGeo:
> > Well, I guess the motivation is probably similar for many projects that use docker. You can get something up and running locally with one single `docker run xxx` or `docker compose up`, instead of having to 1) clone the repo, 2) set up a virtual environment, 3) install dependencies, and 4) finally run some models.
> 
> Just realized that I was wrong about this. Since the image is not released on docker hub, users still have to 1) clone the repo, 2) build the image, and 3) run docker compose
> 
> If our image is on docker hub, then it's possible to just do `docker run xxx` without having to clone the repo.

Yes. Fair enough. I vote to be involved in the docker hub in the future.

AdamZh0u:
Hey! I think I need to make it clearer about two different workflows for using Mesa with Docker:

For Developers:
The current Docker workflow (as described in the README under [Running Mesa in Docker](https://github.com/projectmesa/mesa?tab=readme-ov-file#running-mesa-in-docker)) works great for someone who wants to contribute to the Mesa project. Just run `docker compose up -d` after cloning Mesa. Since we have build . pointing to the Dockerfile, you don't need to run docker build. The -e flag in pip3 install -e /opt/mesa[rec] means you can edit Mesa code and see changes immediately. You can also switch to your own models by changing MODEL_DIR in docker-compose.yml. This is really convenient for anyone interested in developing both ABM models and the Mesa project.

For Users:
If just using Mesa (not developing it) to build ABM models, a different approach might work better. We could publish Mesa images to DockerHub (though this would need maintenance), letting users develop their models in a pre-built Mesa environment. This would simplify the setup process significantly.

@SongshGeo Thanks for catching that! Just checked - actually this is already covered in the README where it lists Docker Compose installation as the first step.

wang-boyu:
> For Users: If just using Mesa (not developing it) to build ABM models, a different approach might work better. We could publish Mesa images to DockerHub (though this would need maintenance), letting users develop their models in a pre-built Mesa environment. This would simplify the setup process significantly.

@projectmesa/maintainers do we want this? Things needed from our end would be (feel free to add more if I miss any):

- update our documentation on how to use the image from docker hub directly
- update our release workflow to tag and release docker image with version number on each release, in addition to current pypi release
- future maintenance

This was previously discussed in https://github.com/projectmesa/mesa/issues/558#issuecomment-413712735. Interestinly there's an unmaintained Mesa image under CoMESE: https://hub.docker.com/r/comses/mesa.

PR #2510, opened 2024-11-16
Updated docs and check_model param

nissu99:
## Summary
This PR adds documentation for the SolaraViz instantiation and adds a new test case to handle arguments and keyword arguments in solaraviz.
## Bug / Issue
 SolaraViz docs #2494 

## Implementation
changes in docs and explicit check in _check_model_params.

## Testing

Screenshot from 2024-11-17 16-30-08

## Additional Notes
<!-- Add any additional information that may be relevant for the reviewers, su

PR #2536, opened 2024-12-06
Add sample and choices to CellCollection

### LINKED ISSUE #2530: Cell space: Place agents in AgentSet randomly on (empty) cells
Often you create a bunch of Agents, and then you need to place them on the grid. Either you loop over them, or do some other unwieldy thing, none of it is elegant.

So it would be great if there was a simple function that would get all agents and place them on, optionally empty, cells. This could be implemented faster than placing them individually, since now only once has to be checked which cells are available / empty.

It's especially handy to be used together with `create_agents` (https://github.com/projectmesa/mesa/pull/2351), combined you can create and place all your agents with two lines of code.

```Python
some_agents = Ant.create_agents(n=100)
self.space.place_agents(some_agents, only_empty=True)
```

quaquel:
This adds 2 new methods to cell collection: `sample_random_cells` and `choices_random_cells`. These methods make it possible to sample cells with or without replacement from a cell collection. This closes #2530. 

Implementation-wise, the methods are CellCollection specific wrappers around `random.sample` and `random.choices`.  Note also that both methods return a list. A CellCollection cannot contain duplicates (it is, in essence, a fixed set). But both sampling and choices can return duplicates, so we cannot return a cell collection. 

The main use case for these methods is in conjunction with the `Agent.create_agents`. You can. now do

```Python
class MyAgent(CellAgent):
        super().__init__(model)
        self.cell = cell

# let's create 10 agents, placed in 10 random empty cells
MyAgent.create_agents(model, 10, self.space.empties.sample_random_cells(k=10))

# let's create 10 agents, placed in 10 random cells
MyAgent.create_agents(model, 10, self.space.all_cells.choices_random_cells(k=10))


```

### DIFF
--- a/docs/tutorials/visualization_tutorial.ipynb
+++ b/docs/tutorials/visualization_tutorial.ipynb
@@ -3,7 +3,9 @@
   {
    "cell_type": "markdown",
    "metadata": {},
-   "source": "# Visualization Tutorial"
+   "source": [
+    "# Visualization Tutorial"
+   ]
   },
   {
    "cell_type": "markdown",
@@ -51,10 +53,18 @@
    ]
   },
   {
-   "metadata": {},
    "cell_type": "code",
-   "outputs": [],
-   "execution_count": null,
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Mesa version: 3.1.0\n"
+     ]
+    }
+   ],
    "source": [
     "import mesa\n",
     "print(f\"Mesa version: {mesa.__version__}\")\n",
@@ -66,10 +76,10 @@
    ]
   },
   {
-   "metadata": {},
    "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
    "outputs": [],
-   "execution_count": null,
    "source": [
     "def agent_portrayal(agent):\n",
     "    return {\n",
@@ -79,15 +89,17 @@
    ]
   },
   {
-   "metadata": {},
    "cell_type": "markdown",
-   "source": "In addition to the portrayal method, we instantiate the model parameters, some of which are modifiable by user inputs. In this case, the number of agents, N, is specified as a slider of integers."
+   "metadata": {},
+   "source": [
+    "In addition to the portrayal method, we instantiate the model parameters, some of which are modifiable by user inputs. In this case, the number of agents, N, is specified as a slider of integers."
+   ]
   },
   {
-   "metadata": {},
    "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
    "outputs": [],
-   "execution_count": null,
    "source": [
     "model_params = {\n",
     "    \"n\": {\n",
@@ -104,8 +116,8 @@
    ]
   },
   {
-   "metadata": {},
    "cell_type": "markdown",
+   "metadata": {},
    "source": [
     "Next, we instantiate the visualization object which (by default) displays the grid containing the agents, and timeseries of values computed by the model's data collector. In this example, we specify the Gini coefficient.\n",
     "\n",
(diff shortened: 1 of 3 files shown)


EwoutH:
Awesome, thanks for working on this!

Personally, since they are so similar, I would go for a single method (i.e. `random` or `draw`) with a keyword argument (`replace` or `allow_duplicates`). Like:

```Python
# Without replacement (equivalent to previous sample_random_cells)
cells = collection.random(k=10, replace=False)

# With replacement (equivalent to previous choices_random_cells)
cells = collection.random(k=10, replace=True)
```

quaquel:
I have been going back and forth on this and in the end decided to stick close to the naming in the random library. Note that this also makes it possible to fully shadow the arguments from both methods which is virtually impossible if we collapse it into a single method with a boolean.

EwoutH:
That's a fair point. In that case I would maybe suggest `random_sample` and `random_choice`. Cells is implied, because it's a method of a CellCollection, and starting them both with random helps when looking through methods alphabetically (in IDEs, etc.)

quaquel:
> Cells is implied, because it's a method of a CellCollection

That is not true. We already have a `select_random_agent` method on cell collection.

EwoutH:
I would say that’s because those are agents selected from a collection of cells. That warrant naming it that way.

In the AgentSet we also name methods `select` and `do`, not `select_agents` and `do_agents`.

quaquel:
Ok, but than I prefer simply `sample`, `choices`, (and perhaps `choice`) which is fully consistent with the naming of these methods in `random`.

EwoutH:
I think the `random_` suffix adds some value here (it’s also `random.sample`, and a sample or choice can also be non-random) but I’m also good without the suffix.

docs/tutorials/visualization_tutorial.ipynb
```diff
@@ -320,42 +352,42 @@
...
+      ]
      },
      "metadata": {},
      "output_type": "display_data"
     }
    ],
-   "execution_count": 12
+   "source": [
+    "page = SolaraViz(\n",
+    "    model1,\n",
```
tpike3, line 383: With the merging of #2510 can you change this to `model` vs `model1`

docs/tutorials/visualization_tutorial.ipynb
```diff
@@ -366,35 +398,35 @@
...
+       "Cannot show ipywidgets in text"
+      ]
      },
      "metadata": {},
      "output_type": "display_data"
     }
    ],
-   "execution_count": 13
+   "source": [
+    "Histogram(model1)"
```
tpike3, line 428: Change to `model` instead of `model1`

docs/tutorials/visualization_tutorial.ipynb
```diff
@@ -118,10 +130,28 @@
...
+   "outputs": [
+    {
+     "data": {
+      "application/vnd.jupyter.widget-view+json": {
+       "model_id": "7a906b76fdb74d2784c4f152f18931e7",
+       "version_major": 2,
+       "version_minor": 0
+      },
+      "text/html": [
+       "Cannot show widget. You probably want to rerun the code cell above (<i>Click in the code cell, and press Shift+Enter <kbd>⇧</kbd>+<kbd>↩</kbd></i>)."
```
tpike3, line 145: Can you clear the cell outputs to remove these

mesa/experimental/cell_space/cell_collection.py
```diff
@@ -110,6 +110,41 @@ def select_random_agent(self) -> CellAgent:
         """
         return self.random.choice(list(self.agents))
 
+    def sample_random_cells(self, k: int = 1, counts: list | None = None) -> list[Cell]:
+        """Sample k unique cells from the cell collection.
+
+        See random.sample for more information.
```
tpike3, line 116: Trying to improve the API would it be better to be more explicit here 

`See Python's random.sample at python.docs.org for more information`

tpike3:
This is awesome! I had some nitpicky comments that I am not tied to at all. 

I am good with merging I think only issue is changing the `model1` in the visualization tutorial

Corvince:
Oh, just saw this PR after my comments in #2530 . It does indeed seems to work to use `cells` directly. Honestly, I am not sure of the benefit of these methods, since they are indeed just thin wrappers and don't involve any logic themselves. But, yes, it helps discoverability so I am also not against it.

quaquel:
I had overlooked the cells property when I made this PR. I am now, like @Corvince, not sure whether we need to merge this. I prefer to keep the API simple. Accidentally, we still have `select_random_agent` which is also in a way redundant because you can do `self.random.choice(cellcollection.agents)`. 

@EwoutH what do you think?

EwoutH:
I agree it's more Pythonic to use built-ins. Good catch @Corvince.

The fact that we didn't realize it for this long, does say something about the discoverability. I think we can cover it with good documentation, but definitely something we would want to highlight in example models and a future tutorial.

quaquel:
closing this as not needed as @Corvince  pointed out.

Should we add a deprecation warning to `select_random_agent` as well?

Issue #2497, opened 2024-11-12
SolaraViz: Add Slider controlling `play_interval` by default


EwoutH:
Aside from the current buttons, having by default a Slider which controls the `play_interval` in SolaraViz would be very useful. It would allow controlling the speed directly from SolaraViz, not only on initialisation.

wang-boyu:
Hi @EwoutH would you like to close this? Or is there anything missing from https://github.com/projectmesa/mesa/pull/2540 that needs to be further developed

EwoutH:
Yes, can indeed be closed!

PR #2540, opened 2024-12-08
Add Interactive Play Interval Control to Mesa Visualization

AdamZh0u:
### Summary
This PR adds a slider control to dynamically adjust simulation speed in Mesa's visualization interface. #2497 

### Motive
- Fixed play interval was limiting user control
- Need flexible speed control for observing different phenomena

### Implementation
- Added play interval slider (1ms-500ms)
- Made interval reactive for real-time updates
- Updated controllers to use dynamic interval

### Usage Examples
The play interval slider appears automatically in the Controls panel of any Mesa visualization, no additional configuration needed.

- Before

- After

### Additional Notes
- Maintains backward compatibility
- Safe range limits prevent performance issues

### DIFF
--- a/mesa/visualization/solara_viz.py
+++ b/mesa/visualization/solara_viz.py
@@ -111,7 +111,7 @@ def SolaraViz(
         with solara.Card("Controls"):
             solara.SliderInt(
                 label="Play Interval (ms)",
-                value=reactive_play_interval.value,
+                value=reactive_play_interval,
                 on_value=lambda v: reactive_play_interval.set(v),
                 min=1,
                 max=500,
@@ -121,14 +121,14 @@ def SolaraViz(
                 ModelController(
                     model,
                     model_parameters=reactive_model_parameters,
-                    play_interval=reactive_play_interval.value,
+                    play_interval=reactive_play_interval,
                 )
             else:
                 SimulatorController(
                     model,
                     simulator,
                     model_parameters=reactive_model_parameters,
-                    play_interval=reactive_play_interval.value,
+                    play_interval=reactive_play_interval,
                 )
         with solara.Card("Model Parameters"):
             ModelCreator(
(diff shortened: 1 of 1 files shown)


quaquel:
I like this PR, but I defer to @Corvince's view on it since he has been doing most of the work on the visualization stuff.

mesa/visualization/solara_viz.py
```diff
@@ -102,24 +102,33 @@ def SolaraViz(
...
 
     with solara.AppBar():
         solara.AppBarTitle(name if name else model.value.__class__.__name__)
 
     with solara.Sidebar(), solara.Column():
         with solara.Card("Controls"):
+            solara.SliderInt(
+                label="Play Interval (ms)",
+                value=reactive_play_interval.value,
+                on_value=lambda v: reactive_play_interval.set(v),
```
Corvince, line 115: I think you can replace the 2 value lines above with just `value=reactive_play_interval` . Since its a reactive variable it will automagically extract the value and update

Corvince (approved):

Corvince:
Thanks for tagging me @quaquel, I wasn't looking into mesa a lot the last weeks, so this would have gone unnoticed by me. Nice, focused PR. I approved, but I also left a comment that you could try out @AdamZh0u . But it is also fine as-is.

AdamZh0u:
@Corvince Thanks for the comments! You helped me realize an important point - the play_interval should be passed as a reactive object to controllers. Without maintaining its reactive nature, the controllers won't be able to dynamically update when the interval changes. I've tested the changes and confirmed that we can now adjust the simulation speed in real-time while the model is running.

mesa/visualization/solara_viz.py
AdamZh0u: Thank you @Corvince! I checked the SliderInt documentation and found that it accepts both `int` and `solara.Reactive[int]` values:
```python
SliderInt(
    value: Union[int, solara.Reactive[int]] = 0,
    on_value: Optional[Callable[[int], None]] = None,
):
```
Based on this, I've removed the .value for better reactivity and cleaner code.

quaquel:
Thanks a lot for this! I'll give @Corvince time to have another look but it seems ready to be merged.

Corvince:
Thanks again @AdamZh0u for this PR! Merged!

PR #2542, opened 2024-12-10
create agents added

peach280:
MoneyModel changed


quaquel:
The point of `create_agents` is to create _n_ agents in one go, and thus getting rid of the for loop for creating agents. This PR does not use this functionality at all. Please update the PR to properly reflect the power of `create_agents`.

quaquel:
closing because of a lack of follow up based on provided feedback.

PR #2543, opened 2024-12-10
Wolf-sheep to use `create_agent`

quaquel:
This updates the wolf sheep example to use `create_agents` for wolves and sheep. It also showcases how we can easily sample cells from a cell collection. 

I have deliberately not touched grass patched. It is possible to rewrite this to use create_agent, but to make this nice and clean requires a few more changes. Moreover, ideally, this should be rewritten to use a property layer for grass anyway.

### DIFF
--- a/mesa/examples/advanced/wolf_sheep/agents.py
+++ b/mesa/examples/advanced/wolf_sheep/agents.py
@@ -4,7 +4,9 @@ from mesa.experimental.cell_space import CellAgent, FixedAgent
 class Animal(CellAgent):
     """The base animal class."""
 
-    def __init__(self, model, energy=8, p_reproduce=0.04, energy_from_food=4, cell=None):
+    def __init__(
+        self, model, energy=8, p_reproduce=0.04, energy_from_food=4, cell=None
+    ):
         """Initialize an animal.
 
         Args:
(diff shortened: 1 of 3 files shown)


EwoutH:
Looks good! Could you use keyword arguments for clarity?

quaquel:
> Looks good! Could you use keyword arguments for clarity?

see the update and let me know which you prefer

EwoutH:
Thanks, I think this helps.

Do you know why the tests fail?

FAILED tests/test_examples.py::test_wolf_sheep - ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

quaquel:
> Do you know why the tests fail?

I have a hunch, but will be tonight most likely before I have time again. I'll get it fixed then

mesa/examples/advanced/wolf_sheep/agents.py
```diff
@@ -4,7 +4,7 @@
 class Animal(CellAgent):
     """The base animal class."""
 
-    def __init__(self, model, energy, p_reproduce, energy_from_food, cell):
+    def __init__(self, model, energy=0, p_reproduce=0.2, energy_from_food=1, cell=None):
```
EwoutH, line 7: What’s the motivation behind these default values?

quaquel: I need default values because you wanted to use keyword arguments on create_agents. These default values have no further meaning then that.

EwoutH (approved):
Thanks!

PR #2544, opened 2024-12-11
Shift sugarscape example to using create_agent

quaquel:
Shift sugarscape example to using create_agent. 

I want to update this example with property layers at some point, so I only moved the Trader agent over.


tpike3 (approved):
LGTM--- Create agent is very cool and makes sense about the property layer

PR #2545, opened 2024-12-12
Change pd_grid example to use create_agents

quaquel:
This modifies the prisoner dilemma example to use create_agents instead of a for loop.


tpike3 (approved):
LGTM

PR #2547, opened 2024-12-13
Add agent state management system with discrete & continuous states

EwoutH:
_This PR builds on discussion #2529._

### Summary
This PR adds a new state management framework to Mesa, enabling both discrete and continuous state management in agents. It provides a clean, minimal API for managing agent states that can change both discretely (at specific moments) and continuously (over time), as well as composed states that derive their values from other states.

### Motive
Agent-based models often need to track various agent states that change in different ways. While discrete state changes are straightforward to implement, continuous changes (like energy depletion, resource growth) are more challenging to model efficiently. Current approaches often require manual calculations each timestep or complex event scheduling.

This framework addresses several common challenges:
1. Efficient handling of continuously changing values
2. Clean separation of state logic from agent behavior
3. Support for derived states that depend on other states
4. Automatic time-based updates only when needed

### Implementation
The implementation adds four new classes:

1. `State`: Base class defining the state interface
2. `DiscreteState`: For states that change at specific moments
3. `ContinuousState`: For states that change continuously over time
4. `CompositeState`: For states derived from other states

And a mixin class:
- `StateAgent`: Extension of Mesa's Agent class with state management capabilities

Key implementation details:
- States are updated only when their values are accessed
- Continuous states use rate functions to calculate changes
- Composite states automatically update their dependencies
- Clean integration with Mesa's existing time management
- Type-safe state modifications

### Usage Examples
Here are the key API examples:

```python
# Basic API usage
agent.add_state(DiscreteState("mood", "happy"))                                   # Add discrete state
agent.add_state(ContinuousState("energy", 100, lambda current, elapsed: -0.1*elapsed))  # Add continuous state
agent.get_state("mood")                                                           # Get state value
agent.set_state("mood", "sad")                                                    # Set discrete state
agent.add_state(CompositeState("health", [agent.states["energy"]], lambda e: "healthy" if e > 50 else "weak"))  # Add composite state
agent.update_states()                                                             # Update all states to current time
```

Here are three small, complete examples showing different use cases:

```python
# Example 1: Simple Predator-Prey Agent
class Wolf(StateAgent):
    def __init__(self, model):
        super().__init__(model)
        # Energy depletes at 10% per time unit
        self.add_state(ContinuousState("energy", 100, lambda c, t: -0.1 * t))
        self.add_state(DiscreteState("status", "searching"))
        
    def step(self):
        self.update_states()
        if self.get_state("energy") < 30:
            self.set_state("status", "hunting")
```

```python
# Example 2: Growing Plant with Multiple States
class Plant(StateAgent):
    def __init__(self, model):
        super().__init__(model)
        # Height grows logistically
        def growth(current, elapsed):
            max_height = 100
            growth_rate = 0.1
            return growth_rate * current * (1 - current/max_height) * elapsed
            
        self.add_state(ContinuousState("height", 1, growth))
        self.add_state(ContinuousState("water", 100, lambda c, t: -0.2 * t))
        self.add_state(CompositeState("status", 
                                    [self.states["water"]],
                                    lambda w: "healthy" if w > 20 else "wilting"))
    
    def step(self):
        self.update_states()
        if self.get_state("water") < 20:
            # Plant would die or need watering
            pass
```

```python
# Example 3: Social Agent with Multiple Interacting States
class SocialAgent(StateAgent):
    def __init__(self, model):
        super().__init__(model)
        # Base happiness changes slowly
        self.add_state(ContinuousState("happiness", 50, lambda c, t: -0.05 * t))
        # Energy depletes faster when happiness is low
        def energy_change(current, elapsed):
            happiness = self.get_state("happiness")
            depletion_rate = -0.2 if happiness < 30 else -0.1
            return depletion_rate * elapsed
            
        self.add_state(ContinuousState("energy", 100, energy_change))
        # Overall mood depends on both happiness and energy
        def compute_mood(happiness, energy):
            if happiness > 70 and energy > 70:
                return "excellent"
            elif happiness > 40 and energy > 40:
                return "good"
            else:
                return "poor"
                
        self.add_state(CompositeState("mood", 
                                    [self.states["happiness"], 
                                     self.states["energy"]],
                                    compute_mood))
    
    def step(self):
        self.update_states()
        current_mood = self.get_state("mood")
        if current_mood == "poor":
            # Agent needs rest or social interaction
            pass
```

#### Future Enhancements:
1. Adding state history tracking
2. Supporting threshold events
3. Adding more specialized state types
4. Adding batch state updates
5. Supporting stochastic state changes

### Related Work
- This builds on ideas from discrete event simulation (#2066)
- Complements the existing time management system
- Could be further integrated with the spaces / PropertyLayer system in the future
- This fits in a bigger picture of Agents being able to make better decisions and more complex behavior based on their internal states. See:
  - https://github.com/projectmesa/mesa/discussions/2538
  - https://github.com/projectmesa/mesa/discussions/2526

#### To review
- This PR probably should integrate with https://github.com/projectmesa/mesa/pull/2291. @quaquel let me know what you think.
- The API is quite cumbersome. One thing I want to improve is using the direct access method for states, so you can do just `self.energy` in an Agent, like we recently added to the PropertyLayer Cell Space integration (see https://github.com/projectmesa/mesa/pull/2512#issuecomment-2498833465)

### DIFF
--- a/mesa/experimental/states.py
+++ b/mesa/experimental/states.py
@@ -1,0 +1,140 @@
+"""State management system for Mesa agents.
+
+This module provides a flexible state management system for Mesa agents, supporting
+both discrete and continuous state changes. It enables agents to maintain multiple
+states that can change either explicitly (discrete) or based on time (continuous),
+and includes support for composite states derived from other states.
+
+Core Classes:
+   State: Base class defining the state interface
+   DiscreteState: States with explicit value changes
+   ContinuousState: States that change over time
+   CompositeState: States computed from other states
+   StateAgent: Mesa Agent subclass with state management
+"""
+
+from collections.abc import Callable
+from typing import Any
+
+from mesa import Agent
+
+
+class State:
+    """Base class for all states."""
+
+    def __init__(self, name: str, initial_value: Any):
+        """Create a new state."""
+        self.name = name
+        self._value = initial_value
+        self._last_update_time = 0
+        self.model = None  # Set when state is added to agent
+
+    @property
+    def value(self) -> Any:
+        """Get current state value."""
+        raise NotImplementedError
+
+    def update(self, time: float) -> None:
+        """Update state to current time."""
+        raise NotImplementedError
+
+
+class DiscreteState(State):
+    """A state with discrete values that change explicitly."""
+
+    @property
+    def value(self) -> Any:
+        """Get the current state value."""
+        return self._value
+
+    @value.setter
+    def value(self, new_value: Any) -> None:
+        """Set the state value."""
+        self._value = new_value
+
+    def update(self, time: float) -> None:
+        """DiscreteStates only update when value is explicitly changed."""
+
+
+class ContinuousState(State):
+    """A state that changes continuously over time."""
+
+    def __init__(
+        self,
+        name: str,
+        initial_value: float,
+        rate_function: Callable[[float, float], float],
+    ):
+        """Create a new continuous state."""
+        super().__init__(name, initial_value)
+        self.rate_function = rate_function
+
+    @property
+    def value(self) -> float:
+        """Calculate and return current value based on elapsed time."""
+        current_time = self.model.time
+        if current_time > self._last_update_time:
+            self.update(current_time)
+        return self._value
+
+    def update(self, time: float) -> None:
+        """Update state value based on elapsed time."""
+        elapsed = time - self._last_update_time
+        self._value = self._value + self.rate_function(self._value, elapsed)
+        self._last_update_time = time
+
+
+class CompositeState(State):
+    """A state derived from other states."""
+
+    def __init__(
+        self,
+        name: str,
+        dependent_states: list[State],
+        computation_function: Callable[..., Any],
+    ):
+        """Create a new composite state."""
+        self.dependent_states = dependent_states
+        self.computation_function = computation_function
+        super().__init__(name, None)  # Value computed on first access
+
+    @property
+    def value(self) -> Any:
+        """Compute value based on dependent states."""
+        state_values = [state.value for state in self.dependent_states]
+        return self.computation_function(*state_values)
+
+    def update(self, time: float) -> None:
+        """Update all dependent states."""
+        for state in self.dependent_states:
+            state.update(time)
+
+
+class StateAgent(Agent):
+    """An agent with integrated state management."""
+
+    def __init__(self, model) -> None:
+        """Create a new agent with state management."""
+        super().__init__(model)
+        self.states = {}  # name -> State mapping
+
+    def add_state(self, state: State) -> None:
+        """Add a new state to the agent."""
+        state.model = self.model
+        self.states[state.name] = state
+
+    def get_state(self, name: str) -> Any:
+        """Get the current value of a state."""
+        return self.states[name].value
+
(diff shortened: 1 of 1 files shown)


EwoutH:
I updated the PR to:
- Allow directe attribute access (`agent.some_state`)
- Remove the now unneeded naming construct
- Added a check that states are up-to-date before fetching them

The API is now much cleaner!

Corvince:
I haven't seen the discussion and haven't looked at the implementation yet, but this reads fantastic! The API seems simple and powerful. Really looking forward to how this evolved, but congratulations already!

EwoutH:
Thanks! Only thing I'm a bit worried about is performance.

Don't make things states that don't need to be a state I guess.

EwoutH:
The next thing is that I would like to do is implement triggers for state changes:
- A `DiscreteState` can trigger on every change, or to/from a certain state
- A `ContinuousState` can trigger on reaching (below / above) a threshold value

These thresholds can be read (Boolean value) but could also emit an event to an event scheduler.

This would tightly integrate with potentially Mesa Observables and DEVS, so we need a bit of coordination on this.

EwoutH:
I also looked a bit into how it could integrate with Mesa Observables. I have a rough Proof of Concept here:
- https://github.com/EwoutH/mesa/commit/2f1a29f902d9611a07ba7b563649ec0cd7a0e63d

@quaquel I would love to hear your take on it.

quaquel:
There is a lot going on here. I need time to look at it closely and make up my mind. Given that is experimental, I believe we can push foward regardless. 

However, some quick reactions

1. Playing devil's advocate, state means attribute, so what does this really add?
2. Playing devil's advocate, explicit is better than implicit. Your Wolf example leaves a lot implicit (i.e., how energy is being updated automagically). Is all this implicitness of updating states desirable?
3. Regarding signals, this makes it even more complicated. I am not sure you can simply subclass Observable. Observable is a class-level attribute. Related, DerivedState is analogous to Computed. In fact, in javascript, a computed is typically explained as being a derived state, while a signal is considered a state. The complexity in case of MESA is that we want to combine this with objects/classes, while in javascript signals, computed, and effect are used within a functional programming paradigm. I like the idea of making this work with/on top of mesa signals. It would open up reactive ABMs (which currently don't exist at all to my knowledge).

EwoutH:
Thanks for thinking critically about it.

The main advantage is to decouple state updates from the main, fixed timestep, agent logic. Especially if you have a DEVS scheduler, you can now update states continuously, instead of them being literal step functions that get updated only on the whole time steps.

With the thresholds, it also allows events to fire exactly upon reaching a certain state.

It does add complexity and maybe makes some things more implicit/hidden. I would like to explore where this is worth it.

EwoutH:
@projectmesa/maintainers and others, I could use some additional eyes on this.

1. Is this feature useful?
2. Conceptually sound?
3. Implementation good?

And any other suggestions ideas. I mainly need to know if this is an idea worth pushing further, or not.

jackiekazil:
I am with @quaquel on all comments, but I want to +1 -- it is experimental - so I am less hesitant to stop an integration. 
My first concern was -- how this helps or hinders performance .... I saw the comment on this as well. This is something we should evaluate.

Overall though -- this is potentially very interesting!

quaquel:
see also #2851, closing this for now.

PR #2546, opened 2024-12-12
Switch sugarscape to using property layers

quaquel:
This changes sugarscape to use property layers for sugar and spice instead of a resource agent. 

I haven't yet updated the visualization side fully to take advantage of the property layer visualizations in MESA, I can do this in this PR or in a subsequent one, whichever is prefered.

PR #2548, opened 2024-12-13
bug fixes for draw_property_layers

quaquel:
While trying to update sugarscape to use `draw_property_layer`, I encountered a few bugs.
1. The integration into cell spaces includes its own version of `PropertyLayer`, so the `isinstance` check needs to account for this.
2. There was a transpose in the imshow that should have been removed as part of the larger matplotlib visualization overhaul because there `draw_property_layer ` shifted to using `origin="lower"`
3. In the sugarscape update (#2546) I forgot to remove some path stuff from app.py.

### DIFF
--- a/mesa/visualization/mpl_space_drawing.py
+++ b/mesa/visualization/mpl_space_drawing.py
@@ -181,8 +181,9 @@ def draw_property_layers(
         property_layers = space._mesa_property_layers
 
     for layer_name, portrayal in propertylayer_portrayal.items():
+        print(layer_name)
         layer = property_layers.get(layer_name, None)
-        if not isinstance(layer, PropertyLayer):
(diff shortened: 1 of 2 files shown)


mesa/visualization/mpl_space_drawing.py
```diff
@@ -181,8 +181,12 @@ def draw_property_layers(
         property_layers = space._mesa_property_layers
 
     for layer_name, portrayal in propertylayer_portrayal.items():
+        print(layer_name)
```
EwoutH, line 184: Should this line be added?

EwoutH (approved):

mesa/visualization/mpl_space_drawing.py
quaquel: no

Issue #2437, opened 2024-10-30
Minor refactoring of VoroinoiGrid


quaquel:
VoroinoiGrid contains a `cell_coloring_property`. This is visualization specific and does not belong in the grid class, but should be part of some kind of grid drawing/property layer portrayal of voroinoi grids. 

Moreover, visual inspection in #2340 suggests that something might be at least visually off about voroinoi grids becasue agents don't appear cleanly in the center of the cell.

PR #2549, opened 2024-12-14
Try to fix issue #2437

LunnyRia:
### Summary
This PR removes cell_coloring_property from VoronoiGrid and sets it as an argument when visualizing.

### Bug / Issue
Related to [Minor refactoring of VoroinoiGrid #2437](https://github.com/projectmesa/mesa/issues/2437).

### Implementation
In `voronoi.py`, there are three instances where the `cell_coloring_property` attribute is mentioned. All of these instances should be commented out (but not deleted). By examining `mpl_space_drawing.py`, I found that this attribute is used to set the transparency (`alpha`) of polygons. Based on the last line of `voronoi.py`, I understand that the developer intended to set its value to `0`. However, I don’t quite understand why it was defined as a `str`, as indicated by `cell_coloring_property: str | None = None`.

To enable setting the transparency of Voronoi polygons during visualization, I modified `mpl_space_drawing.py`. In the `draw_voronoi_grid` function, I added this attribute as a new parameter. GPT and Claude suggested more refined approaches for this visualization, but I want to first confirm that my modification doesn’t introduce any bugs and hear your opinions on it.

### Testing
Using the `.pre-commit-config.yaml` file from the root folder of the Mesa project, I ran `pre-commit run --all-files` locally, and all checks passed.

### Additional Notes
I just started learning Mesa and look forward to your suggestions. Thank you!


quaquel:
Thanks for taking a stab at this issue. 

1. Can you make the title a bit more descriptive?
2. Why don't you delete the old code?
3. This version of the PR removes functionality: i.e., we can no longer control the alpha of the individual voroinoi cells. How difficult would it be to modify `draw_voronoi_grid` to accept a color and/or alpha kwarg?

EwoutH:
Hi @LunnyRia! Thanks for the PR. Would you like to continue working on this? If so, is the feedback from quaquel helpful? And do you need anything from us?

quaquel:
superseded by #2557

Issue #2534, opened 2024-12-05
Cell space: Convenient method/practice to deny agents access (or remove cells)


EwoutH:
It would be nice if you could create inaccessible parts in the cell space, like having "walls" in your space. Maybe even completely being able to remove cells.

Neighbours e.d. should be updated.

quaquel:
Why would you want to create inaccessible parts? Can you give some examples of phenomena that you would like to model?

A critical design choice of the cell spaces is that the way they are wired up is static and cached. This is, in part, what makes them fast. Having dynamic cell spaces would break that design choice and result in a potentially substantial slowdown. 

Moreover, you can easily do something like this using property layers. Add a boolean layer indicating which parts of the space a given agent or agent class can access and use this in your custom movement logic. 

There is one use case for dynamic cell spaces, and that is for changing networks. There are quite a few open issues about dynamic networks, and neither the old-style network nor the new-style network supports this.

EwoutH:
> Why would you want to create inaccessible parts? Can you give some examples of phenomena that you would like to model?

Basically everything with pathfinding or segregation. Emergency workers finding people in an  (unknown) environment, a river dividing an ecoystem, a highway through a neighbourhood, etc.

Ideally you could also deny certain types / groups of agents access.

Not sure on implementation, haven't mapped the solution space yet. But a PropertyLayer could work. Removing cells might also be interesting, since it could even increase performance since there are both less cells and less neighbours to check. You could might be able to create vast sparse spaces this way.

quaquel:
> Basically everything with pathfinding or segregation. Emergency workers finding people in an (unknown) environment, a river dividing an ecoystem, a highway through a neighbourhood, etc.

I am not sure I would use a cell space for most of these examples. Moreover, the emergency worker example reflect the knowledge of the agents and should thus be handled via agent state not via the environment itself. 


>  Removing cells might also be interesting, since it could even increase performance since there are both less cells and less neighbours to check. You could might be able to create vast sparse spaces this way.

There is a difference between having a dynamic discrete space (i.e., the environment is dynamic) vs. the case where different agents have different understandings of the environment (and where this understanding might itself be evolving). The former is not possible with the current cell spaces. The later is not supported out of the box by MESA. I also doubt whether we should even support that out of the box because there is a 101 ways that users of mesa migth want to implement this.

EwoutH:
Slightly related:
- https://github.com/projectmesa/mesa/issues/2565

[...]

PR #2564, opened 2024-12-22
Extend Wolf Sheep example to include PropertyLayer

mgemaakbar:
This pull request is in response to @tpike3 's reply on Matrix discussion: https://matrix.to/#/!CAAdpmuJhdNJkxurrp:matrix.org/$sSOIip3pT1L5C6wrpzhvoM64Y4wYcpSsUHZSZXKwfFE?via=matrix.org&via=gitter.im

At the time this PR is created (Dec 22) the changes are still not that well tested. I would like to hear what you think first about my idea to extend this Wolf Sheep example, before testing it more and possibly adding more changes to it.

My idea of extending the Wolf Sheep is pretty simple:  use `PropertyLayer` to store information which cells that are cliff. If an animal (wolf or sheep) steps into a cliff cell, then the animal dies. The coordinate in `PropertyLayer` is set `true` if the cell is a cliff, `false` if it is not.
 
Please let me know what you think :) Thank you in advance.


EwoutH:
Thanks for your PR!

I just merged https://github.com/projectmesa/mesa/pull/2503, which also modified this example. You rebase your branch and resolve any conflicts? If you need any help with that let me know!

mgemaakbar:
> Thanks for your PR!
> 
> I just merged #2503, which also modified this example. You rebase your branch and resolve any conflicts? If you need any help with that let me know!

Yes I already did rebase and resolved the conflict.

EwoutH:
Thanks! like the idea of barriers / cliffs, it's widely discussed at the moment:
- https://github.com/projectmesa/mesa/issues/2534
- https://github.com/projectmesa/mesa/discussions/2556

I don't think the model dynamics will be nice if animals just walk like they currently do and suddenly die because a cell is marked as a cliff. Either, I would like the cliff to be inaccessible, so they don't even consider moving there, and choose another cells. This of course can mean they walk into a dead end (surrounded by cliffs), don't have any movement options anymore, and thus get eaten.

mgemaakbar:
> Thanks! like the idea of barriers / cliffs, it's widely discussed at the moment:
> 
> * [Cell space: Convenient method/practice to deny agents access (or remove cells) #2534](https://github.com/projectmesa/mesa/issues/2534)
> * [Wall Implementation and representation #2556](https://github.com/projectmesa/mesa/discussions/2556)
> 
> I don't think the model dynamics will be nice if animals just walk like they currently do and suddenly die because a cell is marked as a cliff. Either, I would like the cliff to be inaccessible, so they don't even consider moving there, and choose another cells. This of course can mean they walk into a dead end (surrounded by cliffs), don't have any movement options anymore, and thus get eaten.

Would you like me to implement that in this PR? It's actually the initial idea that came to my mind, but decided to make it 'stepping into a cliff = dies'.  

Also there might be a case where an animal (wolf or sheep) is randomly generated in the middle of 4 neighboring cliff (hence not being able to move at all forever).

Or, if you have another idea to include `PropertyLayer` in this Wolf Sheep example and you would like me to implement it, please feel free to tell me. This cliff idea is just an idea that I came up with.

EwoutH:
> Would you like me to implement that in this PR?

Yes

> Also there might be a case where an animal (wolf or sheep) is randomly generated in the middle of 4 neighboring cliff (hence not being able to move at all forever).

Was also thinking about this. Ideally cliffs are not generated fully random, but I'm not sure on a simple and elegant solution yet.

mgemaakbar:
> Either, I would like the cliff to be inaccessible, so they don't even consider moving there, and choose another cells. This of course can mean they walk into a dead end (surrounded by cliffs), don't have any movement options anymore, and thus get eaten.

I already updated the PR with this. But I changed the name cliff -> wall, since I think it fits better.

I have a question. In this code, we exit `move()` if the a sheep is surrounded by wolves. https://github.com/projectmesa/mesa/pull/2503/files#diff-a075828ed0ee3952964600d1f989ef36da3673f0a8e5a170ec124f6510581e8dR72

But in that case. the energy still gets decremented. https://github.com/projectmesa/mesa/pull/2503/files#diff-a075828ed0ee3952964600d1f989ef36da3673f0a8e5a170ec124f6510581e8dR42

Is it a bug, or is it an intended behavior?

EwoutH:
Okay, maybe take a step back here, before you invest too much in the current approach. Because:
1. I'm not 100% sure if walls add value to this implementation
2. If you purely want to demonstrate PropertyLayers, it might be more useful to add something like "soil fertility" (makes grass grow faster/slower) or elevation (takes more/less energy to move).
3. If we add walls, I think they should be added more deliberately then just "random noise". Maybe like you place ships in [Battleship](https://en.wikipedia.org/wiki/Battleship_(game)), or procedurally (learn something from Factorio: FFF [219](https://factorio.com/blog/post/fff-219) and [401](https://factorio.com/blog/post/fff-401)).

So, we have to take a step back, and determine your goal:
- Do you want to demonstrate the PropertyLayer (in a simple and elegant way)?
- Do you want to demonstrate how to make cells inaccessible for Agents ([#2534](https://github.com/projectmesa/mesa/issues/2534))?

The approach we often take when things 

1. Get the problem sharply and neatly defined
2. Propose a solution
3. Discuss and agree on a conceptual approach for the solution
4. Explore the implementation for the solution (API, data structures, whatever is needed)
5. Implement it

The first 4 can best happen in a [Discussion](https://github.com/projectmesa/mesa/discussions). If you have those first 4 very clearly defined, 5 becomes often trivial (or at least straightforward).

mgemaakbar:
> Do you want to demonstrate the PropertyLayer (in a simple and elegant way)?
Yes, I think my goal is this.

> If you purely want to demonstrate PropertyLayers, it might be more useful to add something like "soil fertility" (makes grass grow faster/slower) or elevation (takes more/less energy to move).

This is a good idea, should I try this?

> The first 4 can best happen in a [Discussion](https://github.com/projectmesa/mesa/discussions). If you have those first 4 very clearly defined, 5 becomes often trivial (or at least straightforward).

Should I create a discussion about the "soil fertility" idea there?

Thank you for your feedback

quaquel:
Having a separate discussion on soil fertility sounds like a good idea. However, with sugarscape, we already have a simple property layer example. So, at a minimum, this example should add value and be different from what is show cased there.

EwoutH:
> Should I create a discussion about the "soil fertility" idea there?

Yes, let's do this.

> However, with sugarscape, we already have a simple property layer example. So, at a minimum, this example should add value and be different from what is show cased there.

It could demonstrate how two PropertyLayers interact: The soil layer is static (stays the same during the run), and the grass layers is updated based on the soil layer.

mgemaakbar:
Hi, sorry for the late response. I made the discussion here https://github.com/projectmesa/mesa/discussions/2578

EwoutH:
What's the current state of this PR? Do you need help or input on anything?

mgemaakbar:
Hi, sorry for late response. I have been busy with other university course. I have replied to your latest [reply](https://github.com/projectmesa/mesa/discussions/2578#discussioncomment-11728167) on the [discussion](https://github.com/projectmesa/mesa/discussions/2578)

I have a question. I saw you mentioned this PR on this [issue](https://github.com/projectmesa/mesa/issues/2534). Would you like me to keep this PR so I can contribute to this https://github.com/projectmesa/mesa/issues/2534 issue too? 

I can create separate PR to implement the PropertyLayer demonstration on Wolf Sheep as we discuss [here].(https://github.com/projectmesa/mesa/discussions/2578)

mgemaakbar:
I already updated this PR according to the discussion in https://github.com/projectmesa/mesa/discussions/2578. If you want to know how the spatial distribution is visualized, use this code

```
import numpy as np
from scipy.ndimage import gaussian_filter

def generate_grass_regrowth_time_array():
  rows, cols = 100, 100

  seeds = np.zeros((rows, cols))
  max_grass_regrowth_time = 15  

  # Generate seed points with random values
  for _ in range(max_grass_regrowth_time):
      x, y = np.random.randint(0, rows), np.random.randint(0, cols)
      seeds[x, y] = np.random.randint(1, max_grass_regrowth_time)

  # Smooth the array to create clustered patterns using SciPy's Gaussian filter
  filtered_array = gaussian_filter(seeds, sigma=10)

  # Normalize the array to the range [1, max_grass_regrowth_time]
  filtered_array = (filtered_array - np.min(filtered_array)) / (np.max(filtered_array) - np.min(filtered_array)) * (max_grass_regrowth_time - 1) + 1
  filtered_array = filtered_array.astype(int)

  return filtered_array

# # Visualize the spatial distribution
plt.figure(figsize=(8, 8))
plt.imshow(generate_grass_regrowth_time_array(), cmap='YlGn', origin='upper')
plt.colorbar(label='Time for Grass Growth (days)')
plt.title('Spatial Distribution of Grass Growth Time (Normalized to [1, max_grass_regrowth_time])')
plt.xlabel('X Coordinate')
plt.ylabel('Y Coordinate')
plt.show()
```

quaquel:
closed as stale

EwoutH:
I spend an non-trivial amount of time looking for comments like https://github.com/mesa/mesa/pull/2564#issuecomment-2562921839 before writing the Mesa development process guidelines in #3135.

In the end I wrote them from scratch, and of course, now I find another.

Issue #1572, opened 2022-12-30
Add prose linter to CI


rht:
**What's the problem this feature will solve?**

We currently use codespell, but we could have a more uniform documentation style with a prose linter.

**Describe the solution you'd like**

One possible solution is using [Vale](https://github.com/errata-ai/vale). But there are other [alternatives](https://github.com/errata-ai/vale#functionality) as well.

**Additional context**
See https://github.com/projectmesa/mesa-examples/pull/14#issuecomment-1364454346

rht:
@EwoutH any opinion on those prose linter tools?

manvendra5345:
Hi everyone!
I am Manvendra Singh Solanki from Cse Ai-Ml background and i am also interested in working in this project , I know python , java , HTML , CSS  . I can adapt very fast and have great learning capacity but these thing are common what i think i have different in me is that i want to learn from thoes who are better than me and catch up  to them and even beat them and always keep learning.

I also want to contribute and learn but i am new to this platform please guide me .

Email:- singhmanvendra2003@gmail.com

EwoutH:
> @EwoutH any opinion on those prose linter tools?

Don't know it, pitch it to me :)

fammu03:
Hey is this issue still open? I would like to try and work on it.

rht:
Yes, you can work on it. Vale is the one I have found to be used by multiple projects.

sanika-n:
Hi, Sanika here, I just created a pull request which hopefully addresses this issue.. I integrated vale into the CLI workflow and configured so that vale gets triggered anytime anyone pushes code or submits a  PR, this is my first time contributing  in mesa, so pls let me know if it is alright:)

EwoutH:
Thanks for the PR. I looked into it a bit, and I think a prose linter seems useful, and Vale looks like it's the most commonly used one and actively developed. So I support adding it, if it doesn't add too much overhead to our processes.

So go ahead, I left some more technical feedback in the PR.

sanika-n:
Thank you, will try to resolve them as soon as possible

Edu92337:
Hey @sanika-n , @EwoutH , I would like to contribute to this issue. Could you assign it to me?

Sahil-Chhoker:
@Edu92337, you can look at the implementation and discussions in PR #2568 before working on the issue.

pragam-m25:
[...]

PR #2568, opened 2024-12-24
added vale to CLI

sanika-n:
### Summary
I added Vale, a prose linter to CLI as requested in "Add prose linter to CI #1572 " issue. So, each time a user pushes/ creates a PR, the CLI automatically runs, and its suggestions and warnings can be viewed. Currently, I have configured it to only check on .md files, but that can be changed according to use case by altering the " .vale.ini "file.

### Screenshots
Currently after running on the existing repo, vale made a lot of recommendation and suggestions. I intentionally created a "bad" .md file titled TESTVALE.md just to test it, the results are attached below:
image
image
It gave some errors and warnings on the existing files as well. I have attached a log to that below.
[4_Run Vale.txt](https://github.com/user-attachments/files/18241496/4_Run.Vale.txt)


EwoutH:
Thanks for the PR!

It's currently over a thousands lines of code spread over 45 files, which is quite challenging to review. I also prefer not to add such amounts of code to our codebase. Are all the files critical for this functionality to function? Anything you might be able to cleanup?

.github/workflows/vale.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  vale:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+
+      - name: Set up Vale
+        run: |
+          curl -sSL https://github.com/errata-ai/vale/releases/download/v2.12.0/vale_2.12.0_Linux_64-bit.tar.gz | tar -xz -C /tmp
```
EwoutH, line 20: It seems the latest version is [v3.9.1](https://github.com/errata-ai/vale/releases/tag/v3.9.1), is there a specific reason you're using [v2.12.0](https://github.com/errata-ai/vale/releases/tag/v2.12.0)?

sanika-n:
@EwoutH, I updated the version of vale and also suppressed vale to allow the use of "agent". I had actually cloned code from this 
repo to -> https://github.com/errata-ai/Microsoft/tree/master/Microsoft   to serve as the style for vale and hence the large amount of files, I tried to find a work around, but the only option seems to be to use it as a submodule...
I did integrate vale to pre-commit and it does fail as not all md files are following vale's rules.
image


However, the reason for failure in the checks above, (pre-commit.ci -pr) says that it failed because it wasn't able to find vale, 
I am not able to figure out why that is happening
image

EwoutH:
Maybe it’s easier to start with a minimal implementation, using only pre-commit, and then extend from there.

sanika-n:
@EwoutH, Hi, I think I figured it out it is working now:)
image

quaquel:
With 45 files changed, this PR has become impossible to review. I see a lot of value in adding this, however, would it be possible to have only this one change? Next, in follow up pr's we can go throught the code base to fix all issues that are flagged after vale has been activated.

sanika-n:
Sure, so from what I understand you would like me to cut down on the number of files right? I think one thing that I could is change the style from Microsoft to Write-Good (https://github.com/errata-ai/write-good/tree/master/write-good) this only has 9 yml files...Would that be fine?

quaquel:
My bad, I had not carefully checked the PR. It is not 45 files _changed_, but 1 file changed (pre-commit) and 44 new files required for vale. Can you briefly explain the options and why you went with the Microsoft style?

sanika-n:
Sure, so officially these are the styles that vale recognizes:
image
(https://github.com/errata-ai/packages)

I chose Microsoft because it has a wider range of error catching abilities and can focus on a lot more writing elements compared to some other options. Write-good is alright too but it is more basic and won't be able to catch errors that Microsoft can.

quaquel:
Finally had time to look closer at Vale. It seems [they suggest](https://github.com/errata-ai/vale-action) to put the `styles` and `.vale` folders inside the `.github` folder. Can you ensure the layout follows their suggestions? Once that is done, I think this is good to go.

sanika-n:
Done :thumbsup:

.pre-commit-config.yaml
```diff
@@ -1,34 +1,47 @@
...
-    # Run the linter.
-    - id: ruff
-      types_or: [ python, pyi, jupyter ]
-      args: [ --fix ]
-    # Run the formatter.
-    - id: ruff-format
-      types_or: [ python, pyi, jupyter ]
--   repo: https://github.com/asottile/pyupgrade
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.8.1
```
quaquel, line 6:
```suggestion
    # Ruff version.
    rev: v0.8.1
```

quaquel:
I think this looks good, but it would be good if one of the other maintainers also takes a look.

EwoutH:
Thanks for working on it.

When I review an PR one of the first things I do is judge the costs-benefit ratio of the PR. Each new feature adds complexity, is something that can break, and is something we need to update and maintain.

With this PR, I'm not yet fully convinced that ratio/proportionality is there.
- Benefits: We already have Codespell for spelling/language, a code linter (ruff), a docstring linter (ruff-format) and a few smaller checkers for things like whitespaces. That seems to cover most stuff, I don't see the distinct value of Vale for us.
- Costs: This PR adds a huge fileset that probably needs to be updated periodically. Can that be automated? A git submodule might be possible, but since it's the first, we need to update that.

So, while neither on their own are a problem, I currently don't see the ratio here. Either (personally) I need to be convinced of the benefits a prose linter adds, or the complexity needs to come way down (thing of one or at most two files that can be updated automatically periodically). Can we use the Microsoft or Google standard in some way without copying all their files to our repo?

sanika-n:
I completely understand, I did try to find alternatives to copying their repo, but I couldn't find much results... will surely update if I find a workaround

EwoutH:
I'm closing this PR as stale, but feel free to open a new one (or continue the discussion) if new insights pop up!

PR #2528, opened 2024-12-03
Fixes links for mesa examples in ReadMe.

Sahil-Chhoker:
## Summary
Fixed link references for examples in Readme

## Bug / Issue
Closes #2527

PR #2570, opened 2024-12-27
Fix links in example README.md

jackiekazil:
Fix broken links.

### DIFF
--- a/mesa/examples/README.md
+++ b/mesa/examples/README.md
@@ -6,32 +6,32 @@ More user examples and showcases can be found in the [mesa-examples](https://git
 ## Basic Examples
 The basic examples are relatively simple and only use stable Mesa features. They are good starting points for learning how to use Mesa.
 
-### [Boltzmann Wealth Model](examples/basic/boltzmann_wealth_model)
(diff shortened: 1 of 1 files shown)


quaquel:
This breaks the links on read the docs. See #2448 and #2528. 

I don't know of a way to use the readme when generating the [read the docs examples landing page](https://mesa--2570.org.readthedocs.build/2570/examples.html) and have links work both within github and within readthedocs.

EwoutH:
This issue keeps coming back, so we should try to fix it somehow.

One thing that might work is move the Readme, for example by breaking the Readme into a Basic and Advanced section, so the relative links would be the same for GitHub and RTD.

quaquel:
> This issue keeps coming back, so we should try to fix it somehow.

I agree

> One thing that might work is move the Readme, for example by breaking the Readme into a Basic and Advanced section, so the relative links would be the same for GitHub and RTD.

That is not the problem. The issue is that readthedocs requires `examples/basic/ect`, while github uses `basic/ect`. So, one option is to just splice it the examples part when generating the readthedocs file. This could be done in docs/conf.py and then `setup_examples_pages`.

jackiekazil:
Shall I close this?

PR #2575, opened 2024-12-28
Meta agents

tpike3:
### Summary
This PR is useful for creating meta-agents that represent groups of agents with interdependent characteristics.

New meta-agent classes are created dynamically using the provided name, attributes and functions of sub agents, and unique attributes and functions.

supersedes #2561 
  
### Motive

This method is for dynamically creating new agents (meta-agents).

Meta-agents are defined as agents composed of existing agents.

Meta-agents are created dynamically with a pointer to the model, name of the meta-agent,
iterable of agents to belong to the new meta-agents, any new functions for the meta-agent,
any new attributes for the meta-agent, whether to retain sub-agent functions,
whether to retain sub-agent attributes.

Examples of meta-agents:
- An autonomous car where the subagents are the wheels, sensors,
battery, computer etc. and the meta-agent is the car itself.
- A company where the subagents are employees, departments, buildings, etc.
- A city where the subagents are people, buildings, streets, etc.

Currently meta-agents are restricted to one parent agent for each subagent/
one meta-agent per subagent.

Goal is to assess usage and expand functionality.
### Implementation

Method has three paths of execution:

1. Add agents to existing metaagent
2. Create new meta-agent instance of existing metaagent class
3. Create new meta-agent class

Added `meta_agents.py` in experimental
Added tests in test-agent.py
Added alliance formation model in basic examples

### Usage Examples

I added a basic example of alliance formation using the bilateral shapley value 

Step 0- 50 Agents: 
image

Step 8 - 17 Agents of increasing hierarchy added dynamically during code execution:
image

### Additional Notes
 Currently restricted to one parent agent and one meta-agent per agent. Goal is to assess usage and expand functionality.


EwoutH:
Did you consider incorporating the `AgentSet` into Meta-agents one way or another (inheritance, composition or otherwise)?

tpike3:
> Did you consider incorporating the `AgentSet` into Meta-agents one way or another (inheritance, composition or otherwise)?

I did, however just to get meta-agents started and integrated I avoided it since it will add a layer of complications and I was worried about collisions and MRO issues -- Ways forward in no particular order are add AgentSet so each meta-agent has the AgentSet functionality, integrate threading so meta-agents can be on their own thread in 3.13 forward, allow for greater combinatorics (e.g. agents can be in multiple meta-agents)

EwoutH:
Cool, I will try to dive in and do a proper review tomorrow or Monday.

jackiekazil:
I am playing catch-up on this. This is interesting.  I do not have strong opinions on the setup, and I feel like @EwoutH and @quaquel have it covered. 

Question for everyone regarding adding an example in the core examples folder: If memory serves me correctly -- we were adamant about only having five examples with ones that build off of those for maintenance purposes. In what cases do we justify an additional example to the core folder as opposed to the examples folder? (I ask because our original examples folder grew to the maintenance issue that it is because we were trying to demonstrate all the functionality we had in at least one location).

tpike3:
> I am playing catch-up on this. This is interesting. I do not have strong opinions on the setup, and I feel like @EwoutH and @quaquel have it covered.
> 
> Question for everyone regarding adding an example in the core examples folder: If memory serves me correctly -- we were adamant about only having five examples with ones that build off of those for maintenance purposes. In what cases do we justify an additional example to the core folder as opposed to the examples folder? (I ask because our original examples folder grew to the maintenance issue that it is because we were trying to demonstrate all the functionality we had in at least one location).

Well clearly I am going to have a bias on this, however, I did consider challenge of an excess of core examples and the reason I built in basic examples as an exception is because as far I know this would be a unique capability of Mesa compared to NetLogo, MASON etc. Putting it in the basic examples would make it more prominent and easier for users to access, ideally further increasing Mesa's competitiveness as the ABM library of choice. 

A second option would be putting it in mesa-examples and then adding some documentation in getting started. 

Or I could just put it in Mesa-examples.

I am good with whatever the group decides, just let me know.

EwoutH:
Let's focus on the functionality first and then work out the details on the examples.

quaquel:
For various experimental features, I added some examples into the folder with the new experimental code. Once the API started to stabilize, I removed those examples. That might be a good middle ground here as well. The example helps clarify the experimental feature.

EwoutH:
I have some initial thoughts, mainly on the conceptual and API design.

Our `AgentSet` provides already a lot of the capabilities needed by MetaAgents, so I would think as a MetaAgents as an AgentSet that can also perform things on their own. This would allow us to leverage existing functionality while making the API more consistent and intuitive.

Here's how we could redesign this:

```python
class MetaAgent(mesa.Agent):
    """An agent that is composed of other agents and can act as a unit."""
    
    def __init__(self, model, components=None):
        super().__init__(model)
        # Use AgentSet to manage components - inheriting all its powerful features
        self._components = AgentSet(components or [], random=model.random)
        
    @property
    def components(self):
        """Read-only access to components as an AgentSet."""
        return self._components

    # Component management
    def add_component(self, agent):
        self._components.add(agent)
        
    def remove_component(self, agent):
        self._components.discard(agent)
```

This gives us several benefits:
1. All AgentSet methods (`select`, `sort`, `do`, `map`, `agg`, etc.) are available on the components
2. Consistent with how Mesa already manages agents
3. Automatic handling of weak references and agent lifecycle
4. Easy aggregation of component properties

For example, here's how your alliance formation could look using this approach:

```python
class Alliance(MetaAgent):
    def __init__(self, model, components=None):
        super().__init__(model, components)
        # Compute alliance properties from components using AgentSet methods
        self.power = self.components.agg("power", sum) * 1.1  # With synergy bonus
        self.position = self.components.agg("position", np.mean)
        
    @classmethod
    def evaluate_potential(cls, agents: AgentSet) -> float:
        """Evaluate if these agents would make a good alliance using AgentSet methods."""
        total_power = agents.agg("power", sum)
        positions = agents.get("position")
        position_spread = max(positions) - min(positions)
        return total_power * (1 - position_spread)
```

We could even add some convenience methods to AgentSet to support meta-agent operations:

```python
class AgentSet:
    def to_meta_agent(self, meta_agent_class, **kwargs):
        """Convert this AgentSet into a MetaAgent of the specified class."""
        return meta_agent_class(self.model, components=self, **kwargs)
    
    def find_combinations(self, size_range=(2,5), evaluation_func=None, min_value=None):
        """Find valuable combinations of agents in this set."""
        combinations = []
        for size in range(*size_range):
            for candidate_group in itertools.combinations(self, size):
                group_set = AgentSet(candidate_group, random=self.random)
                if evaluation_func:
                    value = evaluation_func(group_set)
                    if min_value is None or value >= min_value:
                        combinations.append((group_set, value))
        return combinations
```

Then alliance formation becomes very natural:

```python
def step(self):
    # Find potential alliances among base agents
    base_agents = self.agents.select(agent_type=BaseAgent)
    potential_alliances = base_agents.find_combinations(
        evaluation_func=Alliance.evaluate_potential,
        min_value=10
    )
    
    # Form some alliances
    for agents, value in potential_alliances:
        if self.random.random() < 0.3:  # Some formation probability
            alliance = agents.to_meta_agent(Alliance)
```

Some open questions to consider:
1. Should we allow agents to be part of multiple meta-agents? The current PR restricts this but it might be unnecessarily limiting.
2. How should we handle property updates when components change? Should meta-agents recompute their properties automatically?
3. Should we add specific support in the Model class for tracking meta-agents separately from base agents?

I think this approach would make the feature more intuitive and maintainable while preserving all the functionality of the current implementation. It would also make it easier to extend with new capabilities in the future.

What do you think about moving in this direction? Happy to elaborate on any of these points.

quaquel:
Just some quick thoughts:

1. I agree with @EwoutH  suggestion to use composition. So let `MetaAgent` extend `Agent`, and let it have an AgentSet attribute. Not sure about the `components` label. Since it is a MetaAgent, something like `constituting_agents` would be better (but is too long). `components` to me is to vague. Likewise, I would simplify `add_component` to `add_agent` etc. Also, you can just do `self.random` instead of `model.random`. 
2. I don't think the default MetaAgent class should make a choice on how to handle property updates. This is very case-specific. With the sketched functionality, you can do whatever you want in your own custom subclass.
3. There is no need for extra support in `Model`, since you can just use agents_by_type to get the MetaAgent instances.
4. I would not add a method to AgentSet to create MetaAgents. First, MetaAgent is still experimental. Second, in my view, it violates the separation of concern. If you want something like that, it might make more sense to a factory method to MetaAgent, so you can do `MetaAgent.from_agentset`.

tpike3:
Thanks @EwoutH and @quaquel I appreciate the time you are spending on this, which came out unexpectedly. 

So for expanding the MetaAgent to have AgentSet, that works for me. On the attribute name maybe `nodes` or `subagents`, but I am not sure I like any of those. I have no disagreements with any of @quaquel other concerns. 

**For practical way forward:** 
@EwoutH I believe code is sharing knowledge and ideas so feel free to edit my PR as you see fit or improve it in a new better PR. My only concern is how are you passing new meta agent functions or making a sub agent function now the meta-agent function? 

There is some more nuance to the bilateral shapley value where you do not want to do mass aggregation and strictly speaking I am not doing it exactly right, but that is off topic. However, it leads to the larger conceptualization of meta-agents, which you can read or not at your leisure.  

**For less practical off the top of my head considerations**
How do we deal with the diversity of use cases: 
- One use would be similar to the alliance formation, where user wants the meta-agent to have the alliance formation function on its own so it can use its meta attributes to assess formation with other metaagents, which to keep it clean means it needs its own version of the function. 
- Another path would be like the smart car example where different parts, at the desired level of granularity, would have their attributes and functions, FLIR sensor, turning mechanism, battery etc but the car metaagent would have functions to manage and control the subcomponents. 
- Stretch goal would be at some point we would add the functionality for the subcomponents to create emergent functions from the constituent parts, thinking with RL and generative AI right now

Let me know you thoughts.

EwoutH:
@tpike3 I created an initial implementation on the [`meta_agents_2`](https://github.com/EwoutH/mesa/tree/meta_agents_2) branch in https://github.com/EwoutH/mesa/commit/377fb3db7b2c84a4e5db6c95f7eb99f6ee99bba3. If you want, you can continue from there (or use part of it, whatever works for you).

tpike3:
> ```python
> @classmethod
> ```

Thanks @EwoutH! -- merging the code right now and yaaa I should have totally integrated AgentSet from the beginning

tpike3:
*Status Update*

So I realized we are looking at the problem in different not necessarily mutually exclusive ways. So what I am working on is this---

From @EwoutH a base MetaAgent optimizes use of AgentSet

From me the ability to dynamically create multiple agent types

Things to do --

- [ ] update syntax to dynamically create agent types
- [ ] update tests
- [ ] add single level alliance formation model

Let me more if you have more thoughts

quaquel:
I finally had time to take another look at this. Based on #2538, my understanding is that meta agents are composed of agents and have their own behavior and state which might be based on the behavior and state of its constituting agents. I think the ideas here are very interesting, but also hard to get right (which is why I don't know of any other ABM library that has something similar).

I am, however, a bit confused, about `create_multi_levels`. 

First, `add_attributes` wil just add the value of the last agent in `agents`. See the code below. If you have multiple agents of the same type, `meta_attributes[name] = value` will just refer to the value of the last agent in `agent`, so what's the point of this?

```python
if retain_subagent_attributes:
    for agent in agents:
        for name, value in agent.__dict__.items():
            if not callable(value):
                meta_attributes[name] = value

for key, value in meta_attributes.items():
    setattr(meta_agent_instance, key, value)
```

Second, `add_functions`. As an aside, this inner function is misnamed, and should be `add_methods`. Basically, this dynamically adds all methods from the constituting agents to the MetaAgent. Again, I am not sure about the conceptual motivation for this. Take the alliance example, the alliance has behavior that is uniquely its own but is collective over the alliance members. Second, it might trigger specific behavior in the alliance members. Now, some behavior might indeed be shared. _E.g._, individual agents can form into an alliance and alliances can form into larger alliances. Although, I would argue that the way in which an individual agent agrees to join an alliance is different from how an existing alliance agrees to join into a larger alliance. In the first case, the individual can decide itself, in the second case either all individual members decide themselves, or there is some voting procedure. Regardless, the logic of the behavior is not the same for individuals and for groups.

tpike3:
> I finally had time to take another look at this. Based on #2538, my understanding is that meta agents are composed of agents and have their own behavior and state which might be based on the behavior and state of its constituting agents. I think the ideas here are very interesting, but also hard to get right (which is why I don't know of any other ABM library that has something similar).
> 
> I am, however, a bit confused, about `create_multi_levels`.
> 
> First, `add_attributes` wil just add the value of the last agent in `agents`. See the code below. If you have multiple agents of the same type, `meta_attributes[name] = value` will just refer to the value of the last agent in `agent`, so what's the point of this?
> 
> ```python
> if retain_subagent_attributes:
>     for agent in agents:
>         for name, value in agent.__dict__.items():
>             if not callable(value):
>                 meta_attributes[name] = value
> 
> for key, value in meta_attributes.items():
>     setattr(meta_agent_instance, key, value)
> ```
> 
> Second, `add_functions`. As an aside, this inner function is misnamed, and should be `add_methods`. Basically, this dynamically adds all methods from the constituting agents to the MetaAgent. Again, I am not sure about the conceptual motivation for this. Take the alliance example, the alliance has behavior that is uniquely its own but is collective over the alliance members. Second, it might trigger specific behavior in the alliance members. Now, some behavior might indeed be shared. _E.g._, individual agents can form into an alliance and alliances can form into larger alliances. Although, I would argue that the way in which an individual agent agrees to join an alliance is different from how an existing alliance agrees to join into a larger alliance. In the first case, the individual can decide itself, in the second case either all individual members decide themselves, or there is some voting procedure. Regardless, the logic of the behavior is not the same for individuals and for groups.

@quaquel Sorry for the delay, I was visiting some family over the holiday 

However, I made a bunch of updates to ideally allow more user options. Below I walk through each one, and I think that address your points you outline excepted for function to method, which I will change probably tomorrow since we are getting weather. 

First - Based on @EwoutH comments I created `MetaAgents` as stand alone, this allows users to explicitly create meta_agents of the desired type. 

Example use case I am thinking is an autonomous systems in a warehouse. So it has parts like wheels, FLIR sensors, computer etc, now it can have a meta agent that calls the unique attributes and methods of each of its subagents.  

Second - I added multi-level agents as a stand alone and separate function. This is like the alliance formation and dynamically adds new agent classes MetaAgents at increasing hierarchies.  

I also added the use of `itertools.combinations` and associated functions, but I still locked down the multi-level agents just because of the exponential growth of combinatorics. Going above into higher numbers of combinations (e.g. combinations of 10 vs 2), quickly overwhelms the model 

Example use case is the provided example which is a more pure instantiation of the bilateral shapely value, but to your point an individual agent may now join a level 3 agent as it position changed.  This, however, would be dependent on the example model but the multi_level agent code does allow for it by changing the logical of the multi-level alliance formation model.

Finally multi-level agents does allow for adding the sub agents attributes and methods. The thought behind this is unlike the first example that is explicit, this can allow for recombination and emergent agents. 

Example use case and just to keep with with alliance: I once did a tribal affiliation model of the Libyan civil war. In this case some tribes may be coming with control of road networks and certain attributes and methods for dealing with that, while another tribe may come with control of oil fields and likewise have unique functions and attributes, this would allow for a singular agent who has leaders managing both assets as well as subgroups managing their respective territory. I concede this is an arguable feature. 

The hard question to me, which ideally by dabbling in this unique and difficult landscape will help inspire the community to explore, is how one can assess how well the groups have coalesced together and what is their chances for descending into infighting and how do you manage that computationally. In one sense this expands experimental to not only do great Mesa improvements (like agentset, propertylayer, continuous space etc) but also to try some novel and questionable features to ideally get a larger community to explore the space and push the whole ABM field forward. 


Let me know what you think. Thanks again for the review!

  

In the end I really only did an example 2 but in future pull requests could add examples for 1 and 3. 

Let me know what you think.

tpike3:
@quaquel and @EwoutH ready for review

EwoutH:
> Sorry for the delay, I was visiting some family over the holiday

Let’s stop apologizing for this stuff, we’re all doing this in our free time. Thanks for working on it!

I will try to review the ContiniousSpace soon, and then the PR hopefully later this week.

Exciting to have so much serious work going on!

EwoutH:
Bit busy this week with job applications, I will try to review Thursday or Friday. If I haven't reviewed by next weekend, also please remind me!

EwoutH:
@tpike3 I can do a call now

EwoutH:
I thought about this a bit more.

The current function-based approach, while flexible, has significant drawbacks. It makes the code harder to understand and maintain, provides limited IDE support and type hints, and makes error handling more complex. Most importantly, by focusing on dynamic creation through functions, we lose the natural object-oriented patterns that make Mesa intuitive to use. The motivation for an inheritance-based approach is to provide users with a clear, type-safe interface that follows Mesa's existing patterns while still maintaining the flexibility needed for dynamic agent creation.

Instead of having separate implementations, we could have a single `MetaAgent` base class that includes factory methods for dynamic creation. This would provide:

1. Clear base functionality through inheritance (addressing IDE support and type safety concerns)
2. Dynamic creation capabilities through class methods (maintaining the flexibility needed for alliance formation)
3. Better encapsulation of the meta-agent logic

For example:

```python
class MetaAgent(mesa.Agent):
    def __init__(self, model, components=None):
        super().__init__(model)
        self._components = mesa.AgentSet(components or [], random=model.random)
        
    @classmethod
    def create(cls, model, name, components, attributes=None, methods=None):
        """Factory method for dynamic meta-agent creation"""
```

This addresses your examples:
- Autonomous systems: Create specialized classes (Robot, Car) inheriting from MetaAgent
- Alliance formation: Use the factory method to dynamically create new alliance types
- Tribal affiliations: Combine both approaches - base classes for known structures, dynamic creation for emergent groups

```Python
# Case 1: Autonomous Systems - Direct inheritance
class Robot(MetaAgent):
    def __init__(self, model, sensors=None, wheels=None):
        components = (sensors or []) + (wheels or [])
        super().__init__(model, components)
        
    def move(self):
        wheels = self.components.select(type=Wheel)
        return all(wheel.functional for wheel in wheels)

# Case 2: Alliance Formation - Dynamic creation 
class Alliance(MetaAgent):
    @classmethod
    def form_alliance(cls, model, agents, level):
        def evaluate_merger(self, other):
            """Evaluate potential merger between alliances of the same level."""
            if self.level == other.level:
                return self.power * other.power
            return 0
            
        return cls.create(
            model=model,
            name=f"Level{level}Alliance",
            components=agents,
            attributes={"level": level},
            methods={"evaluate_merger": evaluate_merger}
        )

# Case 3: Tribal Affiliations - Hybrid approach
class Tribe(MetaAgent):
    def __init__(self, model, territory, components=None):
        super().__init__(model, components)
        self.territory = territory

    @classmethod
    def merge_tribes(cls, model, tribes):
        """Create a new meta-tribe from existing tribes."""
        def manage_territory(self):
            """Calculate territory management efficiency."""
            return self.territory * len(self.components)
            
        territory = sum(tribe.territory for tribe in tribes)
        
        return cls.create(
            model=model,
            name="MergedTribe",
            components=tribes,
            attributes={"territory": territory},
            methods={"manage_territory": manage_territory}
        )
```

Curious what you (and @quaquel and @Corvince) think about this approach!

EwoutH:
Maybe this is actually less intuitive... let's discuss in the call!

EwoutH:
Usage examples for both:

1. Simple Component Grouping

```python
# Original Approach
components = [CPU(model), RAM(model), GPU(model)]
computer = create_meta_agent(
    model=model,
    name="Computer",
    components=components,
    attributes={"power_state": "off"},
    methods={"power_on": lambda self: setattr(self, "power_state", "on")}
)

# Class Approach
class Computer(MetaAgent):
    def __init__(self, model, components):
        super().__init__(model, components)
        self.power_state = "off"
    
    def power_on(self):
        self.power_state = "on"

computer = Computer(model, [CPU(model), RAM(model), GPU(model)])
```

2. Dynamic Alliance Formation

```python
# Original Approach
traders = [Trader(model) for _ in range(3)]
alliance = create_meta_agent(
    model=model,
    name="TradeAlliance",
    components=traders,
    attributes={"market_power": sum(t.resources for t in traders)},
    methods={"trade": lambda self: self.market_power * 1.2}
)

# Class Approach
class TradeAlliance(MetaAgent):
    @classmethod
    def form_alliance(cls, model, traders):
        def calculate_trade(self):
            return self.market_power * 1.2
            
        return cls.create(
            model=model,
            name="TradeAlliance",
            components=traders,
            attributes={"market_power": sum(t.resources for t in traders)},
            methods={"trade": calculate_trade}
        )

alliance = TradeAlliance.form_alliance(model, traders)
```

3. Hierarchical Organization

```python
# Original Approach
employees = [Employee(model) for _ in range(5)]
department = create_meta_agent(
    model=model,
    name="Department",
    components=employees
)
company = create_meta_agent(
    model=model,
    name="Company",
    components=[department],
    methods={"reorganize": lambda self: len(self.components)}
)

# Class Approach
class Department(MetaAgent):
    def __init__(self, model, employees):
        super().__init__(model, employees)
        self.budget = sum(e.salary for e in employees)

class Company(MetaAgent):
    def __init__(self, model, departments):
        super().__init__(model, departments)
    
    def reorganize(self):
        return len(self.components)

department = Department(model, employees)
company = Company(model, [department])
```

4. Emergent Behavior

```python
# Original Approach
swarm = create_meta_agent(
    model=model,
    name="Swarm",
    components=drones,
    methods={"flock": lambda self: [d.move() for d in self.components]}
)

# Class Approach
class Swarm(MetaAgent):
    def flock(self):
        for drone in self.components:
            drone.move()
```

5. Resource Management

```python
# Original Approach
resource_pool = create_meta_agent(
    model=model,
    name="ResourcePool",
    components=resources,
    attributes={"total": sum(r.amount for r in resources)},
    methods={"allocate": lambda self: self.total >= amount}
)

# Class Approach
class ResourcePool(MetaAgent):
    def __init__(self, model, resources):
        super().__init__(model, resources)
        self.total = sum(r.amount for r in resources)
    
    def allocate(self, amount):
        return self.total >= amount
```

tpike3:
### Meta-Agents Phase 2

I got some time to update Meta-Agents for 3.2

#### Motivation

This is a cleaner variant of some of my earlier work on multi-level mesa (Thanks AgentSet!). Feedback from that work and recent interactions from students has always at least found the idea of agent groups in a meta-agent intriguing, so I believe it has the potential to be very worthwhile for Mesa Users

#### Overview

Allows users to either create meta-agents deliberately or create meta-agents through emergently through interaction of agents. These two approaches are captured in the alliance formation examples (emergent creation)  and the pseudo warehouse model (deliberate creation) 

There are two keys parts: 
 - One main class - `MetaAgent` which mirrors Mesa's Agent Class
 - One Main Function `create_meta-agents` function. 
 -  It also has two functions for evaluating whether or not agents should form a meta-agent.

#### PR Issues

This became and overly large PR due to the following (apologies)

- I switched IDEs and made some poor assumptions about how it handled command lines, virtual environments, and the code in my IDE, which caused me some commit fits and in my frustration led to me doing a git merge main instead of rebase. 
- I made some updates to configs docstring cutoff based on some JOSS feedback which impacted some other files (e.g. Agent.py and ListLike docstring)

#### Way Ahead

If this proves a useful feature, next step would be to allow agents to belong in multiple meta-agents. This could also then lead to a network based scheduler, that allows for active and dormant links based on agent situation/stimulus.

EwoutH:
I'm going to try to give this another good look this weekend!

tpike3:
Close PR to reopen a cleaner version

PR #2595, opened 2025-01-04
Contributing: Add link to video of release process

EwoutH:
I recorded a screen capture of me creating the Mesa v3.1.2 release: https://youtu.be/JE44jkegmns. This PR adds this URL to the Contributing guide for future reference.

Sorry for the bad audio, was recorded straight from my laptop.

CC @projectmesa/maintainers


jackiekazil (approved):
I never thought of adding something like this! LGTM!

PR #2596, opened 2025-01-04
Add render interval slider to control visualization update frequency

### LINKED ISSUE #2579: Allow updating visualisation every n steps (instead of every step)
With many different plots the plotting of all the matplotlib components takes up significant time in the visualisations. An easy and simple way to speed this up would be only plotting the graphs every N steps. We could add a slider to the visualisation that defaults to 1 and allows updating the graphs less frequently for performance.

HMNS19:
### Summary
This PR adds a render interval slider to the SolaraViz visualization system, allowing users to control how frequently plots and visualizations are updated during model simulation. This feature significantly improves performance when working with complex visualizations or multiple plots.

### Motive
When running models with multiple or complex visualizations, plotting components (especially Matplotlib) can become a performance bottleneck. This enhancement allows users to reduce visualization update frequency while maintaining full simulation speed, addressing the performance issues raised in #2579.

### Implementation
- Added a `render_interval` parameter to `SolaraViz` that defaults to 1
- Implemented a slider control (1-100, step size 2) to adjust the render interval
- Modified the `do_step()` function in both `ModelController` and `SimulatorController` to:
  - Run multiple simulation steps based on the render interval
  - Only update visualizations after completing all steps in the interval
- Added proper documentation and type hints for the new parameter

### Usage Examples
The render interval slider appears automatically in the visualization interface.

_TODO: Add screenshot_

### Additional Notes
- The render interval is separate from the play interval - it controls visualization updates while the play interval controls simulation speed
- The implementation maintains backward compatibility and doesn't affect model execution logic

### DIFF
--- a/mesa/examples/advanced/pd_grid/model.py
+++ b/mesa/examples/advanced/pd_grid/model.py
@@ -14,7 +14,13 @@ class PdGrid(mesa.Model):
     payoff = {("C", "C"): 1, ("C", "D"): 0, ("D", "C"): 1.6, ("D", "D"): 0}
 
     def __init__(
-        self, width=50, height=50, activation_order="Random", payoffs=None, seed=None,stepsize=1,
+        self,
+        width=50,
+        height=50,
+        activation_order="Random",
+        payoffs=None,
+        seed=None,
+        stepsize=1,
     ):
         """
         Create a new Spatial Prisoners' Dilemma Model.
--- a/mesa/examples/advanced/pd_grid/model.py
+++ b/mesa/examples/advanced/pd_grid/model.py
@@ -20,7 +20,6 @@ class PdGrid(mesa.Model):
         activation_order="Random",
         payoffs=None,
         seed=None,
-        stepsize=1,
     ):
         """
         Create a new Spatial Prisoners' Dilemma Model.
--- a/mesa/examples/advanced/pd_grid/model.py
+++ b/mesa/examples/advanced/pd_grid/model.py
@@ -14,12 +14,7 @@ class PdGrid(mesa.Model):
     payoff = {("C", "C"): 1, ("C", "D"): 0, ("D", "C"): 1.6, ("D", "D"): 0}
 
     def __init__(
-        self,
-        width=50,
-        height=50,
-        activation_order="Random",
-        payoffs=None,
-        seed=None,
+        self, width=50, height=50, activation_order="Random", payoffs=None, seed=None
     ):
         """
         Create a new Spatial Prisoners' Dilemma Model.
--- a/mesa/examples/advanced/pd_grid/model.py
+++ b/mesa/examples/advanced/pd_grid/model.py
@@ -14,7 +14,7 @@ class PdGrid(mesa.Model):
     payoff = {("C", "C"): 1, ("C", "D"): 0, ("D", "C"): 1.6, ("D", "D"): 0}
 
     def __init__(
-        self, width=50, height=50, activation_order="Random", payoffs=None, seed=None
+        self, width=50, height=50, activation_order="Random", payoffs=None, seed=None,stepsize=1,
     ):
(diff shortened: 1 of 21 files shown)


EwoutH:
@HMNS19 thanks for the PR! Is it already ready for a first round of review?

HMNS19:
@EwoutH  The work is almost ready for the first round of review. For the most part, the execution speed has improved, but it’s not as fast as expected for a few models. I’m currently working on optimizing those. I should have it ready in a day or two. Let me know if you'd like me to proceed with the review or wait until it's fully optimized.

EwoutH:
One thing I notice is that you change every example model. Ideally, I would like a generalized solution, directly in SolaraViz.

Like in this PR:

- https://github.com/projectmesa/mesa/pull/2540

CC @Corvince @AdamZh0u

HMNS19:
Understood, will work on it. Thank you for the feedback and the reference!

EwoutH:
Thanks, code looks good, have you tested it?

I will review later today.

HMNS19:
Yes, I have tested the example models, and they are working as expected. Most of them exhibit a display latency of less than 1 second, while the Epstein model has a slightly longer latency of around 2 seconds.

EwoutH:
This works perfect, and is really useful! Glad to see it could be done the same way as in #2540.

I do find the naming currently not the most intuitive, both of `play_interval` and `step_interval`. Maybe we can think of some better names, any suggestions?

HMNS19:
Sure, renaming `step_interval` to `render_interval` makes much more sense. However, I’m wondering what we should do about the `step button` in that case. Since  ideally clicking on step should increment step by 1, but it currently depends on the value of the `step_interval slider`. Should we modify its name to match `render_interval` so the users don't get confused about why the `step button`'s value also changes.
And if we keep the `step` button to increment the step by 1 regardless of the slider value, it could cause a mismatch between the current step and the current plot, as the value of the slider would differ.

Would you like me to proceed with any changes based on this?

EwoutH:
That's a very good catch. For now, I'm okay with leaving it as is, to not make this PR too big. But maybe document this well in the docstring / comments.

CC @quaquel @Corvince.

EwoutH:
I would like to call it `render_interval` by the way.

HMNS19:
I've made the necessary changes and renamed it to render_interval as you suggested. Let me know if any other changes are needed!

mesa/examples/advanced/pd_grid/model.py
```diff
@@ -14,7 +14,12 @@ class PdGrid(mesa.Model):
     payoff = {("C", "C"): 1, ("C", "D"): 0, ("D", "C"): 1.6, ("D", "D"): 0}
 
     def __init__(
-        self, width=50, height=50, activation_order="Random", payoffs=None, seed=None
+        self,
+        width=50,
+        height=50,
+        activation_order="Random",
+        payoffs=None,
+        seed=None,
```
EwoutH, line 22:
```suggestion
        self, width=50, height=50, activation_order="Random", payoffs=None, seed=None
```

mesa/visualization/solara_viz.py
```diff
@@ -390,7 +414,7 @@ def ModelCreator(
           or are dictionaries containing parameter details such as type, value, min, and max.
         - The `seed` argument ensures reproducibility by setting the initial seed for the model's random number generator.
         - The component provides an interface for adjusting user-defined parameters and reseeding the model.
-
+        -
```
EwoutH, line 417:
```suggestion
```

EwoutH:
@HMNS19 Looks good! I cleaned up the PR a little bit (you changed some unrelated files) and updated the PR title and description.

Finally, could you a screenshot of how the slider now looks to the PR description?

EwoutH (approved):

Issue #2559, opened 2024-12-20
Tracking issue: JOSS paper


EwoutH:
## Overview
This tracks the development of the Mesa JOSS paper that aims to provide a comprehensive overview of Mesa in its current state (Mesa 3). The paper will live on the [`paper_2024`](https://github.com/projectmesa/mesa/tree/paper_2024) branch.

Authors/maintainers, feel free  to edit this tracking issue directly and open PRs into the `paper_2024` branch.

### Content Structure & Organization
- [x] Rewrite paper to focus on Mesa as a complete framework rather than centering on 3.0 changes (@EwoutH) #2560
- [x] Add introductory paragraph about agent-based modeling
- [x] Restructure content into three main sections:
  - ABM Framework (agents/management, spaces, time management)
  - Visualization
  - Experimentation
- [x] Add Acknowledgements section

### Technical Elements
- [x] Integrate experimental features into main narrative sections
- [x] Review and update discussion of spaces (focusing on new-style spaces)

### References & Citations
- [x] Add more application references (https://github.com/projectmesa/mesa/pull/2572)
- [x] Include Mesa-Geo paper citation
- [x] Review and incorporate suggested bibtex entries
- [x] Add additional relevant references

### Author Management
- [x] Collect remaining ORCIDs and confirm affiliations https://github.com/projectmesa/mesa/pull/2567

## Next Steps
- [x] 1. Merge current PR into paper_2024 branch
- [x] 2. Implement standalone paper rewrite
- [x] 3. Individual PRs can be opened against paper_2024 branch for specific improvements

## Coordination
- Authors can submit changes via PR or provide feedback through GitHub suggestions

## Notes
- Paper should be readable standalone without requiring familiarity with previous Mesa papers
- Focus on presenting Mesa in its current form while mentioning 3.0 features in context
- Consider 5-year publication cycle (previous papers: 2015, 2020, likely 2025)

## Review process
- https://github.com/openjournals/joss-reviews/issues/7620

EwoutH:
On Jackie's request, there is now a Google Docs in which we continue to edit, suggest and review: https://docs.google.com/document/d/1HdoNDecjRvvmYhMyr9lZqPj3mmTOa3lrFv1a8_NCk38/edit?usp=sharing

@projectmesa/maintainers would appreciate everyone's review!

jackiekazil:
It looks like it is currently too long-- 1500+ and it needs to be <1000. I
am going to work on suggesting cuts to trim down.

[...]

PR #2597, opened 2025-01-05
paper: Update space section 

jackiekazil:
This took longer than I wanted, but I hope this gets at what I thought needed to be edited in this section. 
The edits include: 

- Explicitly highlight new additions in Mesa 3, so readers understand what has changed
- Explicitly highlight how/where to import from while these things are in the experimental stage 
- Other minor wording/grammar updates

Let me know if something needs to be adjusted. 
Related: #2559


quaquel (approved):
Very well done. I think this strikes a middle ground between what I would prefer and what @jackiekazil has been arguing for.

EwoutH (approved):
This is great, thanks!

jackiekazil:
Thank you both!

PR #2598, opened 2025-01-06
Bugfix for non deterministic rng behavior

quaquel:
This fixes a bug in how the numpy rng is seeded. If instantiating a model with `Model(seed=42)`, rng should default to 42, not `None`. The net result of this bug is that at the moment,  `Model(seed=42) will leave the numpy rgn undeterministic. This adds an explicit test to ensure that both model.random and model.rng behave in a deterministic manner when seeded with the same initial value and fixes the bug.

### DIFF
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -94,7 +94,7 @@ class Model:
             self._seed = seed  # this allows for reproducing stdlib.random
 
             try:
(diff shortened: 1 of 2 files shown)


EwoutH (approved):
Good catch!

Issue #2592, opened 2025-01-03
bug in ContinuousSpace.get_neighbors


quaquel:
**Describe the bug**
Currently, `get_neighbors` will fail to identify neighbors if two agents occupy the exact same position. 

The line that causes the issue is shown below. if `include_center` is `True`, this will return all agents on the position. If if `include_center` is `False`, it will return no agents.  However, in case of 2 agents on the same location, neither is correct. 

```
        neighbors = [
            self._index_to_agent[x] for x in idxs if include_center or dists[x] > 0
        ]
```
**Expected behavior**
Return any neighbors that are not the agent for which you are calling `get_neighbors`. 

**To Reproduce**
Create a space with 2 agents in the same location and call get_neighbors with the position of either agent. 

```
model = Model(seed=42)
space = ContinuousSpace(1, 1, torus=False)
agent1 = Agent(model)
space.place_agent(agent1, (0.5, 0.5))

agent2 = Agent(model)
space.place_agent(agent2, (0.5, 0.5))

# returns no agents, should return agent2
space.get_neighbors(agent1.pos, radius=1, include_center=False)

# returns both agents, should return only agent2
space.get_neighbors(agent1.pos, radius=1, include_center=True)
```

Sahil-Chhoker:
Upon looking at code I found that it is impossible to find the agent which is passed in the method, because you are passing the position, and since every agent has the same position, if you try to remove the so called current agent, you will have to remove a random one or none at all.

To overcome this one can change the type of the `pos` variable, that is if an `agent` is passed in the method, we can track that agent and exclude it from the result.

Can something like this work?
```python
if isinstance(pos, Agent):
            self.current_agent = pos
            pos = pos.pos
        else:
            self.current_agent = None

...

neighbors = [
            self._index_to_agent[x] for x in idxs
        ]

if not self.current_agent:
    return neighbors
else:
    neighbors.remove(self.current_agent)
    return neighbors

quaquel:
You are absolutely right that this bug cannot be fixed inside the method as it currently stands. However, changing the method signature is a breaking change and could have a profound impact on existing models. So, that is not an ideal solution. 

Practically speaking, with #2584 almost ready to go, I am thinking of the following
1. explicitly document the current buggy behavior. This is the minimum thing to do.
2. Add a new method with a different signature with the correct behavior. Not sure about this in light of #2584. 
[...]

PR #2599, opened 2025-01-06
Basic bugfix for bug ContinuousSpace.get_neighbors

quaquel:
This is a basic solution to #2592. In short, `ContinuousSpace.get_neighbors` takes an optional `include_center` keyword argument. If this is set to true, all agents on center will be removed, potentially missing neighbors. For details, see #2592.

This takes the most basic approach:
1. Document the behavior explicitly.
2. Update the boids example to not include center and explicitly filter out self. 

The reason for this basic fix is that any proper fix would be backward incompatible. Proper solutions would require distinguishing between getting agents in radius of a given coordinate and getting neighbors for a given agent. This would involve changing method names and changing signatures.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -1415,6 +1415,11 @@ class ContinuousSpace:
                             coordinates. i.e. if you are searching for the
                             neighbors of a given agent, True will include that
                             agent in the results.
+
+        Notes:
+            If 1 or more agents are located on pos, include_center=True will remove these agents
+            from the results.
+
         """
         if self._agent_points is None:
             self._build_agent_cache()
--- a/mesa/examples/basic/boid_flockers/agents.py
+++ b/mesa/examples/basic/boid_flockers/agents.py
@@ -58,7 +58,8 @@ class Boid(Agent):
 
     def step(self):
         """Get the Boid's neighbors, compute the new vector, and move accordingly."""
-        self.neighbors = self.model.space.get_neighbors(self.pos, self.vision, False)
+        neighbors = self.model.space.get_neighbors(self.pos, self.vision, True)
+        self.neighbors = [n for n in neighbors if n is not self]
 
         # If no neighbors, maintain current direction
         if not self.neighbors:


EwoutH:
Sounds like it’s working like intended, right? All agents with the same location are not considered neighbors.

Specifically for flocking, since agents are on continuous coordinates, isn’t the chance extremely small a neighbor has the exact same location (fp32 precision?) as another agent?

quaquel:
> Sounds like it’s working like intended, right? All agents with the same location are not considered neighbors.

I beg to differ. To me, if a given agent X asks for its neighbors, I expect to get all neighbors even if there happens to be an agent Y on the exact same location this agent Y is still a neighbor of agent X. 

> Specifically for flocking, since agents are on continuous coordinates, isn’t the chance extremely small a neighbor has the exact same location (fp32 precision?) as another agent?

Yes, this is very much an edge-case kind of issue, which is why the documentation solution is fine as far as I am concerned.

mesa/space.py
```diff
@@ -1415,6 +1415,13 @@ def get_neighbors(
                             coordinates. i.e. if you are searching for the
                             neighbors of a given agent, True will include that
                             agent in the results.
+
+        Notes:
+            If 1 or more agents are located on pos, include_center=True will remove these agents
+            from the results. So, if you really want to get the neighbors of a given agent,
```
tpike3, line 1421: I thing I am missing something in this sentence. Based on the change above and my understanding fo the issues shouldnt this sentence read 

If 1 or more agents are located on pos, include_center=False will remove these agents from the results and the given agent (i.e. self). So, if you really want to get the neighbors of a given agent, you should set include_center=True  and then filter the list of agents to remove the given agent (i.e. self).

quaquel: slightly rephrased drawing on your formulation.

mesa/space.py
```diff
@@ -1415,6 +1415,13 @@ def get_neighbors(
                             coordinates. i.e. if you are searching for the
                             neighbors of a given agent, True will include that
                             agent in the results.
+
+        Notes:
+            If 1 or more agents are located on pos, include_center=True will remove all these agents
```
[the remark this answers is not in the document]
quaquel: yes

EwoutH (approved):
Looks good, thanks!

EwoutH:
I updated the PR title and body to reflect the final changes. Feel free to adjust further or let me know if you don't like it.

PR #2609, opened 2025-01-10
Fixed hex-space draw function to avoid overlaps

### LINKED ISSUE #2438: Improve visual styling of grid drawing
Grids are optionally drawn for all discrete spaces. However, the look is not always esthetically pleasing. It might be worth to spend some time to tune how grids are drawn for all spaces.

Sahil-Chhoker:
### Summary
This PR fixes the overlapping of hex-tiles.
#### Before:
Screenshot 2025-01-10 154840  Screenshot 2025-01-10 161539

#### After:
image image

#### Before:
image

#### After:
image

### Bug / Issue
fixes #2438 

### Implementation
The old code used `RegularPolygon` to gather hexagon shapes and assemble them using `PatchCollection`. In the new implementation, all six vertices are calculated using simple math and the origin coordinates, then combined at the end using `LineCollection` to avoid duplicate edges.

No breaking changes have been made to the API.

### Testing
Tested using building a custom model, found no errors.

### DIFF
--- a/mesa/visualization/mpl_space_drawing.py
+++ b/mesa/visualization/mpl_space_drawing.py
@@ -347,7 +347,7 @@ def draw_hex_grid(
 
     def setup_hexmesh(width, height):
         """Helper function for creating the hexmesh with unique edges."""
-        edges = []
+        edges = set()
         size = 1.0
         x_spacing = np.sqrt(3) * size
         y_spacing = 1.5 * size
@@ -378,8 +378,7 @@ def draw_hex_grid(
             for v1, v2 in pairwise([*vertices, vertices[0]]):
                 # Sort vertices to ensure consistent edge representation and avoid duplicates.
                 edge = tuple(sorted([tuple(np.round(v1, 6)), tuple(np.round(v2, 6))]))
-                if edge not in edges:
-                    edges.append(edge)
+                edges.add(edge)
 
         # Return LineCollection for hexmesh
         return LineCollection(edges, linestyle=":", color="black", linewidth=1, alpha=1)
@@ -387,8 +386,6 @@ def draw_hex_grid(
     if draw_grid:
         ax.add_collection(setup_hexmesh(space.width, space.height))
 
-    # Set aspect ratio to 'equal' to ensure hexagons appear regular
-    ax.set_aspect("equal")
     return ax
 
 
--- a/mesa/visualization/mpl_space_drawing.py
+++ b/mesa/visualization/mpl_space_drawing.py
@@ -191,11 +191,10 @@ def draw_property_layers(
             continue
 
         data = layer.data.astype(float) if layer.data.dtype == bool else layer.data
-        width, height = data.shape  # if space is None else (space.width, space.height)
 
-        if space and data.shape != (width, height):
+        if (space.width, space.height) != data.shape:
             warnings.warn(
-                f"Layer {layer_name} dimensions ({data.shape}) do not match space dimensions ({width}, {height}).",
+                f"Layer {layer_name} dimensions ({data.shape}) do not match space dimensions ({space.width}, {space.height}).",
                 UserWarning,
                 stacklevel=2,
             )
@@ -208,6 +207,7 @@ def draw_property_layers(
 
         # Draw the layer
         if "color" in portrayal:
+            data = data.T
             rgba_color = to_rgba(portrayal["color"])
             normalized_data = (data - vmin) / (vmax - vmin)
             rgba_data = np.full((*data.shape, 4), rgba_color)
@@ -231,7 +231,7 @@ def draw_property_layers(
             if isinstance(cmap, list):
                 cmap = LinearSegmentedColormap.from_list(layer_name, cmap)
             im = ax.imshow(
-                data,
+                data.T,
                 cmap=cmap,
                 alpha=alpha,
                 vmin=vmin,
--- a/mesa/visualization/mpl_space_drawing.py
+++ b/mesa/visualization/mpl_space_drawing.py
@@ -318,7 +318,7 @@ def draw_hex_grid(
     size = 1.0
     x_spacing = np.sqrt(3) * size
     y_spacing = 1.5 * size
-    
+
     loc = arguments["loc"].astype(float)
     # Calculate hexagon centers
     loc[:, 0] = loc[:, 0] * x_spacing + (loc[:, 1] % 2) * (x_spacing / 2)
@@ -333,9 +333,11 @@ def draw_hex_grid(
     y_max = space.height * y_spacing
 
     # Add padding that accounts for the hexagon points
-    x_padding = size * np.sqrt(3)/2  # Distance from center to rightmost point of hexagon
+    x_padding = (
+        size * np.sqrt(3) / 2
+    )  # Distance from center to rightmost point of hexagon
     y_padding = size  # Distance from center to topmost point of hexagon
-    
+
     # Plot limits to perfectly contain the hexagonal grid
     # Determined through physical testing.
     ax.set_xlim(-2 * x_padding, x_max + x_padding)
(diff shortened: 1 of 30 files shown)


quaquel:
Thanks for this! I quickly scanned the code, and it made sense. I want to take a closer look, hopefully soon, to see if stuff can be simplified a bit. In the meantime, could you add a before and after of, say, a 20x20 grid next to what you already provided (which looks great, btw)?

Sahil-Chhoker:
Thank you for your response! I’ve added grid images with a 10x10 layout (as the 20x20 grid didn’t look as clean in the screenshot) while keeping the same line-dot pattern for better clarity.

mesa/visualization/mpl_space_drawing.py
```diff
@@ -323,62 +315,84 @@ def draw_hex_grid(
...
+    x_spacing = np.sqrt(3) * size
+    y_spacing = 1.5 * size
 
     loc = arguments["loc"].astype(float)
-
-    logical = np.mod(loc[:, 1], 2) == 0
-    loc[:, 0][logical] += 0.5
-    loc[:, 1] *= offset
+    # Calculate hexagon centers
+    loc[:, 0] = loc[:, 0] * x_spacing + (loc[:, 1] % 2) * (x_spacing / 2)
```
quaquel, line 325: So I think something is not right. Look at the top row in both figures. You'll see that they are differently aligned. This means that the neigborhood of a given cell as shown in the grid is not the same as in the underlying data. You can also check this by placing an agent in the very first cell. 

This is due to this line of code: `loc[:, 1] % 2`. Note how I check for 0, rather than 1.

mesa/visualization/mpl_space_drawing.py
```diff
@@ -323,62 +315,84 @@ def draw_hex_grid(
...
+            # Calculate center position for each hexagon with offset for even rows
+            x = col * x_spacing + (row % 2) * (x_spacing / 2)
+            y = row * y_spacing
+
+            vertices = get_hex_vertices(x, y)
+
+            # Edge logic, connecting each vertex to the next
+            for i in range(len(vertices)):
+                v1 = vertices[i]
+                v2 = vertices[(i + 1) % len(vertices)]
```
quaquel, line 379: I think that it should be possible to simplify this quite a bit. 

If I understand this code correctly, `v1` and `v2` are the start and end coordinates of the line that is to be drawn. So you could do something like

```python
from itertools import pairwise

for v1, v2 in pairwise(vertices+ [vertices[0]]):


```

mesa/visualization/mpl_space_drawing.py
```diff
@@ -323,62 +315,84 @@ def draw_hex_grid(
...
+
+                # Sort vertices to ensure consistent edge representation
+                edge = tuple(sorted([tuple(np.round(v1, 6)), tuple(np.round(v2, 6))]))
```
quaquel, line 382: 3 clarifying questions:
* why needs this to be sorted?
* why are you rounding here?
* why use numpy for rounding but sorted for sorting?

mesa/visualization/mpl_space_drawing.py
```diff
@@ -323,62 +315,84 @@ def draw_hex_grid(
...
+                edges.add(edge)
+
+        # Convert to LineCollection format
+        edges_list = [np.array(edge) for edge in edges]
```
quaquel, line 386: why not ensure that the edges are already in a list in the foregoing iteration?

quaquel (changes requested):
Again, thanks a lot for this PR. It's great to see how you solved the problem!

I have left a variety of suggestions for improving the code of this PR. If you have any questions, don't hesitate to ask.

mesa/visualization/mpl_space_drawing.py
quaquel: This issue is still not resolved.

mesa/visualization/mpl_space_drawing.py
```diff
@@ -323,62 +316,79 @@ def draw_hex_grid(
...
+    y_padding = size  # Distance from center to topmost point of hexagon
+
+    # Plot limits to perfectly contain the hexagonal grid
+    # Determined through physical testing.
+    ax.set_xlim(-2 * x_padding, x_max + x_padding)
+    ax.set_ylim(-2 * y_padding, y_max + y_padding)
+
+    def setup_hexmesh(width, height):
+        """Helper function for creating the hexmesh with unique edges."""
+        edges = []
```
quaquel, line 350:
```suggestion
        edges = set()
```

mesa/visualization/mpl_space_drawing.py
```diff
@@ -323,62 +316,79 @@ def draw_hex_grid(
...
+            x = col * x_spacing + (row % 2) * (x_spacing / 2)
+            y = row * y_spacing
+
+            vertices = get_hex_vertices(x, y)
+
+            # Edge logic, connecting each vertex to the next
+            for v1, v2 in pairwise([*vertices, vertices[0]]):
+                # Sort vertices to ensure consistent edge representation and avoid duplicates.
+                edge = tuple(sorted([tuple(np.round(v1, 6)), tuple(np.round(v2, 6))]))
+                if edge not in edges:
```
quaquel, line 381: this operation is expensive on a list, hence my suggestion above to change `edges` to a set.

mesa/visualization/mpl_space_drawing.py
```diff
@@ -323,62 +316,79 @@ def draw_hex_grid(
...
+                    edges.append(edge)
```
quaquel, line 382:
```suggestion
                    edges.add(edge)
```

mesa/visualization/mpl_space_drawing.py
```diff
@@ -323,62 +316,79 @@ def draw_hex_grid(
...
-        ax.add_collection(
-            setup_hexmesh(
-                space.width,
-                space.height,
-            )
-        )
+        ax.add_collection(setup_hexmesh(space.width, space.height))
+
+    # Set aspect ratio to 'equal' to ensure hexagons appear regular
+    ax.set_aspect("equal")
```
quaquel, line 391: Please leave this out. Users can do this in their post_process function. Adding it here renders this function inconsistent with the others.

quaquel:
There is still one issue unresolved regarding the alignment of hexes. Make the change shown below:


```python

loc[:, 0] = loc[:, 0] * x_spacing + ((loc[:, 1]-1) % 2) * (x_spacing / 2)
```

Sahil-Chhoker:
@quaquel I don't know if you were able to see the review I started, in which I clearly attached images of your suggestion not working, so just to make sure I am attaching one image again, your suggestion offsets the placement of agents. I don't know if this is the intended behavior, if it is I shall change the code.
image

quaquel:
I had missed that reply. Sorry. I'll try to take a closer look myself later this week.

quaquel:
Finally had time to look at this again. My suggestion only changed the agent location, but the same correction evidently also needs to be made to the hex centers: so `x = col * x_spacing + (row % 2 == 0) * (x_spacing / 2)`.

However, while figuring this out, I discovered 2 other bugs (which are also present in the current code.
1. If there are no agents, the modifications to loc given an error because  loc = arguments["loc"].astype(float) is empty. 
2. See the figure below. I created a simple grid with 1 central agent in blue and its neighbors in orange. 

Unknown

This is most definitely not right and seems to be related to mixing up row and column indexing. I'll try to figure this out asap.

Sahil-Chhoker:
I have added the reviewed changes, including the new empty `loc` check. And correct me if I'm wrong, the issue with neighbors existed in the previous code and was not introduced by this PR, right?

quaquel:
> the issue with neighbors existed in the previous code and was not introduced by this PR, right?

Yes this issue predates your PR. See #2632

quaquel:
I have merged the bugfix and updated this PR by rebasing.

Sahil-Chhoker:
@quaquel If there are no issues with the current approach, I would really appreciate it if this PR could be merged.

quaquel:
I hope to some final checks later this afternoon CET time.

quaquel:
Ok, I checked stuff visually, and it looks all ok now. Merging this.

Thanks again!

quaquel (approved):

PR #2610, opened 2025-01-10
Move mesa.experimental_cell to mesa.discrete_space

quaquel:
As part of the stabilization of cell spaces, this PR moves everything to `mesa.discrete_space`. I have kept the `mesa.experimental.cell_space`, which should continue to function but will issue a deprecation warning. 

I updated all tests and all examples that use cell spaces.

PR #2622, opened 2025-01-16
added color-bar for spice

sanika-n:
### Summary
In the [Sugarscape Model](https://github.com/projectmesa/mesa/tree/main/mesa/examples/advanced/sugarscape_g1mt), I noticed that while running it, the color bar was shown only for sugar and not for spice, so I added a color bar for spice as well.

### Before
image

### After
image


tpike3 (approved):
Thanks @sanika-n!

Notes on check failures: 
- benchmarks -- fails due to no scipy import -- will be fixed in #2610
- tests didnt reach diff target, based sized and purpose of merge not relevant

Issue #2285, opened 2024-09-07
Multiline plot is not yet documented


rht:
**What's the problem this feature will solve?**
#1941 was merged without documentation because it modified an internal function. It should have public documentation.

**Additional context**
Example usage for [virus on network](https://github.com/projectmesa/mesa-examples/tree/main/examples/virus_on_network)
```python
page = JupyterViz(
    VirusOnNetwork,
    model_params,
    measures=[
        # This is subject to change, as it could have been
        # {"Infected": {"color": "tab:red"}}, but let's keep it simple.
        {"Infected": "tab:red", "Susceptible": "tab:green", "Resistant": "tab:gray"},
        # Alternatively, if you don't want to manually specify colors:
        # ["Infected", "Susceptible", "Resistant"],
        make_text(get_resistant_susceptible_ratio),
    ],
    name="Virus Model",
    agent_portrayal=agent_portrayal,
)
```
It should look like [this](https://github.com/projectmesa/mesa/pull/1941#issuecomment-1879897461).

sanika-n:
@rht , I have created a PR #2624 which I think deals with the solara version of this issue, I think the 'make_plot_component' which I added the documentation for is what is currently used for multiline plot

PR #2624, opened 2025-01-16
Updated Docs

sanika-n:
### Added 'make_space_component' and 'make_plot_component' to Docs
These 2 functions were left out as they were [internal functions](https://github.com/projectmesa/mesa/blob/main/mesa/visualization/components/__init__.py).
I guess this PR also resolves this issue #2285 

image

image


quaquel:
You have this now listed under matplotlib, but these two methods are backend (i.e., altair or mpl) agnostic. I suggest moving the docs to the Jupyter visualization heading.

quaquel (approved):
Thanks!

PR #2628, opened 2025-01-18
Adding a copy option at the top of the code written in the docs

PrashantChoudhary13579:
### Summary
Adding a copy option at the top of the code written in the docs

### Motive
New users who just want to explore the things it would be easier for them if a copy to clipboard (copy) option is available for the code of the model written in the docs 

### Implementation

- #2619 

This, Discussion contains the implementation, Example, Outputs.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -47,6 +47,7 @@ extensions = [
     "sphinx.ext.viewcode",
     "sphinx.ext.napoleon",  # for google style docstrings
     "myst_nb",  # For Markdown and Jupyter notebooks
+    "sphinx_copybutton" # For copying the code in the documentation
 ]
 
 # Add any paths that contain templates here, relative to this directory.
--- a/README.md
+++ b/README.md
@@ -34,21 +34,16 @@ To install our latest stable release (3.0.x), run:
 pip install -U mesa
 ```
 
-To install our latest pre-release, run:
-
-``` bash
-pip install -U --pre mesa
-```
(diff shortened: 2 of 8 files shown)


docs/conf.py
```diff
@@ -47,6 +47,7 @@
     "sphinx.ext.viewcode",
     "sphinx.ext.napoleon",  # for google style docstrings
     "myst_nb",  # For Markdown and Jupyter notebooks
+    "sphinx_copybutton" # For copying the code in the documentation
```
tpike3, line 50: The read the docs build is failing because 

```
Extension error:
Could not import extension sphinx_copybutton (exception: No module named 'sphinx_copybutton')
```

If I remember correctly you also need to add it to the `pyproj.toml`

tpike3 (changes requested):
Thanks for doing this @PrashantChoudhary13579 ; the copy code should be a nice addition; 

We just need to make sure the read the docs build passes in the checks

PrashantChoudhary13579:
> Thanks for doing this @PrashantChoudhary13579 ; the copy code should be a nice addition;
> 
> We just need to make sure the read the docs build passes in the checks

Thank you! I'm glad you think so. I hope this addition improves the documentation experience for everyone!

docs/conf.py
PrashantChoudhary13579: @tpike3 I tried as you say and it passes all the test cases. Thanks a lot !

tpike3 (approved):

PR #2633, opened 2025-01-20
benchmarks.yml: Install SciPy and use uv for pip install

EwoutH:
Fixes the broken benchmark workflow.

- Install SciPy, since that's now a dependency
- Use uv for pip install, which is way faster


quaquel (approved):
looks good to me

PR #2636, opened 2025-01-21
Remove version specific docs

EwoutH:
Update version references in docs to be more generic, avoiding the need to update them with each release.

Also, SolaraViz has been stable and well-tested through the Mesa 3.0.x series. Remove the experimental warnings and version-specific notes from documentation, as it is now a stable component of Mesa 3.1.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -177,8 +177,6 @@ The results are returned as a list of dictionaries, which can be easily converte
 ### Visualization
 Mesa now uses a new browser-based visualization system called SolaraViz. This allows for interactive, customizable visualizations of your models.
 
-> **Note:** SolaraViz is experimental and still in active development for Mesa 3.0. While we attempt to minimize them, there might be API breaking changes between Mesa 3.0 and 3.1. There won't be breaking changes between Mesa 3.0.x patch releases.
-
 > **Note:** SolaraViz instantiates new models using `**model_parameters.value`, so all model inputs must be keyword arguments.
 
 Ensure your model's `__init__` method accepts keyword arguments matching the `model_params` keys.
--- a/mesa/visualization/__init__.py
+++ b/mesa/visualization/__init__.py
@@ -1,10 +1,4 @@
-"""Solara based visualization for Mesa models.
-
-.. note::
-    SolaraViz is experimental and still in active development for Mesa 3.0. While we attempt to minimize them, there might be API breaking changes between Mesa 3.0 and 3.1.
-
-    There won't be breaking changes between Mesa 3.0.x patch releases.
-"""
+"""Solara based visualization for Mesa models."""
 
 from mesa.visualization.mpl_space_drawing import (
     draw_space,
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ can be displayed in browser windows or Jupyter.*
 
 ## Using Mesa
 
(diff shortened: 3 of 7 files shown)


docs/getting_started.md
```diff
@@ -177,8 +177,6 @@ The results are returned as a list of dictionaries, which can be easily converte
 ### Visualization
 Mesa now uses a new browser-based visualization system called SolaraViz. This allows for interactive, customizable visualizations of your models.
 
-> **Note:** SolaraViz is experimental and still in active development for Mesa 3.0. While we attempt to minimize them, there might be API breaking changes between Mesa 3.0 and 3.1. There won't be breaking changes between Mesa 3.0.x patch releases.
```
quaquel, line 180: I am not sure about this removal. Yes solara as basis is fine, but I am not convinced yet fully by the api side of things.

mesa/visualization/__init__.py
```diff
@@ -1,10 +1,4 @@
-"""Solara based visualization for Mesa models.
-
-.. note::
-    SolaraViz is experimental and still in active development for Mesa 3.0. While we attempt to minimize them, there might be API breaking changes between Mesa 3.0 and 3.1.
```
quaquel, line 4: see my earlier comment

quaquel (commented):
fine except the solara experimental stuff.

docs/getting_started.md
EwoutH: We can't keep it experimental forever right? Maybe we need to commit to the current API and move to 4.0 if we want to change / overhaul it again.

quaquel: Nothing has changed fundamentally compared to where we were when releasing 3.0. So, all the reasons we had back then for keeping it experimental still stand. The passing of time alone is not a reason for making something suddenly stable.

In fact, the recent bugs we have found underline the immaturity of the code, further validating this original choice to keep it experimental.

EwoutH: Not sure if I fully align on the time passing statement, but in general sentiment I agree.

Shall we change the note to simply say 3.x?

> **Note:** SolaraViz is experimental and still in active development in Mesa 3.x. While we attempt to minimize them, there might be API breaking changes in minor releases.

docs/getting_started.md
```diff
@@ -177,8 +177,6 @@ The results are returned as a list of dictionaries, which can be easily converte
 ### Visualization
 Mesa now uses a new browser-based visualization system called SolaraViz. This allows for interactive, customizable visualizations of your models.
```
quaquel, line 179:
```suggestion

Note: SolaraViz is experimental and still in active development in Mesa 3.x. While we attempt to minimize them, there might be API breaking changes in minor releases.
```

mesa/visualization/__init__.py
```diff
@@ -1,10 +1,4 @@
...
-
-    There won't be breaking changes between Mesa 3.0.x patch releases.
-"""
+"""Solara based visualization for Mesa models."""
```
quaquel, line 1:
```suggestion
"""Solara based visualization for Mesa models."""

.. note::
    SolaraViz is experimental and still in active development in Mesa 3.x. While we attempt to minimize them, there might be API breaking changes in minor releases.
```

quaquel:
I used your suggested text in both places, so we can more forward with this.

PR #2639, opened 2025-01-23
bugfix for draw_property_layer

### LINKED ISSUE #2635: Property layer visualization doesn't work properly on rectangular grids.
**Describe the bug**
Property layer visualization only works on square grids.

**Expected behavior**
Visualization of property layer should work for entire grid.

#### Minimal Example
```python
from mesa.experimental.cell_space.grid import OrthogonalMooreGrid
from mesa.experimental.cell_space.grid import HexGrid
from mesa import Model
import math
from mesa.experimental.cell_space import CellAgent
from mesa.visualization import SolaraViz, make_space_component, make_plot_component
from mesa.experimental.cell_space.property_layer import PropertyLayer
import numpy as np

class myAgent(CellAgent):
    def __init__(self, model, cell=None):
        super().__init__(model) 
        self.number = 0 
        self.cell = cell

    def step(self):
        self.number = self.random.randint(1, 5)
    
 
class myModel(Model): 
    def __init__(self, seed=None, width=6,  height=4, n_agents=4):
        super().__init__(seed=seed) 
        self.width = width
        self.height = height
        self.num_agents = n_agents 
            (self.width, self.height), capacity=math.inf, torus=False, random=self.random
        )

        self.test_layer = PropertyLayer("test", (self.grid.width, self.grid.height), default_value=1, dtype=int)
        self.test_layer.data = np.random.randint(2, 10, (self.width, self.height))
        self.grid.add_property_layer(self.test_layer)

        myAgent(self, self.grid.all_cells.cells[14])
        
            model_reporters={
                "number": lambda m: sum(agent.number for agent in m.agents)
            },
            agent_reporters={"number": "number"},
        )

        self.running = True

    def step(self):


def portrayal(agent):
    if agent is None:
        return None
    
    if agent.unique_id == 1:
        return {
            "color": "tab:green", 
            "size": 100,
        }

    return {
        "color": "tab:orange",
        "size": 70,
    }


def post_process(ax):
    ax.set_aspect("equal")
    ax.set_xticks([])
    ax.set_yticks([])
    ax.get_figure().set_size_inches(8, 8)

property_layer_portrayal = {
        "test": {
            "color": "orange",
            "alpha": 0.7,
        }
    }

model_params = {
    "seed": {
        "type": "InputText",
        "value": 42,
        "label": "Random Seed",
    },
    "n_agents": {
        "type": "SliderInt",
        "value": 4,
        "min": 1,
        "max": 10,
        "step": 1,
        "label": "Number of Agents",
    },
    "width": 6,
    "height": 5,
}

model = myModel()

space_component = make_space_component(
    agent_portrayal=portrayal, draw_grid=True, post_process=post_process, propertylayer_portrayal=property_layer_portrayal
)

plot_component = make_plot_component(
    {"number": "tab:blue"},
)

page = SolaraViz(
    model,
    components=[space_component, plot_component],
    model_params=model_params,
    name="Test Model",
)
page
```

#### Visualization for the above example

Image

#### Versions I am running this example on:
Solara: 1.43.0
mesa: 3.1.3
matplotlib: 3.10.0
python: 3.13.1

quaquel:
This closes #2635. Axes.imshow uses row, col, while our property layer data is x, y. This solves this by transposing the data in the axes.imshow command.

### DIFF
--- a/mesa/visualization/mpl_space_drawing.py
+++ b/mesa/visualization/mpl_space_drawing.py
@@ -192,8 +192,9 @@ def draw_property_layers(
 
         data = layer.data.astype(float) if layer.data.dtype == bool else layer.data
         width, height = data.shape  # if space is None else (space.width, space.height)
+        data = data.T
 
-        if space and data.shape != (width, height):
+        if space.dimensions != (width, height):
             warnings.warn(
                 f"Layer {layer_name} dimensions ({data.shape}) do not match space dimensions ({width}, {height}).",
                 UserWarning,
--- a/mesa/visualization/mpl_space_drawing.py
+++ b/mesa/visualization/mpl_space_drawing.py
@@ -192,7 +192,7 @@ def draw_property_layers(
 
         data = layer.data.astype(float) if layer.data.dtype == bool else layer.data
 
-        if space.dimensions != data.shape:
+        if (space.width, space.height) != data.shape:
             warnings.warn(
                 f"Layer {layer_name} dimensions ({data.shape}) do not match space dimensions ({width}, {height}).",
                 UserWarning,
--- a/mesa/visualization/mpl_space_drawing.py
+++ b/mesa/visualization/mpl_space_drawing.py
@@ -194,7 +194,7 @@ def draw_property_layers(
 
         if (space.width, space.height) != data.shape:
             warnings.warn(
-                f"Layer {layer_name} dimensions ({data.shape}) do not match space dimensions ({width}, {height}).",
+                f"Layer {layer_name} dimensions ({data.shape}) do not match space dimensions ({space.width}, {space.height}).",
                 UserWarning,
                 stacklevel=2,
             )
@@ -207,6 +207,7 @@ def draw_property_layers(
 
         # Draw the layer
         if "color" in portrayal:
+            data = data.T
             rgba_color = to_rgba(portrayal["color"])
             normalized_data = (data - vmin) / (vmax - vmin)
             rgba_data = np.full((*data.shape, 4), rgba_color)
@@ -216,7 +217,7 @@ def draw_property_layers(
                 layer_name, [(0, 0, 0, 0), (*rgba_color[:3], alpha)]
             )
             im = ax.imshow(
-                rgba_data.T,
+                rgba_data,
                 origin="lower",
             )
             if colorbar:
--- a/mesa/visualization/mpl_space_drawing.py
+++ b/mesa/visualization/mpl_space_drawing.py
@@ -191,10 +191,8 @@ def draw_property_layers(
             continue
 
         data = layer.data.astype(float) if layer.data.dtype == bool else layer.data
-        width, height = data.shape  # if space is None else (space.width, space.height)
-        data = data.T
 
-        if space.dimensions != (width, height):
+        if space.dimensions != data.shape:
             warnings.warn(
                 f"Layer {layer_name} dimensions ({data.shape}) do not match space dimensions ({width}, {height}).",
                 UserWarning,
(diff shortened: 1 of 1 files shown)


EwoutH:
@Sahil-Chhoker since you put in the original issue, would you like to review?

Sahil-Chhoker:
@EwoutH Sure, I would love to!

mesa/visualization/mpl_space_drawing.py
```diff
@@ -191,9 +191,8 @@ def draw_property_layers(
             continue
 
         data = layer.data.astype(float) if layer.data.dtype == bool else layer.data
-        width, height = data.shape  # if space is None else (space.width, space.height)
 
-        if space and data.shape != (width, height):
+        if (space.width, space.height) != data.shape:
             warnings.warn(
                 f"Layer {layer_name} dimensions ({data.shape}) do not match space dimensions ({width}, {height}).",
                 UserWarning,
```
Sahil-Chhoker, line 198: Does checking here the width and height really necessary, because there are already checks in the property layer implementation that prevent such edge case from ever occurring. Refer to [link](https://github.com/projectmesa/mesa/blob/main/mesa/experimental/cell_space/property_layer.py#L242)

mesa/visualization/mpl_space_drawing.py
```diff
@@ -191,9 +191,8 @@ def draw_property_layers(
             continue
 
         data = layer.data.astype(float) if layer.data.dtype == bool else layer.data
-        width, height = data.shape  # if space is None else (space.width, space.height)
 
-        if space and data.shape != (width, height):
+        if (space.width, space.height) != data.shape:
             warnings.warn(
                 f"Layer {layer_name} dimensions ({data.shape}) do not match space dimensions ({width}, {height}).",
```
Sahil-Chhoker, line 197: `width` and `height` does not exist, change them to `space.width`, `space.height`

mesa/visualization/mpl_space_drawing.py
```diff
@@ -217,7 +216,7 @@ def draw_property_layers(
                 layer_name, [(0, 0, 0, 0), (*rgba_color[:3], alpha)]
             )
             im = ax.imshow(
-                rgba_data,
+                rgba_data.T,
```
Sahil-Chhoker, line 219: This approach is flawed. It will give a TypeError since let's say, we have a grid of shape (10,15), therefore the `rgba_data` will be of shape (10, 15, 4) and its transpose will result in shape (4, 15, 10) which can't be processed by `imshow()`. By transposing the data before RGBA operations, you ensure the correct (height, width, channels) format that Matplotlib requires, preventing dimensional misalignments that would otherwise cause TypeErrors during image rendering.

mesa/visualization/mpl_space_drawing.py
```diff
@@ -191,9 +191,8 @@ def draw_property_layers(
             continue
 
         data = layer.data.astype(float) if layer.data.dtype == bool else layer.data
-        width, height = data.shape  # if space is None else (space.width, space.height)
```
Sahil-Chhoker, line 194: I would suggest adding these two lines:
```python
data = data.T
height, width = data.shape
```

Sahil-Chhoker (changes requested):

mesa/visualization/mpl_space_drawing.py
quaquel: in the new-style cell spaces, we check, but not in the old style _PropertyGrid.

mesa/visualization/mpl_space_drawing.py
quaquel: good catch.

mesa/visualization/mpl_space_drawing.py
quaquel: I fixed your comments in a slightly different way but the effect is the same

Sahil-Chhoker:
This was a fairly simple fix, @EwoutH this PR can be merged!

EwoutH (approved):

Issue #2436, opened 2024-10-30
Shift from dict for agent_potrayal to an AgentPortrayaStyle class


quaquel:
Currently, `agent_portrayal` should return a dict. Valid fields are `color`, `marker`, `size`, and `zorder`. In #2430 and i[n a comment on #2389](https://github.com/projectmesa/mesa/issues/2389#issuecomment-2439549903), it has been suggested to shift to an `AgentPortrayaStyle` class instead. The benefit of this is that is much easier to write your own agent_portrayal function. It also ensures that only valid fields are used. It might make it easier to write agent_portrayal functions that are indifferent about the plotting backend (i.e., altair or matplotlib). Last, it becomes trivial for all `draw_x` functions in `mesa.visualization.compoments.matplotlib` to be modifies so agent_portrayal becomes an optional keyword argument.

Sahil-Chhoker:
@quaquel, I personally believe that using a dedicated class for agent portrayal is a great idea. I've implemented a simple version based on my vision.

```python
@dataclass
class AgentPortrayalStyle:

    color: str | tuple = "tab:blue"
    marker: str = "o"
    size: int = 50
    zorder: int = 1
    valid_marker: list ["o", "s", "^", ...]

    def __post_init__(self):
        """Validate the attributes after initialization."""
        if self.color is not None:
            # Validate color
            try:
                mcolors.to_rgb(self.color)
            except ValueError:
                raise ValueError(f"Invalid color specification: {self.color}")

        if self.marker is not None:
            # Validate Markers
            if self.marker not in VALID_MARKERS:
                raise ValueError(f"Invalid marker '{self.marker}'.)
        
        if self.size is not None and not isinstance(self.size, Number):
            raise ValueError(f"Size must be a number, got {type(self.size)}")
        
        if self.zorder is not None and not isinstance(self.zorder, int):
            raise ValueError(f"Zorder must be an integer, got {type(self.zorder)}")

    def to_dict(self) -> dict:
        """Convert the style to a dictionary."""
        return {k: v for k, v in self.__dict__.items() if v is not None}
```

#### Example usage can be something like:
```python
def portrayal(agent):
    if agent is None:
        return None
    
    style = AgentPortrayalStyle(
        color="tab:red" if agent.energy > 50 else "tab:blue",
        size=agent.size * 10,
        marker="^" if agent.is_predator else "o",
        zorder=2 if agent.is_predator else 1
    )
    return style.to_dict() # convert to dict
```
[...]

PR #2640, opened 2025-01-23
Added conditional checks for empty grid placement

### LINKED ISSUE #2638: draw_space only works if there is at least one agent
**Describe the bug**
If you try to draw a space and there are no agents in the space, you get an index error

**Expected behavior**
the space should simply be drawn without any agents

**To Reproduce**

```python
from mesa.experimental.cell_space import OrthogonalMooreGrid, CellAgent
from mesa import Model
from mesa.visualization import draw_space

model = Model(seed=42)
space = OrthogonalMooreGrid((10,5), random=model.random)

fig, ax = plt.subplots()

draw_space(space, ax=ax)
plt.show()
```

Sahil-Chhoker:
### Summary
Added `if` checks to confirm existence of agents before trying to place them on the grid, which before resulted in a error, now just displays the empty grid.

### Bug / Issue
fixes #2638 

### Additional Notes
No changes made to the `draw_hex_grid` function since the required changes are included in #2609.

### DIFF
--- a/mesa/visualization/mpl_space_drawing.py
+++ b/mesa/visualization/mpl_space_drawing.py
@@ -276,8 +276,9 @@ def draw_orthogonal_grid(
     s_default = (180 / max(space.width, space.height)) ** 2
     arguments = collect_agent_data(space, agent_portrayal, size=s_default)
 
-    # plot the agents
-    _scatter(ax, arguments, **kwargs)
+    if arguments["loc"].size > 0:
+        # plot the agents
+        _scatter(ax, arguments, **kwargs)
 
     # further styling
     ax.set_xlim(-0.5, space.width - 0.5)
@@ -433,10 +434,12 @@ def draw_network(
     # this assumes that nodes are identified by an integer
     # which is true for default nx graphs but might user changeable
     pos = np.asarray(list(pos.values()))
-    arguments["loc"] = pos[arguments["loc"]]
 
-    # plot the agents
-    _scatter(ax, arguments, **kwargs)
+    if arguments["loc"].size > 0:
+        arguments["loc"] = pos[arguments["loc"]]
+
+        # plot the agents
+        _scatter(ax, arguments, **kwargs)
 
     # further styling
     ax.set_axis_off()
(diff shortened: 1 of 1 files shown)


quaquel:
This is related to #2638 and #2642.

I need to think about the best way forward to address this. Yes, we can add a bunch of if statements as done here. However, I am inclined to see this as an architectural problem. At the moment, we don't distinguish carefully between drawing the space (e.g., the grid structure), the drawing of properties of the space (i.e., property layer portrayal), and the drawing of agents within the space. These are effectively all collapsed into a single draw_space method. I am inclined to redesign the API and explicitly separate these three things slightly. If you pass the same axes (in the case of matplotlib), you can draw them on top of each other, but if you pass separate axes, you can also portray them more cleanly. Any thoughts on this are welcome, preferably in #2642.

Sahil-Chhoker:
I think we should implement an `AgentPortrayalStyle` class as proposed in #2436 first. Then we can do something similar with the property layer portrayal as well:
```python
class PropertyLayerPortrayal:
    def __init__(self, layer_name, colormap=None, color=None, alpha=1.0, vmin=None, vmax=None):
        self.layer_name = layer_name
        self.colormap = colormap
        self.color = color
        # etc...
```

And then we can separate the space drawing into three parts. Maybe something like this:
```python
class SpaceRenderer:
    def draw_structure(self, space, ax=None, **kwargs):
       ...
    def draw_properties(self, space, propertylayer_portrayal, ax=None, **kwargs):
       ...
    def draw_agents(self, space, agent_portrayal, ax=None, **kwargs):
       ...

    # Composite method(not the best name)
    def draw_all(self, space, agent_portrayal=None, property_config=None, ax=None, **kwargs):
       ...
 ```
 
 But this design may feel complex for beginners while also breaking the current API.

quaquel:
@Sahil-Chhoker I like the direction of thought: encapsulate the portrayal of agents, space, and propertylayers into their own class, have three dedicated draw functions for each of them, and an overarching one that calls each of them. I'll try to find time to hash this out a bit further in the coming days.

> But this design may feel complex for beginners while also breaking the current API.

1. I don't know yet whether it will be complex. Ideally it should be simple for default cases yet be powerfull enough to cover whatever you want.
2. the current API is experimental, so we don't have to worry about it too much.
3. As shown in #2644, its not too difficult to maintain the current API while raising warnings pointing the user to the new API. This should ease transition.

Sahil-Chhoker:
@quaquel, As I'm expecting, does the implementation look like [this](https://github.com/projectmesa/mesa/issues/2389#issuecomment-2439679944) in your mind?

quaquel:
The space drawer idea you link to was heavily inspired by altair's API. I still like the idea, and I think it is worth exploring further. However, it is also a more radical departure from the current API.  Having `<class>PortrayalStyle` objects might be a simpler solution that is more compatible with the current design, while still eliminating the difficult to memorize dicts and nested dicts. In short, I am not sure.

PR #2646, opened 2025-01-26
Fix: Property layer visualization for HexGrid

### LINKED ISSUE #2433: Grid and property layer for hexgrid
#2430 adds support for plotting hexgrids. It comes with an optional grid using a patchcollection. This results in some overdrawing of lines wich is less than ideal. 

Also, property layers won't display correctly yet because the coordinate systems are different. 

The solution to both might be to create a polycollection (as used in matplotlib's hexbin).

Sahil-Chhoker:
### Summary
Property layers were not properly visualized for HexGrids, and this PR aims to resolve that issue. It utilizes the same coordinate system as the `draw_hex_grid` function and generates a `PolyCollection` that overlays color values on the grid.

### Bug / Issue
fixes #2433 

### Before
Screenshot 2025-01-26 203400

### After
Screenshot 2025-01-26 203521

### Additional Notes:
I am not entirely sure if this was the best approach to address the issue. If there is a more effective or elegant solution, I would greatly appreciate your recommendations.

### DIFF
--- a/mesa/visualization/mpl_space_drawing.py
+++ b/mesa/visualization/mpl_space_drawing.py
@@ -11,7 +11,8 @@ import itertools
 import warnings
 from collections.abc import Callable
 from itertools import pairwise
-from typing import Any
+from typing import Any, Iterator
+from functools import lru_cache
 
 import networkx as nx
 import numpy as np
@@ -175,6 +176,21 @@ def _get_hex_vertices(
     return vertices
 
 
+@lru_cache(maxsize=1024, typed=True)
+def _get_hexmesh(width: int, height: int, size: float) -> Iterator[list[tuple[float, float]]]:
+    """
+    Generate hexagon vertices for the mesh. Yields list of vertex coordinates for each hexagon
+    """
+    x_spacing = np.sqrt(3) * size
+    y_spacing = 1.5 * size
+    
+    for row, col in itertools.product(range(height), range(width)):
+        # Calculate center position with offset for even rows
+        x = col * x_spacing + (row % 2 == 0) * (x_spacing / 2)
+        y = row * y_spacing
+        yield _get_hex_vertices(x, y, size)
+
+
 def draw_property_layers(
     space, propertylayer_portrayal: dict[str, dict[str, Any]], ax: Axes
 ):
@@ -192,32 +208,12 @@ def draw_property_layers(
 
     """
 
-    def get_hexmesh(width: int, height: int) -> list[list[tuple[float, float]]]:
-        """Create hexagon vertices for the mesh."""
-        hexagons = []
-        size = 1
-        x_spacing = np.sqrt(3) * size
-        y_spacing = 1.5 * size
-
-        for row in range(height):
-            for col in range(width):
-                # Calculate center position with offset for even rows
-                x = col * x_spacing + (row % 2 == 0) * (x_spacing / 2)
-                y = row * y_spacing
-                vertices = _get_hex_vertices(x, y, size)
-                hexagons.append(vertices)
-
-        return hexagons
-
     try:
         # old style spaces
         property_layers = space.properties
     except AttributeError:
         # new style spaces
         property_layers = space._mesa_property_layers
-
-    # Check space type once
-    is_hex_grid = isinstance(space, HexGrid)
 
     for layer_name, portrayal in propertylayer_portrayal.items():
         layer = property_layers.get(layer_name, None)
@@ -259,28 +255,7 @@ def draw_property_layers(
                 f"PropertyLayer {layer_name} portrayal must include 'color' or 'colormap'."
             )
 
-        if is_hex_grid:
-            width, height = data.shape
-
-            # Generate hexagon mesh
-            hexagons = get_hexmesh(width, height)
-
-            # Normalize colors
-            norm = Normalize(vmin=vmin, vmax=vmax)
-            colors = data.ravel()  # flatten data to 1D array
-
-            if "color" in portrayal:
-                normalized_colors = np.clip(norm(colors), 0, 1)
-                rgba_colors = np.full((len(colors), 4), rgba_color)
-                rgba_colors[:, 3] = normalized_colors * alpha
-            else:
-                rgba_colors = cmap(norm(colors))
-
-            # Draw hexagons
-            collection = PolyCollection(hexagons, facecolors=rgba_colors, zorder=-1)
-            ax.add_collection(collection)
-        else:
-            # Rectangular grid rendering
+        if isinstance(space, OrthogonalGrid):
             if "color" in portrayal:
                 data = data.T
                 normalized_data = (data - vmin) / (vmax - vmin)
@@ -297,6 +272,32 @@ def draw_property_layers(
                     vmax=vmax,
                     origin="lower",
                 )
+
+        elif isinstance(space, HexGrid):
+            width, height = data.shape
+
+            # Generate hexagon mesh
+            hexagons = _get_hexmesh(width, height, size=1)
+
+            # Normalize colors
+            norm = Normalize(vmin=vmin, vmax=vmax)
+            colors = data.ravel()  # flatten data to 1D array
+
+            if "color" in portrayal:
+                normalized_colors = np.clip(norm(colors), 0, 1)
+                rgba_colors = np.full((len(colors), 4), rgba_color)
+                rgba_colors[:, 3] = normalized_colors * alpha
+            else:
+                rgba_colors = cmap(norm(colors))
+
+            # Draw hexagons
+            collection = PolyCollection(hexagons, facecolors=rgba_colors, zorder=-1)
+            ax.add_collection(collection)
+
+        else:
+            raise NotImplementedError(
+                f"PropertyLayer visualization not implemented for {type(space)}."
+            )
 
         # Add colorbar if requested
         if colorbar:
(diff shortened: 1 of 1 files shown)


quaquel:
Thanks for this. The example looks great.  I am taking a look at the code now.

mesa/visualization/mpl_space_drawing.py
```diff
@@ -205,46 +242,69 @@ def draw_property_layers(
...
-            if colorbar:
-                plt.colorbar(im, ax=ax, label=layer_name)
+            elif isinstance(cmap, str):
+                cmap = plt.get_cmap(cmap)
         else:
             raise ValueError(
                 f"PropertyLayer {layer_name} portrayal must include 'color' or 'colormap'."
             )
 
+        if is_hex_grid:
```
quaquel, line 262: I suggest flipping this arround. It is best practice that the most common clause in an if else comes first. So orthogonal grids are more common than hex_grids.

mesa/visualization/mpl_space_drawing.py
```diff
@@ -175,13 +191,34 @@ def draw_property_layers(
...
+        size = 1
+        x_spacing = np.sqrt(3) * size
+        y_spacing = 1.5 * size
+
+        for row in range(height):
+            for col in range(width):
+                # Calculate center position with offset for even rows
+                x = col * x_spacing + (row % 2 == 0) * (x_spacing / 2)
+                y = row * y_spacing
+                vertices = _get_hex_vertices(x, y, size)
```
quaquel, line 207: This is repetetive with lines 417 - 420. Might it be possible to add an additional helper method that returns all vertices for a given hexgrid? If so, we can even cash this using `functools.cache` for a small speedup when plotting.

mesa/visualization/mpl_space_drawing.py
```diff
@@ -205,46 +242,69 @@ def draw_property_layers(
...
+            # Normalize colors
+            norm = Normalize(vmin=vmin, vmax=vmax)
+            colors = data.ravel()  # flatten data to 1D array
+
+            if "color" in portrayal:
+                normalized_colors = np.clip(norm(colors), 0, 1)
+                rgba_colors = np.full((len(colors), 4), rgba_color)
+                rgba_colors[:, 3] = normalized_colors * alpha
+            else:
+                rgba_colors = cmap(norm(colors))
```
quaquel, line 277: can't this be moved outside the if else since it should be largely the same for both types of grids?

quaquel (commented):
jJust a few questions and ideas. Looks very good already.

Sahil-Chhoker:
Are you sure it's fine as it is? I had a few questions about it:  

1. Regarding the repetitive part, I think I’ve got an answer for now. I’ll get back to you as early as tomorrow.  
2. I'm hardcoding the value of `zorder` to -1. Is this the correct approach?

quaquel:
I gave it a first quick pass and focussed on the high-level stuff. The zorder one is a good question (and actually interacts with #2642). Ideally (but beyond this pr), zorder should become controllable for property layers to explicitly control layering.

mesa/visualization/mpl_space_drawing.py
Sahil-Chhoker: I don't think that is possible, since `imshow()` is not used here, therefore there is no need to transpose the data whereas in `OrthogonalGrid` there is.

mesa/visualization/mpl_space_drawing.py
Sahil-Chhoker: Makes sense!

Sahil-Chhoker:
Thank you for the review! I have added the suggested changes. Please let me know if you have any additional recommendations to further improve the code quality.

mesa/visualization/mpl_space_drawing.py
```diff
@@ -159,6 +160,37 @@ def draw_space(
     return ax
 
 
+# Helper function for getting the vertices of a hexagon given the center and size
+def _get_hex_vertices(
+    center_x: float, center_y: float, size: float
```
quaquel, line 165: I suggest turning size into a keyword argument with a default value of 1.

mesa/visualization/mpl_space_drawing.py
```diff
@@ -159,6 +160,37 @@ def draw_space(
     return ax
 
 
+# Helper function for getting the vertices of a hexagon given the center and size
+def _get_hex_vertices(
```
quaquel, line 164: perhaps move this to an inner function in _get_hexmesh?

mesa/visualization/mpl_space_drawing.py
```diff
@@ -350,38 +410,15 @@ def setup_hexmesh(width, height):
...
-            x = col * x_spacing + (row % 2 == 0) * (x_spacing / 2)
-            y = row * y_spacing
-
-            vertices = get_hex_vertices(x, y)
-
+        for vertices in _get_hexmesh(width, height, size):
             # Edge logic, connecting each vertex to the next
             for v1, v2 in pairwise([*vertices, vertices[0]]):
-                # Sort vertices to ensure consistent edge representation and avoid duplicates.
+                # Sort vertices to ensure consistent edge representation
```
quaquel, line 418: why do you need to sort them?

Sahil-Chhoker: To avoid duplicate edges like let's say (1, 2) and (2, 1). On sorting both becomes (1, 2), therefore no duplicates!

quaquel:
I think this is ready to be merged. @Sahil-Chhoker do you have any outstanding doubts? If not, I'll merge this later today.

Sahil-Chhoker:
Thanks @quaquel, I think I'm fine.

quaquel (approved):

Sahil-Chhoker:
@quaquel, I ran into a bug while testing this feature again with the `lru_cache` functionality. 
The function call `hexagons = _get_hexmesh(width, height)` won't display the grid. I don't know exactly what might be causing this, but it can be fixed by either of these two options:
1. `hexagons = _get_hexmesh(width, height, size=1.0)`
2. Removing `lru_cache`.

I am very sorry for pushing code without properly testing it.

quaquel:
can you share a minimum example with the bug? I might be able to help diagnose it.

Sahil-Chhoker:
Sorry for the late response, here is my example I was testing the code with:
```python
from mesa.experimental.cell_space.grid import HexGrid
from mesa import Model
import math
from mesa.experimental.cell_space import CellAgent
from mesa.experimental.cell_space.property_layer import PropertyLayer
import numpy as np

class myAgent(CellAgent):
    def __init__(self, model, cell=None):
        super().__init__(model)
        self.cell = cell
    
 
class myModel(Model): 
    def __init__(self, seed=None, width=10,  height=15 , n_agents=4):
        super().__init__(seed=seed) 
        self.width = width
        self.height = height 
        self.num_agents = n_agents
        self.grid = HexGrid( 
            (self.width, self.height), capacity=math.inf, torus=False, random=self.random
        )
 
        self.test_layer = PropertyLayer("test", (self.width, self.height), default_value=1, dtype=int)
        self.test_layer.data = np.random.randint(2, 10, (self.width, self.height))
        self.grid.add_property_layer(self.test_layer)
        myAgent(self, self.grid.all_cells.cells[17])


def portrayal(agent):
    if agent is None:
        return None
    
    return {
        "color": "tab:orange",
        "size": 70,
    }


def post_process(ax):
    ax.set_aspect("equal")
    ax.set_xticks([])
    ax.set_yticks([])
    ax.get_figure().set_size_inches(8, 8) 

property_layer_portrayal = { 
        "test": { 
            "colormap": "Blues",
            "alpha": 0.7,
            "vmin": 0, 
            "vmax": 10,
        }
    }


model_params = { 
    "seed": {
        "type": "InputText",
        "value": 42,
        "label": "Random Seed",
    },
    "n_agents": {
        "type": "SliderInt",
        "value": 4,
        "min": 1,
        "max": 10,
        "step": 1,
        "label": "Number of Agents",
    },
    "width": 6,
    "height": 5,
}

model = myModel()

space_component = make_space_component(
    agent_portrayal=portrayal, draw_grid=True, post_process=post_process, propertylayer_portrayal=property_layer_portrayal
)

page = SolaraViz(
    model,
    components=[space_component],
    model_params=model_params,
    name="Test Model",
)
page
```

Sahil-Chhoker:
@quaquel, While testing the code mentioned above, I discovered another bug: whenever the model's step count is increased, the hex grid simply disappears. After diagnosing the issue, I found that `lru_cache` was once again the culprit. I'm not sure why it’s causing so many problems, but I would recommend removing it from the code entirely.

quaquel:
What happens if you use a normal cache instant of `lru_cache`?

Sahil-Chhoker:
> What happens if you use a normal cache instant of `lru_cache`?

`cache` behaves like `lru_cache`, but it stores results infinitely, unlike `lru_cache`, which deletes the least recently used values based on its `max_size`. Ultimately, what I mean is that the results were the same and buggy.

quaquel:
I think I know what's going on. `_get_hexmesh` is a generator rather than a function.

Sahil-Chhoker:
I thought it was cleaner than storing results in a list and returning them. Should I change it into a function instead?

quaquel:
It depends on what we prefer. In case of a normal function. we can memoize the return allowing us to effectively generate the hexmesh vertices only once.

Sahil-Chhoker:
Now that I have this knowledge that generators don't go well with caching (thanks to you), I think we should change this to a function.

quaquel:
Can you put in a new PR which does this? Otherwise, I'll pick up tomorrow.

PR #2651, opened 2025-01-28
Uses array for hex grid property layer fix

Sahil-Chhoker:
### Summary
`lru_cache` does not work with generators, therefore `_get_hexmesh` is changed to a function.

This was a problem in #2646 that used generators which don't go well along with caching. This PR simply removes the generator and adds a list where all the results are stored.

### DIFF
--- a/mesa/visualization/mpl_space_drawing.py
+++ b/mesa/visualization/mpl_space_drawing.py
@@ -9,7 +9,7 @@ for a paper.
 import contextlib
 import itertools
 import warnings
-from collections.abc import Callable, Iterator
+from collections.abc import Callable
 from functools import lru_cache
 from itertools import pairwise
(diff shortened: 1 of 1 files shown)


Sahil-Chhoker:
@quaquel I think this should do it.

quaquel (approved):
Yes this should do it.

Can you make the PR text a bit more descriptive by linking it to the previous PR? I'll merge later today. 

Thanks for the quick fix!

Sahil-Chhoker:
Jan, I’d also love to hear why caching doesn’t work with generators when you have time.

quaquel:
The cache returns the same generator object. This object is exhausted after the first iteration, so subsequent iterations never happen.

quaquel (approved):

Issue #2580, opened 2024-12-30
Add example of visualizing PropertyLayers to example models


EwoutH:
Currently none of the Example models visualize the PropertyLayers, it would be useful if at least one of them did. This could be [Sugarscape Constant Growback Model with Traders](https://github.com/projectmesa/mesa/tree/main/mesa/examples/advanced/sugarscape_g1mt) for example.

See:
- https://github.com/projectmesa/mesa/pull/2548

quaquel:
I looked at the sugarscape example, and the current portrayal of property layers is not compatible with what would be required for sugarscape. The issue is that currently the entire layer is shown, and you cannot mask part of the layer as done in sugarscape.

EwoutH:
Can you set `alpha` to `0` for certain values you want to hide?

quaquel:
alpha is applied to the entire layer

EwoutH:
It seems I did allow scaling over alpha instead of color in the implementation (https://github.com/projectmesa/mesa/pull/2336): https://github.com/projectmesa/mesa/pull/2336/commits/7c8e98bab81edf14e1d682582ea94bc6a61ef688

> Any [matplotlib color](https://matplotlib.org/stable/api/colors_api.html#module-matplotlib.colors) is valid. The highest value will have that color with 0% opacity, the lowest with 100% opacity (which can again be scaled with modifying vmin and vmax).

EwoutH:
Go ahead! Do you have an particular example in mind?

sanika-n:
I was thinking of sticking with what you said and using "color" with sugarscape

PR #2653, opened 2025-01-29
Added property layer viz to sugarscape

sanika-n:
This PR addresses Issue #2580, I have replaced the normal matplotlib visualization in sugarscape to the property layer visualization with matplotlib as backend.
### Before
Screenshot 2025-01-16 141514

### After 
image

### DIFF
--- a/mesa/examples/advanced/sugarscape_g1mt/app.py
+++ b/mesa/examples/advanced/sugarscape_g1mt/app.py
@@ -8,13 +8,7 @@ def agent_portrayal(agent):
 
 
 propertylayer_portrayal = {
-    "sugar": {
-        "color": "blue",
-        "alpha": 0.8,
-        "colorbar": True,
-        "vmin": 0,
-        "vmax": 10,
-    },
+    "sugar": {"color": "blue", "alpha": 0.8, "colorbar": True, "vmin": 0, "vmax": 10},
     "spice": {"color": "red", "alpha": 0.8, "colorbar": True, "vmin": 0, "vmax": 10},
 }
 
--- a/mesa/examples/advanced/sugarscape_g1mt/app.py
+++ b/mesa/examples/advanced/sugarscape_g1mt/app.py
@@ -1,47 +1,30 @@
-import numpy as np
-import solara
-from matplotlib.figure import Figure
-
 from mesa.examples.advanced.sugarscape_g1mt.model import SugarscapeG1mt
 from mesa.visualization import Slider, SolaraViz, make_plot_component
+from mesa.visualization.components.matplotlib_components import make_mpl_space_component
 
 
-def SpaceDrawer(model):
-    def portray(g):
-        layers = {
-            "trader": {"x": [], "y": [], "c": "tab:red", "marker": "o", "s": 10},
-        }
+def agent_portrayal(agent):
+    return {"marker": "o", "color": "red", "size": 10}
 
-        for agent in g.all_cells.agents:
-            i, j = agent.cell.coordinate
-            layers["trader"]["x"].append(i)
-            layers["trader"]["y"].append(j)
-        return layers
 
-    fig = Figure()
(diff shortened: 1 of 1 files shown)


mesa/examples/advanced/sugarscape_g1mt/app.py
```diff
@@ -1,48 +1,31 @@
...
-    return solara.FigureMatplotlib(fig)
+propertylayer_portrayal = {
+    "sugar": {
+        "color": "blue",
+        "alpha": 0.8,
+        "colorbar": True,
+        "vmin": 0,
+        "vmax": 10,
+    },
+    "spice": {"color": "red", "alpha": 0.8, "colorbar": True, "vmin": 0, "vmax": 10},
```
tpike3, line 18: @sanika-n The formatting here seems off --- are you using the[ `ruff` linter](https://github.com/projectmesa/mesa/blob/main/CONTRIBUTING.md#testing-and-code-standards)

tpike3 (commented):
Thanks @sanika-n -- this code shows the power of property layer. 

Just one question as the formatting does not seem to be uniform.

mesa/examples/advanced/sugarscape_g1mt/app.py
sanika-n: Oh, I actually wrote spice in the same way as sugar but when I ran ruff it was changing the formatting of spice and it didn't allow me to keep it like sugar, so now I have made sugar formatted like spice...

quaquel:
I had a look at sugarscape a few weeks back as well. It made me realize that the property layer portrayal needs some more work at the API level (see also #2640). However, in the meantime, this looks fine to me. However, could you experiment a bit with colormaps that are more distinct that the current red and blue? Likewise, what happens if you lower the alpha to say 0.5?

sanika-n:
So here are some plots with different values of alpha, the reason why there is not much opacity even alpha=1 is that I set vmax to 10, and the sugar and spice levels in this example only go up to 4. 
image

Given that the first property layer is sugar and the second is spice, in areas where both spice and sugar equal 10 we will only see a bright red because both colors are completely opaque (alpha=1 and value=vmax) and therefore spice will cover up sugar. I set vmax to 10 to avoid this. However, this case is rare, and I noticed it doesn't come up much in this example, and setting vmax to 4 made the plot more vibrant.
image


Coming to colormaps, I tried a few configurations, but for most of them, I wasn't too happy with how it turned out. The colors were mixing quite a lot with each other.

Example 1:
```python
propertylayer_portrayal = {
    "sugar": {"colormap": "winter", "alpha": 1, "colorbar": True, "vmin": 0, "vmax": 4},
    "spice": {"colormap": "afmhot", "alpha": 0.5, "colorbar": True, "vmin": 0, "vmax": 10},
}
```
image


Example 2
```Python
propertylayer_portrayal = {
    "sugar": {"colormap": "winter", "alpha": 1, "colorbar": True, "vmin": 0, "vmax": 4},
    "spice": {"colormap": "Reds", "alpha": 0.4, "colorbar": True, "vmin": 0, "vmax": 10},
}
```
image


Example 3:
```python
propertylayer_portrayal = {
    "sugar": {"colormap": "viridis", "alpha": 0.5, "colorbar": True, "vmin": 0, "vmax": 4},
    "spice": {"colormap": "plasma", "alpha": 0.5, "colorbar": True, "vmin": 0, "vmax": 4},
}
```
image

quaquel:
Thanks for this. I'll think about this a bit more (also in light of the various other visualization and API issues that are popping up). I am inclined to accept this PR while we hash out a better overall API level solution.

tpike3:
> Thanks for this. I'll think about this a bit more (also in light of the various other visualization and API issues that are popping up). I am inclined to accept this PR while we hash out a better overall API level solution.

I concur @quaquel

quaquel:
Thanks!

Issue #2652, opened 2025-01-28
Tutorial environment Binder


bertheling:
It is no longer possible to test the tutorials from Binder. An error occurs on the first script. This might be due to the fact that an old version of MESA is used in Binder (3.0.0b1).

PR #2655, opened 2025-01-29
Update Binder environment to use latest Mesa version (#2652)

aarav-shukla07:
### Summary

The Binder environment for Mesa tutorials is currently broken because it uses an outdated version of Mesa (3.0.0b1). This causes errors when running the tutorials, preventing users from testing Mesa examples on Binder. The issue impacts new contributors and learners who rely on Binder to explore Mesa without setting up a local environment.

### Bug / Issue

- what is the bug:
       The Binder environment for Mesa tutorials is failing because it uses an outdated version of Mesa (3.0.0b1). This 
       causes errors when running the tutorials.
- why does it happen:
       The environment.yml file in the binder directory specifies an old version of Mesa (3.0.0b1), which is no longer 
       compatible with the latest dependencies or tutorial code.
- Impact of the bug:
       - Users cannot test Mesa tutorials on Binder, which limits accessibility and usability for new contributors and learners.
       - The broken environment prevents users from exploring Mesa’s features and examples without setting up a local development environment.

### Implementation

- Updated the environment.yml file to use the latest version of Mesa.

- Simplified the dependencies to reduce build time and avoid potential conflicts.

- Verified that the Binder environment now works by testing the tutorials locally and on Binder.

### Testing

Steps via which testing can be done :
- Go to https://mybinder.org/.

- Enter the URL of the Mesa repository and click Launch.

- Once the environment is built, navigate to the examples directory.

- Open and run the introtutorial.ipynb tutorial.

- Confirm that the tutorials run without errors and produce the expected output.

### DIFF
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -2,7 +2,7 @@ name: mesa-tutorials
 channels:
   - conda-forge
 dependencies:
-  - python=3.9
+  - python
   - numpy
   - pip
   - pip:
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -1,8 +1,8 @@
-name: example-environment
+name: mesa-tutorials
 channels:
   - conda-forge
 dependencies:
-  - python
+  - python=3.9
   - numpy
   - pip
   - pip:


binder/environment.yml
```diff
@@ -1,8 +1,8 @@
-name: example-environment
+name: mesa-tutorials
 channels:
   - conda-forge
 dependencies:
-  - python
+  - python=3.9
```
tpike3, line 5: Can you change this to just `python` -- we actually require [>=3.11](https://github.com/projectmesa/mesa/blob/c58af091aa74959c8b438adf1f2a915e035b60c2/pyproject.toml#L9) 

However if we pin it to a python version, then it is one more thing to update

tpike3 (changes requested):
Thanks for catching this @aarav-shukla07 !

I had one quick request

binder/environment.yml
aarav-shukla07: Hi @tpike3 ,
Thanks for the feedback! I’ve updated the environment.yml file to remove the pinned Python version and ensure compatibility with Python >=3.11. The environment has been tested locally and on Binder, and the tutorials are working as expected.
Please let me know if there’s anything else I need to address.

tpike3 (approved):

PR #2668, opened 2025-02-07
fixing the draw_orthogonal_grid function

sanika-n:
## Bug
I noticed that using make_mpl_space_component to create a plot, always created weird patterns in the background when the model used orthogonal grids. The red dots are agents, but the gray patterns were actually the rendition of the orthogonal grid by the draw_orthogonal_grid function. 

### Before
image

The logic seems to be correct, but the grid is being drawn using ":", this makes it easy to confuse between the horizontal and the vertical lines causing visually confusing renditions. To resolve this, I just changed ":" to "-" in the code.

### After
image
 
To replicate an issue of this kind you could change draw_grid to True in [sugarspace](https://github.com/projectmesa/mesa/blob/7c778609e08247829ea86d8310659e32a353b5ef/mesa/examples/advanced/sugarscape_g1mt/app.py#L20C4-L20C21)

### DIFF
--- a/mesa/visualization/mpl_space_drawing.py
+++ b/mesa/visualization/mpl_space_drawing.py
@@ -350,9 +350,9 @@ def draw_orthogonal_grid(
     if draw_grid:
         # Draw grid lines
         for x in np.arange(-0.5, space.width - 0.5, 1):
-            ax.axvline(x, color="gray", linestyle=":")
+            ax.axvline(x, color="gray", linestyle="-")
         for y in np.arange(-0.5, space.height - 0.5, 1):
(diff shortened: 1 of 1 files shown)


quaquel:
This is related to #2438 and, more broadly, to #2640 and #2642. So rather than shift from one predetermined style to another, I favor changing the API so that the user can explicitly control this (but with meaningful defaults).

sanika-n:
So do you recommend that I add the feature to mention the type of grid, or should I just close this pull request?

quaquel:
I suggest describing this issue in #2642 and then closing this PR as the best way forward.

Issue #1150, opened 2022-01-30
Rotate Shape based on Heading


aiwhoo:
**What's the problem this feature will solve?**
Relay visual feedback of heading for debugging and verification.

**Describe the solution you'd like**
It'd be nice to be able to load in an image of the agent and rotate it based on the heading. For example, in the continous space, if a boid flocker is moving at a 45 degree angle, instead of the red dot, an image of a bird could be loaded which is facing at the 45 degree angle.

**Additional context**
This is primarily working with the boid flocker example in the mesa/examples directory.

EwoutH:
Yes, I strongly agree this should be implemented. Thanks for reporting!

On my personal list is first finishing the big project of implementing different agent types in Mesa (called [breeds](http://ccl.northwestern.edu/netlogo/docs/dict/breed.html) in NetLogo) (#1142), but after that I will try to take a look into this.

If anyone else is up to speed with this part of the code and wants to implement this, please do so, since I'm not.

aiwhoo:
@EwoutH there was an individual that just recently shared a video Showcasing his model which happened to have this. Mesa admin should view his code and consider integration. His video demonstrates core agent based modeling functionality that should be within mesa.

https://youtu.be/pOF1H84xPik

He stated he’ll upload to GitHub by end of week.

EwoutH:
That looks awesome! I would love to see the source code, it might even be useful for the Example library: https://github.com/projectmesa/mesa/tree/main/examples

aiwhoo:
@EwoutH reach out to him to show Interest! That will encourage him to share the code.

The main issue with the mesa/examples is that it lacks examples. The flocker example is the only one using the continuous space, but it doesn’t encompass the other mesa coding styles/features.  Anything added to the examples would be a win in my book!

aiwhoo:
@EwoutH The author posted his github link: https://github.com/ValerioB88/EvoAgent/tree/master!!!!

PR #2670, opened 2025-02-08
Fixed Marker in mpl

sanika-n:
### Bug
In agent portrayal, the marker is supposed to be able to take on all matplotlib [marker types](https://matplotlib.org/stable/api/markers_api.html#module-matplotlib.markers). But the tuple way of defining markers is currently not compatible in MESA. This issue can be replicated by replacing the marker in any one of the examples (here I have taken boid_flockers) with say (3, 0, 30). This is supposed to convert the agents to a 3 sided polygon tilted by 30 degrees, if we replaced 3 with 4 we will get a square instead. 
The main reason for the error was that during agent data collection, numpy was turning the tuples into arrays. I have fixed the issue in this PR.

I just ran boid_flockers using tuples in the following code after the fix:
```python
def boid_draw(agent):
    neighbors = len(agent.neighbors)

    if neighbors <= 1:
        return {"color": "red", "size": 20, "marker": "o"}
    elif neighbors >= 2:
        return {"color": "green", "size": 20, "marker": (3, 0, 60)}
```
image

I am not exactly sure, but I was wondering if merging this fix will help us later implement a solution to #1150, I can try working on that for boid_draw if it is feasible...


mesa/visualization/mpl_space_drawing.py
```diff
@@ -638,8 +646,8 @@ def _scatter(ax: Axes, arguments, **kwargs):
                     f"{entry} is specified in agent portrayal and via plotting kwargs, you can only use one or the other"
                 )
 
-    for mark in np.unique(marker):
-        mark_mask = marker == mark
+    for mark in set(marker):
```
sanika-n, line 649: I changed it to a set here because np.unique only works on similar datatypes as far as I know. And if we have a mix of tuples and the normal strings, np.unique wont work

mesa/examples/basic/boid_flockers/app.py
quaquel, line 1: This is a good way to test hybrid markers. Nice!

quaquel (approved):

Issue #2666, opened 2025-02-06
Requested Edits from JOSS


jackiekazil:
A review with requested edits have come in from JOSS. The checklist is below. Let's use this ticket for tracking.

References:
* Proof article for line number reference: https://github.com/openjournals/joss-papers/blob/joss.07668/joss.07668/10.21105.joss.07668.pdf
* Requested edits with full comment: https://github.com/openjournals/joss-reviews/issues/7668#issuecomment-2639856634
* Parent ticket: https://github.com/projectmesa/mesa/issues/2559

--------------

### Major comments

  - [x] Experimental features: cell spaces and event-based scheduling (lines 76 & 86) (#2675)
    - I believe these features should be completed before publication. Publishing a work in progress is not ideal, particularly since this is not the first paper on this package and these features appear to be close to a stable release.
    - Additionally, when I attempted to test these new features - which I find to be very interesting innovations - I found the online documentation lacking sufficient information.
    - If completing these features before publication is not feasible, I recommend moving them to a dedicated section such as Future Plans or Next Steps.

### Comments on the documentation (@EwoutH)

  - [x] On the index page, I would add a sentence explaining what the "rec" and "all" extensions are needed for respectively (e.g. the visualization package?). #2672

  - [x] I would suggest to put the "Overview of the MESA library" section within "Getting started" as a seperate page, as it is a handy reference that is more accessible than the API documentation, also after getting started. #2673

  - [x] In addition, it would be great for the overview to cover all features of MESA. Compared to the paper, spaces and time advancement are missing here. #2673

  - [x] **[from Marti]** In the "Mesa core examples" section, there's confusion between referring to the mesa-examples repository while code snippets correspond to files in https://github.com/projectmesa/mesa/tree/main/mesa/examples. It would be better to put everything in one place and mention in the documentation that users can clone it and go to each example's folder instead of having them copy each snippet in a separate file. #2706

  - [x] **[from Marti]** It may be a good idea to explicitly mention the `number_processes` argument of `mesa.batch_run` in the [Batchrunner API](https://mesa.readthedocs.io/stable/apis/batchrunner.html) page. The documentation should also explain why it's "strongly advised to only run in parallel using a normal python file (so don't try to do it in a jupyter notebook)". #2707

### Comments on the introductory tutorial (@tpike3)
[...]

PR #2673, opened 2025-02-09
docs: Split off the overview page, extend with spaces/activation

EwoutH:
Based on feedback from the paper, this splits the overview (again) off to a separate page. It does add one more item to the top bar, but I think that's worth it.

This PR consists of two commits:
- a2e38b9 to split off the overview to a separate page 
- f59fa99 to add space and time advancement sections to overview

Check out the new overview page rendered on Readthedocs: https://mesa--2673.org.readthedocs.build/2673/overview.html

Part of #2666 based on feedback of @jofmi.


quaquel (approved):
Fine with me. 

At what point do we remove the migration guide from the top bar?

PR #2674, opened 2025-02-09
Feat: Added command center

Sahil-Chhoker:
### Summary
This feature is inspired by [this comment](https://github.com/projectmesa/mesa/issues/2176#issuecomment-2223093927) from issue #2176.

This pull request aims to add a command center into Solara Viz to execute commands directly from the browser to your model.

### Implementation
This PR adds a solara component named `CommandCenter` into the sidebar directly. Runs commands written in the text area directly in the global scope using `exec` function and updates the visualization to show the changes.
image

Every function and property available through model can be used and updated using the command center. Some examples:

1. ```model.step()```
2. ```print(len(model.agents))```
3. Something like this can also work:
```python 
  import random
  
  agents = model.agents_by_type[ContinuousSpaceAgent]
  
  for agent in agents:
     agent.position = [random.random(), random.random()]
```

**For custom imports:**
```python
from agents import myAgent

page = SolaraViz(
    model,
    components=[space_component],
    model_params=model_params,
    name="Test Model",
    additional_imports={
        "myAgent": myAgent, # for custom classes
        "np" : "import numpy as np", # for strings
    }
)
page
```

now both `np` and `myAgent` can be used in the command center.


Sahil-Chhoker:
Some known limitations:
1. Ruff error [S102](https://docs.astral.sh/ruff/rules/exec-builtin/).
2. Ugly spellcheck:
image

Sahil-Chhoker:
@EwoutH Since you were the one who recommended this feature over on issue #2176, I would love to hear your thoughts on this.

EwoutH:
This is super cool. I'm in a long distance train this afternoon, I will try to look at it then!

quaquel:
I think this is cool. However, the unchecked use of exec is a massive security risk. Just try `import shutil; shutil.rmtree("/"),`.....

I don't have time to dig a bit deeper, but e.g., https://www.stackhawk.com/blog/command-injection-python/ might be a good starting point to think about this a bit more.

EwoutH:
Really like the idea so far!

There are a few minor things. For example, to make it practical, it should probably import at least the Agent types used (in this example Wolf, Sheep, etc.)

image

quaquel:
I suggest also checking the [code](https://docs.python.org/3/library/code.html) module, which allows you to attach a full console to a running python process.

Sahil-Chhoker:
I did some research on making `exec()` safer, and with a little help from AI, I came up with this final implementation. It throws an error for dangerous commands. Also, since the server runs on the client's machine, it's unlikely they'd try to sabotage themselves.

Sahil-Chhoker:
> I suggest also checking the [code](https://docs.python.org/3/library/code.html) module, which allows you to attach a full console to a running python process.

The `code` module could be useful for turning the command center into a command console. I also tried building it separately, but with my current Solara skills, it looks quite ugly.

quaquel:
> The code module could be useful for turning the command center into a command console. I also tried building it separately, but with my current Solara skills, it looks quite ugly.

Thanks for checking! Might it be an idea to open a separate discussion or issue where you share what you did? One dream feature would be to have a full variable explorer and console when running mesa models through the GUI. This PR is a stepping stone towards that vision and anything you learned can be valuable for others. I'll try to review this PR more closely hopefully over the weekend.

Sahil-Chhoker:
>  One dream feature would be to have a full variable explorer and console when running mesa models through the GUI.

That sounds like a great feature! I have opened an issue (#2683) for discussions. I decided to start from scratch to make it easier to incorporate valuable input. I would greatly appreciate any help you can provide!

mesa/visualization/solara_viz.py
```diff
@@ -157,6 +164,8 @@ def SolaraViz(
             )
         with solara.Card("Information"):
             ShowSteps(model.value)
+        with solara.Card("Command Center"):
```
quaquel, line 167: do we want the command center to be allways there, or make it optional in some way?

mesa/visualization/solara_viz.py
```diff
@@ -586,3 +595,156 @@ def ShowSteps(model):
     """Display the current step of the model."""
     update_counter.get()
     return solara.Text(f"Step: {model.steps}")
+
+
+class _SecureCodeValidator(ast.NodeVisitor):
```
quaquel, line 600: what did you use as inspiration, if any for this?

quaquel (commented):
Should we not move all command center stuff into its own seperate module?

A few clarifying questions to start this review.

Sahil-Chhoker:
> Should we not move all command center stuff into its own separate module?

I realized the same after implementing it, but I think the development of console is going smooth, that can replace the command center entirely. We can have an separate file for that.

mesa/visualization/solara_viz.py
Sahil-Chhoker: I personally would prefer it to be always be there, it will encourage the users to play around with it and know it better.

mesa/visualization/solara_viz.py
Sahil-Chhoker: I didn't find any source exactly resembling this, but after search a bit I found some open discussions on stackoverflow suggesting the use of `ast`, after doing a little research over it, I used AI to assemble those ideas into a prototype.

mesa/visualization/solara_viz.py
quaquel: From a design point of view, I prefer giving users the choice but show it in all our examples.

Someone might want to built a UI with very limited possibility for interaction, or only show the animation to a client.

Sahil-Chhoker: Makes sense, does a boolean parameter in `SolaraViz` work?

quaquel: I was thinking more along the lines of treating it as a component so you could just at it to the list of components:

```python
page = SolaraViz(
    model,
    components=[SpaceComponent, SomeGraphComponent, Console],
    model_params=model_params,
    name="my fancy model",
)

```

Sahil-Chhoker: Yeah, that works as well, in fact that is more intuitive.

Corvince:
I just want to say that I am really impressed with the work done here and in #2683 . Definitely something that has been on my frontend wishlist for a long time. But I am not sure how to proceed from here, is this PR here superseded by the work done in #2683? Or should this here be merged before implementing the changes there? Or is the scope of both features different?

Sahil-Chhoker:
Thanks @Corvince for your reply!

> But I am not sure how to proceed from here, is this PR here superseded by the work done in https://github.com/projectmesa/mesa/issues/2683? 

Yes the work done in #2683 will replace the work done by this PR. This PR is here to get the review on the the feature `_SecureCodeValidator`, because it is going to be reused in the work in #2683. I would really appreciate if you could give some useful insights on that.

mesa/visualization/solara_viz.py
```diff
@@ -586,3 +595,156 @@ def ShowSteps(model):
...
+    """Validates Python code for potentially dangerous operations."""
+
+    FORBIDDEN_BUILTINS = {
+        "eval",
+        "exec",
+        "compile",  # Code execution
+        "open",  # File system access
+        "globals",
+        "locals",  # Global state manipulation
+    }
```
quaquel, line 610: most of these make sense to me

is the model itself by default available in the namespace?

also, we can refine this based on experiences

mesa/visualization/solara_viz.py
```diff
@@ -586,3 +595,156 @@ def ShowSteps(model):
...
+
+    FORBIDDEN_ATTRIBUTES = {"__code__", "__globals__", "__builtins__"}
+
+    def __init__(self):
+        self.errors = []
+
+    def visit_Call(self, node):
```
quaquel, line 617:
```suggestion
    def visit_call(self, node):
```

mesa/visualization/solara_viz.py
```diff
@@ -586,3 +595,156 @@ def ShowSteps(model):
...
+        elif (
+            isinstance(node.func, ast.Attribute)
+            and node.func.attr in self.FORBIDDEN_ATTRIBUTES
+        ):
+            self.errors.append(
+                f"Access to special attribute '{node.func.attr}' is not allowed for security reasons"
+            )
+        self.generic_visit(node)
+
+    def visit_Attribute(self, node):
```
quaquel, line 633:
```suggestion
    def visit_attribute(self, node):
```

mesa/visualization/solara_viz.py
```diff
@@ -586,3 +595,156 @@ def ShowSteps(model):
...
+        ]
+
+        for attr in attrs:
+            for pattern in dangerous_patterns:
+                if re.match(pattern, attr):
+                    self.errors.append(
+                        f"Access to system module '{attr}' is not allowed"
+                    )
+
+        self.generic_visit(node)
```
quaquel, line 651: what is the overhead of this security validation?

quaquel:
Although I am sympathetic to @Corvince's points, I don't fully agree. One use case is that I am giving a model with a UI to some novice student. She will run the model locally, but might not be intimately familiar with python or even with programming. In such a context, a minimum level of protection of what can be done via the console seems desirable.

Corvince:
Good point, you are right, I haven't considered the use-case of someone who isn't a model developer, but is playing with a model locally. But I am not sure how much of a risk that is. Its not *that* easy to cause serious harm. If you are worried about that you shouldn't let users use the computer unsupervised ;)

mesa/visualization/solara_viz.py
Sahil-Chhoker: I ran some tests, and most of the time, the overhead from the security validator is more than double the execution time of the code.

mesa/visualization/solara_viz.py
Sahil-Chhoker: Sorry, I don't think I exactly understand what you mean here. Can you add a little more context.

mesa/visualization/solara_viz.py
quaquel: that is a lot, it might be an argument to leave out validation at all and just give a very clear warning in the docs about security.

Sahil-Chhoker: Yeah it's a lot, I didn't think it to be so performance heavy. But since it could be used for smaller models, I think a warning in the docs should do the job (personal view).

EwoutH:
Can I help move this PR forward?

Sahil-Chhoker:
Thanks, @EwoutH! I would love to hear your input on the `SecureCodeValidator` as well.

EwoutH:
I think the bare minimum could be done to make sure users don’t terminate their PC with one wrong command, but not more than that. I agree with Corvince there isn’t a convincing attack angle and that’s not easy to do serious harm.

quaquel:
I assume this is superseded by #2697 and so can be closed?

Sahil-Chhoker:
Yeah this PR can be closed!

PR #2681, opened 2025-02-11
Move Game of life to cell space

quaquel:
This moves game of life over to use cell spaces.


EwoutH:
Game of Life isn’t in the benchmark models. Why did the performance benchmark fail in this PR?

quaquel:
> Game of Life isn’t in the benchmark models. Why did the performance benchmark fail in this PR?

Because of the `__init__.py` issue fixed in #2687. As you can see, it now passes.

EwoutH (approved):

PR #2684, opened 2025-02-12
Move Schelling to discrete space

quaquel:
This moves the schelling model over to using the new stable cell spaces.


EwoutH (approved):
I really like the new API, thanks for updating this example!

PR #2688, opened 2025-02-13
Move virus_on_a_network to discrete_space

quaquel:
This moves virus_on_a_network to discrete_space and cleans up the code.

### DIFF
--- a/mesa/examples/basic/virus_on_network/agents.py
+++ b/mesa/examples/basic/virus_on_network/agents.py
@@ -1,6 +1,6 @@
 from enum import Enum
 
-from mesa import Agent
+from mesa.discrete_space import FixedAgent
 
 
 class State(Enum):
@@ -9,7 +9,7 @@ class State(Enum):
     RESISTANT = 2
 
 
-class VirusAgent(Agent):
+class VirusAgent(FixedAgent):
     """Individual Agent definition and its properties/interaction methods."""
 
     def __init__(
@@ -20,6 +20,7 @@ class VirusAgent(Agent):
         virus_check_frequency,
         recovery_chance,
         gain_resistance_chance,
+        cell
     ):
         super().__init__(model)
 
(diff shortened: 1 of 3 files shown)


mesa/examples/basic/virus_on_network/model.py
```diff
@@ -60,23 +56,23 @@ def __init__(
...
+            list(self.grid.all_cells),
+        )
 
         # Infect some nodes
-        infected_nodes = self.random.sample(list(self.G), self.initial_outbreak_size)
-        for a in self.grid.get_cell_list_contents(infected_nodes):
+        infected_nodes = CellCollection(
+            self.random.sample(list(self.grid.all_cells), self.initial_outbreak_size),
+            random=self.random,
+        )
```
EwoutH, line 74: Non blocking, but might be interesting if we can make this API-wise a bit cleaner.

mesa/examples/basic/virus_on_network/agents.py
```diff
@@ -29,19 +30,14 @@ def __init__(
...
-            for agent in self.model.grid.get_cell_list_contents(neighbors_nodes)
-            if agent.state is State.SUSCEPTIBLE
-        ]
-        for a in susceptible_neighbors:
-            if self.random.random() < self.virus_spread_chance:
-                a.state = State.INFECTED
+        for agent in self.cell.neighborhood.agents:
+            if (agent.state is State.SUSCEPTIBLE) and (
+                self.random.random() < self.virus_spread_chance
+            ):
```
EwoutH, line 39: Can we use `agents.select()` for this?

EwoutH (approved):
Looks good, one potential improvement comment and one potential follow-up.

mesa/examples/basic/virus_on_network/agents.py
quaquel: No. `cell.neighborhood.agents` is a list, not an agentset. I have tested changing this in discrete_space to agentset and it results in massive performance loss.

mesa/examples/basic/virus_on_network/model.py
quaquel: Not sure what you are suggesting. I could write the example to not use a cell collection, but this kept the code closer to the original code and illustrates how users can create their own CellCollection.

PR #2689, opened 2025-02-14
added dark mode

sanika-n:
## This PR adds the Solara Dark theme to Mesa

By default, the theme is set to match the user's system settings, however, it can be controlled by clicking on the toggle at the right end of the app bar once solara is launched

### Dark Mode (Auto for me)
image

### Light Mode
image


EwoutH (approved):
Easy, nice and uncomplicated. I like it!

I will leave the merge for another maintainer.

PR #2690, opened 2025-02-14
Series of minor paper updates

jackiekazil:
RE: #2666 

Fixes 

- [x] l34 What does pure python mean? Does it not also use C, JS/React, etc.? (@jackiekazil)
- [x] l96 Please add a link or reference regarding the Solara package (@jackiekazil)
- [x] l143 Maybe say "want" instead of "need"? (@jackiekazil)
- [x] Citation error, noticed in passing.


EwoutH (approved):
Changes looks good to me, thanks!

jofmi (approved):

PR #2692, opened 2025-02-15
fix multiple citations in joss paper

wang-boyu:
Part of https://github.com/projectmesa/mesa/issues/2666.

Apparently it needs a `;` not a `,` when there are more than one citations.


EwoutH (approved):
Thanks for catching this!

Issue #2691, opened 2025-02-15
Network Visualization error when Agent Portrayal has inconsistent fields


n1starck:
**Describe the bug**
When trying to visualize a network-based model, the draw_network function raises an error. I was working on making the error reproducible with the Virus on Network example code and noticed that the Stable and Latest versions use two different methods of creating the Network. The latest version seemed to tolerate visualizing the model with more agents than nodes (assuming the cell capacity was large enough). After adapting my code it visualizes the model with some agents greater than the number of nodes but fails when other agents are introduced, so it does not seem like it is strictly a function of the number of agents.

![Image](https://github.com/user-attachments/assets/dcdbde24-2701-46cb-8071-abce3c11be91)

**Expected behavior**
The draw_network code uses the agent_portrayal function to visualize the various type and number of agents in each node of the network. 

**To Reproduce**
 I haven't been able to reproduce the error with the VoN example. 

**Additional context**
I am using Mesa Version 3.1.0 in JupyterLab on MacOS Sequoia

<hr>

_Updated description - please check_

### Description
When trying to visualize a network-based model with the `draw_network` function, I encountered an error. Initially, I thought it was related to having more agents than nodes, but through investigation with the help of @quaquel, we discovered that the actual issue was in the agent portrayal function.

The visualization fails when one agent type has fields in its portrayal that other agent types don't have. In my specific case, I was using `"edgecolors": "black"` for one agent type (OCO) but not for others, which caused the error.

### To Reproduce
1. Create a network model with multiple agent types
2. Create an agent portrayal function where one agent type returns a different set of visual properties than others
3. Try to visualize the model using `draw_network`

### Expected Behavior
Either:
1. The visualization should handle inconsistent portrayal fields gracefully, or
2. There should be clearer error messages indicating that all agent types must have the same portrayal fields

### Solution
The issue is resolved by ensuring that all agent types return the same set of fields in the agent portrayal function, even if the values for those fields differ.

### Additional Context
This issue affects both the stable and development versions of Mesa. I was using Mesa Version 3.1.0 in JupyterLab on MacOS Sequoia.

quaquel:
The latest version of the model uses the new style `mesa.discrete_space.Network`. I am surprised about the difference in behavior between this and the old style `mesa.space.NetworkGrid`. 

The offending code is in `mesa.visualization.mpl_space_drawing.draw_network`

```python
    # gather agent data
    s_default = (180 / max(width, height)) ** 2
[...]

PR #2693, opened 2025-02-15
Fix draw_network: Handle multiple agents per node correctly (#2691)

aarav-shukla07:
### Summary

The draw_network() function previously failed when there were more agents than nodes in a network visualization. This PR fixes the issue by properly handling multiple agents at the same node, ensuring they are displayed correctly without overlapping.

### Bug / Issue

Linked Issue: [#2691](https://github.com/projectmesa/mesa/issues/2691)
Problem:
- The draw_network() function assumed a 1:1 mapping of agents to nodes, causing errors when multiple agents occupied a single node.
- Indexing errors (IndexError) occurred due to incorrect NumPy array handling, leading to crashes in _scatter().
- The function overlapped agents at the same node, making visualization unreadable.

Expected Behavior:
- The function should allow multiple agents per node and display them without overlap.
- The visualization should not crash when draw_network() is called on a network with more agents than nodes.

Actual Behavior (Before Fix):
- ❌ IndexError due to NumPy array misalignment.
- ❌ Agents stacked on top of each other, making visualization unusable.
- ❌ _scatter() failed when no agents were present.

### Implementation

The fix introduces the following changes:
(1) Grouped agents per node manually
- Instead of using missing methods (get_all_cell_contents()), we now directly track agents using space.agents.

(2)  Adjusted agent positions to prevent overlap
- Used a circular spread algorithm (cos & sin functions) to position agents around their node.

(3)  Fixed IndexError by ensuring arguments["loc"] is always a 2D NumPy array
- Used .reshape(-1, 2) to maintain (N,2) format, avoiding NumPy misalignment errors.

(4)  Prevented _scatter() from failing on empty agent lists
- Checked if arguments["loc"].shape[0] > 0 before calling _scatter().

Code Changes:
- Modified mpl_space_drawing.py to correctly handle agent visualization for network models.
- Ensured agents are evenly distributed around their node rather than overlapping.

### Testing

(1)  Ran all tests in test_components_matplotlib.py
- Result: 7 passed, 1167 warnings (Warnings unrelated to this PR).

(2) Manually tested visualization with network models containing more agents than nodes
- Before Fix: ❌ Overlapping agents, crashes when no agents.
- After Fix: ✅ Agents spread properly, no crashes.

### Additional Notes
- This fix is backward-compatible and does not affect any other functionality in Mesa.
- No new dependencies were added.
- The fix ensures that future changes to draw_network() will properly support multiple agents per node.


mesa/visualization/mpl_space_drawing.py
```diff
@@ -483,17 +483,41 @@ def draw_network(
...
+    node_to_agents = {}
+    for agent in space.agents:  # ✅ Directly access the agent list
+        if hasattr(agent, "pos"):  # Ensure agent has a position
+            if agent.pos not in node_to_agents:
+                node_to_agents[agent.pos] = []
+            node_to_agents[agent.pos].append(agent)
+
+    # Adjust positions for multiple agents per node
+    adjusted_positions = []
+    for node in space.G.nodes:
```
quaquel, line 495: This does not work for `mesa.discrete_space.Network`

aarav-shukla07:
hi @quaquel 
Thanks for your review!
I have updated the code to ensure compatibility with mesa.discrete_space.Network by properly handling agent positions and preventing KeyError: None. Now, all 7 tests are passing. Let me know if further improvements are needed!

mesa/visualization/mpl_space_drawing.py
aarav-shukla07: hi @quaquel
Thanks for your review!
I have updated the code to ensure compatibility with mesa.discrete_space.Network by properly handling agent positions and preventing KeyError: None. Now, all 7 tests are passing. Let me know if further improvements are needed!

EwoutH:
@quaquel are you able to further review this PR?

quaquel:
I am skeptical about this PR.  It started from #2691, but that contained some misdiagnosis of what is going wrong. See the last 2 responsens there. 

So, what this PR leaves is a way of visualizing multiple agents that occopy the same cell by drawing them in a circle arround the cell. I am doubtfull that this is a good overall default solution.

Sahil-Chhoker:
I personally believe that visualizing multiple agents by alternating their alpha and size is a better approach than arranging them differently.

tpike3:
@aarav-shukla07 Thank you for doing this as it did result in some good conversations. Although based on the discussion above we are not going to merge your contribution in Mesa. 

The value of having  different ways of visualizing agents is always useful. Another way to do this may be to create a new example or variant of an existing example [Mesa examples](https://github.com/projectmesa/mesa-examples) that shows users different ways to represent agents in the same cell.

tpike3:
Closing

PR #2695, opened 2025-02-17
Paper: Fix Statement of Need; Acknowledgements, and Applications.

jackiekazil:
This ticket addresses the following in #2666 -- 

- [ ] l16 The Statement of Need could be more clearly structured. Currently, it reads more like a historical recap. I recommend organizing it along these lines: (1) Why such a package is needed, (2) The current state of the field, and (3) What MESA contributes — focusing on its present state rather than its 2014 origins. Additionally, please refer to the Statement of Need and State of the Field requirements outlined in the JOSS checklist above to ensure alignment with the journal's expectations. (@jackiekazil)
- [ ] l109 I would suggest to move the "Applications" section to the start of the paper (@jackiekazil)
- [ ] l134 Acknowledgements are usually placed after the conclusions (@jackiekazil)

### Software paper

- [ ] A statement of need: Does the paper have a section titled 'Statement of need' that clearly states what problems the software is designed to solve, who the target audience is, and its relation to other work?
- [ ] State of the field: Do the authors describe how this software compares to other commonly-used packages?


jackiekazil:
@jofmi - RE: statement of need -- let me know if I missed any of the requirements or suggestions.

jofmi (approved):
looks good! :)

PR #2696, opened 2025-02-18
Update boid_flockers example include flight direction

sanika-n:
I updated boid_flockers so that the direction of movement of agents is shown in the simulation.
I did this by changing the marker to caret-down (a triangle with a larger base so that the direction of motion is not ambiguous) and then rotating the marker according to the direction of motion of the agent. Here are some images of boid_flockers after this update:
image

image

### DIFF
--- a/mesa/examples/basic/boid_flockers/agents.py
+++ b/mesa/examples/basic/boid_flockers/agents.py
@@ -58,7 +58,7 @@ class Boid(ContinuousSpaceAgent):
         self.separate_factor = separate
         self.match_factor = match
         self.neighbors = []
-        self.angle = 0.0
+        self.angle = 0.0  # represents the angle at which the boid is moving
 
     def step(self):
         """Get the Boid's neighbors, compute the new vector, and move accordingly."""
--- a/mesa/examples/basic/boid_flockers/agents.py
+++ b/mesa/examples/basic/boid_flockers/agents.py
@@ -58,6 +58,7 @@ class Boid(ContinuousSpaceAgent):
         self.separate_factor = separate
         self.match_factor = match
         self.neighbors = []
+        self.angle = 0.0
 
     def step(self):
         """Get the Boid's neighbors, compute the new vector, and move accordingly."""
--- a/mesa/examples/basic/boid_flockers/app.py
+++ b/mesa/examples/basic/boid_flockers/app.py
@@ -1,6 +1,9 @@
 import os
 import sys
 
+import numpy as np
+from matplotlib.markers import MarkerStyle
+
 sys.path.insert(0, os.path.abspath("../../../.."))
 
 from mesa.examples.basic.boid_flockers.model import BoidFlockers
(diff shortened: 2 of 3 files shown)


EwoutH (commented):
Awesome, this is really useful. Code itself looks good!

Why don't the non-blocking boids have an direction?

Have you any idea on the performance impact?

Sahil-Chhoker:
Really sorry @EwoutH @sanika-n, as much as I would love to review this PR, my exams got me all tied up for about next 10 days.

EwoutH:
No worries, it’s all voluntarily. Thanks for communicating your availability clearly, I can pick this one up.

EwoutH:
@sanika-n a simple initial test is running the old model for 1000 steps and (manually) timing how long it takes, and then doing the same for the updated model. Make sure to use the same seed in both cases.

sanika-n:
> Awesome, this is really useful. Code itself looks good!
> 
> Why don't the non-blocking boids have an direction?
> 
> Have you any idea on the performance impact?

Just added the direction element to all the boids, and these are the timings that I got after updating the non-flocking boids

Random seed 42

Altered version 
1000 steps: 11min 44sec
50 steps:  15.84s
25 steps: 8.18s 

Original version 
1000 steps: 5min 22s 
50 steps: 10.33s  
25 steps: 5.42s

EwoutH:
@sanika-n thanks for the benchmarks. A 2x slowdown is more that I would like. It would be interesting to see where the performance degradation is coming from:
1. The angle calculation `np.degrees(np.arctan2(...))`
2. Marker transformation & drawing

If it's the former, we could try to vectorize it. If it's the latter, we could add a marker cache. Something like:
```Python
import numpy as np
from matplotlib.markers import MarkerStyle

# Pre-compute markers for different angles (e.g., every 10 degrees)
MARKER_CACHE = {}
for angle in range(0, 360, 10):
    marker = MarkerStyle(10)
    marker._transform = marker.get_transform().rotate_deg(angle)
    MARKER_CACHE[angle] = marker

def boid_draw(agent):
    neighbors = len(agent.neighbors)
    
    # Calculate the angle
    deg = np.degrees(np.arctan2(agent.direction[0], agent.direction[1]))
    # Round to nearest 10 degrees
    rounded_deg = round(deg / 10) * 10 % 360
    
    if neighbors <= 1:
        return {"color": "red", "size": 20, "marker": MARKER_CACHE[rounded_deg]}
    elif neighbors >= 2:
        return {"color": "green", "size": 20, "marker": MARKER_CACHE[rounded_deg]}
```

sanika-n:
Thank you so much! 
I tried to implement vectorization, hope it is in line with what you meant...

EwoutH:
Glad to hear it helped, that’s a serious speed-up!

It might be good to add a few comments what we’re doing and why.

Conceptually and functionally we’re good. @quaquel, could you do a code/implementation review?

quaquel:
@sanika-n can you update this to be in line with the base branch?

My agenda is a bit of a mess, but will try to find a few hours soon to go through various open PRs.

EwoutH (approved):
This is great, thanks a lot!

EwoutH:
Thanks for reminding us @tpike3. @sanika-n please remind us if we leave you hanging so long next time!

PR #2699, opened 2025-02-22
Optimize model  execution to stop immediately when required

HMNS19:
This PR updates the `do_step` function to exit early when `running.value` is `False`. Previously, it would continue executing `n` steps even after `running` was set to stop. Now, it stops immediately when `running.value` becomes `False`.  

This applies to models running in both `ModelController` and `SimulatorController`.

### DIFF
--- a/mesa/experimental/devs/simulator.py
+++ b/mesa/experimental/devs/simulator.py
@@ -110,7 +110,7 @@ class Simulator:
                 "simulator has not been setup, call simulator.setup(model) first"
             )
 
-        while True:
+        while True and self.model.running:
             try:
                 event = self.event_list.pop_event()
             except IndexError:  # event list is empty
@@ -348,7 +348,7 @@ class ABMSimulator(Simulator):
                 "simulator has not been setup, call simulator.setup(model) first"
             )
 
-        while True:
+        while True and self.model.running:
             try:
                 event = self.event_list.pop_event()
             except IndexError:
--- a/mesa/visualization/solara_viz.py
+++ b/mesa/visualization/solara_viz.py
@@ -239,9 +239,10 @@ def ModelController(
     def do_step():
         """Advance the model by the number of steps specified by the render_interval slider."""
         for _ in range(render_interval.value):
+            if not running.value:
+                break
             model.value.step()
-
-        running.value = model.value.running
+            running.value = model.value.running
 
         force_update()
 


HMNS19:
I noticed that the checks didn’t pass in my PR due to an `AttributeError` in the test. The issue seems to be that `model.running` isn’t set in the mock object for simulator models. Since I added a feature to stop the model when `running` is False, the test fails because it tries to access `model.running`, which doesn’t exist in the mock.

Would it make sense to update the test setup to include `model.running = True`, or should we handle this differently?

EwoutH:
Thanks for the PR! Could you explain clearer what the old behavior is, why it was wrong, what the new behavior is, and why that's right?

HMNS19:
Right now, in the Schelling model, if I set the `step interval` to 100 (with default values for every other parameter), it runs all 100 steps even if it should stop earlier. For example, if `running.value` turns False at step 47, it still keeps going till 100, which isn’t how it should work. This PR fixes that by checking `running.value` before every step and stopping the simulation when it’s False.

- before

Screenshot 2025-03-01 144907
- after
Screenshot 2025-03-01 144823



This works fine in` ModelController` and  `SimulatorController ` but the checks fail because the test setup doesn’t include `model.running` in simulator.py which is used to check if the model is running.

We need `model.running` in SimulatorContoller because the condition to stop the simulation is checked in simulator.py. But in the test cases, `model.running` isn’t intialized to `True` in the mock object for simulator models, which causes an AttributeError. Since the code in this PR used `model.running`, the test fails because it tries to access `model.running`, which doesn’t exist in the mock. This is the same issue that caused the checks to fail in my PR earlier.

An alternative would be to copy how ModelController handles this into SimulatorController by directly checking `running.value`  is True in solaraviz.py itself, instead of `model.running` . This would solve the issue without failing the checks, but it would also require using the for loop for the step interval just like ModelController does.

quaquel:
I am trying to understand this PR and what motivates it. In my understanding, the overarching issue is that `model.running` is not used by Solara to stop a model when `model.running` is set to `False.` Is this correct?

Next, when trying to fix this, it has knock-on consequences for event scheduling as well. Again, is this understanding correct?

HMNS19:
My bad for not being clear. This PR ensures the model stops exactly when it's supposed to, instead of running extra steps. This is essentially implementing [#2332](https://github.com/projectmesa/mesa/pull/2332) in [#2596](https://github.com/projectmesa/mesa/pull/2596).  

> The overarching issue is that `model.running` is not used by Solara to stop a model when `model.running` is set to `False`.  

Right. We do have a variable `running.value` in `solara_viz.py` that is responsible for stopping the frontend, but it's only updated after running `n` steps (as specified by `step interval`).  

Now, `ModelController` only required minor changes, which could be handled with `running.value` itself. These changes are already included in this PR.  

For `SimulatorController`, initially, I considered the following approach:  

```python
def do_step():
    for _ in range(render_interval.value):
        if not running.value:
                break
        simulator.run_for(1)
        running.value = model.value.running
    force_update()

```
This would ensure that the model stops at the correct step, and it would also pass the build checks.

But since `run_for` runs the simulator from the current time to current time + time_delta, I thought it would be better to refrain from using a for loop and instead keep the current approach:

```python
    def do_step():
        simulator.run_for(render_interval.value)
        running.value = model.value.running
        force_update()

``` 
while handling `self.model.running` in another suitable place, since this approach doesn't allow checking `running.value` before every step. This is why I opted to use `self.model.running` in `simulator.py` which caused the build checks to fail.

Wanted to get your input on which of these approaches seems more suitable.

quaquel:
1. Can you ensure this PR is up to date with main?
2. I am inclined to separate this into 2 PRs. One is just solving the use of `model.running` bug for normal ABM models. The other would focus on adding support for `model.running` to event scheduling. 
3. Beyond this PR, I find it confusing that we have two places where a `running` variable is declared: once in the model and once in `ModelController` for SolaraViz purposes. Ideally, I would like to find a way of reducing this back to a single variable in just `Model`. However, I am not up to speed with how all this interacts with the threading stuff that has been added recently. @Corvince or @EwoutH: any thoughts?

EwoutH:
@HMNS19 are you still (planning to) working on this?

HMNS19:
Hey @quaquel  @EwoutH, thanks for checking in. I was on a short break, but I’m back now and will get this updated

EwoutH:
Awesome! Let us know if you need anything.

quaquel:
with all the work on the new time api, this is outdated or soon to be outdated. I  am closing this PR.

PR #2708, opened 2025-03-07
JOSS Tutorial Fixes

tpike3:
### Summary
Updates for JOSS paper #2666

### Motive
Per request

### Implementation
- update line lengths (to 79) in tutorials
- edit intro tutorial per JOSS requests
- edit visualization tutorial per JOSS request
- remove pyprojoect.toml info about line-length being taken care of through ruff formatter
- add .virtual documents to .gitignore 
- test_agent was reformatted due to pyproject.toml changes

### Additional Notes
The line-length is not formatting in pyproject.toml as discussed [here](https://github.com/projectmesa/mesa/blob/f511a4bc57340cb2dd0ba4b0af76307b37aea0ca/pyproject.toml#L130) what I am confused about is why it is only identifying some issues on my computer (like 1) but lots of issues when the workflow starts.So I will need to dig into this


pyproject.toml
```diff
@@ -95,7 +95,7 @@ path = "mesa/__init__.py"
 # Hardcode to Python 3.11.
 # Reminder to update mesa-examples if the value below is changed.
 target-version = "py311"
-extend-exclude = ["docs", "build"]
+extend-exclude = ["build"]
```
EwoutH, line 98: This addresses the large package size right? How much does it help?

EwoutH (approved):

EwoutH:
Thanks Tom!

@martibosch could you checkout the [rendered docs](https://mesa--2708.org.readthedocs.build/2708/) and confirm this resolves your comments?

@tpike3 are @jofmi’s comments also addressed in this PR?

pyproject.toml
tpike3: Do you mean line-length or am I missing something? 

This put docs through the ruff linter and formatter, but I also ignored a lot (see per-file-ignores)

To address the line-length across Mesa we need to add to E501 to [tool.ruff.lint] select  and if we want to be compliant with pep 8 we would set the line-length to 79. I am trying to figure out why I am still get a discrepancy between the pre-commit on my machine and the pre-commit in github actions.

tpike3:
> Thanks Tom!
> 
> @martibosch could you checkout the [rendered docs](https://mesa--2708.org.readthedocs.build/2708/) and confirm this resolves your comments?
> 
> @tpike3 are @jofmi’s comments also addressed in this PR?

I addressed all the comments on the checklist except the choose version, this seems to be a readthedocs default, but their documentation on it is not very clear.

pyproject.toml
EwoutH: Sorry I confused this with another comment, you're right.

Personally I won't worry too much about this.

EwoutH:
It was added in this PR:
- https://github.com/projectmesa/mesa/pull/2324

tpike3:
> It was added in this PR:
> 
> * [Readthedocs: Add version switch and update URL #2324](https://github.com/projectmesa/mesa/pull/2324)
> 
> If it’s easier, we could also just reverse this commit: [8d1dd13](https://github.com/projectmesa/mesa/commit/8d1dd13019974263952bf728300449975c81564f)

I removed yesterday and it was added to the PR. Every time I mess with the docs its like I have to relearn the whole process. Ohh well maybe this time it will stick. :)

EwoutH (approved):
Sounds good, merge when you’re ready.

Issue #2705, opened 2025-03-04
Enhancing Command Console


Sahil-Chhoker:
This issue is here to list some features that can improve the working of `CommandConsole` recently added in Mesa through PR #2697.

- [ ] Arrow key navigation for commands.
- [ ] Integrate the input field into the console (like in actual consoles).
- [ ] Extend the auto update that just using `force_update()`.
- [ ] Nice placement window for the console other than sidebar.
- [ ] Improve autoscroll.
- [ ] Inherit all the imports from `app.py` into the command console environment.

Sahil-Chhoker:
@quaquel I think we will need to discuss the placement of console, and what other functionalities we could need that `force_update()` is not able to perform. You did the test for console, how exactly was `force_update()` wasn't working if that was the case.

quaquel:
I tested it for Boltzman, but it should generalize. When using the console to change the wealth of an agent, the space plot did update, but the gini plot did not change.

The reason for this is obvious: gini is calculated in the model and the plot just extracts this number from the model. Because GINI is not a reactive variable, it is not updated when you change the wealth of an agent. So, the problem is not with console or `force_update` per se. 

One solution is to use `mesa.experimental.mesa_signals`. For example, GINI could be turned into a `Computed`, while `agent.wealth` becomes an `Observable`. In this case, the moment you change the wealth of any single agent from the console, this will result in `model.gini` being set to `dirty`. When `force_update` than triggers the extraction of gini from the model, this will result in recalculating gini (because it's dirty), and the plot should be updated accordingly. This, however, is something to be tested (but also showcases nicely why I added mesa_signals).

quaquel:
`mesa_signals` is based on many things and emerged out of close to 1.5 years of discussions and exchanges of ideas. A good starting point is probably reading #2281 and #2291. 

But in short, I have worked in the past with simulation libraries in Java that used the observable pattern for any UI-related functionality. This pattern itself is well over 20 years old by now. So, based on input from the other developers, I started reading up on more modern developments of this, most notably what is called signals in JavaScript. A good, rather technical read is https://github.com/tc39/proposal-signals. `mesa_signals` is a stand-alone implementation of these ideas, tailored to the need of MESA.

[...]

PR #2725, opened 2025-03-18
Feat: Added arrow key navigation

Sahil-Chhoker:
This PR adds arrow key navigation (point 1 of issue #2705) in the `CommandConsole`. 
A problem faced during the implementation of this feature was that we can't apply same event listener (`keyup` in my case) to multiple events. Issue for this is opened [here](https://github.com/widgetti/solara/issues/1014).
Current fix I found was use all the available `key` related events, i.e., `keypress` for Enter, `keyup` for UpArrow, `keydown` for DownArrow. But the navigation works as expected.

### DIFF
--- a/mesa/visualization/command_console.py
+++ b/mesa/visualization/command_console.py
@@ -292,8 +292,10 @@ class ConsoleManager:
     def get_entries(self) -> list[ConsoleEntry]:
         """Get the list of console entries."""
         return self.history
-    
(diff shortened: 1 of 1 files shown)


EwoutH (approved):
Clean diff, useful feature. Thanks!

Final merge for another maintainer.

tpike3:
Thanks @Sahil-Chhoker

PR #2727, opened 2025-03-18
Feat: Added Model sharing through links

### LINKED ISSUE #2724: Sharing of Models through links
I have been working on sharing Mesa models through simple links using PyCafe, and this is my current progress. I believe this feature could be a valuable addition to Mesa, but I'm unsure how to elegantly incorporate it.

After reading the [PyCafe API documentation](https://py.cafe/docs/api), here’s what I have so far:

```python
from urllib.parse import quote
import base64
import json
import gzip

def get_pycafe_link(files: dict[str, str], requirements: str | None = None) -> str:
    """
    Generate a PyCafe link for a given set of files and dependencies.
    """

    # Extract the content of the app.py file
    app = ""
    for file in files:
        if file == "app.py":
            with open(files[file], "r") as f:
                app += f.read()

    requirements = (
        requirements
        or """mesa
        matplotlib
        numpy
        networkx
        solara
        altair
        pandas"""
    )

    # Extract the content of the other files
    file_list = []
    for file in files:
        file_dict = {}
        if file == "app.py":
            continue
        with open(files[file], "r") as f:
            file_dict["name"] = file
            file_dict["content"] = f.read()
            file_list.append(file_dict)

    json_object = {"code": app, "requirements": requirements, "files": file_list}

    json_text = json.dumps(json_object)
    # Compress using gzip to shorten the URL
    compressed_json_text = gzip.compress(json_text.encode("utf-8"))
    # Encode in base64
    base64_text = base64.b64encode(compressed_json_text).decode("utf-8")
    c = quote(base64_text)
    
    return f"https://py.cafe/snippet/solara/v1#c={c}"
```

### Usage:
```python
files = {
    "model.py": "wolf_sheep/model.py",
    "agents.py": "wolf_sheep/agents.py",
    "app.py": "wolf_sheep/app.py"
}

print(get_pycafe_link(files)) # This prints a shareable link of a pycafe project in the terminal.
```

### File Structure:
```
link_share/
├── wolf_sheep/
│   ├── agents.py
│   ├── model.py
│   └── app.py
└── test.py
```

I would love to hear opinions on this feature and how it could be incorporated into Mesa. Do you think it's worth adding? If so, what would be the best way to integrate it?

Sahil-Chhoker:
### Summary
This PR provides a basic model sharing through a button on the solara web-interface. It uses PyCafe to generate links.

### Motive
As explained in #2724. Closes #2724.

### Implementation
A new module named `model_share.py` is added to mesa that handles the link generating and a new button labeled "Open on PyCafe" on the solara's `AppBar`, that redirects to the generated link that contains all the files your model uses. Strict naming convention is followed for autodetection i.e. `model.py`, `agents.py` and `app.py`.

image

### Additional Notes
There are many uncertainties and assumptions in this implementation like the button only works if you are present in that directory and I am unsure about if this is even the correct approach to handle this. But having a draft to work with will make things easier that's why this PR is opened. Please list your opinions.

### DIFF
--- a/mesa/model_share.py
+++ b/mesa/model_share.py
@@ -1,0 +1,86 @@
+"""This module contains functions to generate a shareable py.cafe link containing Python code and dependencies."""
+
+import base64
+import gzip
+import json
+import os
+from pathlib import Path
+from urllib.parse import quote
+
+
+def get_pycafe_link(
+    files: dict[str, str] | None = None,
+    requirements: str | None = None,
+    autodetect_files: bool = False,
+) -> str:
+    r"""Generate a shareable py.cafe link containing Python code and dependencies.
+
+    Args:
+        files(dict[str, str] | None): Map of file names to paths, if files are not in the current directory.
+
+        requirements(str | None): Package dependencies, one per line other than the essential ones.
+
+        autodetect_files(bool): If True, scans directory for Python files instead of using `files`. Mutually exclusive with `files`.
+
+    Example:
+            >>> files = {
+            >>>     "model.py": "wolf_sheep\\model.py",
+            >>>     "agents.py": "wolf_sheep\\agents.py",
+            >>>     "app.py": "wolf_sheep\\app.py"
+            >>> }
+            >>> get_pycafe_link(files=files)
+
+    Returns:
+        str: URL with encoded application code and dependencies.
+
+    """
+    requirements = (
+        requirements or "mesa\nmatplotlib\nnumpy\nnetworkx\nsolara\naltair\npandas"
+    )
+
+    app = ""
+    file_list = []
+
+    if autodetect_files:
+        all_files = _scan_python_files()
+        for file in all_files:
+            with open(file) as f:
+                if file.endswith("app.py"):
+                    app += f.read()
+                else:
+                    file_dict = {}
+                    file_dict["name"] = os.path.basename(file)
+                    file_dict["content"] = f.read()
+                    file_list.append(file_dict)
+    else:
+        for file in files:
+            with open(files[file]) as f:
+                file_content = f.read()
+                if file == "app.py":
+                    app += file_content
+                else:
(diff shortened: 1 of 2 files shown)


mesa/model_share.py
EwoutH, line 1: Is this the most logical place for this file? (also looking at my co-maintainers)

EwoutH (approved):
Wow this is really cool! Can we also add this button to our Readthedocs examples?

I think this is worth experimenting this, so green checkmark from me. I will leave the merge to another maintainer.

Sahil-Chhoker:
Thanks for your reply @tpike3

> * Make it a "Share with PYCAFE"  (Thinking is for those that dont know py.cafe it gives a more obvious use of the button)

Makes sense, will do!

> * With the uncertainties I would not make it a default button, we dont want a users visualization to break for a feature they may not want. We should advertise though, so would need to add it to the visualization tutorial and default on all examples. Maybe a Boolean kwarg?

It makes sense to not make it a default button, but some things I want to clear are that the viz. won't break even if the button does not work as expected, in such case on clicking on the button an empty solara pycafe project opens.
And another thing I want to propose is that we should have a shared account on PyCafe for Mesa, and we can host every example in the main repo their and add their respective links. Since the button functionality works great when you are in the current working directory, it can be a user only button (if this even makes any sense).
> * With the regards to the location, I am thinking in the docs. This is about sharing Mesa/ models/ approaches which is effectively what the docs does, but can go either way

Makes sense, will do!

Sahil-Chhoker:
@coderabbitai review

jackiekazil (approved):
@Sahil-Chhoker looks good to me. Excited about this addition.

Sahil-Chhoker:
@jackiekazil sorry for leaving this PR unfinished, I plan to complete the work as part of my GSoC commitment.

jackiekazil (changes requested):
Holding for completion - not ready to merge.

PR #2732, opened 2025-03-20
Fixed move_agent_to_one_of to handle empty positions correctly

aarav-shukla07:
### Summary

This PR improves the move_agent_to_one_of method in space.py by ensuring that agents do not move when no valid positions are available. Previously, if an empty list was provided, the agent would still move unpredictably, leading to incorrect behavior.
### Bug / Issue

Fixes part of issue [#1903](https://github.com/projectmesa/mesa/issues/1903).

Issue Details:
- The function previously moved agents even when no valid positions were available.
- Tests such as test_move_agent_empty_list and test_move_agent_no_eligible_cells failed due to incorrect movement.

Expected Behavior:
- If no valid positions exist, the agent should stay in place instead of moving randomly.
- The function should handle empty lists gracefully without causing unintended movement.
### Implementation

Added an early return condition to move_agent_to_one_of when pos is empty.
If handle_empty is "warning", it logs a warning but does not move the agent.
If handle_empty is "error", it raises an exception instead of moving the agent randomly.
Ensured that the agent only moves when a valid target position exists.
### Testing

Re-ran full test suite:
```
pytest tests/test_space.py
```
### Additional Notes
- This implementation follows best practices by keeping the logic simple and efficient.
- The solution aligns with the reviewer’s request for an elegant way to place agents on empty cells.
- Further refinements (if any) can be discussed with maintainers for future improvements.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -462,7 +462,7 @@ class _Grid:
     def move_agent_to_one_of(
         self,
         agent: Agent,
-        pos: list[Coordinate],
+        pos: list[Coordinate] = None,
         selection: str = "random",
         handle_empty: str | None = None,
     ) -> None:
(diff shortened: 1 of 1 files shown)


EwoutH:
Thanks for the PR!

You made an improvement to the old space, which is currently maintenance only. So I don’t know if we can accept it, but I will let the other maintainers weigh in.

With older issues it’s always worth checking in if they are still relevant.

The PR looks well structured by the way!

EwoutH:
It happens to the best of us :)

> Should I check if a similar improvement is needed in the newer implementation?

@quaquel I think this feature is already covered by the new API, or is it not yet?

quaquel:
> I think this feature is already covered by the new API, or is it not yet?

Depends a bit on how you define "this feature". CellCollection has a `select_random_cell` method. The `closest` behavior is not readily available but is trivial to do yourself. Just take the absolute difference in coordinates between two cells and sum across coordinates (so using city block distance).

EwoutH:
@aarav-shukla07 you could study the current discrete space API, and if you come up with an idea to improve/extend it, propose it to us :)

aarav-shukla07:
Hey @EwoutH  @quaquel ,
Thanks for the clarification and insights!

I'll study the current discrete space API in more detail and see if I can come up with any ideas to improve or extend it.

If I find a meaningful improvement, I'll propose it to you for discussion before implementing it. 

Thanks again for your guidance! Looking forward to contributing further.

EwoutH:
Perfect, I’m closing this PR, looking forward to your future work!

PR #2733, opened 2025-03-21
paper: Clarify model.agents AgentSet initialization and access

EwoutH:
This PR implements @martibosch's feedback by explaining how the AgentSet class is instantiated and accessed in Mesa. The updated text clarifies that agents are automatically registered with the model and added to AgentSet collections that are accessible through the model's properties.

Added line:
> When agents are created, they automatically register with the model via `model.register_agent(self)`. This registration automatically adds the agent to an `AgentSet` that's accessible through the model's `agents` property. Additional `AgentSet` instances for each agent type are maintained and available through `model.agents_by_type`. These collections are automatically updated when agents are added or removed from the model.


quaquel:
perhaps add how the removal works, otherwise fine with me.

EwoutH:
Thanks, considering the length limitation I’m going to leave it at that.

PR #2743, opened 2025-03-30
Support multiple functions in AgentSet `agg` method

EwoutH:
### Summary
This PR enhances the `AgentSet.agg` method to accept lists and tuples of aggregation functions, enabling multiple aggregations in a single method call.

### Motive
Currently, when analyzing agent attributes, users need to make separate calls to `agg` for each aggregation function (min, max, sum, etc.). This creates performance overhead since the attribute values must be collected multiple times.

This enhancement improves both performance and code readability by supporting multiple aggregation functions in a single call, allowing the attribute values to be collected just once and reused across all aggregation functions.

### Implementation
- Modified the `agg` method signature to accept a single callable or an iterable of callables
- Updated the return type to handle returning either a single value or a list of results
- Updated docstring with examples to demonstrate the new functionality
- Added comprehensive tests to verify all use cases

### Usage Examples
```python
# Before (multiple calls required)
min_wealth = model.agents.agg("wealth", min)
max_wealth = model.agents.agg("wealth", max)
total_wealth = model.agents.agg("wealth", sum)

# After (single call)
min_wealth, max_wealth, total_wealth = model.agents.agg("wealth", [min, max, sum])

# Also works with custom functions
stats = model.agents.agg("wealth", [min, max, np.mean, custom_aggregation_function])
```

### Additional Notes
This enhancement is completely backward compatible, so existing code will continue to work without changes. The method will now return a tuple when multiple functions are provided, maintaining the same order as the input function list.

It might also reduce future datacollector complexity, because aggregation can be done efficiently outside it.

I also considered allowing multiple `attribute` values, but I decided against it since it adds complexity, its use case is less clear and the performance improvements are negligible.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -382,18 +382,18 @@ class AgentSet(MutableSet, Sequence):
         return res
 
     def agg(
-        self, attribute: str, func: Callable | list[Callable] | tuple[Callable, ...]
+        self, attribute: str, func: Callable | Iterable[Callable]
     ) -> Any | list[Any]:
         """Aggregate an attribute of all agents in the AgentSet using one or more functions.
 
         Args:
             attribute (str): The name of the attribute to aggregate.
-            func (Callable | list[Callable] | tuple[Callable]):
+            func (Callable | Iterable[Callable]):
                 - If Callable: A single function to apply to the attribute values (e.g., min, max, sum, np.mean)
-                - If list or tuple: Multiple functions to apply to the attribute values
+                - If Iterable: Multiple functions to apply to the attribute values
 
         Returns:
-            Any | tuple[Any, ...]: Result of applying the function(s) to the attribute values.
+            Any | [Any, ...]: Result of applying the function(s) to the attribute values.
 
         Examples:
             # Single function
@@ -407,7 +407,7 @@ class AgentSet(MutableSet, Sequence):
         if isinstance(func, Callable):
             return func(values)
         else:
-            return tuple(f(values) for f in func)
+            return [f(values) for f in func]
 
     @overload
     def get(
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -404,10 +404,10 @@ class AgentSet(MutableSet, Sequence):
         """
         values = self.get(attribute)
 
-        if isinstance(func, list | tuple):
-            return tuple(f(values) for f in func)
-        else:
+        if isinstance(func, Callable):
             return func(values)
+        else:
+            return tuple(f(values) for f in func)
 
     @overload
     def get(
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -381,18 +381,33 @@ class AgentSet(MutableSet, Sequence):
 
         return res
 
-    def agg(self, attribute: str, func: Callable) -> Any:
-        """Aggregate an attribute of all agents in the AgentSet using a specified function.
+    def agg(
+        self, attribute: str, func: Callable | list[Callable] | tuple[Callable, ...]
+    ) -> Any | list[Any]:
+        """Aggregate an attribute of all agents in the AgentSet using one or more functions.
 
         Args:
             attribute (str): The name of the attribute to aggregate.
-            func (Callable): The function to apply to the attribute values (e.g., min, max, sum, np.mean).
+            func (Callable | list[Callable] | tuple[Callable]):
+                - If Callable: A single function to apply to the attribute values (e.g., min, max, sum, np.mean)
+                - If list or tuple: Multiple functions to apply to the attribute values
 
         Returns:
-            Any: The result of applying the function to the attribute values. Often a single value.
+            Any | tuple[Any, ...]: Result of applying the function(s) to the attribute values.
+
+        Examples:
+            # Single function
+            avg_energy = model.agents.agg("energy", np.mean)
+
+            # Multiple functions
+            min_wealth, max_wealth, total_wealth = model.agents.agg("wealth", [min, max, sum])
         """
         values = self.get(attribute)
-        return func(values)
+
+        if isinstance(func, list | tuple):
+            return tuple(f(values) for f in func)
+        else:
+            return func(values)
 
     @overload
     def get(
--- a/tests/test_agent.py
+++ b/tests/test_agent.py
@@ -413,15 +413,15 @@ def test_agentset_agg():
 
     # Test with list of functions
     min_max_energy = agentset.agg("energy", [min, max])
-    assert min_max_energy == (1, 10)
+    assert min_max_energy == [1, 10]
 
     # Test with tuple of functions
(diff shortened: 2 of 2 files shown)


Sahil-Chhoker:
Its amazing that you still trying to improve the `AgentSet`, I may not have the expertise to review it but I love reading the code for such features.

Ben-geo:
I feel this is more intuitive and easier to navigate, also I am guessing easier to maintain as we can easily add/ remove components

here the title should be Adding space right
[](https://mesa--2731.org.readthedocs.build/2731/tutorials/1_adding_space.html)

also when learning I tend to switch between overview of mesa library and specific component ( might just be my preference ) so it would be nice if overview was present in section navigation ( overview is there on top bar but switching between overview and specific section requires users to go back to getting started)

EwoutH:
> Its amazing that you still trying to improve the `AgentSet`

It's so fun working on it, you can go from idea to complete PR in less than an hour.

> I may not have the expertise to review it but I love reading the code for such features.

Give it a try, it's the quickest way to get there ;)

<hr>

@Ben-geo I think you meant to comment on Tom's PR.

Ben-geo:
> @Ben-geo I think you meant to comment on Tom's PR.

yea I was going through your pr and posted it here by mistake TT

mesa/agent.py
```diff
@@ -381,18 +381,33 @@ def map(self, method: str | Callable, *args, **kwargs) -> list[Any]:
...
+            # Single function
+            avg_energy = model.agents.agg("energy", np.mean)
+
+            # Multiple functions
+            min_wealth, max_wealth, total_wealth = model.agents.agg("wealth", [min, max, sum])
         """
         values = self.get(attribute)
-        return func(values)
+
+        if isinstance(func, list | tuple):
```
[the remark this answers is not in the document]
EwoutH: Python 3.11 introduced the pipe syntax, which we require, so this is actually correct.

quaquel (approved):
looks good to me!

EwoutH:
@quaquel thanks! Before merging, two questions.

> I also considered allowing multiple `attribute` values, but I decided against it since it adds complexity, its use case is less clear and the performance improvements are negligible.

1. Do you agree with this?

2. I want to add a `return_values` boolean in a follow up PR, that allows to also return all values, which internally is already present in `values`. Do you see any problems or things to consider (API wise or other)?

quaquel:
1. We have discussed this before I think and this was why we kept `agg` simple. Indeed, having both multiple attributes and multiple methods is too much. However, I might see a use case for multiple attributes, single operation. However, documenting and type hinting all that becomes messy. 
2. We have split various other `AgentSet` methods to get rid of a return_value boolean, so I am not greatly in favor.

EwoutH:
Thanks for your perspective. Agree on both in some extend.
1. I was thinking of allowing multiple attributes **_or_** multiple methods, but not both. However, I think the former case is both less used and less beneficial performance wise. So I think I will leave it at this.
2. I'm a bit in doubt, but I think if you want all values anyway you could just run a `.get()` yourself and then run the function directly instead of through `.agg`. On the other hand, having this a one-liner is quite elegant. But especially the combination with this PR makes it a bit complicated (since you could have 4 different return types if you adopt both).

I will think about this a bit more.

Sahil-Chhoker:
While I have no problem with this PR, here are some things I would like to know:
1. What was the need of this feature? (maybe point me there)
2. Was 'it' possible before this?

Sahil-Chhoker (approved):

mesa/agent.py
```diff
@@ -381,18 +381,33 @@ def map(self, method: str | Callable, *args, **kwargs) -> list[Any]:
 
         return res
 
-    def agg(self, attribute: str, func: Callable) -> Any:
-        """Aggregate an attribute of all agents in the AgentSet using a specified function.
+    def agg(
+        self, attribute: str, func: Callable | list[Callable] | tuple[Callable, ...]
+    ) -> Any | list[Any]:
```
sanika-n, line 386: According to the docstring the output should be Any | tuple[Any, ...] right?

sanika-n: Other than this, it looks good to me

EwoutH: Perfect, thanks for catching that! I indeed changed the output format, but forget to update the docstring.

Do you also find it logically we output a tuple (and not a list)?

Sahil-Chhoker: > Do you also find it logically we output a tuple (and not a list)?

Are we doing it to make the output read only?

sanika-n: same, I don't think there is a reason we would need it to be read only

EwoutH: In that case I would say, for consistency, let's require the input to be the list (so not a tuple) and then output also a list.

quaquel: In fact the input just has to be an iterable

EwoutH: The only thing is that at some point you have to check if it's a function or an list/tuple/iterable. But yeah, we could check for iterable instead of list/tuple.

quaquel: you can check the other way around. So check if callable else handle it as collection of callable

adamamer20 (approved):
Nice addition!

> 1. I was initially considering allowing either multiple attributes **_or_** multiple methods, but not both. That said, I feel the case of multiple attributes is less common and provides limited performance benefits. So, for now, I think I’ll keep things as they are.

Personally, I’d still vote in favor of supporting multiple attributes as well—it gives the user more flexibility and freedom in how they want to use the function.  
From an implementation perspective, you could use `typing.overload` to define valid combinations of parameters and return types. This way, the IDE will offer typing suggestions depending on the input.

```py
from typing import overload, Collection, Any, Callable
from collections.abc import Iterable

    @overload
    def agg(
        self,
        attribute: str,
        func: Callable
    ) -> Any:
        ...

    @overload
    def agg(
        self,
        attribute: Collection[str],
        func: Callable
    ) -> dict[str, Any]:
        ...

    @overload
    def agg(
        self,
        attribute: str,
        func: Iterable[Callable]
    ) -> tuple[Any, ...]:
        ...

    @overload
    def agg(
        self,
        attribute: Collection[str],
        func: Iterable[Callable]
    ) -> dict[str, tuple[Any, ...]]:
        ...

    def agg(
        self,
        attribute: str | Collection[str],
        func: Callable | Iterable[Callable]
    ) -> Any | dict[str, Any] | tuple[Any, ...] | dict[str, tuple[Any, ...]]:
        # implementation
```

mesa/agent.py
EwoutH: Done in 007c005a6f0a8ef6b875bd60d877df8be3edd1d6

EwoutH: Now outputs a list: b5b572244b1e3d976a7b801602980bf7802729c5

PR #2744, opened 2025-04-01
Memory module proposition - v2

colinfrisch:
# Memory PR upgrade

Based on the feedbacks from the previous PR [(#2735)](https://github.com/projectmesa/mesa/pull/2735) that I made on a proposition for a memory module in mesa (thanks again @EwoutH, @Corvince, @quaquel, if you also have insights for this one, im open to discussion !)

## About the upgrade

Switching to a object format for optimized storage : MemoryEntry is now an object. One of the perks of using an object is not being obligated to have a uniqueID creation module as an object already has an intrinsic uniqueid ```id(my_object)```. It saves storage space as well as lines of code.

image

### ShortTermMemory : more memory-efficient and reactive (efficient store and pop functionality)
- Holds fixed number of recent MemoryEntry objects (automatic removing with dque)
- Optimized for quick retrieval of recent interactions similar to the previous proposition : ```get()```, ```get_by_entry_type()```, etc.
- Can be cleared/reset when an agent session ends

### LongTermMemory : more computational-efficient (efficient navigation)
- Hash map storage for better organization 
- Can be searched (soon, still working on it... I'd greatly appreciate your feedbacks and ideas)

### Justification of memory transfering
```tell_to()``` feature still here implemented directly in the memory (yay), renamed as ```communicate()``` to be more explicit
I see this as "communicating in general is just a way to make a memory of something (like a sentence) and transmitting it to another being". I understand it's unusual, but if the memory module works well, it could mean opening agent communication without building other complex modules. I'm very open to discussion, so please tell me if you have any suggestions regarding this way of doing things !

#### A couple of examples of the structure of the new memory

```python

# Observation from environment
MemoryItem(
    entry_content="The space next to me looks crowded",
    entry_type = "coordinates",
    entry_step = 43,
    metadata={
        "conversation_id": "conv_789",
        "current_loc": agent.pos,
    }
)

# coordinates transfered from another agent
MemoryItem(
    entry_content= (1,3),
    entry_type = "coordinates",
    entry_step = 43,
    metadata={
        "external_agent_id" : external.id,
    }
)


```

## Example usage

```python
# Creating a memory for an agent
agent.memory = Memory(model=self.model, stm_capacity=10)

# Storing information
entry_content= "The space next to me looks crowded",
entry_type = "coordinates",
entry_step = 43,
entry_metadata={
    "conversation_id": "conv_789",
    "current_loc": agent.pos,
}

entry = agent.memory.remember_short_term(
    model,
    entry_content,
    entry_type,
    entry_step,
    entry_metadata
)

# Recalling information
position_data = agent.memory.recall(position_entry_id)

# Finding entries by type in the short term memory
all_position_entries = agent.memory.get_by_type_st("position", include_short_term = True, include_long_term = False,)

# Sharing memory with another agent
agent.memory.tell_to(position_id, external_agent)

# Removing an entry
agent.memory.forget(position_entry_id)
```

## A little bit of performance analysis

The comparaison was done between the last memory structure (from [#2735](https://github.com/projectmesa/mesa/pull/2735)) and the new ShortTermMemory datastructure.

image

## What I want to do next 

- I would like to transform get_by_entry_type into a more general ```search()``` method looking like this :

```python
def search(self, query: str = None, entry_type: List[str] = None, 
            limit: int = 5) -> List[Tuple[MemoryEntry]]:
    """
    Search in short-term memory.
    Implementation could return the n-first entries with relevance scores based on :
        - Exact search
        - Substring search
        - Regex

    Useful for LLM-based agents, but not only. 
    I would 
    """
    return
```

## Summary by CodeRabbit

- **New Features**
  - Introduced a comprehensive memory management module that integrates immediate (short-term) and persistent (long-term) capabilities for improved agent performance.
  - Enhanced event handling with prioritized scheduling, efficient recording, and flexible filtering of memory entries to deliver more dynamic simulation behavior.
  - Optimized overall system efficiency by providing advanced control over memory events, ensuring more reliable and responsive agent interactions.


wang-boyu:
Thanks for the extensive work on this memory module. I had a similar question as @Corvince mentioned in the first PR (https://github.com/projectmesa/mesa/pull/2735#issuecomment-2768793410), about how this can be useful or necessary for all agent types in general. I understand that it is commonly used in LLM-based agents, but I'm trying to understand it from a more general perspective. For example, in our gis examples [Rainfall](https://github.com/projectmesa/mesa-examples/tree/main/gis/rainfall) and [Urban Growth](https://github.com/projectmesa/mesa-examples/tree/main/gis/urban_growth), agents are raindrops and land parcels. How do these inanimate objects have memories?

Would it make sense to relate the current design of memory module, to the history of agents' attributes? Or if they are not the same, what's the difference between the two?

There is an experimental feature for observables: https://github.com/projectmesa/mesa/pull/2291. Imagine I set `pos` (i.e., agents' position) as an observable and I save each agent's past positions as a list inside each one of them. What's the difference of using the observable feature and the memory module here? Is it possible to combine this two together?

colinfrisch:
I see what you mean @wang-boyu. I didn't really design it to fit in all scenarios the same way we don't necessarily us mesa signals for all simulations. The idea was to give the users a new tool and open a simple new range of experimentations with this. Examples I gave in answer to Corvince's feedback were : 

- prey/predator models where remembering certain areas could be useful to the agent (and allow to model patterns like migrations, etc.)
- Basic social interaction models, where remembering who is trustworthy are not could also allow to model a few interesting effects (social networks forming, basic approach strategies, etc.)

These two examples use basic memory (and communication), but require the agents to have individual memories to make it interesting. The memory sharing feature also sets the base for communication without having to build another complicated module. I understand that it would be particularly useful with LLMs, but my opinion is that it can go beyond that.

I'll definitely try to make a model more complex than the foraging ants one to try to show more complex agent patterns with this memory module in the near future !

EwoutH:
Are you still working on this PR?

colinfrisch:
> Are you still working on this PR? Can I help move it forward?

I'm currently working on a model using this memory module to compare it with the same model without. I should finish it by the end of the week. If it's not concluding, I'll post a recap of what can be learned from this PR and close it.

EwoutH:
Awesome, very curious!

colinfrisch:
Back with some insights from building more models depending on agents storing and exchanging information.

I built a virus vs antibody model to simulate an immune reaction declined as a confrontation between antibody agents and virus agents. The antibodies use a short term and a long term memory. I find the result quite interesting, and I'll open a PR to mesa_examples soon, but it's already available [on my model repository](https://github.com/colinfrisch/ABM_mesa_models/tree/main/virus_antibody) if you want to take a look.

The conclusion I had is that indeed, a complex memory module is not necessary for such models because memory in agents can simply be created in the init method, and they usually don't need complex retrieval features or very different types of storing :

```python
self.st_memory : deque = deque()
self.lt_memory : list = []
```

These experimentations also helped me to reflect on the added value of this memory module. All in all, it has some, but it resides in :
- Easily storing extensive metadata for each souvenir (type, and other metadata chosen by the user)
- The ability to store very different types of information in a homogenous way
- The searching features (get_by_type, search with keywords, etc.)
- The efficient memory structure and performance overhead in case of extensive storing of different big memory entries (cf. figure in the PR comment comparing the new short-term memory structure to the old one)

Which are not necessarily needed in most basic agent simulations.

Overall, it’s really what we are looking for in terms of memory LLM because they will need memories of very different types to be easily retrievable (text, reflexions, coordinates, states, etc.). 

Thanks again @EwoutH and @wang-boyu for your help and feedbacks, I'm closing this PR until further notice.

EwoutH:
Thanks for the detailed update!

colinfrisch:
> Have you discovered any feature that might be missing or worth adding to the AgentSet?

@EwoutH I actually struggled quite a bit in terms of agent referencing (problems with agents still referenced even after being removed). I don't know yet if it's specific to the way I built my model ou to the way that weak referencing is handled in the AgentSet class, but I'll definitely look into it.

Also, from what I understood, we have to create a new AgentSet each time that we want to create new agents in the simulation : for example, in the virus/antibody model, the agents can duplicate under certain conditions. There is already an `_update` method in `AgentSet`,  but it is private. Could it be interesting, or simply cleaner for agent organisation and statistics, to be able to add agents to a specific AgentSet (to have only 2 big AgentSets in my model : one for the antibodies, one for the viruses) ? Is there a feature that I am missing ?

EwoutH:
`model.agents` is initialized by default and handled fully automatically. You shouldn’t (have to) modify it manually. If you find a reproduceable bug, please report.

Aside from that, you can assign any agent to any agentset whenever you want. Have you found bugs there or are missing features?

Some (reproduceable) code examples would really help!

colinfrisch:
> Aside from that, you can assign any agent to any agentset whenever you want. Have you found bugs there or are missing features?
Alright, definitely missed that. I had a not very clean but working code so I thought that was it. Definitely going to modify this, thanks for the help !

> model.agents is initialized by default and handled fully automatically. You shouldn’t (have to) modify it manually. If you find a reproduceable bug, please report.
I'm going to refactor my code to hadle agents more cleanly in their AgentSets. If I still have issues with agent referencing that don't appear to be linked to my code, I will make a reproducible example.

quaquel:
Just a couple of pointers which might be helpfult.

1. You can get all agents in a model via `model.agents`
2. You can get all agents of a particular class via `model.agents_by_type[Wolf]`
3. Explicitly use `agent.remove()` when an agent is "_killed_"

In my experience, with these 3 pointers, it is rare to manage your own `AgentSet` instances.

quaquel:
thanks for the example. The simplest solution is to turn target into a weakref: `weakref.ref(some_aagent)`. If you want to access the aggent you need to do `if (agent := target()) is not None:` and then you are fine. In this way, agents can still be removed from the model as normal while all weakrefs resolve to none once the weak referenced agent has been removed. The other solution is that agents know who holds a reference to them and extend `remove` so that all those references are cleared.

Let me know if this is not clear enough. It took me a while to get my head around weakrefs when I first learned about them.

colinfrisch:
@quaquel 
Alright I get it, thanks ! I had encountered this notion of weak referencing before in the docs but didn't get the full sense of it... 
Do you think that it could be an interesting idea to explore the automating of weak referencing when a user makes an agent refer to another, or are there cases were "strong" referencing is needed ?

(Thank you also for the model.agents tip - used it a few times while building but it slipped out of my mind)

colinfrisch:
> There is, as far as I know, no way of making refs weak automatically. However, it might be good to cover weakrefs and the importance of `agent.remove()` and how to extend this yourself in [the docs under best practices](https://mesa.readthedocs.io/latest/best-practices.html).

Alright ! I'm going to work on finding useful tips to write about it in the docs, and I'll open a PR if I do encounter interesting things.

PR #2747, opened 2025-04-02
Debug option for errors in visualisation

### LINKED ISSUE #2745: Is there debugging in solara ?
### problem

I noticed that when there is an error while visualising a mesa model with solara, it only displays the error type, not the classic debugging error (document, line, etc.). 
For example, I got an error `AttributeError: 'my_object' object has no attribute 'len'`, but I didn't know what part it was referring to, as I was using dependencies to different files that I created.

### Solution requested 

Is there a simple option to display the full error message if needed ?

---
Did a PR for this in #2747

colinfrisch:
Resolves #2745

### What the problem was

I noticed that when there is an error while visualising a mesa model with solara, it only displays the error type, not the classic debugging error (document, line, etc.).
For example, I got an error AttributeError: 'my_object' object has no attribute 'len', but I didn't know what part it was referring to, as I was using dependencies to different files that I created. It would be useful to have a possibility to see the full error for debugging.

### How it fixes it

I only added an option to display the error if needed.
Now, instead of displaying  `Error in step : [YOUR ERROR]`, it displays `Error in step: [YOUR ERROR]. Press 'e' for full traceback `, if you press any other key, it doesn't show anything. It also does not prevent you from refreshing your visualisation or resetting it.

### DIFF
--- a/mesa/visualization/solara_viz.py
+++ b/mesa/visualization/solara_viz.py
@@ -273,10 +273,8 @@ def ModelController(
                 do_step()
                 if use_threads.value:
                     visualization_pause_event.set()
-        except Exception as e:
-            debug = input(f"Error in step: {e}. Press 'e' for full traceback ")
-            if debug == "e":
-                traceback.print_exc()
+        except Exception as _:
+            traceback.print_exc()
             return
 
     def visualization_task():
@@ -286,12 +284,8 @@ def ModelController(
                     visualization_pause_event.wait()
                     visualization_pause_event.clear()
                     force_update()
-            except Exception as e:
-                debug = input(
-                    f"Error in visualization_task: {e}. Press 'e' for full traceback "
-                )
-                if debug == "e":
-                    traceback.print_exc()
+            except Exception as _:
+                traceback.print_exc()
 
     solara.lab.use_task(
         step, dependencies=[playing.value, running.value], prefer_threaded=True
--- a/mesa/visualization/solara_viz.py
+++ b/mesa/visualization/solara_viz.py
@@ -27,6 +27,7 @@ import asyncio
 import inspect
 import threading
 import time
+import traceback
 from collections.abc import Callable
 from typing import TYPE_CHECKING, Literal
 
@@ -273,7 +274,9 @@ def ModelController(
                 if use_threads.value:
                     visualization_pause_event.set()
         except Exception as e:
-            print(f"Error in step: {e}")
+            debug = input(f"Error in step: {e}. Press 'e' for full traceback ")
+            if debug == "e":
+                traceback.print_exc()
             return
 
     def visualization_task():
@@ -284,7 +287,11 @@ def ModelController(
                     visualization_pause_event.clear()
                     force_update()
             except Exception as e:
-                print(f"Error in visualization_task: {e}")
+                debug = print(
+                    f"Error in visualization_task: {e}. Press 'e' for full traceback "
+                )
+                if debug == "e":
+                    traceback.print_exc()
 
     solara.lab.use_task(
         step, dependencies=[playing.value, running.value], prefer_threaded=True
--- a/mesa/visualization/solara_viz.py
+++ b/mesa/visualization/solara_viz.py
@@ -287,7 +287,7 @@ def ModelController(
                     visualization_pause_event.clear()
                     force_update()
             except Exception as e:
-                debug = print(
+                debug = input(
                     f"Error in visualization_task: {e}. Press 'e' for full traceback "
                 )
                 if debug == "e":


tpike3 (approved):
Very cool @colinfrisch! I also have been frustrated by the lack of traceback with the visualization and adding the `e` option is inspired, I would have just given them the traceback.

Sahil-Chhoker:
Hello @colinfrisch, thanks for this PR, great work!
I tested it, and it works great! The only issue is that I have to press keys in the terminal. I would also have preferred it to just display the traceback automatically.

colinfrisch:
@Sahil-Chhoker thanks for the feeback ! I kind of liked the idea to have the option to keep a clean interface, especially if we use the terminal for other prints, but if everybody agrees, I'll just make the error plainly display :)
I feel like @tpike3 also liked the "type 'e'" option , but I may have misinterpreted his comment...

Sahil-Chhoker:
> I feel like @tpike3 also liked the "type 'e'" option , but I may have misinterpreted his comment.

I think he is suggesting to remove it, can you confirm @tpike3?

Also have you tried testing it over on your model which you were having issues with? Because that is the best option to confirm its functionality. Also take a look at this [article](https://stackoverflow.com/questions/3702675/catch-and-print-full-python-exception-traceback-without-halting-exiting-the-prog), if your haven't.

tpike3:
I do like the idea of the 'e' option, but granted I have not used it. However, this is a fairly minor change and still an improvement, so always explore-exploit, so my bias would be to keep the 'e' and see how it plays out very easy to change.  

Either way good discussion @Sahil-Chhoker  and @colinfrisch

sanika-n (approved):

sanika-n:
Looks good to me:)
I personally don't have a strong opinion on keeping "e", so I say we keep it at leave it for the user to decide

Sahil-Chhoker (changes requested):
> I haven't seen anything contradictory to this approach in the article that you sent me @Sahil-Chhoker, but I learned much. Did you think otherwise regarding the method?

I just wanted you to read the alternatives, and I think the current implementation is just fine.

I found an issue about your "Press 'e'" feature, it blocks your current thread and therefore the error message on the frontend does not print but is just stuck as is.

Look at the image, here there is a mistake in `agents.py` file but running the model does nothing, its stuck on step 0 and you won't know the error until you see the terminal and I don't like this behavior.
image


But that's just my opinion, I leave the final decision to you, @tpike3.

Sahil-Chhoker (approved):

sanika-n:
> > I haven't seen anything contradictory to this approach in the article that you sent me @Sahil-Chhoker, but I learned much. Did you think otherwise regarding the method?
> 
> I just wanted you to read the alternatives, and I think the current implementation is just fine.
> 
> I found an issue about your "Press 'e'" feature, it blocks your current thread and therefore the error message on the frontend does not print but is just stuck as is.
> 
> Look at the image, here there is a mistake in `agents.py` file but running the model does nothing, its stuck on step 0 and you won't know the error until you see the terminal and I don't like this behavior. image
> 
> But that's just my opinion, I leave the final decision to you, @tpike3.

Just ran a test on this myself, you are right, when there is an error, your front end just freezes and you don't realize it is an error until you check the terminal.. that can be an issue ig

Sahil-Chhoker:
Solara frontend did indeed fail when we are just showing the error because again there is nothing blocking the main thread. Not sure about the full error, but in my opinion the full error is useful even if its just visible on the terminal.

> In all cases, I'd be glad to try and find out how to display the full error on the front end !

You are free to do so if you want, just update or maybe open a follow-up PR if this PR gets merged.

colinfrisch:
I'm quite intrigued that you guys found the fact that solara freezes while there is an error specifically with this feature, because it happened to me every time that there was an error in the program...
I'll definitely look into it !

tpike3:
> @Sahil-Chhoker @sanika-n I'm quite intrigued that you guys found the fact that solara freezes while there is an error specifically with this feature, because it happened to me every time that there was an error in the program... I'll definitely look into it !

We have this same problem with `batch_run` is there is an error on a different processor it will not reach I/O. Curious what you find @colinfrisch! Nice catch @Sahil-Chhoker (Again great dicussion!)

colinfrisch:
Sorry for these manipulations... I accidentally closed the PR so I reopened it. By curiosity, what are the potential next steps before the PR being merged ?

Sahil-Chhoker:
> What are the potential next steps before the PR being merged ?

Remove the `input` statements and change them with simple `print` statements.

colinfrisch:
> Remove the `input` statements and change them with simple `print` statements.
Done !

Sahil-Chhoker (approved):
I am fine with these changes.

colinfrisch:
> I am fine with these changes.

Great ! Then we just need the review from @Corvince and it's ready to merge !

tpike3 (approved):

EwoutH:
Congrats on having your first PR merged!

Issue #2729, opened 2025-03-19
Using DEVSimulator: How to prevent simulation from stopping when no events are present?


D-viddddd:
### Problem Description:
I am using `DEVSimulator` in a Mesa-based simulation, and I noticed that when there are no events in the queue, the simulation stops running. This behavior is problematic when I want to keep the simulation running indefinitely and allow new events to be scheduled dynamically.

### Expected Behavior:
- When there are no events, the simulation should not stop completely but rather wait for new events to be scheduled (or continue advancing time at a fixed interval).
- Is there a built-in way to handle this behavior in `DEVSimulator`?

### Current Workaround:
To prevent the simulation from stopping, I am considering:
1. **scheduling a periodic "idle" event** that does nothing but keeps the queue active.
such as:
```python
def idle_event(model):
    # idle event can do nothing or just update the time
    pass

# Schedule a periodic idle event after the simulator is initialized (e.g. before run_for())
def schedule_idle_events(simulator, interval=1):
    # schedule an idle event every interval seconds
    simulator.schedule_event_relative(idle_event, interval, function_kwargs={'model': simulator.model})
    # call itself to ensure loop
    simulator.schedule_event_relative(schedule_idle_events, interval, function_kwargs={'simulator': simulator, 'interval': interval})

# schedule the idle event immediately once initializing the simulator
schedule_idle_events(simulator, interval=1)
```

2. **Modifying the event processing loop**(function `run_until`) to wait instead of stopping when the queue is empty.
such as:
```python
def run_until(self, end_time):
    while self.time < end_time:
        try:
            event = self.event_list.pop_event()
        except IndexError:
            # the queue is empty, wait for a while, and then continue the cycle
            time.sleep(0.1)
            continue
        self.time = event.time
        event.execute()
```
Is there a recommended way to handle this within `DEVSimulator`, or is this expected behavior?

### Steps to Reproduce:
1. Create a `DEVSimulator`-based simulation.
2. Run it until all scheduled events are completed.
3. Observe that the simulation stops running.

### System Info:
- Mesa version: 3.1.4
- Python version: 3.11.6
- OS: macOS Sequoia 15.2

Any guidance on the best approach would be greatly appreciated!

quaquel:
[...]

PR #2752, opened 2025-04-08
Fix: keep simulation running when no events exist in DEVSimulator.

rafia-10:
**Summary**
Modified run_until() to allow the simulation to continue running even when the event list is empty, by implementing idle event handling.

**Motive**
The simulation previously stopped when no events were in the queue. This change ensures the simulation remains active, waiting for new events.

**Implementation**
Updated the run_until() method to prevent termination when the event queue is empty. Introduced a waiting mechanism to keep the simulator running, even in the absence of events.

**Usage Examples**
# The simulation continues to run even when there are no events in the queue.
simulator.run_until(1000)

**Additional Notes**
This change improves long-running simulations by ensuring they don’t stop prematurely. The new behavior is helpful in cases where events are dynamically scheduled.

### DIFF
--- a/mesa/experimental/devs/simulator.py
+++ b/mesa/experimental/devs/simulator.py
@@ -115,7 +115,9 @@ class Simulator:
                 event = self.event_list.pop_event()
             except IndexError:  # event list is empty
                 self.time = end_time
-                break
+                # Wait briefly instead of stopping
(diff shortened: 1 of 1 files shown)


quaquel:
As indicated in #2729, there is a simple work around if you want this behavior. This PR breaks officials specs for DEVS so I am closing this PR.

PR #2753, opened 2025-04-11
Debug option for errors in visualisation front end

colinfrisch:
This is a followup of #2747 (merged)

## What the problem was
I noticed that when there is an error while visualising a mesa model with solara, it does not display the error in the front-end.
After my first PR, the error now displays only in the terminal, but the user can still have a bit of trouble figuring out what is happening if everything just freezes.

## How it fixes it
Uses `solara.use_reactive` to store the short version of the error as a [reactive state](https://solara.dev/documentation/api/hooks/use_reactive#use_reactive) in the simulation.
I then added the display of the condensed error in the front end. 
The first PR was about getting the full error in the terminal. However, after a few tests, I think that the "condensed" error is enough for the front end, in order to maintain clarity in the interface  :

mesa_error_viz

_Test error on the [foraging ants model](https://github.com/colinfrisch/ABM_mesa_models/tree/main/foraging_ants_V1) I built : to reproduce this exact error, clone the [ABM_mesa_models](https://github.com/colinfrisch/ABM_mesa_models) repo, and uncomment line 141 in `agents.py`_

## Features
- Displays a short error message in the control column of visualization
- Message error disappears and doesn't block simulation if "Reset" button is pressed
- Still displays the full debugging error in the terminal

Thanks again to @EwoutH, @tpike3, @Sahil-Chhoker, and @sanika-n for the feedback in the previous PR, I'd love to hear your thoughts on this follow-up.


sanika-n:
Really cool! Just tried it out
The code also looks good to me:)

EwoutH:
Thanks!

@colinfrisch, could you rebase your branch on our `main` and resolve the conflicts?

@sanika-n is this otherwise good to merge (ideally use the review tool to comment/approve)

Sahil-Chhoker:
Nice initiative, @colinfrisch! I'm not opposing your feature by saying this—just sharing a personal preference. I actually like it when my frontend breaks and shows the error directly; it's easier for me to see the full traceback on the screen rather than navigating through the terminal.

That said, this is just my opinion, and I’ll leave the review and decision to @sanika-n.

colinfrisch:
> @colinfrisch, could you rebase your branch on our `main` and resolve the conflicts?

Alright, I'll do this asap !

mesa/visualization/solara_viz.py
```diff
@@ -389,7 +400,8 @@ def step():
                 if use_threads.value:
                     visualization_pause_event.set()
         except Exception as e:
-            print(f"Error in step: {e}")
+            error_message.value = f"error in step: {e}"
```
Sahil-Chhoker, line 403: What happens if the error message is too long? Won't it clutter the front-end?

colinfrisch:
> I actually like it when my frontend breaks and shows the error directly; it's easier for me to see the full traceback on the screen rather than navigating through the terminal.

@Sahil-Chhoker, I don't think that this PR prevents the simulation from breaking down and displaying the error on the full screen if there's an error in the build. 
It's tackling rather the simulation part once everything is built. Right now, if there is the simulation starts running correctly and an error comes up during the simulation, everything just freezes and, after the first PR, an error is displayed in the terminal. 
This current PR simply notifies the user that there is an error instead of just freezing and waiting. 
However, if you've tested a case where there should be a full breakdown and it doesn't happen anymore, I'd be interested to know about it :)

Sahil-Chhoker:
No, I think I get it, sorry I was referring to something else, you're good.

mesa/visualization/solara_viz.py
colinfrisch: I agree with you that the printed statement shouldn't be to long, and that is why I only print the exception and not the full debugging statement). For now, even if the exception is very long, the use of the `Error` module in solara code allows to simply extends the error in a nice way : 

Screenshot 2025-04-11 at 15 18 34

<br/><br/>
After a little non exhaustive research, the longest exception that i found is `PendingDeprecationWarning`, which still fits quite well in the box.

colinfrisch:
> This is great stuff. @colinfrisch can you resolve the merge conflicts? I'll try to review asap afterwards.

@quaquel @EwoutH, rebase and conflits should be handled. Don't hesitate to let me know if you see any other thing I should do :)

Sahil-Chhoker (approved):
I like this PR, it's an improvement and the code is also minimal, I am approving the changes and leave the merge to @quaquel.

One little thing though, maybe add a comment after `traceback.print_exc()` about what it does.

mesa/visualization/solara_viz.py
Sahil-Chhoker: Great! No problems from my side then.

EwoutH (approved):
Great stuff, will give Jan some time to review (and merge).

PR #2789, opened 2025-05-19
Make random number generator optional in creation of AgentSet, take R…

verisimilidude2:
Merge branch into main

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -162,30 +162,36 @@ class AgentSet(MutableSet, Sequence):
         which means that agents not referenced elsewhere in the program may be automatically removed from the AgentSet.
 
     Notes:
-        A `UserWarning` is issued if `random=None`. You can resolve this warning by explicitly
-        passing a random number generator. In most cases, this will be the seeded random number
-        generator in the model. So, you would do `random=self.random` in a `Model` or `Agent` instance.
+        If random is None then the random number generator in the model of the first agent is used.
+        If the agents list is empty a user warning is issued and the agent set is an empty list
+        with no random number generator
 
     """
 
-    def __init__(self, agents: Iterable[Agent], random: Random | None = None):
+    def __init__(self,
+                 agents: Iterable[Agent],
+                 random: Random | np.random.Generator | None = None):
         """Initializes the AgentSet with a collection of agents and a reference to the model.
 
         Args:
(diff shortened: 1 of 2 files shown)


verisimilidude2:
Change int cast that Ruff complained about

quaquel:
Looks good to me. A few minor requests

1. Can you expand the PR description
2. Can you resolve the conflict


Thanks!

jackiekazil:
Agree with @quaquel. 
Thank you for the contribution @verisimilidude2! Looking forward to getting this merged in with those updates.

verisimilidude2:
@quaquel I'm not sure how one changes the PR description.  The diff on the note on the AgentSet class pretty much describes what I did:
Was:
      A `UserWarning` is issued if `random=None`. You can resolve this warning by explicitly
        passing a random number generator. In most cases, this will be the seeded random number
        generator in the model. So, you would do `random=self.random` in a `Model` or `Agent` instance.
Now:
        If random is None then the random number generator in the model of the first agent is used.
        If the agents list is empty a user warning is issued and the agent set is an empty list
        with no random number generator

EwoutH:
What's the status of this PR? And what's needed to move it forward?

quaquel:
I guess it's stale. I requested 2 small things, neither of which has been done. 

Fixing the PR descritption can be done by us. I just merged in the main branch to check if there are any new conflicts or failing tests.

EwoutH:
@quaquel I updated the title and description, could you review it?

Code itself is good?

mesa/agent.py
```diff
@@ -162,30 +162,39 @@ class AgentSet(MutableSet, Sequence):
...
+        is an empty list and a default random number generator.  This can make models non-reproducible.
+        If your code may create an AgentSet with no agents please pass a random number generator explicitly.
 
     """
 
-    def __init__(self, agents: Iterable[Agent], random: Random | None = None):
+    def __init__(
+        self,
+        agents: Iterable[Agent],
+        random: Random | np.random.Generator | None = None,
```
quaquel: the typing here is not correct.

first, np.randomGeneratpr os deprecated
second, and critical, we don't yet use numpy inside agentset but just stdlib random, so the original typing was correct. 


```suggestion
        random: Random | None = None,
```

quaquel (approved):
one typing suggestion, looks fine otherwise

EwoutH:

PR #2788, opened 2025-05-19
docs/nix flake

hussainsultan:
### Summary

add docs for using nix flake and the runnable app
### Additional Notes

depends on #2785

PR #2785, opened 2025-05-19
add nix flake

hussainsultan:
### Summary

add a nix flake with uv2nix 
### Implementation

- uv2nix
- shells available: `impure`, `uv2nix`

PR #2790, opened 2025-05-20
Move to uv and pin versions

derkweijers:
### Summary

Currently versions are not pinned in the project. This can lead to unexpected behavior when various people are working on the project at the same time, as versions of packages might differ based on when the project was cloned.

`uv` is gaining popularity quickly, it is extremely fast and is able to lock dependencies in a file called `uv.lock`. By locking the dependencies. This will result in a reproducable environment across various systems. It also reduces the chance of someone accidentally using a major new version of a dependency.

### Motive

Generally speaking, it's best practice to provide version constraints on dependencies that are used in a project. This way, you prevent accidental major version bumps and introducing irregular behavior between systems. It also makes it possible to update packages in a controlled, automatic and predictable way with tools like Dependabot and Renovate.

At the same time, this will also enable the optimization of the Dockerfile, decreasing the size and speeding up the build time.

### Implementation

`pyproject.toml` was updated with version ranges that should currently be supported by Mesa. This also generated a `uv.lock` which holds the exact versions.

### Usage Examples

`uv sync` -> generate a virtual environment
`uv run <some tool>` -> run a tool/script within the virtual environment
`uv add` -> add a dependency to the project
`uv add --optional <extra> <dependency>` -> add a dependency to one of the extra groups

### Additional Notes

This is still compatible with `pip` for people who are not using `uv` yet.

The docs of uv are located here: https://docs.astral.sh/uv/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,10 +35,10 @@ classifiers = [
 ]
 readme = "README.md"
 dependencies = [
-  "numpy",
-  "pandas",
-  "tqdm",
-  "scipy"
+  "numpy>=2.2.6,<3",
+  "pandas>=2.2.3,<3",
+  "scipy>=1.15.3,<2",
+  "tqdm>=4.67.1,<5",
 ]
 dynamic = ["version"]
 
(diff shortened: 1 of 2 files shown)


jackiekazil:
@derkweijers it seems that @hussainsultan added uv.lock in #2785, but also dependent on resolution is #2788. 
I am thinking the best way to is remove uv in this PR? Let me know your thoughts.

derkweijers:
Found the issue, I've introduced 2 small bugs when fixing Ruff findings in the example repo. Apologies!

Fixed in the this PR: https://github.com/projectmesa/mesa-examples/pull/278.
Once merge, I'll trigger the actions for this PR again so it can be reviewed.

EwoutH:
Move to `uv` is great!

What’s the maintenance/update plan/strategy for the fixed dependencies?

jackiekazil:
@EwoutH periodic update?
@derkweijers - do you have another suggestion?

derkweijers:
Apologies for the delayed response. I was traveling back home and spending some time with the family.

Personally I've been using [Dependabot](https://github.com/dependabot) over the past few years. Uv support has been slow, so for my own projects I'm migrating to [Renovate](https://docs.renovatebot.com/) which does the same thing.

Both projects will update dependencies, lockfiles, GitHub Actions. Renovate is also able to update base images that are specified in Dockerfiles.

In my opinion if this PR gets adopted/merged, setting up Dependabot or Renovate would be one of the next steps.

EwoutH:
While in packages is common to pin dependency versions, in Python libraries it's best practice not to. This way we stay compatible with as many user systems as possible. We only pin a (minimal) version if we know we use a very new specific feature from a library.

I do like to move to `uv` though. Would you like to open a PR for that?

EwoutH:
I’m going to close this PR since we don’t require pinned dependencies, but feel free to open a new PR for the move to `uv`!

PR #2791, opened 2025-05-20
Remove unneeded int cast

derkweijers:
### Summary

`round()` implicitely returned an int, which was later cast to an int again. This can be avoided.

### Implementation

I've removed the call to `int()` and gave a second argument of `None` to `round()` so it explicitly returns an int.

### Testing

This makes Ruff happy again, as it was failing because of the double int cast.

### Additional Notes


quaquel (approved):
Thanks!

PR #2797, opened 2025-05-27
Removing Deprecations Warnings from `agent_portrayal` and `propertylayer_portrayal` until next release.

Sahil-Chhoker:
### Summary
Removes Deprecations Warnings from `agent_portrayal` and `propertylayer_portrayal`, linked [comment](https://github.com/projectmesa/mesa/pull/2786#issuecomment-2910540569).

### DIFF
--- a/mesa/visualization/mpl_space_drawing.py
+++ b/mesa/visualization/mpl_space_drawing.py
@@ -93,12 +93,6 @@ def collect_agent_data(
         aps: AgentPortrayalStyle
 
         if isinstance(portray_input, dict):
-            warnings.warn(
-                "Returning a dict from agent_portrayal is deprecated and will be removed "
-                "in a future version. Please return an AgentPortrayalStyle instance instead.",
-                DeprecationWarning,
-                stacklevel=2,
-            )
             dict_data = portray_input.copy()
 
             agent_x, agent_y = get_agent_pos(agent, space)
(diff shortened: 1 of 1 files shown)


EwoutH (approved):
Perfect!

We can revert this commit after a few releases.

EwoutH:
Wait, actually, I think you can just change it to a [`PendingDeprecationWarning`](https://docs.python.org/3/library/exceptions.html#PendingDeprecationWarning) instead of removing it.

EwoutH:
It's quite a niche thing I happened to know by accident.

Thanks!

EwoutH:
I think we can update these to full DepreciationWarnings again.

We do have to pick a version when to deprecate. Following full SemVer is needs to be 4.0, but I think we made an exception for the visualization, so we can also do 3.5 for example.

We should also add a few lines to the [migration guide](https://github.com/projectmesa/mesa/blob/main/docs/migration_guide.md).

Sahil-Chhoker:
I actually have no idea how versions and migration guide works. But I think we'll need to update the migration guide with the new visualization changes because it changed a lot with the introduction of portrayal components and space renderer.

EwoutH:
@Sahil-Chhoker we follow https://semver.org/. It basically comes down to only make breaking changes between major releases (x.0.0). Minor releases (x.y.0) are for new features and patch releases (x.y.z.) is for fixing bugs, updating docs, maintenance, etc.

For visualisation we have made an exception, we don't apply SemVer strictly to keep development speed up.

If we plan to make a breaking change, we normally deprecate it at least one minor release before removing it (which would be a major release). We do that by adding notes to the docs and release notes, making sure it isn't used in any examples anymore, making sure the replacement is fully ready and stable, etc. We also add notes to the migration guide.

So here we plan to remove something, and the replacement is already ready, examples are updated, so we can deprecate it. I think everything is already done for that, except adding a note to the migration guide on how to update and actually updating to the full `DepreciationWarnings`.

Does this help?

PR #2806, opened 2025-06-24
fix : prevent breakdown when citizen_density (Initial Agent Density s…

colinfrisch:
# 🐛 Bug fix

Noticed tiny bug on civil violence : the model breaks down when citizen_density (Initial Agent Density slider) is set to 0.
Fix is : model density starts now at 0.1


EwoutH (approved):
Thanks for spotting and fixing this, I think it makes sense.

I will leave the PR open for ~24 hours for additional comments/reviews.

tpike3 (approved):
LGTM

PR #2809, opened 2025-07-10
Renderer null check fix in solaraviz

Sahil-Chhoker:
Explicity check if the renderer is None or not in `SolaraViz`. 
Normal check was causing a error on the solara interface.


tpike3 (approved):
LGTM

PR #2817, opened 2025-07-22
Fix: Grid not showing in altair if both draw_structure and draw_agnets are called.

Sahil-Chhoker:
### Summary
If both `draw_structure` and `draw_agnets` are called in altair backend, the grid won't show because the axis are overridden when layering the two charts.

### Implementation
Resolved the axis for both x and y to independent to prevent overriding of axis because of charts.

## Summary by CodeRabbit

* **Refactor**
  * Improved the display of layered spatial charts by setting their axes to be independent, enhancing clarity when viewing multiple chart layers.


Sahil-Chhoker:
@coderabbitai full review

PR #2829, opened 2025-09-01
Fix pre-commit failure

tpike3:
## Summary
- Updates tests to ensure code passes CI test before release 3.3.0

## Bug / Issue
 - pre-commit is failing due to ruff update

## Implementation
 - Add `# noqa: PLC041` for model imports in tests
 - Move repeated imports to top of file

PR #2831, opened 2025-09-02
Release3.3

tpike3:
Release for v3.3.0
- updates release history
- updates `__init__.py`


EwoutH (approved):
Ideally the ruff stuff should be fixed in a separate PR, but otherwise, looks good!

tpike3:
The ruff stuff is also #2829, I was being sloppy with branching... sorry about that. I will close that PR

PR #2832, opened 2025-09-06
Release updates

tpike3:
Update `init` to 3.4.0.dev per [release process](https://github.com/projectmesa/mesa/blob/main/CONTRIBUTING.md#releases)

### DIFF
--- a/mesa/__init__.py
+++ b/mesa/__init__.py
@@ -24,7 +24,7 @@ __all__ = [
 ]
 
 __title__ = "mesa"
-__version__ = "3.3.0"
+__version__ = "3.3.0.dev"
 __license__ = "Apache 2.0"
 _this_year = datetime.datetime.now(tz=datetime.UTC).date().year
 __copyright__ = f"Copyright {_this_year} Project Mesa Team"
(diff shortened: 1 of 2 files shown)


mesa/__init__.py
```diff
@@ -24,7 +24,7 @@
 ]
 
 __title__ = "mesa"
-__version__ = "3.3.0"
+__version__ = "3.3.0.dev"
```
EwoutH: I think this should be either `3.3.1.dev` or `3.4.0.dev`

EwoutH: (or `4.0.0.dev`, but I don’t think we’re planning that already, right?)

quaquel: 3.4 dev makes the most sense to me. 3.3.1 would be for bug fixes.

Is there also a 3.3 maintenance branch?

tpike3: Sorry I am behind... will do

tpike3: Updated

tpike3:
Apologies I didn't follow good dev practices with separate branches so I added the fix to #2836 (fix broken links in README) in this commit.

PR #2841, opened 2025-10-07
Fix: Auto-increment seed across batch_run iterations

EwoutH:
### Summary
This PR updates `batch_run` to provide explicit control over random seeds for multiple replications. A new `rng` keyword argument is added that accepts a single seed value or an iterable of seed values, while the `iterations` parameter is deprecated.

### Motive
When using `batch_run()` with a single seed value and multiple iterations, all iterations used the same seed, producing identical results instead of independent replications (see #2835). This made it impossible to run meaningful parameter sweeps with reproducible but varied random states across iterations.

```python
# Previous problematic behavior
parameters = {'seed': 42}
batch_run(MyModel, parameters, iterations=10)
# All 10 iterations use seed=42 → identical results
```

### Implementation
- Added new `rng` keyword argument to `batch_run` that accepts either a single seed value or an iterable of seed values
- Deprecated the `iterations` parameter with a `DeprecationWarning` directing users to the migration guide
- Added validation to prevent using both `iterations` and `rng` simultaneously
- The implementation inspects the model's signature to determine whether to use `seed` or `rng` as the parameter name when passing to the model
- Each seed value in the iterable creates a separate run, giving users explicit control over reproducibility
- Updated tests to cover the new functionality and edge cases
- Updated the batch_run tutorial notebook with the new recommended usage
- Added migration guide entry for Mesa 3.4.0

### Usage Examples
```python
import numpy as np
import sys

# Create 5 random seed values
rng = np.random.default_rng(42)
seed_values = rng.integers(0, sys.maxsize, size=(5,))

results = mesa.batch_run(
    MoneyModel,
    parameters=params,
    rng=seed_values.tolist(),  # Pass the 5 seed values
    max_steps=100,
    number_processes=1,
    data_collection_period=1,
    display_progress=True,
)
```

For a single run with a specific seed:
```python
results = mesa.batch_run(MyModel, {}, rng=42)
```

### Additional Notes
- Using the deprecated `iterations` parameter will issue a `DeprecationWarning` but continues to work (passing `None` as the seed for each iteration, maintaining backward compatibility)
- The results now include a `seed` column showing which seed was used for each run
- This is a breaking change for users who relied on the implicit behavior of `iterations` - they should update their code to explicitly provide seed values via `rng`

### DIFF
--- a/tests/test_batch_run.py
+++ b/tests/test_batch_run.py
@@ -97,6 +97,7 @@ class MockModel(Model):
             fixed_model_param: fixed model parameters
             enable_agent_reporters: whether to enable agent reporters
             n_agents: number of agents
+            seed : random seed
             kwargs: keyword arguments
         """
         super().__init__(seed=seed, **kwargs)


quaquel:
I agree that this needs to be fixed. However, using subsequent integers with Mersenne Twister, Python's default RNG, is a bad idea. 

From wikipedia: "_A consequence of poor diffusion is that two instances of the generator, started with initial states that are almost the same, will usually output nearly the same sequence for many iterations_". Also, using a seed with many zeros (like 42) is actually bad as well. One option is to just use `time.time()` every single time and return this seed value for reproducibility. 

As an aside, numpy's rng is much better and I believe we should move all mesa code over to using this while deprecating the use of python's stdlib random library.

tpike3:
Considering how important this is, maybe we should just go all in and do the switch to numpy and its rng and then have seed options like system time and hierarchical seeding?

Does it have to be a breaking change? Could we keep the old behavior and just add a warning?

quaquel:
> Does it have to be a breaking change? Could we keep the old behavior and just add a warning?

Moving the internals over should be possible as a non-breaking change.

mesa/batchrunner.py
```diff
@@ -170,6 +170,14 @@ def _model_run_func(
         Return model_data, agent_data from the reporters
     """
     run_id, iteration, kwargs = run
+
+    # Handle seed uniqueness across iterations
+    if "seed" in kwargs and kwargs["seed"] is not None and iteration > 0:
+        seed_value = kwargs["seed"]
+        if isinstance(seed_value, (int, float)) and not isinstance(seed_value, bool):
+            kwargs = kwargs.copy()
+            kwargs["seed"] = int(seed_value) + iteration
```
quaquel:
```suggestion
            kwargs["seed"] = seed_value + time.time()
```

This is all that is needed to ensure a much better spread of seeding values and thus better randomness.

EwoutH:
> As an aside, numpy's rng is much better and I believe we should move all mesa code over to using this while deprecating the use of python's stdlib random library.

> Considering how important this is, maybe we should just go all in and do the switch to numpy and its rng and then have seed options like system time and hierarchical seeding?

Considering this, do we want to move forward with this PR?

> return this seed value for reproducibility.

Where/how should we do this (without breaking API)?

quaquel:
> Considering this, do we want to move forward with this PR?

Shifting to numpy rng requires changes in e.g., `CellCollection` and `AgentSet`, it's independent from this PR. 


> Where/how should we do this (without breaking API)?

The return is `List[Dict[str, Any]]`, so in prinicple you could just insert a seed kwarg into the dict. 

Alternatively, you can keep stuff as is and just document the behavior. 

Or, perhaps even better: raise a ValueError if `seed` and `iterations` don't match. So, if you do `iterations=10` and `seed=[5,]` you raise a `ValueError` because the number of iterations and the number of seeds don't match. 

In my view, we might even consider deprecating `iterations` in favor of only seed, where seed is either a single `SeedLike` or a list of `SeedLike`.

quaquel:
@EwoutH, I have updated this PR as discussed yesterday. I added a new kewword argument `rng` and deprecated `iterations`. I have also updated the tests accordingly. 

While reviewing the code, I noticed the current use of `run_id` and `iteration` in the return value from `batch_run`. `run_id` is just the sum total of runs, while `iteration ` gives the iteration-id of particular experiment. I am inclined to modify this (probably in a separate PR). In my mind, it makes sense to have 3 pieces of information: an identifier for the experiment (i.e., the exact parameter settings), an identifier for the iteration/replication (this might also just be the seed value instead of an integer starting from 0), and possibly an identifier for which run in total it is. Currently, `batch_run` only gives you the second and third, so grouping by experiment can be a bit tricky, while this is critical when calculating, e.g., the average across replications.

quaquel:
1. It's technically not breaking because iterations will continue to work. It just issues a deprecation warning. 
2. Yes, I'll take a look at adding it to the migration guide.
3. If you use arguments instead of keyword arguments, you deserve what you get :). Also, `iterations` is still there and still functions, but with a warning. So there is no problem.
4. Yes, I was planning to update the docs.

EwoutH:
Good stuff.

I was thinking (not for this PR), wouldn’t it be useful if the results wasn’t just a dataframe, but an object?

quaquel:
> I was thinking (not for this PR), wouldn’t it be useful if the results wasn’t just a dataframe, but an object?

I am not sure. With the workbench, I have never seen the need for anything other than a dataframe with the experiments and a dictionary with the outcome names as keys and a numpy array as values. It might be different, however, if you want to store agent-level data over time. Still, for those use cases, you don't want to keep everything in memory.

quaquel:
@EwoutH, this is ready for review. I added the docs, migration guide, and fixed the tests.

docs/migration_guide.md
```diff
@@ -3,6 +3,39 @@ This guide contains breaking changes between major Mesa versions and how to reso
...
+    MoneyModel,
+    parameters=params,
+    rng=rng_values.tolist(), # we pass the 5 seed values to rng
+    max_steps=100,
+    number_processes=1,
+    data_collection_period=1,
+    display_progress=True,
+)
+
+
```
EwoutH: Remove white lines
```suggestion
```

tests/test_batch_run.py
```diff
@@ -158,9 +196,81 @@ def test_batch_run():  # noqa: D103
...
+    assert result == [
+        {
+            "RunId": 0,
+            "iteration": 0,
+            "Step": 1000,
+            "reported_model_param": 42,
+            "AgentID": 1,
+            "agent_id": 1,
+            "agent_local": 250.0,
+            "seed": 42,
```
EwoutH, line 219: There seems to be a lot of duplicate information in this output format. Maybe we should clean this up (at some point)

mesa/batchrunner.py
```diff
@@ -70,11 +77,34 @@ def batch_run(
         batch_run assumes the model has a `datacollector` attribute that has a DataCollector object initialized.
 
     """
+    if iterations is not None and rng is not None:
+        raise ValueError(
+            "you cannot use both iterations and rng at the same time. Please only use rng."
+        )
+    if iterations is not None:
+        warnings.warn(
+            "iterations is deprecated, please use rng instead",
```
EwoutH:
```suggestion
            "The `iterations` keyword argument is deprecated, please use `rng` instead.
            See https://mesa.readthedocs.io/latest/migration_guide.html#batch-run",
```

mesa/batchrunner.py
```diff
@@ -70,11 +77,34 @@ def batch_run(
...
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        rng = [
+            None,
+        ] * iterations
```
EwoutH: bit more readable
```suggestion
        rng = [None] * iterations
```

mesa/batchrunner.py
```diff
@@ -70,11 +77,34 @@ def batch_run(
...
+    if not isinstance(rng, Iterable):
+        rng = [rng]
+
+    # establish to use seed or rng as name for parameter
+    model_parameters = inspect.signature(Model).parameters
+    rng_kwarg_name = "rng"
+    if "seed" in model_parameters:
+        rng_kwarg_name = "seed"
```
EwoutH, line 99: This behavior should be concisely explained in the batch_run docstring and tutorial

EwoutH (commented):
Looks good and complete, thanks a lot. I have a few minor comments/suggestions.

tests/test_batch_run.py
quaquel: I agree, but it's beyond the scope of this PR.

mesa/batchrunner.py
quaquel: I can commit it, but ruff is likely to change it back again.

mesa/batchrunner.py
quaquel: why?

Using both seed and rng on a model already gives an error. This just ensures that the seed is set to the keyword argument specified by the user in their model class, whether it is seed or rng.

EwoutH: Fair, that's true. Give me a moment to get up to speed with it myself.

EwoutH:
Thanks, I think I can resolve the last open comments and merge.

Can you make sure the PR description represents the final state of this PR?

quaquel:
updated the original starting post

quaquel (approved):
I think this is ready to go. @EwoutH, I'll leave the merge up to you.

EwoutH:
Thanks, I will give it a final sweep tonight.

EwoutH:
I fixed a small layout thing in the migration guide and extended the PR description a bit.

@quaquel one last thing, I noticed the parameter inspection logic:

```python
model_parameters = inspect.signature(Model).parameters
rng_kwarg_name = "rng"
if "seed" in model_parameters:
    rng_kwarg_name = "seed"
```

This inspects the base `Model` class, which has both `seed` and `rng` in its signature:

```python
def __init__(
    self,
    *args: Any,
    seed: float | None = None,
    rng: RNGLike | SeedLike | None = None,
    **kwargs: Any,
) -> None:
```

Since `seed` is always present in `Model`, the condition is always `True`, so `rng_kwarg_name` will always be `"seed"`. The code effectively behaves the same as just hardcoding `rng_kwarg_name = "seed"`.

Was the intent to inspect `model_cls` (the user's model) instead of `Model` (the base class) to adapt to different model signatures? Or is always using `"seed"` the desired behavior and this could be simplified?

EwoutH:
@quaquel if you could take a look at the comment above, we could merge this PR.

quaquel:
> Since seed is always present in Model, the condition is always True, so rng_kwarg_name will always be "seed". The code effectively behaves the same as just hardcoding rng_kwarg_name = "seed".

`model` will generally be a user-specified subclass of `Model` and so is unlikely to contain both `seed` and `rng`. 

> Was the intent to inspect model_cls (the user's model) instead of Model (the base class) to adapt to different model signatures? Or is always using "seed" the desired behavior and this could be simplified?

So yes, that was the intent. Will fix it tomorrow.

EwoutH:
Thanks, glad to see this merged!

EwoutH:
Follow-up: test_batch_run is now producing some warnings: https://github.com/projectmesa/mesa/actions/runs/19961904483/job/57244101008#step:7:69

They either should be caught (emitting a warning is the new expected behavior) or the tests should be updated.

`DeprecationWarning` should also be updated to `FutureWarning`.

PR #2842, opened 2025-10-07
Drop Python 3.11, require Python 3.12+ and update to modern type parameter syntax

EwoutH:
This PR updates Mesa to require Python 3.12 as the minimum version and adopts the new [PEP 695](https://peps.python.org/pep-0695/) type parameter syntax introduced in Python 3.12.

Python 3.11 support is dropped following [SPEC 0](https://scientific-python.org/specs/spec-0000/). The Mesa 3.3.x release series will keep supporting Python 3.11.

**Key changes:**
- Updated minimum Python version from 3.11 to 3.12 across `pyproject.toml`, documentation, and pre-commit configuration
- Replaced legacy `Generic[T]` + `TypeVar` syntax with modern type parameter syntax (`class MyClass[T: Bound]:`)
- Removed redundant `Generic` inheritance where classes already inherit from generic base classes
- Applied the new syntax to `DiscreteSpace`, `CellCollection`, `Grid`, and decorator type hints

Please review. I would like to fast-forward merge this myself.

### DIFF
--- a/.github/workflows/build_lint.yml
+++ b/.github/workflows/build_lint.yml
@@ -35,8 +35,6 @@ jobs:
         include:
           - os: ubuntu
             python-version: "3.12"
-          - os: ubuntu
-            python-version: "3.11"
           # Disabled for now. See https://github.com/projectmesa/mesa/issues/1253
           #- os: ubuntu
           #  python-version: 'pypy-3.8'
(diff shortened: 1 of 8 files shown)


quaquel (approved):
looks good to me.

PR #2849, opened 2025-10-15
Fix missing variable g in the tutorial

pazmiller:
A minor fix of an undefined variable, g, under the AgentSet tutorial in the documentation.

### Summary
An undefined variable, g, which I encountered when learning from the tutorial. A simple bug.

### Bug / Issue
The histplot variable, g, was not defined but was called in the next line.

### Implementation
Defined the variable

### Testing
All tests were passed


EwoutH:
CI failure is unrelated, so merging.

Issue #2859, opened 2025-10-27
PropertyLayer Visualization is Broken in HexGrids


flucco:
**Bug Description**

_Note on coordinates: In Mesa, 2D PropertyLayer arrays have shape (width, height), and so (row, column) in a PropertyLayer array = (x,y) in the grid._ 
 
In the latest version of the build, when using the Matplotlib backend for Solara visualization, the values of a PropertyLayer don't show up in the right places on a HexGrid. The color associated with (x,y) will not appear at (x,y) in the hexagonal grid shown on the screen.*

What I would expect is that everything associated with the grid cell (x,y), such as agents and PropertyLayer values, should appear at (x,y) in the visualization. This is how the visualization of an orthogonal Grid behaves.

As far as I can tell, it's the visualization that has an issue, not the HexGrid class itself.

*Precisely, the color value layout is what you would get if you matched a flattened (W x H) array of color values to a flattened (H x W) array of grid cells those color values should correspond to. 

**Reproduction**

I created a modified version of the Sugarscape example to show this bug. I changed the backend to matplotlib and the grid to a HexGrid for the last two screenshots. I also added a file "sugarscape-asymmetrical.txt", which is a version of "sugarscape.txt" with several rows zeroed out. This creates a white line in the grid. It's much easier to see the bug this way.

_**Behavior with Orthogonal Grid**_

Appearance of visualization at start:

<img width="276" height="200" alt="Image" src="https://github.com/user-attachments/assets/e5d96502-154e-4baf-a1a2-150142b25cc3" />

Note the position of the white stripe.

Appearance of visualization at timestep 1:

<img width="276" height="200" alt="Image" src="https://github.com/user-attachments/assets/0ac51525-c689-4254-b1d4-02b14868c624" />

We can see that agents are in white squares because they eat whatever's in their square.

**_Behavior with Hexagonal Grid_**

Appearance of visualization at start:

<img width="276" height="200" alt="Image" src="https://github.com/user-attachments/assets/ff55d09c-0e46-4589-bde9-635de7f8f24d" />

The white stripe is clearly in the wrong place. 

Appearance of visualization at timestep 1:

<img width="276" height="200" alt="Image" src="https://github.com/user-attachments/assets/9e21303c-b6e3-41bc-9179-c04f91395d75" />

We can see that white squares appear, but not where the agents are. This is because the color values are assigned to the wrong grid cells, whereas the agents are shown in the correct grid cells.

[...]

PR #2864, opened 2025-10-28
Fix PropertyLayer Visualization on HexGrids

flucco:
### Summary

Fix a visualization error for PropertyLayers on HexGrids where color values were displayed in the wrong places. Add a transpose operation to put the color values in the right order.

### Bug / Issue

Issue [2859](https://github.com/projectmesa/mesa/issues/2859). I found that when trying to display PropertyLayers on Hexagonal Grids, color values are consistently shown in the wrong places on the grid. The visualizations for hexagonal grids look dramatically different from the visualizations for orthogonal grids. You would expect the visualizations to look similar besides the small differences caused by the hexagonal cell layout.

### Implementation

The issue was in the draw_propertylayer function in matplotlib_backend.py. The code for the HexGrid visualization flattened the W x H PropertyLayer data into a 1D array, then matched those values to a collection of hexagonal grid cells arranged in H x W order. The different orders led to incorrect placement of values on the grid.

The one-line fix is to add a transpose operation (.T) before the ravel operation that flattens the data. This creates a 1D array of color values in the correct H x W order. I picked this approach because it's simple and matches the implementation used for orthogonal grids. Now, all cases use a simple transpose operation to adjust the data for graphing purposes.

### Testing
I tested this fix in a modified version of the Sugarscape example, described in the issue and available in [this branch] (https://github.com/flucco/mesa/tree/bug_example) of my fork. I used the original sugarscape map, 50x50 modifications of that map which added asymmetrical elements, and modified maps that had varied W x H ratios. 

Before the fix, the values aren’t shown in the right places on the grid. Agents should be in white squares due to consumption of resources, but the white squares appear in other locations. Here is the comparison at timestep 1 of Orthogonal and Hexagonal behaviors using the 60x40 grid. (Note: I changed the size of the agents to 7 and alpha to 0.8 as it's easier to see the color values). See also the examples shown in the issue. 

Orthogonal:

ovn_step1_60_40

Hexagonal:

hex_step1_60_40

Note the incorrect placement of the color values, as shown by the differences from the orthogonal grid and how the agents are not in white squares. Instead, the white squares that should correspond to the agents appear elsewhere in the grid.

After the fix, the behavior is correct in all cases I tested, and closely resembles the visualization of the original orthogonal model.

Hexagonal after fix:

hex_step1_60_40_FIXED

I tested the fixed code on all the basic tests given and ran the formatting checks just to make sure. 

### Additional Notes

- I admittedly didn’t do much additional testing because the change is extremely minimal, adding only a single operation in a single line of code. I believe it should only affect the precise thing I was looking to fix. Please let me know if there's anything else it would make sense to test.

- I deleted my pull request and redid it because I didn't have the updated examples in my main branch. I wasn't sure if I should keep them in there but they're very easy to take out. All the examples are in the sugarscape_hex folder. The only actual change is to matplotlib_backend.py. (Also, I initially changed mpl_space_drawing.py, but that code is correct and shouldn't have been modified, so I changed it back).


Sahil-Chhoker:
Thanks for the detailed explanation and for fixing the issue. You’re right — the data was mapped incorrectly, and transposing it does indeed fix the problem, so this PR looks good to me. I also tried running your example but couldn’t get it to work in Altair — not sure why.

You can go ahead and remove the example and the related files from this PR now.

EwoutH:
Thanks for the detailed bug report (https://github.com/projectmesa/mesa/issues/2859) and PR @flucco!

@Sahil-Chhoker thanks for picking up the review.

@quaquel you might also find this interesting and want to skim this through.

quaquel:
I agree with @Sahil-Chhoker, the transpose is the right fix. However, all the other changes seem not to be relevant so once those are removed this can be merged.

flucco:
> Thanks for the detailed explanation and for fixing the issue. You’re right — the data was mapped incorrectly, and transposing it does indeed fix the problem, so this PR looks good to me. I also tried running your example but couldn’t get it to work in Altair — not sure why.
> 
> You can go ahead and remove the example and the related files from this PR now.

Thank you! I will remove my example folder from the PR. It's still in my bug_example branch for anyone's future reference.

RE Altair, the example I made only works with Solara on the Matplotlib backend. I only ever used the Matplotlib code in my own project, so that was the only bug I found. After looking at the Altair version myself, I think there is a separate issue with the Altair visualizations in the latest version of the code. I haven't dug into it much but if you run the unmodified Sugarscape example with the current code, you'll see a bug that looks _very_ similar to what I described and fixed here. So either there's a similar problem in the Altair code or the example itself needs modification to work with the latest updates.

EwoutH:
@flucco feel free to open a new PR if that's easier! Whatever works for you

quaquel:
> So either there's a similar problem in the Altair code or the example itself needs modification to work with the latest updates.

I would not be surprised if an identical issue exists in altair. This transpose stuff has tricked me up several times before with property layer visualization.

Sahil-Chhoker:
I also dug around a bit, the property layer data mapping for `altair` and `matplotlib` is also flipped.

PR #2867, opened 2025-10-28
ruff fixes in visualization

### LINKED ISSUE #2865: Update Ruff to v0.13 and Fix RUF059 Violations
### Description
We should update Ruff from v0.12.11 to v0.13.3 (latest version). The update introduces some new stabilized rules that will require fixes to our codebase.

### Current Status
A PR updating `.pre-commit-config.yaml` to use Ruff v0.13.3 (https://github.com/projectmesa/mesa/pull/2840) currently fails due to violations of the newly stabilized `RUF059` (unused-unpacked-variable) rule.

### 1. Update Ruff Version
The version update in `.pre-commit-config.yaml` is already prepared (https://github.com/projectmesa/mesa/pull/2840).

#### 2. Fix RUF059 Violations
The `RUF059` rule (previously in preview, now stable in v0.13) flags unpacked variables that are never used. We have 18 violations across our codebase where `fig` is unpacked from `plt.subplots()` but never used.

**Affected files:**
- `mesa/visualization/mpl_space_drawing.py` (6 occurrences)
- `mesa/visualization/space_drawers.py` (5 occurrences)
- `tests/test_components_matplotlib.py` (7 occurrences)

**Example violation:**
```python
if ax is None:
    fig, ax = plt.subplots()  # fig is never used
```

**Recommended fix:**
Replace `fig` with `_` to indicate it's intentionally unused:
```python
if ax is None:
    _, ax = plt.subplots()
```

#### 3. Fix RUF043 Violations (Optional)
There are also 3 `RUF043` violations (pytest-raises-ambiguous-pattern) where regex metacharacters in pytest `match=` patterns should be escaped or use raw strings. These should be addressed as well:
- `tests/test_discrete_space.py:912`
- `tests/test_portrayal_components.py:131`
- `tests/test_solara_viz.py:319`

### Migration Notes
According to the Ruff v0.13 changelog, most users should be able to upgrade without issues. The main breaking changes don't affect our configuration:
- ✅ We don't rely on deprecated rule group selections
- ✅ We don't use the deprecated macOS config directory
- ✅ Removed rules (PD901, UP038) were already deprecated

### Action Items
- [x] Fix all RUF059 violations by replacing unused `fig` variables with `_`
- [x] Fix RUF043 violations by escaping regex patterns or using raw strings
- [x] Update Ruff version to v0.13.3
- [x] Verify all pre-commit hooks pass

quaquel:
The stable version of Ruff (v0.13.3) creates two new errors for MESA: [RUF059](https://docs.astral.sh/ruff/rules/unused-unpacked-variable/) and [RUF043](https://docs.astral.sh/ruff/rules/pytest-raises-ambiguous-pattern/). This PR fixes both following the suggested fixes from ruff. 

Closes #2865

### DIFF
--- a/tests/test_solara_viz.py
+++ b/tests/test_solara_viz.py
@@ -300,7 +300,7 @@ def test_model_creator():  # noqa: D103
         handle_error=False,
     )
 
-    with pytest.raises(ValueError, re.escape("Missing required model parameter")):
+    with pytest.raises(ValueError, match=re.escape("Missing required model parameter")):
         solara.render(
             ModelCreator(
                 solara.reactive(ModelWithRequiredParam(param1="mock")), user_params={}


EwoutH:
Awesome, thanks. Could you explain briefly what `re.escape` does?

quaquel:
I just followed the suggested fix for [RUF043 ](https://docs.astral.sh/ruff/rules/pytest-raises-ambiguous-pattern/). 

From the [docs](https://docs.python.org/3/library/re.html#re.escape): "Escape special characters in pattern. This is useful if you want to match an arbitrary literal string that may have regular expression metacharacters in it."

PR #2868, opened 2025-10-29
Fix visualization error for PropertyLayers on HexGrids. Add tranpose …

flucco:
(New pull request that includes just the one commit)
### Summary

Fix a visualization error for PropertyLayers on HexGrids where color values were displayed in the wrong places. Add a transpose operation to put the color values in the right order.

### Bug / Issue

Issue [2859](https://github.com/projectmesa/mesa/issues/2859). I found that when trying to display PropertyLayers on Hexagonal Grids, color values are consistently shown in the wrong places on the grid. The visualizations for hexagonal grids look dramatically different from the visualizations for orthogonal grids. You would expect the visualizations to look similar besides the small differences caused by the hexagonal cell layout.

### Implementation

The issue was in the draw_propertylayer function in matplotlib_backend.py. The code for the HexGrid visualization flattened the W x H PropertyLayer data into a 1D array, then matched those values to a collection of hexagonal grid cells arranged in H x W order. The different orders led to incorrect placement of values on the grid.

The one-line fix is to add a transpose operation (.T) before the ravel operation that flattens the data. This creates a 1D array of color values in the correct H x W order. I picked this approach because it's simple and matches the implementation used for orthogonal grids. Now, all cases use a simple transpose operation to adjust the data for graphing purposes.

### Testing
I tested this fix in a modified version of the Sugarscape example, described in the issue and available in [this branch] (https://github.com/flucco/mesa/tree/bug_example) of my fork. I used the original sugarscape map, 50x50 modifications of that map which added asymmetrical elements, and modified maps that had varied W x H ratios. 

Before the fix, the values aren’t shown in the right places on the grid. Agents should be in white squares due to consumption of resources, but the white squares appear in other locations. Here is the comparison at timestep 1 of Orthogonal and Hexagonal behaviors using the 60x40 grid. (Note: I changed the size of the agents to 7 and alpha to 0.8 as it's easier to see the color values). See also the examples shown in the issue. 

Orthogonal:

ovn_step1_60_40

Hexagonal:

hex_step1_60_40

Note the incorrect placement of the color values, as shown by the differences from the orthogonal grid and how the agents are not in white squares. Instead, the white squares that should correspond to the agents appear elsewhere in the grid.

After the fix, the behavior is correct in all cases I tested, and closely resembles the visualization of the original orthogonal model.

Hexagonal after fix:

hex_step1_60_40_FIXED

I tested the fixed code on all the basic tests given and ran the formatting checks just to make sure. 

### Additional Notes

- It was pointed out that this example doesn't seem to work on Altair but I think this might be due to unrelated issues.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -45,7 +45,7 @@ jobs:
         run: python global_benchmark.py
       # Upload benchmarks, checkout PR branch, download benchmarks
       - name: Upload benchmark results
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v5
         with:
           name: timings-main
           path: benchmarks/timings_1.pickle
@@ -61,7 +61,7 @@ jobs:
       - name: Install Mesa of the PR branch
         run: pip install --no-deps .
       - name: Download benchmark results
-        uses: actions/download-artifact@v5
+        uses: actions/download-artifact@v6
         with:
           name: timings-main
           path: benchmarks
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,7 +41,7 @@ jobs:
         run: hatch build
       - name: Upload package as artifact to GitHub
         if: github.repository == 'projectmesa/mesa' && startsWith(github.ref, 'refs/tags')
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v5
         with:
           name: package
           path: dist/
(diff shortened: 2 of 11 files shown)


Sahil-Chhoker (approved):
@flucco, Thanks for this!

EwoutH:
@Sahil-Chhoker this can be merged, right?

Be sure to add a label to the PR so it ends up in the right place in the release notes (“bug”) in this case.

Sahil-Chhoker:
@EwoutH Yes, this PR can be merged, do you want me to do it?

EwoutH:
Yes, go ahead!

EwoutH:

Issue #2879, opened 2025-11-06
ReadtheDocs Section Navigation


tpike3:
**Describe the bug**
Section navigation for overview.md is not correct

**Expected behavior**
Should show sections of overview instead shows first part of the tutorials

**To Reproduce**
Go to https://mesa.readthedocs.io/stable/overview.html and see section navigation

<img width="1622" height="837" alt="Image" src="https://github.com/user-attachments/assets/95848331-4416-4402-9248-b849d4a7b05e" />

quaquel:
Not sure I understand. According to you what should be in the section navigation?

unnati-jaiswal24:
Hi, I have seen the problem with the navigation section on the overview.md page which is showing the tutorial sections instead of the overview sections. The sidebar should show the current page we are in ,in overview.md file to make navigation easier for users.

I am interested in working on fixing this issue. please assign this issue to me so I can start seeking to sidebar configuration, toctree directives, and related theme settings to resolve it.

Thanks

quaquel:
I am not sure I agree. At the top of the page, you can see "Getting Started", "Overview", etc. so it is redundant to also show this on the left hand side. The section navigation shows the content within a given entry. So, e.g., within getting started or within overview. As it currently does.

dhiraj-143r:
I saw you’re working on this issue. Should I also work on it ?

quaquel:
It's not yet clear exactly what the issue is, so let's first get a clarification from @tpike3

dhiraj-143r:
ok

dhiraj-143r:
I will be waiting for the clarification

tpike3:
Apologies, was in a bit of a rush when I wrote this. To @quaquel points there is the "on this page" in the upper right, I made clarifications to the issue, and am repeating here. 


### Improved issue discussion: 

Section Navigation needs improvement/ reorganization/ removal. It lists the first 6 of 11 tutorials and the best practices, which is nonsensical (appreciating I did the last big docs updates that produced this result)

This section navigation is also apparent in [Getting Started](https://mesa.readthedocs.io/stable/getting_started.html) page, overview on section navigation carries redundant information as shown here when you expand overview. 

<img width="1043" height="661" alt="Image" src="https://github.com/user-attachments/assets/8ad9fd31-a4db-4841-9e70-92640f0ee196" />

Now how to "fix it" is another question.

[...]

PR #2881, opened 2025-11-07
Fix: Improve sidebar navigation 

unnati-jaiswal24:
fixed the bug. 🐛
Refference: doc📃 section navigation:
The changes which will now navigates to the correct ✅documentation .
index.md : Now ensures the main toctree, separates overview and tutorials .
overview.md : Updated its heading , and removed the toctree section as it was keeping hidden links .
conf.py: Added the max depth of headings similar to those of index.md.
This will now help user to find and jump to the part of documentation user wanted to read.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -131,6 +131,10 @@ html_theme = "pydata_sphinx_theme"
 # documentation.
 html_theme_options = {
     "navbar_start": ["navbar-logo"],
+    "navigation_depth": 4,
+    "show_nav_level": 2,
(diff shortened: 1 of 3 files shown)


quaquel:
Thanks for this PR. As pointed out in #2879, however, I am unsure whether there is a problem and whether this PR is necessary. I will await a response from @tpike3 to clarify exactly what the problem is, according to him.

quaquel:
closed and superseded by #2883.

Again thanks for pitching in with your solution.

PR #2883, opened 2025-11-08
Update overview.md

quaquel:
This PR fixes an inconsistency in the section navigation of the docs, as reported in #2879. 

In short, the section navigation for "getting started" and "overview" are not correct. The section overview should give links to the "heading 2" level content for a given section. However, getting started gives inconsistent links to heading 1 sections, while the section navigation for overview misses relevant stuff.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -172,7 +172,7 @@ html_static_path = ["_static"]
 
 # Custom sidebar templates, maps document names to template names.
 html_sidebars = {
-    "getting started": [],
+    "getting_started": [],
     "migration_guide": [],  # No sidebar migration
 }
 


EwoutH:
Could you add before / after screenshots?

docs/conf.py
```diff
@@ -172,6 +172,7 @@
 
 # Custom sidebar templates, maps document names to template names.
 html_sidebars = {
+    "overview": [],
```
EwoutH: This is a handy trick to know!

with this, are the other changes still needed?

tpike3:
@EwoutH  You can see it on read the docs at --- https://mesa--2883.org.readthedocs.build/2883/getting_started.html

docs/conf.py
quaquel: Yes, because the toctree's were wrong.

tpike3: @quaquel Could you add "getting started" to this list as it section navigation is empty.

Thanks for doing this! I am good with your changes and will give it an approve just to move this along as quickly as possible.

quaquel: I fixed it as requested and merged the PR after confirming the section overviews now render correctly.

PR #2885, opened 2025-11-12
Add generic type parameters to Agent, AgentSet, and Model

SiddharthBansal007:
### Summary
Implements [PEP 695](https://peps.python.org/pep-0695/) generic typing for `Agent`, `AgentSet`, and `Model` classes to enable static type checkers to preserve concrete agent and model types throughout the codebase.

### Motive
Previously, type checkers could not infer specific agent or model types:
- `model.agents` returned untyped `AgentSet[Agent]` instead of the specific agent type
- `agent.model` returned generic `Model` instead of the concrete model type
- Method chaining on `AgentSet` lost type information

This made it difficult to catch type errors at development time and reduced IDE autocomplete effectiveness.

### Implementation
- Added generic type parameter `M` to `Agent[M: Model]` to track the specific model type
- Added generic type parameter `A` to `Model[A: Agent]` to track the specific agent type
- Added generic type parameter `A` to `AgentSet[A: Agent]` for type-safe collections
- Updated all method signatures to use generic type variables (`filter_func: Callable[[A], bool]`, `key: Callable[[A], Any]`)
- Added `@overload` decorators to `AgentSet.__getitem__` to distinguish between int and slice indexing
- Updated internal storage type hints (`_agents: dict[A, None]`, `_agents_by_type: dict[type[A], AgentSet[A]]`)

### Usage Examples
```python
class MyAgent(Agent["MyModel"]):
    def custom_method(self) -> None:
        pass

class MyModel(Model[MyAgent]):
    pass

model = MyModel()
agent = MyAgent(model)

# Type checker now knows:
reveal_type(agent.model)  # MyModel (not just Model)
reveal_type(model.agents)  # AgentSet[MyAgent] (not AgentSet[Agent])
reveal_type(model.agents[0])  # MyAgent (not Agent)

# Method chaining preserves types
filtered: AgentSet[MyAgent] = model.agents.select(lambda a: a.pos is not None)
filtered[0].custom_method()  # Type checker knows custom_method exists
```

### Additional Notes
- Requires Python 3.12+ for PEP 695 syntax
- No runtime behavior changes; purely improves static type checking
- Known limitation: `Agent[M: Model]` and `Model[A: Agent]` create a circular type bound dependency, which is acceptable for this use case
- `create_agents` uses independent type variable `T` to support subclass creation

### DIFF
--- a/mesa/discrete_space/cell_agent.py
+++ b/mesa/discrete_space/cell_agent.py
@@ -106,9 +106,8 @@ class FixedAgent(Agent, FixedCell):
         """Remove the agent from the model."""
         super().remove()
 
-        # fixme we leave self._mesa_cell on the original value
-        #  so you cannot hijack remove() to move patches
         self.cell.remove_agent(self)
+        self._mesa_cell = None
 
 
 class Grid2DMovingAgent(CellAgent):
--- a/tests/discrete_space/test_discrete_space.py
+++ b/tests/discrete_space/test_discrete_space.py
@@ -1164,3 +1164,20 @@ def test_select_random_empty_cell_fallback():
     # Ensure the property layer data was actually correct (the fallback relies on this)
     assert grid.empty.data[5, 5]
     assert not grid.empty.data[0, 0]
+
+
+def test_fixed_agent_removal_state():
+    """Test that a FixedAgent's cell is None after removal."""
+    model = Model()
+    cell1 = Cell((1,), capacity=None, random=random.Random())
+    agent = FixedAgent(model)
+    agent.cell = cell1
+
+    assert agent in cell1.agents
+    assert agent.cell == cell1
+
+    # Remove the agent
+    agent.remove()
+
+    assert agent not in cell1.agents
+    assert agent.cell is None


EwoutH:
Thanks for tackling this issue, improved type hints are definitely useful.

The implementation looks solid, but I wanted to mention: since we recently dropped Python 3.11 support and now require 3.12+, we can actually use the newer [PEP 695](https://peps.python.org/pep-0695/) syntax instead of the `Generic[T]` + `TypeVar` approach.

Python 3.12 introduced cleaner syntax for generics that eliminates the need for the conditional `TYPE_CHECKING` blocks and the manual `TypeVar` declarations. Here's what it would look like:

**Instead of:**
```python
if TYPE_CHECKING:
    M = TypeVar("M", bound=Model)
    A = TypeVar("A", bound="Agent")
else:
    M = TypeVar("M")
    A = TypeVar("A")

class Agent(Generic[M]):  # noqa: UP046
    def __init__(self, model: M, *args, **kwargs) -> None:
        self.model: M = model
```

**We can now write:**
```python
class Agent[M: Model]:
    def __init__(self, model: M, *args, **kwargs) -> None:
        self.model: M = model
```

Same goes for `AgentSet` and the method-level generics:
```python
class AgentSet[A: Agent](MutableSet[A], Sequence[A]):
    # ...

@classmethod
def create_agents[T: Agent](cls: type[T], model: M, n: int, *args, **kwargs) -> AgentSet[T]:
    # ...
```

The PEP 695 syntax is more readable, eliminates the runtime/type-checking split, and is the idiomatic way to do this in Python 3.12+. The `# noqa: UP046` comments you added are actually there because Ruff knows this syntax exists but couldn't be used when supporting 3.11 - now we can.

Would you be up for refactoring this to use the new syntax?

SiddharthBansal007:
@EwoutH  Thanks, i have made the following changes

mesa/agent.py
```diff
@@ -24,13 +24,11 @@
 import numpy as np
 
 if TYPE_CHECKING:
```
EwoutH, line 25: Could you check if we still need this? (honestly I don't know)

mesa/agent.py
```diff
@@ -62,7 +60,9 @@ def __init__(self, model: Model, *args, **kwargs) -> None:
         """
         super().__init__(*args, **kwargs)
 
-        self.model: Model = model
+        # Preserve the more specific model type for static type checkers.
+        # At runtime this remains the Model instance passed in.
```
EwoutH: Comments are not needed
```suggestion
```

mesa/model.py
```diff
@@ -107,11 +110,9 @@ def __init__(
         # setup agent registration data structures
         self._agents = {}  # the hard references to all agents in the model
         self._agents_by_type: dict[
-            type[Agent], AgentSet
-        ] = {}  # a dict with an agentset for each class of agents
-        self._all_agents = AgentSet(
-            [], random=self.random
-        )  # an agenset with all agents
```
EwoutH, line 120: please restore the original comments

EwoutH (commented):
Thanks! Few minor points

mesa/agent.py
SiddharthBansal007: ok, I will make the neccessary changes

mesa/agent.py
SiddharthBansal007: As far as I went through the codebase, I think numpy is needed
 Example usecase (Line 146)
``` @property
def rng(self) -> np.random.Generator:
    """Return a seeded np.random rng."""
    return self.model.rng
```
    
2) for the TYPE_CHECKING, i have used it to prevent circular imports, do let me know if i am wrong in any of the both cases

mesa/model.py
SiddharthBansal007: Made the changes, please review the new push

SiddharthBansal007:
@EwoutH incase you missed the changes, could you look upon the changes

mesa/agent.py
```diff
@@ -85,7 +83,9 @@ def advance(self) -> None:  # noqa: D102
         pass
 
     @classmethod
-    def create_agents(cls, model: Model, n: int, *args, **kwargs) -> AgentSet[Agent]:
+    def create_agents[T: Agent](
+        cls: type[T], model: Model, n: int, *args, **kwargs
+    ) -> AgentSet[T]:
```
EwoutH, line 87: Why is `T` used here for `Agent`? Wouldn't `A` be more logical?

Please motivate first. But if you agree:
```suggestion
    def create_agents[A: Agent](
        cls: type[A], model: Model, n: int, *args, **kwargs
    ) -> AgentSet[A]:
```

mesa/agent.py
```diff
@@ -213,7 +213,7 @@ def select(
         filter_func: Callable[[Agent], bool] | None = None,
         at_most: int | float = float("inf"),
         inplace: bool = False,
-        agent_type: type[Agent] | None = None,
+        agent_type: type[A] | None = None,
     ) -> AgentSet:
```
EwoutH: We can also add a type on the return, right?
```suggestion
    ) -> AgentSet[A]:
```

mesa/model.py
```diff
@@ -12,21 +12,24 @@
...
+from typing import TYPE_CHECKING, Any
 
 import numpy as np
 
 from mesa.agent import Agent, AgentSet
 from mesa.mesa_logging import create_module_logger, method_logger
 
+if TYPE_CHECKING:
+    pass
+
```
EwoutH: This empty block can be removed, right?
```suggestion
```

mesa/model.py
```diff
@@ -12,21 +12,24 @@
 from collections.abc import Sequence
 
 # mypy
-from typing import Any
+from typing import TYPE_CHECKING, Any
```
EwoutH: Can be removed I think.
```suggestion
from typing import Any
```

mesa/agent.py
```diff
@@ -307,15 +307,15 @@ def sort(
             else self._update(sorted_agents)
```
EwoutH, line 306: In the sort function, `[A]` can also be added.

```Python
    def sort(
        self,
        key: Callable[[Agent], Any] | str,
        ascending: bool = False,
        inplace: bool = False,
    ) -> AgentSet[A]:
```

mesa/model.py
```diff
@@ -107,9 +110,9 @@ def __init__(
         # setup agent registration data structures
         self._agents = {}  # the hard references to all agents in the model
```
EwoutH: I think this is possible:
```suggestion
        self._agents = dict[A, None]  # the hard references to all agents in the model
```

EwoutH:
@SiddharthBansal007 sorry for the late review. I've a few more (minor) comments, can you address them?

Please don't adopt them without thinking, but critically check.

mesa/agent.py
quaquel: In my understanding of typing, T is just the common default to use. But I am by no means an expert yet on type hinting in python.

quaquel:
@EwoutH, I picked this up and made a bunch of changes. Can you check to see if it's good to go at this point? I'd like to get this merged.

mesa/agent.py
```diff
@@ -212,8 +212,8 @@ def select(
         filter_func: Callable[[Agent], bool] | None = None,
```
EwoutH: Should this be `Callable[[A], bool]`?
```suggestion
        filter_func: Callable[[A], bool] | None = None,
```

mesa/agent.py
```diff
@@ -284,7 +284,7 @@ def sort(
         key: Callable[[Agent], Any] | str,
```
EwoutH: Idem dito?
```suggestion
        key: Callable[[A], Any] | str,
```

mesa/agent.py
```diff
@@ -501,7 +501,7 @@ def set(self, attr_name: str, value: Any) -> AgentSet:
             setattr(agent, attr_name, value)
         return self
 
-    def __getitem__(self, item: int | slice) -> Agent:
+    def __getitem__(self, item: int | slice) -> A:
```
EwoutH: This function doesn't always return an Agent right, but can also return a list of agents? (slice)

Do we need to overload (`@overload`) here?

mesa/space.py
```diff
@@ -57,8 +57,6 @@
 GridContent = Agent | None
 MultiGridContent = list[Agent]
 
-F = TypeVar("F", bound=Callable[..., Any])
-
 
 def accept_tuple_argument[F: Callable[..., Any]](wrapped_function: F) -> F:
```
EwoutH, line 59: Is it correct that `F = TypeVar("F")` was removed, but `F` is still used here?
```suggestion
def accept_tuple_argument[F: Callable[..., Any]](wrapped_function: F) -> F:
```

mesa/agent.py
```diff
@@ -23,13 +23,11 @@
...
 
 if TYPE_CHECKING:
-    # We ensure that these are not imported during runtime to prevent cyclic
-    # dependency.
     from mesa.model import Model
     from mesa.space import Position
 
 
-class Agent:
+class Agent[M: Model]:
```
EwoutH, line 30: Since the `AgentSet` uses `WeakKeyDictionary` which requires agents to be hashable, maybe we should add:
```suggestion
class Agent[M: Model](Hashable):
```

EwoutH (commented):
Thanks for picking this up @quaquel. I agree it would be very useful to have this in Mesa sooner than later.

A few suggestions and questions. Please check them carefully, this is not my core area of expertise.

mesa/agent.py
quaquel: I have no idea how this works exactly, so I will hold off on this for now.

mesa/space.py
quaquel: I reverted all changes in space.py since this is maintenance only.

quaquel:
@EwoutH, I addressed your comments. I think this is ready for now.

I want to guard against scope creep, so as long as these changes are fine, I suggest we move forward. From my own first impressions, some type warnings seem to be gone when using this PR.

mesa/agent.py
EwoutH: I was already thinking this might be too much for this PR, but this is an issue I encountered more often (Agents hashing not being possible).

I wrote a follow-up:
- https://github.com/mesa/mesa/issues/3101

mesa/space.py
EwoutH: Perfect.

EwoutH (approved):
Looks good to me! @SiddharthBansal007 thanks for the initial PR and @quaquel for landing it.

Issue #2884, opened 2025-11-10
Deprecating stdlib random


quaquel:
Mesa currently supports both the sdlib random library (via model.random) and numpy's random number generators (via model.rng). While disucssing [a fix for a bug in the batchrunner](https://github.com/projectmesa/mesa/pull/2841#issuecomment-3378574529), I mentioned in passing that it might be good to shift all internals over to numpy's random number generation.

Why? Python's stdlib random uses [Mersenne Twister](https://en.wikipedia.org/wiki/Mersenne_Twister) and only exists at the global level. Both are a problem. Mersenne Twister is a proven but by now quite dated algorithm for the generation of pseudo-random numbers with various known issues. Most importantly, adjacent seed values (eg., 42, 43, 44) produce very similar random numbers, and using low integers (e.g., 42, 101) as a seed results in poor sequences of random numbers. 

Stdlib random only has global state, so whether you use model.random, or just random does not matter. Moreover, if you have two models running side by side using, e.g., threading, you get race conditions, and you cannot guarantee reproducibility. 

Both problems are solved in numpy's random number generators. You can have multiple independent instances of a generator (so, no global state, only instance-level state), and the algorithms available produce higher-quality sequences of random numbers and are much less sensitive to the values with which the algorithm is seeded. 

Therefore, I propose deprecating `model.random` and favoring `model.rng`. Moreover, I would like to do this in a [spec 7](https://scientific-python.org/specs/spec-0007/) compliant way. However, before we can do this, all internals will have to be moved over. A quick first list (to be updated) of classes that will require changes is
- [ ] AgentSet
- [ ] All discrete spaces
- [ ] CellCollection
- [ ] ContinuousSpace

EwoutH:
In the [spirit](https://peps.python.org/pep-0020/) of Python, I agree.

> There should be one-- and preferably only one --obvious way to do it.

Let's deprecate and remove with 4.0.

tpike3:
I also concur

quaquel:
Ok, I am traveling next week, so I'll try to find some time to open a PR starting this process. My current idea is to move over 1 class per PR: for each class, add rng as a keyword argument (spec-7 compliant), issue a deprecation warning when random is used, and move the internals over to use rng.

schrodingers-cat-T10:
@quaquel hello there , can i assist you with the switch with numpy ?

PR #2888, opened 2025-11-18
Deprecate seed in favor of rng

quaquel:
This PR addressed #2884. 

It does several things

1. Adds a decorator for annotating methods that currently take seed or random as a kwarg. The decorator results in a `FutureWarning` when seed or random is being used instead of the new preferred rng kwarg. I use. FutureWarning because this is intended for end users (See https://docs.python.org/3/library/warnings.html). 
2. Adds this decorator to all methods that have seed or random as a kwarg
3. Starts adding rng to all decorated classes and update their inner workings to use numpy generators instead of stdlib random generators. 

**reminder**
we need to fix the set_seed widget on the solara side as well!

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -45,7 +45,7 @@ jobs:
         run: python global_benchmark.py
       # Upload benchmarks, checkout PR branch, download benchmarks
       - name: Upload benchmark results
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: timings-main
           path: benchmarks/timings_1.pickle
@@ -61,7 +61,7 @@ jobs:
       - name: Install Mesa of the PR branch
         run: pip install --no-deps .
       - name: Download benchmark results
-        uses: actions/download-artifact@v6
+        uses: actions/download-artifact@v7
         with:
           name: timings-main
           path: benchmarks
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,7 +41,7 @@ jobs:
         run: hatch build
       - name: Upload package as artifact to GitHub
         if: github.repository == 'mesa/mesa' && startsWith(github.ref, 'refs/tags')
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: package
           path: dist/
--- a/docs/tutorials/1_adding_space.ipynb
+++ b/docs/tutorials/1_adding_space.ipynb
@@ -272,7 +272,7 @@
     "        self.num_agents = n\n",
     "        # Instantiate an instance of Moore neighborhood space\n",
     "        self.grid = OrthogonalMooreGrid(\n",
-    "            (width, height), torus=True, capacity=10, random=self.random\n",
+    "            (width, height), torus=True, capacity=5, random=self.random\n",
     "        )\n",
     "\n",
     "        # Create agents\n",
--- a/docs/tutorials/2_collecting_data.ipynb
+++ b/docs/tutorials/2_collecting_data.ipynb
@@ -370,7 +370,7 @@
     "last_step = agent_wealth.index.get_level_values(\"Step\").max()  # Get the last step\n",
     "end_wealth = agent_wealth.xs(last_step, level=\"Step\")[\n",
     "    \"Wealth\"\n",
-    "]  # Get the welath of each agentat the last step\n",
+    "]  # Get the wealth of each agent at the last step\n",
     "# Create a histogram of wealth at the last step\n",
     "g = sns.histplot(end_wealth, discrete=True)\n",
     "g.set(\n",
--- a/mesa/examples/advanced/epstein_civil_violence/app.py
+++ b/mesa/examples/advanced/epstein_civil_violence/app.py
@@ -64,8 +64,11 @@ chart_component = make_plot_component(
 )
 
 epstein_model = EpsteinCivilViolence()
-renderer = SpaceRenderer(epstein_model, backend="matplotlib")
-renderer.draw_agents(citizen_cop_portrayal)
+renderer = SpaceRenderer(epstein_model, backend="matplotlib").setup_agents(
+    citizen_cop_portrayal
+)
+# Specifically, avoid drawing the grid to hide the grid lines.
+renderer.draw_agents()
 renderer.post_process = post_process
 
 page = SolaraViz(
--- a/mesa/examples/advanced/pd_grid/app.py
+++ b/mesa/examples/advanced/pd_grid/app.py
@@ -45,7 +45,11 @@ plot_component = make_plot_component("Cooperating_Agents", backend="altair", gri
 # Initialize model
 initial_model = PdGrid()
 # Create grid and agent visualization component using Altair
-renderer = SpaceRenderer(initial_model, backend="altair").render(pd_agent_portrayal)
+renderer = (
+    SpaceRenderer(initial_model, backend="altair")
+    .setup_agents(pd_agent_portrayal)
+    .render()
+)
 
 # Create visualization with all components
 page = SolaraViz(
--- a/mesa/examples/advanced/sugarscape_g1mt/app.py
+++ b/mesa/examples/advanced/sugarscape_g1mt/app.py
@@ -57,11 +57,16 @@ model = SugarscapeG1mt()
 # Here, the renderer uses the Altair backend, while the plot components
 # use the Matplotlib backend.
 # Both can be mixed and matched to enhance the visuals of your model.
-renderer = SpaceRenderer(model, backend="altair").render(
-    agent_portrayal=agent_portrayal,
-    propertylayer_portrayal=propertylayer_portrayal,
-    post_process=post_process,
+renderer = (
+    SpaceRenderer(model, backend="altair")
+    .setup_agents(agent_portrayal)
+    .setup_propertylayer(propertylayer_portrayal)
 )
+# Specifically, avoid drawing the grid to hide the grid lines.
+renderer.draw_agents()
+renderer.draw_propertylayer()
+
+renderer.post_process = post_process
 
 # Note: It is advised to switch the pages after pausing the model
 # on the Solara dashboard.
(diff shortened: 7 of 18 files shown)


quaquel:
As can be seen in the performance benchmarks, shifting from stdlib random to numpy.random.default_rng comes with a performance hit. I dug into this a bit more. The two main random calls that are made in these benchmark models are `shuffle` and `choice`. `np.random.default_rng().shuffle()` is about 20% faster than `random.shuffle()`. However, `np.random.default_rng().choice` is several orders of magnitude slower than `random.choice()`. In fact, for the specific case of choosing a single item randomly from a collection, it is advised to use `collection[rng.integers(0, len(collection)]` instead, but this is still twice as expensive as `random.choice`. 

More broadly, `numpy.random.default_rng()` is designed for generating and operating on numpy arrays. So, calls to create a single random number (e.g., `rng.random`, `rng.integers`) are all much slower than their stdlib random alternatives. However, once the size becomes larger, they become increasingly faster than a list expression with stdlib random. So, shifting to `rng` comes with additional implementation considerations to work around this performance hit. See some of the updated examples for what that can look like.

Thoughts are welcome. In your opinion, is the improved quality of the random number generators and the fact that you have local state with numpy worth the performance hit?

quaquel:
Given how common it is to draw a single random number (i.e., `random.random()`, I was wondering about adding a shorthand for this and see how fast this was.

```python
def test_rand(rng, initial_size=100):
    initial_values = rng.random(size=initial_size)

    while True:
        for entry in initial_values:
            yield entry
        initial_value = rng.random(size=initial_size)

def draw():
    return next(a)

a = test_rand(rng, 200)  
```

And then I timed these as shown below

```
%%timeit
[random.random() for _ in range(250)]
8.88 μs ± 72.4 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

%%timeit
[draw() for _ in range(250)]
20.7 μs ± 130 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

%%timeit
[next(a) for _ in range(250)]
16.4 μs ± 121 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

%%timeit
[rng.random() for _ in range(250)]
67.9 μs ± 273 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

So, a naive shift from `random.random` to `rng.random`, is almost an order of magnitude slower. We can effectively halve this by using the generator and advancing this generator directly via `next`, wrapping this `next` call up in another function adds some overhead, but it is still a lot faster than a naive replacement. So, I am considering adding a method to the model, `model.rand`, that performs this task. The other option would be to figure out how to subclass `numpy.random.Generator` and add this `rand` method to it, because then you can just do `self.rng.rand` as a much faster alternative for `rng.random()`.

EwoutH:
> More broadly, `numpy.random.default_rng()` is designed for generating and operating on numpy arrays.

I don't know if it's a good idea, but: Can you pre-generate an array (like a 100 or a 1000), use those numbers, and everytime you used up the array generate a new one? Basically generate random numbers in batch?

quaquel:
> I don't know if it's a good idea, but: Can you pre-generate an array (like a 100 or a 1000), use those numbers, and everytime you used up the array generate a new one? Basically generate random numbers in batch?

yes, that is exactly what I show here: https://github.com/projectmesa/mesa/pull/2888#issuecomment-3557219776

tpike3:
@quaquel Your stuff is always great and this is no exception. I don't think I have anything substantive to add. From my perspective it is better to improve the rng even if it has a performance hit. 

Then the next question becomes, are there ways to mitigate the performance hit while ensuring optimal randomness. Two options on the table:
1. wrapper
2. subclass

I think 1 is the best bet. Subclassing in python to a generator using c-extensions **may** create unexpected challenges that **may** have unintended impacts on user simulations. 

At this point I would recommend 1, but always open to discussion!

EwoutH:
Can we add a Model lever keyword argument, that defaults to using NumPy RNG but allows you to use random if you need te performance?

quaquel:
I'll try to take another look at this somewhere this week. As a first step, I'll add the wrapper idea to the model class so you can just do `model.rand` if you need a single number. Next, I'll rerun the benchmarks to determine the performance difference. 

@EwoutH, nothing prevents you from using stdlib.random in your own models in the future. If we decide to make `model.rng` the only option in MESA 4, you can still have `MyCustomModel.random` as well but you would have to add this to the model yourself.

quaquel:
The benchmarks are a bit funky via github actions, so I ran them locally. The results are shown below. The TLDR from my perspective is that the performance loss is now acceptable and the benefits of using numpy's random number generation outweight the loss of performance.

This is the benchmarking with a new `model.rand` method for generating individual random numbers on unit interval. Internally, this wraps a numpy array of a prespecified length that is replaced once exhausted. For further details, see my comment with the code above. The `model.rand` method is only used by the WolfSheep benchmark, but it significantly reduces the performance loss compared to a naive use of `rng.random()`.  

There is still some performance loss as can be seen, but this is more manageable. Moreover, some of this is likely due to the overhead of the deprecation decorator I added as part of this PR. I'll run a local benchmark later, removing this decorator, to get a clearer sense of the performance loss.

/opt/anaconda3/bin/python /Users/jhkwakkel/Documents/GitHub/mesa/benchmarks/compare_timings.py 
| BoltzmannWealth | small  | 🔴 +10.1% [+9.0%, +11.0%]  | 🔴 +33.9% [+33.4%, +34.2%] |
| BoltzmannWealth | large  | 🔴 +9.6% [+9.2%, +10.0%]   | 🔴 +27.2% [+24.8%, +29.0%] |
| Schelling       | small  | 🔴 +15.5% [+14.9%, +16.0%] | 🔵 -1.0% [-2.0%, -0.1%]    |
| Schelling       | large  | 🔴 +12.1% [+11.5%, +12.5%] | 🔴 +62.7% [+59.2%, +65.4%] |
| WolfSheep       | small  | 🔴 +16.5% [+15.8%, +17.3%] | 🔴 +30.4% [+23.3%, +37.7%] |
| WolfSheep       | large  | 🔴 +15.1% [+14.9%, +15.4%] | 🔴 +25.0% [+22.6%, +26.6%] |
| BoidFlockers    | small  | 🔵 +0.5% [-0.1%, +1.1%]    | 🔵 -0.3% [-0.5%, -0.1%]    |
| BoidFlockers    | large  | 🔵 -2.0% [-4.3%, -0.1%]    | 🔵 -1.1% [-2.5%, -0.1%]    |

EwoutH:
> I'll run a local benchmark later, removing this decorator, to get a clearer sense of the performance loss.

Curious for this. Because 25 to 60% runtime is not trivial.

quaquel:
> Curious for this. Because 25 to 60% runtime is not trivial.

It's a bit tricky to directly compare the runtime of the old and new cases for both wolf sheep and Schelling. In the case of Schelling, if everyone is happy, no further moves take place, and thus the model runs much faster. In the case of wolf sheep, if all the sheep die quickly, again the model runs much faster afterwards. Since the random numbers are different, the fraction of models that reach this fast "state" might very well be different.

Based on my testing of individual operations (e.g., random.shuffle vs. rng.shuffle), there is a performance loss of about 30%-50% on these individual operations. How that adds up in a bigger model is more difficult to assess because these individual operations are just a small part of the overall model. In many ways, Boltzmann might be the best one to look at because this model has virtually no overhead from agent logic.

Additionally, I would like to delve a bit deeper by conducting some line profiling to identify where performance is being lost exactly.

Accidentally, I just ran a quick benchmark with all the decorators removed, and the overhead of the deprecation decorator is just a few percent.

quaquel:
@EwoutH, @tpike3 , I had a closer look at what is happening with the runtime. In short, there are 2 key changes here

* `random.choice` --> `some_collection[rng.integers(0, len(some_collection))]`
* `random.shuffle` --> `rng.shuffle`

There is also `random.random`, but that is not part of the core of Mesa and is only used in a single benchmark model, so we can ignore that for now.

To start with the good news, `rng.shuffle` is an order of magnitude faster than `random.shuffle`. If I test this change in isolation, the benchmark models run 5%-10% faster. 

The problem is with replacing `random.choice`. A naive replacement with `rng.choice`, is almost 2 orders of magnitude slower (if I read these units correctly...).

```
%%timeit
random.choice(a)
181 ns ± 9.22 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

%%timeit
rng.choice(a)
7.07 μs ± 69.4 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```
This is, in fact, a known issue. There are various issues related to this, both in the NumPy git repository and on Stack Overflow. One often made suggestion is to use `rng.integers` if you want to choose only a single random item from a collection. This does make a difference:

```python
%%timeit
a[rng.integers(0, len(a))]
892 ns ± 4.18 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

So still almost an order of magnitude slower, but at least we are in the same time units again. This is the version I have used in this PR.

So how does all this relate to the reported benchmarks? Most of our benchmark models rely on random walks, random selections of agents, or random selection of a cell to move to. All of these rely on `rng.choice`/`some_collection[rng.integers(0, len(some_collection))]`. Thus, it is not surprising that the benchmarks are what they are. However, I would caution against over-interpreting this. Many real-world models in my experience don't rely as much on `random.choice`. 

Additionally, I believe there may be room for further improvements. In particular, the random choice inside cell collections can probably be tuned a bit more by doing something similar to the earlier example of `rng.random` by wrapping `rng.integers` in a generator. 

I remain in favor of proceeding with this PR. The proper control of random number generation, avoiding race conditions, and the higher quality of the random numbers are all crucial for the proper use of ABMs.

EwoutH:
What would you like to do with this PR?

quaquel:
> What would you like to do with this PR?

Deprecating seed as a keyword argument on model already helps a lot. 

I still would prefer standardizing on a single random number generator, but the performance on `rng.choice` is just atrocious.

EwoutH:
> Deprecating seed as a keyword argument on model already helps a lot.

Could you open a PR for that?

quaquel:
I am closing this but might revisit it when working on mesa 4

PR #2898, opened 2025-11-29
Add AI/LLM use guidelines to Code of Conduct

EwoutH:
Set expectations for meaningful AI use in contributions:
- Contributors must understand and be able to explain their work
- AI should support learning, not replace critical thinking
- Low-effort AI-generated submissions are not acceptable

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -36,6 +36,16 @@ Examples of unacceptable behavior include:
 * Other conduct which could reasonably be considered inappropriate in a
   professional setting
 
+### AI and LLM Use
+
+AI tools can be valuable for learning, exploring ideas, and accelerating work. We welcome their thoughtful use. To ensure contributions remain meaningful and contributors continue to grow, we ask that you:
+
+* Understand what you submit. Be able to explain and defend any contribution you make. If you cannot, it is not ready to submit.
+* Stay in the driver's seat. Use AI to support your learning and work, not to replace the effort and critical thinking that make you a better developer.
+* Engage meaningfully. Low-effort, AI-generated submissions (issues, PRs, or proposals) without genuine personal engagement are not acceptable.
+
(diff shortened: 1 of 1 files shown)


colinfrisch (approved):
I think it's a good message to send !

adamamer20 (approved):

EwoutH:
Thanks for the reviews. I will give others a bit of time to comment/review, and with no objections will merge tomorrow.

CODE_OF_CONDUCT.md
```diff
@@ -36,6 +36,16 @@ Examples of unacceptable behavior include:
 * Other conduct which could reasonably be considered inappropriate in a
   professional setting
 
+### AI and LLM Use
+
+AI tools can be valuable for learning, exploring ideas, and accelerating work. We welcome their thoughtful use. To ensure contributions remain meaningful and contributors continue to grow, we ask that you:
```
jackiekazil, line 41: I would make one change -- add something explicit about how the AI/LLM is being use -- 

- Is the AI being used to write code?
- Is the AI being used to write the PR? 

It doesn't need to be an exhaustive list - it is just unclear how the AI is being applied.

EwoutH: Thanks for your input. I don’t think you can list what’s “right” and “wrong” use of AI, It depends too much on the situation, skill and complexity.

quaquel: At TU Delft, we ask students to add a short section on AI use to e.g. a thesis. Here they must indicate how generative AI was used. This might offer a middle ground and is something we can add to the PR template.

EwoutH: Something worth considering I think. The level of detail might be hard to get right.

EwoutH:
For now I’m merging this. We can always update this and/or the templates in future PRs.

PR #2872, opened 2025-11-02
Deprecate `agent_portrayal` and update Migration guide

Sahil-Chhoker:
### Summary
Updates the migration guide inline with the new visualization API. Also activates the deprecation warnings for portrayal components.

### Motive
Updating the migration guide to help users upgrade to the latest Mesa version.

### Implementation
Documented API changes with a comparison between the old and new versions.

PR #2900, opened 2025-11-29
docs: Add deprecation policy to contributor guide

EwoutH:
#### Summary
Adds a new "Deprecation Policy" section to our contributor guide, establishing clear guidelines for how we deprecate features in Mesa.

#### Motive
As discussed in #1909 and various PR reviews (e.g., #2841 and #2872), we've been making ad-hoc decisions about deprecation practices. This has led to repeated discussions about which warning type to use, when deprecations can be added, and what prerequisites are needed. Having a documented policy will streamline future PRs and ensure consistent user experience.\

Please review critically:
- if you agree with the deprecation policy as described
- if it's clear what needs to be done to deprecate a feature (so if it's a good reference)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -200,67 +200,68 @@ A recorded video of this process is [available here](https://youtu.be/JE44jkegmn
 ### Deprecation policy
 Mesa follows [Semantic Versioning](https://semver.org/) (SemVer) and has a structured deprecation process to ensure users have time to migrate to new APIs while maintaining backward compatibility.
 
-#### Prerequisites for Deprecation
-Before deprecating any feature, all of the following must be complete:
+#### Deprecation process
+##### Step 1: Introduce alternative
+When implementing a replacement feature:
 
-1. Stable alternative available: The replacement feature must be fully implemented and stable (not experimental).
-2. Documentation updated: All relevant docs and tutorials must reflect the new approach.
-3. Migration guide entry added: A clear entry in the [Migration Guide](https://github.com/projectmesa/mesa/blob/main/docs/migration_guide.md) explaining what changed and how to update code.
-4. Examples updated: All example models in the repository must use the new API.
-
-#### Deprecation Timeline
-Mesa uses a two-phase deprecation process:
-
-**Phase 1: Introduction of Alternative**
-
-When a new alternative is implemented, all four items from the checklist above _may_ already be added. In addition, you may add a `PendingDeprecationWarning` to the old feature. This signals to developers that a change is coming but doesn't yet require immediate action.
+- Add the new, stable alternative
+- Optionally add a `PendingDeprecationWarning` to the old feature (signals intent in source code, hidden by default)
 
 ```python
 warnings.warn(
-    "This feature will be deprecated in a future release. "
-    "Consider using new_feature() instead.",
+    "old_feature() will be deprecated in a future release. "
+    "Use new_feature() instead.",
     PendingDeprecationWarning,
     stacklevel=2,
 )
 ```
 
-**Phase 2: Active Deprecation**
-The new alternative must be available for _at least one minor release_ before the old feature receives a `DeprecationWarning`. All four items from the checklist above _must_ be implemented. When adding the deprecation warning:
+##### Step 2: Complete prerequisites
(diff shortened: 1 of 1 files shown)


jackiekazil:
LGTM - thank you for adding this.

Sahil-Chhoker:
I'm a little confused, like in #2893 I've changed the code and examples but the migration guide and tutorials remain as is, following this document I should also update the docs and migration guide but then won't the users be unable to see the API before it was changed and won't be able to follow the tutorials.

I was thinking of updating the tutorials and migration guide once the change will be changed to `DeprecationWarning` from `PendingDeprecationWarning`.

EwoutH:
Great question! The prerequisites don’t all need to happen at once, they can be done anywhere between adding the alternative and the final deprecation. Notice the difference between _may_ and _must_ in the policy. The key is that everything is in place before/when we add the `DeprecationWarning`.

So your approach in #2893 is fine: add the alternative with a `PendingDeprecationWarning`, then update docs/tutorials/migration guide in follow-up PRs before/when switching to `DeprecationWarning`.

Regarding visibility of the old API: these changes are to our `main` branch, but tagged releases and their docs remain available. Users on older versions can always access the corresponding documentation.

quaquel:
Just one clarifying question: why use `PendingDeprecationWarning` and `DeprecationWarning` instead of using `FutureWarning`?

EwoutH:
@quaquel This is actually an interesting question. I looked into this and the distinction comes down to intended audience:

- `DeprecationWarning`/`PendingDeprecationWarning`: For developers, hidden by default
- `FutureWarning`: For end users, shown by default

For Mesa, our users actively write code against our API: they subclass `Model` and `Agent`, call our methods directly. They're the developers who need to update their code, and they *can* act on warnings.

So it depends a bit if you see them as end-users or developers. They're probably both.

Libraries with similar dynamics (NumPy, pandas, scikit-learn) use `FutureWarning` for public API changes for this reason. With the current two-stage approach, users running their models normally would never see any warning unless they explicitly enable them or run pytest.

While `DeprecationWarning` is shown for code in `__main__`, this doesn't really help for library deprecations. The deprecated methods live in Mesa's modules, not in the user's script, so even a single-file model triggers warnings from imported code. The `__main__` exception essentially never applies here.

I would like to check this though, because I do remember having seen `DeprecationWarning`s here and there when building models.

But I think we should indeed update to `FutureWarning` for public API changes. Learned something today!

EwoutH:
pandas has a whole spec around this: [PDEP-17](https://pandas.pydata.org/pdeps/0017-backwards-compatibility-and-deprecation-policy.html). They start with `DeprecationWarning`, then switch to `FutureWarning` in the last minor release before removal. This gives library maintainers early warning (via pytest/CI) while ensuring end users see it before it breaks.

We could adopt a similar pattern: `DeprecationWarning` for initial deprecation, switching to `FutureWarning` one release before removal (or one minor release after the replacement is fully available).

quaquel:
> So it depends a bit if you see them as end-users or developers. They're probably both.

This is something I have been struggling with myself as well. I see them more as end-users, but technically they are developers since they use our library in their own code.

> While DeprecationWarning is shown for code in __main__, this doesn't really help for library deprecations. The deprecated methods live in Mesa's modules, not in the user's script, so even a single-file model triggers warnings from imported code. The __main__ exception essentially never applies here.

If you run your own custom mesa model from within main, as any sane person should do, you'll get your deprecation warnings wherever we put them. What matters is just that you call mesa stuff that raises a deprecation warning directly or indirectly from a main. 

>  We could adopt a similar pattern: DeprecationWarning for initial deprecation, switching to FutureWarning one release before removal (or one minor release after the replacement is fully available).

I agree that this is the best approach going forward.

EwoutH:
I validated the behavior above, it's correctly described.

I also thought about it a bit more, and `DeprecationWarning` is the most unpredictable: If they pop-up depends on how you structure your code. So let's not use them, and let's only use `PendingDeprecationWarning` (as placeholder) and `FutureWarning`.

I rewrote the guide to reflect that, and also made the order a bit more clear (it's now 4 logical steps).

Please rereview.

EwoutH:
Merging, future additions/changes can be made later.

PR #2851, opened 2025-10-17
Add ContinuousStates and -Observables

EwoutH:
### Summary
Adds `ContinuousObservable` to mesa_signals, enabling agent states that change continuously over time with automatic threshold detection and signal emission. Demonstrates the feature with an enhanced Wolf-Sheep predator-prey model where energy depletes continuously rather than at discrete time steps.

### Motive
Agent-based models often need to represent continuously changing states like energy depletion, resource growth, or temperature changes. Currently, modelers must manually calculate state changes at each step or schedule discrete update events, leading to:
- Balancing trade-offs between update frequency and performance
- Difficulty detecting exact threshold crossings
- No standardized patterns for time-varying states
- Complex coordination of multiple changing states

This addresses discussion #2529 (Continuous States) and provides a foundational building block toward the behavioral framework outlined in discussion #2538, enabling more realistic agent behaviors driven by internal state dynamics.

### Implementation
**Core Components:**

1. **`ContinuousObservable` descriptor**: Extends `Observable` to track values that change over time according to a rate function. Uses lazy evaluation - values are only recalculated when accessed, making it efficient even with many agents.
2. **`ContinuousState` helper class**: Internal state tracker storing the current value, last update time, rate function, and threshold set. Handles time-based calculation and threshold crossing detection.
3. **Threshold management**: `HasObservables.add_threshold()` provides a clean API for registering callbacks when values cross specific thresholds (upward or downward). Uses the existing signal/observer pattern - thresholds emit `"threshold_crossed"` signals.
4. **Time integration**: Uses the new unified `model.time` (see #2903).

**Key design decisions:**
- Lazy evaluation prevents unnecessary calculations when states aren't accessed
- Threshold storage uses a set (values only), callbacks managed through signal subscriptions to avoid duplicate invocations
- Linear integration for simplicity; extensible to more complex methods
- Fully integrated with existing `Computed` properties and dependency tracking

### API usage examples
```python
# Continuous value with rate per time unit
energy = ContinuousObservable(initial_value: float, rate_func: (value, elapsed, agent) -> float)

# Set/adjust value (emits "change" + checks thresholds)
self.energy = 120.0
self.energy += 5.0

# Thresholds (fires "threshold_crossed" with signal.threshold, signal.direction)
self.add_threshold("energy", 0.0, callback)

# Subscribe to signals
self.observe("energy", "change", on_change)
self.observe("energy", "threshold_crossed", on_threshold)

# Computed properties that depend on observables
is_hungry = Computable()
[...]

PR #2903, opened 2025-11-30
Add `model.time` as universal source of truth for simulation time

EwoutH:
### Summary
Introduces `model.time` as the single, canonical source of simulation time in Mesa. All components (DataCollector, visualization, user code, simulators) can now rely on `model.time` to get the current simulation time, regardless of whether the model uses simple stepping or discrete event simulation.

### Motive
Previously, time in Mesa was fragmented across different locations depending on the simulation mode:
- Simple models used `model.steps` as a proxy for time
- Discrete event simulations stored time in `simulator.time`
- Components needing time had to implement fallback chains like:
  ```python
  def _get_time(self):
      if hasattr(model, "simulator") and hasattr(model.simulator, "time"):
          return model.simulator.time
      if hasattr(model, "time"):
          return model.time
      return float(model.steps)
  ```

This caused problems for features that need consistent time access, such as the `ContinuousObservable` in #2851 and the conceptual model of space discussed in #2585. The discussion in #2228 (which originated in https://github.com/projectmesa/mesa/pull/2223) established consensus that Mesa needs a universal truth for time.

### Implementation
**Model class changes:**
- Add `time: float` attribute initialized to `0.0`
- ~Add `step_duration: float` parameter (default `1.0`) controlling time advancement per step~
- Add `_simulator: Simulator | None` to track if a simulator controls time
- Modify `_wrapped_step()` to only auto-increment time when no simulator is attached

**Simulator class changes:**
- Remove internal `self.time` attribute
- Write directly to `model.time` during event execution
- Add deprecated `time` property delegating to `model.time` for backward compatibility
- Register with model via `model._simulator = self` in `setup()`
- Reset `model.time` to `start_time` in `reset()`

Design kept minimal: a simple attribute rather than a property or `RunControl` class, following YAGNI principles while solving the immediate problem.

### Usage Examples
Basic usage (time auto-increments with steps):
```python
model = Model()
model.step()
print(model.time)  # 1.0
model.step()
print(model.time)  # 2.0
```

~Custom step duration (useful for staged activation):~
```python
model = Model(step_duration=0.25)
model.step()
print(model.time)  # 0.25
model.step()
print(model.time)  # 0.5
```

With discrete event simulator (simulator controls time):
```python
model = Model()
simulator = DEVSimulator()
simulator.setup(model)

simulator.schedule_event_absolute(some_function, 2.5)
simulator.run_until(3.0)
print(model.time)  # 3.0
```

Components can now simply use `model.time`:
```python
class MyDataCollector:
    def collect(self, model):
        timestamp = model.time  # Always works, no fallbacks needed
```

### Compatibility
- `simulator.time` still works but emits a `DeprecationWarning` directing users to `model.time` (considering DEVS is still experimental, this is quite gentle)
- Code directly accessing `simulator.time` without triggering the property (unlikely) would break (again, experimental, and all test pass)

### Additional Notes
- Closes the implementation side of #2228
- Enables cleaner implementation of #2851 (`ContinuousObservable`)
- In the future we might want to allow `datetime` values for `time`.

### DIFF
--- a/mesa/experimental/devs/simulator.py
+++ b/mesa/experimental/devs/simulator.py
@@ -69,7 +69,7 @@ class Simulator:
         """Simulator time (deprecated)."""
         warnings.warn(
             "simulator.time is deprecated, use model.time instead",
-            DeprecationWarning,
+            FutureWarning,
             stacklevel=2,
         )
         return self.model.time


quaquel:
1. I suggest making `step_duration` protected: `_step_duration`. It is good that users can control it, but we reserve the right to modify how the controlling step duration works in the future without breaking Semantic Versioning. I still would like to think through a more detailed way for users to set up time, including support for date time and calendars. This will interact with `step_duration` and so I don't want to be locked into this yet. `model.time` is generic enough that I am fine with locking this in as part of the public api.
2. I am not sure about the API for the simulators. You now do

```python
model = Model()
simulator = DEVSimulator()
simulator.setup(model)
```

However, in the Wolfsheep example, I have moved `simulator.setup` into the `__init__` method of the `Model` class. I am inclined to make this the preferred api by adding `simulator` as a keyword argument to `Model`. and make `model.simulator` part of the public API of the model class if `simulator is not None`.

As an aside: what are people's thoughts on stabilizing the simulator stuff?

EwoutH:
Thanks for getting back.

> 1. I suggest making `step_duration` protected: `_step_duration`.

Maybe we need to think this a bit further through before going one way or the other. I also want to add `datetime` support for example.

Do what do you think about a separate `RunControl` class vs integrating it in the model? In this effort, I started with it, but realized it wasn't needed and mainly added API complexity.

> I am not sure about the API for the simulators. 

I just copied (previous) best-practices. It can be that's now outdated. Should I update it?

> As an aside: what are people's thoughts on stabilizing the simulator stuff?

Perfectly ok with it. Used it intensively without problems during my thesis.

quaquel:
1. I would suggest using composition for `RunControl`. So, a dedicated class that can be passed to `Model.__init__` or something along those lines. 
2. I suggest keeping this PR focused on just making `model.time` part of the public API and minimizing all other changes. Having `model.time` as the sole truth is valuable as such. Let's worry about expanding it in separate PR's.
3. The simulator stuff is fine as is for now. I'll make a separate PR asap as part of a move towards stabilizing it.

EwoutH:
> 1. I suggest making `step_duration` protected: `_step_duration`.

Done.

EwoutH:
@quaquel can you give this a formal review?

mesa/experimental/devs/simulator.py
```diff
@@ -61,9 +62,17 @@ def __init__(self, time_unit: type, start_time: int | float):
...
-        self.time = self.start_time
-        self.model = None
+        self.model: Model | None = None
+
+    @property
+    def time(self) -> float:
+        """Simulator time (deprecated)."""
+        warnings.warn(
+            "simulator.time is deprecated, use model.time instead",
+            DeprecationWarning,
```
quaquel: should this not be up dated in line with the new policy?

alternatively, should this actually raise a warning at all, or can we julst leave this as a shorthand (just as agent.random, agent.rgn being shorthands for agent.model.random/agent.model.rng?

EwoutH: > should this not be up dated in line with the new policy?

Yes, this should be a `FutureWarning` indeed (since this is still experimental we can skip the `PendingDeprecationWarning` step). Good catch!

> or can we julst leave this as a shorthand (just as agent.random, agent.rgn being shorthands for agent.model.random/agent.model.rng?

I think I would like there to be only one, explicit way (following [zen](https://peps.python.org/pep-0020/)).

> There should be one-- and preferably only one --obvious way to do it.

EwoutH:
```suggestion
            FutureWarning,
```

EwoutH:
Removed `step_duration` for now, to keep don't over invest into `steps` since we're moving to `time`.
- https://github.com/mesa/mesa/pull/3007

PR #2910, opened 2025-12-02
New issue 2.2

ShreyasN707:
This PR resolves multiple UserWarning messages caused by type mismatches in PropertyLayer instantiations within the test suite. The warnings were generated when integer defaults (e.g., 0) were used while the underlying data arrays were dtype=float64.

- Bug / Issue

Related to Issue: “2.2 PropertyLayer Default Value Type Mismatch (~4 warnings)”

The following warning appeared in several tests:

UserWarning: Default value 0 (int) might not be best suitable with dtype=float64.

This happened because test cases instantiated PropertyLayer with integer default values, while the internal representation expected float64. The mismatch produced noisy test output but did not affect functionality.

- Implementation

The fix updates the affected test cases to ensure that the default values passed to PropertyLayer explicitly match the expected dtype.

- Changes include:

Added import numpy as np where needed.

Replaced integer/float literals (0, 0.0, 10.0) with explicit np.float64() values.

Ensured all PropertyLayer defaults used in tests match dtype=float64.

- Modified files:

tests/test_components_matplotlib.py

tests/test_space_renderer.py

- Testing

Ran the full test suite using pytest — all tests pass.

Confirmed that the previous warnings are no longer emitted.

Used a small helper script (check_warnings.py) to assert that no dtype-related UserWarning appears.

Ran formatting and linting via pre-commit run --all-files to ensure code quality.

- Additional Notes

Fix applies only to test files; no changes were made to the PropertyLayer implementation.

This cleanup improves signal-to-noise in test output, making future debugging easier.

No functional or behavioral changes to Mesa’s core library.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ ci:
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.14.3
+  rev: v0.14.7
   hooks:
     # Run the linter.
     - id: ruff
@@ -14,7 +14,7 @@ repos:
     - id: ruff-format
       types_or: [ python, pyi, jupyter ]
 -   repo: https://github.com/asottile/pyupgrade
-    rev: v3.21.0
+    rev: v3.21.2
     hooks:
     -   id: pyupgrade
         args: [--py312-plus]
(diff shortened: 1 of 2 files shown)


ShreyasN707:
will fix my local copy and raise a new PR

PR #2929, opened 2025-12-05
tests: Set batch_run tests to use 100 steps instead of 1000

EwoutH:
Update all `mesa.batch_run()` calls to use `max_steps=100` and adjust expected test results accordingly (`Step: 100, agent_local: 25.0`).


EwoutH:
Never mind, they actually don't take up significant time. I had another issue.

PR #2931, opened 2025-12-05
Fix random number generator warnings

Mani212005:
Description
This PR addresses the random number generator warnings that were causing ~45+ UserWarnings across the test suite. The warnings indicated that models could be non-reproducible when RNG is not explicitly specified.

Changes Made:

1. Added shared rng fixture (tests/conftest.py)
    * Created a pytest fixture that provides a deterministic [random.Random(42)]( instance
    * Can be used across all test files for reproducible randomness
2. Fixed Conway's Game of Life example ([model.py]
    * Added [random=self.random]) parameter to initialization
    * Ensures the grid uses the model's seeded RNG for reproducibility
3. Added tests using the rng fixture (tests/test_discrete_space.py)
    * test_orthogonal_moore_grid_with_rng: Tests Moore grid with provided RNG
    * test_orthogonal_von_neumann_grid_with_rng: Tests Von Neumann grid with provided RNG
    * test_cell_agent_with_rng: Tests CellAgent with provided RNG

Testing:

pytest tests/test_discrete_space.py -v
# Result: 30 passed

pytest tests/ -v 2>&1 | grep -i "random number generator not specified"
# Result: No warnings found

Impact
✅ Eliminates all "Random number generator not specified" warnings
✅ Improves model reproducibility
✅ No breaking changes - backwards compatible
✅ Provides pattern for future test development with the rng fixture

Closes # 2904 part 2.1 . Thank you.


tests/conftest.py
quaquel, line 1: I would suggest moving this into the test_discrete_space.py file itself. As it is, we don't use common fixtures across tests.

quaquel:
superseded by #2983

PR #2934, opened 2025-12-08
Update dict-based agent portrayals to use `AgentPortrayalStyle` in tests

falloficaruss:
## Summary

This PR addresses part of issue #2904 (section 1.1) by updating the older visualization test suite to replace deprecated dict-based agent portrayals with `AgentPortrayalStyle` and add tests to confirm the emitted `FutureWarning` from older dict-based agent portrayals.

## Changes Made

- tests/test_examples_viz.py: Updated 9 agent portrayal functions to AgentPortrayalStyle.
- tests/test_solara_viz_updated.py: Updated dict-based portrayal to AgentPortrayalStyle.
- tests/test_backends.py: Added tests for deprecated dict-based portrayal to confirm they emit `FutureWarning`.

## Impact
- Eliminates FutureWarning: Returning a dict from agent_portrayal is deprecated messages from the test suite.
- Verifies deprecated dict-based portrayal still works (backward compatibility) while emitting the expected warning.
- All visualization tests passing.
- Code follows new Mesa API standards using AgentPortrayalStyle.

## Testing
Ran the following tests successfully:
```bash
pytest tests/test_examples_viz.py tests/test_solara_viz_updated.py tests/test_backends.py -v
```

---
**Changes by @Sahil-Chhoker**:

Updated the description to match the PR’s functionality.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -45,7 +45,7 @@ jobs:
         run: python global_benchmark.py
       # Upload benchmarks, checkout PR branch, download benchmarks
       - name: Upload benchmark results
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: timings-main
           path: benchmarks/timings_1.pickle
@@ -61,7 +61,7 @@ jobs:
       - name: Install Mesa of the PR branch
         run: pip install --no-deps .
       - name: Download benchmark results
-        uses: actions/download-artifact@v6
+        uses: actions/download-artifact@v7
         with:
           name: timings-main
           path: benchmarks
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,7 +41,7 @@ jobs:
         run: hatch build
       - name: Upload package as artifact to GitHub
         if: github.repository == 'mesa/mesa' && startsWith(github.ref, 'refs/tags')
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: package
           path: dist/
--- a/mesa/examples/advanced/epstein_civil_violence/app.py
+++ b/mesa/examples/advanced/epstein_civil_violence/app.py
@@ -64,8 +64,11 @@ chart_component = make_plot_component(
 )
 
 epstein_model = EpsteinCivilViolence()
-renderer = SpaceRenderer(epstein_model, backend="matplotlib")
-renderer.draw_agents(citizen_cop_portrayal)
+renderer = SpaceRenderer(epstein_model, backend="matplotlib").setup_agents(
+    citizen_cop_portrayal
+)
+# Specifically, avoid drawing the grid to hide the grid lines.
+renderer.draw_agents()
 renderer.post_process = post_process
 
 page = SolaraViz(
--- a/mesa/examples/advanced/pd_grid/app.py
+++ b/mesa/examples/advanced/pd_grid/app.py
@@ -45,7 +45,11 @@ plot_component = make_plot_component("Cooperating_Agents", backend="altair", gri
 # Initialize model
 initial_model = PdGrid()
 # Create grid and agent visualization component using Altair
-renderer = SpaceRenderer(initial_model, backend="altair").render(pd_agent_portrayal)
+renderer = (
+    SpaceRenderer(initial_model, backend="altair")
+    .setup_agents(pd_agent_portrayal)
+    .render()
+)
 
 # Create visualization with all components
 page = SolaraViz(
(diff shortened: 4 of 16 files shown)


EwoutH:
Thanks!

Would you like to give a shot at fixing all remaining agent_portrayal warnings?

falloficaruss:
> Thanks!
> 
> Would you like to give a shot at fixing all remaining agent_portrayal warnings?

Sure, will try

falloficaruss:
Hey @EwoutH , please check if the fixes I applied are correct and worth merging this PR.

tests/test_backends.py
```diff
@@ -67,13 +67,6 @@ def agent_portrayal_style(agent):
     data = mb.collect_agent_data(DummySpace(), agent_portrayal_style)
     assert "loc" in data and data["loc"].shape[0] == 1
 
-    # Test with dict-based portrayal
-    def agent_portrayal_dict(agent):
-        return {"size": 5, "color": "red", "marker": "o"}
-
-    data = mb.collect_agent_data(DummySpace(), agent_portrayal_dict)
-    assert "loc" in data and data["loc"].shape[0] == 1
-
```
EwoutH, line 76: I think we can do here:
```python
    # Test with dict-based portrayal (deprecated, emits FutureWarning)
    def agent_portrayal_dict(agent):
        return {"size": 5, "color": "red", "marker": "o"}

    with pytest.warns(FutureWarning):
        data = mb.collect_agent_data(DummySpace(), agent_portrayal_dict)

    assert "loc" in data and data["loc"].shape[0] == 1
```
That way we make sure that the old behavior also keeps working until we remove it.

tests/test_backends.py
```diff
@@ -203,13 +196,6 @@ def agent_portrayal_style(agent):
     data = ab.collect_agent_data(DummySpace(), agent_portrayal_style)
     assert "loc" in data and data["loc"].shape[0] == 1
 
-    # Test with dict-based portrayal
-    def agent_portrayal_dict(agent):
-        return {"size": 5, "color": "red", "marker": "o"}
-
-    data = ab.collect_agent_data(DummySpace(), agent_portrayal_dict)
-    assert "loc" in data and data["loc"].shape[0] == 1
-
```
EwoutH, line 212: Idem dito
```Python
    # Test with dict-based portrayal (deprecated, emits FutureWarning)
    def agent_portrayal_dict(agent):
        return {"size": 5, "color": "red", "marker": "o"}

    with pytest.warns(FutureWarning):
        data = ab.collect_agent_data(DummySpace(), agent_portrayal_dict)

    assert "loc" in data and data["loc"].shape[0] == 1
```

tests/test_backends.py
falloficaruss: Okay, will work on it.

tests/test_backends.py
falloficaruss: This too!

falloficaruss:
Hey @EwoutH verified the backward compatibilty too.

EwoutH:
@Sahil-Chhoker I think it looks good, could you give a final check?

tests/test_examples_viz.py
```diff
@@ -192,8 +194,9 @@ def test_boltzmann_wealth_model(solara_test, page_session: playwright.sync_api.P
     model = BoltzmannWealth(seed=42)
 
     def agent_portrayal(agent):
-        color = agent.wealth  # we are using a colormap to translate wealth to color
-        return {"color": color}
+        return AgentPortrayalStyle(
+            color="tab:purple" if agent.wealth > 0 else "tab:grey"
+        )
```
Sahil-Chhoker: Why aren't you setting color to `agent.wealth` here, this is a crucial test to check if the backend is properly converting `int` values to a colormap, but now you are just returning two colors.

falloficaruss: Thanks for catching that out, I'm trying to fix it now

falloficaruss:
Hey @Sahil-Chhoker , I've updated the test to use color=agent.wealth. I came across a bug in mpl_space_drawing.py: the backend was attempting to copy the scalar color value (an integer in this case) directly to edgecolors, which caused Matplotlib to crash.

I’ve pushed a fix for that backend issue along with the updated test, so AgentPortrayalStyle now correctly handles scalar values for colormapping."

Let me know if I'm wrong or if I need to change anything.

Sahil-Chhoker:
Sorry for the delay @falloficarus22, and thanks for tracking the bug, this is indeed the right fix for it but can you make a separate PR with just the fix since this PR is only for the tests, first we'll merge the fix then this PR.

falloficaruss:
> Sorry for the delay @falloficarus22, and thanks for tracking the bug, this is indeed the right fix for it but can you make a separate PR with just the fix since this PR is only for the tests, first we'll merge the fix then this PR.

Okay, will do.

falloficaruss:
> Good Work!

Thanks!

Issue #2722, opened 2025-03-15
SolaraViz Visualisation won't start if Mesa-Model's __init__-parameters are not set as keyword-arguments.


Lezhor:
I'm not sure if this can be categorized as a Bug Report since I did find the cause for the bug. But since I don't have the time to create a Pull-Request of my own I'm gonna describe the possible solution here for you.

**Describe the bug**
So I was following the Introduction Visualisation Tutorial and stumbled onto something: My `MoneyModel` constructor looked like this:
```python
class MoneyModel(Model):
    def __init__(self, n: int, width: int, height: int, seed: float | None = None) -> None:
        super().__init__(seed=seed)
        # Other code
```

I later noticed the Tutorial explicitly said to set the arguments to default values using kwargs like this:
```python
class MoneyModel(Model):
    def __init__(self, n: int = 10, width: int = 10, height: int = 10, seed: float | None = None) -> None:
        super().__init__(seed=seed)
        # Other code
```
The first variant threw the following Error Message whilst the second one did not:
```
  File "C:\<path-to-venv>\mesa\visualization\solara_viz.py", line 427, in <lambda>
    lambda: _check_model_params(model.value.__class__.__init__, fixed_params),
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\<path-to-venv>\mesa\visualization\solara_viz.py", line 480, in _check_model_params
    raise ValueError(f"Missing required model parameter: {name}")
ValueError: Missing required model parameter: n
```

**Expected behavior**
I don't see how it must be necessary to specify constants in the Model (keyword arguments) for SolaraViz to work. I mean - The parameters are passed via the `model_parameters` object one needs to define during the Tutorial anyways.

**To Reproduce**
Already explained in the Sections above. Also I use mesa 3.1.4 and if you need to know the versions of other correlating libraries, here they are:
```toml
[tool.poetry.dependencies]
python = "^3.12"
simpy = "^4.1.1"
numpy = "^2.2.3"
matplotlib = "^3.10.1"
mesa = {extras = ["rec"], version = "^3.1.4"}
seaborn = "^0.13.2"
solara = "^1.44.1"
```

**How to Fix this Issue?**

I actually dug into the SolaraViz Code and found the source of the problem. 

In `solara_vis.py` in the `ModelCreator` function there is this Code-Snippet:

[...]

PR #2941, opened 2025-12-10
Add test for SolaraViz with required model parameters

ShreyasN707:
- Summary

Adds a minimal regression test ensuring that SolaraViz correctly handles models with required constructor parameters when those parameters are supplied through model_params.

Related Issue : #2722

- Background

Users previously reported a ValueError: Missing required model parameter when using models with required parameters (e.g., MoneyModel(n, width, height)), even though the parameters were passed via model_params.
This behavior has already been fixed in the current codebase.

- What This PR Does

This PR adds a minimal test to confirm the fix.
No code changes were required.

- Test Added :
   tests/test_solara_required_params.py

Result:

     1 passed, 1 warning in 4.78s

### DIFF
--- a/tests/test_solara_required_params.py
+++ b/tests/test_solara_required_params.py
@@ -3,6 +3,7 @@
 This reproduces the issue where models with required parameters (no default values)
 should work with SolaraViz when those parameters are provided via model_params.
 """
+
 import solara
 
 from mesa import Model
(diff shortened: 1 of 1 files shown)


ShreyasN707:
This PR had some previous unwanted commits . I have raised a new PR #2942

PR #2942, opened 2025-12-10
Added test for SolaraViz with required model parameters

ShreyasN707:
- Summary

Adds a minimal regression test ensuring that SolaraViz correctly handles models with required constructor parameters when those parameters are supplied through model_params.

Related Issue : #2722

- Background

Users previously reported a ValueError: Missing required model parameter when using models with required parameters (e.g., MoneyModel(n, width, height)), even though the parameters were passed via model_params.
This behavior has already been fixed in the current codebase.

- What This PR Does

This PR adds a minimal test to confirm the fix.
No code changes were required.

- Test Added :
   tests/test_solara_required_params.py

Result:

     1 passed, 1 warning in 4.78s


ShreyasN707:
@EwoutH @quaquel Please review my PR.

tests/test_solara_required_params.py
```diff
@@ -0,0 +1,50 @@
...
+    # The parameters are provided via model_params
+    try:
+        solara.render(SolaraViz(model, model_params=model_params), handle_error=False)
+        print("✓ Test passed: SolaraViz accepts required parameters via model_params")
+    except ValueError as e:
+        if "Missing required model parameter" in str(e):
+            print(f"✗ Test FAILED: {e}")
+            raise
+        else:
+            raise
```
quaquel: I don't believe this is the proper way to write these tests.

1. Please consider adding this to the test_solara_viz.
2. Please test the code for both raising and not raising the exception.

ShreyasN707: Okay I’ll move the test into test_solara_viz and rewrite it using proper pytest assertions.  I’ll also add explicit coverage for both cases: when the exception should be raised and when it shouldn’t . I’ll push an update shortly.

ShreyasN707: @quaquel I have updated the tests.

quaquel:
this is undergoing changes because of the introduction of scenarios. Once this is fixed, we should look at the testing again.

PR #2943, opened 2025-12-10
Fix links from mesa renaming. (#2880)

jackiekazil:
A few minor links needed updating from the migration.


EwoutH (approved):
Thanks! Missed these because the where added after I opened the PR. Should have done a rebase.

Great catch!

PR #2946, opened 2025-12-11
Fix for Model.reset_rng

### LINKED ISSUE #2940: Model.reset_rng() creates new RNG instead of restoring initial state
I noticed the RNG behavior in Model, Calling reset_rng() without any arguments doesn’t return the numpy RNG to the state it had when the model was created. Instead, it seems to create a completely new RNG instance seeded from system entropy.

This makes the RNG sequence after a reset different from the sequence right after initialization, even when the model is created with a fixed seed. Since the docstring mentions “if None, reset using the current seed,” this looks unintentional.

The model also stores the initial RNG state in self._rng, but that saved state isn’t used during reset.

- Expected behavior

If a model is initialized with a deterministic seed, calling:

```python
     model.reset_rng()
```

should bring the numpy RNG back to its initial state so that the sequence of random values matches what we would get from a fresh model run.

- To Reproduce
```python
from mesa import Model

model = Model(seed=42)
initial = model.rng.random()

# Move RNG forward
for _ in range(100):
    model.rng.random()

# Try to reset back to initial state
model.reset_rng()

after_reset = model.rng.random()

print(initial, after_reset)
# These two values do not match
```

This happens consistently: the RNG after reset does not match the original starting state.

@EwoutH I can work on this! Please let me know.

quaquel:
### Summary
`Model.reset_rng` did not correctly reset to the original seed if called with `seed=None`. 

### Bug / Issue
This closes #2940. 

### Implementation
Use `Model._rgn`, which contains the state information about the numpy random number generator that is being used. The code ensures that not only the seed value is reused, but also the correct `BitGenerator` is used.

### Testing
added 2 tests that cover the bug

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -13,31 +13,31 @@ We're excited to welcome six new contributors to the Mesa project in this releas
 
 ## What's Changed
 ### 🐛 Bugs fixed
-* Fix visualization error for PropertyLayers on HexGrids. Add transpose … by @flucco in https://github.com/projectmesa/mesa/pull/2868
-* Fix: Property layer data mapping for both Altair and Matplotlib backends by @Sahil-Chhoker in https://github.com/projectmesa/mesa/pull/2869
-* Fix: AgentSet initialization should not require explicit random number generator by @verisimilidude2 in https://github.com/projectmesa/mesa/pull/2789
+* Fix visualization error for PropertyLayers on HexGrids. Add transpose … by @flucco in https://github.com/mesa/mesa/pull/2868
+* Fix: Property layer data mapping for both Altair and Matplotlib backends by @Sahil-Chhoker in https://github.com/mesa/mesa/pull/2869
+* Fix: AgentSet initialization should not require explicit random number generator by @verisimilidude2 in https://github.com/mesa/mesa/pull/2789
 ### 🔍 Examples updated
-* Fix missing variable g in the tutorial by @pazmiller in https://github.com/projectmesa/mesa/pull/2849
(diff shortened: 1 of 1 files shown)


mesa/model.py
```diff
@@ -231,8 +232,19 @@ def reset_rng(self, rng: RNGLike | SeedLike | None = None) -> None:
...
+        bit_generators = {
+            entry.__name__: entry
+            for entry in [PCG64, PCG64DXSM, MT19937, Philox, SFC64]
+        }
+
+        if rng is None:
+            rng = self._rng
+            alg = bit_generators[rng["bit_generator"]]()
+            alg.state = rng
+            self.rng = np.random.default_rng(alg)
```
EwoutH: Could we do:
```suggestion
        if rng is None:
            # Restore from saved initial state
            bg_class = getattr(np.random, self._rng["bit_generator"])
            bg = bg_class()
            bg.state = self._rng
            self.rng = np.random.Generator(bg)
```

quaquel:
@EwoutH clever solution, now I just need an approval and this can be merged.

quaquel:
test seems to be failing will check in an hour

EwoutH:
It seems to be the `codecov`config, not the tests. Probably due the org name change.

quaquel:
The tests are failling because of problems with uploading the coverage information. @EwoutH this might be related to the mesa move. Do you have time to check?

EwoutH:
Tried some stuff, didn't help.

Apparently it sometimes takes a few days after a organization name change. We can ignore it for now.

macOS is an upstream NetworkX meets Python 3.14.1 bug. Will be fixed upstream.

We don't have any required checks, so go ahead and merge.

PR #2950, opened 2025-12-12
Refactor: Replace generic Exception with RuntimeError in Cell.add_agent

falloficaruss:
## Description

 This PR addresses a FIXME in mesa/discrete_space/cell.py by replacing a generic Exception with a more specific RuntimeError.

## Changes

- Modified Cell.add_agent to raise RuntimeError("Cell is full") instead of Exception("ERROR: Cell is full") when attempting to add an agent to a cell that has reached its capacity.

## Motivation:

Using specific exception types allows for better error handling and testing. RuntimeError is semantically more appropriate here as it indicates a state error where the operation cannot be performed due to the current state of the cell (being full).

## Summary by CodeRabbit

* **Chores**
  * Improved error handling for cell capacity constraints by using more specific exception types.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


falloficaruss:
@coderabbitai review

falloficaruss:
@coderabbitai review

quaquel:
I am not sure using a `RuntimeError` is the appropriate solution here. One reason for the FIXME was that I am inclined to add a bunch of exceptions to MESA itself.

mesa/discrete_space/cell.py
```diff
@@ -110,9 +110,7 @@ def add_agent(self, agent: CellAgent) -> None:
         self.empty = False
 
         if self.capacity and n >= self.capacity:
-            raise Exception(
-                "ERROR: Cell is full"
-            )  # FIXME we need MESA errors or a proper error
+            raise RuntimeError("Cell is full")
```
EwoutH: I think that the type of error isn't the primary concern, but the message. I think something like this would already be far more useful.
```suggestion
            raise Exception(
                f"Cannot add agent {agent.unique_id} to cell at {self.coordinate}: "
                f"cell is full (capacity: {self.capacity}, current: {n})"
            )  # FIXME we might want dedicated MESA errors types
```
@quaquel if we make a Mesa error type, could we use that to (automatically) include some information about:
- the model state (`model.time` for example by default)
- where the error occurred (which agent, model, space, etc.)

quaquel: Not sure. I tend to keep errors simple with just a message. Including where is not needed because you already have the full stack trace, so you know where it occurred.

quaquel:
I have opened a discussion on mesa exceptions in the mesa 4.0 discussion. I am closing this pull request, but I have included a link to it for reference. Depending on how we decide to move forward, we might reopen this or at least draw heavily from it.

PR #2408, opened 2024-10-23
Cleanup and restructure EpsteinCivilViolence and PdGrid examples

EwoutH:
Cleanup and restructure the EpsteinCivilViolence and PdGrid advanced example models. This PR is similar to what was done in the basic examples in https://github.com/projectmesa/mesa/pull/2365.

It draws the discrete spaces as was supported in https://github.com/projectmesa/mesa/pull/2386 and updates the imports to be absolute, similar to https://github.com/projectmesa/mesa/pull/2402.

EpsteinCivilViolence and PdGrid now:
- Have all their Agent code in an `agents.py` file.
- Use a flat structure.
- Use the newest SolaraViz API
- Have the old visualization code removed

Screenshot_522
Screenshot_523

A few things to note:
- Handeling agent_portrayals is still finicky
- I couldn't get the square shape to work
- EpsteinCivilViolence shows remarkable stable behavior. Is this correct?
- Claude 3.5 Sonnet went full YOLO and found a way to get a choice selector to work:
    ```Python
    "activation_order": {
        "type": "Select",
        "value": "Random",
        "values": PdGrid.activation_regimes,
        "label": "Activation Regime"
    }
    ```
    @Corvince Is this a proper approach? If so, does that mean that #2376 isn't needed anymore?
- Real interesting changes in behavior between the random, sequential and staged activation regimes! Sequential and staged quickly go to all cooperating or all non-cooperating. Maybe we should double check if this works as intended.

Issue #2376, opened 2024-10-16
Viz: Implement Choice UserParam


EwoutH:
The old visualisation contained a Choice class which you could use to select one of multiple discrete/categorial options.

https://github.com/projectmesa/mesa/blob/53ce2945a1dfaca526e3b64965672940bec6f7a9/examples/advanced/pd_grid/pd_grid/server.py#L12-L16

It would be nice to have that in the new SolaraViz module. Maybe an boolean Checkbox would also be useful. For full context, NetLogo has [these options](https://ccl.northwestern.edu/netlogo/docs/interfacetab.html) you can place:

![image](https://github.com/user-attachments/assets/dd72904f-6360-4730-9dec-d6591425116d)

@Corvince would you take a shot at this?

quaquel:
Many of these are already supported in the current code pace. Whe have space visualization and line plots for output. Text based output is available in solara and demonstrated in various examples (e.g. virus on a network). 

SolaraViz supports Sliders for both int and float, a drop-down menu bar for select, and a selection via a checkbox. 

So what is missing in your view?

PR #2955, opened 2025-12-16
examples: Add Activation Order and Grid Type selectors to EpsteinCivilViolence

Nithurshen:
### Summary

This PR adds interactive selectors to the `EpsteinCivilViolence` example, allowing users to dynamically choose the Activation Order (Random vs. Sequential) and Grid Type (Moore vs. Von Neumann) directly from the Solara GUI.

* Follow-up to discussion in PR #2408
* Demonstrates usage of selectors mentioned in #2376

### Motive

In the discussion for PR #2408, maintainer @quaquel noted that the activation order (Random vs. Sequential) significantly impacts model stability and suggested making it a user input. Additionally, it was observed that swapping the grid topology (Moore vs. Von Neumann) drastically changes agent behavior. Exposing these parameters allows users to explore these dynamics easily without modifying the code.

### Implementation

- `model.py`: Updated `EpsteinCivilViolence.__init__` to accept `activation_order` and `grid_type` arguments. Added logic to initialize the corresponding Grid class (`OrthogonalMooreGrid` or `OrthogonalVonNeumannGrid`) and handle the scheduling logic in the `step()` method.

- `app.py`: Added `activation_order` and `grid_type` to `model_params` as `Select` widgets, enabling the dropdown menus in the Solara sidebar.

- `model.py` : Updated docstrings to reflect the new parameters.

### Usage Examples

**Before**:
before

**After:**
after

Run the example:

```Bash
solara run mesa/examples/advanced/epstein_civil_violence/app.py
```
In the sidebar, use the Activation Order dropdown to switch between "Random" and "Sequential".

Use the Grid Type dropdown to switch between "Moore" (8 neighbors) and "Von Neumann" (4 neighbors).

### Additional Notes

- Defaults are set to Von Neumann and Random to match the behavior established in recent updates.

- "Sequential" activation relies on `self.agents.do("step")`, while "Random" uses `self.agents.shuffle_do("step")`.

### DIFF
--- a/docs/tutorials/10_comparing_scenarios.ipynb
+++ b/docs/tutorials/10_comparing_scenarios.ipynb
@@ -14,14 +14,14 @@
    "metadata": {},
    "source": [
     "If you want to get straight to the tutorial checkout these environment providers:<br>\n",
-    "(with Google Account) [![Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/mesa/mesa/blob/main/docs/tutorials/8_comparing_scenarios.ipynb)<br>\n",
-    "(No Google Account) [![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/mesa/mesa/main?labpath=docs%2Ftutorials%2F8_comparing_scenarios.ipynb) (This can take 30 seconds to 5 minutes to load)\n",
+    "(with Google Account) [![Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/mesa/mesa/blob/main/docs/tutorials/10_comparing_scenarios.ipynb)<br>\n",
+    "(No Google Account) [![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/mesa/mesa/main?labpath=docs%2Ftutorials%2F10_comparing_scenarios.ipynb) (This can take 30 seconds to 5 minutes to load)\n",
     "\n",
     "*If you are running locally, please ensure you have the latest Mesa version installed.*\n",
     "\n",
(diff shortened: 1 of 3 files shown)


Nithurshen:
The doc build failed on `tutorials/1_adding_space.ipynb` with a 'Cell is full' error, which seems unrelated to my changes in the Epstein example.

mesa/examples/advanced/epstein_civil_violence/model.py
```diff
@@ -48,14 +51,22 @@ def __init__(
...
     ):
         super().__init__(seed=seed)
         self.movement = movement
         self.max_iters = max_iters
+        self.activation_order = activation_order
 
-        self.grid = mesa.discrete_space.OrthogonalVonNeumannGrid(
-            (width, height), capacity=1, torus=True, random=self.random
-        )
+        if grid_type == "Moore":
```
quaquel: consider changing both this if else statement and the one in step to a match case structure:

```python
match activation_order:
    case "Von Neumann":
        pass. # your original if code
    case "Moore"
        pass #your original else code
    case _:
        raise ValueError("unknown value of activation_order")
```

This ensures that we get a value error if an invalid value is passed.

quaquel:
Thanks for this. I have merged main back into your branch which should solve the failing docs. I also have one suggestion. Looks good to me otherwise.

mesa/examples/advanced/epstein_civil_violence/model.py
Nithurshen: Done. Please review it now.

EwoutH:
Thanks for the PR! I appreciated the really clean diff and detailed description with screenshots.

quaquel:
Thanks!

PR #2963, opened 2025-12-17
Fix PropertyLayer default value type mismatch warnings

falloficaruss:
### Summary
Fixed false warnings in `PropertyLayer` when _default_value_ is compatible with the specified _dtype_. The previous implementation used `isinstance()` checks that incorrectly flagged compatible values (e.g., `float 10.0` with `dtype=float` or `np.float64`) as type mismatches. This fix eliminates ~4 false warnings in test files while maintaining proper warnings for actual incompatibilities and precision loss cases.

### Bug / Issue
The `PropertyLayer` class was generating false `UserWarning` messages about _dtype_ mismatches when the _default_value_ was actually compatible with the specified _dtype_. This occurred because:

1. **Root Cause**: The code used `isinstance(default_value, dtype)` which doesn't work correctly with _NumPy dtypes_. For example:
   - `isinstance(10.0, float)` returns `True` (correct)
   - `isinstance(10.0, np.float64)` returns `False` (incorrect, even though 10.0 is compatible)

2. **Impact**: This caused false warnings in multiple test files:
   - `tests/test_space_renderer.py:46` - `PropertyLayer("test", [2, 2], default_value=0, dtype=int)`
   - `tests/test_solara_viz.py:121` - `PropertyLayer(..., default_value=10.0, dtype=float)`
   - `tests/test_solara_viz_updated.py:136` - `PropertyLayer(..., default_value=10.0, dtype=float)`
   - `tests/test_components_matplotlib.py:240` - `PropertyLayer(..., default_value=0, dtype=int)`

3. **Expected Behavior**: Warnings should only appear for actual incompatibilities (e.g., _string_ -> _int_) or precision loss (e.g., `10.5` -> `int`).

### Implementation
Replaced the `isinstance()`-based type checking with _NumPy-aware dtype_ compatibility checking in both `PropertyLayer` implementations:

**Files Modified:**
- `mesa/discrete_space/property_layer.py`
- `mesa/space.py`
- `tests/test_discrete_space.py`

**Key Changes:**
1. **Dtype Normalization**: Use `np.dtype(dtype)` to normalize both Python types (`int`, `float`) and NumPy types (`np.int64`, `np.float64`) to _NumPy dtype_ objects.

2. **Conversion Testing**: Instead of type checking, attempt actual conversion using `np.array([default_value], dtype=dtype_obj)` to verify compatibility.

3. **Precision Loss Detection**: Added specific check for precision loss when converting float -> int with decimal values (e.g., `10.5` -> `int`).

4. **Error Handling**: Only warn when conversion fails (`ValueError`, `TypeError`, `OverflowError`) or when precision would be lost.

### Testing
**Manual Testing:**
Created and ran test cases to verify the fix works correctly:

1.  `default_value=0, dtype=int` -> No warning (compatible)
2.  `default_value=10.0, dtype=float` -> No warning (compatible)
3.  `default_value=0, dtype=float` -> No warning (compatible)
4.  `default_value=10.0, dtype=np.float64` -> No warning (compatible)
5.  `default_value=0, dtype=np.int64` -> No warning (compatible)
6.  `default_value=10.5, dtype=int` -> Warns correctly (precision loss)

**Affected Test Files:**
The following test files should no longer generate false warnings:
- `tests/test_space_renderer.py:46`
- `tests/test_solara_viz.py:121`
- `tests/test_solara_viz_updated.py:136`
- `tests/test_components_matplotlib.py:240`

**Verification:**
- Code passes linting checks (no errors)
- Logic verified with direct _dtype_ compatibility tests
- Both `PropertyLayer` implementations fixed consistently

### Additional Notes
- **Backward Compatibility**: This change is fully backward compatible. It only affects when warnings are raised, not the actual functionality of `PropertyLayer`.
- **Consistency**: Both `PropertyLayer` implementations (`mesa.space.PropertyLayer` and `mesa.discrete_space.PropertyLayer`) were updated to use the same logic for consistency.
- **Future Considerations**: The fix properly handles both Python native types and _NumPy types_, making it more robust for users who prefer explicit _NumPy dtypes_.
- **Related**: This addresses the "PropertyLayer Default Value Type Mismatch" warnings mentioned in #2904 (section 2.2)

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -642,25 +642,16 @@ class PropertyLayer:
                 f"Width and height must be positive integers, got {width} and {height}."
             )
         # Check if the dtype is suitable for the data
-        dtype_obj = np.dtype(dtype)
         try:
-            # Try to convert the value to the target dtype to check compatibility
-            _ = np.array([default_value], dtype=dtype_obj)[0]
-            # Check if conversion would lose precision
-            if (
-                dtype_obj.kind == "i"
-                and isinstance(default_value, float)
-                and default_value != int(default_value)
-            ):
-                warn(
-                    f"Default value {default_value} ({type(default_value).__name__}) might not be best suitable with dtype={dtype_obj.name} (loss of precision).",
+            if dtype(default_value) != default_value:
+                warnings.warn(
+                    f"Default value {default_value} will lose precision when converted to {dtype.__name__}.",
                     UserWarning,
                     stacklevel=2,
                 )
-        except (ValueError, TypeError, OverflowError) as e:
-            # Value cannot be converted to the target dtype
+        except (ValueError, TypeError) as e:
             raise TypeError(
-                f"Default value {default_value} ({type(default_value).__name__}) is not compatible with dtype={dtype_obj.name}."
+                f"Default value {default_value} is incompatible with dtype={dtype.__name__}."
             ) from e
 
         self.data = np.full((width, height), default_value, dtype=dtype)


mesa/discrete_space/property_layer.py
```diff
@@ -75,9 +75,25 @@ def __init__(
...
+                warnings.warn(
+                    f"Default value {default_value} ({type(default_value).__name__}) might not be best suitable with dtype={dtype_obj.name} (loss of precision).",
+                    UserWarning,
+                    stacklevel=2,
+                )
+        except (ValueError, TypeError, OverflowError):
+            # Value cannot be converted to the target dtype
             warnings.warn(
-                f"Default value {default_value} ({type(default_value).__name__}) might not be best suitable with dtype={dtype.__name__}.",
+                f"Default value {default_value} ({type(default_value).__name__}) might not be best suitable with dtype={dtype_obj.name}.",
```
quaquel: should this not just be an exception instead of a warning?

mesa/space.py
```diff
@@ -642,9 +642,25 @@ def __init__(
...
+                    f"Default value {default_value} ({type(default_value).__name__}) might not be best suitable with dtype={dtype_obj.name} (loss of precision).",
+                    UserWarning,
+                    stacklevel=2,
+                )
+        except (ValueError, TypeError, OverflowError):
+            # Value cannot be converted to the target dtype
             warn(
-                f"Default value {default_value} ({type(default_value).__name__}) might not be best suitable with dtype={dtype.__name__}.",
+                f"Default value {default_value} ({type(default_value).__name__}) might not be best suitable with dtype={dtype_obj.name}.",
                 UserWarning,
```
quaquel: There is code repetition here.

falloficaruss: I don't see the repitition, but I will check and make required changes.

quaquel:
Can you also check why builds are failing now?

falloficaruss:
> Can you also check why builds are failing now?

I need to update tests, working on it.

PR #2966, opened 2025-12-18
Fix SolaraViz multipage rendering when renderer is absent

### LINKED ISSUE #2965: Components don’t render on pages other than `0` when the renderer isn’t passed.
**Describe the bug**
Components doesn't render on solara web view if their page is anything other than `0` and renderer is not passed. Most likely, this happens because the renderer acts as an anchor that is always on page `0`, and the pages are calculated relative to it.

**Expected behavior**
Component should render on any page that is specified, the existence of any pages before or after it should not matter.

Currently it opens up the numbered page with nothing on it, it should plot the respective component on that page.

**To Reproduce**
Take any example and try to set page of plot component to anything other than `0` and remove `renderer` from `SolaraViz`. Or try to pass any custom component with page number other than `0` in absence of renderer.

falloficaruss:
### Summary
Fixes a bug where visualization components assigned to any page number other than 0 failed to render when no space renderer was provided. It also resolves a Solara hook validation warning in the `ComponentsView` component.

### Bug / Issue

In `SolaraViz`, if a user provides custom components and assigns them to a page other than 0 (e.g., `components[(MyComponent, 1)])` while omitting the renderer argument, the components never appear. The UI comparison logic was incorrectly matching the user-defined `page_id` against the internal 0-based `current_tab_index`.

Expected: The active tab should display the components assigned to the page representing that tab. Actual: Tab 0 attempted to find a component with `page_id == 0`. If none existed (because the user started at Page 1), the screen remained blank.

### Implementation
- Logic Correction: In `mesa/visualization/solara_viz.py`, updated `ComponentsView` to compare the `current_tab_index` against the _enumeration index (i)_ of the sorted pages rather than the `page_id` itself. This ensures the active tab always renders the content assigned to that sequential position.

### Testing
- Reproduction Case: Created a test model with a single button component assigned to Page 1 and no renderer. Verified that without the fix, the button is missing from the active tab, and with the fix, it renders correctly.
- Regressions: Ran `pytest tests/test_solara_viz.py` to ensure that standard visualizations (where renderer is Page 0) still function correctly.

### Additional Notes
This change makes the page numbering more robust; for example, if a user passes `[(CompA, 5), (CompB, 10)]`, Tab 0 will now correctly show `CompA` and Tab 1 will show `CompB`.

Closes #2965


ShreyasN707 (approved):
The updated tab/page handling looks correct and resolves the rendering behavior cleanly.

One small optional suggestion for maintainability: adding a brief comment near the `i == current_tab_index` check explaining why page IDs are intentionally decoupled from tab indices could help prevent future regressions.

@Sahil-Chhoker it looks good to me.

falloficaruss:
> The updated tab/page handling looks correct and resolves the rendering behavior cleanly.
> 
> One small optional suggestion for maintainability: adding a brief comment near the `i == current_tab_index` check explaining why page IDs are intentionally decoupled from tab indices could help prevent future regressions.
> 
> @Sahil-Chhoker it looks good to me.

I didn't feel the need to add a comment while making changes but I think you have a good point. Thanks for pointing out.

Sahil-Chhoker:
@ShreyasN707 Thanks for you review, I'll give it a final review as early as tomorrow, but I have to point out that 
> Hook Warning Fix: Removed an early return in ComponentsView that occurred before solara.use_state calls. This fixes a UserWarning regarding the "rules of hooks" when the component list was empty.

is already addressed in PR #2925 so I would like you to remove the changes regarding that.

Sahil-Chhoker (approved):
Sorry for the delay @falloficarus22, great work!

Issue #2968, opened 2025-12-18
Grid.select_random_empty_cell hangs on full grids


ShreyasN707:
I noticed Grid.select_random_empty_cell appears to be able to enter an infinite loop when the grid has no empty cells available.

When _try_random is enabled (the default), the method keeps sampling random cells inside a while True loop and only returns once an empty cell is found. If the grid is fully occupied, this condition is never met, so the call never returns and the simulation ends up hanging.

From reading the implementation, this looks like an unguarded edge case.

## Expected behavior

select_random_empty_cell should always terminate in a predictable way.
If no empty cells exist, I would expect it to either fall back to the deterministic implementation in DiscreteSpace or fail cleanly with a clear error, rather than hanging indefinitely.

## To Reproduce

Minimal reproduction:

Create a small grid (for example, 2×2).

Fill all cells with agents so that no empty cells remain.

Call grid.select_random_empty_cell().

### Example:
```python

from mesa.discrete_space.grid import OrthogonalMooreGrid

grid = OrthogonalMooreGrid((2, 2))
for cell in grid.all_cells:
    cell.add_agent(object())  # fill all cells

# This call does not return
grid.select_random_empty_cell()

```
Once the grid is full, the function hangs and never returns.

### Additional context

 - This seems to occur specifically in the _try_random branch due to the unbounded while True loop.

 - A possible fix could be to limit the number of random attempts and then fall back to  DiscreteSpace.select_random_empty_cell(), keeping the current optimization while guaranteeing termination.

- Could you please take a look and confirm whether this behavior is expected? 

@EwoutH @quaquel I can raise a PR fixing this issue!

quaquel:
good catch, go ahead with a PR.

ShreyasN707:
> good catch, go ahead with a PR.

Sure! Will raise a PR soon.

EwoutH:
> * This seems to occur specifically in the _try_random branch due to the unbounded while True loop.

Good catch. Another reason to be very careful with using `while True`.

codebyNJ:
@quaquel @EwoutH! I've implemented a fix that addresses this issue. 

Changes made:

- Added an early exit check when the grid is completely full, raising `IndexError` with a clear message.
- Replaced the unbounded `while True` loop with a bounded loop `( max_attempts = total_cells * 2 )`.
- Added adaptive strategy that switches to list-based selection when grid is >95% full.
[...]

PR #2976, opened 2025-12-19
some changes in grid.py

ShreyasN707:
Fixed an issue where select_random_empty_cell could enter an infinite loop when the grid was fully occupied, causing the simulation to hang.

Issue:#2968

When select_random_empty_cell was called on a full grid, the internal while True loop never found an empty cell and never terminated. This resulted in a hard freeze instead of predictable behavior.

Expected behavior: the method should either return an empty cell or fail deterministically when none exist.
Actual behavior: the call hung indefinitely.

## Implementation

Updated mesa/discrete_space/grid.py to guarantee termination:

- Added an explicit check to confirm whether any empty cells exist before attempting random sampling.

- Replaced the unbounded while True loop with a bounded loop derived from the grid dimensions.

- Added a deterministic fallback to super().select_random_empty_cell() when no empty cell can be found.

- This ensures the method always finishes and preserves existing parent-class behavior.

## Testing

- Reproduction test: Verified that calling select_random_empty_cell on a fully occupied 2×2 grid no longer hangs and instead returns immediately via the fallback.

- Regression testing: Ran the full test suite (pytest) with no failures.

- Manual check: Ran the Schelling example model to confirm normal grid behavior remains unchanged.


quaquel:
superseded by by ##3014

PR #2978, opened 2025-12-20
Fix reproducibility warnings by adding explicit random parameters

codebreaker32:
### Summary
Fixes `UserWarning` about missing random number generators in tests and examples by explicitly passing `random` parameters to grid constructors.

### Motive
Several tests were generating warnings that "Random number generator not specified, this can make models non-reproducible."

### Implementation
Added explicit `random` parameter to all affected grid and cell constructors:
- Conway's Game of Life: passes `random=self.random` to inherit model's RNG
- Test files: passes `random=random.Random(42)` for deterministic test behavior

### Usage Examples
```python
# Before
grid = OrthogonalMooreGrid((width, height), capacity=1, torus=True)  # Warning!
```
```python
# After
grid = OrthogonalMooreGrid((width, height), capacity=1, random=self.random, torus=True)
```

Part of #2904.

### DIFF
--- a/tests/visualization/test_backends.py
+++ b/tests/visualization/test_backends.py
@@ -1,11 +1,11 @@
 """Test the backends of the visualization package."""
 
+import random
 import types
 from unittest.mock import MagicMock
 
 import numpy as np
 import pytest
-import random 
 
 from mesa.discrete_space.grid import OrthogonalMooreGrid
 from mesa.discrete_space.property_layer import PropertyLayer
--- a/mesa/examples/basic/conways_game_of_life/model.py
+++ b/mesa/examples/basic/conways_game_of_life/model.py
@@ -10,7 +10,9 @@ class ConwaysGameOfLife(Model):
         """Create a new playing area of (width, height) cells."""
         super().__init__(seed=seed)
         # Use a simple grid, where edges wrap around.
-        self.grid = OrthogonalMooreGrid((width, height), capacity=1, random=self.random, torus=True)
+        self.grid = OrthogonalMooreGrid(
+            (width, height), capacity=1, random=self.random, torus=True
+        )
 
         # Place a cell at each location, with some initialized to
         # ALIVE and some to DEAD.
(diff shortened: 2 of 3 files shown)


EwoutH:
@quaquel is it now best-practice that we need to pass `random` to everything? Can't all components that have access to a `model` instance just use `self.model.random` or such?

quaquel:
> @quaquel is it now best-practice that we need to pass random to everything? Can't all components that have access to a model instance just use self.model.random or such?

That is already how it's done. `CellCollection`, and the spaces are not aware of the model, so they cannot fall back on it. In contrast, `Agent` has the model and so uses this. `AgentSet` also falls back on it via the model reference found in the first agent in the set.

EwoutH:
While I get where the design originates from, I find this very confusing for end-users. Especially from the concept of guided learning with easy entry and gracefully learning more powerful features (@tpike3 how do you always call that?). Adding a space is often one of the first things people do, and now they get a warning and they have to dive into random number generators and which parts of spaces have access to what model design. That's all unneeded complexity introduced too early to 

I get that we need the whole `random` passing from a reproducibility standpoint, and I acknowledge that may add some complexity here and there. I support adopting [SPEC 7](https://scientific-python.org/specs/spec-0007/). But how can we make it so that the end-users doesn't encounter its complexity so early, but can gradually learn it?

quaquel:
We debated this before while developing both the discrete space and the new experimental continuous space. 

From a learning perspective, I don't see the problem as long as the tutorials and examples always pass `random=model.random`/`rng=model.rng`, which provides a clear pattern for learners to follow. 

From an implementation point of view, I also don't know of a simple solution, apart from trying to get `model.random` from the agent as per `AgentSet`, but this will add overhead via an if else or try catch and a property:

```python
@properyt
def random:
    try:
        return getattr(self, "_random"
    except AttribtueError:
       # figure it out via agents


```

EwoutH:
@quaquel code wise, is this good to go?

tests/visualization/test_backends.py
```diff
@@ -1,5 +1,6 @@
 """Test the backends of the visualization package."""
 
+import random
```
EwoutH, line 3: Is this the right `random` that's imported?

EwoutH (approved):

PR #2981, opened 2025-12-21
Fix: Handle capacity=None in Cell.is_full property

### LINKED ISSUE #2980: Bug: Cell.is_full type comparison error with capacity=None
**Describe the bug**
The Cell.is_full property compares an integer to None when checking if a cell with infinite capacity is full. This happens because cells default to capacity=None for infinite capacity, and the code does 
len(self.agents) == self.capacity
, which evaluates to something like 100 == None. While this returns the correct result (False), it's a type comparison error that violates type safety.

Location: 
mesa/discrete_space/cell.py
 lines 135-137

Image

**Expected behavior**
The method should explicitly handle the None case instead of relying on int-to-None comparison. When capacity=None, it should return False without comparing types.

**To Reproduce**
```
from mesa import Model

# cell with infinite capacity
cell = Cell((0, 0), capacity=None)

# Add agents
model = Model()
for i in range(100):
    agent = CellAgent(model)
    agent._mesa_cell = cell
    cell._agents.append(agent)

# This does: len(cell.agents) == cell.capacity
# Which is: 100 == None
print(f"is_full: {cell.is_full}") 
```

**Additional context**
Currently works by accident since int == None returns False, but should be fixed for code clarity and type safety. Also changed == to >= in the fix to handle edge cases where agents might exceed capacity.

Nithin9585:
Fixes #2980

This PR fixes the type comparison error in `Cell.is_full` when `capacity=None`.

### Changes
- Add explicit `None` check for infinite capacity
- Change `==` to `>=` to handle edge cases where agents exceed capacity
- Add comprehensive tests covering None capacity, finite capacity, and edge cases

### Testing
All new tests pass:
- [test_cell_is_full_with_none_capacity](cci:1://file:///c:/Users/Nithi/OneDrive/Pictures/Desktop/gsoc/mesa/tests/test_cell_is_full_fix.py:9:0-24:32)
- [test_cell_is_full_with_finite_capacity](cci:1://file:///c:/Users/Nithi/OneDrive/Pictures/Desktop/gsoc/mesa/tests/test_cell_is_full_fix.py:27:0-51:31)  
- `test_cell_is_full_with_capacity_exceeded`

### DIFF
--- a/mesa/discrete_space/cell.py
+++ b/mesa/discrete_space/cell.py
@@ -135,7 +135,7 @@ class Cell:
     def is_full(self) -> bool:
         """Returns a bool of the contents of a cell."""
         if self.capacity is None:
-            return False  
+            return False
         return len(self.agents) >= self.capacity
 
     @property
--- a/tests/test_discrete_space.py
+++ b/tests/test_discrete_space.py
@@ -605,28 +605,28 @@ def test_cell():
 def test_cell_is_full_with_none_capacity():
     cell = Cell((0, 0), capacity=None)
     assert cell.is_full is False
-    
+
     model = Model()
     for _ in range(100):
         agent = CellAgent(model)
         agent._mesa_cell = cell
         cell._agents.append(agent)
-    
+
     assert cell.is_full is False
 
 
 def test_cell_is_full_with_finite_capacity():
     cell = Cell((0, 0), capacity=3)
     model = Model()
-    
+
     assert cell.is_full is False
-    
+
     cell.add_agent(CellAgent(model))
     assert cell.is_full is False
-    
+
     cell.add_agent(CellAgent(model))
     assert cell.is_full is False
-    
+
     cell.add_agent(CellAgent(model))
     assert cell.is_full is True
 


tests/test_cell_is_full_fix.py
EwoutH, line 1: I don’t think this needs to be so extensive and be its own file. Just one or two test cases in the existing ones should be sufficient.

quaquel: Can you move these tests into test_discrete_space.py?

Nithin9585: @quaquel i have moved the tests into test_discrete_space.py

Nithin9585: @EwoutH  I have also minimised the test cases

quaquel:
From my perspective, this is ready to be merged. I'll wait for @EwoutH before merging.

EwoutH:
Nice clean final diff, thanks!

(we do require docstring for every function, a single line is often enough for tests. See https://github.com/mesa/mesa/commit/6420ae28cc609eea646426f1edefdc88212dab32)

PR #2983, opened 2025-12-21
Fix: IndexError in select_random_agent when cell collection is empty

### LINKED ISSUE #2982: CellCollection.select_random_agent() crashes when no agents in cells
I was building a simple money exchange model where agents on a grid give money to random neighbors. My model kept crashing and I tracked it down to `CellCollection.select_random_agent()` raising `IndexError` when there are no agents in the cells.

This happens in sparse grids or when agents die/move, leaving neighborhoods empty. The method tries to call `random.choice()` on an empty list, which crashes the entire model.

**Location:** `mesa/discrete_space/cell_collection.py`, line 111

**Expected behavior**

The method should either:
1. Raise a more descriptive error explaining no agents are available, OR
2. Have a parameter to handle empty cases gracefully

Currently it just crashes with a generic `IndexError` that doesn't explain the real problem.

**To Reproduce**

Here's a minimal example showing the crash:

```python
from mesa import Model
from mesa.discrete_space import OrthogonalMooreGrid, CellAgent

class MoneyAgent(CellAgent):
        super().__init__(model)
        self.cell = cell
        self.wealth = 100
    
    def step(self):
        if self.wealth > 0:
            # This crashes when neighborhood is empty!
            other = self.cell.neighborhood.select_random_agent()
            other.wealth += 1
            self.wealth -= 1

class MoneyModel(Model):
    def __init__(self, n_agents=5, width=10, height=10):
        super().__init__()
        self.grid = OrthogonalMooreGrid((width, height), random=self.random)
        
        # Sparse grid - most cells empty
        for _ in range(n_agents):
            cell = self.random.choice(list(self.grid.all_cells))
            MoneyAgent(self, cell)
    
    def step(self):

# Run it
model = MoneyModel()
model.step()  # Crashes!
```

**Error Output:**

```
  File "test_crash_real.py", line 52, in <module>
    model.step()
  File "mesa/model.py", line 136, in _wrapped_step
    self._user_step(*args, **kwargs)
  File "test_crash_real.py", line 41, in step
  File "mesa/agent.py", line 356, in shuffle_do
    getattr(agent, method)(*args, **kwargs)
  File "test_crash_real.py", line 21, in step
    other = self.cell.neighborhood.select_random_agent()
  File "mesa/discrete_space/cell_collection.py", line 111, in select_random_agent
    return self.random.choice(list(self.agents))
  File "random.py", line 347, in choice
    raise IndexError('Cannot choose from an empty sequence')
IndexError: Cannot choose from an empty sequence
```

**Root Cause**

Looking at the code in `cell_collection.py`:

```python
def select_random_agent(self) -> CellAgent:
    """Select a random agent."""
    return self.random.choice(list(self.agents))  #  Crashes if empty!
```

When `self.agents` is empty (no agents in the cells), `list(self.agents)` returns `[]`, and `random.choice([])` raises `IndexError`.

**Proposed Fix**

```python
def select_random_agent(self) -> CellAgent:
    """Select a random agent."""
    agents_list = list(self.agents)
    if not agents_list:
        raise ValueError("No agents available in the cell collection")
    return self.random.choice(agents_list)
```

I'm happy to submit a PR if this looks like a valid issue.

Nithurshen:
### Summary

This PR fixes a crash in `CellCollection.select_random_agent()` where calling the method on a collection with no agents raised an `IndexError`.

It implements a hybrid solution (consensus reached in #2982 ):

Raises a descriptive LookupError if the collection is empty and no default is provided.
Adds a default parameter . Users with sparse grids can pass `default=None` to avoid the overhead of exception handling.

### Bug / Issue

Fixes #2982 
In sparse grids or simulations where agents move/die, it is common for a cell's neighborhood to contain no agents. What happened previously is that `select_random_agent` attempted to call `self.random.choice()` on `list(self.agents)`. If `self.agents` was empty, `random.choice` raised `IndexError: Cannot choose from an empty sequence`, causing the entire model to crash. The method now checks for empty collections first. If `default` is not provided it raises `LookupError: Cannot select random agent from empty collection`. If `default` is provided then it returns the default value.

### Implementation

1) Modified `mesa/discrete_space/cell_collection.py`
```python
    def select_random_agent(self, default=_no_default) -> CellAgent | None:
        """Select a random agent from the collection.

        Args:
            default: Value to return if the collection is empty.
                     If not provided, raises LookupError.

        Returns:
            CellAgent: A random agent, or the default value if provided and collection is empty.

        Raises:
            LookupError: If collection is empty and no default is provided.
        """
        agents = list(self.agents)

        if not agents:
            if default is _no_default:
                raise LookupError("Cannot select random agent from empty collection")
            return default

        return self.random.choice(agents)
```

2) Added a new test case `test_select_random_agent_empty_safe` in `tests/test_cell_spaces.py`:
```python
def test_select_random_agent_empty_safe():
    """Test that select_random_agent returns None when no agents are present."""
    rng = random.Random(42)
    empty_collection = CellCollection([], random=rng)
    with pytest.raises(LookupError):
        empty_collection.select_random_agent()
    assert empty_collection.select_random_agent(default=None) is None
    assert empty_collection.select_random_agent(default="Empty") == "Empty"
```

### Testing

**Bug Reproduction Script**, credit @Nithin9585
```python
from mesa import Model
from mesa.discrete_space import OrthogonalMooreGrid, CellAgent

class MoneyAgent(CellAgent):
        super().__init__(model)
        self.cell = cell
        self.wealth = 100
    
    def step(self):
        if self.wealth > 0:
            # This crashes when neighborhood is empty!
            other = self.cell.neighborhood.select_random_agent()
            other.wealth += 1
            self.wealth -= 1

class MoneyModel(Model):
    def __init__(self, n_agents=5, width=10, height=10):
        super().__init__()
        self.grid = OrthogonalMooreGrid((width, height), random=self.random)
        
        # Sparse grid - most cells empty
        for _ in range(n_agents):
            cell = self.random.choice(list(self.grid.all_cells))
            MoneyAgent(self, cell)
    
    def step(self):

# Run it
model = MoneyModel()
model.step()  # Crashes!
```

**Output Before**
```command-line
(venv) nithurshen@Nithurshens-MacBook mesa % python3 temp.py  
  File "/Users/nithurshen/Projects/mesa/temp.py", line 32, in <module>
    model.step()  # Crashes!
    ~~~~~~~~~~^^
  File "/Users/nithurshen/Projects/mesa/mesa/model.py", line 136, in _wrapped_step
    self._user_step(*args, **kwargs)
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/Users/nithurshen/Projects/mesa/temp.py", line 28, in step
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/Users/nithurshen/Projects/mesa/mesa/agent.py", line 356, in shuffle_do
    getattr(agent, method)(*args, **kwargs)
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/Users/nithurshen/Projects/mesa/temp.py", line 13, in step
    other = self.cell.neighborhood.select_random_agent()
  File "/Users/nithurshen/Projects/mesa/mesa/discrete_space/cell_collection.py", line 111, in select_random_agent
    return self.random.choice(list(self.agents))
           ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "/Users/nithurshen/.pyenv/versions/3.13.8/lib/python3.13/random.py", line 351, in choice
    raise IndexError('Cannot choose from an empty sequence')
IndexError: Cannot choose from an empty sequence
```

**Output After**
```command-line
(venv) nithurshen@Nithurshens-MacBook mesa % python3 ./temp.py
  File "/Users/nithurshen/Projects/mesa/./temp.py", line 32, in <module>
    model.step()  # Crashes!
    ~~~~~~~~~~^^
  File "/Users/nithurshen/Projects/mesa/mesa/model.py", line 136, in _wrapped_step
    self._user_step(*args, **kwargs)
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/Users/nithurshen/Projects/mesa/./temp.py", line 28, in step
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/Users/nithurshen/Projects/mesa/mesa/agent.py", line 356, in shuffle_do
    getattr(agent, method)(*args, **kwargs)
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/Users/nithurshen/Projects/mesa/./temp.py", line 13, in step
    other = self.cell.neighborhood.select_random_agent()
  File "/Users/nithurshen/Projects/mesa/mesa/discrete_space/cell_collection.py", line 122, in select_random_agent
    raise LookupError("Cannot select random agent from empty collection")
LookupError: Cannot select random agent from empty collection
```

### Additional Notes

This design balances the Pythonic "EAFP" principle with the performance needs of Agent-Based Modeling.

Maintainers preferred raising an Exception by default to avoid type-checking ambiguity.

Benchmarks provided by @Nithin9895 in #2982 showed that using `default=None` is ~1.2x faster than try/except blocks on sparse grids (5% density). This opt-in path allows high-performance simulations to avoid the overhead of creating/catching exceptions thousands of times per step.

Credits:

@Nithin9585 for the bug report, reproduction script, and performance benchmarks.

@quaquel and @EwoutH for the design discussion regarding EAFP vs. Performance.

### DIFF
--- a/mesa/discrete_space/cell_collection.py
+++ b/mesa/discrete_space/cell_collection.py
@@ -28,7 +28,7 @@ if TYPE_CHECKING:
 
 T = TypeVar("T", bound="Cell")
 
-_no_default = object()
+RAISES = object()
 
 
 class CellCollection[T: Cell]:
@@ -102,7 +102,7 @@ class CellCollection[T: Cell]:
         """Select a random cell."""
         return self.random.choice(self.cells)
 
-    def select_random_agent(self, default=_no_default) -> CellAgent | None:
+    def select_random_agent(self, default=RAISES) -> CellAgent | None:
         """Select a random agent from the collection.
 
         Args:
@@ -118,7 +118,7 @@ class CellCollection[T: Cell]:
         agents = list(self.agents)
 
         if not agents:
-            if default is _no_default:
+            if default is RAISES:
                 raise LookupError("Cannot select random agent from empty collection")
             return default
 


quaquel:
I disagree with this fix. In my view, returning None is a bad habit because it forces the user to perform a check on the return type. It is, in fact, more Pythonic, in my experience, to use try-except structures for control flow. See for example [the first answer to this question on stack overflow](https://stackoverflow.com/questions/16138232/is-it-a-good-practice-to-use-try-except-else-in-python).

Nithurshen:
> I disagree with this fix. In my view, returning None is a bad habit because it forces the user to perform a check on the return type. It is, in fact, more Pythonic, in my experience, to use try-except structures for control flow. See for example [the first answer to this question on stack overflow](https://stackoverflow.com/questions/16138232/is-it-a-good-practice-to-use-try-except-else-in-python).

I have updated the PR according to the agreed fix in #2986. Kindly review it

mesa/discrete_space/cell_collection.py
```diff
@@ -28,6 +28,8 @@
 
 T = TypeVar("T", bound="Cell")
 
+_no_default = object()
```
quaquel: Why did you choose to do it this way instead of using None as in dict.get? If the idea is to mimic a well established pattern like dict.get, why deviate from it?

Nithurshen: The reason for the sentinel is that we want to distinguish between two specific scenarios:
1) `select_random_agent()` (No argument)  Should raise `LookupError`.
2) `select_random_agent(default=None)` (Explicit None)  Should return `None`

If I defined the signature as `def select_random_agent(self, default=None):`, then 1 would automatically return `None` instead of raising an error. The function would have no way of knowing if the user forgot the argument (which should error) or explicitly passed `None` (which should not).

quaquel: Logically, I understood what was happening. I am just not sure what the cleanest solution is. Conceptually, we are taking inspiration from `dict.get`, which uses `None`. However, `dict.get` will never raise an exception, so it's not a perfect analogy.

So, I am fine with the suggested approach. It offers both exceptions as default, but also control, including the use of None. 

2 minor quibbles: first, since this becomes part of the signature of the method, naming this matters. `no_default` is not really clear. In 32986, @Nithin9585 uses `RAISES` which is perhaps clearer. Second, for the same reason, I would not use an underscore in the name.

Nithurshen: I have renamed it to `RAISES`, as you requested.

PR #2985, opened 2025-12-22
Support AgentPortrayalStyle in Altair visualization components

EwoutH:
### Problem
The Altair visualization components (`altair_components.py`) were not updated to support the new `AgentPortrayalStyle` class introduced in the visualization refactoring. When users returned an `AgentPortrayalStyle` instance from their `agent_portrayal` function, the code attempted dictionary-style assignment (`agent_data["x"] = x`), causing a `TypeError: 'AgentPortrayalStyle' object does not support item assignment`.

This also triggered a deprecation warning in tests that were still using the old dict-based portrayal format.

### Changes
- Added `_portrayal_to_dict()` helper function to convert `AgentPortrayalStyle` instances to dict format for Altair chart generation
- Updated `_get_agent_data_old__discrete_space()`, `_get_agent_data_new_discrete_space()`, and `_get_agent_data_continuous_space()` to use the helper function
- Updated `test_call_space_drawer` to use `AgentPortrayalStyle` instead of deprecated dict format
- Maintains full backward compatibility with existing dict-based portrayals

### Technical Details
Altair requires dict-formatted data for chart generation and only uses `color` and `size` fields from the agent portrayal (other fields become tooltips). The helper function extracts these fields from `AgentPortrayalStyle` instances while preserving the original dict-based behavior for legacy code.

Follow-up on https://github.com/mesa/mesa/pull/2786.

### DIFF
--- a/mesa/visualization/components/altair_components.py
+++ b/mesa/visualization/components/altair_components.py
@@ -88,6 +88,23 @@ def SpaceAltair(
     solara.FigureAltair(chart)
 
 
+def _portrayal_to_dict(portrayal_result, agent):
+    """Convert AgentPortrayalStyle to dict or return dict as-is."""
+    from mesa.visualization.components import AgentPortrayalStyle  # noqa
+
+    if isinstance(portrayal_result, AgentPortrayalStyle):
+        # Convert AgentPortrayalStyle to dict for Altair
+        # Altair only uses 'color' and 'size' from the portrayal
+        agent_data = {"id": getattr(agent, "unique_id", id(agent))}
+        if portrayal_result.color is not None:
+            agent_data["color"] = portrayal_result.color
+        if portrayal_result.size is not None:
+            agent_data["size"] = portrayal_result.size
+        return agent_data
+    else:
+        return portrayal_result
+
+
 def _get_agent_data_old__discrete_space(space, agent_portrayal):
     """Format agent portrayal data for old-style discrete spaces.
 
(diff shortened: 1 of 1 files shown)


tests/test_solara_viz.py
```diff
@@ -129,7 +130,7 @@ def __init__(self, seed=None):
     model = MockModel()
 
     def agent_portrayal(agent):
-        return {"marker": "o", "color": "gray"}
+        return AgentPortrayalStyle(marker="o", color="gray")
```
Sahil-Chhoker, line 133: This is an old file, only kept for backwards compatibility, the updated file has all the necessary tests so you can leave this out of this PR.

EwoutH: Right, so the correct behavior is to emit an warning here.

But then isn't it weird that this edge case in `altair` behavior was only catched by the old test file and not the new one?

Sahil-Chhoker: This file contains the test for legacy backends, that's why it's kept, so it makes sense that it caught the error not the newer one.

EwoutH: But is Altair tested at all on this aspect in the new test set? Was it validated it tested all new behavior? Because it started to fail when I tried to resolve a deprecation warning by updating old to new behavior in the old test set. The new one so apparently didn't include it.

Sahil-Chhoker: You're right, older `altair` doesn't have any other tests, let's update it and can you rename the file as well to `test_solara_viz_legacy_backends` or something along those lines

EwoutH: Sounds good!

PR #2988, opened 2025-12-22
Fix: IndexError in batch_run with sparse data collection

### LINKED ISSUE #2987: BatchRunner crashes with sparse data collection (IndexError)
**Describe the bug**

`batch_run()` crashes with `IndexError: list index out of range` when a model uses sparse data collection (collecting data only at specific steps, not every step). The bug is in `batchrunner.py` line 256, which incorrectly uses step numbers as list indices.

**Expected behavior**

`batch_run()` should work correctly when models collect data only at certain steps (e.g., every 5 steps, or only at the end). This is a common performance optimization to save memory.

**To Reproduce**

```python
from mesa import Agent, Model

class SimpleAgent(Agent):
    def __init__(self, model):
        super().__init__(model)
        self.value = 0
    
    def step(self):
        self.value += 1

class TestModel(Model):
    def __init__(self, rng=None):
        super().__init__(rng=rng)
        self.agent = SimpleAgent(self)
        
            model_reporters={"AgentValue": lambda m: m.agent.value}
        )
        self.running = True
    
    def step(self):
        # Only collect data every 5 steps (sparse collection)
        if self.steps % 5 == 0:
        
        self.agent.step()
        
        if self.steps >= 10:
            self.running = False

# This crashes
results = batch_run(
    TestModel,
    parameters={},
    iterations=1,
    max_steps=10,
    data_collection_period=1,
    number_processes=1,
)
```

**Error Output:**
```
IndexError: list index out of range

Traceback:
  File "mesa/batchrunner.py", line 256, in _collect_data
    model_data = {param: values[step] for param, values in dc.model_vars.items()}
                         ~~~~~~^^^^^^
```

**Root Cause:**

At line 256 in `batchrunner.py`:
```python
model_data = {param: values[step] for param, values in dc.model_vars.items()}
```

The code uses `step` (a step number like 0, 5, 10) as a list index. But `values` is a list with only 3 items (indices 0, 1, 2). When it tries `values[5]` or `values[10]`, it crashes.

**Fix Needed:**

Map step numbers to the correct list indices instead of using step numbers directly.

Nithin9585:
Fixes #2987 

This PR resolves an `IndexError` that occurs in `batch_run()` when models use sparse data collection (collecting data only at specific steps rather than every step).

## Problem

The `_collect_data()` function in `batchrunner.py` incorrectly used step numbers as list indices:

```python
model_data = {param: values[step] for param, values in dc.model_vars.items()}
```

When a model collects data sparsely (e.g., only at steps 0, 5, 10), the `model_vars` list has only 3 items (indices 0, 1, 2). Attempting to access `values[5]` or `values[10]` causes an `IndexError`.

## Solution

Changed the implementation to map step numbers to their corresponding collection indices:

```python
available_steps = sorted(dc._agent_records.keys())
if step not in available_steps:
    step = max((s for s in available_steps if s <= step), default=0)

try:
    collection_index = available_steps.index(step)
except ValueError:
    collection_index = 0

model_data = {param: values[collection_index] for param, values in dc.model_vars.items()}
```

## Testing

Added `test_batch_run_sparse_collection()` to verify the fix works with models that collect data every N steps.

Test output:
```
tests/test_batch_run.py::test_batch_run_sparse_collection PASSED [100%]
```

## Changes

- `mesa/batchrunner.py`: Fixed the IndexError in `_collect_data()`
- `tests/test_batch_run.py`: Added regression test for sparse data collection

### DIFF
--- a/tests/test_batch_run.py
+++ b/tests/test_batch_run.py
@@ -538,3 +538,38 @@ def test_batch_run_missing_step():
 
     # Should handle sparse collection - may have fewer results
     assert len(results) >= 0
+
+
+def test_batch_run_empty_collection_edge_case():
+    """Test batch_run with edge case: requesting data before any collection happens."""
+
+    class EmptyCollectionModel(Model):
+        """Model that doesn't collect any data initially."""
+
+        def __init__(self, *args, **kwargs):
+            self.schedule = None
+            super().__init__()
+            self.datacollector = DataCollector(
+                model_reporters={"Value": lambda m: m.time}
+            )
+            # Don't collect initial state - this creates the edge case
+
+        def step(self):
+            super().step()
+            # Only collect on final step
+            if self.time == 3:
+                self.datacollector.collect(self)
+
+    # Request data for early steps when nothing has been collected yet
+    results = mesa.batch_run(
+        EmptyCollectionModel,
+        parameters={},
+        number_processes=1,
+        rng=[None],
+        max_steps=3,
+        data_collection_period=1,
+        display_progress=False,
+    )
+
+    # Should handle empty collections gracefully
+    assert len(results) >= 0


falloficaruss:
Hey, isn't it better to simply use `values[-1]` to get the latest collected data instead of trying to map steps to indices. This would be safer and work for all cases.

Nithin9585:
@falloficarus22 Good point! But if data exists at steps [0, 5, 10] and we request step 3, wouldn't `values[-1]` return step 10's data when we actually want step 0's? 

How would your approach handle different requested steps?

falloficaruss:
> @falloficarus22 Good point! But if data exists at steps [0, 5, 10] and we request step 3, wouldn't `values[-1]` return step 10's data when we actually want step 0's?
> 
> How would your approach handle different requested steps?

Well I thought it was possible to do with `values[-1]` and tried to apply the changes locally and I'm running into multiple errors.
So, I guess this approach is fine.

Your pre-commit is failing (maybe you didn't `ruff check` and `ruff format` the code).

Nithin9585:
@quaquel @EwoutH can you please review this pr ?

quaquel:
I tried to resolve the conflicts with main, but failed. Tests are falling at the moment. Do you have time to investigate?

Nithin9585:
@quaquel I've resolved the merge conflict!

quaquel:
> @quaquel I've resolved the merge conflict!

Thanks! I'll try to look at later today.

EwoutH:
Thanks for the review, and reviewing.

I think this is a bug fix right, and should be labeled as such?

quaquel:
it's not so much a bug fix but an enhancement of what used to be an IndexError (at least in my understanding)

PR #2990, opened 2025-12-22
Fix/cell capacity zero

ahmednabiled:
Fix a bug in `Cell.add_agent()` where cells with `capacity=0` incorrectly allowed unlimited agents instead of rejecting all agents.

## Problem
The condition `if self.capacity and n >= self.capacity:` treated `capacity=0` as falsy, which skipped the capacity check entirely. This meant cells with zero capacity could hold unlimited agents, which is incorrect behavior.

**Example of the bug:**
```python
cell = Cell(coordinate=(0, 0), capacity=0)
agent = CellAgent(model)
cell.add_agent(agent)  # Should raise exception, but didn't
print(len(cell.agents))  # Output: 1 (should be 0)
```

## Solution
Changed the condition from `if self.capacity` to `if self.capacity is not None` to properly distinguish between:
- `capacity=None` → unlimited agents allowed
- `capacity=0` → no agents allowed
- `capacity=1, 2, 3...` → limited by the specified capacity

## Changes Made
- Fixed the capacity check logic in `mesa/discrete_space/cell.py`
- Added test case for `capacity=0` in `tests/test_discrete_space.py`

## Testing
✓ Existing capacity tests still pass
✓ New test verifies `capacity=0` raises exception when adding agents
✓ All cell-related tests pass

## Impact
This fix addresses a logical bug in the capacity checking mechanism. Please review:
- Code that may have relied on the previous `capacity=0` behavior (which allowed unlimited agents) could be affected. This is considered fixing unintended behavior rather than a regression.
- No expected impact on code using `capacity=None` (unlimited) or positive capacity values.

Reviewers are welcome to discuss whether this change aligns with the intended design and if any deprecation or migration guidance is needed.

### DIFF
--- a/mesa/discrete_space/cell.py
+++ b/mesa/discrete_space/cell.py
@@ -134,7 +134,9 @@ class Cell:
     @property
     def is_full(self) -> bool:
         """Returns a bool of the contents of a cell."""
-        return len(self.agents) == self.capacity
+        if self.capacity is None:
+            return False
+        return len(self.agents) >= self.capacity
 
     @property
     def agents(self) -> list[CellAgent]:
--- a/mesa/discrete_space/cell_collection.py
+++ b/mesa/discrete_space/cell_collection.py
@@ -28,6 +28,8 @@ if TYPE_CHECKING:
 
 T = TypeVar("T", bound="Cell")
 
+RAISES = object()
+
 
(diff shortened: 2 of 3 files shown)


quaquel:
A very specific logical corner case. Thanks for spotting the bug, fixing it and including tests.

PR #2997, opened 2025-12-23
Fix #2996: Prevent infinite loop in Grid.select_random_empty_cell()

### LINKED ISSUE #2996: Infinite loop in Grid.select_random_empty_cell() when no empty cells exist
**Describe the bug**

When a grid is completely full and you call `select_random_empty_cell()`, the program hangs forever in an infinite loop. No exception is raised, making it very hard to debug.

This happens in real models when populations grow and fill up all available space, which is pretty common in agent-based simulations.

**Expected behavior**

Should raise a clear error message instead of hanging:

```python
ValueError: Cannot select random empty cell: grid is full.
```

**To Reproduce**

```python
from mesa import Model
from mesa.discrete_space.cell_agent import CellAgent

class SimpleAgent(CellAgent):
        super().__init__(model)
        self.cell = cell

model = Model()
grid = OrthogonalMooreGrid(dimensions=(3, 3), capacity=1, random=model.random)

for cell in grid.all_cells:
    SimpleAgent(model, cell)

empty_cell = grid.select_random_empty_cell()
```

Run this and the program will hang forever at the last line.

**Why this happens**

The code uses `while True` which never exits when there are no empty cells:

```python
if self._try_random:
    while True:
        cell = self.all_cells.select_random_cell()
        if cell.is_empty:
            return cell
```

Nithin9585:
Fixes #2996

## Problem
The `Grid.select_random_empty_cell()` method enters an infinite loop when all cells are occupied, causing the program to hang forever with no error message.

## Solution
- Added check for empty cells before the `while True` loop
- Raises `ValueError` with helpful message when grid is full
- Prevents silent hang and provides clear feedback to users

## Changes
- [mesa/discrete_space/grid.py](cci:7://file:///c:/Users/Nithi/OneDrive/Pictures/Desktop/gsoc/mesa/mesa/discrete_space/grid.py:0:0-0:0): Added empty cell validation
- [tests/test_discrete_space.py](cci:7://file:///c:/Users/Nithi/OneDrive/Pictures/Desktop/gsoc/mesa/tests/test_discrete_space.py:0:0-0:0): Added test case for full grid scenario

### DIFF
--- a/mesa/discrete_space/grid.py
+++ b/mesa/discrete_space/grid.py
@@ -150,6 +150,14 @@ class Grid(DiscreteSpace[T], HasPropertyLayers):
         # https://github.com/mesa/mesa/issues/1052 and
         # https://github.com/mesa/mesa/pull/1565. The cutoff value provided
         # is the break-even comparison with the time taken in the else branching point.
+
+        # Check if there are any empty cells to avoid infinite loop
+        if not self.empties:
+            raise ValueError(
+                "Cannot select random empty cell: grid is full. "
+                "Check grid.empties or grid.exists_empty_cells() before calling."
+            )
+
         if self._try_random:
             while True:
                 cell = self.all_cells.select_random_cell()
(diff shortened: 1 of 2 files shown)


codebreaker32:
See #2976. It is also addressing this same issue

Nithin9585:
@codebreaker32 Hey
Looks like we both found the same bug independently - I was working from issue #2996 (duplicate of #2968) and didn't see your PR. Sorry about that! i will close this !

PR #3002, opened 2025-12-23
Fix: Method Reporter Validation in DataCollector

vedantvakharia:
## Summary

This PR fixes a bug in `DataCollector._validate_model_reporter()` where **method reporters were never validated** due to incorrect conditional logic.

---

### Original Code (lines 166-168 in `datacollection.py`)

```python
# Type 2: Method of class/instance
if not callable(reporter) and not isinstance(reporter, types.LambdaType):
    pass
```

### Why This Was Wrong

| Expression | For a Method | Result |
|------------|--------------|--------|
| `callable(self.some_method)` | `True` | Methods are callable |
| `not callable(reporter)` | `False` | Always False for methods |
| `if` condition | `False and ...` | **Never executes** |

**Method reporters were completely skipped during validation.**

---

## The Fix

```python
# Type 2: Method of class/instance (bound methods are callable)
if callable(reporter) and not isinstance(reporter, types.LambdaType):
    try:
        reporter()  # Call without args for bound methods
    except Exception as e:
        raise RuntimeError(
            f"Method reporter '{name}' failed validation: {e!s}"
        ) from e
```

### Key Changes

| Aspect | Original | Fixed |
|--------|----------|-------|
| Condition | `not callable(reporter)` | `callable(reporter)` |
| Matches methods? | ❌ Never | ✅ Yes |
| Validation | `pass` (nothing) | Calls `reporter()` |
| Error handling | None | Wraps in `RuntimeError` with reporter name |

---

## Why `reporter()` Instead of `reporter(model)`?

Looking at how `collect()` calls method reporters (line 341):

```python
# Bound methods: called WITHOUT arguments (self is already bound)
self.model_vars[var].append(deepcopy(reporter()))
```

Bound methods like `self.some_method` already have `self` bound, so they're called with **no arguments**.

---

## Comparison Table

| Test Scenario | Original Code | Fixed Code |
|---------------|---------------|------------|
| Method referencing non-existent attribute | ❌ Not caught, unclear `AttributeError` | ✅ `RuntimeError` with reporter name |
| Method that raises exception | ❌ Raw exception | ✅ Wrapped with context |
| Working method reporter | ✅ Works | ✅ Works |
| Lambda + method mix | ✅ Works | ✅ Works |

---

## Tests Added

Added `TestMethodReporterValidation` class to `tests/test_datacollector.py`:

| Test | Description |
|------|-------------|
| `test_broken_method_reporter_raises_runtime_error` | Broken methods raise `RuntimeError` with reporter name |
| `test_working_method_reporter_succeeds` | Valid methods continue to work |
| `test_method_reporter_exception_is_wrapped` | Exceptions wrapped with context |
| `test_mixed_lambda_and_method_reporters` | Lambdas and methods work together |
| `test_method_reporter_called_without_args` | Methods called with `reporter()` |

---

## Test Results

```
$ pytest tests/test_datacollector.py -v
20 passed  # 15 original + 5 new
```

---

## Files Changed

| File | Change |
|------|--------|
| `mesa/datacollection.py` | Fixed method reporter validation condition |
| `tests/test_datacollector.py` | Added `TestMethodReporterValidation` class (5 tests) |

---

## Backwards Compatibility

✅ All 363 existing tests pass
✅ Working method reporters unaffected
✅ Only broken reporters now fail with clear error message

### DIFF
--- a/tests/test_datacollector.py
+++ b/tests/test_datacollector.py
@@ -641,6 +641,8 @@ class TestMethodReporterValidation(unittest.TestCase):
 
         # Validation call + actual collect call = at least 1
         self.assertGreaterEqual(model.call_count, 1)
+
+
 def test_mutable_data_independence():
     """Test that mutable agent data is deep-copied, preventing historical records from changing."""
 


vedantvakharia:
Hi @quaquel I have resolved the conflicts and accepted both changes. The tests are passing. Could you please re-approve so the pr can be merged?

PR #2994, opened 2025-12-23
Reorganize tests to mirror source module structure

EwoutH:
### Description
Reorganizes the test directory structure to mirror the source code organization in `mesa/`, making it easier to locate and maintain tests. If you now want to update a `visualization` test for example, you can dive straight into `tests/visualization`

### Changes
- Created `tests/discrete_space/` for discrete space-related tests
- Created `tests/experimental/` for experimental module tests  
- Created `tests/visualization/` for visualization-related tests
- Created `tests/examples/` for example tests
- Kept structure flat at one level deep (no nested subdirectories)
- Root-level module tests remain at top level

### New structure & comparison
```
tests/
├── __init__.py
├── read_requirements.py
├── test_agent.py
├── test_batch_run.py
├── test_datacollector.py
├── test_mesa_logging.py
├── test_model.py
├── test_space.py
├── test_import_namespace.py
├── test_lifespan.py
├── discrete_space/
│   ├── test_discrete_space.py
│   ├── test_grid.py
│   ├── test_hexgrid_torus_validation.py
├── experimental/
│   ├── test_continuous_space.py
│   ├── test_devs.py
│   ├── test_mesa_signals.py
│   └── test_meta_agents.py
├── visualization/
│   ├── test_backends.py
│   ├── test_components_matplotlib.py
│   ├── test_portrayal_components.py
│   ├── test_solara_viz.py
│   ├── test_solara_viz_updated.py
│   ├── test_space_drawer.py
│   └── test_space_renderer.py
└── examples/
    ├── test_examples.py
    ├── test_examples_viz.py
    └── test_end_to_end_viz.sh
```
image

Issue #3004, opened 2025-12-23
Coverage reporting 0% after test reorganization


EwoutH:
After reorganizing test files into subdirectories, Codecov consistently reports 0% coverage, even when source files are modified.

### What's known
Test execution
- ✅ All 379 tests are discovered correctly
- ✅ All 379 tests run successfully
- ✅ Pytest is working as expected

Coverage.py behavior
- ❌ Coverage.py reports: "No data was collected (no-data-collected)"
- ❌ Warning from: `/opt/hostedtoolcache/Python/3.14.2/x64/lib/python3.14/site-packages/coverage/control.py:957`
- The warning suggests coverage.py ran but didn't measure any lines as executed

Changes made
1. **Test file reorganization** (PR #2994):
   - Moved tests from `tests/` root into subdirectories:
     - `tests/discrete_space/`
     - `tests/experimental/`
     - `tests/examples/`
     - `tests/visualization/`
   - Updated import in `tests/test_space.py`: `from tests.discrete_space.test_grid import MockAgent`
2. **Configuration changes attempted**:
   - Moved `omit = ["tests/*"]` between `[tool.coverage.run]` and `[tool.coverage.report]` - no effect

Current configuration
`pyproject.toml`:
```toml
[tool.coverage.run]
source = ["mesa"]
branch = true
omit = ["tests/*"]
```

`codecov.yaml`:
```yaml
coverage:
  status:
    project:
      default:
        target: 80%
        threshold: 1%

ignore:
  - "benchmarks/**"
  - "mesa/visualization/**"

comment: off
```

### What doesn't explain this
- ❌ "No source files changed" - we modified source files
- ❌ "Changed files are ignored" - `mesa/experimental/` is not ignored
- ❌ "Tests aren't running" - all 379 tests execute successfully
- ❌ "Wrong omit configuration" - tried both configurations

### The mystery
Coverage.py is running (we see its warning), pytest is running all tests successfully, but coverage.py claims it measured zero lines. This suggests coverage.py is either:
1. Not tracing the code during test execution, or
2. Collecting data but unable to write/report it, or
3. Running in a mode where it's not hooked into the test execution

### Questions to investigate
1. Is coverage.py actually being invoked during pytest execution in CI?
2. Are there any pytest plugins or configurations that might interfere with coverage tracing?
3. Does the test reorganization somehow break coverage.py's ability to trace execution paths?
4. Is there a mismatch between where coverage.py expects source files vs where they actually are?

### Links
[...]

PR #3006, opened 2025-12-24
CI: Restore coverage collection after test reorganization

falloficaruss:
## Summary
Improves coverage collection after test reorganization in #2994 by adopting `pytest`'s native coverage integration and measuring all code including tests.

## Problem
After moving tests into nested subdirectories, coverage.py reported 0% coverage. The root cause was a combination of:
1. Path resolution issues with nested test packages
2. Tests not being measured in coverage reports

## Solution
This PR implements reviewer feedback to adopt coverage best practices:

### Configuration Changes (`pyproject.toml`):
- **Added `source = ["."]`**: Measures all code in the project directory
- **Kept `source_pkgs = ["mesa"]`**: Specifically tracks the mesa package
- **Added `relative_files = true`**: Fixes path resolution for nested directories
- **Removed `omit = ["tests/*"]`**: Now measures test coverage (critical for verifying all tests run)
- **Added `[tool.coverage.paths]`**: Maps installed package paths for accurate coverage reporting

### CI Changes (`.github/workflows/build_lint.yml`):
- **Replaced** `coverage run -m pytest` + `coverage xml` (2 commands from #3005)
- **With** `python -Im pytest -p pytest_cov --cov --cov-report=xml --cov-report=markdown-append:"${GITHUB_STEP_SUMMARY}" --cov-context=test tests/` (single command)
- **Benefits**:
  - `-Im` flag ensures testing installed package, not git checkout
  - `-p pytest_cov` loads plugin early to avoid import issues
  - `--cov` (no value) measures everything per coverage config
  - Markdown report appears in GitHub Actions summary (better UX)
  - `--cov-context=test` enables per-test coverage tracking

## Advantages Over PR #3005
1. **Single command** instead of two separate commands
2. **Measures test coverage** to verify all tests actually run
3. **More accurate** with `source = ["."]` + `source_pkgs` combination
4. **Tests installed package** using `python -Im` flag

## Testing
- Local verification shows 84.75% total coverage (up from 0%)
- All 379 tests pass
- Test files now measured in coverage
- Coverage XML generated successfully
- Branch coverage: 74.91%

## Coverage by Package
- Core modules: 94-99%
- Discrete space: 96.55%
- Experimental: 94%+
- Visualization: 65.87%

## Credits
Implementation incorporates excellent feedback from @webknjaz

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -99,6 +99,15 @@ source_pkgs = ["mesa"]
 relative_files = true
 branch = true
 
+[tool.coverage.paths]
+_site-packages-to-flat-layout-mapping = [
+  'mesa',
+  '*/mesa',
+  '*/lib/pypy*/site-packages/mesa',
+  '*/lib/python*/site-packages/mesa',
+  '*/Lib/site-packages/mesa',
+]
+
 [tool.hatch.build.targets.wheel]
 packages = ["mesa"]
 


Nithin9585:
Nice fix @falloficarus22!  Small note: original CI used `coverage run -m pytest`, not pytest-cov. So root cause was path resolution (fixed by `relative_files = true`), not a plugin conflict. Works great!

falloficaruss:
> Nice fix @falloficarus22! Small note: original CI used `coverage run -m pytest`, not pytest-cov. So root cause was path resolution (fixed by `relative_files = true`), not a plugin conflict. Works great!

Yeah but this is still a workaround, works until maintainers find the fix for the root cause.

EwoutH:
Thanks for the PR. What are the main advantages of this workaround over the other one?

falloficaruss:
> Thanks for the PR. What are the main advantages of this workaround over the other one?

- The key advantage is this workaround will do the same thing in a single command rather than 2 seperate commands.
- Also added `--cov-context=test` to enable per-test coverage tracking which might help in debugging.
- I think it may be better to use the standard pytest command rather than directly using coverage.py.

pyproject.toml
```diff
@@ -94,10 +94,14 @@ repository = "https://github.com/mesa/mesa"
...
 [tool.coverage.run]
-source = ["mesa"]
+source_pkgs = ["mesa"]
+relative_files = true
 branch = true
 omit = ["tests/*"]
 
+[tool.coverage.report]
+omit = ["tests/*"]
+
```
webknjaz, line 110: Best not to skip collecting coverage of the test modules, nor hide it from reports.
```suggestion
```

https://nedbatchelder.com/blog/202008/you_should_include_your_tests_in_coverage.html / https://nedbatchelder.com/blog/201908/dont_omit_tests_from_coverage.html / https://nedbatchelder.com/blog/201106/running_coverage_on_your_tests.html

In fact, it'd be a good idea to include an additional reported check in the Codecov config, requiring 100% code coverage on the test folder: https://github.com/ansible/awx-plugins/blob/274b32a/.codecov.yml#L33-L38.

One might argue that this is out of the scope of this PR and request a follow-up PR but can you prove that all the tests *actually* run in CI w/o that?

.github/workflows/build_lint.yml
```diff
@@ -73,9 +73,7 @@ jobs:
       - run: pip install uv
       - run: uv pip install --system .[dev]
       - run: playwright install chromium-headless-shell
-      - run: |
-          coverage run -m pytest tests/
-          coverage xml
+      - run: pytest --cov=mesa --cov-report=xml --cov-context=test tests/
```
webknjaz: Use `--cov` w/o a value to enable the plugin but not limit what's measured. It's actually a bit messy since it corresponds to `source` in the coveragepy config and there's no corresponding flag for `source_pkgs`, which means it probably refereces a folder and not an importable. Plus, as I mentioned elsewhere, test modules should *really* be measured too.

It's a good idea to also add `-p pytest_cov` which makes pytest load the plugin earlier than the standard entry point-based auto-discovery mechanism.

Finally, I'm adding `--cov-report=markdown-append:"${GITHUB_STEP_SUMMARY}"` since it's a nice touch to have this integrated in GHA: 

```suggestion
      - run: pytest -p pytest_cov --cov --cov-report=xml --cov-report=markdown-append:"${GITHUB_STEP_SUMMARY}" --cov-context=test tests/
```

Follow-up PR action items:
* [ ] Move `tests/` into a dedicated testpaths setting in `.pytest.toml`
* [ ] Move `--cov` and `--cov-context=test` into addopts in the pytest config too — these would be useful to be auto-added for both local contributors and in CI
* [ ] Keep `--cov-report=xml` and `--cov-report=markdown-append:"${GITHUB_STEP_SUMMARY}"` in CI since locally contributors don't need those reports
* [ ] Configure html coverage explicitly so that the contributors could pass `--cov-report=html` locally and have everything show up nicely by default: https://github.com/ansible/awx-plugins/blob/274b32a/.coveragerc#L1-L3
* [ ] Look into replacing `pytest` with `python -Im pytest` (`-I` disables auto-adding the current working directory to `$PYTHONPATH`/`sys.path` so you'd be sure you're testing what's in `site-packages/` and not the Git checkout copy)

pyproject.toml
webknjaz, line 1: You may want to play with hiding less important things in all reports: https://github.com/ansible/awx-plugins/blob/274b32a/.coveragerc#L17-L19. But it's best to do so combined with setting up HTML as I mentioned in another comment — this one should show everything because it's to be used interactively.

pyproject.toml
```diff
@@ -94,10 +94,14 @@ repository = "https://github.com/mesa/mesa"
 ignore-words-list = ["ist", "hart", "nD"]
 
 [tool.coverage.run]
-source = ["mesa"]
+source_pkgs = ["mesa"]
```
webknjaz, line 98: Both are needed to measure all the things and be able to check them automatically (like checking that all the tests run). I've found that this combo of settings works best when Codecov is integrated:
```suggestion
source = ["."]
source_pkgs = ["mesa"]
```

pyproject.toml
webknjaz, line 1: When the copy of the library in `site-packages` is tested correctly, you might need to set up mappings in `[paths]`: https://github.com/ansible/awx-plugins/blob/274b32a/.coveragerc#L5-L12.

webknjaz (commented):
Here's a few notes from me.

webknjaz:
> * Also added `--cov-context=test` to enable per-test coverage tracking which might help in debugging.

Note that this is only displayed in the HTML report and Codecov doesn't know how to show such contexts (it only shows own platform-specific flags you pass when uploading).


> * I think it may be better to use the standard pytest command rather than directly using coverage.py.

This is often considered controversial but I tend to like this way too.

DipayanDasgupta:
> @falloficarus22 I was looking into the root cause based on webknjaz's feedback in the issue. I verified locally that we can solve this without a workaround by updating `pyproject.toml` to handle the path resolution correctly.
>
> If you apply these changes, it should fix the coverage reporting comprehensively:
>
> **pyproject.toml**
> ```toml
> [tool.coverage.run]
> source_pkgs = ["mesa"]  # Targets the package specifically
> branch = true
> relative_files = true   # Fixes the path mapping from subdirectories
> # omit = ["tests/*"]    # Removed to ensure tests are measured
> ```
>
> And then in the workflow, simply use:
> `pytest -p pytest_cov --cov --cov-report=xml --cov-report=markdown-append:"${GITHUB_STEP_SUMMARY}" --cov-context=test tests/`
>

falloficaruss:
@webknjaz Thanks for the detailed review and suggestions. As I have mentioned earlier this is supposed to be a temporary workaround (an improvement over #3004). I'll work on moving `tests/` into dedicated test paths but doing so is out of scope for this PR. I guess there will be follow-up PRs coming up by me.

webknjaz:
@falloficarus22 almost. You keep adding an empty report section which is not necessary w/o any settings in it.

I usually recommend accepting suggestions via GH UI – it is likely that you'll get them committed exactly as they were meant to be, and the added bonus is that it's easier to credit people this way. Otherwise, trying to reproduce that locally is rather manual and thus error-prone: https://hynek.me/til/easier-crediting-contributors-github/.

With this, you'd click commit in-browser and then pull the changes locally. Afterwards, you'd be able to make corrections/rebasing/cleanups as needed.

pyproject.toml
```diff
@@ -94,9 +94,12 @@ repository = "https://github.com/mesa/mesa"
...
 
 [tool.coverage.run]
-source = ["mesa"]
+source = ["."]
+source_pkgs = ["mesa"]
+relative_files = true
 branch = true
-omit = ["tests/*"]
+
+[tool.coverage.report]
```
webknjaz:
```suggestion
```

webknjaz: @falloficarus22 in case you're not familiar with this feature on GitHub, here's the doc: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request#applying-suggested-changes

webknjaz:
I thought that https://github.com/mesa/mesa/pull/3006/files#r2646764846 can wait but now I realized that it must be included into this PR.

Look at https://github.com/mesa/mesa/actions/runs/20505498180?pr=3006#summary-58919356331 — it lists both relative paths (with 0% for anything starting with `mesa/`) and absolute coverage (with something actually measured under `..../site-packages/mesa/`). And because the modules from both locations are listed, that 0% ends up contributing to low overall coverage number.

You can take https://github.com/ansible/awx-plugins/blob/274b32a/.coveragerc#L5-L12 as is and convert it from INI into TOML, integrate that into `pyproject.toml` and this will fix the problem. You should end up with all paths being relative and CWD merged with site-packages.

---

As a side note, I noticed that *some* test modules don't have 100% coverage. This means that there's dead code in tests that's never exercised (but people may think that it's being run in CI, which would be misleading). This is a good candidate for figuring out in another follow-up PR — maybe, some code can be deleted, otherwise some of it might need to be marked with nocover pragma comments.

.github/workflows/build_lint.yml
```diff
@@ -73,9 +73,7 @@ jobs:
       - run: pip install uv
       - run: uv pip install --system .[dev]
       - run: playwright install chromium-headless-shell
-      - run: |
-          coverage run -m pytest tests/
-          coverage xml
+      - run: pytest -p pytest_cov --cov --cov-report=xml --cov-report=markdown-append:"${GITHUB_STEP_SUMMARY}" --cov-context=test tests/
```
webknjaz: While on it, it's perhaps just commit this too:
```suggestion
      - run: python -Im pytest -p pytest_cov --cov --cov-report=xml --cov-report=markdown-append:"${GITHUB_STEP_SUMMARY}" --cov-context=test tests/
```

pyproject.toml
```diff
@@ -94,9 +94,10 @@ repository = "https://github.com/mesa/mesa"
 ignore-words-list = ["ist", "hart", "nD"]
 
 [tool.coverage.run]
-source = ["mesa"]
+source = ["."]
+source_pkgs = ["mesa"]
+relative_files = true
 branch = true
-omit = ["tests/*"]
```
webknjaz, line 101:
```suggestion

[tool.coverage.paths]
_site-packages-to-flat-layout-mapping = [  # currently in use
  'mesa',
  '*/mesa',
  '*/lib/pypy*/site-packages/mesa',
  '*/lib/python*/site-packages/mesa',
  '*/Lib/site-packages/mesa',
]
_site-packages-to-src-mapping = [  # future possibility
  'src',
  '*/src',
  '*\src',
  '*/lib/pypy*/site-packages',
  '*/lib/python*/site-packages',
  '*\Lib\site-packages',
]

```

webknjaz (changes requested):
This should address the last missing piece: https://github.com/mesa/mesa/pull/3006#issuecomment-3691677788 / https://github.com/mesa/mesa/pull/3006/files#r2646764846

pyproject.toml
falloficaruss: That's quite good suggestion but I'm going to skip adding `src mapping` for now because mesa doesn't use `/src` layout as of now. It is a future possibility and a story for another PR.

falloficaruss:
`coverage build` has failed but has nothing to do with this PR. The problem is in `tests/examples/test_examples_viz.py` and can be fixed in another PR.

pyproject.toml
webknjaz: It's technically not harmful to have it in the config. But okay. I was just worried that when someone gets to do another restructuring, they might end up being confused once again. But at least it should be possible to set up coverage enforcements that would catch this.

webknjaz: It's technically not harmful to have it in the config. But okay. I was just worried that when someone gets to do another restructuring, they might end up being confused once again. But at least it should be possible to set up coverage enforcements that would catch this.

webknjaz:
@falloficarus22 if that's the case, could you please close and then reopen the PR right away? This will trigger a new CI run. Or ask the maintainers to restart that one job. I'd like to see it post the coverage report on the CI run web page (that job Markdown summary as a table). The diff looks good but it's a good idea to double-check that the mapping works correctly before finally green-lighting the merge.

falloficaruss:
Weird, but I think the PR is now good enough to be merged.

webknjaz (approved):
Not weird, really. When the tests themselves are flaky, this is expected. They can usually be make more reliable or the easiest hack would be adding a pytest plugin for auto-restarting the unstable tests.

Anyway.. I see that the report looks like what I'd expect (within the context I learned): https://github.com/mesa/mesa/actions/runs/20526264823#summary-58974397795.

You're good to go now. No more notes from me.

EwoutH:
Thanks for both your effort!

@falloficarus22 could you update the PR description to be fully up to date with the current diff and everything we’ve learned? I would like to have it as a reference for if we ever need to revisit it. Focus on helping us understand what every option does and how they interact.

falloficaruss:
@EwoutH updated the PR description. Please tell me if you would like to make any changes in this PR.

webknjaz:
> `-Im` flag ensures testing installed package, not git checkout

This is partially accurate. `-m` is the flag that invokes the target module through the runpy interface (`pytest.__main__`),  but it also adds `.` (CWD) to the import paths. And `-I` counters that bit of the side effect, making the interpreter ignore any `PYTHON*` env vars, including `PYTHONPATH` which influences `sys.path`.

`pytest` itself may also manipulate `sys.path` internally and this ability is not suppressed by `-I` — the relevant behavior usually depends on the presence of the `__init__.py` modules in the test dirs (this also affects whether it's possible to have relative imports in tests or absolute ones).

EwoutH (approved):
Great, thanks!

PR #3008, opened 2025-12-24
Update `from __future__ import annotations` comment

EwoutH:
The `from __future__ import annotations` statements were originally added to enable the `|` union operator for type hints in Python 3.9. Now that Python 3.10 is the minimum supported version (where [PEP 604](https://peps.python.org/pep-0604/) union syntax is natively available), these imports are no longer necessary and can be safely removed from agent.py, model.py, and space.py.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -3,8 +3,7 @@
 Core Objects: Agent and AgentSet.
 """
 
-# Mypy; for the `|` operator purpose
-# Remove this __future__ import once the oldest supported Python is 3.10
+# Postpone annotation evaluation to avoid NameError from forward references (PEP 563). Remove once Python 3.14+ is required.
 from __future__ import annotations
 
 import contextlib
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -3,8 +3,7 @@
 Core Objects: Model
 """
 
-# Mypy; for the `|` operator purpose
-# Remove this __future__ import once the oldest supported Python is 3.10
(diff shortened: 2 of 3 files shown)


EwoutH:
Okay, so mainly the comments are outdated. `from __future__ import annotations` is still needed, but now for something else than it was orginally added.

> ### Why `from __future__ import annotations` is still needed
> The `from __future__ import annotations` import enables PEP 563's behavior, which converts all annotations to strings at compile time instead of evaluating them as Python expressions.
> 
> **Why it was originally added to Mesa:**
> These imports were added with the comment "for the `|` operator purpose" to enable Python 3.9 compatibility. PEP 604 (Python 3.10+) introduced the `|` union operator for type hints (e.g., `int | str`), and the future import was needed to use this syntax on Python 3.9.
> 
> **Why we can't remove it yet:**
> While the `|` operator now works natively on Python 3.10+, the future import serves a different critical purpose: it prevents `NameError` exceptions from forward references. Without it, Python eagerly evaluates annotations at class/function definition time. In our code, `Agent.__init__` has a type hint `model: Model`, but `Model` is defined in a separate file. On Python 3.13 and earlier, removing the future import causes Python to try looking up `Model` immediately when the `Agent` class is defined, resulting in `NameError: name 'Model' is not defined`.
> 
> **The path forward:**
> PEP 649 (Python 3.14+) solves this with deferred evaluation—annotations are compiled into a lazy `__annotate__` function that's only called when accessed. Once Python 3.14 is our minimum version, these imports can be safely removed. Until then, they must stay to maintain compatibility with Python 3.10-3.13.

PR #3012, opened 2025-12-24
Fix - ABMSimulator.run_until() Step Rescheduling Bug

vedantvakharia:
## Summary

This PR fixes a critical bug in `ABMSimulator.run_until()` where `model.step` was never rescheduled after the first tick due to an incorrect function comparison. The simulator would only execute one step regardless of the specified end time.

### Bug Location
`mesa/experimental/devs/simulator.py`, line 373 (ABMSimulator.run_until)

### Root Cause
```python
# BUGGY CODE
if event.fn() == self.model.step:
    self.schedule_event_next_tick(self.model.step, priority=Priority.HIGH)
```

The bug occurs because:
1. `event.fn` is a **weak reference** (`WeakMethod` or `weakref.ref`)
2. `event.fn()` **dereferences** the weak reference, returning the actual callable
3. The code then compares the callable's **return value** to `self.model.step` (due to `==` operator precedence)
4. `model.step()` returns `None` or a Model instance, which never equals the method itself
5. **Result:** The condition is always `False`, so `model.step` is never rescheduled

### Impact
- `ABMSimulator.run_until(10)` would only execute **1 step** instead of 10
- The entire ABMSimulator was effectively broken for multi-step simulations

## The Solution

Instead of relying on fragile function identity comparison (which is problematic with weak references and bound methods), we introduce an explicit **EventType** enum to categorize events:

```python
class EventType(IntEnum):
    DEFAULT = 0      # Regular user-scheduled events
    MODEL_STEP = 1   # Auto-rescheduled step events
```

The fix replaces the broken comparison:
```python
# FIXED CODE
if event.event_type == EventType.MODEL_STEP:
    self.schedule_event_next_tick(
        self.model.step,
        priority=Priority.HIGH,
        event_type=EventType.MODEL_STEP,
    )
```

### 1. `mesa/experimental/devs/eventlist.py`

**Added `EventType` enum:**
```python
class EventType(IntEnum):
    """Types of simulation events."""
    DEFAULT = 0
    MODEL_STEP = 1
```

**Added `event_type` parameter to `SimulationEvent.__init__`:**
```python
def __init__(
    self,
    time: int | float,
    function: Callable,
    priority: Priority = Priority.DEFAULT,
    function_args: list[Any] | None = None,
    function_kwargs: dict[str, Any] | None = None,
    event_type: EventType = EventType.DEFAULT,  # NEW
) -> None:
```

**Additional fix:** Changed empty exception to descriptive error:
```python
# Before
if not callable(function):
    raise Exception()  # No message

# After
if not callable(function):
    raise TypeError(f"function must be callable, got {type(function).__name__}")
```

### 2. `mesa/experimental/devs/simulator.py`

**Updated import:**
```python
from .eventlist import EventList, EventType, Priority, SimulationEvent
```

**Added `event_type` parameter to all scheduling methods:**
- `schedule_event_now()`
- `schedule_event_absolute()`
- `schedule_event_relative()`
- `schedule_event_next_tick()` (ABMSimulator)

**Fixed `ABMSimulator.setup()` to mark step as MODEL_STEP:**
```python
def setup(self, model):
    super().setup(model)
    self.schedule_event_next_tick(
        self.model.step, priority=Priority.HIGH, event_type=EventType.MODEL_STEP
    )
```

**Fixed `ABMSimulator.run_until()` to use event_type check:**
```python
if event.event_type == EventType.MODEL_STEP:
    self.schedule_event_next_tick(
        self.model.step,
        priority=Priority.HIGH,
        event_type=EventType.MODEL_STEP,
    )
```

### 3. `mesa/experimental/devs/__init__.py`

**Exported `EventType`:**
```python
from .eventlist import EventType, Priority, SimulationEvent
__all__ = ["ABMSimulator", "DEVSimulator", "EventType", "Priority", "SimulationEvent", "Simulator"]
```

### 4. `tests/test_devs.py`

**Added 3 new test cases** to verify the fix and prevent regression.

### Test 1: `test_abm_simulator_step_rescheduling`
Verifies that the simulator correctly executes the expected number of steps.

```python
def test_abm_simulator_step_rescheduling():
    """Test that ABMSimulator correctly reschedules steps in run_until."""
    simulator = ABMSimulator()
    model = Model()
    
    simulator.setup(model)
    simulator.run_until(10)
    
    # With the bug, only 1 step would execute. With the fix, 10 steps should execute.
    assert model.steps == 10, f"Expected 10 steps, got {model.steps}"
    assert model.time == 10.0
```

### Test 2: `test_abm_simulator_event_type_tracking`
Verifies that the initial step event is correctly marked as MODEL_STEP.

```python
def test_abm_simulator_event_type_tracking():
    """Test that event types are correctly tracked in ABMSimulator."""
    from mesa.experimental.devs.eventlist import EventType
    
    simulator = ABMSimulator()
    model = Model()
    simulator.setup(model)
    
    events = simulator.event_list.peak_ahead(1)
    assert len(events) == 1
    assert events[0].event_type == EventType.MODEL_STEP
```

### Test 3: `test_user_events_not_rescheduled`
Verifies that user-scheduled events with DEFAULT type are not automatically rescheduled.

```python
def test_user_events_not_rescheduled():
    """Test that user-scheduled events are not automatically rescheduled."""
    simulator = ABMSimulator()
    model = Model()
    simulator.setup(model)
    
    user_fn = MagicMock()
    simulator.schedule_event_next_tick(user_fn)  # DEFAULT type
    
    simulator.run_for(5)
    
    user_fn.assert_called_once()  # Only called once, not rescheduled
    assert model.steps == 5       # But model steps ran 5 times
```

## Test Output

```
$ pytest tests/test_devs.py -v

============================= test session starts =============================
platform win32 -- Python 3.13.11, pytest-9.0.2, pluggy-1.6.0
collected 8 items

tests/test_devs.py::test_devs_simulator PASSED                           [ 12%]
tests/test_devs.py::test_abm_simulator PASSED                            [ 25%]
tests/test_devs.py::test_abm_simulator_step_rescheduling PASSED          [ 37%]
tests/test_devs.py::test_abm_simulator_event_type_tracking PASSED        [ 50%]
tests/test_devs.py::test_user_events_not_rescheduled PASSED              [ 62%]
tests/test_devs.py::test_simulator_time_deprecation PASSED               [ 75%]
tests/test_devs.py::test_simulation_event PASSED                         [ 87%]
tests/test_devs.py::test_eventlist PASSED                                [100%]

============================== 8 passed in 4.94s ==============================
```

### DIFF
--- a/mesa/experimental/devs/simulator.py
+++ b/mesa/experimental/devs/simulator.py
@@ -258,6 +258,11 @@ class Simulator:
             SimulationEvent: the simulation event that is scheduled
 
         """
+        # --- NEW CODE START ---
+        if time_delta < 0:
+            raise ValueError("Cannot schedule event in the past")
+        # --- NEW CODE END ---
+
         event = SimulationEvent(
             self.model.time + time_delta,
             function,


quaquel:
I have to investigate the bug a bit more. I don't fully follow your explanation. 

Additionally, I fundamentally dislike the idea of adding complexity through an event type enum. There must be a cleaner way to solve this.

PR #3014, opened 2025-12-25
fix: prevent infinite loop in select_random_empty_cell via heuristic fallback

### LINKED ISSUE #2968: Grid.select_random_empty_cell hangs on full grids

DipayanDasgupta:
**Closes #2968**

This PR fixes the infinite loop in `Grid.select_random_empty_cell` when the grid is full, while optimizing for performance on large grids. It includes a comprehensive regression test suite.

### The Issue
The original implementation used an unbounded `while True` loop to find a random empty cell. On a full grid, this condition is never met, causing the simulation to hang indefinitely.

### The Solution: Heuristic Fallback
Instead of checking every cell (O(N)) or looping indefinitely, this implementation uses a constant-time heuristic:

1.  **Try Random (O(1)):** It attempts to find an empty cell via random sampling up to **50 times**. This preserves the performance benefits for sparse grids.
2.  **Fallback (Deterministic):** If no empty cell is found after 50 tries (implying the grid is dense or full), it immediately falls back to `super().select_random_empty_cell()`. 

The parent method correctly builds a list of empty cells and selects one, or raises an `IndexError` if the grid is truly full.

### Changes
*   **Source:** Modified `mesa/discrete_space/grid.py` to replace `while True` with a bounded loop.
*   **Tests:** Added `tests/discrete_space/test_grid_sampling.py` with comprehensive coverage:
    1.  **Full Grid:** Verifies fallback is called and `IndexError` is raised.
    2.  **Sparse Grid:** Verifies random sampling returns a cell quickly (fast path).
    3.  **Nearly Full Grid:** Verifies the correct empty cell is found even when probability is low.
    4.  **Disabled Random:** Verifies `_try_random=False` forces immediate fallback.
*   **Documentation:** Preserved original context comments regarding the cutoff trade-offs.

### Verification
*   Passed all new unit tests locally.
*   Passed strict `ruff` linting and formatting.
*   Verified that `UserWarning` regarding RNG is suppressed in tests by using explicit fixtures.

### DIFF
--- a/mesa/discrete_space/grid.py
+++ b/mesa/discrete_space/grid.py
@@ -142,6 +142,14 @@ class Grid(DiscreteSpace[T], HasPropertyLayers):
 
     def select_random_empty_cell(self) -> T:  # noqa
         # Use a heuristic: try random sampling first for performance (O(1))
+        # FIXME:: basically if grid is close to 99% full, creating empty list can be faster
+        # FIXME:: note however that the old results don't apply because in this implementation
+        # FIXME:: because empties list needs to be rebuild each time
+        # This method is based on Agents.jl's random_empty() implementation. See
+        # https://github.com/JuliaDynamics/Agents.jl/pull/541. For the discussion, see
+        # https://github.com/mesa/mesa/issues/1052 and
+        # https://github.com/mesa/mesa/pull/1565. The cutoff value provided
+        # is the break-even comparison with the time taken in the else branching point.
         if self._try_random:
             # Limit attempts to avoid infinite loops on full grids
             for _ in range(50):


DipayanDasgupta:
@coderabbitai review and give feedback on improvements possible

DipayanDasgupta:
Thanks for the feedback! I have implemented the requested changes:
Added Comprehensive Tests: Created tests/discrete_space/test_grid_sampling.py which explicitly verifies:
Full grid: Confirms IndexError is raised via fallback.
Sparse grid: Confirms random sampling works quickly.
Nearly full grid: Confirms the correct empty cell is found (handling CellCollection indexing correctly).
Disabled random: Confirms _try_random=False forces immediate fallback.
Linting & Docs: Added the missing docstring to select_random_empty_cell and removed the unused # noqa directive.
All local tests and pre-commit hooks are passing

mesa/discrete_space/grid.py
```diff
@@ -140,23 +140,23 @@ def _validate_parameters(self):
...
+        """Select a random empty cell with heuristic fallback optimization."""
         # https://github.com/mesa/mesa/issues/1052 and
         # https://github.com/mesa/mesa/pull/1565. The cutoff value provided
         # is the break-even comparison with the time taken in the else branching point.
+        #
+        # FIX: Added a limit (50 attempts) to prevent infinite loops on full grids.
+        # See https://github.com/mesa/mesa/issues/2968
         if self._try_random:
-            while True:
+            for _ in range(50):
```
quaquel, line 155: I suggest turning this into an attribute. As a general rule, it's bad to have implicit constants. By making this an attribute (e.g., self._n_tries_random), users can also potentially control the behavior.

tests/discrete_space/test_grid_sampling.py
```diff
@@ -0,0 +1,84 @@
+"""Unit tests for Grid.select_random_empty_cell heuristic behavior."""
```
quaquel: please include these in the overall test file rather than adding new test files.

DipayanDasgupta:
@quaquel implemented it, _n_tries_random is now an instance attribute (default 50).
Moved the new tests into tests/discrete_space/test_grid.py as requested.

quaquel:
> Moved the new tests into tests/discrete_space/test_grid.py as requested.

Are you sure you commited this change because I still see the old test file under files changed at the top.

tests/visualization/test_solara_viz.py
```diff
@@ -1,4 +1,4 @@
-"""Test Solara visualizations."""
+"""Test Solara visualizations - Modern API."""
```
quaquel: This should not be part of the PR. Please remove any changes you have made here (probably some branching issue?)

quaquel:
I have taken the liberty to make a few additional changes to your PR so it can be merged, which I will do later today. 

Thanks so far.

tests/visualization/test_solara_viz.py
DipayanDasgupta: Ya im sorry, I was working on another branch , and probably forgot to switch, which caused all this confusion. I'm so sorry.

mesa/discrete_space/grid.py
```diff
@@ -141,22 +141,16 @@ def _validate_parameters(self):
...
-        # FIXME:: currently just a simple boolean to control behavior
-        # FIXME:: basically if grid is close to 99% full, creating empty list can be faster
-        # FIXME:: note however that the old results don't apply because in this implementation
-        # FIXME:: because empties list needs to be rebuild each time
-        # This method is based on Agents.jl's random_empty() implementation. See
-        # https://github.com/JuliaDynamics/Agents.jl/pull/541. For the discussion, see
-        # https://github.com/mesa/mesa/issues/1052 and
-        # https://github.com/mesa/mesa/pull/1565. The cutoff value provided
-        # is the break-even comparison with the time taken in the else branching point.
+        # Use a heuristic: try random sampling first for performance (O(1))
```
quaquel, line 144:
```suggestion
        # Use a heuristic: try random sampling first for performance (O(1))
        # FIXME:: basically if grid is close to 99% full, creating empty list can be faster
        # FIXME:: note however that the old results don't apply because in this implementation
        # FIXME:: because empties list needs to be rebuild each time
        # This method is based on Agents.jl's random_empty() implementation. See
        # https://github.com/JuliaDynamics/Agents.jl/pull/541. For the discussion, see
        # https://github.com/mesa/mesa/issues/1052 and
        # https://github.com/mesa/mesa/pull/1565. The cutoff value provided
        # is the break-even comparison with the time taken in the else branching point.
```

PR #3015, opened 2025-12-25
docs changed as Solaraviz only accepts keyword model arguments

### LINKED ISSUE #2494: SolaraViz docs

Akshay-krish:
Mesa models must now be instantiated using **keyword arguments** because SolaraViz reconstructs models via **kwargs. 
This aligns the documentation with the expected behavior.

This change will close #2494


quaquel:
This interacts with the need to support the new scenarios. So I am closing this issue in favor of a more comprehensive solution that ensures that scenarios also work.

PR #3022, opened 2025-12-26
Implementation of Vale

### LINKED ISSUE #1572: Add prose linter to CI

vedantvakharia:
## Summary
Closes #1572
This PR implements **Vale**, a syntax-aware linter for prose, to ensure consistent documentation style across the project. It includes configurations for GitHub Actions (CI) and local pre-commit hooks.

### 1. Configuration (`.vale.ini`)
* **Base Style:** Implements the **Google** style guide.
    * *Reasoning:* Google's guide was chosen over Microsoft's as it focuses more on technical clarity and developer documentation, whereas Microsoft's rules can be overly corporate/verbose for this context.
* **Modifications:**
    * Disabled Vale's internal spell checker to avoid redundancy (since Mesa already uses `codespell`).
    * Disabled specific Google rules that felt too strict for the current documentation state.
    * **Severity:** Google style alerts are set to `suggestion` (non-blocking) so they do not prevent commits.

### 2. Custom Project Rules (`.github/styles/mesa.yml`)
* Added a dedicated style file for project-specific terminology.
* **Current Status:** The rules are currently commented out/placeholders.
* **Behavior:** These rules are configured to trigger `error` (blocking) level alerts in pre-commit, ensuring critical terminology (e.g., "Mesa" vs "mesa") is enforced strictly once enabled.

### 3. Workflow Integration
* **Pre-commit:** Added Vale to `.pre-commit-config.yaml` for local linting.
* **CI:** Added `.github/workflows/vale.yml` to run checks on Pull Requests.
* **Gitignore:** Added Google styles to `.gitignore` to prevent cluttering the repository with external dependency files.

### 4. Exclusions
* Excluded `HISTORY.md` and `CODE_OF_CONDUCT.md` from linting to preserve historical logs and standard community text.

## Questions for Reviewers
1.  **Custom Rules:** `mesa.yml` is currently a placeholder. Are there specific terms (besides "Mesa" and "NumPy") you would like strictly enforced immediately?
2.  **Workflow:** I have enabled Vale in both CI and pre-commit. Please let me know if you prefer to restrict it to CI only.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ ci:
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.14.7
+  rev: v0.14.10
   hooks:
     # Run the linter.
     - id: ruff
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,7 +32,7 @@ discuss via [Matrix] OR via [an issue].
 - If implementing a new feature, include some documentation in docs folder.
 - Make sure that your submission works with a few of the examples in the examples repository. If adding a new feature to mesa, please illustrate usage by implementing it in an example.
 - Make sure that your submission passes the [GH Actions build]. See "Testing and Standards below" to be able to run these locally.
-- Make sure that your code is formatted according to [the black] standard (you can do it via [pre-commit]).
+- Make sure that your code is formatted according to [the ruff] standard (you can do it via [pre-commit]).
 - Push your changes to your fork on Github: `git push origin NAME_OF_BRANCH`.
 - [Create a pull request].
 - Describe the change w/ ticket number(s) that the code fixes.
@@ -78,8 +78,8 @@ Start with creating your own models, for fun. Once you have some experience, mov
 :target: https://codecov.io/gh/mesa/mesa
 ```
 
-```{image} https://img.shields.io/badge/code%20style-black-000000.svg
-:target: https://github.com/psf/black
+```{image} https://img.shields.io/badge/code%20style-ruff-000000.svg
+:target: https://github.com/astral-sh/ruff
 ```
 
 As part of our contribution process, we practice continuous integration and use GH Actions to help enforce best practices.
@@ -104,7 +104,7 @@ We test by implementing simple models and through traditional unit tests in the 
 py.test --cov=mesa tests/
 ```
 
-With respect to code standards, we follow [PEP8] and the [Google Style Guide]. We use [ruff format] (a more performant alternative to `black`) as an automated code formatter. You can automatically format your code using [pre-commit], which will prevent `git commit` of unstyled code and will automatically apply black style so you can immediately re-run `git commit`. To set up pre-commit run the following commands:
+With respect to code standards, we follow [PEP8] and the [Google Style Guide]. We use [ruff format] (a more performant alternative to `black`) as an automated code formatter. You can automatically format your code using [pre-commit], which will prevent `git commit` of unstyled code and will automatically apply ruff style so you can immediately re-run `git commit`. To set up pre-commit run the following commands:
 
 ```bash
 pip install pre-commit
@@ -326,7 +326,7 @@ A special thanks to the following projects who offered inspiration for this cont
 [18f's foia]: https://github.com/18F/foia-hub/blob/master/CONTRIBUTING.md
 [18f's midas]: https://github.com/18F/midas/blob/devel/CONTRIBUTING.md
 [an issue]: https://github.com/mesa/mesa/issues
-[black]: https://github.com/psf/black
+[ruff]: https://github.com/astral-sh/ruff
 [clone your repository]: https://help.github.com/articles/cloning-a-repository/
 [create a pull request]: https://help.github.com/articles/creating-a-pull-request/
 [django]: https://github.com/django/django/blob/master/CONTRIBUTING.rst
--- a/docs/tutorials/4_visualization_basic.ipynb
+++ b/docs/tutorials/4_visualization_basic.ipynb
@@ -6,6 +6,7 @@
    "source": [
     "# Visualization - Basic Dashboard\n",
     "\n",
+    "\n",
     "### The Boltzmann Wealth Model "
    ]
   },
@@ -60,6 +61,14 @@
     "\n",
     "import mesa\n",
     "from mesa.discrete_space import CellAgent, OrthogonalMooreGrid\n",
+    "\n",
+    "# Check Mesa version for visualization compatibility\n",
+    "if mesa.__version__.startswith((\"3.0\", \"3.1\", \"3.2\")):\n",
+    "    print(\n",
+    "        f\"⚠️  Mesa {mesa.__version__} detected. Visualization features require Mesa 3.3+\"\n",
+    "    )\n",
+    "    print(\"To upgrade: pip install --upgrade mesa\")\n",
+    "\n",
     "from mesa.visualization import SolaraViz, SpaceRenderer, make_plot_component\n",
     "from mesa.visualization.components import AgentPortrayalStyle"
    ]
--- a/docs/tutorials/5_visualization_dynamic_agents.ipynb
+++ b/docs/tutorials/5_visualization_dynamic_agents.ipynb
@@ -6,6 +6,7 @@
    "source": [
     "# Visualization - Dynamic Agents\n",
     "\n",
+    "\n",
     "### The Boltzmann Wealth Model "
    ]
   },
@@ -60,6 +61,14 @@
     "\n",
     "import mesa\n",
     "from mesa.discrete_space import CellAgent, OrthogonalMooreGrid\n",
+    "\n",
+    "# Check Mesa version for visualization compatibility\n",
+    "if mesa.__version__.startswith((\"3.0\", \"3.1\", \"3.2\")):\n",
+    "    print(\n",
+    "        f\"⚠️  Mesa {mesa.__version__} detected. Visualization features require Mesa 3.3+\"\n",
+    "    )\n",
+    "    print(\"To upgrade: pip install --upgrade mesa\")\n",
+    "\n",
     "from mesa.visualization import SolaraViz, SpaceRenderer, make_plot_component\n",
     "from mesa.visualization.components import AgentPortrayalStyle"
    ]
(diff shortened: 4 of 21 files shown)


HISTORY.md
```diff
@@ -1,147 +1,8 @@
+<!-- vale off -->
+
 ---
 title: Release History
 ---
```
quaquel, line 5: how come the content of this file is fully removed. That should not happen.

quaquel:
Thanks for this PR. I am curious about @EwoutH take on this. But as is clear from some of the changed, files, it does seem helpful in fixing basic typo's.

HISTORY.md
vedantvakharia: Apologies for that! It was an accidental deletion that happened. I have restored the original content in the latest commit. It should be correct now.

Also, can you also tell what custom rules you need to be implemented? and whether you need vale in both CI and pre-commit?

quaquel: > Also, can you also tell what custom rules you need to be implemented? 

No idea, which is why I asked @EwoutH for input

> and whether you need vale in both CI and pre-commit?

precommit makes the most sense to me, we run ruff also on precommit if I am not mistaken.

.github/styles/Mesa/README.md
EwoutH, line 1: Please reduce this document to the bare essentials, we don’t need a copy of the Vale documentation (I found Understanding Vale Output and Resources the most useful sections).

Focus on a written-out explanation how it works and why it’s configured this exact way (in our CI). All the other stuff can go.

.github/workflows/vale.yml
```diff
@@ -0,0 +1,53 @@
...
+    name: Run Vale
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install Vale
+        run: |
+          wget https://github.com/errata-ai/vale/releases/download/v3.0.7/vale_3.0.7_Linux_64-bit.tar.gz
```
EwoutH: I don’t like hardcoded versions. Are there alternatives?

.github/styles/Mesa/mesa.yml
EwoutH, line 1: Can configuration also be done in pyproject.toml?

.pre-commit-config.yaml
```diff
@@ -28,3 +29,32 @@ repos:
   rev: v2.4.1
   hooks:
     - id: codespell
+
+# Prose linting with Vale
+- repo: https://github.com/errata-ai/vale
```
EwoutH, line 34: Do we both need to be in GitHub actions and Pre-commit? If not, I prefer only being in pre-commit.

HISTORY.md
```diff
@@ -1,6 +1,9 @@
+<!-- vale off -->
```
EwoutH, line 1: I don’t understand how vale is off here, but there are still changes in the document

vedantvakharia: I first made vale run through the file to identify any spelling errors in the file. Then I turned vale off for that file to avoid noise from false positives, specifically regarding spell checking.

**The Issue:**
Vale's default spell checker treats technical terms (e.g., variable names, module paths, library-specific jargon) as spelling errors if they aren't in its standard dictionary. Since `HISTORY.md` is a log of code changes, it is naturally full of these terms.

**What I tried first:**
I initially attempted to fix this by building a project-specific vocabulary (`accept.txt`), adding valid technical terms to it so Vale would recognize them.

**Why that didn't work:**
I quickly found that the number of unique variable names and technical terms in the history log is massive. Maintaining an `accept.txt` list for every single new parameter or method mentioned in the changelog would become a significant maintenance burden (the file would grow indefinitely and require constant updates).


Since we already use `codespell` (which is better tuned for codebases), I decided it was cleaner to rely on `codespell` for catching typos in these specific files and use Vale purely for enforcing style/prose in the documentation, where it shines. If you need vale to be enabled in those 2 files, I can make changes to the file and enable it.

.pre-commit-config.yaml
vedantvakharia: There is a technical catch with using only pre-commit.ci:

The pre-commit.ci bot runs in an isolated environment and cannot run vale sync to download the external Google styles. This causes Runtime Error unless the styles are present locally.

If we want to use only pre-commit, I will need to commit the entire Google style definition folder (inside .github/styles/Google) directly into the repository so the bot can access them.

**Why I chose the Action:** I saw that in PR #2568, there was hesitation about adding the Microsoft style files directly to the repo. To respect that preference and keep the repo clean, I used a GitHub Action to download the styles on the fly, rather than committing them.

Are you comfortable with committing the Google style files (~33 YAML files) to the repository? If so, I can switch to pre-commit only. Otherwise, the GitHub Actions workflow provides a cleaner solution.
Let me know your preference!

.github/styles/Mesa/mesa.yml
vedantvakharia: Unfortunately, Vale does not support `pyproject.toml` for its configuration. Vale is language-agnostic (written in Go) and exclusively uses the `.vale.ini` format for its main configuration file.

This is standard practice across all projects using Vale, including large projects like GitLab and Qt Documentation. Since Vale is designed to work with prose in any programming language ecosystem (not just Python), it doesn't integrate with Python-specific configuration files like `pyproject.toml`.

The `.vale.ini` file at the repository root is required and follows Vale's official documentation and community best practices.

**References:**
- [Vale Configuration Documentation](https://vale.sh/docs/topics/config/)
- [GitLab's [.vale.ini](cci:7://file:///C:/Users/Vedant/mesa/.vale.ini:0:0-0:0)](https://gitlab.com/gitlab-org/gitlab/-/blob/master/.vale.ini)

.pre-commit-config.yaml
EwoutH: So a GitHub Action can dynamically load them while pre-commit cannot?

.github/styles/Mesa/mesa.yml
EwoutH: Sounds reasonable, thanks

.github/styles/Mesa/README.md
vedantvakharia: Thanks for the feedback @EwoutH! I've addressed your comments:
**README.md**: Streamlined from 326 to ~70 lines, focusing only on Mesa-specific configuration

Let me know if you'd like any more adjustments.

.github/workflows/vale.yml
vedantvakharia: Thanks for the feedback @EwoutH! I've addressed your comments:
**Hardcoded version**: Replaced manual installation with official `errata-ai/vale-action@v2.1.1`

Let me know if you'd like any other adjustments!

HISTORY.md
EwoutH: Would it make sense to limit it to certain folders?

.github/styles/Mesa/README.md
EwoutH: This is great, thanks

.pre-commit-config.yaml
vedantvakharia: Yes, exactly.

- GitHub Actions allows us to run shell commands, so the workflow I added explicitly runs vale sync to download the styles before the check starts.
- pre-commit.ci is a stricter, sandboxed environment. It does not run arbitrary setup commands (like vale sync) before the hooks. It expects all configuration files to be present in the repository itself.

That is why the Action works 'out of the box' while pre-commit.ci fails unless we commit the style files.

EwoutH: Right, so in our case, what does vale in pre-commit do?

HISTORY.md
vedantvakharia: Yes, we can certainly do that.

Currently, the configuration targets the `docs/` folder AND `*.md` (files in the root directory).

If we limit Vale strictly to the `docs/` folder:
* **Pro:** It automatically ignores `HISTORY.md` and `CODE_OF_CONDUCT.md` without needing any exclude rules.
* **Con:** It would stop linting the main `README.md` of the project (since it sits in the root), and that is often the most visible file for new users.

I stuck with the current approach (checking root files + manual excludes) specifically to keep the main `README.md` covered.

.pre-commit-config.yaml
vedantvakharia: In our current setup, Vale in pre-commit acts as an immediate feedback tool. Provided the developer has run `vale sync` once on their machine, pre-commit catches style errors *before* they push code. This saves them from waiting for CI to fail. We keep the pre-commit hook solely for the benefit of local developers (so they can lint while they code). We rely on the GitHub Action to perform the actual "official" check on the server.

HISTORY.md
EwoutH: Let’s start in this PR with `/docs` only, and then we can discuss if we want to extend it later.

vedantvakharia: I have updated the PR to include only the changes in `/docs`.

quaquel: @EwoutH, can you confirm that this PR is now ready to be merged?

EwoutH:
Running pre-commit locally, I get:

image

What do I need to do to resolve this?

quaquel:
my hunch is that some vale init needs to be run locally to download the relevant styles.

vedantvakharia:
Hi @EwoutH, thanks for testing this locally!

The error `style 'Google' does not exist` happens because I haven't committed the Google style definitions to the repository to keep our git history clean and avoid bloating the repo with upstream style files.

To fix this, you just need to fetch the styles once. Please run:

```bash
vale sync
```

quaquel:
Might it be an idea to update `contributing.md` to explicitly explain this?

EwoutH:
Good idea, agreed.

@vedantvakharia could you add a small snippet under https://github.com/mesa/mesa/blob/main/CONTRIBUTING.md#testing-and-code-standards?

EwoutH:
While I appreciate the effort, I would suggest revering this PR. pre-commit is already complex, especially for new contributors, and this makes it even more complex. I don't think that complexity weights up against the limited value vale adds.

Maybe we can run it separately, in an isolated GitHub action, once a month or so.

quaquel:
I don't understand your point. I haven't run into any issues with vale yet in precommit.

PR #3024, opened 2025-12-27
Fix network visualization bug

codebyNJ:
Fixes a critical bug where agents on network nodes with non-contiguous IDs were silently dropped from visualization.

**Problem:** Code used node IDs as array indices instead of dictionary keys, causing IndexError that was silently suppressed.

**Solution:** Replace array indexing with dictionary lookup with proper error handling.

**Before:** Nodes [0,1,5,10,15] → 0/5 agents visible (all lost)
**After:**  Nodes [0,1,5,10,15] → 5/5 agents visible 

**Tests:** 6 comprehensive unit tests added covering:
- Non-contiguous node IDs
- Missing node positions with warnings
- Large and negative node IDs
- Edge cases (empty, single node, duplicates)

**Files changed:**
- `mesa/visualization/space_renderer.py` - Fix implementation
- `tests/visualization/test_space_renderer.py` - Unit tests

**Checklist:**
- [x] Bug fix (non-breaking change)
- [x] Tests added and passing
- [x] FIXME comment resolved
- [x] No breaking changes

### DIFF
--- a/tests/examples/test_examples_viz.py
+++ b/tests/examples/test_examples_viz.py
@@ -72,12 +72,12 @@ def run_model_test(
         # Display and capture the updated visualizations
         display(space_viz)
         page_session.wait_for_selector("img")
-        changed_space = page_session.locator("img").screenshot()
+        changed_space = page_session.locator("img").first.screenshot()
 
         if measure_config:
             display(graph_viz)
             page_session.wait_for_selector("img")
-            changed_graph = page_session.locator("img").screenshot()
+            changed_graph = page_session.locator("img").last.screenshot()
 
         # Convert screenshots to base64 for comparison
         initial_space_encoding = base64.b64encode(initial_space).decode()


Sahil-Chhoker:
@codebyNJ Thanks for this PR and finding out the bug, can you give me an example on which this bug can be reproduced, just drop it in a comment like this so it doesn't take much space:

<details>
<summary> Example </summary>

```python

import ...

```
</details>

codebyNJ:
@Sahil-Chhoker I have reproduced.

**To Reproduce**

```python
import networkx as nx
from mesa import Agent, Model
from mesa.space import NetworkGrid
from mesa.visualization import SpaceRenderer
import numpy as np
import contextlib

# Create network with non-contiguous node IDs
G = nx.Graph()
G.add_nodes_from([0, 1, 5, 10, 15])
G.add_edges_from([(0, 1), (1, 5), (5, 10), (10, 15)])

model = Model()
space = NetworkGrid(G)
model.grid = space

for node_id in [0, 1, 5, 10, 15]:
    agent = Agent(model)
    space.place_agent(agent, node_id)

renderer = SpaceRenderer(model)

# Demonstrate the actual bug - what old code did
print("BEFORE FIX (old code with array indexing):")

pos_array = np.array([[0.1, 0.2], [0.3, 0.4], [0.5, 0.6], [0.7, 0.8], [0.9, 1.0]])
node_ids = np.array([[0], [1], [5], [10], [15]], dtype=float)

mapped_positions_old = []
for node_id in node_ids[:, 0].astype(int):
    with contextlib.suppress(IndexError):  # This is what the old code did!
        pos = pos_array[node_id]
        mapped_positions_old.append(pos)

print(f"Agents mapped: {len(mapped_positions_old)}/5")
print(f"Agents silently dropped: {5 - len(mapped_positions_old)}/5")
print(f"Which nodes were dropped: {[int(n) for n in node_ids[:, 0] if int(n) >= len(pos_array)]}")

# Demonstrate the fix - what new code does
print("\nAFTER FIX (new code with dictionary lookup):")

pos_dict = {
    0: np.array([0.1, 0.2]),
    1: np.array([0.3, 0.4]),
    5: np.array([0.5, 0.6]),
    10: np.array([0.7, 0.8]),
    15: np.array([0.9, 1.0])
}

mapped_positions_new = []
for node_id in node_ids[:, 0].astype(int):
    if node_id in pos_dict:  # Dictionary lookup - works for any node ID
        mapped_positions_new.append(pos_dict[node_id])

print(f"Agents mapped: {len(mapped_positions_new)}/5")
print(f"All agents visible: {len(mapped_positions_new) == 5}")
```

**Explaination**

Networks with non-contiguous node IDs (like `[0, 1, 5, 10, 15]` ) caused the old code to silently fail when using array indexing. By replacing array indexing with dictionary lookup, ensures all agents are properly visualized regardless of node ID values.

PR #3031, opened 2025-12-28
Fix memory leak in ContinuousSpace agent removal

### LINKED ISSUE #3029: Fix memory leak in ContinuousSpace agent removal
When removing an agent from `ContinuousSpace`, the internal index mapping (`_index_to_agent`) is corrupted. The code shifts subsequent agents to fill the gap but fails to delete the old index entry of the last shifted agent. This creates a "Ghost Agent" entry: two indices point to the same agent, and the dictionary size grows larger than the number of active agents.

**Expected behavior**
The `_index_to_agent` dictionary should accurately reflect the active agents. When an agent is removed:
1. The removed agent's entry is deleted.
2. Subsequent agents are re-indexed.
3. The old index of the last shifted agent is deleted.
4. `len(_index_to_agent)` == `len(active_agents)`.

**To Reproduce**
Run the following script to reproduce the issue:

```python
from mesa import Agent, Model
from mesa.experimental.continuous_space import ContinuousSpace
from mesa.experimental.continuous_space.continuous_space_agents import ContinuousSpaceAgent
import numpy as np

class DummyAgent(ContinuousSpaceAgent):
    pass

model = Model()
space = ContinuousSpace(dimensions=[[0, 10], [0, 10]], torus=False)

# Create 3 agents
agents = []
for i in range(3):
    agent = DummyAgent(space, model)
    agent.position = np.array([float(i), float(i)])
    agents.append(agent)

print(f"Initial agents: {len(space.active_agents)}")
print(f"Initial index map size: {len(space._index_to_agent)}")

# Remove the middle agent (index 1)
agent_to_remove = agents[1]
print(f"Removing agent at index: {space._agent_to_index[agent_to_remove]}")
space._remove_agent(agent_to_remove) 

print(f"Agents after removal: {len(space.active_agents)}")
print(f"Index map size after removal: {len(space._index_to_agent)}")

# Check for stale index
expected_size = len(space.active_agents)
actual_size = len(space._index_to_agent)

if actual_size > expected_size:
    print("BUG DETECTED: _index_to_agent has stale entries!")
    print(f"Expected size: {expected_size}, Actual size: {actual_size}")
    last_index = actual_size - 1
    print(f"Entry at index {last_index}: {space._index_to_agent.get(last_index)}")
```

**Additional context**
- **Severity:** High (Memory Leak + Logic Error)
- **Location:** `mesa/experimental/continuous_space/continuous_space.py`
- **Logic Demonstration:**
  Imagine a list `[A, B, C]`. Removing `B` shifts `C` to index 1.
  - **Correct Result:** `[A, C]`
  - **Buggy Result:** `[A, C, C]` (The old `C` at index 2 remains).

Nithin9585:
Fixes #3029

## Description

When removing an agent from [ContinuousSpace](cci:2://file:///c:/Users/Nithi/OneDrive/Pictures/Desktop/gsoc/mesa/mesa/experimental/continuous_space/continuous_space.py:14:0-272:9), the internal index mapping (`_index_to_agent`) was corrupted. The code shifted subsequent agents to fill the gap but failed to delete the old index entry of the last shifted agent. This created a "Ghost Agent" entry where two indices pointed to the same agent, causing the dictionary to grow larger than the number of active agents.

## Root Cause

In the [_remove_agent](cci:1://file:///c:/Users/Nithi/OneDrive/Pictures/Desktop/gsoc/mesa/mesa/experimental/continuous_space/continuous_space.py:139:4-161:72) method, when agents were shifted down to fill gaps, the old index entry of the last shifted agent was never deleted.

**Example:** Removing agent B from `[A, B, C]`:
- Before: `_index_to_agent = {0: A, 1: B, 2: C}`
- After (buggy): `_index_to_agent = {0: A, 1: C, 2: C}` (ghost entry at index 2)
- After (fixed): `_index_to_agent = {0: A, 1: C}`

## Changes

- Added tracking of `last_old_index` during agent shifting
- Added cleanup logic to remove stale index entries
- Added comprehensive test coverage with 3 scenarios

## Impact

- **Severity**: High (memory leak + data corruption)
- **Breaking Changes**: None
- **Performance**: Negligible overhead

### DIFF
--- a/mesa/experimental/continuous_space/continuous_space.py
+++ b/mesa/experimental/continuous_space/continuous_space.py
@@ -148,7 +148,7 @@ class ContinuousSpace:
         self._index_to_agent.pop(index, None)
         del self.active_agents[index]
 
-        # Shift all subsequent agents down by 1
+        # Shift all subsequent agents up by 1
         for agent in self.active_agents[index::]:
             old_index = self._agent_to_index[agent]
             self._agent_to_index[agent] = old_index - 1


mesa/experimental/continuous_space/continuous_space.py
```diff
@@ -149,10 +149,15 @@ def _remove_agent(self, agent: Agent) -> None:
         del self.active_agents[index]
 
         # we update all indices
+        last_old_index = None
         for agent in self.active_agents[index::]:
             old_index = self._agent_to_index[agent]
             self._agent_to_index[agent] = old_index - 1
             self._index_to_agent[old_index - 1] = agent
+            last_old_index = old_index
```
quaquel: By definition, `last_old_index = self.active_agents[-1]`, so there is little need to update this inside the loop.

mesa/experimental/continuous_space/continuous_space.py
```diff
@@ -149,10 +149,15 @@ def _remove_agent(self, agent: Agent) -> None:
...
+
+        if last_old_index is not None:
+            self._index_to_agent.pop(last_old_index, None)
```
quaquel: this can only happen in the number of agents is 0, right?

quaquel:
The bug is even a bit more severe, at least theoretically. Since dictionaries are ordered in Python by insertion, the old index in `_index_to_agent` will be used, resulting in the wrong agent being returned. 

edit: at present, `_index_to_agent` is not used by the class itself for anything, so although the bug being fixed here exists, it has not affected the behavior and performance significantly (outside of the slight memory leak).

mesa/experimental/continuous_space/continuous_space.py
Nithin9585: It happens when removing the last agent (nothing to shift), not specifically when count reaches 0. But since line 148 already deletes that index, the None check just prevents redundant cleanup.

mesa/experimental/continuous_space/continuous_space.py
Nithin9585: since the last shifted agent is always at `active_agents[-1]`, its old index is just `len(active_agents)` (after deletion). I can simplify this to avoid updating `last_old_index` on every iteration.

Nithin9585:
@quaquel Good point about dict ordering! The fix prevents both the memory leak and potential wrong lookups if `_index_to_agent` gets used in the future. Thanks for the thorough review!

mesa/experimental/continuous_space/continuous_space.py
```diff
@@ -148,12 +148,16 @@ def _remove_agent(self, agent: Agent) -> None:
         self._index_to_agent.pop(index, None)
         del self.active_agents[index]
 
-        # we update all indices
+        # Shift all subsequent agents down by 1
```
quaquel:
```suggestion
        # Shift all subsequent agents up by 1
```

quaquel:
Thanks!

PR #3036, opened 2025-12-29
Add initialization check in Simulator.run_for()

codebreaker32:
### **Summary**
Implemented a mandatory initialization check in `Simulator.run_for()` to explicitly validate that the simulator has been set up before execution.

### **Motive**
The `run_for` method previously contained a _#fixme comment noting the need for an initialization check_. Without this check, attempting to run the simulator without a model lead to error
```plaintext
Nonetype has no attribute 'time'
```

### **Implementation**
Replaced the placeholder comment with a conditional check.
```python
if self.model is None:
    raise RuntimeError(
        "Simulator not set up. Call simulator.setup(model) first."
    )
```

### **Usage Examples**
This change ensures that improper usage is caught immediately:

- Forgot to call sim.setup(model)
```python
try:
    sim.run_for(10)
except RuntimeError as e:
    print(e) 
    # Output: Simulator not set up. Call simulator.setup(model) first.
```

### DIFF
--- a/mesa/datacollection.py
+++ b/mesa/datacollection.py
@@ -37,6 +37,8 @@ import warnings
 from copy import deepcopy
 from functools import partial
 
+from mesa import Agent
+
 with contextlib.suppress(ImportError):
     import pandas as pd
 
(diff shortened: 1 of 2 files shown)


mesa/experimental/devs/simulator.py
```diff
@@ -163,8 +163,14 @@ def run_for(self, time_delta: int | float):
...
 
+        Raises:
+            Exception if simulator.setup() has not yet been called
+
         """
-        # fixme, raise initialization error or something like it if model.setup has not been called
+        if self.model is None:
+            raise RuntimeError(
+                "Simulator not set up. Call simulator.setup(model) first."
+            )
```
quaquel: why not do

```python
try:
    end_time = self.model.time + time_delta
except AttributeError as e:
    raise RuntimeError(
                "Simulator not set up. Call simulator.setup(model) first."
            ) from e
else:
    self.run_until(end_time)

```

which is a more pythonic way to achieve the same thing.

codebreaker32:
Made the changes @quaquel

quaquel:
Can you update tests/experimental/test_devs.py with appropriate tests that cover both the bug and the updated behavior? Once those tests are in, this seems good to go.

codebreaker32:
I have added the required tests also

PR #3038, opened 2025-12-29
Fix: Add deepcopy to agent reporters to prevent mutable reference lea…

### LINKED ISSUE #3035: DataCollector Mutable Reference Leak
The `DataCollector` fails to create deep copies of agent-level data when collecting. When an agent reporter returns a mutable object (list, dict, numpy array, etc.), the `DataCollector` stores a reference to that object instead of a copy. As the agent modifies this object in subsequent steps, all historical records are retroactively updated to reflect the current state, completely invalidating longitudinal data analysis.

**Expected behavior**

When collecting agent data at each step, the `DataCollector` should preserve the state of mutable objects at that specific time step. Historical records should remain unchanged when agents modify their attributes in later steps.

For example:
- Step 1: Agent has `grades = []`, DataCollector should record `[]`
- Step 2: Agent has `grades = [85]`, DataCollector should record `[85]`
- Step 3: Agent has `grades = [85, 92]`, DataCollector should record `[85, 92]`

When reviewing historical data, Step 1 should still show `[]`, Step 2 should show `[85]`, and Step 3 should show `[85, 92]`.

**To Reproduce**

```python
from mesa.model import Model
from mesa.agent import Agent

class TestAgent(Agent):
    def __init__(self, model):
        super().__init__(model)
        self.my_list = []  # Mutable attribute
    
    def step(self):
        self.my_list.append(self.model.steps)

class TestModel(Model):
    def __init__(self):
        super().__init__()
        self.agent = TestAgent(self)
        
        # Track the mutable list
            agent_reporters={"MyList": lambda a: a.my_list}
        )
    
    def step(self):
        self.agent.step()

# Run simulation
model = TestModel()
model.step()  # Step 1: list is []
model.step()  # Step 2: list is [1]
model.step()  # Step 3: list is [1, 2]

# Check historical data
print(df)

# BUG: All steps show [1, 2, 3] instead of their historical values
# Expected:
#   Step 1: []
#   Step 2: [1]
#   Step 3: [1, 2]
# Actual:
#   Step 1: [1, 2, 3]
#   Step 2: [1, 2, 3]
#   Step 3: [1, 2, 3]
```

**Additional context**

- This bug affects any mutable data type: lists, dicts, sets, numpy arrays, custom objects, etc.
- The `DataCollector` already uses `deepcopy()` for model-level reporters (line 330 in `datacollection.py`) but not for agent-level reporters
- The fix is to apply `deepcopy()` to agent reporter results in the `_record_agents()` method (around line 284)

This bug silently corrupts historical data, making research conclusions based on Mesa simulations potentially invalid when tracking mutable agent attributes.

Nithin9585:
Fixes #3035

The `DataCollector` was storing references to mutable objects (lists, dicts, numpy arrays, etc.) returned by agent reporters instead of creating deep copies. This caused all historical records to retroactively update when agents modified these objects in subsequent steps, completely invalidating longitudinal data analysis.

## The Problem

When an agent reporter returns a mutable object, the `DataCollector` stores a reference to that object. As the simulation progresses and the agent modifies the object, the historical data changes to reflect the current state rather than the state at the time of collection.

**Example:**
```python
# Step 1: Agent has list = []
datacollector.collect()  # Should record []

# Step 2: Agent has list = [1]  
datacollector.collect()  # Should record [1]

# Bug: When checking Step 1's data later, it shows [1] instead of []
```

## The Solution

Apply `deepcopy()` to agent reporter results in both `_record_agents()` and `_record_agenttype()` methods, consistent with the existing behavior for model-level reporters.

## Changes Made

- **`mesa/datacollection.py`**: Add `deepcopy()` to agent reporter results (lines 284, 296)
- **`tests/test_datacollector_mutable_fix.py`**: Add test to verify mutable data independence

## Testing

- New test `test_mutable_data_independence` passes
-  All existing DataCollector tests pass (15/15)
-  No regressions introduced

## Impact

- **Data Integrity**: Historical agent data now correctly preserves state at each collection time
- **Compatibility**: Works with all mutable types (lists, dicts, sets, numpy arrays, custom objects)

### DIFF
--- a/mesa/datacollection.py
+++ b/mesa/datacollection.py
@@ -37,6 +37,8 @@ import warnings
 from copy import deepcopy
 from functools import partial
 
+from mesa import Agent
+
 with contextlib.suppress(ImportError):
     import pandas as pd
 
@@ -290,11 +292,20 @@ class DataCollector:
     def _record_agents(self, model):
         """Record agents data in a mapping of functions and agents."""
         rep_funcs = self.agent_reporters.values()
+        # Immutable types that don't need deepcopy
+        python_immutable_types = (str, int, bool, float, tuple)
 
         def get_reports(agent):
             _prefix = (agent.model.steps, agent.unique_id)
-            reports = tuple(deepcopy(rep(agent)) for rep in rep_funcs)
-            return _prefix + reports
+            reports = []
+            for rep in rep_funcs:
+                value = rep(agent)
+                # Only deepcopy mutable types to avoid performance overhead
+                if isinstance(value, python_immutable_types):
+                    reports.append(value)
+                else:
+                    reports.append(deepcopy(value))
+            return _prefix + tuple(reports)
 
         agent_records = map(get_reports, model.agents)
         return agent_records
@@ -302,18 +313,25 @@ class DataCollector:
     def _record_agenttype(self, model, agent_type):
         """Record agent-type data in a mapping of functions and agents."""
         rep_funcs = self.agenttype_reporters[agent_type].values()
+        # Immutable types that don't need deepcopy
+        python_immutable_types = (str, int, bool, float, tuple)
 
         def get_reports(agent):
             _prefix = (agent.model.steps, agent.unique_id)
-            reports = tuple(deepcopy(rep(agent)) for rep in rep_funcs)
-            return _prefix + reports
+            reports = []
+            for rep in rep_funcs:
+                value = rep(agent)
+                # Only deepcopy mutable types to avoid performance overhead
+                if isinstance(value, python_immutable_types):
+                    reports.append(value)
+                else:
+                    reports.append(deepcopy(value))
+            return _prefix + tuple(reports)
 
         agent_types = model.agent_types
         if agent_type in agent_types:
             agents = model.agents_by_type[agent_type]
         else:
-            from mesa import Agent  # noqa: PLC0415
-
             if issubclass(agent_type, Agent):
                 agents = [
                     agent for agent in model.agents if isinstance(agent, agent_type)


tests/test_datacollector_mutable_fix.py
```diff
@@ -0,0 +1,49 @@
+"""Test DataCollector correctly deep-copies mutable agent data."""
+
+import unittest
```
quaquel: Can you rewrite these tests to use pytest and include them in the proper file that contains the tests for the data collector?

Nithin9585: Done! Converted to pytest and moved into [tests/test_datacollector.py](cci:7://file:///c:/Users/Nithi/OneDrive/Pictures/Desktop/gsoc/mesa/tests/test_datacollector.py:0:0-0:0). All tests passing

quaquel:
Thanks for this PR and for addressing my initial comments. 

As you can see in the benchmark, deepcopy adds substantial overhead. So, I am not sure whether this is the ideal solution for this bug. It is a wasteful operation in the case of immutables, and expensive in the case of complicated, nested data structures. Moreover, if I am not mistaken, returning collections will break some of the downstream behavior of the data collector, such as returning dataframes. So I am curious if there are alternative solutions that don't have this overhead.

One option is to simply not allow mutables as a return to datacollector. Which, in my experience, should not be a significant problem since people typically collect simple data (i.e., numbers instead of mutable collections).

codebyNJ:
@Nithin9585 @quaquel 

I guess we can use type-aware approach. 
```python
def get_reports(agent):
        _prefix = (agent.model.steps, agent.unique_id)
        reports = []
        for rep in rep_funcs:
            value = rep(agent)
            # only copy known mutable types
            if isinstance(value, (list, dict, set, bytearray)):
                reports.append(deepcopy(value))
            elif hasattr(value, '__array__'):  # numpy arrays
                reports.append(value.copy())
            else:
                reports.append(value)  # immutable types: int, float, str, tuple
        return _prefix + tuple(reports)
```

quaquel:
I have to think about this a bit more. Your code, as given, will solve the problem only for known mutable types, but will still silently work fine for user-defined classes. I am inclined instead to limit data that can be collected to only known built-in immutable types. Which is a much tighter constraint. 

Alternatively, you could do something like

```python
python_immutable_types = [str, int, long, bool, float, tuple,]

if insintance(value, *python_immutable_types)
   reports.append(value)
else:
   reports.append(copy.deepcopy(value))
```

At least, this avoids the expensive deepcopy in 80% of all cases, if not more.

Nithin9585:
@quaquel  Implemented your suggested approach:
```
python
python_immutable_types = (str, int, bool, float, tuple)
if isinstance(value, python_immutable_types):
    reports.append(value)
else:
    reports.append(deepcopy(value))
```
This avoids deepcopy overhead for primitives (80%+ of cases) while still fixing the bug. All 16 tests pass.

PR #3041, opened 2025-12-30
Fix/datacollector missing attribute error

codebyNJ:
## Problem

DataCollector silently skipped reporters for non-existent agent attributes, 
causing silent data loss.

**Solution**
Added validation to raise AttributeError with informative message including:
- Agent ID
- Agent type
- Missing attribute name  
- Reporter name

**Changes**
- Modified `mesa/datacollection.py`
- Added tests to `tests/test_datacollector.py`

**Testing**
- All 18 tests passing
- No regressions
- Backward compatible

**To Reproduce** 
Run the below code

```python
from mesa import Agent, Model

class TestAgent(Agent):
    def __init__(self, model):
        super().__init__(model)
        self.wealth = 100
    
    def step(self):
        pass

class TestModel(Model):
    def __init__(self):
        super().__init__()
        agent = TestAgent(self)
        self.add_agent(agent)
    
    def step(self):
        for agent in self.agents:
            agent.step()

model = TestModel()

dc = DataCollector(
    agent_reporters={
        "wealth": "wealth",
        "health": "health",
        "status": "status"
    }
)

dc.collect(model)
print(dc.get_agent_vars_dataframe())
```

### DIFF
--- a/.github/workflows/build_lint.yml
+++ b/.github/workflows/build_lint.yml
@@ -73,9 +73,7 @@ jobs:
       - run: pip install uv
       - run: uv pip install --system .[dev]
       - run: playwright install chromium-headless-shell
-      - run: |
-          coverage run -m pytest tests/
-          coverage xml
+      - run: python -Im pytest -p pytest_cov --cov --cov-report=xml --cov-report=markdown-append:"${GITHUB_STEP_SUMMARY}" --cov-context=test tests/
       - uses: codecov/codecov-action@v5
         with:
           fail_ci_if_error: true
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -177,6 +177,42 @@ All roles are expected to:
 
 Mesa grows through good ideas and contributions. We're all volunteers working together to make Mesa better. Don't hesitate to reach out to any maintainer to discuss your interests and potential growth within the project!
 
+## Mesa examples structure and policy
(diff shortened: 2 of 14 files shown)


EwoutH:
Thanks for the PR.

It now changes over a 1000 lines of code, that can’t be right?

codebyNJ:
@EwoutH sorry it's a wrong push I will update it. It's just few lines update.

codebyNJ:
@EwoutH I have updated the PR.

mesa/datacollection.py
```diff
@@ -215,7 +215,13 @@ def _new_agent_reporter(self, name, reporter):
             attribute_name = reporter
 
             def attr_reporter(agent):
-                return getattr(agent, attribute_name, None)
+                # Validate that the attribute exists on the agent
+                if not hasattr(agent, attribute_name):
```
quaquel: I don't like this design. Its not pythonic because it involves checking if something exists before using it. Is it possible to use try catch instead? So something like

```python
try: 
    return getattr(agent, attribute_name)
except AttributeError as e:
     raise AttribeError( f"Agent {agent.unique_id} of type {type(agent).__name__} "
                                    f"has no attribute '{attribute_name}' (reporter: '{name}')"
                    )) from e
```

codebyNJ: Yes it's possible. I have updated the PR with your suggested methodology.

PR #3044, opened 2025-12-30
docs: clarify SolaraViz keyword-only model arguments

Arjun-Sasi:
### What this PR does

- Documents that SolaraViz requires Mesa models to be instantiated using keyword-only arguments.
- Adds a clear example showing supported vs unsupported usage.

### Why this is needed

SolaraViz internally creates models using keyword arguments, which is currently undocumented and can confuse new users.

### DIFF
--- a/mesa/experimental/devs/eventlist.py
+++ b/mesa/experimental/devs/eventlist.py
@@ -25,8 +25,11 @@ from collections.abc import Callable
 from enum import IntEnum
 from heapq import heapify, heappop, heappush, nsmallest
 from types import MethodType
-from typing import Any
+from typing import TYPE_CHECKING, Any
 from weakref import WeakMethod, ref
+
+if TYPE_CHECKING:
+    from mesa import Model
 
 
 class Priority(IntEnum):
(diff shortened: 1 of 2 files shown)


Sahil-Chhoker:
@Arjun-Sasi thanks for the PR and clarifying the issue, but this is not the optimal place to put this, can you add a markdown cell and add this as a warning in the tutorial file [here](https://github.com/mesa/mesa/blob/main/docs/tutorials/4_visualization_basic.ipynb?short_path=9bb8e00#L343).

Arjun-Sasi:
Hi @Sahil-Chhoker ,Thank you for the feedback.I have now,

 Added a Markdown warning cell and Updated tutorial examples (e.g., MoneyModel(...)) to use keyword arguments consistently, ensuring alignment with SolaraViz requirements.

Sahil-Chhoker:
Thanks for making changes @Arjun-Sasi, looking at the ReadtheDocs build, this graph is missing:
image

and the graph you build using updated model gives this graph:
image

can you check what's wrong?

Arjun-Sasi:
Hi @Sahil-Chhoker ,Changes made in Mark Down and example to ensure that users include All parameters when converting from positional to keyword arguments.

PR #3045, opened 2025-12-30
Fix network visualization bug: Replace array indexing with dictionary lookup

### LINKED ISSUE #3023: Network visualization silently drops agents on non-contiguous nodes due to IndexError suppression
**Describe the bug**
Agents placed on network nodes with non-contiguous IDs (e.g., nodes [0, 1, 5, 10, 15]) are silently dropped from visualization without any error message. The bug occurs in `mesa/visualization/space_renderer.py:168-172` where `IndexError` is suppressed using `contextlib.suppress()`

**Expected behavior**
All agents should be visible in visualization at their correct positions regardless of node ID values. If position mapping fails, users should receive a clear warning message.

**To Reproduce**

1. Create a NetworkGrid with non-contiguous node IDs
2. Try to visualize the network
3. Observe that agents on nodes 5, 10, and 15 are missing from visualization
4. No error message or warning is shown

```python
import networkx as nx
from mesa import Agent, Model
from mesa.space import NetworkGrid
from mesa.visualization import SpaceRenderer

G = nx.Graph()
G.add_nodes_from([0, 1, 5, 10, 15])  # Non-contiguous IDs
G.add_edges_from([(0, 1), (1, 5), (5, 10), (10, 15)])

model = Model()
space = NetworkGrid(G)
model.space = space

# Place agents on each node
for node_id in [0, 1, 5, 10, 15]:
    agent = Agent(model)
    space.place_agent(agent, node_id)

# Create visualization
renderer = SpaceRenderer(model, space, agent_portrayal=lambda a: {"color": "red"})
```

**Additional context**

- Affects any NetworkGrid or Network space visualization
- Common with real-world networks that have arbitrary node IDs
- Also affects dynamic networks where nodes are added/removed during simulation
- Root cause: Using node IDs as array indices instead of dictionary keys

codebyNJ:
Fixes #3023. @Sahil-Chhoker This is a new PR I have raised a cleaner version.

**Problem:** Code used node IDs as array indices instead of dictionary keys, causing IndexError that was silently suppressed.

**Solution:** Replace array indexing with dictionary lookup with proper error handling.

**Before:** Nodes [0,1,5,10,15] → 0/5 agents visible (all lost)
**After:**  Nodes [0,1,5,10,15] → 5/5 agents visible 

**Tests:** 6 comprehensive unit tests added covering:
- Non-contiguous node IDs
- Missing node positions with warnings
- Large and negative node IDs
- Edge cases (empty, single node, duplicates)

**Files changed:**
- `mesa/visualization/space_renderer.py` - Fix implementation
- `tests/visualization/test_space_renderer.py` - Unit tests

**To Reproduce**

```python
import networkx as nx
from mesa import Agent, Model
from mesa.space import NetworkGrid
from mesa.visualization import SpaceRenderer
import numpy as np
import contextlib

# Create network with non-contiguous node IDs
G = nx.Graph()
G.add_nodes_from([0, 1, 5, 10, 15])
G.add_edges_from([(0, 1), (1, 5), (5, 10), (10, 15)])

model = Model()
space = NetworkGrid(G)
model.grid = space

for node_id in [0, 1, 5, 10, 15]:
    agent = Agent(model)
    space.place_agent(agent, node_id)

renderer = SpaceRenderer(model)

# Demonstrate the actual bug - what old code did
print("BEFORE FIX (old code with array indexing):")

pos_array = np.array([[0.1, 0.2], [0.3, 0.4], [0.5, 0.6], [0.7, 0.8], [0.9, 1.0]])
node_ids = np.array([[0], [1], [5], [10], [15]], dtype=float)

mapped_positions_old = []
for node_id in node_ids[:, 0].astype(int):
    with contextlib.suppress(IndexError):  # This is what the old code did!
        pos = pos_array[node_id]
        mapped_positions_old.append(pos)

print(f"Agents mapped: {len(mapped_positions_old)}/5")
print(f"Agents silently dropped: {5 - len(mapped_positions_old)}/5")
print(f"Which nodes were dropped: {[int(n) for n in node_ids[:, 0] if int(n) >= len(pos_array)]}")

# Demonstrate the fix - what new code does
print("\nAFTER FIX (new code with dictionary lookup):")

pos_dict = {
    0: np.array([0.1, 0.2]),
    1: np.array([0.3, 0.4]),
    5: np.array([0.5, 0.6]),
    10: np.array([0.7, 0.8]),
    15: np.array([0.9, 1.0])
}

mapped_positions_new = []
for node_id in node_ids[:, 0].astype(int):
    if node_id in pos_dict:  # Dictionary lookup - works for any node ID
        mapped_positions_new.append(pos_dict[node_id])

print(f"Agents mapped: {len(mapped_positions_new)}/5")
print(f"All agents visible: {len(mapped_positions_new) == 5}")
```

**Explaination**

Networks with non-contiguous node IDs (like `[0, 1, 5, 10, 15]` ) caused the old code to silently fail when using array indexing. By replacing array indexing with dictionary lookup, ensures all agents are properly visualized regardless of node ID values.


Sahil-Chhoker:
@codebyNJ Thanks for the PR. After reviewing the problem again, it appears we have two distinct issues: the Node ID mapping error (which you identified) and the layout synchronization race condition (identified by @codebreaker32).

While this PR fixes the mapping logic, we also need the graceful failure handling from @codebreaker32's approach to prevent crashes during frame delays. I’d like you two to collaborate on a hybrid solution in this PR. Ideally, we should combine the accuracy of dictionary lookups with `NaN` masking for missing nodes. Please update the tests to cover both scenarios efficiently.

codebyNJ:
@Sahil-Chhoker 

Hybrid fix combining approaches from #3045 and #3065

This PR addresses two distinct issues:

Non-contiguous node ID mapping - Using dictionary lookup instead of array indexing
Race condition resilience - Using NaN masking for missing nodes (from @codebreaker32's approach in #3065)
Missing agents are now hidden (NaN) rather than placed at origin, which is safer and prevents visual artifacts.

The below is the benchmarks after the hybrid fix
image

tpike3:
@Sahil-Chhoker I think the hybrid suggestion and collaboration is genius!

I will throw out a thought that anyone can feel free to dismiss, but in the hybrid vein, instead of a `for loop` (Line 171), what if you vectorized the operation (the mask approach of #3065)? I think this should further improve performance.

codebyNJ:
@tpike3 I tried out your idea locally and also benchmarked by implementing vectorization using `np.unique()` with boolean masking. 

If you want me to make the change I am ready to make it. 

### Vectorization Results 

The vectorized approach (using `np.unique()` with boolean masking) is **consistently faster**:

| Scenario | Loop | Vectorized | Improvement |
|----------|------|-----------|-------------|
| Small networks (10 nodes) | 112 µs | 59 µs | **-48%** |
| Non-contiguous nodes | 175 µs | 62 µs | **-65%** |
| Large networks | 787 µs | 748 µs | **-5%** |
| Race conditions | 153 µs | 31 µs | **-80%** |

image

Sahil-Chhoker:
@codebyNJ thanks for benchmarking both approaches, can you implement the vectorized approach if it doesn't introduce the discrete `node_id` problem again.

codebyNJ:
@Sahil-Chhoker I have updated with the vectorization technique. No it does not introduce the discrete `node_id` problem.

codebyNJ:
@Sahil-Chhoker I have updated with vectorization technique. It does not introduce discrete `node_id` issue again.

mesa/visualization/space_renderer.py
```diff
@@ -153,23 +152,38 @@ def _map_coordinates(self, arguments):
...
+            # Warn if significant nodes missing (likely layout issue, not race condition)
+            if missing_nodes and len(missing_nodes) > len(pos_dict) / 10:
+                sample = missing_nodes[: min(5, len(missing_nodes))]
+                warnings.warn(
+                    f"Many nodes {sample}{'...' if len(missing_nodes) > 5 else ''} not found "
+                    f"in position layout ({len(missing_nodes)}/{len(node_ids)} agents). "
+                    f"This may indicate the network layout needs to be updated or regenerated.",
+                    UserWarning,
+                    stacklevel=2,
+                )
```
Sahil-Chhoker, line 182: Did you run into such a problem while testing?

Sahil-Chhoker (approved):
Thanks @codebyNJ for this PR, this is approved from my side but it can take a look from @quaquel.

mesa/visualization/space_renderer.py
codebyNJ: I didn't encounter this during testing of Issue #3023 specifically, but I added it as defensive handling since we're now using dictionary lookup. Without the warning, missing nodes would silently become `NaN` (invisible) with no indication to the user.

If you think this warning is unnecessary, I can remove it and just let missing nodes silently become `NaN`. Let me know your preference!

PR #3048, opened 2025-12-31
Fix "Ghost Agents" in ContinuousSpace: Allow neighbors at same position

### LINKED ISSUE #3047: ContinuousSpace "Ghost Agents" Bug-agents at same position become invisible
In `ContinuousSpace.get_neighbors()`, agents at the **exact same position** are invisible to each other when using `include_center=False`. The code uses `distance == 0` as a proxy for "same object" when it actually means "same location."

**Buggy code** (`mesa/space.py:1431`):
```python
neighbors = [
    self._index_to_agent[x] for x in idxs if include_center or dists[x] > 0
]
```

## To Reproduce

```python
from mesa.space import ContinuousSpace
from mesa.agent import Agent
from mesa.model import Model

class TestAgent(Agent):
        super().__init__(model)
        self.unique_id = unique_id

model = Model()
space = ContinuousSpace(100, 100, False)

agent_a = TestAgent(1, model)
agent_b = TestAgent(2, model)

# Both agents at same position
pos = (50.0, 50.0)
space.place_agent(agent_a, pos)
space.place_agent(agent_b, pos)

neighbors = space.get_neighbors(pos, radius=10.0, include_center=False)
print(f"Neighbors: {len(neighbors)}")  # Prints: 0 (BUG - should be 1)
```

**Expected:** Find 1 neighbor (Agent B)  
**Actual:** Find 0 neighbors (Agent B is invisible)

## Impact

This **silently breaks** models that allow overlapping agents:
- Collision detection (can't detect overlaps)
-  Disease transmission (misses contact at dist=0)
-  Mating/reproduction (can't find partners at same location)
-  Crowding analysis (agents at same point see 0 neighbors)

## Root Cause

The code assumes: `distance == 0` ⟹ `same object`  
Reality: `distance == 0` ⟹ `same location` (different objects can overlap)

## Proposed Solution

Add optional `agent` parameter to filter by identity instead of distance:

```python
def get_neighbors(
    self, pos, radius, include_center=True, agent=None  # NEW
):
    # ... existing code ...
    
    if agent is not None:
        # Filter by identity
        neighbors = [
            self._index_to_agent[x] for x in idxs 
            if include_center or self._index_to_agent[x] is not agent
        ]
    else:
        # Legacy behavior (backward compatible)
        neighbors = [
            self._index_to_agent[x] for x in idxs 
            if include_center or dists[x] > 0
        ]
```

Nithin9585:
Closes #3047

In `ContinuousSpace.get_neighbors()`, if you set `include_center=False`, the code currently filters out **all agents at distance 0**. 

This means if two agents are at the **exact same position** (e.g. `(50, 50)`), they become invisible to each other.
- **Expected:** Agent A sees Agent B (distance 0).
- **Actual:** Agent A sees nobody (because the code assumes `distance 0` = `self`).

This breaks models relying on collision detection, contact transmission, or mating at the same location.

##  The Solution
I updated `get_neighbors` to accept an optional `agent` parameter.

```python
def get_neighbors(self, pos, radius, include_center=True, agent=None):
```

1. **If YOU PASS the agent:** The code filters by **identity**. It explicitly removes *that specific agent* from the result, but keeps everyone else (even if they are at the same position).
   * *Result:* "Ghost agents" are now visible.

2. **If YOU DON'T pass the agent (Legacy):** The code keeps the old behavior (filtering by distance > 0) to ensure backward compatibility for existing models.

##  Verification
I added 4 comprehensive tests to `tests/test_space.py`:
1. `test_overlapping_agents_detected_with_agent_parameter`: Verifies two agents at `(10,10)` can see each other.
2. `test_overlapping_agents_legacy_behavior`: Verifies old behavior is preserved when no agent is passed.
3. `test_multiple_overlapping_agents`: Verifies multiple agents at the same spot are handled correctly.
4. `test_partial_overlap_scenario`: Verifies a mix of overlapping and nearby agents.

All existing tests passed (`pytest tests/test_space.py`).

##  API Changes
- **Modified:** `ContinuousSpace.get_neighbors` signature changed to `(..., agent=None)`.
- **Non-breaking:** The parameter is optional, so existing code will run exactly as before.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@ ci:
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.14.7
+  rev: v0.14.10
   hooks:
     # Run the linter.
     - id: ruff
--- a/docs/tutorials/4_visualization_basic.ipynb
+++ b/docs/tutorials/4_visualization_basic.ipynb
@@ -6,6 +6,7 @@
    "source": [
     "# Visualization - Basic Dashboard\n",
     "\n",
+    "\n",
     "### The Boltzmann Wealth Model "
    ]
   },
@@ -60,6 +61,14 @@
     "\n",
     "import mesa\n",
     "from mesa.discrete_space import CellAgent, OrthogonalMooreGrid\n",
+    "\n",
+    "# Check Mesa version for visualization compatibility\n",
+    "if mesa.__version__.startswith((\"3.0\", \"3.1\", \"3.2\")):\n",
+    "    print(\n",
+    "        f\"⚠️  Mesa {mesa.__version__} detected. Visualization features require Mesa 3.3+\"\n",
+    "    )\n",
+    "    print(\"To upgrade: pip install --upgrade mesa\")\n",
+    "\n",
     "from mesa.visualization import SolaraViz, SpaceRenderer, make_plot_component\n",
     "from mesa.visualization.components import AgentPortrayalStyle"
    ]
--- a/docs/tutorials/5_visualization_dynamic_agents.ipynb
+++ b/docs/tutorials/5_visualization_dynamic_agents.ipynb
@@ -6,6 +6,7 @@
    "source": [
     "# Visualization - Dynamic Agents\n",
     "\n",
+    "\n",
     "### The Boltzmann Wealth Model "
    ]
   },
@@ -60,6 +61,14 @@
     "\n",
     "import mesa\n",
     "from mesa.discrete_space import CellAgent, OrthogonalMooreGrid\n",
+    "\n",
+    "# Check Mesa version for visualization compatibility\n",
+    "if mesa.__version__.startswith((\"3.0\", \"3.1\", \"3.2\")):\n",
+    "    print(\n",
+    "        f\"⚠️  Mesa {mesa.__version__} detected. Visualization features require Mesa 3.3+\"\n",
+    "    )\n",
+    "    print(\"To upgrade: pip install --upgrade mesa\")\n",
+    "\n",
     "from mesa.visualization import SolaraViz, SpaceRenderer, make_plot_component\n",
     "from mesa.visualization.components import AgentPortrayalStyle"
    ]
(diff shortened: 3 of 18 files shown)


quaquel:
Thanks for this PR. A couple of remarks

1. `mesa.space` is likely to be deprecated in Mesa 4.0. So I am hesitant to add new functionality to it
2. What you are actually saying with this pr is that include_center is conceptually wrong. This should always have been an agent instance instead of just a boolean. You are adding a keyword argument to work around this. 
3. The bug this solves is highly theoretical. It is rare for two agents to occupy the exact same float position in a continuous space. 

To be clear: none of these remarks invalidate this PR.

Nithin9585:
@quaquel 

The change is minimal and backward-compatible, so it shouldn't add tech debt for the 4.0 transition. Happy to check if this issue affects the new space implementation too.

**Re: include_center design**  
You're right this is a workaround for a conceptual flaw. `include_center` conflates identity ("exclude self") with distance ("exclude distance 0"). If you'd prefer, I can add a deprecation warning to guide users toward the `agent` parameter.

**Re: theoretical bug**  
While accidental position overlap is rare, there are real use cases where it's intentional:
- Contact-based transmission models (disease spread when agents occupy same location)
- Collision detection (physics simulations)
- Mating/grouping behaviors (agents converging to identical coordinates)

For these models, this bug blocks core functionality agents literally can't detect each other when they need to.

quaquel:
1. I like the idea of a deprecation warning and pointing users to the new keyword argument. At least within Mesa 3.x, that will remain valid. 
2. The problem does not exist in the new continuous spaces. Long story short: the new spaces are designed around an agent-centric API. So, `ContinuousSpaceAgent.get_neighbors_in_radius` removes self, while `ContinuousSpace.get_agents_in_radius` works from points and just returns a list of agents and their distances. It's up to the caller to do any post-processing, which is exactly how `ContinuousSpaceAgent.get_neighbors_in_radius` works. 
3. I see your point on whether the bug is theoretical or not, but with the new and faster continuous space, there is a superior alternative in place already.

quaquel:
Given the maintenance only status of mesa.space, @EwoutH, and I decided to close this PR.

PR #3053, opened 2026-01-02
test: ensure DataCollector raises ValueError when collecting data for…

disgruntled-penguin:
… non-Agent types

This PR adds a test case to `tests/test_datacollector.py` to ensure the `DataCollector` raises a `ValueError` when `collect` is called with an object that is not a Mesa Agent

- Added `test_collect_with_non_agent` to verify that passing a non-agent object (like a string or integer) to `DataCollector.collect` raises a `ValueError`
- 
- Fixes implicit behavior where non-agent objects might be processed silently or cause confusing errors later

### DIFF
--- a/mesa/batchrunner.py
+++ b/mesa/batchrunner.py
@@ -28,6 +28,7 @@ put the code inside an ``if __name__ == '__main__':`` code black as shown below:
 
 """
 
+import bisect
 import inspect
 import itertools
 import multiprocessing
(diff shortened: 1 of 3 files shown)


tests/test_datacollector.py
```diff
@@ -303,6 +303,20 @@ class NonExistentAgent(Agent):
...
             self.assertTrue(non_existent_data.empty)
 
+    def test_invalid_agent_type_error(self):
+        """Test that passing a non-Agent class raises ValueError during collection."""
+        data_collector = self.model.datacollector
+
+        class NotAnAgent:
+            pass
+
+        data_collector._new_agenttype_reporter(NotAnAgent, "foo", lambda a: 1)
```
quaquel: why are you messing with a private method here?

tests/test_datacollector.py
```diff
@@ -303,6 +303,20 @@ class NonExistentAgent(Agent):
...
+
+        with self.assertRaises(ValueError) as cm:
```
quaquel, line 314: You test here at the highest level (i.e., data_collector.collect). This makes it difficult for me to understand exactly where the value error is coming from. Might it be possible to make this test more specific (e.g., use `data_collector._record_agents).

Also, is the updated validation not able to detect this problem in the first place?

To clarify: I did not implement the data collector and have never used it myself outside of teaching, so I am relying on the source code to try to understand what this PR is adding.

tests/test_datacollector.py
disgruntled-penguin: ooh, I wasn't thinking about it being a private method, is it okay if i ill update it rn?

tests/test_datacollector.py
disgruntled-penguin: oh, I did not see this before, apologies
The value error is raised in `_record_agenttype()` when it checks if the agent_type is a valid agent subclass.

I've alreadyjust  updated this test in my latest commit to use the DataCollector constructor instead of calling `_new_agenttype_reporter()`  The test now creates a DataCollector with the invalid agent type and verifies that `collect()` raises the error.

I think about validation, though, currently `DataCollector.__init__` doesn't validate agenttype_reporters, it just stores them. The check only happens later during `collect()`.  Should I update `DataCollector` to strictly validate these types during `init` in this pr?

quaquel: I wouldn't bother with that. For Mesa 4.0, we are likely to revamp data collection anyway. If you can just make the test a bit more focussed (next to the other appreciated change), I think this is good to go.

disgruntled-penguin: got it, thank you! just to clarify, ill be using _record_agenttype() in place of collect()

PR #3056, opened 2026-01-02
feat: Add SignalType enum for type-safe signal definitions.

codebyNJ:
`mesa_signals` uses plain strings for signal types. Typos are only caught at runtime, there is no IDE autocomplete or discoverability (FIXME in source).

**Solution**
Add` SignalType(str, Enum) (e.g., SignalType.CHANGE)` and use it internally while preserving full backward compatibility with string signal names.

**Changes**
- `mesa_signal.py` - added SignalType, updated usages
- `__init__.py` - exported SignalType

**To Reproduce**
```python
from mesa.experimental.mesa_signals import Observable

class M:
    val = Observable()

m = M()

# Typo in signal type — currently only caught at runtime
try:
    m.observe("val", "chang", lambda s: print("changed"))
except ValueError as e:
    print("Runtime error:", e)

# Preferred (type-safe) usage after this change:
from mesa.experimental.mesa_signals import SignalType
m.observe("val", SignalType.CHANGE, lambda s: print("changed"))
m.val = 10  # triggers handler
```

**Testing**
- All mesa_signals tests pass.
- Ruff linting passed.
- Benchmarks show no regressions.
- Backward compatible: SignalType.CHANGE == "change" and existing string usage continues to work.

### DIFF
--- a/docs/migration_guide.md
+++ b/docs/migration_guide.md
@@ -3,7 +3,29 @@ This guide contains breaking changes between major Mesa versions and how to reso
 
 Non-breaking changes aren't included, for those see our [Release history](https://github.com/mesa/mesa/releases).
 
-## Mesa 3.4.0
+## Mesa 3.5
+### model steps deprecated
+`model.steps` is deprecated and will be removed in Mesa 4.0. Use `model.time` instead.
+
+Previously, time in Mesa was fragmented: simple models used `model.steps` as a proxy for time, while discrete event simulations stored time separately in `simulator.time`.
+`model.time` is now the single, canonical source of simulation time in Mesa.
+
+For most use cases, simply replace `model.steps` with `model.time`. Since `model.time` is a float that increments by 1.0 per step by default, the numerical values will be equivalent.
+
+```python
+# Old
+current = model.steps
+
+# New
+current = model.time
+
+# If you need an integer step count, cast to int
+current = int(model.time)
+```
+
+- Ref: [PR #3136](https://github.com/projectmesa/mesa/pull/3136), [PR #2903](https://github.com/projectmesa/mesa/pull/2903)
+
+## Mesa 3.4
 
 ### batch run
 `batch_run` has been updated to offer explicit control over the random seeds that are used to run multiple replications of a given experiment. For this a new keyword argument, `rng` has been added and `iterations` will issue a `DeprecationWarning`. The new `rng` keyword argument takes a valid value for seeding or a list of valid values. If you want to run multiple iterations/replications of a given experiment, you need to pass the required seeds explicitly.
@@ -29,7 +51,7 @@ results = mesa.batch_run(
 )
 ```
 
-## Mesa 3.3.0
+## Mesa 3.3
 
 Mesa 3.3.0 is a visualization upgrade introducing a new and improved API, full support for both `altair` and `matplotlib` backends, and resolving several recurring issues from previous versions.
 For full details on how to visualize your model, refer to the [Mesa Documentation](https://mesa.readthedocs.io/latest/tutorials/4_visualization_basic.html).
--- a/mesa/batchrunner.py
+++ b/mesa/batchrunner.py
@@ -202,7 +202,7 @@ def _model_run_func(
     run_id, iteration, kwargs = run
 
     model = model_cls(**kwargs)
-    while model.running and model.steps < max_steps:
+    while model.running and model._steps < max_steps:
         model.step()
 
     data = []
@@ -212,9 +212,9 @@ def _model_run_func(
         recorded_steps = model.datacollector._collection_steps
     except AttributeError:
         # Fallback for legacy models without _collection_steps
-        steps = list(range(0, model.steps, data_collection_period))
-        if not steps or steps[-1] != model.steps - 1:
-            steps.append(model.steps - 1)
+        steps = list(range(0, model._steps, data_collection_period))
+        if not steps or steps[-1] != model._steps - 1:
+            steps.append(model._steps - 1)
     else:
         match data_collection_period:
             case -1:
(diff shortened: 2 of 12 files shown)


quaquel:
I know there was an idea of using an enum in the fixme, but this can also be handled via `Literal[]` type hinting. What are the merits of both approaches and why would you argue that enums are better? Moreover, as indicated in the FIXME, there is a lot more going on here that might not be doable with a simple enum (hence the fixme comment).

codebyNJ:
@quaquel Here's my comparison of both approaches:

**Literal[] Pros**
- Zero runtime overhead
- Purely static, no new classes

**Literal[] Cons**
- No runtime validation (typos pass silently at runtime)
- No discoverability (`list(SignalType)` not possible)
- Limited IDE support compared to enums
- Cannot be extended by subclasses

**Enum Pros**
- Runtime validation catches errors
- Full IDE autocomplete and discoverability
- `SignalType(str, Enum)` has no overhead (inherits from str)
- Extensible: subclasses can add domain-specific signals
- Introspectable: `list(SignalType)` works

**Regarding the Broader FIXME**

> "Ideally you can define signal_types throughout the class hierarchy and they are just combined together"

This actually **favors enums** because:
1. Custom enums can inherit/extend `SignalType`
2. Subclasses like `ObservableList` could define:
   ```python
   class ListSignalType(SignalType):
       APPEND = "append"
       REMOVE = "remove"

quaquel:
> No runtime validation (typos pass silently at runtime)

I don't think this is correct. `HasObservables.observe` raises a value error if you try to subscribe to a signal type that is not being emitted by the observable. 

> Limited IDE support compared to enums

I mostly agree, although support for typing and, by extension, Literal is improving rapidly in most IDEs. 

> Extensible: subclasses can add domain-specific signals
> Introspectable: list(SignalType) works

I am not entirely sure how subclassing Enums works, so I would need an example or some time to investigate this myself. However, ideally, it should be trivial to obtain an overview of all observables and, for each observable, the signals it can emit.

codebyNJ:
@quaquel 

**Runtime validation:**
You're right - I should have been more precise. `HasObservables.observe` does validate at runtime, but only when you call `observe()`. The issue is that typos in signal names passed to `emit()` or when checking `if signal_type in ...` won't be caught unless there's an active observer. With enums, typos are caught immediately at parse time by the IDE and at runtime whenever the enum is referenced.

**Extensibility and introspection:**
For enum subclassing, here's an example:

```python
from enum import Enum

class BaseSignals(str, Enum):
    CHANGED = "changed"
    REMOVED = "removed"

class CellSignals(BaseSignals):
    AGENT_ADDED = "agent_added"
    AGENT_REMOVED = "agent_removed"

# Works with type checkers
def handle_signal(signal: CellSignals): ...

# Introspection
list(CellSignals)  # All signals including inherited ones
BaseSignals.CHANGED in CellSignals  # True

**Comment by quaquel** [comment]
Thanks. I'll try to take a quick look myself later today at your example as well as reacquaint myself with this code to see if enums indeed solve all issues I had in mind when I wrote the fixme.

**Comment by codebyNJ** [comment]
Sure got it.

**Comment by quaquel** [comment]
As a further exploration of this idea, could you also update observable_collections? Here, more signals exist, so it's a nice test to see if using enums works as intended.

**Comment by codebyNJ** [comment]
> As a further exploration of this idea, could you also update observable_collections? Here, more signals exist, so it's a nice test to see if using enums works as intended.

Sure I will update it.

**Comment by codebyNJ** [comment]
@quaquel I have updated the `observable_collections` also with enum.

File: mesa/experimental/mesa_signals/observable_collections.py [hunk 1]
```diff
@@ -13,22 +14,75 @@
...
 
 from .mesa_signal import BaseObservable, HasObservables
 
 __all__ = [
+    "ListSignalType",
     "ObservableList",
 ]
 
 
+class ListSignalType(str, Enum):
```
quaquel, line 28: why does this not extent SignalType?

mesa/experimental/mesa_signals/observable_collections.py
```diff
@@ -75,7 +129,9 @@ def __setitem__(self, index: int, value: Any) -> None:
         """
         old_value = self.data[index]
         self.data[index] = value
-        self.owner.notify(self.name, old_value, value, "replace", index=index)
+        self.owner.notify(
+            self.name, old_value, value, ListSignalType.REPLACE, index=index
```
quaquel, line 133: I do like the clean code due to the enum here. So yes, enums seem like the way to go.

mesa/experimental/mesa_signals/observable_collections.py
```diff
@@ -13,22 +14,75 @@
...
+
+
 class ObservableList(BaseObservable):
     """An ObservableList that emits signals on changes to the underlying list."""
 
     def __init__(self):
         """Initialize the ObservableList."""
         super().__init__()
-        self.signal_types: set = {"remove", "replace", "change", "insert", "append"}
+        self.signal_types: set = {
```
quaquel: I guess this should be simpler, because it's just all fields of the `ListSignalEnum`

mesa/experimental/mesa_signals/observable_collections.py
codebyNJ: Python Enums can’t be cleanly subclassed, so inheriting SignalType isn’t viable. I added ListSignalType as (str, Enum) to preserve runtime string-equality (e.g. ListSignalType.INSERT == "insert" == SignalType.INSERT), keep full backward compatibility, and get IDE/type-safety for list-specific signals.

mesa/experimental/mesa_signals/observable_collections.py
codebyNJ: I replaced the manual string set with an automatic set of enum members so it always reflects ListSignalType. Added a one-line comment explaining the choice.

mesa/experimental/mesa_signals/observable_collections.py
quaquel: Thanks for the clarification.

PR #3058, opened 2026-01-02
Fix: Correct data retrieval in batch_run when DataCollector.collect() called multiple times per step

### LINKED ISSUE #3057: [BUG] `batch_run` silently returns incorrect data
The `batch_run` utility in `mesa/batchrunner.py` relies on a critical but incorrect assumption: that the simulation step number `N` directly corresponds to the list index `N` in the `DataCollector`'s internal storage.

Specifically, in `mesa/batchrunner.py`:
```python
# The code retrieves data using the step number as a specific list index
model_data = {param: values[step] for param, values in dc.model_vars.items()}
```

If a model calls `datacollector.collect()` more frequently (e.g., twice per step for sub-step resolution) or less frequently (e.g., conditionally) than exactly once per step, this assumption breaks.

*   **Scenario A (Over-collection):** If `collect()` is called twice per step, Step 5 is at Index 10. `batch_run` requests `Index 5`, returning data effectively from Step 2.5. This results in **silent data corruption** where the time axis is completely decoupled from the data values.
*   **Scenario B (Sparse-collection):** If `collect()` is called every 10 steps, Step 100 is at Index 10. `batch_run` requests `Index 100`, causing an `IndexError` crash.

**Expected behavior**
`batch_run` should retrieve data corresponding to the specific **Model Step**, regardless of how many items are in the `DataCollector` list. 

If the user requests data for "Step 5", `batch_run` should look up the record where `step == 5`, rather than simply grabbing the 5th item in the list.

**To Reproduce**
Run the following script. It simulates a model that collects data twice per step (e.g., "start" and "end" of a step).

```python
import unittest
from mesa.model import Model

class CorruptModel(Model):
    def __init__(self, *args, **kwargs):
        self.schedule = None
        super().__init__()
            model_reporters={"RealStep": lambda m: m.steps}
        )
        # Collect INITIAL state

    def step(self):
        super().step()
        # Collect data TWICE per step to simulate sub-step resolution 
        # or separate 'move' and 'trade' phases

class TestBatchRunnerTimeDilation(unittest.TestCase):
    def test_time_dilation(self):
        # Run for 5 steps
        # Expected collections: Init(1) + 5 steps * 2 collections = 11 items.
        
        # BatchRunner will try to read 'Step 5'.
        # It naively grabs index 5.
        # Index 5 actually corresponds to the 2nd collection of Step 2 (Step 2b).
        
        results = batch_run(
            CorruptModel,
            parameters={},
            iterations=1,
            max_steps=5,
            data_collection_period=1, 
            display_progress=False
        )
        
        last_result = results[-1]
        reported_step = last_result["Step"] # BatchRunner says this is Step 5
        actual_step_data = last_result["RealStep"] # The data proves it is Step 2 or 3
        
        print(f"BatchRunner reports: Step {reported_step}")
        print(f"Actual data contains: Step {actual_step_data}")
        
        # This assert WILL FAIL
        assert reported_step == actual_step_data, \
            f"FATAL: BatchRunner returned data from Step {actual_step_data} when asked for Step {reported_step}"

if __name__ == '__main__':
    unittest.main()
```

**Output:**
```
BatchRunner reports: Step 5
Actual data contains: Step 3
FATAL: BatchRunner returned data from Step 3 when asked for Step 5
```

**Additional context**
*   **Severity:** High.
*   **Why it is dangerous:** In the "over-collection" scenario, the code does **not** crash. It produces a `DataFrame` that looks valid but contains scientifically invalid time-series data (e.g., epidemic curves will appear compressed or dilated).
*   **Proposed Fix:** The `DataCollector` structure might need to support key-based lookup (dictionary of steps), or `batch_run` needs to search the list for the correct step index rather than assuming `index == step`.

Nithin9585:
**fixes #3057**

## Problem
The `batch_run` utility incorrectly assumed that the step number equals the list index when retrieving data from `DataCollector`. This caused wrong data retrieval when `DataCollector.collect()` was called multiple times within a single model step (e.g., collecting data at initialization, mid-step, and end-of-step).

## Example Scenario
- Model runs for 5 steps
- `collect()` called 3 times per step (initialization, mid-step, end-of-step)
- Total collections: 15 entries in the data list
- When `batch_run` requested data for step 5, it accessed index 5 instead of indices 12-14
- **Result**: Retrieved data from step 2 instead of step 5

### Changes Made

**DataCollector** (`mesa/datacollection.py`):
- Added `_collection_steps` list to track which model time each collection occurred at
- Appends `model.time` to `_collection_steps` on every `collect()` call
- Uses `model.time` (instead of deprecated `model.steps`) throughout for consistency with #2903

**BatchRunner** (`mesa/batchrunner.py`):
- Modified `_collect_data()` to use binary search (`bisect.bisect_right`)
- Maps model time values to correct collection indices via `_collection_steps`
- Handles multiple collections per time step correctly

**Tests** (`tests/test_batch_run.py`):
- Added `test_batch_run_time_dilation` to verify the fix
- Tests scenario with 3 collections per step over 5 steps
- Validates correct data retrieval for specific time steps
- Uses `model.time` in all test model reporters for consistency

### DIFF
--- a/tests/test_datacollector.py
+++ b/tests/test_datacollector.py
@@ -2,6 +2,8 @@
 
 import unittest
 
+import pandas as pd
+
 from mesa import Agent, Model
 from mesa.datacollection import DataCollector
 
@@ -212,6 +214,21 @@ class TestDataCollector(unittest.TestCase):
 
         with self.assertRaises(Exception):
             data_collector.add_table_row("Final_Values", {"final_value": 10})
+
+    def test_table_ignore_missing(self):
+        """Test table collection with ignore_missing=True."""
+        data_collector = self.model.datacollector
+        # ["agent_id", "final_value"]
+
+        row = {"agent_id": 999}
+        data_collector.add_table_row("Final_Values", row, ignore_missing=True)
+
+        table_df = data_collector.get_table_dataframe("Final_Values")
+        # the last row should have 999 and None
+        last_row = table_df.iloc[-1]
+        self.assertEqual(last_row["agent_id"], 999)
+
+        self.assertTrue(pd.isna(last_row["final_value"]))
 
     def test_exports(self):
         """Test DataFrame exports."""


mesa/datacollection.py
```diff
@@ -355,6 +356,8 @@ def get_reports(agent):
     def collect(self, model):
         """Collect all the data for the given model object."""
         if self.model_reporters:
+            if hasattr(self, "_collection_steps"):
```
quaquel, line 359: As per #2903, it would be better to use model.time. `model.step` is likely to be deprecated in the near future.

mesa/batchrunner.py
```diff
@@ -253,7 +254,31 @@ def _collect_data(
         )
     dc = model.datacollector
 
-    model_data = {param: values[step] for param, values in dc.model_vars.items()}
+    if hasattr(dc, "_collection_steps"):
```
quaquel, line 257: What is going in in this if block? The pr description is about data collection, yet this changes the batchrunner.

mesa/datacollection.py
Nithin9585: Thank you for the feedback! You're absolutely right. I'll update the implementation to use `model.time` instead of `model.steps` . I'll push the updated code shortly.

mesa/batchrunner.py
Nithin9585: The fix requires changes in both DataCollector and BatchRunner because they work together to solve the bug.

**The Root Cause:**
The bug is in BatchRunner's [_collect_data()](cci:1://file:///c:/Users/Nithi/OneDrive/Pictures/Desktop/gsoc/mesa/mesa/batchrunner.py:245:0-284:38) function, specifically this line:
```python
model_data = {param: values[step] for param, values in dc.model_vars.items()}

Nithin9585:
@quaquel Thank you for the feedback! . I've updated the implementation to use `model.time` instead of `model.steps` throughout the fix to align with the deprecation plan from #2903.

**Changes made:**

**In [mesa/datacollection.py](cci:7://file:///c:/Users/Nithi/OneDrive/Pictures/Desktop/gsoc/mesa/mesa/datacollection.py:0:0-0:0):**
- `_collection_steps.append(model.time)` (line you commented on)
- [_record_agents()](cci:1://file:///c:/Users/Nithi/OneDrive/Pictures/Desktop/gsoc/mesa/mesa/datacollection.py:299:4-318:28): Changed agent prefix to use `agent.model.time`
- [_record_agenttype()](cci:1://file:///c:/Users/Nithi/OneDrive/Pictures/Desktop/gsoc/mesa/mesa/datacollection.py:320:4-353:32): Changed agent prefix to use `agent.model.time`  
- [collect()](cci:1://file:///c:/Users/Nithi/OneDrive/Pictures/Desktop/gsoc/mesa/mesa/datacollection.py:355:4-392:17): Changed dictionary keys to use `model.time` for `_agent_records` and `_agenttype_records`

**In [tests/test_batch_run.py](cci:7://file:///c:/Users/Nithi/OneDrive/Pictures/Desktop/gsoc/mesa/tests/test_batch_run.py:0:0-0:0):**
- Updated all test model reporters to use `m.time` instead of `m.steps`
- Updated conditional logic in test models to use `self.time`

This ensures consistency throughout the fix and avoids mixing deprecated (`model.steps`) and new (`model.time`) approaches.

All tests pass with these changes. Pushing the update now.

quaquel:
Can you update the PR description to reflect the shift to using model.time? Otherwise, this seems good to go.

Nithin9585:
@quaquel I've updated the PR description to reflect the shift to using `model.time` instead of `model.steps`.

PR #3065, opened 2026-01-04
Fix race around condition in space_renderer

### LINKED ISSUE #3064: Race condition in SpaceRenderer._map_coordinates for NetworkGrid
There is a race condition in `SpaceRenderer._map_coordinates `when rendering NetworkGrid environments. Currently, the method uses `contextlib.suppress(IndexError)` to handle cases where an agent's node index exceeds the bounds of the cached layout positions (pos).

The current logic is "all-or-nothing." If the simulation adds a new node/agent (e.g., node 100) before the visualization layout refreshes (which only knows nodes 0-99), the `IndexError` triggers. The suppress block catches this error but aborts the entire coordinate mapping assignment. 

### Expected behavior 

- The visualization should be robust against transient race conditions.

- Valid Agents: Agents with valid node indices should still be mapped and drawn correctly, even if other agents in the same batch are invalid.

- Invalid Agents: Agents with out-of-bounds indices (waiting for layout updates) should be safely ignored or hidden (e.g., set to NaN) for that single frame, rather than freezing the visualization or rendering garbage data for the entire batch. 

Performance: Benchmarking with 5,000 agents shows a negligible overhead of ~0.13ms per frame vs the old unsafe method.

Proposed Logic:

```Python

# Handle race condition 

# 1. Initialize result with NaNs (hidden from plot)
mapped_locs = np.full((len(x), 2), np.nan)

# 2. Identify valid indices that exist in the current layout
if len(pos) > 0:
    valid_mask = (x >= 0) & (x < len(pos))

    # 3. Map only valid agents
    if np.any(valid_mask):
        mapped_locs[valid_mask] = pos[x[valid_mask]]

mapped_arguments["loc"] = mapped_locs
```

codebreaker32:
### Summary
There is a race condition in `SpaceRenderer._map_coordinates `when rendering NetworkGrid environments. Currently, the method uses `contextlib.suppress(IndexError)` to handle cases where an agent's node index exceeds the bounds of the cached layout positions (pos).

The current logic is "all-or-nothing." If the simulation adds a new node/agent (e.g., node 100) before the visualization layout refreshes (which only knows nodes 0-99), the `IndexError` triggers. The suppress block catches this error but aborts the entire coordinate mapping assignment. 

### Bug / Issue
Fix #3064 

### Implementation
```Python

# Handle race condition 

# 1. Initialize result with NaNs (hidden from plot)
mapped_locs = np.full((len(x), 2), np.nan)

# 2. Identify valid indices that exist in the current layout
if len(pos) > 0:
    valid_mask = (x >= 0) & (x < len(pos))

    # 3. Map only valid agents
    if np.any(valid_mask):
        mapped_locs[valid_mask] = pos[x[valid_mask]]

mapped_arguments["loc"] = mapped_locs
```
### Testing
Benchmarking with 5,000 agents shows a negligible overhead of ~0.13ms per frame vs the old unsafe method.
If you're fixing the visualisation, add before/after screenshots.

### DIFF
--- a/.github/styles/Mesa/README.md
+++ b/.github/styles/Mesa/README.md
@@ -1,0 +1,77 @@
+# Vale Configuration for Mesa
+
+Vale is a syntax-aware prose linter that helps maintain consistent, high-quality documentation across Mesa's markdown and reStructuredText files.
+
+## How Vale is Configured for Mesa
+
+### Base Style: Google Developer Style Guide
+
+Mesa uses the Google style guide as its foundation because it prioritizes technical clarity and developer documentation over corporate/marketing tone. This aligns well with Mesa's technical, academic audience.
+
+### Configuration Files
+
+- **`.vale.ini`**: Main configuration at repository root
+- **`.github/styles/Mesa/mesa.yml`**: Custom Mesa-specific terminology rules (e.g., "Mesa" vs "mesa", "NumPy" vs "numpy")
+- **`.github/workflows/vale.yml`**: CI workflow that runs on documentation changes
+
+### Alert Levels
+
+Mesa uses a tiered approach:
+
+- **`suggestion`** (default): Google style recommendations are informational only, non-blocking
+- **`warning`**: Important issues shown in CI (via `--minAlertLevel=warning`)
+- **`error`**: Mesa branding rules that must be enforced (e.g., correct capitalization)
+
+### Disabled Google Rules
+
+Many Google rules are disabled for technical documentation:
+
+- **Acronyms, Passive Voice, Semicolons**: Common and acceptable in technical writing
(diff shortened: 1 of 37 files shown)


codebreaker32:
Please review @EwoutH @quaquel

Sahil-Chhoker:
I'm currently reviewing PR #3045 by @codebyNJ and that feels like the proper solution to this problem, can't say anything now but will let know when the review is over.

quaquel:
Can you take a look at the conflicts in space renderer and ensure they are correctly resolved? Once that is done, I'll give a quick last look before merging it.

codebreaker32:
This bug is resolved in #3045 through collaborative effort(see [here](https://github.com/mesa/mesa/pull/3045#issuecomment-3729598871). Thus closing this PR

PR #3069, opened 2026-01-04
Optimize Cell is_empty/is_full

codebyNJ:
### Summary
Optimize `Cell` to remove an avoidable **O(n)** copy and improve error semantics.  
This prevents unnecessary allocations on frequent `remove_agent()` calls and makes capacity errors catchable.

### Bug / Issue
- `remove_agent()` did `self.empty` = `self.is_empty` → `len(self.agents)` → `self._agents.copy()` O(n) every removal.

### Implementation
 **`cell.py`**
- `is_empty` ⇒ `return len(self._agents) == 0`
- `is_full` ⇒ `return len(self._agents) >= self.capacity`

### Testing
- Added a concise test in `test_discrete_space.py` verifying that no list replacement occurs and that `.agents` still returns a copy.
- Updated capacity-related tests to expect a `ValueError`.
- Unit: all discrete_space tests passed locally (72/72).

### DIFF
--- a/tests/discrete_space/test_discrete_space.py
+++ b/tests/discrete_space/test_discrete_space.py
@@ -598,12 +598,12 @@ def test_cell():
     cell1.add_agent(CellAgent(model))
     assert cell1.is_full
 
-    with pytest.raises(ValueError):
+    with pytest.raises(Exception):
         cell1.add_agent(CellAgent(model))
 
     # Test capacity=0 (no agents allowed)
     cell_zero = Cell((1,), capacity=0, random=random.Random())
-    with pytest.raises(ValueError):
+    with pytest.raises(Exception):
         cell_zero.add_agent(CellAgent(model))
 
 


mesa/discrete_space/cell.py
```diff
@@ -110,9 +110,7 @@ def add_agent(self, agent: CellAgent) -> None:
         self.empty = False
 
         if self.capacity is not None and n >= self.capacity:
-            raise Exception(
```
quaquel, line 113: This is an API change, so we cannot simply do this given semver.

There is an open disucssion on exceptions for Mesa 4.0, so until that is resolved, please don't change this.

codebyNJ: Ok got it. I will remove the exception changes and update the PR.

codebyNJ: The PR is updated. Kindly re-run the benchmark score.

Issue #3072, opened 2026-01-05
Make `cell.empty` read only


quaquel:
**What's the problem this feature will solve?**
Currently, cells have two ways of getting their full/empty state: `cell.empty` and `cell.is_empty`. The former is a view into the underlying `empty` property layer on the space. The latter is the actual check if self-agents are empty. 

It is easy to invalidate the property layer by setting `cell.empty` directly yourself, and since it is part of the public API, nothing will warn the user or prevent the user from doing so

**Describe the solution you'd like**
Ideally, we add a way to add read-only property layer descriptors, while still allowing the cell to privately change its value.

**Additional context**
See also #3068 for more background on this.

Nithin9585:
---

Good idea! I can implement this with:

1. New `ReadOnlyPropertyDescriptor` class (raises error on write)
2. Add `_space` reference to cells
3. Internal `_update_property()` method for managed properties
4. Mark `empty` layer as read-only

This prevents users from doing `cell.empty = False` while still allowing internal updates via `add_agent`/`remove_agent`. Reading remains backward compatible.

Happy to work on a PR if this approach works for you!

---

quaquel:
I am not sure that the sketched outline is what I prefer. Most notably, we deliberately made `Cell` unaware of the space, so adding `_space` runs counter to a key design choice. Also, I prefer to have a generic solution in place so you can mark any property layer as read-only.

codebreaker32:
I propose implementing a **read-only interface with a private backing field**. This effectively mimics standard Python encapsulation (where a public read-only property exposes a private writable attribute), adapted dynamically for the `PropertyLayer` system.

This solves the issue without introducing new classes, simply by extending the existing `PropertyDescriptor`.

### The Proposal

We configure the system so that the `Grid` attaches **two** descriptors to the `Cell` for the same data layer:

1. **Public Interface (`cell.empty`):** A **Read-Only** descriptor. This prevents users from manually invalidating the state.
2. **Private Backing (`cell._empty`):** A **Writable** descriptor. The `Cell` uses this internal attribute to manage its state.

### Implementation Details

**1. `mesa/discrete_space/property_layer.py`**
We modify `PropertyDescriptor` to accept a `read_only` flag and update `HasPropertyLayers` to wire up the private backing descriptor automatically.

```python
# Modified PropertyDescriptor
class PropertyDescriptor:
    def __init__(self, property_layer: PropertyLayer, read_only: bool = False):
        self.layer = property_layer
[...]

PR #3074, opened 2026-01-05
Refactor PropertyLayer to implement NumPy interface

codebreaker32:
### Summary

Refactors `PropertyLayer` to serve as a direct wrapper around the underlying NumPy array, enabling standard NumPy syntax (slicing, broadcasting and methods)

### Motive

The current `PropertyLayer` class acts as an overly complex wrapper that duplicates native NumPy functionality with custom methods like `set_cells` and `modify_cells`. This design:

* **Violates Pythonic principles:** It hides standard operations behind a slower, restrictive API ("explicit is better than implicit").
* **Hinders Performance:** Methods like `modify_cells` rely on `np.vectorize` for lambdas, which is significantly slower than native ufuncs and broadcasting.
* **Blocks Future Architecture:** As discussed regarding the [`model.world`](https://github.com/mesa/mesa/discussions/2585) vision, layers should ideally be flexible views on data. The current design tightly couples data to legacy access patterns, acting as an obstacle to the "Single Source of Truth" concept.

### Implementation

1. **Direct exposure of NumPy data**

* Replaced the internal attribute `_mesa_data` with a public attribute `data`
* `data` is now a **public NumPy array**, enabling slicing, masking, and aggregation without wrappers
* Eliminates unnecessary `np.copyto` overhead.

2. **More efficient in-place updates**

* `set_cells` and `modify_cells` now operate directly on `self.data`.
* Conditional updates use boolean masks instead of rebuilding arrays.

3. **Improved `modify_cells` semantics**

### DIFF
--- a/mesa/discrete_space/property_layer.py
+++ b/mesa/discrete_space/property_layer.py
@@ -26,6 +26,7 @@ from mesa.discrete_space import Cell
 Coordinate = Sequence[int]
 T = TypeVar("T", bound=Cell)
 
+
 class PropertyLayer:
     """A class representing a layer of properties in a two-dimensional grid.
 
@@ -131,12 +132,10 @@ class PropertyLayer:
             mask = condition(self.data)
             target_data = self.data[mask]
 
-        if(isinstance(operation,np.ufunc)):
-            if(ufunc_requires_additional_input(operation)):
+        if isinstance(operation, np.ufunc):
+            if ufunc_requires_additional_input(operation):
                 if value is None:
-                    raise ValueError(
-                        "This ufunc requires an additional input value."
-                    )
+                    raise ValueError("This ufunc requires an additional input value.")
                 self.data[mask] = operation(target_data, value)
             else:
                 self.data[mask] = operation(target_data)


quaquel:
Since you are now raising a deprecation warning, all tests are failing. So, the tests need to be updated in line with the new API. 

Performance gains look modest but are still very welcome.

codebreaker32:
I believe it is fit to review now @quaquel

mesa/discrete_space/property_layer.py
```diff
@@ -88,40 +74,100 @@ def __init__(
...
         )
-        layer.set_cells(data)
+        layer.data = data
         return layer
 
+    # NumPy Array Interface
+
+    def __array__(self, dtype=None):
+        """Allow the layer to be passed directly to NumPy functions."""
+        return np.asarray(self.data, dtype=dtype)
```
quaquel: Is this "cast" of np.asarray really needed? 

Also, what happens is I pass a different dtype to the dtype of self.data?

mesa/discrete_space/property_layer.py
```diff
@@ -88,40 +74,100 @@ def __init__(
...
+    def __ifloordiv__(self, other):
+        """In-place floor division."""
+        self.data //= other
+        return self
+
+    def __ipow__(self, other):
+        """In-place power."""
+        self.data **= other
+        return self
+
```
quaquel: are all these methods needed for backward compatibility?

mesa/discrete_space/property_layer.py
```diff
@@ -88,40 +74,100 @@ def __init__(
...
             value: The value to be used for the update.
             condition: (Optional) A callable that returns a boolean array when applied to the data.
         """
+        warnings.warn(
+            "set_cells is deprecated and will be removed in a future version. "
+            "Use direct NumPy indexing on the layer object instead (e.g. layer[:] = value or layer[mask] = value).",
+            DeprecationWarning,
+            stacklevel=2,
+        )
         if condition is None:
```
quaquel: consider turning this into an annotation so you can simply annotate all depracated methods with `@deprecated`.

mesa/discrete_space/property_layer.py
```diff
@@ -131,37 +177,31 @@ def modify_cells(
...
-            # Vectorize non-ufunc operations
-            vectorized_operation = np.vectorize(operation)
-            modified_data = vectorized_operation(self._mesa_data)
+        warnings.warn(
+            "modify_cells is deprecated and will be removed in a future version. "
+            "Use direct NumPy operations on the layer object instead (e.g. layer += 1).",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        mask = condition(self.data) if condition is not None else slice(None)
```
quaquel, line 160: You now have not only deprecated this, but also written a new and perhaps faster implementation of `vectorized_operation`. Is this new implementation realy needed if we deprecate the method?

mesa/discrete_space/property_layer.py
```diff
@@ -170,34 +210,40 @@ def select_cells(self, condition: Callable, return_list=True):
...
 
 class HasPropertyLayers:
-    """Mixin-like class to add property layer functionality to Grids.
-
-    Property layers can be added to a grid using create_property_layer or add_property_layer. Once created, property
-    layers can be accessed as attributes if the name used for the layer is a valid python identifier.
+    """Mixin class to add property layer functionality to Grids.
 
+    Property layers can be added to a grid using create_property_layer or add_property_layer.
+    Once created, property layers can be accessed as attributes.
```
quaquel: why did you change this docstring? is the valid python identfier stuff not valid anymore?

mesa/discrete_space/property_layer.py
```diff
@@ -271,11 +314,17 @@ def set_property(
...
-        self._mesa_property_layers[property_name].set_cells(value, condition)
+        # Refactored to bypass deprecated set_cells
+        layer = self._mesa_property_layers[property_name]
+        if condition is None:
+            layer.data[:] = value
+        else:
+            mask = np.vectorize(condition)(layer.data)
+            layer.data[mask] = value
 
     def modify_properties(
```
quaquel, line 263: Can this not also be deprecated?

mesa/discrete_space/property_layer.py
```diff
@@ -406,24 +454,23 @@ def __getattr__(self, name: str) -> Any:  # noqa: D105
...
+                f"'{type(self).__name__}' object already has a property layer with name '{key}'"
+            )
         else:
-            if key in layers:
-                raise AttributeError(
-                    f"'{type(self).__name__}' object already has a property layer with name '{key}'"
-                )
-            else:
-                super().__setattr__(key, value)
+            super().__setattr__(key, value)
```
quaquel: these changes don't seem to be needed, why did you make them?

When reviewing PRs, most maintainers heavily rely on the visual diff. Adding changes that are not needed for the PR (including also editorial or visual (e.g., capitalization, punctuation) docstring changes) makes reviewing more difficult than it needs to be.

mesa/discrete_space/property_layer.py
codebreaker32: Yes, np.asarray is the standard implementation for the __array__ protocol. It ensures that passing the layer to np.array() works as expected without unnecessary copying if the types match.

If a different dtype is passed (e.g. np.array(layer, dtype=int)), np.asarray handles the casting automatically and returns a new array with that type.

mesa/discrete_space/property_layer.py
codebreaker32: No, they are actually the opposite! These methods aren't for backward compatibility; they implement the new features that allow the PropertyLayer to behave like a standard NumPy array.

They directly replace the bloated wrapper methods we are deprecating:

1. __iadd__ / __isub__  etc. replace modify_cells, enabling clean syntax like layer += 1.
2. __setitem__ replaces set_cells, enabling standard assignment like layer[0, 0] = 5.
3. __array__ replaces aggregate, allowing direct use of functions like np.mean(layer).

Without these dunder methods, users would be forced to access .data for every single operation.

mesa/discrete_space/property_layer.py
codebreaker32: The new implementation removes the private _mesa_data property wrapper that the old method relied on. Since the method had to be updated to use self.data, I implemented it using standard NumPy logic rather than duplicating the old, more complex code structure.

If you think we should stick to old implementation, I will revert this change

mesa/discrete_space/property_layer.py
codebreaker32: Sorry for the unrelated changes, I will keep it in mind for future PRs

mesa/discrete_space/property_layer.py
quaquel: ok, clear. You essentially went one step further than what I thought this PR would do. I would already have been happy with just doing `space.property_layer.data *= 0.5`. 

What is the performance difference between these operations and doing them via `.data`?

codebreaker32: The dunder methods (like __imul__) are just thin wrappers, so the only overhead is a single Python function call. For any meaningful grid size, the actual NumPy operation dominates the execution time.

codebreaker32:
Hi @quaquel 
I was working on it for past days so I made some unrelated changes unintentionally. Sorry for that
Now I've reverted them to clean up the diff

The PR is now back to just the essentials:
1. Added the new dunder methods to support the NumPy array interface.
2. Added deprecation warnings to the old methods (set_cells, etc.) without removing them.

Tests have been updated to handle the warnings.

mesa/discrete_space/property_layer.py
codebreaker32: I have reverted to the old code

tests/discrete_space/test_discrete_space.py
```diff
@@ -638,30 +638,6 @@ def test_cell_is_full_with_finite_capacity():
     assert cell.is_full is True
 
 
-def test_is_empty_no_list_copy():
```
quaquel, line 641: why is this test removed?

mesa/discrete_space/property_layer.py
quaquel: I am still curious, so I might test it myself tomorrow. 


Otherwise, this looks almost ready to go.

tests/discrete_space/test_discrete_space.py
codebreaker32: reverted to original

mesa/discrete_space/property_layer.py
quaquel: I have run some local tests. I am not sure that this is working as intended.

1. For many operations, doing it via layer.data is about 50% faster.
2. Stuff like model.layer *= 0.5 does not work
3. Stuff like model.layer[model.layer < 0.5] = 1 does not work.

I am wondering whether we should not simplify this PR by dropping the new magic methods and just use layer.data. A subsequent PR can then add the magic methods and ensure they all work properly.

codebreaker32: No problem but could you share the results or errors that you got, It will further help me to debug

codebreaker32:
Hi @quaquel I have now removed all the dunder(magic) methods and accordingly updated the PR description

mesa/discrete_space/property_layer.py
quaquel: So, I got a couple of different errors:

```python
from mesa.examples import BoltzmannWealth
model = BoltzmannWealth(n=90)
model.grid.create_property_layer("test")
model.grid.test *= 1
```

This raises an attribute error: `AttributeError: 'OrthogonalMooreGrid' object already has a property layer with name 'test'`.

```python
model.grid.test[model.grid.test<0.5]
```

This raises a type error `'<' not supported between instances of 'PropertyLayer' and 'float'`, probably because the relevant dunder methods for logical operations are missing.

There was another one that I don't recall, but it only triggered when using `%%timeit` in a notebook, making it even weirder.

quaquel:
How do you want to proceed? Just merge this for now or do you want to take a second stab at the dunder methods as part of this PR. I am fine either way.

codebreaker32:
I think we should merge this first so that it would be easier for anyone(or us) working on #3072

mesa/discrete_space/property_layer.py
```diff
@@ -139,27 +151,28 @@ def modify_cells(
...
         else:
             # Vectorize non-ufunc operations
             vectorized_operation = np.vectorize(operation)
-            modified_data = vectorized_operation(self._mesa_data)
+            modified_data = vectorized_operation(self.data)
 
-        self._mesa_data = np.where(condition_array, modified_data, self._mesa_data)
+        self.data = np.where(condition_array, modified_data, self.data)
 
+    @deprecated("Use np.argwhere(condition(layer)) or boolean masks instead.")
```
quaquel:
```suggestion
    @deprecated("Use np.argwhere(condition(layer.data)) or boolean masks instead.")
```

mesa/discrete_space/property_layer.py
```diff
@@ -174,19 +187,22 @@ def select_cells(self, condition: Callable, return_list=True):
...
 
-        condition_array = condition(self._mesa_data)
+        condition_array = condition(self.data)
         if return_list:
             return list(zip(*np.where(condition_array)))
         else:
             return condition_array
 
+    @deprecated(
+        "Use NumPy aggregate functions directly on the layer object (e.g. np.mean(layer))."
```
quaquel:
```suggestion
        "Use NumPy aggregate functions directly on the layer object (e.g. np.mean(layer.data))."
```

mesa/discrete_space/property_layer.py
```diff
@@ -265,6 +281,9 @@ def remove_property_layer(self, property_name: str):
         delattr(self.cell_klass, property_name)
         self.cell_klass._mesa_properties.remove(property_name)
 
+    @deprecated(
+        "Use direct NumPy assignment on the property layer instead (e.g. grid.layer_name[:] = value)."
```
quaquel:
```suggestion
        "Use direct NumPy assignment on the property layer instead (e.g. grid.layer_name.data[:] = value)."
```

mesa/discrete_space/property_layer.py
```diff
@@ -277,6 +296,7 @@ def set_property(
         """
         self._mesa_property_layers[property_name].set_cells(value, condition)
 
+    @deprecated("Use direct NumPy operations on the property layer instead.")
```
quaquel:
```suggestion
    @deprecated("Use direct NumPy operations on the property layer data field instead.")
```

quaquel:
sounds good, some last comments to update the deprecation warning to reflect the current recommended usage.

quaquel:
Sorry, one more question/thought: we move `_mesa_data` to `data`. However, in pandas, the numpy array is always available via `DataFrame.values`, so might that not be a more intuitive name to use?

codebreaker32:
Sure, we can switch to that name

quaquel:
@EwoutH, what do you think? `PropertyLayer.data` or `PropertyLayer.values` as the name for getting the numpy array behind the property layer. 

I am fine either way, and we should not overthink it.

codebreaker32:
I did a quick search to check the conventions.

While Pandas uses `.values`, they are slowly moving away from it in favor of `.to_numpy()`. On the other hand, Xarray and NumPy Masked Arrays (which are arguably closer to our grid structure) both use .data to expose the underlying array.

I think sticking with `.data` makes more sense given those precedents.

Checkout : https://stackoverflow.com/questions/13187778/convert-pandas-dataframe-to-numpy-array#:~:text=Use%20df.,values%20%2C%20here's%20why.&text=It's%20time%20to%20deprecate%20your,Index%20and%20Series%20objects%20only.

EwoutH:
I think `.values` is slightly more intuitive. But both would work.

I while I really appreciate this effort, I would really like to put of making an (breaking) API change until we have figured out how PropertyLayers are going to fit in the bigger picture.

If some changes could be made without changing the API, like by making the implementations behind the current methods more elegant or NumPy native, that could already happen.

quaquel:
> I while I really appreciate this effort, I would really like to put of making an (breaking) API change until we have figured out how PropertyLayers are going to fit in the bigger picture.

Everything here is backward compatible, so there are no API-breaking changes; just a cleaner and faster alternative API.

EwoutH:
We're deprecating stuff, that's basically an announced API break. Just not enforced for now.

Also I have my doubts if we should already add another API. While direct nparray access is probably useful in some cases, we might also expose it through the spaces those propertylayers are attached to.

quaquel:
> Also I have my doubts if we should already add another API. While direct nparray access is probably useful in some cases, we might also expose it through the spaces those propertylayers are attached to.

This is exactly what this PR does: `space.property_layer.data`, while already working, with th is PR works much more cleanly and becomes the preferred method. That is: this PR makes doing `space.property_layer.data` faster by eliminating intermediate steps, and highlights to users that this faster approach is the preferred method going forward.

quaquel:
@codebreaker32, I had a longer in person conversation with @EwoutH. Given the ongoing conversations about spaces in Mesa 4, we decided not to deprecate anything yet. So, if you can remove those warnings, we think this should be good to go.

codebreaker32:
Thanks for the update, @quaquel.

I will go ahead and remove the deprecation warnings.

One question: Can we still optimize the existing functions (like `set_cells`, `modify_cells`)? I could rewrite them to use the direct `.data` operations internally, so they are faster and cleaner, even if we aren't officially deprecating them yet.

quaquel:
yes, feel free to rewrite them but then also profile the difference

EwoutH:
> I could rewrite them to use the direct `.data` operations internally, so they are faster and cleaner

Yes, this is fine and appreciated.

This also could be separate PR(s), that might make testing and reviewing easier.

codebreaker32:
### Summary of Changes

This update simplifies the `PropertyLayer` implementation, improves performance, and makes NumPy integration more explicit and predictable.

#### 1. **Direct exposure of NumPy data**

* Removed the `@property data` getter/setter indirection.
* `data` is now a **public NumPy array**, enabling slicing, masking, and aggregation without wrappers
* Eliminates unnecessary `np.copyto` overhead.

#### 2. **More efficient in-place updates**

* `set_cells` and `modify_cells` now operate directly on `self.data`.
* Conditional updates use boolean masks instead of rebuilding arrays.

#### 3. **Improved `modify_cells` semantics**

* Operations now update **only the selected cells**, not the whole grid.

codebreaker32:
Hi @quaquel @EwoutH,

I have updated the PR. I initially tried a simple rename from `.data` to `.values`, but it caused cascading failures. The reason is that the existing `data` attribute was actually a `@property` with specific setter logic (calling `set_cells`) that legacy code relies on.

I will change it in subsequent PRs

quaquel:
Can you update the original PR starting post to reflect the final version? I hope to review this later today, and if I don't see anything strange, merge it.

quaquel:
> I will change it in subsequent PRs

I am fine with leaving it as `.data` for now. @EwoutH and I discussed removing the `PropertyLayer` class in Mesa 4 in favour of directly using numpy arrays. Additionally, changing the name now would be a deprecation and a hassle we can avoid.

EwoutH:
@codebreaker32 Thanks for the PR. Could you check if the PR description is now fully up to date with the final implementation in this merged PR?

codebreaker32:
Yes, the PR description is now up to date

PR #3081, opened 2026-01-07
Version bump to 3.5.0-dev

quaquel:
version bump in init, which was not done after the release of 3.4

### DIFF
--- a/mesa/__init__.py
+++ b/mesa/__init__.py
@@ -24,7 +24,7 @@ __all__ = [
 ]
 
 __title__ = "mesa"
-__version__ = "3.4.0"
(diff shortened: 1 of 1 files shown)


EwoutH:
Version bump needs to happen indeed.

I think we'll be targeting 4.0.0 on `main`, right?

quaquel:
We can always go to 4.0 if we so decide. At least this makes testing easier for me because I can see whether I am testing against my installed version or my Git version. 

Also, with the many PRs that have been coming in we might do a new 3.x release anyway.

EwoutH (approved):
Fair, no harm in that.

PR #3087, opened 2026-01-07
Optimise select_random_empty_cell() in grid.py

codebreaker32:
### Summary

This PR optimizes the fallback mechanism in `Grid.select_random_empty_cell` to use vectorized NumPy operations, replacing the slow O(N) Python iteration. It also fixes a critical bug in `HasPropertyLayers.select_cells` preventing the correct masking of empty cells.

### Motive

**1. Performance on Dense Grids:**
The `select_random_empty_cell` method uses a fast random heuristic (O(1)). However, when the grid is nearly full (dense), this heuristic fails, and the code previously fell back to iterating over all cells:
`return self.random.choice(list(self.empties))`
This triggered the descriptor protocol for every single cell, causing a massive performance penalty. By switching to `self.select_cells(only_empty=True)`, we bypass the `Cell` objects entirely and scan the underlying NumPy array directly.

**2. Bug Fix:**
The `select_cells(only_empty=True)` method in `property_layer.py` was previously broken. It attempted to apply a logical mask to the `PropertyLayer` *object* instead of its underlying `.data` array, causing operations to fail or return incorrect results.( I missed it in #3074 )

### Implementation

1. **`mesa/discrete_space/grid.py`**:
* Updated `select_random_empty_cell` to call `self.select_cells(only_empty=True)` in the fallback block.

2. **`mesa/discrete_space/property_layer.py`**:
* Fixed `select_cells`: Changed `self._mesa_property_layers["empty"]` to `self._mesa_property_layers["empty"].data`.

### Usage Examples

There is no change to the public API.

### Additional Notes

* **Dependencies**: This relies on the `PropertyLayer` architecture being present.

### DIFF
--- a/mesa/discrete_space/property_layer.py
+++ b/mesa/discrete_space/property_layer.py
@@ -420,5 +420,5 @@ class PropertyDescriptor:
 
 def ufunc_requires_additional_input(ufunc):  # noqa: D103
     # NumPy ufuncs have a 'nin' attribute indicating the number of input arguments
-    # For binary ufuncs (like np.add), nin is 2
-    return ufunc.nin > 1
+    # For binary ufuncs (like np.add), nin is 2    # codespell:ignore
+    return ufunc.nin > 1  # codespell:ignore
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -91,7 +91,7 @@ homepage = "https://github.com/mesa/mesa"
 repository = "https://github.com/mesa/mesa"
 
 [tool.codespell]
-ignore-words-list = ["ist", "hart", "nD", "nin"]
+ignore-words-list = ["ist", "hart", "nD"]
 
 [tool.coverage.run]
 source = ["."]


mesa/discrete_space/grid.py
```diff
@@ -157,8 +157,9 @@ def select_random_empty_cell(self) -> T:  # noqa
                 if cell.is_empty:
                     return cell
 
-        # Fallback to the robust parent method (O(N)) if random sampling fails
-        return super().select_random_empty_cell()
+        empty_coords = self.select_cells(only_empty=True)
+        random_coord = self.random.choice(empty_coords)
+        return self._cells[random_coord]
```
quaquel: would it not be a lot faster to do something like


```python

coords = np.argwhere(self.empty.data)

```

codebreaker32: You are right.

quaquel: but if that works, I am wondering whether its not just allways faster to do this instead of the random trying. 

I know that there has been some research on this before (see. mesa.space for details). I am just not sure how those findings translate to the new discrete_space.

codebreaker32: I looked into the existing `mesa.space` implementation(which led me to https://github.com/mesa/mesa/issues/1052), and it actually uses a specific cutoff formula (cutoff_empties) to switch between random sampling and list generation. However, I believe numpy definitely shifts the **cutoff point** so we might reduce the **N**

quaquel: Yeah, my thinking exactly. I suppose some of the testing will need to be rerun to determine the cutoff in this case.

quaquel:
So this crashed:

```
  return self._cells[random_coord]
           ~~~~~~~~~~~^^^^^^^^^^^^^^
TypeError: cannot use 'numpy.ndarray' as a dict key (unhashable type: 'numpy.ndarray')
```

Basically, the numpy array for the coordinates needs to be cast to a tuple I guess. Also, it highlights the need for test coverage.

update: I fixed it for you after confirming that it was indeed working. Main thing now is to add test coverage.

mesa/discrete_space/grid.py
```diff
@@ -157,8 +159,9 @@ def select_random_empty_cell(self) -> T:  # noqa
                 if cell.is_empty:
                     return cell
 
-        # Fallback to the robust parent method (O(N)) if random sampling fails
-        return super().select_random_empty_cell()
+        empty_coords = np.argwhere(self.empty.data)
+        random_coord = self.random.choice(empty_coords)
+        return self._cells[random_coord]
```
```suggestion
        return self._cells[tuple(random_coord)]
```

codebreaker32:
I will add tests for it today only.

codebreaker32:
Hi @quaquel 
I also want to point out one minor inconsistency
`ufunc.nargs` counts both inputs and outputs (e.g., np.add has nargs=3) and `ufunc.nin` counts only input(np.add has nin=2) 
```python
def ufunc_requires_additional_input(ufunc):  # noqa: D103
    # NumPy ufuncs have a 'nargs' attribute indicating the number of input arguments
    # For binary ufuncs (like np.add), nargs is 2
    return ufunc.nargs > 1
```
Isn't it misleading then?

quaquel:
> Isn't it misleading then?

I guess so. I just copied this from`mesa.space`, but I guess it should be `ufunc.nin > 1`.

codebreaker32:
`ufunc.nin` resulted in a linting(codespell) error

```plaintext
ruff (legacy alias)......................................................Passed
ruff format..............................................................Passed
pyupgrade................................................................Passed
trim trailing whitespace.................................................Passed
check toml...............................................................Passed
check yaml...............................................................Passed
codespell................................................................Failed- hook id: codespell- exit code: 
65:mesa/discrete_space/property_layer.py:422: nin ==> inn, min, bin, nine
mesa/discrete_space/property_layer.py:423: nin ==> inn, min, bin, nine
```

quaquel:
> ufunc.nin resulted in a linting(codespell) error

What did you change exactly? I guess the problem triggers on the comment, not the actual code?

codebreaker32:
> What did you change exactly? I guess the problem triggers on the comment, not the actual code?

Both i.e method and comments. Yes the problem was triggered by comment

Update: Replaced `nargs` with `nin` and updated `pyproject.toml` to add "nin" as ignored-word

mesa/discrete_space/property_layer.py
```diff
@@ -419,6 +419,6 @@ def __set__(self, instance: Cell, value):  # noqa: D105
 
 
 def ufunc_requires_additional_input(ufunc):  # noqa: D103
-    # NumPy ufuncs have a 'nargs' attribute indicating the number of input arguments
-    # For binary ufuncs (like np.add), nargs is 2
-    return ufunc.nargs > 1
+    # NumPy ufuncs have a 'nin' attribute indicating the number of input arguments
+    # For binary ufuncs (like np.add), nin is 2
```
quaquel:
```suggestion
    # For binary ufuncs (like np.add), nin is 1
```

pyproject.toml
```diff
@@ -91,7 +91,7 @@ homepage = "https://github.com/mesa/mesa"
 repository = "https://github.com/mesa/mesa"
 
 [tool.codespell]
-ignore-words-list = ["ist", "hart", "nD"]
+ignore-words-list = ["ist", "hart", "nD", "nin"]
```
quaquel: if nin is between quotes in the comment, so \`nin\`, does this trigger codespell as well? If not, I would prefer solving it that way rather than adding it to the ignored words.

mesa/discrete_space/property_layer.py
codebreaker32: nin is 2 for np.add and 3 for nargs

pyproject.toml
codebreaker32: the error is caused by 
```python
    # For binary ufuncs (like np.add), nin is 2   
    return ufunc.nin > 1  
```
To fix:
```python
    # For binary ufuncs (like np.add), nin is 2    # codespell:ignore
    return ufunc.nin > 1  # codespell:ignore
```

Should I do it instead of changing `pyproject.toml`

codebreaker32:
I think its ready to be merged now.

PR #3088, opened 2026-01-07
Optimize `_Grid.move_to_empty` by removing unnecessary sort

Sonu0305:
### Summary
This change optimizes the `_Grid.move_to_empty` method in `mesa/space.py` by removing an unnecessary sorting operation, improving its performance.

### Motive
The previous implementation involved sorting the `self.empties` set before making a random choice. Since sets are inherently unordered, sorting them on every call is a computationally expensive and unnecessary operation, especially in large grids with many empty cells.

### Implementation
The implementation was modified to convert the `self.empties` set directly to a list instead of sorting it. This avoids the `O(n log n)` complexity of sorting in favor of the `O(n)` complexity of list conversion.

### Additional Notes
The selection of an empty cell remains random. The order of elements in a set is arbitrary, so sorting it before making a random choice does not make the choice any more or less random.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -566,7 +566,7 @@ class _Grid:
                 if self.is_cell_empty(new_pos):
                     break
         else:
-            new_pos = agent.random.choice(sorted(self.empties))
+            new_pos = agent.random.choice(list(self.empties))
         self.remove_agent(agent)
         self.place_agent(agent, new_pos)
 


quaquel:
The reason for sorting is to maintain the model's determinism. If you convert a set to a list, the same model, run several times, can produce different results.

Sonu0305:
> The reason for sorting is to maintain the model's determinism. If you convert a set to a list, the same model, run several times, can produce different results.

Perhaps, can we achieve the same deterministic outcome with better performance by changing how we select the element, rather than giving up on performance entirely?

quaquel:
I'm not aware of a method for that. So, you are welcome to come up with something and properly profile it. In the mean time, I am closing this PR.

PR #3095, opened 2026-01-09
Fix:Add agenttype_reporters support to batch_run

### LINKED ISSUE #2804: Collecting data from different types of agents (agenttype_reporters)
Hi,

I am not sure if I am misunderstanding the usage of `agenttype_reporters` parameter in `DataCollector()` function, but I cannot make it work. For example, I was trying to implement it in the Money Model (https://mesa.readthedocs.io/latest/tutorials/2_collecting_data.html) by replacing 

```python
    model_reporters={"Gini": compute_gini}, 
    agent_reporters={"Wealth": "wealth"}
)
```

with 

```python
    model_reporters={"Gini": compute_gini}, 
    agenttype_reporters={MoneyAgent: {'Wealth': 'wealth'}}
)
```

In this case no errors are popping up, but at the end there are no `Wealth` data in the results. Running using `batch_run()`:

```
params = {"width": 10, "height": 10, "n": range(5, 105, 5)}

results = mesa.batch_run(
    MoneyModel,
    parameters=params,
    iterations=5,
    max_steps=100,
    number_processes=1,
    data_collection_period=-1,
    display_progress=True,
)
results_df = pd.DataFrame(results)
```

Am I missing something? If I use both `agent_reporters` and `agenttype_reporters` at the same time, then only measures defined in `agent_reporters` are saved in the results. Mesa version 3.2.0

Thank you!

BhoomiAgrawal12:
### Summary
Fixes `agenttype_reporters` not collecting data when using `batch_run()`, making it impossible to use selective agent data collection with parameter sweeps and parallel execution.

### Bug / Issue
Closes #2804

Users reported that `agenttype_reporters` worked perfectly in regular model runs but  failed to collect data when using `batch_run()`. 

**What was expected:** Data from `agenttype_reporters` should appear in batch_run results, just like `agent_reporters` data does.

**What actually happened:** The agenttype data was silently dropped - no errors just missing columns in the results DataFrame.

### Implementation
The issue was in the `_collect_data()` helper function in `mesa/batchrunner.py`. This function extracts data from the DataCollector after model execution, but it only looked at `_agent_records` and completely ignored `_agenttype_records`.

**Changes made:**
1. Added iteration through `dc._agenttype_records` in `_collect_data()` 
2. Process agenttype data using the same structure as agent data
3. Combine both into the `all_agents_data` list for consistent output

It handles all edge cases:
- Models with only `agent_reporters`
- Models with only `agenttype_reporters` 
- Models with both reporter types
- Models with neither (model-only data)

### Testing
Added  test coverage in `tests/test_batch_run.py`:

1. **`test_batch_run_agenttype_reporters()`**: Verifies agenttype_reporters data is collected correctly
   - Tests with different parameter values (n_agents: [3, 5])
   - Validates correct number of rows (8 agents = 8 rows)
   - Confirms all expected columns are present

2. **`test_batch_run_agenttype_and_agent_reporters()`**: Tests mixed usage of both reporter types
   - Ensures both data types coexist in results
   - Verifies no data loss from either reporter type
   - Checks proper handling of duplicate agent entries

**Test Results:**
- All 20 batch_run tests pass (18 existing + 2 new)
- All datacollection tests pass (no regressions)

### DIFF
--- a/mesa/discrete_space/grid.py
+++ b/mesa/discrete_space/grid.py
@@ -19,6 +19,8 @@ from itertools import product
 from random import Random
 from typing import Any, TypeVar
 
+import numpy as np
+
 from mesa.discrete_space import Cell, DiscreteSpace
 from mesa.discrete_space.property_layer import (
     HasPropertyLayers,
@@ -157,8 +159,9 @@ class Grid(DiscreteSpace[T], HasPropertyLayers):
                 if cell.is_empty:
                     return cell
 
-        # Fallback to the robust parent method (O(N)) if random sampling fails
-        return super().select_random_empty_cell()
+        empty_coords = np.argwhere(self.empty.data)
+        random_coord = self.random.choice(empty_coords)
+        return self._cells[tuple(random_coord)]
 
     def _connect_single_cell_nd(self, cell: T, offsets: list[tuple[int, ...]]) -> None:
         coord = cell.coordinate
--- a/mesa/discrete_space/property_layer.py
+++ b/mesa/discrete_space/property_layer.py
@@ -343,7 +343,7 @@ class HasPropertyLayers:
         # Apply the empty mask if only_empty is True
         if only_empty:
             combined_mask = np.logical_and(
-                combined_mask, self._mesa_property_layers["empty"]
+                combined_mask, self._mesa_property_layers["empty"].data
             )
 
         # Apply conditions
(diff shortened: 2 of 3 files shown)


BhoomiAgrawal12:
Thanks @EwoutH for considering my contribution! I would like to work more on this project, it would be great if you can suggest me something that I can pick up. I would be happy to implement.

Issue #3046, opened 2025-12-30
Solara UI crashing while rendering Altair charts for certain Mesa Visualization examples


PrimalMentat15:
### **Describe the bug**

When using Mesa’s Solara-based visualization stack with Altair charts fails to render for some of the examples. The Solara UI fails to load entirely and causes a crash.

* So far i have only encountered this issue with only specific models, namely
1. [mesa/examples/basic/virus_on_network](https://github.com/mesa/mesa/tree/main/mesa/examples/basic/virus_on_network)
2. [mesa/examples/basic/boltzmann_wealth_model](https://github.com/mesa/mesa/tree/main/mesa/examples/basic/boltzmann_wealth_model)
3. [mesa/examples/advanced/pd_grid](https://github.com/mesa/mesa/tree/main/mesa/examples/advanced/pd_grid)
4. [mesa/examples/advanced/sugarscape_g1mt](https://github.com/mesa/mesa/tree/main/mesa/examples/advanced/sugarscape_g1mt)

---

### **Expected behavior**

Altair charts should render correctly in Solara-based Mesa visualizations when following the standard Mesa development setup, without running into UI crashes.

---

### **To Reproduce**

**Environment (example):**

* Python: 3.11
* Mesa: `main` branch (current)
* Altair: 6.0.0 (currently being installed implicitly via `pip install -e ".[dev]"`)
* Solara: 1.30.0 (currently being installed implicitly via `pip install -e ".[dev]")

**Steps:**

1. Clone the Mesa repository.
2. Install dependencies using:

   ```bash
   pip install -e ".[dev]"
   ```
3. Run any Mesa example model that uses Solara + Altair for visualization.
4. Launch the Solara frontend (`solara run ...` or via the provided example runner).
5. Observe that the Altair charts do not render and the UI crashes.

<img width="1864" height="887" alt="Image" src="https://github.com/user-attachments/assets/a4366a6d-f91d-4238-a9b2-408a095e3142" />

PrimalMentat15:
i request to be assigned to this issue as i would like to work on a solution for it

quaquel:
is this python 3.11 specific or does it also happens under 3.12?

PrimalMentat15:
> is this python 3.11 specific or does it also happens under 3.12?

[...]

PR #3098, opened 2026-01-09
fix altair rendering issue 

champ-byte:
### Summary

Rendering issue for mesa examples with altair backend due to the latest version altair==6 
### Bug / Issue

Error:
image

python version used: 3.13.5
altair version ==6

The error occurs because the software is looking for specific older versions of the Vega-Lite format (v4 or v5) within the data's MIME bundle, but only a newer version (v6) is present
 
### Implementation

Since the issue is due to widgetti/solara (mentioned here: [widgetti/solara/issues/1127]) and due to mesa , I think the simplest temporary fix is to downgrade the altair version <6 .

Another approach would be the whitelist or blacklist version and render an html iframe as a fallback , but again this won't be future proof and unnecessary hassle.
We cannot directly add a fallback by using try and catch(making it future-proof) since , Solara catches the error internally and displays it in the UI, so Python's try-catch can't intercept it. The error happens during the frontend rendering phase, not during Python execution.

### Additional Notes
If I am missing on any point or you would want me to follow a certain approach I would be happy to do that !

### DIFF
--- a/docs/apis/experimental.md
+++ b/docs/apis/experimental.md
@@ -37,3 +37,15 @@ This namespace contains experimental features. These are under development, and 
 .. automodule:: experimental.continuous_space.continuous_space_agents
    :members:
 ```
+
+## Scenarios
+
+```{eval-rst}
+.. automodule:: experimental.scenarios.scenario
+   :members:
+```
+
+```{eval-rst}
+.. automodule:: experimental.scenarios.model_with_scenario
+   :members:
+```
--- a/docs/migration_guide.md
+++ b/docs/migration_guide.md
@@ -3,7 +3,29 @@ This guide contains breaking changes between major Mesa versions and how to reso
 
 Non-breaking changes aren't included, for those see our [Release history](https://github.com/mesa/mesa/releases).
 
-## Mesa 3.4.0
+## Mesa 3.5
+### model steps deprecated
+`model.steps` is deprecated and will be removed in Mesa 4.0. Use `model.time` instead.
+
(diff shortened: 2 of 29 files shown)


champ-byte:
I found out that widgetti/solara currently required altair<6 but python 3.14 supports [altair==6 only](https://github.com/vega/altair/issues/3905) , hence I have made the neccessary changes in pyproject.toml

I have also made changes so that the code now detects the altair version and renders accordingly. If older altair is used then original older rendering is used else it uses the new VegaLite rendering method, this probably makes it future proof if solara doesn't fix the issue.

champ-byte:
Fixes #3046

champ-byte:
Hi , @quaquel is there anything I can help with in this regard ?

Sahil-Chhoker:
Thanks @champ-byte for this PR, can you include a image of how the newer `VegaLite` graphs look if they are any different from the legacy ones?

champ-byte:
> Thanks @champ-byte for this PR, can you include a image of how the newer `VegaLite` graphs look if they are any different from the legacy ones?

Hi, Thank you looking into the PR. The graphs look exactly the same but still attaching both the graphs for confirmation.

Legacy version:
image

Newer version:
image

EwoutH:
As of Solara 1.57.1 Altair 6 is supported (see https://github.com/widgetti/solara/issues/1127), does that have implications for this PR? Should we revert (part of) it?

Sahil-Chhoker:
The edits by this PR works but now are unnecessary due to the `solara` update, I think this PR can entirely be reverted as we don't need to pin the versions and there is no need to use the `VegaLite` spec for newer `altair` version.

PR #3103, opened 2026-01-09
Add explicit support for Scenarios

quaquel:
This PR adds explicit support for running scenarios/experiments too Mesa. As discussed [in the Mesa 4 goals discussion](https://github.com/mesa/mesa/discussions/2972#discussioncomment-15427010), it would be really convenient to have proper support for scenarios in Mesa. This PR is a first stab at achieving this.

### the problem this solves
Running computational experiments, or scenarios, with ABMs is at the core of ABM use. However, Mesa does not currently offer first-class support for this. Instead, users have to roll their own, or Mesa makes assumptions (in the GUI and the batch runner). Basically, all scenario parameters are assumed to be passed as keyword arguments to the `Model.__init__`. I personally tend to complement this with class-level attributes that I can easily change (e.g., `MyAgent.rationality = "expected_utility"`). As a side effect of all this is that if you want to experiment with specific parameters deep down in your agents, you have to pass them first to `Model.__init__` and then to the `__init__` of your custom agent. Moreover, if you want to experiment over a wide range of parameters, your `Model.__init__` becomes unwieldy with too many keyword arguments.

### implementation details
This PR adds a new `Scenario` class and updates the `Model` class to optionally use it. The scenario class takes keyword arguments only, and all keywords become available as attributes. I also decided to explicitly encapsulate the seeding of the random number generator into the scenario class using the spec-7 compliant rng argument. 

With all in place, `model.scenario` becomes your single source of truth for scenario parameters (hence the explicit inclusion of `rng`). No need to pass stuff around anymore via `__init__`'s of tinker with class variables.

### API

```python
scenario = Scenario(rng=42, density=0.8, minory_pc=0.5, homophily=0.5, radius=1)

model = Schelling(height=20, width=20, scenario=scenario)

# inside model
	
def __init__(self, height: int = 20, width: int = 20, scenario:Scenario|None=None):

	self.density = self.scenario["density"]
	self.minority_pc = self.scenario.minority_pc  # attribute access also work

	# rest proceeds as normal, see basic Schelling example

# inside agent
def __init__(self, model, cell, agent_type):
	super().__init__(model)
	self.cell = cell
	self.agent_type = agent_type

	self.homophily = model.scenario["homophily"]
	self.radius = model.scenario["radius"]
```

### DIFF
--- a/mesa/experimental/scenarios/__init__.py
+++ b/mesa/experimental/scenarios/__init__.py
@@ -2,4 +2,6 @@
 
 from .scenario import Scenario
 
-__all__ = ["Scenario",]
+__all__ = [
+    "Scenario",
+]


EwoutH:
I really really like the idea of this. Kind of obvious, weird we didn't think of it earlier.

> I personally tend to complement this with class-level attributes that I can easily change (e.g., `MyAgent.rationality = "expected_utility"`). As a side effect of all this is that if you want to experiment with specific parameters deep down in your agents, you have to pass them first to `Model.__init__` and then to the `__init__` of your custom agent.

Oh yes this is always a chore.

I would change one thing:

```Pyhton
    self.density = self.scenario["density"]
    self.minority_pc = self.scenario["minority_pc"]
```
What if we can stop doing this? Instead, why not just access scenario parameters directly?

```Python
    # Access scenario parameters directly
    self.grid = MultiGrid(width, height, torus=True, rng=self.scenario.rng)
    
    # Create agents based on scenario parameters
    num_agents = int(width * height * self.scenario.density)
    num_minority = int(num_agents * self.scenario.minority_pc)
```
In this design, it's always explicit what's scenario-based, and what's just a static model variable. You also prevent making errors in the Model init.

Implementation wise, I think you can use a (frozen) dataclass for this.

We could even allow passing an equivalent `policy` object, that works exactly the same, but now you can semantically separate them.

Anyway, thanks for initiating this, really interesting direction!

quaquel:
> What if we can stop doing this? Instead, why not just access scenario parameters directly?
>  You also prevent making errors in the Model init.

See my point 3. Basically, I am using a dict internally. I realize now that I can make attribute access work via `__getattr__` or some related magic. I doubt however that this will prevent typos and stuff like that because of all this works via the dynamic nature of python and thus won't be available to static language analysis tools (like type checking, autocomplete). 

> Implementation-wise, I think you can use a (frozen) dataclass for this.

That was also on my list of things to check, but again, I am not sure due to its dynamic nature. I don't want to force users to subclass Scenario to make any of this work. 


update:
1. attribute style, or dot style access works (and did work for the original commit). Because the keyword arguments are stored in the internal `__dict__`, they will be found automatically if you use attribute-style access. I enhanced this a bit, so deleting and setting can also use attribute access. 
2. Had a quick look at dataclasses. It can be done via make_dataclass, but that still requires the user to do something like

```python
Scenario = make_dataclass(Scenario, [("a", int), ("b", float)])
``` 
So rather cumbersome for no apparent benefit that I can see.

quaquel:
> We could even allow passing an equivalent `Policy` object, that works exactly the same, but now you can semantically separate them.

From a policy analysis point of view, I agree. It makes good sense to separate policies and scenarios. However, this is a very specific vocabulary. Scenario is more common and often used as a catch-all for model input. The other option that I am considering is to just call it `Experiment`.

quaquel:
This is ready for review. I added tests, and it's included in readthedocs. 

There are various follow-ups that are possible from this:
1. Update some of the examples to use it
2. Properly integrate it with Solara (interacts with the current requirement for keyword arguments only). 
3. Explore how this might simplify the experiment generation side of `batch_run`

However most of these are stable so its a bit tricky while this is still in experimental.

codebreaker32:
I’ve been following this PR, and I must say this was much needed and excellently implemented
I particularly like how this encapsulates the `rng` and provides a single source of truth for experimental parameters.

While reviewing the current implementation of the `Scenario` class, I noticed a potential technical edge case regarding the mutation protection. The current `__setitem__` and `__setattr__` logic successfully blocks **reassignment** while `model.running` is true:

```python
# This is correctly blocked
scenario.density = 0.8 

```

However, because `__getitem__` returns a direct reference to the internal object, **internal mutation** of mutable objects (like lists or dictionaries) can still occur during a run:

```python
# This bypasses the current protection
scenario.my_list.append(new_value)      # List
scenario.alpha["a"] = 1               # Dictionary
```

>The scenario class is just a dict with protection against changing it

To prevent mid-run changes, this 'mutation bypass' might be worth addressing.

What are your thoughts on whether this level of deep protection is necessary for the initial experimental release?

quaquel:
> However, because __getitem__ returns a direct reference to the internal object, internal mutation of mutable objects (like lists or dictionaries) can still occur during a run

You are absolutely right. 

1. This is Python, so the basic idea is that we are all consenting adults here. That is, rather than enforcing everything very strictly, one has to trust the user and ensure that this kind of information is well-documented.
2. There is a trade-off between simplicity and enforcement. Making the mutation of mutables in scenario classes impossible will be very difficult to achieve. The simplest but brutal solution is to only allow immutables inside the scenario class.
3. In my experience, when doing parameter sweeps, it's unlikely that you'll use lists or other mutables. 
4. The current block on changing scenarios while running was put in there by me as part of some wider thinking about run control. I am not fully sure yet how to proceed with all of it.

EwoutH:
I will try to investigate this further soon.

Quick point: In my our three stage problem solving flow (problem —> conceptual solution—> implementation), I’m in stage 2. So I fully agree this is a problem worth solving, but I have not yet wrapped my head around the conceptual here.

Quick question: Do we need both dict and attribute access? (how much) can we simplify if we only allow one (I would prefer attribute)?

quaquel:
> Quick question: Do we need both dict and attribute access? (how much) Can we simplify if we only allow one (I would prefer an attribute)?

If you drop `__getitem__` and `__setitem__`, you would lose dict-like access. There is probably some further tinkering required to clean up the rest as a result. I personally like dict-like access, but that is because of how I tend to think of computational experiments. I also don't see any downside to having both attribute style and dict style access.

EwoutH:
I looked into getting full IDE autocomplete and type checking for `self.scenario` in model subclasses. I think that would be very valuable in developer experience and reducing errors.

After exploring some options, the dataclass seems like the least complex user implementation, although not perfect:

```python
@dataclass
class SchellingScenario(Scenario):
    density: float
    minority_pc: float
    homophily: float = 0.5
    rng: int | None = None

class SchellingModel(Model):
    scenario: SchellingScenario  # ← Add this one line for full IDE support
    
    def __init__(self, scenario: SchellingScenario):
        
        # Now IDE autocompletes self.scenario.density, .minority_pc, etc.
        num_agents = int(width * height * self.scenario.density)
```
So, users would have to do two things:
1. Subclass Scenario
2. Add a class-level annotation (`scenario: SchellingScenario`)

The second is optional, without it you won't get autocomplete or type-hinting, but it still works functionally.

While subclassing is less elegant than direct instantiating, it does allow users to explicitly think about which variables and types they want in their scenarios, and optionally add default values.

And it does have significant advantages:
- Full autocomplete in the model class
- Type checking catches typos immediately
- Works in agents too: `def __init__(self, model: SchellingModel)` → autocomplete for `self.model.scenario.homophily`
- Self-documenting - the scenario dataclass IS the documentation

I think it is worth considering, curious what you think.

quaquel:
I am hesitant and deliberately decided not to go down that route. It adds yet another thing for users to subclass (next to agent and model). If you want to quickly explore some parameters, you would have to do 3 things. 

1. Update the sampling from e.g. `np.random.qmc.sobol`
2. Change the subclassing of Scenario
3. Create instances of the scenario dataclass based on your numpy array. No simple trick via dataframe.to_dict(orientation='record').
5. Update the use of scenarios inside your model. 

With my sketch, you don't need to worry about changing the subclassing at all (at the loss of autocomplete). 

Still, there are advantages to this approach.
1. You have to explicitly represent the parameters you want to explore over
2. You can also use defaults, giving you more flexibility (i.e., constants in the workbench).
3. Autocomplete and type hinting might work

I am not sure, by the way, that autocomplete and type hinting will work properly with this approach, but that can be tested. At least it will require some very careful type hinting throughout the entire codebase, as well as force our users to use type hinting. Basically, you need to indicate in the `__init__` of the model what the type of scenario is, so that `model.scenario` is then typed properly.

EwoutH:
I don't want to let perfect here be the enemy of done here. If we introduce the basic ability to just pass a `Scenario` class to the model, we can later revise the exact implementation of that class.

The very basis of this PR is allowing:

```Python
class MyModel(Model)
    def __init__(..., scenario=scenario):

	    ... = self.scenario.something
```
I don't think switching the backend to a dataclass or something else will occur breaking changes to the user API if we keep it this simple.

I would prefer attribute-only access for now. That makes one obvious way to do this, and makes it easier to migrate to other implementations without API breaks or migrations.

I also would suggest to users to use `self.scenario.variable` directly as the recommended pattern (in our tutorial, examples, etc.), instead of doing `self.minority_pc = self.scenario.minority_pc`. The former is way more explicit.

We could add a `scenario` property to the `Agent` that points to `self.model.scenario` for readability.

quaquel:
I updated the scenario class to remove dict-like access, leaving only attribute-style access.

I agree on the agent idea. But, to keep the code contained inside experimental at least for mesa 3.x, I have just put the idea inside the docstring of the scenario class.

mesa/experimental/scenarios/scenario.py
```diff
@@ -0,0 +1,76 @@
...
+        """Initialize a Scenario.
+
+        Args:
+            rng: a random number generator or valid seed value for a numpy generator.
+            kwargs: all other scenario parameters
+
+        """
+        self.model: M | None = None
+        self.scenario_id: int = next(self._ids[self.__class__])
+        self.__dict__.update(rng=rng, **kwargs)
```
EwoutH, line 62: It looks like the `rng` is stored as a raw value but then passed to Model. If users pass an actual Generator object, this might not serialize well. Should we consider documenting this or adding validation?

mesa/experimental/scenarios/scenario.py
```diff
@@ -0,0 +1,76 @@
...
+    def __init__(self, *, rng: RNGLike | SeedLike | None = None, **kwargs):
+        """Initialize a Scenario.
+
+        Args:
+            rng: a random number generator or valid seed value for a numpy generator.
+            kwargs: all other scenario parameters
+
+        """
+        self.model: M | None = None
+        self.scenario_id: int = next(self._ids[self.__class__])
```
EwoutH: I'm a bit hesitant to make this Scenario-dependent. That means if you for example run in a Jupyter notebook, and define your Scenario in a different cell, and only rerun your model cell, you will have different numbers.

mesa/experimental/scenarios/model_with_scenario.py
```diff
@@ -0,0 +1,33 @@
+"""Small extension to the base model class to add support for the experimental Scenario."""
+
+from mesa.model import Model
+
+from .scenario import Scenario
+
+
+class ModelWithScenario(Model):
```
EwoutH: I must say, I like this stuff. I won't mind going straight into the main `Model` class itself.

We could probably integrate it quite cleanly by doing something like:

```Python
class Model:
    def __init__(...):
        ...
        # If scenario provided, use its rng
        if scenario is not None:
            if seed is not None or rng is not None:
                raise ValueError("Cannot specify both scenario and seed/rng")
            rng = scenario.rng
            self._scenario = scenario
            scenario.model = self
        else:
            # Create implicit scenario from kwargs for backwards compat
            self._scenario = Scenario(rng=rng, seed=seed, **kwargs)
        ...
```

tests/experimental/test_scenarios.py
```diff
@@ -0,0 +1,46 @@
...
+
+    model = ModelWithScenario(scenario=scenario)
+    model.running = True
+    assert model.scenario.model is model
+
+    with pytest.raises(ValueError):
+        scenario.a = 5
+
+    model = ModelWithScenario()
+    assert model.scenario.rng is None
```
EwoutH: Maybe add these tests, for those edge cases:
```suggestion
    assert model.scenario.rng is None
    
def test_scenario_serialization():
    """Test that scenarios can be pickled/unpickled."""
    scenario = Scenario(a=1, rng=42)
    import pickle
    pickled = pickle.dumps(scenario)
    unpickled = pickle.loads(pickled)
    assert unpickled.a == 1
    assert unpickled.scenario_id == 0

def test_scenario_with_generator():
    """Test scenario with actual Generator object."""
    gen = np.random.default_rng(42)
    scenario = Scenario(rng=gen)
    model = ModelWithScenario(scenario=scenario)
    # Should work without error
    assert model.rng is not None

```

mesa/experimental/scenarios/scenario.py
```diff
@@ -0,0 +1,76 @@
...
+        self.__dict__.update(rng=rng, **kwargs)
+
+    def __iter__(self):  # noqa: D105
+        return iter(self.__dict__.items())
+
+    def __len__(self):  # noqa: D105
+        return len(self.__dict__)
```
EwoutH, line 68: To confirm, it's intended to leave `model` and `scenario_id` out?

mesa/experimental/scenarios/scenario.py
```diff
@@ -0,0 +1,76 @@
...
+            pass
+        super().__setattr__(name, value)
+
+    def to_dict(self):
+        """Return a dict representation of the scenario."""
+        data = self.__dict__.copy()
+        for entry in self.__slots__:
+            if entry != "__dict__":
+                data[entry] = getattr(self, entry)
+        return data
```
EwoutH: I think we can simplify it to:
```suggestion
        return {
            **self.__dict__,
            'model': self.model,
            'scenario_id': self.scenario_id
        }
```
Note that `**self.__dict__` is a shallow copy already.

EwoutH (commented):
Good stuff. Some code nitpicks and suggestions.

I would support going straight into the Model class instead of creating a separate `ModelWithScenario`.

We could keep Scenario itself in the experimental space. This way we can still make breaking changes, but users don't need a new Model.

We can add a warning that `Model.scenario` is a reserved variable name (just like `Model.time`, `.agents`, etc.)

mesa/experimental/scenarios/scenario.py
quaquel: I'm not sure what you mean by serialization/validation.

mesa/experimental/scenarios/scenario.py
quaquel: Again not sure what you mean.

So, yes, whenever you instantiate a Scenario, it increments the counter. But doing it differently, e.g., by storing the index based on all attributes is quite a bit more convoluted and I fail to see the value of doing that. 

The primary use case of scenario_id is when performing a batch run.

mesa/experimental/scenarios/model_with_scenario.py
quaquel: I'll take a closer look to see how difficult it would be to cleanly splice this into the existing model class. It will be a bit tricky since at present you already have seed or rng but not both, which now gets expanded to include seed or rng or scenario.

mesa/experimental/scenarios/scenario.py
quaquel: yes that is deliberate. Basically, len(scenario) just gives you the number of parameters you are changing via the scenario. The model and scenario_id attribute are just bookkeeping and so not part of the size of the scenario.

codebreaker32:
Huge props for this @quaquel! This is a massive improvement. I'm really excited to see how this lands
While I was testing this out with `batch_run`, I noticed a critical memory leak/edge case due to the current bidirectional design. 

Currently, there is a circular reference: Model has self.scenario, and Scenario has self.model.
When batch_run executes, it stores the input parameters (including the scenario object) in the results dictionary. Because the scenario object holds a reference to self.model, batch_run inadvertently returns the entire dead Model instance (including the Grid, Schedule etc) in the results list.

We may fix this by adding custom` __deepcopy__`

quaquel:
Out of curiosity, how did you combine this with batch-run?

For me, this is a stepping stone to making batch run obsolete. In this, scenarios are expected to be simple data containers that only hang around shortly and so the circular reference should not really be an issue.

codebreaker32:
I copied the minimal version of both scenario and model_with_scenario and then passed it to batch_run just to check for this bidirectional dependency

```python
import pytest
from mesa import Model

class Scenario:
    def __init__(self, **kwargs):
        self.model = None  
        self.__dict__.update(kwargs)
    

class ModelWithScenario(Model):
    def __init__(self, scenario=None, rng=None):
        self.scenario = scenario
        
        if self.scenario:
            self.scenario.model = self  
        
            model_reporters={"Step": lambda m: m.steps}
        )
        self.running = True
        self.steps = 0
        self.time = 0

    def step(self):
        self.time += 1
        self.steps += 1


def test_batch_run():
    # 1. Create a Scenario
    sc = Scenario(parameter="Test")
    
    # 2. Execute batch_run 
    results = batch_run(
        ModelWithScenario,
        parameters={"scenario": sc},
        max_steps=5,
        number_processes=1,
        display_progress=False
    )
    
    
    first_result = results[0]
    returned_scenario = first_result['scenario']
    
    print(f"\n[DEBUG] Returned Scenario: {returned_scenario}")
    print(f"[DEBUG] Attached Model: {returned_scenario.model}")
    
    # THE FAIL CONDITION
    error_msg = (
        f"MEMORY LEAK: The 'scenario' object in results holds a reference "
        f"to the '{type(returned_scenario.model).__name__}' instance."
    )
    
    assert returned_scenario.model is None, error_msg

if __name__ == "__main__":
    pytest.main([__file__])

```

codebreaker32:
> For me, this is a stepping stone to making batch run obsolete. In this, scenarios are expected to be simple data containers that only hang around shortly and so the circular reference should not really be an issue.

I strongly support the move to replace `batch_run`(obviously there are many reasons to do so)
```python
# batchrunner.py
stepdata = [
    {
         ...
        **kwargs,  # <--- The culprit
         ...
    }
]
```
This confirms that `batch_run` is too "dumb" to handle sophisticated inputs—it treats everything as static logging data rather than simulation controls.

To solve this lifecycle issue, I imagine we need something that explicitly manages the "Init -> Run -> Extract -> Dispose" loop. Instead of a functional loop, maybe an object-oriented approach like this?

```python
exp = Experiment(MyModel, scenario=my_scenario)
results = exp.run()  

```
Any direction from the experts would be highly appreciated :)

Note: I have gone through the discussion in #2972

mesa/experimental/scenarios/scenario.py
EwoutH: When users run batch experiments or save/load models, Python needs to convert objects to bytes (pickling/serialization). If someone passes an actual `np.random.Generator` object as `rng` instead of an integer seed, it might fail when you try to pickle the scenario or could create subtle bugs when deserializing (the RNG state might not restore correctly).

We could consider: (1) documenting that `rng` should be a seed value, not a Generator object, (2) adding a check that raises an error if someone passes a Generator, or (3) handling both cases by extracting/storing the seed from Generator objects so serialization always works.

Bit of an edge-case, but serialization has troubled me too many times (i.e. hashability of Agents).

mesa/experimental/scenarios/scenario.py
EwoutH: Let's make `scenario_id` private (`_scenario_id`) until we know what we want with it exactly (both for internal use and utility / use-cases for end-users).

EwoutH:
@codebreaker32 this is good stuff, worth following up on.

> Any direction from the experts would be highly appreciated :)

I would recommend starting to open a Discussion. Propose your own ideas and ask questions on what you're not sure about yet.

Start with target audiences and use cases. For who is it meant, and when are they going to use it? Then build up from there.

Also look at other libraries. What does already exist, and what libraries might we want to connect to. For example, What does the [EMAworkbench](https://emaworkbench.readthedocs.io/) take for input? What methods/hooks/data structures can we provide to simplify [this example](https://github.com/quaquel/EMAworkbench/blob/master/ema_workbench/examples/example_mesa.py)?

Exploring, discussing and answering the two points above will give insight in current demand and supply, and thus where the gaps are. That will hopefully help answer questions like:

- How do we handle uncertainty axes? (attributes)
- Do we need a Scenario set/collection?
- Do (and if so how do) we want to provide scenario generation/sampling?

Use case --> existing stuff --> gap --> user API --> high-level design is often a decent process (however not always linear).

Hope it helps, let me know if you need anything else.

mesa/experimental/scenarios/scenario.py
quaquel: It's actually not an edge case because when you run in parallel, its likely that we'll do something like `pool.map(scenarios, run_function)`. Here scenarios is a collection of scenarios. However, the pool will then use pickle to send the individual scenario objects to the workers. So, I'll add the tests you suggested below and expand on this a bit more to properly cover pickling.

quaquel:
> Any direction from the experts would be highly appreciated :)

See also [my answer](https://github.com/mesa/mesa/discussions/2972#discussioncomment-15427010) on the mesa 4 discussion

tests/experimental/test_scenarios.py
quaquel: added

mesa/experimental/scenarios/model_with_scenario.py
quaquel: I have taken a first step at integrating it in Model but it's a bit tricky.

I am inclined to simply deprecate' seed' already (even if we keep `model.random`...) just to simplify the call signature. We can just use rng as an argument to `random.Random` and if that raises an exception, just create a random float on unit interval via `rng.random(1,)` and use this to seed `random.Random`.

quaquel:
> Currently, there is a circular reference: Model has self.scenario, and Scenario has self.model.
When batch_run executes, it stores the input parameters (including the scenario object) in the results dictionary. Because the scenario object holds a reference to self.model, batch_run inadvertently returns the entire dead Model instance (including the Grid, Schedule etc) in the results list.

The problem, of course, stems from the design choice to prevent updating scenarios while the model is running. I am slowly coming to the point where I think we don't need to add this protection explicitly, but rather make it clear that changing the content of a scenario or replacing the scenario on a model while it's running can lead to undefined behavior. This would simplify the problem and also means that a scenario won't need a reference to `model`.

mesa/model.py
EwoutH, line 1: Overall a clean implementation! I like it!

codebreaker32:
I think we should explicitly state in the documentation that users must perform any validation or logic on the Scenario before calling `super().__init__()` in their custom models because doing it afterwards will raise `ValueError`

quaquel:
> I think we should explicitly state in the documentation that users must perform any validation or logic on the Scenario before calling super().__init__() in their custom models because doing it afterwards will raise ValueError

I am not sure what you mean. Could you elaborate?

codebreaker32:
I mean doing this straight away will result in `ValueError` because `self.model.running = True`
```python
class MyModel(Model):
    def __init__(self, scenario):
        if self.scenario.density < 0:
            self.scenario.density = 0

```
Either
1. we can set `self.running = False` then apply logic or validate scenario OR
2. we can call `super().__init__()` afterwards

I think either way, we should explicitly mention it in documentation

EwoutH:
Could we use the new `Model.time` variable, to check `Model.time > 0` instead of `running`?

Maybe don't enforce it now, but first think about time progression and run control.

mariuzka:
> For me, this is a stepping stone to making batch run obsolete. In this, scenarios are expected to be simple data containers that only hang around shortly and so the circular reference should not really be an issue.

Out of curiosity, as I try to understand the direction Mesa is currently going: how could scenarios replace batch_run? Where would the variation of parameter values happen?

quaquel:
>Out of curiosity, as I try to understand the direction Mesa is currently going: how could scenarios replace batch_run? Where would the variation of parameter values happen?

See [my answer](https://github.com/mesa/mesa/discussions/2972#discussioncomment-15427010) on the mesa 4 discussion. In short, `Scenario` just encapsulates a single experiment. 

A next step is to have some helper functions to create a collection of scenarios. Since there are ample libraries out that for creating experimental designs (e.g., salib, numpy.stats.qmc, ema_workbench), I am inclined to start with a helper function that takes a dataframe or numpy array with a list of columns, which returns an iterable of experiments. I have to think a bit about how to splice in seeds in this, but that is a minor detail.

quaquel:
@mariuzka, I just added two helper functions to scenario.py that might help clarify my intended direction: `scenarios_from_dataframe` and `scenarios_from_numpy`. 

While adding these two helper functions, I realized that, from a computational experimentation perspective, there is an important difference between an experiment and any specific stochastic realization of the experiment (i.e., an experiment and a specific seed value). Ideally, there is a way of identifying both easily in post-processing. For example, you might want to group by experiment identifier before taking statistics over the stochastic realizations (i.e., replications using different seeds). 

Conceptually, a model need not be aware of either the scenario identifier or the experiment identifier. But when running experiments using, e.g., a `ProcessPoolExecutor`, it's rather useful if this information is encapsulated in the scenario object. For now, I have solved this by simply adding a keyword argument called `_experiment_id`; however, I am inclined to incorporate this into the Scenario class itself and handle it in a similar manner to `_scenario_id`. So, if it is passed in the init, we use that; otherwise, we use an internal counter to generate it. Thoughts are welcome.

EwoutH:
> I just added two helper functions to scenario.py that might help clarify my intended direction: `scenarios_from_dataframe` and `scenarios_from_numpy`.

While potentially the way to go, I think it's good to have a broader discussion about A) scenario generation and B) scenario object storage. Can we keep the helper function outside this PR, and discuss them for follow-up work in a separate thread?

Maybe open a new discussion in which we can discuss scenarios and experimentation on a higher level.

quaquel:
Let's just get this merged. I removed the two functions for now.

EwoutH (approved):
Looks great, thanks!

Could you update the PR description to reflect the final state of the PR?

Issue #3101, opened 2026-01-09
Make Agent hashability requirement explicit in type system


EwoutH:
### Summary

The `Agent` class should explicitly inherit from `collections.abc.Hashable` to document that agents must be hashable, since `AgentSet` uses `WeakKeyDictionary` which requires hashable keys.

### Background

From the Python [documentation](https://docs.python.org/3/glossary.html):

> **hashable**
> 
> An object is hashable if it has a hash value which never changes during its lifetime (it needs a `__hash__()` method), and can be compared to other objects (it needs an `__eq__()` method). Hashable objects which compare equal must have the same hash value.
> 
> Hashability makes an object usable as a dictionary key and a set member, because these data structures use the hash value internally.
> 
> Most of Python’s immutable built-in objects are hashable; mutable containers (such as lists or dictionaries) are not; immutable containers (such as tuples and frozensets) are only hashable if their elements are hashable. Objects which are instances of user-defined classes are hashable by default. They all compare unequal (except with themselves), and their hash value is derived from their `id()`.

### Current Behavior

`AgentSet` stores agents in a `WeakKeyDictionary`, which requires agents to be hashable. The base `Agent` class relies on Python’s default hashability (based on object `id()`), but this requirement is not explicit in the type signature.

### Proposed Change

```python
from collections.abc import Hashable

class Agent[M: Model](Hashable):
    """Base class for a model agent in Mesa.
    
    Notes:
        Agents must be hashable to be stored in AgentSets. If you override
        __eq__(), you must also implement __hash__() to maintain hashability.
    """
```

### Benefits

- Makes the hashability requirement explicit for type checkers and developers
- Helps prevent bugs where users override `__eq__()` without `__hash__()`, breaking AgentSet functionality
- Improves API documentation through the type system
- No runtime behavior change since `Agent` is already hashable by default​​​​​​​​​​​​​​​​

quaquel:
As evidenced by #3104, doing this seems like a terrible idea because of overhead. Instead, it should just be documented clearly.

For my info, what were you doing rendering agents not hashable? I have never seen this because the hash of an object is just is memory location.

codebreaker32:
I suggest doing this might be a good idea if we want it to enforce in type checkers
```python
if TYPE_CHECKING:    
    from collections.abc import Hashable
    _Base = Hashable
else:
    # At runtime: Use a dummy class
    class _Base:
        pass

class Agent[M: Model](_Base):
    """Base class for a model agent in Mesa.
[...]

PR #3104, opened 2026-01-09
Document `Agent` hashability requirement

Sonu0305:
### Summary
Clarifies the requirement for Mesa agents to be hashable and adds tests to verify this behavior, particularly when `__eq__` is overridden in subclasses.

### Motive
`AgentSet` requires agents to be hashable because it uses them in a `WeakKeyDictionary`. This requirement was previously implicit. An earlier attempt to enforce this via the type system (`collections.abc.Hashable`) was reverted due to significant performance regressions.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -36,8 +36,9 @@ class Agent[M: Model]:
         pos (Position): A reference to the position where this agent is located.
 
     Notes:
-        Agents must be hashable to be stored in AgentSets. If you override
-        __eq__(), you must also implement __hash__() to maintain hashability.
+        Agents must be hashable to be used in an AgentSet.
+        In Python 3, defining `__eq__` without `__hash__` makes an object unhashable,
+        which will break AgentSet usage.
         unique_id is unique relative to a model instance and starts from 1
 
     """


EwoutH:
Wow I didn’t expect this much performance impact. Especially WolfSheep is terrible, it’s the only model that actively removes and adds agents during the run.

I don’t think this is a proportional performance trade-off, glad we have the benchmarks.

Curious if there are more efficient ways to enforce hashibility.

Sonu0305:
> Wow I didn’t expect this much performance impact. Especially WolfSheep is terrible, it’s the only model that actively removes and adds agents during the run.
> 
> I don’t think this is a proportional performance trade-off, glad we have the benchmarks.
> 
> Curious if there are more efficient ways to enforce hashibility.

Right, benchmarks got us safe.
I also agree to @quaquel regarding we should just include in the documentation that I have already added to the `Notes` section of the class `Agent`.

Sonu0305:
I have updated the pr description to be precise.
I am a bit confused regarding how do I cover the lines added to the test file itself? (the lines which are causing the codecov test to fail).

mesa/agent.py
```diff
@@ -36,7 +36,9 @@ class Agent[M: Model]:
         pos (Position): A reference to the position where this agent is located.
 
     Notes:
-          unique_id is unique relative to a model instance and starts from 1
+        Agents must be hashable to be stored in AgentSets. If you override
+        __eq__(), you must also implement __hash__() to maintain hashability.
+        unique_id is unique relative to a model instance and starts from 1
```
quaquel: I disagree with this statement. 

`__eq__` is accessed via `==`, `__hash__` is used for `is`. Equality and identity are not the same thing, so it can be perfectly legitimate to override only `__eq__` because equality is assessed at the attribute level. Two objects of the same type and with the same values for all attributes are equal. However, they are not identical because they occupy two different memory locations.

tests/test_agent.py
```diff
@@ -710,3 +710,46 @@ def custom_agg(values):
...
+                isinstance(other, HashableAgent) and self.unique_id == other.unique_id
+            )
+
+        def __hash__(self):
+            return hash(self.unique_id)
+
+    model = Model()
+    agent = HashableAgent(model)
+    test_set = {agent}
+    assert agent in test_set
```
quaquel: since this has turned into a doc change, are these tests still needed

Sonu0305: Okay then, I'll take them down, thanks.

mesa/agent.py
Sonu0305: This is what exactly was mentioned in the issue's (#3101) proposed change.
However for refinement, does the below make sense to you?
```
    Notes:
        Agents must be hashable to be used in an `AgentSet`. If you override
        `__eq__` in a subclass of `Agent`, Python makes it unhashable. To
        restore hashability, you must also implement `__hash__`.
```

quaquel: No because its not correct. Again, equality (`==`) and identity (`is`) are not the same thing. It is completely valid to overwrite `__eq__` without overwriting `__hash__`. 

```python
class MyObject:
    def __init__(self,a):
        self.a=1

    def __eq__(self, other):
        return self.a==other.a

a = MyObject(1)
b = MyObject(1)

> a is B
False

> a == B 
True

```

Sonu0305: Thanks for the clarification, I think we’re talking past each other slightly.

I fully agree that equality (`==`) and identity (`is`) are different concepts, and that it is conceptually valid to override `__eq__` without changing object identity.

However, in Python 3, defining `__eq__` without defining `__hash__` causes Python to automatically set `__hash__ = None`, making the object unhashable. This is standard behavior per the Python data model.

Since `AgentSet` relies on `WeakKeyDictionary`, agents must remain hashable at runtime. Therefore, overriding `__eq__` on an `Agent` subclass will break `AgentSet` unless `__hash__` is explicitly restored.

To avoid confusion, would the following wording be acceptable?
```
Notes:
    Agents must be hashable to be used in an AgentSet.
    In Python 3, defining __eq__ without __hash__ makes an object unhashable,
    which will break AgentSet usage.
```
This avoids conflating equality with identity and simply documents the Python behavior that affects `AgentSet`.

Sonu0305: This is also around the lines mentioned by @EwoutH in the issue template.

quaquel: that note sounds perfect to me.

Sonu0305: Thanks, I have updated the same.

quaquel (approved):

PR #3109, opened 2026-01-10
Fix batch_run Data Collection to Ensure Accuracy and Capture All Steps

### LINKED ISSUE #2514: Batch runner doesn't collect data on last step
**Describe the bug**
When using batch_run the datacollector misses off the last step. 

**Expected behavior**
When using batch_run the datacollector doens't miss of the last step.

**To Reproduce**

MRE:

```
import mesa
import pandas as pd

class Foo(mesa.Agent):
    def __init__(self, model: mesa.Model) -> None:
        super().__init__(model)
        self.tally = 0
        
    def step(self) -> None:
        self.tally += 1

class FooHaus(mesa.Model):
    def __init__(self, foos) -> None:
        super().__init__()
        for f in range(foos):
            Foo(self)
        self.datacollector = mesa.DataCollector(agent_reporters={
            "Tally": "tally" #Need something in the datacollector
        })
    
    def check_end_condition(self) -> None:
        if self.steps >= 3:
            self.running = False
            print("Integer of final model step = ", self.steps)
            print("Tally from Foo agent = ", self.agents[0].tally)

    def step(self) -> None:
        self.check_end_condition()

print("1st case: Run model then display agent_vars dataframe")

test = FooHaus(1)
test.run_model()
test_results = test.datacollector.get_agent_vars_dataframe()
print(test_results)

print("------------------------------")
print("2nd case: batch_run then display results from data collector")
batch_run_results = mesa.batch_run(
    model_cls=FooHaus,
    parameters={"foos":1},
    number_processes=1,
    iterations=1,
    data_collection_period=1,
    max_steps=1000,
    display_progress=False
)

batch_run_results_df = pd.DataFrame(batch_run_results)
print(batch_run_results_df)
```

END MRE

**Additional context**

The batch_run datacollector also adds a zero step, I'm not sure if this is intended or not but I can see the value of this for viewing an initial state.

I think the issue is in the function _model_run_func in the mes.batchrunner source code when defining the list of steps that _collect_data is run on.
```
steps = list(range(0, model.steps, data_collection_period))
    if not steps or steps[-1] != model.steps - 1:
        steps.append(model.steps - 1)

 for step in steps:
        model_data, all_agents_data = _collect_data(model, step)
```
I believe the issue could be that the steps used for model collection goes up to the penultimate step due to ```model.steps - 1```

codebreaker32:
### Summary

This PR fixes a bug where `batch_run` failed to capture the final step of a simulation and incorrectly handled data in models with non-standard collection frequencies. It transitions the `BatchRunner` from a "prediction-based" step calculation to an "observation-based" approach, ensuring 100% data integrity.

### Bug / Issue

**Related Issue:** Fixes #2514

When using `batch_run`, the `DataCollector` would frequently miss the final step of the model. This was caused by the function `_model_run_func` using a hardcoded `range()` based on `model.steps - 1`, which stopped the data extraction one step short of the actual final state.

However #2514 wasn't clearly able to state the problem i.e, the previous implementation assumed a strict 1-to-1 relationship between model steps and data rows. This led to:

* **Data Loss:** In models with "Time Dilation" (collecting multiple times per step), sub-step data was silently deleted because `range()` only generates one integer per step.
* **Ghost Data:** In sparse collection models, `batch_run` would try to fetch data for steps where none was collected, resulting in duplicated or `NaN` rows.

### Implementation

The logic in `_model_run_func` has been refactored to stop guessing the step indices and instead trust the `DataCollector`'s recorded history.

**Key Changes:**

* **Truth-Sourcing:** Replaced `range(0, model.steps, ...)` with a direct lookup of `model.datacollector._collection_steps`.

### Testing

* **Accuracy Verification:** Confirmed via the provided MRE that the all steps are collected
* **Fault in Previous Code:** Using `range()` only returned 6 rows for a 5-step TimeDilation run (Initial + 5 steps), losing nearly half the data(locally tested using TimeDilationModel in test_batch_run)
* **New Code Result:** Correctly retrieves all 11 rows (Initial + 2 collections per step  5 steps), ensuring `len(results) == 11`.

### Additional Notes

**Performance vs. Accuracy Trade-off:**
By moving away from `range()`, we now handle the actual list of timestamps recorded by the model. While this ensures perfect accuracy for all model types (sparse, dilated, or standard), it introduces an overhead. So I think the point of debate is to decide whether this change is worth the performance tradeoff

MRE 
```python
import mesa
import pandas as pd

class TimeDilationModel(mesa.Model):
    def __init__(self,rng=None):
        super().__init__()
            model_reporters={"Step_Value": lambda m: m.steps}
        )
        # 1. Collect Initial State (Step 0)

    def step(self):
        
        if self.steps >= 2:
            self.running = False


# Logic: 
# Step 0: 1 collection (Init)
# Step 1: 2 collections (Start/End)
# Step 2: 2 collections (Start/End)
# Total Expected Rows: 5

results = mesa.batch_run(
    TimeDilationModel,
    parameters={},
    iterations=1,
    max_steps=10,
    data_collection_period=1,
    display_progress=False
)

df = pd.DataFrame(results)

print("--- RESULTS ---")
print(df[["Step", "Step_Value"]])
print(f"\nTotal Rows: {len(df)}")

expected = 5
if len(df) == expected:
    print("\nSUCCESS: All sub-step data captured.")
else:
    print(f"\nFAILED: Data Loss! Expected {expected} rows, but got {len(df)}.")
```

### DIFF
--- a/mesa/batchrunner.py
+++ b/mesa/batchrunner.py
@@ -219,13 +219,10 @@ def _model_run_func(
         match data_collection_period:
             case -1:
                 steps = [recorded_steps[-1]] if recorded_steps else []
-
             case 1:
                 steps = recorded_steps
-
             case _:
                 steps = recorded_steps[::data_collection_period]
-
     for step in steps:
         model_data, all_agents_data = _collect_data(model, step)
 


mesa/batchrunner.py
```diff
@@ -202,14 +202,29 @@ def _model_run_func(
...
+
+        if data_collection_period == -1:
+            steps = [recorded_steps[-1]] if recorded_steps else []
+
+        elif data_collection_period == 1:
+            steps = recorded_steps
+
+        else:
+            steps = recorded_steps[::data_collection_period]
+
```
quaquel: It's bad practice to put all this inside the try .. catch. 

The only thing that matters here is the first line. The 3 clause if stuff should be moved outside like this


```suggestion
    try:
        recorded_steps = model.datacollector._collection_steps
    except AttributeError:
        ...
     else: 
            if data_collection_period == -1:
                steps = [recorded_steps[-1]] if recorded_steps else []
            elif data_collection_period == 1:
                steps = recorded_steps
            else:
                steps = recorded_steps[::data_collection_period]
```

You could also change the 3-clause if with a match-based approach

```python
match data_collection_period:
    case -1
        steps = [recorded_steps[-1]] if recorded_steps else []
    case 1:
        steps = recorded_steps
    case _:
        steps = recorded_steps[::data_collection_period]
```

quaquel:
Can you ensure proper test code coverage? 

Also, any explanation for the performance change?

codebreaker32:
The new code processes more data because it stops deleting "sub-steps"

**Old Implementation:** The code used `range()` to generate a list of unique integers.
```python
# Old Code
steps = list(range(0, model.steps, data_collection_period))

```


If your model ran for 5 steps but collected data 3 times per step (15 total records), `range(0, 5)` only generated **5** indices. The loop ran 5 times, silently ignoring the other 10 records.
**New Implementation:** The code points directly to the `DataCollector`'s full history.

If the `DataCollector` recorded 15 entries, `recorded_steps` has length 15. The loop `for step in steps:` now executes **15 times** instead of 5.
The "slowdown" is simply the time it takes to process the 10 rows of data that were previously being deleted.
Honestly, I expected far more increased execution time because of this but I guess it is improving performance also by eliminating `list(range(...))` (O(n/k) performance overhead)

The new test results on old code :-
```plaintext
FAILED tests/test_batch_run.py::test_batch_run_sparse_collection - AssertionError: Expected 4 rows for sparse collection, got 20
assert 20 == 4
 +  where 20 = len([{'RunId': 0, 'iteration': 0, 'Step': 0, 'collect_interval': 5, 'rng': 42, 'Value': 4}, {'RunId': 0,
 'iteration': 0, 'Step': 1, 'collect_interval': 5, 'rng': 42, 'Value': 4}, {'RunId': 0, 'iteration': 0, 'Step': 2, 
'collect_interval': 5, 'rng': 42, 'Value': 4}, {'RunId': 0, 'iteration': 0, 'Step': 3, 'collect_interval': 5, 'rng': 42,
 'Value': 4}, {'RunId': 0, 'iteration': 0, 'Step': 4, 'collect_interval': 5, 'rng': 42, 'Value': 4}, {'RunId': 0, 'iteration': 0, 
'Step': 5, 'collect_interval': 5, 'rng': 42, 'Value': 4}, {'RunId': 0, 'iteration': 0, 'Step': 6, 'collect_interval': 5, 'rng': 
42, 'Value': 4}, {'RunId': 0, 'iteration': 0, 'Step': 7, 'collect_interval': 5, 'rng': 42, 'Value': 4}, {'RunId': 0, 'iteration': 
0, 'Step': 8, 'collect_interval': 5, 'rng': 42, 'Value': 4}, {'RunId': 0, 'iteration': 0, 'Step': 9, 'collect_interval': 5, 
'rng': 42, 'Value': 4}, {'RunId': 0, 'iteration': 0, 'Step': 10, 'collect_interval': 5, 'rng': 42, 'Value': 9}, {'RunId': 0,
 'iteration': 0, 'Step': 11, 'collect_interval': 5, 'rng': 42, 'Value': 9}, {'RunId': 0, 'iteration': 0, 'Step': 12, 
'collect_interval': 5, 'rng': 42, 'Value': 9}, {'RunId': 0, 'iteration': 0, 'Step': 13, 'collect_interval': 5, 'rng': 42,
 'Value': 9}, {'RunId': 0, 'iteration': 0, 'Step': 14, 'collect_interval': 5, 'rng': 42, 'Value': 9}, {'RunId': 0, 'iteration':
 0, 'Step': 15, 'collect_interval': 5, 'rng': 42, 'Value': 14}, {'RunId': 0, 'iteration': 0, 'Step': 16, 'collect_interval': 5,
 'rng': 42, 'Value': 14}, {'RunId': 0, 'iteration': 0, 'Step': 17, 'collect_interval': 5, 'rng': 42, 'Value': 14}, {'RunId': 0,
 'iteration': 0, 'Step': 18, 'collect_interval': 5, 'rng': 42, 'Value': 14}, {'RunId': 0, 'iteration': 0, 'Step': 19, 
'collect_interval': 5, 'rng': 42, 'Value': 14}])

FAILED tests/test_batch_run.py::test_batch_run_time_dilation - AssertionError: Data Loss! Expected 11 rows, got 6
assert 6 == 11
 +  where 6 = len([{'RunId': 0, 'iteration': 0, 'Step': 0, 'rng': None, 'RealStep': 0.0}, {'RunId': 0, 'iteration': 0, 
'Step': 1, 'rng': None, 'RealStep': 1.0}, {'RunId': 0, 'iteration': 0, 'Step': 2, 'rng': None, 'RealStep': 2.0}, {'RunId':
 0, 'iteration': 0, 'Step': 3, 'rng': None, 'RealStep': 3.0}, {'RunId': 0, 'iteration': 0, 'Step': 4, 'rng': None, 
'RealStep': 4.0}, {'RunId': 0, 'iteration': 0, 'Step': 5, 'rng': None, 'RealStep': 5.0}])

FAILED tests/test_batch_run.py::test_batch_run_missing_step - AssertionError: Expected 4 rows, got 7

assert 7 == 4
 +  where 7 = len([{'RunId': 0, 'iteration': 0, 'Step': 0, 'rng': None, 'Value': 0.0}, {'RunId': 0, 'iteration': 0, 'Step': 
1, 'rng': None, 'Value': 0.0}, {'RunId': 0, 'iteration': 0, 'Step': 2, 'rng': None, 'Value': 2.0}, {'RunId': 0, 'iteration': 
0, 'Step': 3, 'rng': None, 'Value': 2.0}, {'RunId': 0, 'iteration': 0, 'Step': 4, 'rng': None, 'Value': 4.0}, {'RunId': 0, 
'iteration': 0, 'Step': 5, 'rng': None, 'Value': 4.0}, {'RunId': 0, 'iteration': 0, 'Step': 6, 'rng': None, 'Value': 6.0}])

FAILED tests/test_batch_run.py::test_batch_run_empty_collection_edge_case - AssertionError: Expected 1 
row, got 4
assert 4 == 1
 +  where 4 = len([{'RunId': 0, 'iteration': 0, 'Step': 0, 'rng': None, 'Value': 3.0}, {'RunId': 0, 'iteration': 0, 'Step': 
1, 'rng': None, 'Value': 3.0}, {'RunId': 0, 'iteration': 0, 'Step': 2, 'rng': None, 'Value': 3.0}, {'RunId': 0, 'iteration':
 0, 'Step': 3, 'rng': None, 'Value': 3.0}])

================ 4 failed, 11 passed in 13.85s ===========================

```

quaquel:
Code coverage sits at less than 80%. Can you please ensure full coverage? 

You can click on codecov/patch to see which code paths are not yet covered by tests.

codebreaker32:
Added tests to ensure 100% patch coverage

mesa/batchrunner.py
```diff
@@ -202,14 +202,29 @@ def _model_run_func(
...
+    else:
+        match data_collection_period:
+            case -1:
+                steps = [recorded_steps[-1]] if recorded_steps else []
+
+            case 1:
+                steps = recorded_steps
+
+            case _:
+                steps = recorded_steps[::data_collection_period]
```
quaquel:
```suggestion
        match data_collection_period:
            case -1:
                steps = [recorded_steps[-1]] if recorded_steps else []
            case 1:
                steps = recorded_steps
            case _:
                steps = recorded_steps[::data_collection_period]
```

quaquel (approved):

PR #3111, opened 2026-01-10
Optimize _Grid.move_agent_to_one_of() - hybrid vectorized/loop closest selection

codebyNJ:
### Summary
Improved performance of `_Grid.move_agent_to_one_of(selection="closest")` by using a hybrid strategy: keep the original loop for small candidate sets and use a NumPy-vectorized distance computation for large candidate sets. This preserves semantics and randomness while significantly reducing runtime for large candidate lists.

### Bug / Issue
The original "closest" selection iterated in Python over all candidate positions `(O(n) in Python)`, which becomes expensive for large candidate lists (common in flocking/boid/pathfinding scenarios). A pure vectorized replacement regresses on small lists due to array allocation and conversion overhead. 

- **Expected:** close-by candidate chosen with random tie-breaking. 
- **Actual:** slow performance for large n; pure vectorization → overhead for small n.

### Implementation
`space.py`
- **Added:**

    - `closest_loop(self, agent, pos)` - the loop-based closest selection (preserves tie-handling and uses agent.random.shuffle/choice).
    - `closest_vectorized(self, agent, pos)` - NumPy-vectorized distance computation with torus handling, mask for ties, and final random tie selection using agent.random.

- **Modified:**

  - `[move_agent_to_one_of(...)]` — delegates to` closest_loop` or `closest_vectorized` using a small threshold (50 candidates) to balance conversion overhead vs vectorized speed.
  
**Key details**

- Vectorized path:

  - pos -> np.array(dtype=float); compute diffs; apply torus wrap via np.minimum; distances = sum(diffs**2, axis=1).
  - Identify indices where distances == min; pick one with agent.random.randrange(...).
  - Convert chosen row back to Python coordinate tuple.

- Loop path: unchanged logic encapsulated for clarity/testability.
- Threshold (50) chosen after microbenchmarks; easy to tune or expose as config.

### Additional Notes
No API changes or public surface modifications.

### DIFF
--- a/tests/test_space.py
+++ b/tests/test_space.py
@@ -527,27 +527,42 @@ class TestSingleGrid(unittest.TestCase):  # noqa: D101
         """Test hybrid approach uses loop for small position sets (<50)."""
         agent = self.agents[0]
         agent.pos = (5, 5)
-        possible_positions = [(5 + i, 5 + j) for i in range(-2, 3) for j in range(-2, 3)]
+        possible_positions = [
+            (5 + i, 5 + j) for i in range(-2, 3) for j in range(-2, 3)
+        ]
 
         self.space.move_agent_to_one_of(agent, possible_positions, selection="closest")
 
-        min_dist = min(self.space._distance_squared(p, (5, 5)) for p in possible_positions)
-        closest_positions = [p for p in possible_positions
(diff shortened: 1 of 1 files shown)


quaquel:
As stated in the docstring, `mesa.space` is maintenance-only. This PR does not fix a known or critical bug. So I am closing this PR.

PR #3112, opened 2026-01-11
Fix TypeError in find_combinations when evaluation_func is None

codebyNJ:
### Summary
Fixed a `TypeError` that occurs when calling `find_combinations()` without an `evaluation_func` parameter.

### Bug / Issue
In` meta_agent.py`, the `find_combinations()` function calls` evaluate_combination()` which returns `None` when evaluation_func is `None`. The code then attempts to unpack this `None` value:
```python
group_set, result = evaluate_combination(candidate_group, model, evaluation_func)
```
This causes: `TypeError: cannot unpack non-iterable NoneType object`

### Implementation
File changed: `mesa_agent.py`

**Changed From:**
```python 
group_set, result = evaluate_combination(candidate_group, model, evaluation_func)
if result:
    combinations.append((group_set, result))
```

**To:**
```python
evaluation_result = evaluate_combination(candidate_group, model, evaluation_func)
if evaluation_result is not None:
    group_set, result = evaluation_result
    if result:
        combinations.append((group_set, result))
```

### Testing

- Added test `test_find_combinations_without_evaluation_func` that validates the fix
- All 16 meta_agents tests pass
- ruff checks pass


quaquel:
I am not sure this is the right fix. The entire function is pointless if `evaluation_func is None`. So, my guess is that either `evaluation_func` should become an argument or calling it with None should raise a `ValueError`. @tpike3, can you clarify your intent here?

tpike3 (approved):
@codebyNJ Thanks for the catch and fix!

EwoutH:
> I am not sure this is the right fix. The entire function is pointless if `evaluation_func is None`. So, my guess is that either `evaluation_func` should become an argument or calling it with None should raise a `ValueError`. @tpike3, can you clarify your intent here?

I would still be interested in this.

PR #3113, opened 2026-01-11
pathfinding for `__dict__` in `Cell`

quaquel:
This PR removes `__dict__` and `@cached_property` from `Cell,` This is in line with #3108 and reduces the memory footprint of a grid. This works because only `@cached_property` requires `Cell.__dict__`. Properties, descriptors, but also `@cache` don't need `Cell.__dict__`, because they rely on storage at the class-level, not the instance-level. 

**Motivation for removing `__dict__`**
As indicated by @falloficarus22 in #3108, 

> Mesa environments often consist of millions of cells. Previously, the Cell class included `__dict__` in its`__slots__`, which resulted in an overhead of approximately 300 bytes per instance (due to the presence of an empty dictionary). For high-resolution grids (e.g., 1000x1000), this contributed ~300MB of unnecessary RAM usage. This bottleneck limited the scalability of complex spatial models.

It also replaces the `PropertyDescriptor` with a property factory pattern, as suggested by @codebreaker32 in #3080. But, I still need to include the `cell.empty` read-only content from #3080. 

1. What are the performance implications of this?

So far, it appears that there is a slight performance loss from removing `@cached_property`. This is largely due to the extra method call from the property to `get_neighborhood`, which is still cached. Shifting from `self.cell.neighborhood` to `self.cell.get_neighborhood()` avoids the indirection, which has a small performance benefit. For Mesa 4.0, we may consider removing `cell.neighborhood`, as we currently have two different ways of obtaining a neighborhood with a radius of 1 and the center excluded. 

Shifting from a descriptor to a property halves the time to access it. There might be room for further improvement here. 

2. `cell.empty` in main works because of `Grid` adding a property layer. However, as per #3080, empty is a bit of a weird case. `Cell.empty` is special. The cell itself should be able to write to it, but the user should not be able to. One option is to rename it to `_empty` to clearly signal that it is used internally and should be left alone by the user. In which case, we can still use property layers. Another option is to avoid using `self.add_property_layer` and instead work around it for `empty`. 

Regardless, a benefit of shifting from a descriptor to a property factory is that write-only access becomes trivial. The `property()` function takes as keyword arguments a getter, a setter, and a deleter. By omitting the setter, you have automatically created read-only access.

### DIFF
--- a/mesa/discrete_space/property_layer.py
+++ b/mesa/discrete_space/property_layer.py
@@ -47,6 +47,13 @@ class PropertyLayer:
         """Accessor for the data of the property layer."""
         return self._data
 
+    @data.setter
+    def data(self, value):
+        # FIXME, in mesa 4, we simply should not allow a setter, forcing people
+        #    to always use data[:]
+        #    this holds even if we drop PropertyLayers in favour of raw numpy arrays.
+        self._data[:] = value
+
     def __init__(
         self, name: str, dimensions: Sequence[int], default_value=0.0, dtype=float
     ):
@@ -248,7 +255,6 @@ class HasPropertyLayers:
                 layer._data, docstring=f"accessor for {layer.name}"
             ),
         )
-        # setattr(self.cell_klass, layer.name, PropertyDescriptor(layer))
         self.cell_klass._mesa_properties.add(layer.name)
 
     def remove_property_layer(self, property_name: str):
@@ -256,7 +262,6 @@ class HasPropertyLayers:
 
         Args:
             property_name: the name of the property layer to remove
-            remove_from_cells: whether to remove the property layer from all cells (default: True)
         """
         del self._mesa_property_layers[property_name]
         delattr(self.cell_klass, property_name)
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -36,7 +36,10 @@ class Agent[M: Model]:
         pos (Position): A reference to the position where this agent is located.
 
     Notes:
-          unique_id is unique relative to a model instance and starts from 1
+        Agents must be hashable to be used in an AgentSet.
+        In Python 3, defining `__eq__` without `__hash__` makes an object unhashable,
+        which will break AgentSet usage.
+        unique_id is unique relative to a model instance and starts from 1
 
     """
 
--- a/mesa/batchrunner.py
+++ b/mesa/batchrunner.py
@@ -202,15 +202,27 @@ def _model_run_func(
     run_id, iteration, kwargs = run
 
     model = model_cls(**kwargs)
-    while model.running and model.steps <= max_steps:
+    while model.running and model.steps < max_steps:
         model.step()
 
     data = []
 
-    steps = list(range(0, model.steps, data_collection_period))
-    if not steps or steps[-1] != model.steps - 1:
-        steps.append(model.steps - 1)
-
+    # Use the DataCollector's actual history to capture ALL data (including sub-steps)
+    try:
+        recorded_steps = model.datacollector._collection_steps
+    except AttributeError:
+        # Fallback for legacy models without _collection_steps
+        steps = list(range(0, model.steps, data_collection_period))
+        if not steps or steps[-1] != model.steps - 1:
+            steps.append(model.steps - 1)
+    else:
+        match data_collection_period:
+            case -1:
+                steps = [recorded_steps[-1]] if recorded_steps else []
+            case 1:
+                steps = recorded_steps
+            case _:
+                steps = recorded_steps[::data_collection_period]
     for step in steps:
         model_data, all_agents_data = _collect_data(model, step)
 
--- a/mesa/discrete_space/cell.py
+++ b/mesa/discrete_space/cell.py
@@ -48,6 +48,14 @@ class Cell:
         "random",
     ]
 
+    @property
+    def empty(self) -> bool:  # noqa: D102
+        return self._empty
+
+    @empty.setter
+    def empty(self, value: bool) -> None:
+        self._empty = value
+
     def __init__(
         self,
         coordinate: Coordinate,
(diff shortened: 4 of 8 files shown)


mesa/discrete_space/property_layer.py
```diff
@@ -405,17 +410,30 @@ def __setattr__(self, key, value):  # noqa: D105
                 super().__setattr__(key, value)
 
 
-class PropertyDescriptor:
-    """Descriptor for giving cells attribute like access to values defined in property layers."""
+def create_property_accessors(layer, docstring=None):
+    """Helper function for creating accessor for properties on cells."""
+
```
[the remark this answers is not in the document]
codebreaker32:
```python
def create_property_accessors(layer, read_only=False, docstring=None):
    ...

    if read_only:
        # Returns a property with NO setter
        return property(getter, doc=docstring)

    return property(getter, setter, doc=docstring)
```
Also I suggest we add a `read_only` argument to create_property_accessors. This gives us the capability to protect layers. However, we cannot apply `read_only=True` to the 'empty' layer just yet. `Cell.add_agent` and `Cell.remove_agent` currently write to `self.empty`

mesa/discrete_space/property_layer.py
```diff
@@ -405,17 +410,30 @@ def __setattr__(self, key, value):  # noqa: D105
...
+    def getter(self):
+        return layer.data[self.coordinate]
+
+    def setter(self, value):
+        layer.data[self.coordinate] = value
```
codebreaker32:
```suggestion
    _data = layer.data
    def getter(self):
        return _data[self.coordinate]
    def setter(self, value):
        _data[self.coordinate] = value
```

codebreaker32 (changes requested):
I think this will be faster

mesa/discrete_space/property_layer.py
codebreaker32: Also, I suggest we add a `read_only` argument to `create_property_accessors`. This gives us the capability to protect layers. However, we cannot apply `read_only=True` to the 'empty' layer just yet. `Cell.add_agent` and `Cell.remove_agent` currently write to `self.empty`. If we lock it now, those methods will crash.

```python
def create_property_accessors(layer, read_only=False, docstring=None):
    ...

    if read_only:
        return property(getter, doc=docstring)

    return property(getter, setter, doc=docstring)
```

quaquel: 1. I already tried versions of your suggestion. They reduce access time by about 25%. However, if you assign a new numpy array to data, it invalidates the property, so e.g., sugarscape errors out. So

```python
self.grid.my_layer.data = some_new_array
```

breaks everything. 

2. I agree on the read-only stuff, but I still want to figure out how to handle empty as part of it.

falloficaruss:
In `grid.py` we have
```python
self.cell_klass = type(
            "GridCell",
            (self.cell_klass,),
            {"_mesa_properties": set()},
        )
```
Since this `GridCell` subclass doesn't define __slots__, Python will automatically give it a __dict__, defeating the purpose.

quaquel:
> Since this GridCell subclass doesn't define slots, Python will automatically give it a dict, defeating the purpose.

I have updated the PR to include this change, as well as added a test for it (and fixed it in a few other places not covered in #3108).

mesa/discrete_space/property_layer.py
codebreaker32: Curious if we can do this..

```python
class PropertyLayer:
    def __init__(self, ...):
        self._data = np.full(...) 

    @property
    def data(self):
        return self._data

    @data.setter
    def data(self, value):
        # Enforce in-place update to preserve references
        self._data[:] = value
```

codebreaker32: I ran some simulation tests on my local machine with above changes:-

Setup:-
```python
GRID_SIZE = 1000        # 1000x1000 = 1 Million Cells
N_AGENTS = 10_000       # Simulating 10k agents    
N_STEPS = 50            
TOTAL_READS = N_AGENTS * N_STEPS   # 500,000
TOTAL_WRITES = 1000
```
Result(On main)
```plaintext
Grid: 1000x1000 | Agents: 10000
Simulating 500,000 total property reads.

1. WRITE PERFORMANCE (Layer Assignment)
   Time: 114.7 ns

2. READ PERFORMANCE (Agent Access)
   Total Time: 0.1216 sec
   Per Access: 243.3 ns
```

With these changes (on perf_cell)
```plaintext
Grid: 1000x1000 | Agents: 10000
Simulating 500,000 total property reads.

1. WRITE PERFORMANCE (Layer Assignment)
   Time: 0.35 ms

2. READ PERFORMANCE (Agent Access)
   Total Time: 0.0980 sec
   Per Access: 196.0 ns
```

quaquel: I'm curious about the exact test you ran. Can you share them?

I'm unsure about the setter you suggested. I am considering making setting data impossible. However, it at least gives us a direction to explore further.  Thanks for thinking along.

codebreaker32:
```python
import timeit
import numpy as np
import random
from mesa.discrete_space import OrthogonalMooreGrid

# SETUP
GRID_SIZE = 1000        # 1000x1000 = 1 Million Cells
N_AGENTS = 10_000       # Simulating 10k agents
N_STEPS = 50            # Simulate 50 steps
TOTAL_READS = N_AGENTS * N_STEPS
TOTAL_WRITES = 1000

print(f"Grid: {GRID_SIZE}x{GRID_SIZE} | Agents: {N_AGENTS}")
print(f"Simulating {TOTAL_READS:,} total property reads.")

grid = OrthogonalMooreGrid((GRID_SIZE, GRID_SIZE), torus=False, random=random.Random())
layer = grid.create_property_layer("temperature", default_value=0.0)

# Create raw data for write test
new_map = np.random.rand(GRID_SIZE, GRID_SIZE)

agent_coords = [
    (random.randint(0, GRID_SIZE-1), random.randint(0, GRID_SIZE-1)) 
    for _ in range(N_AGENTS)
]
# Get actual cell objects
agent_cells = [grid._cells[coord] for coord in agent_coords]

# WRITE SPEED
def test_write():
    grid.temperature.data = new_map

# Run 1000 times to average
write_time = timeit.timeit(test_write, number=TOTAL_WRITES) / TOTAL_WRITES

print(f"\n1. WRITE PERFORMANCE ")
if write_time < 0.000001:
    print(f"   Time: {write_time*1e9:.1f} ns")
else:
    print(f"   Time: {write_time*1e3:.2f} ms")



def test_read():
    for cell in agent_cells:
        _ = cell.temperature

# Run N_STEPS loops
read_total_time = timeit.timeit(test_read, number=N_STEPS)
avg_read_time_ns = (read_total_time / TOTAL_READS) * 1e9

print(f"\n2. READ PERFORMANCE (Agent Access)")
print(f"   Total Time: {read_total_time:.4f} sec")
print(f"   Per Access: {avg_read_time_ns:.1f} ns")

```

quaquel: quick update, I have added a getter along the lines suggested by @codebreaker32.

However, this also let me to discover a subtle bug in `PropertyLayer.from_data`, which was already present in main.

quaquel:
I split of the dict/slots stuff into its own stand alone PR. I now understand how it works and its a nice stand alone PR. 

I also split of the `PropertyLayer.from_data` bugfix and the property factory pattern into stand alone PR's. Once those are all merged, I'll return here and see what's left. the main things seems to be #3080/#3072.

PR #3120, opened 2026-01-12
Missing Type Validation in AgentSet Constructor

AnvayKharb:
This pull request improves the robustness of the `AgentSet` class by adding type validation to ensure that only `Agent` instances can be included. It also introduces comprehensive tests to verify this new behavior.

**AgentSet type validation:**

* Added a check in the `AgentSet` initializer to raise a `TypeError` if any item in the input is not an instance of `Agent`, including a descriptive error message listing up to 5 invalid items.

**Testing improvements:**

* Added a new test function `test_agentset_type_validation` to ensure that `AgentSet` correctly raises errors for invalid input types and accepts valid agents. This includes testing with all invalid items, mixed valid/invalid items, and checking the error message content.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -184,7 +184,7 @@ class AgentSet[A: Agent](MutableSet[A], Sequence[A]):
         """
         # Convert to list to validate items before creating WeakKeyDictionary
         agents_list = list(agents)
-        
+
         # Validate all items are Agent instances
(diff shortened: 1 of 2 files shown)


quaquel:
I find this extensive validation to be inconsistent with the fundamental principles of Python as a dynamic language, which also, in part, relies on duck typing. Moreover, if you start doing validation, it would make sense to do it also when adding agents to the set. So, I am closing this PR.

PR #3121, opened 2026-01-12
Ensure Cell only uses __slots__

quaquel:
### Summary
This PR ensures that Cell does not use an internal dict and only relies on `__slots__`. Relying on `__slots__` only has speed and memory benefits. This PR builds on #3108.

### Implementation
It updates Cell itself by removing `__dict__` from `__slots__` and also removes the `@cached_property` for `neighborhood. It also ensures that in grid, where we dynamically subclass `Cell` we include `__slots__` explicitly to avoid automagically adding `__dict__`. This change is also included when unpickling cells. Tests are added to ensure all is behaving as intended

### Additional Notes
Removing `@cached_property` has a small performance penalty, but this can be mitigated by using `get_neighborhood` directly.

Closing #3107

### DIFF
--- a/tests/discrete_space/test_discrete_space.py
+++ b/tests/discrete_space/test_discrete_space.py
@@ -1191,7 +1191,7 @@ def test_pickling_cell():
     """Test pickling of a Cell."""
     cell = Cell((1,), capacity=1, random=random.Random(42))
 
-    unpickled_cell = pickle.loads(pickle.dumps(cell)) # noqa: S301
+    unpickled_cell = pickle.loads(pickle.dumps(cell))  # noqa: S301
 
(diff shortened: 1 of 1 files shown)


EwoutH:
Looks great. Have you ran function/method level benchmarks for this?

quaquel:
Yes, the `neighborhood` property becomes slightly slower, but it's in the nanoseconds and, as seen by the benchmarks, does not matter in the bigger scheme of things.

EwoutH:
Would you like to make `performance` a distinct release-notes label, so these PRs are listed in a separate entry in the release notes?

quaquel:
I'll leave that for you to decide. I don't recall the exact details of the release cycle. I am fine either way.

PR #3122, opened 2026-01-12
Ensure PropertyLayer.from_data() does not have side effects

quaquel:
Currently, ' PropertyLayer.from_data()` uses the numpy array directly. However, this means that any subsequent changes to `layer.data` will also change the original dataframe. This can be a bug (currently affecting Sugarscape). An additional test for this bug is also included.


EwoutH (approved):
Thanks!

PR #3125, opened 2026-01-12
Replace PropertyDescriptor with properties

quaquel:
This PR replaces the `PropertyDescriptor` with normal properties using a property factory pattern. This PR follows up on a suggestion from @codebreaker32 in #3080. It roughly halves the access time to a property in a cell. 

**Implementation details**
1. It uses the `property()` function to assign getters and setters for a property to a cell. 
2. It also encapsulated `layer.data` by wrapping it in property. The setter is most critical here because it ensures that `self._data` is never changed to a different NumPy array, thereby breaking the properties assigned via step 1. 
3. I currently don't add a deleter to the property because I fail to see a use case
4. A future PR can make layers read-only at the cell level by omitting the setter.

### DIFF
--- a/mesa/discrete_space/property_layer.py
+++ b/mesa/discrete_space/property_layer.py
@@ -48,7 +48,7 @@ class PropertyLayer:
     @data.setter
     def data(self, value):
         # FIXME, in mesa 4, we simply should not allow a setter, forcing people
-        #    to allways use data[:]
+        #    to always use data[:]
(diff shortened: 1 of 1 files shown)


EwoutH:
Either the current `PropertyDescriptor` and the new `create_property_accessors` aren't and won't be used directly by users, right?

quaquel:
Yes they are purely internal. As you can see, they don't show up in the `__all___` for `mesa.discrete_space`

quaquel:
See #3113 for some numbers on performance.

Issue #3132, opened 2026-01-14
Mesa 4 development tracking issue


EwoutH:
_This is a  **tracking issue**. It's mean for short, concise status updates to keep track of progress, not for extensive discussions. [Discussions take place here](https://github.com/mesa/mesa/discussions)._

The major discussion on Mesa 4.0 have crystalized and found enough alignment to start development. This is an opportunity for breaking changes to address core architectural limitations and make some clean sheet designs.

Based on discussions in:
- https://github.com/mesa/mesa/discussions/2972

Maintainers and collaborators, feel free to update directly.

## Committed 
Goals with commitment from the maintenance team.

### 1. Unified event scheduling and time progression
- [x] Single `model.time` as universal time source #2903
- [x] `EventGenerator` class for recurring event patterns #3201
- [x] Unify event list between `Model` and `Simulator` #3204
- [x] Refactor step scheduling to use `EventGenerator` internally #3260
- [x] Add public `run_for()`, `run_until()`, `schedule_event()`, `schedule_recurring()` to `Model` #3266
- [x] Migrate examples and tutorials from `model.step()` to `model.run_for()` #3270
- [x] Update wolf-sheep example to use new event scheduling API #3278
- [x] Stabilize event scheduling system from experimental to `mesa.time` #3276
- [x] Deprecate `Simulator`, `ABMSimulator`, and `DEVSimulator` classes #3277
- [ ] Add deprecation warnings to `model.step()` and `run_model()`
- [x] Rewrite agent activation, event scheduling and time progression tutorials #3280
- [x] Remove `model.steps` #3328
- [x] Enforce monotonic time in event scheduling #3343
- [x] Add `pause()` and `resume()` to `EventGenerator` #3431
- [x] Add `next_scheduled_time` property to `EventGenerator` #3423
- [x] Add `compact()` to `EventList` for cancelled event cleanup #3359
- [ ] Remove remaining deprecated APIs (e.g., `Simulator` classes)

Open discussions:
1. Either exposing _default_schedule as a public property (allowing users to modify/disable the default step pattern) or moving away from a default scheduled method completely: [Does Mesa need a "heartbeat"?](https://github.com/mesa/mesa/discussions/2921#discussioncomment-15591609)
2. [Conditional running and stopping API](https://github.com/mesa/mesa/discussions/2921#discussioncomment-15614353).

Related:
- #2921, #2585, #2967

Coordinating lead: @EwoutH

### 2. Space Architecture
[...]

PR #3133, opened 2026-01-14
Target main branch for Mesa 4 development

EwoutH:
Bump version from `3.5.0.dev` to `4.0.0.dev`.

Before merge, a `3.x` branch will be split of.

Part of #3132.

### DIFF
--- a/mesa/__init__.py
+++ b/mesa/__init__.py
@@ -24,7 +24,7 @@ __all__ = [
 ]
 
 __title__ = "mesa"
-__version__ = "3.5.0.dev"
+__version__ = "4.0.0.dev"
 __license__ = "Apache 2.0"
 _this_year = datetime.datetime.now(tz=datetime.UTC).date().year
 __copyright__ = f"Copyright {_this_year} Mesa Team"


quaquel:
when do you plan to merge this and what do you still want to get into 3.5?

EwoutH:
Now that I know the new time API works with current Mesa (amazingly even with batch_run), we can postpone this a bit, and do a 3.5 release first.

Even for Spaces we don’t really have to go to 4.0, since continuous is still experimental and thus can be revised to support discrete spaces underneath.

I think we can be proud on how flexibly we architected Mesa and our development processes.

At some point a cleanup becomes beneficial to remove deprecated features. Might be once we’ve figured out data collection, experimental control and signals.

quaquel:
Fair enough. 

However, it does mean that some of the property layer stuff is stuck for now. Ideally, I would first remove the property layer class itself before moving on. Although it might be possible to move forward without it.

I think we are also close to having at least an experimental alternative to data collection.

EwoutH:
Yeah, and it's also time for a clean-up.

Let's release 3.5 when the most important deprecations and time API are in. Quite a substantial release, also with the scenarios. Then we split of to 4.0 and can continue with a clean slate.

quaquel:
I guess this is redundant now?

EwoutH:
Yes, done in https://github.com/mesa/mesa/commit/0c8b5ce32f7ecc606a0a2384bd914a5ff922c5b8

PR #3134, opened 2026-01-14
Remove batchrunner

EwoutH:
Remove batchrunner module, `batch_run` function and associated docs and tutorials.

The number of modified or removed files demonstrate how deeply `batch_run` was integrated in Mesa.

Part of Mesa 4 development (#3132).


quaquel:
It would be good to track which tutorials are modified so we know which ones we need to return to. Otherwise, this seems fine to me.

quaquel:
I suggest we close this and just do it again. Probably easier than fixing all conflicts.

EwoutH:
Agree

PR #3136, opened 2026-01-14
Deprecate `model.steps` in favor of `model.time`

EwoutH:
_This PR was reverted in #3151 due to advancing insight._

### Summary
Deprecates `model.steps` with a `FutureWarning` directing users to use `model.time` instead. The attribute will be removed in Mesa 4.0.

### Motive
Following the introduction of `model.time` as the universal source of truth for simulation time in #2903, `model.steps` is now redundant.

### Implementation
- Rename `self.steps` to `self._steps` (internal attribute)
- Add `steps` property with getter and setter that:
  - Emit `FutureWarning` on access
  - Delegate to `self._steps`
- Update `_wrapped_step()` to use `self._steps` directly (avoiding deprecation warning in internal code)
- Update docstring to remove `steps` from the public attributes list
- Add migration guide entry

`batch_run` changes are in a separate commit, for an easy cherry pick if needed.

### Usage
Users will see this warning when accessing `model.steps`:

```
FutureWarning: model.steps is deprecated and will be removed in Mesa 4.0.
Use model.time instead.
See: https://mesa.readthedocs.io/latest/migration_guide.html#model-steps-deprecated
```

### Migration Path
```python
# Old
if model.steps >= 100:
    model.running = False

# New
if model.time >= 100:
    model.running = False
```

### Checklist
- [x] Documentation updated to use `model.time`
- [x] Migration guide entry added
- [x] Example models updated to use `model.time`
- [x] Tests updated to use `model.time` (or suppress expected warnings)

### Related
- Follows #2903

### DIFF
--- a/docs/migration_guide.md
+++ b/docs/migration_guide.md
@@ -3,7 +3,29 @@ This guide contains breaking changes between major Mesa versions and how to reso
 
 Non-breaking changes aren't included, for those see our [Release history](https://github.com/mesa/mesa/releases).
 
-## Mesa 3.4.0
+## Mesa 3.5
+### model steps deprecated
+`model.steps` is deprecated and will be removed in Mesa 4.0. Use `model.time` instead.
+
+Previously, time in Mesa was fragmented: simple models used `model.steps` as a proxy for time, while discrete event simulations stored time separately in `simulator.time`.
+`model.time` is now the single, canonical source of simulation time in Mesa.
+
+For most use cases, simply replace `model.steps` with `model.time`. Since `model.time` is a float that increments by 1.0 per step by default, the numerical values will be equivalent.
+
+```python
+# Old
+current = model.steps
+
+# New
+current = model.time
+
+# If you need an integer step count, cast to int
+current = int(model.time)
+```
+
+- Ref: [PR #3136](https://github.com/projectmesa/mesa/pull/3136), [PR #2903](https://github.com/projectmesa/mesa/pull/2903)
+
+## Mesa 3.4
 
 ### batch run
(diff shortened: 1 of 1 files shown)


quaquel:
I agree with this. 

What else to deprecate in 3.5? At least `seed` on model, probably `agent.pos`, `mesa.space`, `batch_run`, and `datacollector.

EwoutH:
My suggestion would be:
- Let's deprecate everything now that we already have a (stable) replacement for
- Wait a bit with deprecations if we don't have a replacement yet

quaquel:
But then its a bit strange that you start removing stuff like the datacollector

EwoutH:
Yeah, I didn’t make this clear indeed:
- Before branching of 3.x maintenance, let’s deprecate everything we already got a replacement for. Those deprecations will be included in a 3.5 release.
- After that, we can deprecate/remove everything else, since that will only be included in 4.0.

quaquel:
So that does mean at least `seed` on Model and `mesa.space`. `seed` is replaced with `rng` in line with spec-7. `mesa.space` is fully replaced. We probably then should also move the continuous space out of experimental and make it stable?

EwoutH:
1) yes
2) I’m a bit afraid we’re going to ask people to move twice, considering we’re continuing development on the ContinuousSpace. So maybe backport a deprecation once we know that and how the final API works.

As soon as we have branched of 3.x, we can move ContinuousSpace out of experimental. Since we’re then in 4.0 alpha, it isn’t immediately API frozen, but it will be after 4.0 stable.

Issue #3131, opened 2026-01-14
mesa signal improvements


quaquel:
Mesa has support for reactive programming via `mesa.experimental.mesa_signals`. Upon reviewing this code again, there may be room for some improvement.

1. Replace AttributeDict with a (frozen) dataclass.
2. The current implementation is designed around the use of descriptors. What about rewriting this to use the property factory pattern recently introduced in property layers?

While exploring these changes, it is crucial to explicitly assess their performance implications. When signals were added, it was seen as a potential building block for much more flexible data collections. Early benchmarks yielded mixed results from a runtime perspective. Some stuff became faster, other stuff became slower because of the signal overhead.

kundan223:
hiu @quaquel  ,  I  am a new contributor it might take me 1-2 days to work on it but can i take this issue ?

quaquel:
Feel free to take a crack. As indicated, some micro benchmarking is a critical part of this. It might be good to first post some results and ideas here before opening a PR. This issue, in part, is figuring out a direction rather than just implementing a predefined solution.

quaquel:
Another thing worth checking out is how signals would function in an async context. If well designed, data processing and storage can be handled separately from the execution of the main model.

ShreyasN707:
@quaquel Just wanted to clarify myself For the micro-benchmarking you mentioned, are there specific usage patterns you consider most representative?
For example:
– high-frequency attribute updates
– many agents subscribing to the same signal
– frequent connect/disconnect cycles

codebreaker32:
I have an approach to move from a class-based descriptor pattern (`Computable`/`Computed`) to a standard Python property factory (`@computed`). 

Currently, accessing a derived property triggers the `Computable.__get__` descriptor, which then fetches a separate `Computed` object to execute the logic. This adds unnecessary function call depth to every attribute access.

I suggest two streamlined components:

1. **`@computed` Decorator:** A factory that returns a standard Python `property`. The "reactive" logic (dependency tracking, dirty checking) is captured in the property's closure, eliminating the descriptor lookup overhead.
2. **`ComputedState`:** A lightweight, internal-only class using `__slots__` to hold the value, dirty flag, and dependencies.

### API Changes (Before vs. After)

**Before (Current)**

```python
class MyAgent(HasObservables):
    wealth = Computable()  # Descriptor

    def __init__(self):
        super().__init__()
        # Manual binding required
        self.wealth = Computed(self._calculate_wealth)

    def _calculate_wealth(self):
[...]

PR #3138, opened 2026-01-15
replace AttributeDict with a dataclass

quaquel:
### Summary
Replace the generic `AttributeDict` helper class with a dedicated frozen dataclass `Message` for improved type safety and developer experience when working with Mesa signals.

### Bug / Issue
The current signal implementation uses `AttributeDict`, a generic dictionary with attribute-like access. This has several limitations:
- No IDE autocomplete support when writing signal handlers
- No type safety - typos in field names fail at runtime
- Less clear intent - the structure of signal data is implicit
- Inconsistent field access patterns

Part of #3131 (addresses item 1)

### Implementation
Replaced `AttributeDict` with a frozen dataclass `Message` in the mesa_signals module:

1. New `Message` dataclass in `signals_util.py` with explicit fields: `name`, `old`, `new`, `owner`, `signal_type`, `additional_kwargs`
2. Updated type hints throughout `mesa_signal.py` to accept `str | SignalType | All` where appropriate
3. Changed `signal.type` → `signal.signal_type` (more explicit, avoids Python keyword)
4. Additional kwargs now stored in `additional_kwargs` dict rather than spread into the signal object

Local benchmarking shows no measurable performance difference.

### Testing
All existing tests pass with updated field names. Type checking and IDE autocomplete now work correctly for signal handlers:
```python
def on_change(signal: Message):  # Type hint enables autocomplete
    assert signal.signal_type == SignalType.CHANGE
    assert signal.old == 10
```

### Additional Notes
This is a breaking change for experimental code:
- `signal.type` → `signal.signal_type`
- Custom kwargs now in `signal.additional_kwargs` dict

Since `mesa_signals` is experimental, breaking changes are allowed per Mesa's policy.

### DIFF
--- a/mesa/experimental/mesa_signals/signals_util.py
+++ b/mesa/experimental/mesa_signals/signals_util.py
@@ -23,16 +23,17 @@ __all__ = [
 if TYPE_CHECKING:
     from mesa.experimental.mesa_signals import ListSignalType, SignalType
 
+
 @dataclass(frozen=True, slots=True)
 class Message:
     """A message class containing information about a signal change."""
+
     name: str
     old: Any
     new: Any
     owner: Any
     signal_type: SignalType | ListSignalType
     additional_kwargs: dict
-
 
 
 def create_weakref(item, callback=None):
--- a/mesa/experimental/mesa_signals/mesa_signal.py
+++ b/mesa/experimental/mesa_signals/mesa_signal.py
@@ -23,9 +23,9 @@ import functools
 import weakref
 from abc import ABC, abstractmethod
 from collections import defaultdict, namedtuple
-from collections.abc import Callable, Generator
+from collections.abc import Callable
 from enum import Enum
-from typing import Any, Generator
+from typing import Any
 
 from mesa.experimental.mesa_signals.signals_util import Message, create_weakref
 


EwoutH:
Quick API check: Are any of these used by downstream users or part of the public API?

1. Direct attribute access on signals for extra kwargs (e.g., `signal.index` vs `signal.additional_args["index"]`)
2. `AttributeDict` class itself (was exported in `__all__`)
3. Custom `__dir__` behavior for introspection/autocomplete on signal objects
4. The `type` field as a string vs `SignalType` enum

If any are user-facing, I think we need to document that properly.

mesa/experimental/mesa_signals/signals_util.py
```diff
@@ -10,37 +10,25 @@
...
 __all__ = [
-    "AttributeDict",
+    "Message",
     "create_weakref",
 ]
 
 
-class AttributeDict(dict):
-    """A dict with attribute like access.
+@dataclass(frozen=True)
```
EwoutH: I think `slots=True` could improve memory usage and attribute access speed for a frequently-instantiated message class:
```suggestion
@dataclass(frozen=True, slots=True)
```

mesa/experimental/mesa_signals/signals_util.py
```diff
@@ -10,37 +10,25 @@
...
-    def __dir__(self):  # noqa: D105
-        # allows us to easily access all defined attributes
-        names = dir({})
-        names.extend(self.keys())
-        return names
+    name: str
+    old: Any
+    new: Any
+    owner: Any
+    type: str
```
EwoutH: Could `type` use the `SignalType` enum directly?
```suggestion
    type: SignalType
```

quaquel:
> Quick API check: Are any of these used by downstream users or part of the public API?

Its in experimental and we don''t even use it in our own examples at the moment. So I would not worry about it too much.

> Custom __dir__ behavior for introspection/autocomplete on signal objects

Not sure what you mean, can you elaborate?

> The type field as a string vs SignalType enum

Yes, I merged that PR after opening this one, so I'll fix that

mesa/experimental/mesa_signals/signals_util.py
quaquel: This is a bit tricky. I have now type-hinted it with both enums for signals: SignalType and ListSignalType, but since you cannot cleanly subclass enums, I will need to find a more generic solution at some point.

Issue #3130, opened 2026-01-13
De-duplicate PropertyLayer


EwoutH:
The PropertyLayer class is now duplicated among:

- https://github.com/mesa/mesa/blob/14a94c93b863b25f6e225f61d0f88ed157e96667/mesa/space.py#L592
- https://github.com/mesa/mesa/blob/14a94c93b863b25f6e225f61d0f88ed157e96667/mesa/discrete_space/property_layer.py#L30

With enhancements being made to the discrete_space version (i.e. #3122 #3125), this will get out of sync sooner or later.

I think this is an artifact of the discrete space being developed in the cell space. If the two currently functionally different, we should discuss how we can structure that with the least amount of duplication.

quaquel:
Since `mesa.space` is maintenance only, I would not worry about it.

EwoutH:
What was the original motivation for full duplication or the PropertyLayer class? Did (extensive) changes need to take place to integrate it with the discrete space?

quaquel:
The discrete space was still experimental, but to move it out of it, you wanted feature parity with the existing property layers.

PR #3142, opened 2026-01-15
Refactor PropertyLayer to reduce code duplication

falloficaruss:
## Description
This PR addresses Issue #3130 by de-duplicating the `PropertyLayer` class found in `mesa/space.py` and `mesa/discrete_space/property_layer.py`.

The `mesa.space.PropertyLayer` now inherits from `mesa.discrete_space.property_layer.PropertyLayer`. This ensures that core logic is centralized while maintaining backward compatibility for the legacy API.

## Changes
- `mesa/space.py`:
    - Imported `PropertyLayer` from `mesa.discrete_space.property_layer` as `NewPropertyLayer`.
    - Refactored `PropertyLayer` to inherit from `NewPropertyLayer`.
    - Adapted `__init__` to map `width` and `height` to the new `dimensions` tuple argument.
    - Implemented/overrode legacy methods (`set_cell`, `modify_cell`, `modify_cells`, `aggregate_property`) to map to the new implementation or delegate appropriately.
    - **Behavior Change**: The legacy `PropertyLayer` now adopts the `data` setter behavior of the new implementation, which enforces in-place updates (`[:] = value`) rather than object replacement. This is a safer behavior that prevents breaking references held by the Grid.

## Verification
- Verified legacy API compatibility (initialization, `set_cell`, `modify_cell`, etc.) with a test script.
- Verified that `data` property assignment now updates in-place.

### DIFF
--- a/mesa/space.py
+++ b/mesa/space.py
@@ -589,8 +589,8 @@ def ufunc_requires_additional_input(ufunc):  # noqa: D103
     return ufunc.nargs > 1
 
 
-
 from mesa.discrete_space.property_layer import PropertyLayer as NewPropertyLayer
+
 
 class PropertyLayer(NewPropertyLayer):
     """A class representing a layer of properties in a two-dimensional grid.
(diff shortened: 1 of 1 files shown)


quaquel:
mesa/space.py is maintenance only and about to be fully removed.

PR #3143, opened 2026-01-15
Cell.remove_agent shift to swap-remove

quaquel:
Cell._agents is a list. Removing agents from lists can be slow. So this is quick check to see if a swap remove is faster. In short, you look up the index of the agent to remove (O(n)), swap this agent with the last agent in the list, and pop off the last agent (O(1)), instead of shifting all agents after index up. Local testing suggests a modest speed-up in some cases.


quaquel:
so no major difference or even a slight loss on wolf sheep. Not worth the increase in code complexity.

PR #3145, opened 2026-01-15
Add `DataRecorder` for reactive Data Storage and `DatasetConfig` for Configuration

codebreaker32:
This PR introduces a decoupled, event-driven data collection architecture designed to handle large-scale simulations efficiently. It separates **what to collect** (`DataRegistry` from #3156 ) from **how to store it** (`DataRecorder`).

The core design philosophy is the separation of concerns:

1. What to collect is defined by the DataRegistry (from #3156).
2. When to collect is controlled by DatasetConfig (intervals, windows, start/end times). However there is an ongoing discussion about replacing it with `Schedule`( see https://github.com/mesa/mesa/discussions/1944#discussioncomment-15766733, https://github.com/mesa/mesa/discussions/2921#discussioncomment-15562881 and #3250)
3. How to store it is handled by DataRecorder implementations (Memory, SQL, Parquet).

#### **A. `BaseDataRecorder` (Abstract Base Class)**

Handles the orchestration logic: subscription, interval checks, and lifecycle management.

* `__init__(model, config=...)`: Attaches to the model.
* `collect()`: Manually triggers data collection.
* `get_table_dataframe(name)`: Retrieves data as a Pandas DataFrame.
* `summary()`: Returns stats and summary.

#### **B. `DatasetConfig` (Configuration)**

Fine-grained control over *when* data is collected for each dataset.

* `interval`: Collection frequency (e.g., every 1 time unit or every 0.5/2 time unit(s)).
* `start_time` / `end_time`: Define a specific collection window. (start_time=0 explicitly means to collect initial state)
* `window_size`: Rolling buffer capacity (e.g., keep only the last 1000 snapshots).

**Example Usage:**

```python
config = {
    "agents_wealth": DatasetConfig(interval=10, window_size=500),
    "model_gini": DatasetConfig(start_time=100)
}
recorder = DataRecorder(model, config=config)
# recorder = JSONDataRecorder(model, config=config)

```

### **Included Recorders**

This PR includes four implementations catering to different use cases:

i) **`DataRecorders` (Default/Memory):**
ii) **`SQLDataRecorder` (SQLite):**
iii) **`ParquetDataRecorder` (Parquet):**
iv) **`JSONDataRecorder` (JSON):**

### **Extensibility: Writing New Backends**
The architecture makes adding new storage backends (e.g., MongoDB, HDF5, CSV) straightforward. Developers only need to inherit from BaseDataRecorder.
All interval logic, validation, and observable subscriptions are handled automatically by the base class.

### Migration

Old
```python
class MyModel(Model):
    def __init__(self):
        ...

            model_reporters={"gini": self.compute_gini},
            agent_reporters={"Wealth": "wealth"}
        )

    def step(self):
        ...          # Step Logic
```

New 
```python
class MyModel(Model):
    def __init__(self):
        ...

        config = {
            "Gini": DatasetConfig(start_time=5, end_time=100, interval=1),
            "Wealth": DatasetConfig(start_time=0, end_time=100, interval=1)
        }

        self.recorder = DataRecorder(self, config=config)
        # self.recorder = SQLDataRecorder(self, config=config) # Easy backend swap

    def step(self):
        ...           # Step Logic
```

Note: Please refer to #3156 for details on the registration API

### Known Limitation

1. With current recorder, capturing the initial state ($t=0$) is non-trivial. If the Recorder is initialized before agents are created, it captures the initial state ($t=0$) immediately, but that state is empty/incomplete. If the Recorder relies solely on the time=0 signal emitted by `Model.__init__`, it will miss the event entirely because the signal fires before the user's `__init__` logic even begins.

To ensure the recorder captures the fully populated initial state, the BaseDataRecorder performs an immediate check upon instantiation. This imposes a strict ordering requirement on the user: The DataRecorder must be instantiated as the absolute last step of the `Model.__init__` method.

This is being solved in #3284.

2. We currently lacks an explicit lifecycle event to signal that a simulation run has finished. The recorder cannot automatically detect when is simulation going to end and we might miss the last step.

Workaround: Users must manually invoke `recorder.finalize()` at the end of their execution block.

Note: Future work is expected to provide standard `RUN_STARTED` and `RUN_ENDED` events, which will automate this process( see https://github.com/mesa/mesa/discussions/2921#discussioncomment-15720005)

### DIFF
--- a/mesa/time/events.py
+++ b/mesa/time/events.py
@@ -165,6 +165,16 @@ class Schedule:
     end: float | None = None
     count: int | None = None
 
+    def __post_init__(self):
+        """Validate schedule parameters."""
+        if not callable(self.interval) and self.interval <= 0:
+            raise ValueError(f"Schedule interval must be > 0, got {self.interval}")
+
+        if self.count is not None and self.count <= 0:
+            raise ValueError(
+                f"Schedule count must be > 0 if provided, got {self.count}"
+            )
+
 
 class EventGenerator:
     """A generator that creates recurring events based on a Schedule.
--- a/tests/test_schedule_run.py
+++ b/tests/test_schedule_run.py
@@ -195,3 +195,20 @@ class TestScheduleRecurring:
         model.schedule_recurring(record, Schedule(interval=1.0, start=1.0, count=3))
         model.run_for(10)
         assert len(log) == 3
+
+
+# --- schedule validation guards ---
+class TestScheduleValidation:
+    def test_rejects_nonpositive_fixed_interval(self):
+        with pytest.raises(ValueError):
+            Schedule(interval=0)
+
+        with pytest.raises(ValueError):
+            Schedule(interval=-1)
+
+    def test_rejects_nonpositive_count(self):
+        with pytest.raises(ValueError):
+            Schedule(interval=1.0, count=0)
+
+        with pytest.raises(ValueError):
+            Schedule(interval=1.0, count=-5)


codebreaker32:
I have modified the only examples that were used for benchmarking to compare timings.

mesa/model.py
```diff
@@ -84,12 +85,15 @@ def __init__(
...
+        """
+        # 1. Initialize Reactive System (REQUIRED for Listeners)
+        HasObservables.__init__(self)
 
-        Notes:
-            you have to pass either seed or rng, but not both.
+        # Register Standard Signals that Listeners subscribe to
+        self._register_signal_emitter("step", {"step"})
+        self._register_signal_emitter("reset", {"reset"})
+        self._register_signal_emitter("end", {"end"})
```
quaquel: 1. These have to be aligned with the new run-control stuff that @EwoutH is working on.
2. We just added a SignalEnum ListSignalEnum, it would be good to put these also into some kind of Enum as well, instead of using strings. I expect that in Mesa 4, we won't allow strings anymore
3. You now use a protected method. Would it not make more sense to instead make `model.time` observable in some sense (this overlaps with point 1)

mesa/model.py
```diff
@@ -137,6 +141,7 @@ def __init__(
         self.scenario = scenario
 
         # Wrap the user-defined step method
+        # This is CRITICAL: It ensures the 'step' signal is emitted for the Listener
```
quaquel: hence my point about making `model.time` observable, because then the signals are emitted automatically.

quaquel:
There is a lot here that I appreciate.

1. I like the explicit use of signals for triggering stuff. However, I think this can be improved and expanded a bit more as indicated in my comments on the code
2. I like how much faster it is.

However, this PR also highlights a bunch of open questions (which is another reason why I like it as a pathfinding PR).
1. The current approach, both here and in the data collector, is to just accumulate data using some kind of dict of lists. However, if one can preallocate an empty data structure of the appropriate size, storage will be faster. So, what if the new `RunControl` idea includes a default `end_time`? Then you can just preallocate stuff. 
2. The current design is a hybrid between a purely reactive push design and the current pull design. It is reactive to updates of model.time/model.step. It pulls the data to be collected, however, from the agent/model. I would be interested in exploring making this second step more reactive (does not have to be done in this PR). The benefit is that model-level aggregate statistics often require this anyway (e.g., Gini requires a table of agent wealth).
3. The current data collector has gotten increasingly complicated over time. In part to cover all kinds of weird use cases. It might be helpful to step back and clarify the various use cases for data collection. 

- We want to collect agent-level data over time. This might be split by agent type, or by any set of agents. And this should also account for the addition or removal of agents over time. (Currently split over agent_reporter and agent_by_type reporters)
- We want to collect agent data at random points in time. For example, agent lifetime data is collected whenever an agent is removed. This use case is handled by the table stuff
- We want to collect aggregate data at the model level (i.e., model reporters)
- Are there any more?

4. What about very explicitly constraining the user on the data collection that is supported by Mesa? For example, what about only allowing "simple" data like numbers?
5. I still think we need to more explicitly separate two distinct things: the observable state of the overall model at any given point in time, and the tracking of this state over time with (fixed) intervals. Any observable model or agent reporter is part of the observable state of the model. The newly added signals in this PR trigger, in essence, a snapshot of this state at a given point in time. Tables, however, are used for things that are not snapshotted in the same way.

mesa/model.py
codebreaker32: 2. Agree

1 & 3
While making model.time observable is cleaner for standard ABMs, it presents a major issues for the Discrete Event (DEVS) support in Mesa:

i. Simultaneous Events: In DEVS, multiple logical events can occur at the exact same timestamp ($t$). If we only listen for changes in model.time, the collector will miss subsequent events occurring at the same $t$ (since the value didn't change).
For eg.
Imagine an agent's bank balance starts at $100.
$t=10.0$ (Event A): Agent receives a paycheck (+$1000$). Balance = $1100.
$t=10.0$ (Event B): Agent pays rent (-$1000$). Balance = $100.$t=11.0$
(Next Step): Time advances.

Result: We missed the intermediate state entirely. Or worse, if we trigger before the events process, we might just see $100 \to 100$ and never know the agent had $1100 for a split second.


It will be difficult to make it consistent. In my opinion, as soon as I will get to see @EwoutH's run-control implementation, it might be easy for me to align with it. As currently, the challenge was me to make it align to current examples so that I can see benchmarks and decide its whether good to move forward or not.

quaquel: I agree. But this highlights an implicit assumption that we must make explicit. What is the frequency with which we want to take snaphots (see my overarching thoughts below) of the model state? The current datacollector leaves that up to the user with its `.collect()` call. You force it to be on each whole integer increment/step. What if the user can just specify this explicitly?

quaquel: Perhaps another way to make the same point: we need to explicitly separate the tracking of the model's observable state from the taking of snapshots of this state. 

In a way, this PR focuses primarily on the latter, but it does not give the user explicit control over the frequency of these snapshots.

codebreaker32:
1. Agree but this depends entirelt on `RunControl` API
2. In Python, making every attribute assignment reactive (using __setattr__ or Traitlets) adds massive overhead. It can slow down the simulation physics by 10x-50x. Push-Trigger / Pull-Data (Current Hybrid) is the sweet spot. We wait for the step to finish (Trigger), then use C-speed vectorization to grab the data (Pull). True "Reactive Data" is likely too slow for Python ABMs.
3.  Before opening this PR, I have done extensive research(explored EMAWorkbench, mesa-geo, some heuristic algorithms like ACO, ABC etc.) and explored mesa `datacollector.
For any ABM, there are three major categories(as you mentioned)
i. Agent Time-Series; State of N agents at time 't'
ii. Event Logs: Sparse event at random times
iii. Model Aggregates: Global stats

One more category that I find fits this is
iv. Spatial/Field Data: Recording the state of the environment itself, e.g., a vegetation layer in a grid, a pheromone trail in an ant colony, or raster data in Mesa-Geo.
Why it doesn't fit any of those?
- It isn't a single Model Aggregate (it's a matrix).
- It isn't strictly Agent Data (storing a 1000x1000 grid as 1 million agent rows is incredibly inefficient).
- It represents a "Dense State" that is best stored as a Tensor or Raster (2D/3D Array) rather than a list of records.
4. I liked this approach but one more thing we might do is to let user decide. For most of the users "simple" data may be enough but for sophisticated users(like developers, researchers etc.), we should give them an option to override or inherit from our Listener to create their own
5. I fully endorse this architectural split. It clarifies the role of the new listener:

- Signals (STEP/RESET): These act as the "Camera Shutter" for taking consistent, synchronized Snapshots of the Model's Observable State (model_vars / agent_vars).
- Tables: These remain the dedicated mechanism for Event Logging (asynchronous events like agent death or interactions) that are sparse and don't fit the snapshot paradigm.
- By formalizing the ModelSignal, we are explicitly decoupling the "Trigger" (The Shutter) from the "Collection" (The Film), which is what allows us to swap in the high-performance columnar engine without changing the model logic. Also this new approach makes collector free to implement any logic 

Update: Added the implementation checklist

mesa/model.py
codebreaker32: I will dig deeper into it

codebreaker32:
One more thing that I would like to ask is should I work on `mesa_signals` making more performative(because this PR relies on it) or is it covered by @EwoutH ?

quaquel:
> In Python, making every attribute assignment reactive (using setattr or Traitlets) adds massive overhead

I agree, but that is why I called it the observable state. Clearly not all attributes should be tracked automatically. It's a concious choice by the modeller to indicate what she want's to track. Also, if designed well, the overhead is not that big. I ran tests on mesa a while ago using Boltzman. I had a table with agent.wealth. This was subscribed to updates of agent.wealth. And then model.gini used this table instead of looping over all agents. Depending on implementation details, this was faster than the current data collector approach. 

> For any ABM, there are three major categories

I agree with the three categories and your additional point on matrices/tensors. The problem in Mesa at the moment is that we lack stand alone support for all of this and instead it is all integrated into the data collector. Conceptually, we might need to have a model.statistics style object. User can define data tables on this which belong to any of four types you specified. Next, the data collector can take snapshos of any of the defined tables and store those in some performant back end (pd.dataframe, polars, database, etc.). We might even have an api given the user more fine grained control over which "data tables" are snapshotted when. So some might be snapshotted on every tick, others might be snapshotted only on some kind of run ended signal. And we might even have a dedicated signal that users can fire if they want to snapshot something. 

The open question then is whether these "data tables" have to be reactive or can be reactive. We might just support both and leave it to the user to decide which is appropriate for their use case.

> One more thing that I would like to ask is should I work on mesa_signals making more performative(because this PR relies on it) or is it covered by @EwoutH ?

I built the original version of mesa_signals and have been looking at improving it over the last couple of days. But input is of course allways welcome. For the record: I want to simplify the Computed/Computable stuff (not critical but might help). The main performance improvemnt most likely is to to figure out how to replace the Observable Descriptor with a property with closures (i.e., the property factory pattern you suggested earlier).

quaquel:
I forced myself to articuate this idea of seperating model state and data collection. Below is a quick sketch of what I mean with having an object that contains the model state you want to track. Your listener object now would just need to iterate over (a subset of) the defined datatables. A lot of your fast data getting code could be moved into the respective DataSet objects. Moreover, this design is easily extendible: just define a new DataSet subclass. 

What is missing here is datasets for variable sets of agents, and the tensor idea that you had. The last one could just be a `ArrayDataSet` and internally you would just have an ndarray of arbitrary dimension. The former is a bit trickier to do (and also very tricky in Mesa at the moment). 

The resulting API would be roughly

```python
class MyModel(Model):

	def __init__(self, rng=None):
		super.__init(self, rng=rng)

		self.model_output = ModelOutput()
		self.model_output.add_table(AgentDataSet("agent_data", self.model.agents, ["wealth", "health", "age"])
		self.model_output.add_table(ModelDataSet("model_data", self, ["gini", "average_age"])

		self.data_collector = CollectorListener()
```

A benefit of this approach is that these tables are now also available to others. For example, in Boltzman, the Gini calculation can now rely on `output[agent_data]`, or you just add a separate  `self.model_output.wealth` dataset. Some of the dirty flag stuff could be used as well. Basically, if we can figure out whether we need to renew the internal data, we can avoid having to collect agent-level data multiple times (as currently happens with gini and agent wealth in the data collector). 

@EwoutH, also curious to get your thoughts on this. 

The sketch for `ModelOutput` and `DataSet`

```python
from .agent import AgentSet
from .model import Model

class DataSet:
    # follows anylogic
    def __init__(self, name, fields):
        self.name = name
        self.fields = fields

        # internal datastructure

    @property
    def data(self):
        raise NotImplementedError

class AgentDataSet(DataSet):
    def __init__(self, name, agents:AgentSet, fields:str|callable|list[str|callable]):
        super().__init__(name, fields)
        self.agents = agents

    @property
    def data(self):
        # gets the data for the fields from the agents
        return ...

class ModelDataSet[M: Model](DataSet):
    def __init__(self, name, model:M, fields:str|callable|list[str|callable]):
        super().__init__(name, fields)
        self.fields = fields

    @property
    def data(self):
        # gets the data for the fields from the agents
        return ...


class TableDataSet(DataSet):
    def __init__(self, name, fields:str|list[str]):
        super().__init__(fields)
        self.datasets = {}

    @property
    def data(self):
        # gets the data for the fields from the agents
        return ...

class ModelOutput:

    def __init__(self):
        self.datasets = {}

    def add_dataset(self, dataset: DataSet):
        pass

    def create_dataset(self, dataset_type, name, fields, *args):
        pass

    def __getitem__(self, name:str):
        return self.datasets[name]
```

codebreaker32:
This is a fantastic architectural pivot. I completely agree and it also perfectly addresses the "Separation of Concern" problem.

Fast Optimisations(or complex logics) now move to Dataset
The DataSet abstraction makes it trivial to add a TensorDataSet( or RasterDataSet for Mesa-Geo later) without having to hack the core listener logic.

>Basically, if we can figure out whether we need to renew the internal data, we can avoid having to collect agent-level data multiple times

By treating the `AgentDataSet` as a stateful object rather than just a pass-through function, we can implement intra-step caching:-
1. The AgentDataSet holds a cached view of the current step's vectors.
2. The Gini reporter requests model.data.agents["wealth"]. The dataset sees its cache is empty/dirty, performs the optimized extraction (the "Fast Path"), caches the result, and returns it.
3. The CollectorListener requests the same data to save it. The dataset returns the cached reference immediately $O(1)$
4. .When the ModelSignal.STEP (or a PRE_STEP signal) fires, it marks the dataset as "dirty," forcing a refresh

Also I would suggest a more cleaner API:
```python
class MoneyModel(Model):
    def __init__(self):
        self.data = DataRegistry(self)

        self.data.track_agents("agents", self.agents, ["wealth", "id"])
        self.data.track_model("model", ["gini", "step_count"])
        
        # (Future) Track Spatial/Tensor
        # self.data.track_grid("terrain", self.grid, "elevation")
        self.datacollector = CollectorListener()
```
For this user only need DataRegistry()

```python
class DataRegistry:
    def __init__(self, model):
        self._model = model
        self._datasets = {}  # The internal storage

    def track_agents(self, name: str, source: AgentSet, reporters: list[str]) -> AgentDataSet:
        ds = AgentDataSet(name=name, agents=source, fields=reporters)
        self._datasets[name] = ds
        return ds

    def track_model(self, name: str="model", reporters: list[str]) -> ModelDataSet:
        ds = ModelDataSet(name=name, model=self._model, fields=reporters)
        self._datasets[name] = ds
        return ds

    def __getattr__(self, name):
        if name in self._datasets:
            return self._datasets[name]
        raise AttributeError(f"No dataset named '{name}'")

    def __iter__(self):
        return iter(self._datasets.values())
```

quaquel:
> By treating the AgentDataSet as a stateful object rather than just a pass-through function

Something else occurred to me. At the moment, we treat agent attributes and their collection as two separate things. However, with property layers and the experimental continuous space, we have attributes that are views into an underlying numpy array. If we can capture that design in a property/descriptor style object, the data table and the agent-level attribute become the same thing. 

Regardless, I'll try to put in a minimum-working PR for the DateRegistry/ModelOutput idea, with a few basic tables that cover many of our use cases.

mesa/model.py
EwoutH: > 1. These have to be aligned with the new run-control stuff that @EwoutH is working on. 

Just finished it:
- https://github.com/mesa/mesa/pull/3155

EwoutH:
Glad to such an engaging conversation!

I just finished https://github.com/mesa/mesa/pull/3155. From my perspective it’s production ready.

I will dive into this tomorrow or Sunday, now my brain needs a bit of rest.

mesa/experimental/collector/collectorlistener.py
```diff
@@ -0,0 +1,262 @@
...
+        agents = model.agents
+
+        # If it's an AgentSet, access the internal dict to skip the __iter__
+        # method overhead, but iterate it directly (Keys).
+        if hasattr(agents, "_agents"):
+            agents = agents._agents
+        elif isinstance(agents, dict):
+            # Only if it's a raw dict
+            agents = agents.values()
+
```
quaquel: This is in fact another argument for having both a dict and weakkeydict version of the agentset. Basically, model.agents and model.agents_by_type can use the dict version and that will make data collection often quite a bit faster.

codebreaker32: I strongly endorse this split

codebreaker32:
Given the architectural shifts we discussed, this PR currently needs to be aligned with:
1. #3155 (which is production ready, thanks to @EwoutH )
2. #3156 

As soon as these foundational pieces are ready, I will return to this PR to refactor the implementation to match the new design.

EwoutH:
Thanks for working on this! I had a quick initial look, there are indeed some interesting ideas here.

If I'm correct this is a hybrid design right? It listens to step events, and once those are observed it triggers a pull process where it gathers the specified data. So you could call this a "pull on signal" pattern?

In general, I think proper user control of A) what gets collected B) when it gets collected and C) how it gets stored is very important.

quaquel:
>If I'm correct this is a hybrid design right? It listens to step events, and once those are observed it triggers a pull process where it gathers the specified data. So you could call this a "pull on signal" pattern?

Yes that is indeed the design.

> In general, I think proper user control of A) what gets collected B) when it gets collected and C) how it gets stored is very important.

With #3156 and this, we separate this all more cleanly instead of trying to wrap it all into a single data collector. The `DataRegistry` covers what gets collected. The signals, as used here, determine when a snapshot of A is taken, and at least conceptually make it easy to store it or write your own backend storage.

codebreaker32:
I have updated the PR to align the `CollectorListener` with the #3156 architecture and the #3155 scheduler. This implementation now serves as the "Conductor" for data collection, supporting both standard `AgentDataSet` and the new experimental `NumpyAgentDataSet` (Zero-Copy).

1. **Separation of Concerns:**
* **Schema (`statistics.py`):** The `DataRegistry` and `DataSet` classes now fully define *what* to collect and *how* to extract it (including compiled `attrgetter` and property-injected Numpy arrays).
* **Orchestration (`collectorlistener.py`):** The `CollectorListener` is now purely responsible for *when* to collect and *how* to store it efficiently.


2. **Scheduler Integration (No Signals):**
* We moved away from `BaseListener` and `model.observe`.
* The `CollectorListener` now schedules itself on the model's `Scheduler` with `Priority.LOW`. This ensures collection happens automatically at the end of every step (after agents act) without manual hooks in `model.step()`.


3. **Lazy Block Storage ( Overhead):**
* **Agent Data:** We avoid iterating agent lists during the simulation step. Instead, we store `(Step, Block)` tuples.
* For `AgentDataSet`, we store the list reference.
* For `NumpyAgentDataSet`, we store a copy of the array.


* **Deferred Processing:** The expensive "explosion" of data (adding step columns, flattening lists) is deferred until the user explicitly requests the DataFrame via `get_table_dataframe()`.

* It seamlessly handles the return types from [`statistics.py`](#3156 ) (switching strategies for `list` vs `np.ndarray`).

quaquel:
Thanks for the update. I'll try to take a closer look at this later today.

mesa/experimental/collector/collectorlistener.py
```diff
@@ -0,0 +1,511 @@
...
+        # Initialize storage for all datasets in registry
+        self._initialize_storage(config or {})
+
+        # Subscribe to model steps observable
+        self._subscribe_to_steps()
+
+    def _initialize_storage(self, user_config: dict[str, dict[str, Any]]) -> None:
+        """Create storage containers for all datasets."""
+        for dataset_name, dataset in self.registry.datasets.items():
+            # SKIP TableDataSet: They manage their own history
```
quaquel: I have been thinking about this as well. Should TableDataSet be a dataset, or should it go directly to the CollectorListener?

mesa/experimental/collector/collectorlistener.py
```diff
@@ -0,0 +1,511 @@
...
+                self._store_data(name, current_step, data_snapshot, storage)
+                storage.config._next_collection = current_step + storage.config.interval
+            except Exception as e:
+                warnings.warn(
+                    f"Collection failed: dataset='{name}', step={current_step}: {e}",
+                    RuntimeError,
+                    stacklevel=2,
+                )
+
+    def _store_data(
```
quaquel: might it be an idea to have a _store_data_first_time instead of if statements internally?

mesa/experimental/collector/collectorlistener.py
```diff
@@ -0,0 +1,511 @@
...
+        self, name: str, step: int, data: Any, storage: DatasetStorage
+    ) -> None:
+        """Store data snapshot based on type."""
+        added_bytes = 0
+
+        # Numpy array (NumpyAgentDataSet) - MUST COPY
+        if isinstance(data, np.ndarray):
+            if data.size > 0:
```
quaquel: I would probably switch the if isinstance to a match statement and write seperate handlers for the data sets.

quaquel:
I really like the clarity of this collector. Just some comments and thoughts that might be useful to further refine this. But I strongly believe that between this PR and #3156, we are finally moving in a promising direction for improving the data collection in Mesa. 

Also, ideally, this collector should be easy to subclass so people can write custom classes if they, e.g., want to store stuff in a database. To this end, it might be useful to either specify an ABC for the listener or at least specify a Protocol.

mesa/experimental/collector/collectorlistener.py
codebreaker32: Likewise. TableDataSet doesn't support "collection", it supports "accumulation" So it technically belongs to Listener

mesa/experimental/collector/collectorlistener.py
codebreaker32: The main trade-off here is **code duplication**. Since we infer the data type lazily (from the first snapshot), both a `_store_first_time` and a `_store_subsequent` method would still need the exact same `isinstance` dispatch chain to handle the different payloads (Numpy vs. List vs. Dict).

If we split them, we would end up repeating the storage logic in both places. I leaned towards keeping the logic co-located to avoid violating DRY (Don't Repeat Yourself).

mesa/experimental/collector/collectorlistener.py
codebreaker32: Old habits die hard

codebreaker32:
Thanks @quaquel! I am glad you find it promising.

>it might be useful to either specify an ABC

I completely agree. I am actually already in the process of defining an ABC for the listener.
The goal is to formalize the contract (specifically the `collect` interface and the observable subscription logic) so that users can easily subclass it for custom backends(like a `SQLListener` or `ParquetListener`)without having to reimplement the signal handling logic.

I will include the ABC structure in the next update.

codebreaker32:
It is really good to see the architecture coming together. This update directly addresses the extensibility concerns raised by @quaquel regarding custom storage backends.

We have introduced an **Abstract Base Class (ABC)** pattern that enforces a clean separation of concerns between the *mechanics of listening* and the *implementation of storage*. As the name implies, we now distinguish between:

1. The Listener (Base Class): Handles the "When" ->Observables, configuration, intervals, and signal orchestration.
2. The Collector (Subclass): Handles the "How" -> Storage implementation (Memory, SQL, Parquet) and data format.

Formally, the Architectural Changes are:-

* **`BaseCollectorListener` (ABC):** Manages the lifecycle, configuration merging, and `model.steps` subscription. It defines the contract (`_store_dataset_snapshot`) that all backends must implement.
* **`CustomListener`:** It focuses solely on efficient storage and data conversion(or compression), stripping away the scheduling logic.

Note: This PR directly interacts with #3156

codebreaker32:
Also, I have implemented three custom backends- SQLListener, JSONListener and ParquetListener

All of them are working and I have tested locally includin CollectorListener. All examples except WolfSheep(10-20% slower) are performing better

quaquel:
> All of them are working and I have tested locally includin CollectorListener. All examples except WolfSheep(10-20% slower) are performing better

Just out of curiosity, how did you test this? I guess you used #3156 in some form?

codebreaker32:
> > All of them are working and I have tested locally includin CollectorListener. All examples except WolfSheep(10-20% slower) are performing better
> 
> Just out of curiosity, how did you test this? I guess you used #3156 in some form?

I copied `statistics.py` on my local branch. If you are curious enough I can push the exact changes in example models that you might need to check yourself

quaquel:
>  copied statistics.py on my local branch. If you are curious enough, I can push the exact changes in example models that you might need to check yourself

No need, I have done something similar locally as well for #3156.

codebreaker32:
I only changed(now reverted) to use Boltzmann and WolffSheep to use new registry(ref #3156) and listener.

Note: see results above

quaquel:
I have been playing around with the examples today as well and started adding tests to the data registry stuff. I hope to wrap that up early next week so we can start to merge the PRs. Results are roughly in line with the numbers shown here, although with wolf sheep and the numpy-based dataset, I had a reduction in runtime.

codebreaker32:
> although with wolf sheep and the numpy-based dataset, I had a reduction in runtime

I tried to use `track_agents_numpy()` but I got some error, currently figuring out the root cause

quaquel:
> I tried to use track_agents_numpy() but I got some error, currently figuring out the root cause

Ok, I won't have access to a computer the rest of the day, but still, please let me know if you find anything.

Also, any other thoughts and feedback on the current state of the dataset PR are welcome. It will likely take a couple of iterations after the integration with this PR to get it completely right. Moreover, some of it also depends on the further development of mesa_signals. 

I ran a quick local test with Schelling, turning `Agent.happy` into an observable and `model.happy` into a computed. It worked, but there was, as expected, a signal storm, so once we add a `batch` context manager, it might be nice to check again.

Also, I am considering adding a `_set_dirty` method to `AgentDataSet`. This gives users a hook to decide whether to regather data. For example, in Boltzman, you might tie it to `model.step` via `@emit`. On every model.step, you first diry the agent dataset, then inside model.step, you activate all agents. Then, when snapshotting data, you only have to gather agent.wealth once.

codebreaker32:
> Also, I am considering adding a `_set_dirty` method to `AgentDataSet`. This gives users a hook to decide whether to regather data. For example, in Boltzman, you might tie it to `model.step` via `@emit`. On every model.step, you first diry the agent dataset, then inside model.step, you activate all agents. Then, when snapshotting data, you only have to gather agent.wealth once

This would really be a very good optimisation. I have also been thinking about something like this in WolfSheep where agents frequently die, reproduce and energy exchanges. this approach would reduce significant overhead

codebreaker32:
> please let me know if you find anything

I was seeing an error earlier as I was missing the auto-registration loop in my local `Agent.__init__`
```python
        for dataset in self._datasets:
            self.model.data_registry[dataset].add_agent(self)
```
Thats why I was getting initialisation errors. Now fixed

mesa/experimental/data_collection/basedatarecorder.py
```diff
@@ -38,23 +38,82 @@ class DatasetConfig:
...
+
+        # Warmup phase: skip first 100 steps, collect 100-200
+        DatasetConfig(interval=1, start_time=100, end_time=200)
     """
 
     interval: int | float = 1
-    start: int | float = 0
+    start_time: int | float = 0
+    end_time: int | float | None = None
+    window_size: int | None = None
```
quaquel, line 66: should this not be an int or a float for devs models?

codebreaker32: window_size is strictly an int because it represents the buffer capacity (the maximum count of snapshots to keep in memory), not a time duration. For example, window_size=100 means "keep the last 100 collected data points," regardless of whether the simulation time is integer or float. It is similar to the "sliding window" approach.

codebreaker32: If you wish to collect time-bound data, you should use start_time and end_time

quaquel: clear. I thought it meant the time between snapshots. Is there a way to control this as well?

codebreaker32: Yes, that is controlled by the `interval` parameter which is a float to support DEVS as well

codebreaker32:
I believe this PR is ready for review. I have updated the description to reflect the final architecture.
Code reviews, testing, and any suggestions are highly welcome!

**Note:** Extensive testing depends on the `DataRegistry` implementation from #3156.

mesa/experimental/data_collection/basedatarecorder.py
quaquel, line 1: 1. python modules use an underscore naming convention, not a camel case naming conventions
2. I moved the dataset stuff into experimental.data_collection. Would it make sense to put these files there as well?

mesa/experimental/data_collection/basedatarecorder.py
```diff
@@ -0,0 +1,305 @@
+"""BaseCollectorListener for the Custom Listeners.
+
+Subclasses must implement:
+    - _store_dataset_snapshot: Store a single dataset snapshot
+    - _get_dataset_dataframe: Retrieve stored data as DataFrame
```
quaquel, line 5: is this really needed for all backends?

mesa/experimental/data_collection/basedatarecorder.py
```diff
@@ -0,0 +1,305 @@
...
+        # Collect every step, keep only last 1000 snapshots
+        DatasetConfig(interval=1, start_time=0, window_size=1000)
+
+        # Warmup phase: skip first 100 steps, collect 100-200
+        DatasetConfig(interval=1, start_time=100, end_time=200)
+    """
+
+    interval: int | float = 1
+    start_time: int | float = 0
+    end_time: int | float | None = None
```
quaquel, line 65: A tricky potential bug is how to handle the data collection when the simulation stops (for whatever stopping condition). If we collect on fixed intervals, we might miss the true end state of the data. How is this handled currently?

mesa/experimental/listeners/baseCollectorListener.py
```diff
@@ -0,0 +1,305 @@
...
+
+        # Load defaults from registry
+        for name in self.registry.datasets:
+            self.configs[name] = DatasetConfig()
+
+        # Override with user config
+        if config:
+            for name, user_cfg in config.items():
+                if name not in self.configs:
+                    warnings.warn(
```
quaquel: why did you handle this as a warning instead of as an exception?

mesa/experimental/listeners/baseCollectorListener.py
```diff
@@ -0,0 +1,305 @@
...
+                        config.update_next_collection(current_time)
+
+                    except Exception as e:
+                        warnings.warn(
+                            f"Initial collection failed: dataset='{name}': {e}",
+                            RuntimeWarning,
+                            stacklevel=2,
+                        )
+        else:
+            warnings.warn(
```
quaquel: This should become unnecessary once #3212 has been merged.

mesa/experimental/listeners/baseCollectorListener.py
```diff
@@ -0,0 +1,305 @@
...
+                        dataset = self.registry.datasets[name]
+                        data_snapshot = dataset.data
+
+                        self._store_dataset_snapshot(name, current_time, data_snapshot)
+
+                        # Important: Advance the next schedule so we don't double-collect
+                        config.update_next_collection(current_time)
+
+                    except Exception as e:
+                        warnings.warn(
```
quaquel: why handle this as a warning?

mesa/experimental/listeners/baseCollectorListener.py
```diff
@@ -0,0 +1,305 @@
...
+
+    def get_all_dataframes(self) -> dict[str, pd.DataFrame]:
+        """Get DataFrames for all datasets."""
+        return {name: self.get_table_dataframe(name) for name in self.configs}
+
+    @abstractmethod
+    def summary(self) -> dict[str, Any]:
+        """Get collection status summary."""
+
+    def __del__(self):
```
quaquel: Is this the right place to put it, because this is only called when the instance is garbage-collected. I am thinking of adding a `model.close` that would allow for better cleanup (e.g., remove all agents, close all datasets, close collectors etc.).

mesa/experimental/listeners/collectorlistener.py
```diff
@@ -0,0 +1,286 @@
...
+if TYPE_CHECKING:
+    from mesa.model import Model
+
+# Optional Polars for high-performance output
+try:
+    import polars as pl
+
+    HAS_POLARS = True
+except ImportError:
+    HAS_POLARS = False
```
quaquel: I suggest leaving this out for now. As part of mesa 4, @EwoutH and me want to take a look at making polars a (optional) dependency. But not before that.

mesa/experimental/listeners/collectorlistener.py
```diff
@@ -0,0 +1,286 @@
...
+class DatasetStorage:
+    """Storage container with sliding window support."""
+
+    blocks: deque = field(default_factory=deque)
+    metadata: dict[str, Any] = field(default_factory=dict)
+    total_rows: int = 0
+    estimated_size_bytes: int = 0
+
+
+class CollectorListener(BaseCollectorListener):
```
quaquel: not sure about the naming of this but don't have a better name myself at the moment either.

mesa/experimental/listeners/collectorlistener.py
```diff
@@ -0,0 +1,286 @@
...
+                storage.blocks.append((time, data_copy))
+                storage.total_rows += len(data_copy)
+                added_bytes = data_copy.nbytes
+
+                if "type" not in storage.metadata:
+                    storage.metadata["type"] = "numpyagentdataset"
+                    storage.metadata["dtype"] = data.dtype
+                    try:
+                        dataset = self.registry.datasets[dataset_name]
+                        storage.metadata["columns"] = list(dataset._attributes)
```
quaquel: Consider moving the second line into the else clause of try, except, else.

mesa/experimental/listeners/collectorlistener.py
```diff
@@ -0,0 +1,286 @@
...
+                storage.blocks.append(row)
+                storage.total_rows += 1
+                added_bytes = 100
+
+                if "type" not in storage.metadata:
+                    storage.metadata["type"] = "modeldataset"
+                    storage.metadata["columns"] = [*list(data.keys()), "time"]
+
+            case np.ndarray() | list():
+                pass  # Empty
```
quaquel: why?

mesa/experimental/data_collection/basedatarecorder.py
codebreaker32: 1. Just my java muscle memory kicking in. I'll modify it follow python conventions
2. Are you suggesting we merge the listener and data_collection folders? I would advise against a full merge. The current structure encourages contributors to add custom listeners by giving them a distinct space.

Alternatively, we could move the core Base and CollectorListener classes into data_collection, but then I recommend keeping a separate directory dedicated to custom implementations.

mesa/experimental/data_collection/basedatarecorder.py
codebreaker32: In my opinion, both are essential:

1. `_store_dataset_snapshot`: Abstracts how data is persisted (e.g., INSERT for SQL vs. write_row for Parquet), allowing the simulation to run without knowing the storage format.

2. `_get_dataset_dataframe`: Abstracts how data is loaded, guaranteeing that the users always receive a standard DataFrame, regardless of the source. I made it abstract after our discussion to keep Pandas a first-class citizen for all listeners

Enforcing both ensures that swapping a backend (e.g., Memory -> SQL) never breaks the visualization or analysis code. Frankly, I kept them abstract assuming that all the custom listeners will reside in Mesa. 

Another idea that is, we can remove the `@abstract` decorator and instead raise `NotImplementedError`. Is it good enough or should I strip it completely?

mesa/experimental/data_collection/basedatarecorder.py
codebreaker32: Good catch! Its not currently handled
I see two options:
1. `listener.finalise()`: Users need to call it to finalise the data collection which will do nothing but call `.collect()`
2. 
```python
with CollectorListener(model, config) as listener:
    model.run_while()    # model.run_model()
    # __exit__ automatically does final collect()
    
df = listener.get_table_dataframe("wealth")
```

Do you have anything in mind?

mesa/experimental/listeners/baseCollectorListener.py
codebreaker32: I changed it to warning to allow flexible configuration (e.g., same config across models with different datasets). Should this be stricter? Happy to make it an exception if that's preferred

mesa/experimental/listeners/baseCollectorListener.py
codebreaker32: yep, I'll remove it

mesa/experimental/listeners/collectorlistener.py
codebreaker32: Sure

mesa/experimental/listeners/collectorlistener.py
codebreaker32: Sure

mesa/experimental/listeners/collectorlistener.py
codebreaker32: This prohibits two special cases to not enter the "default" case
1. case np.ndarray() if data.size <= 0  
"OR"
2. case list() if  not data

I forgot to clean it up ( by keeping `if data.size > 0` in `case` block and same for list())

mesa/experimental/listeners/baseCollectorListener.py
codebreaker32: Using warnings allows simulation to continue even if one dataset fails. Mainly to reuse the configs across different models while I was testing different listeners. I'd replace RuntimeWarning with RuntimeExceptions(here and in many instances) to make it stricter if you suggest

mesa/experimental/listeners/baseCollectorListener.py
codebreaker32: Agree.

mesa/experimental/data_collection/basedatarecorder.py
quaquel: I don't know. It depends in part on API design. Currently, we have `model.datacollector`. After an experiment, you get your results from this. In ongoing work, e.g., on the time API, more is added to the model, making it the central object you interact with. 

With a context manager, you would move the data collection outside of the model itself, which runs counter to this "model as central object" design.

If we instead want to stick with this "model as central object" design, there are at least two options: 
1. Make the new run control API reactive, such that it emits a signal when a run is completed. This, however, in turn would require a clear definition of when a run is completed. 
2. Add some `finalise`-style method as per your suggestion.

@EwoutH, I would appreciate your thoughts on this as well.

mesa/experimental/listeners/baseCollectorListener.py
quaquel: I am inclined to prefer exceptions over warnings for this. Model runs can be expensive. To then realize there was an error in your data collection and you waited a few hours for nothing is very frustrating. If we raise exceptions here, we can allways decide in some new experiment runner how to handle these exceptions.

mesa/experimental/listeners/collectorlistener.py
codebreaker32: How about DataWriter?

mesa/experimental/data_collection/basedatarecorder.py
quaquel: > Are you suggesting we merge the listener and data_collection folders? I would advise against a full merge. The current structure encourages contributors to add custom listeners by giving them a distinct space.

I am inclined to merge them, as both are related to data collection. I am not sure I fully appreciate your argument about the current structure encouraging custom listeners. Why would the folder matter for this?

mesa/experimental/data_collection/basedatarecorder.py
quaquel: your explanation makes sense and I am fine with keeping it as is.

I still need to think through how to tie the new dataset stuff into the UI.

mesa/experimental/data_collection/basedatarecorder.py
codebreaker32: > Why would the folder matter for this?

My reasoning was that a dedicated `listeners` folder signals that storage backends are pluggable extensions, distinct from the core extraction logic.
However, I agree that co-locating them makes more sense functionally. As soon as #3156 merges, I will move them to that folder

mesa/experimental/listeners/jsonListener.py
quaquel, line 1: 1. can you please change the names of all files to use underscores instead of CamelCase?
2. I guess its a Java thing to put each class in its own file (for the record, Java was the first programming language I learned and still occasionally use). Would it not make more sense to put these all into `collector_listeners.py`

codebreaker32: 1. Sure
2. Java is my first language as well

codebreaker32:
I have fixed the naming. Next few things to address:

1. Removing `__del__` in favor of an explicit `model.close()`
2. Implementing logic to capture the final data snapshot when the simulation stops, ensuring no data is lost due to interval timing.
3. Exploring how well this new listener+dataregistry can interact with U

codebreaker32:
>1. Removing __del__ in favor of an explicit model.close() 
>2. Implementing logic to capture the final data snapshot when the simulation stops, ensuring no data is lost due to interval timing.

@quaquel, can we solve both by triggering a finalize() method directly inside model.close()? The only problem that I see is that this enforces a stricter lifecycle: users would be required to call model.close() to guarantee that the final data points are recorded before interacting with the data

quaquel:
> @quaquel, can we solve both by triggering a finalize() method directly inside model.close()? The only problem that I see is that this enforces a stricter lifecycle: users would be required to call model.close() to guarantee that the final data points are recorded before interacting with the data

To be honest, I am not sure yet. It will, in part, depend on how the new TIME API shapes up and how people might use it. 

1. If we have a `run_length` parameter as part of some RunControl specification, we can use this. Basically, if the model is stopped because of run_length, or more generally, any stopping condition, we could detect this and automagically add the last snapshot.
2. Alternatively, we might do a check when calling the to_dataframe method. Basically, is the latests snapshot from the current timestep, if not, add it in. This allows for a more natural workflow where people might use run_until or some kind of stepping through the model over time.

codebreaker32:
I am convinced by the fact that further progress of this PR highly relies on new TIME APIs and DATA REGISTRY Architecture. As soon as those ideas crystallize, I'll come back to update this PR

quaquel:
> I am convinced by the fact that further progress of this PR highly relies on new TIME APIs and DATA REGISTRY Architecture. As soon as those ideas crystallize, I'll come back to update this PR

I agree, a lot hinges on #3212 and the TIME api. I hope to have time to add more tests and further refine #3156 tomorrow, and hopefully next week we can start merging a bunch of things.

EwoutH:
Thanks for working on this. I have an API question (which relates both to this PR and #3156).

If I understand correctly, the current setup requires defining datasets in the registry, then separately constructing a parallel config dict for the listener:

```python
self.data_registry.track_agents(self.agents, "agents_wealth", "wealth")
self.data_registry.track_model(self, "model_data", "gini")

config = {
    "agents_wealth": DatasetConfig(interval=1, start=0),
    "model_data": DatasetConfig(interval=5, start=100, end=500),
}
self.listener = CollectorListener(self, config=config)
```

This names things twice and couples the two by string keys. Could we instead chain `.listen()` onto the track calls?

```python
self.data_registry.track_agents(self.agents, "agents_wealth", "wealth") \
    .listen(interval=1, start=0)

self.data_registry.track_model(self, "model_data", "gini") \
    .listen(interval=5, start=100, end=500)

# Do you even still need this?
self.listener = CollectorListener(self)
```

The `track_*` methods already return the dataset, `.listen()` would attach schedule config to it. The Listener reads configs from the datasets rather than needing a separate dict. Datasets without `.listen()` aren't collected.

<hr>

As a separate question, do we need a `Schedule`  or `Pattern` object that can be reused? It seem that data collection needs it, but event scheduling also (https://github.com/mesa/mesa/discussions/2921#discussioncomment-15562881).

We're designing the same concept twice:

**Data collection (#3145):**
```python
DatasetConfig(interval=1, start=0, end=500, window_size=1000)
```

**Event scheduling (#2921):**
```python
model.pattern(callback, interval=1, start_at=0, end_at=500)
```

Both describe "do something repeatedly on a schedule." The parameters are nearly identical: interval, start, end, count/window. The only difference is what happens at each tick: one stores a snapshot, the other calls a function.

Should we extract a shared `Schedule` (or `Pattern`) object?

```python
# Define a schedule once
every_step = Schedule(interval=1, start=0)
weekly = Schedule(interval=7, start=100, end=500)

# Use it for data collection
self.data_registry.track_agents(self.agents, "wealth", "wealth") \
    .listen(every_step)

self.data_registry.track_model(self, "model_data", "gini") \
    .listen(weekly)

# Use it for event scheduling
model.create_recurring_event(self.migrate, schedule=weekly)
```

This eliminates `DatasetConfig` as a separate concept, gives users one scheduling vocabulary across the framework, and avoids the two systems diverging in parameter names (`start` vs `start_at`, `end` vs `end_at`, etc.).

If an new object is overkill, we should at least consider aligning the API / keyword arguments.

quaquel:
I am intrigued by both ideas.

1. If we add `listen` it would become a method on the dataset. I am not sure that listen is than the right name because the dataset is not the one who does the listening. Rather it should be something like `register` or `take_snapshots`. Ideas are welcome, but I like the cleanliness of the API that would result. It also makes clear in one view which datasets are for internal use and which ones are (also) stored over time. 
2. Not sure about abstracting a schedule or pattern object, but at a minimum, it would be good to at least have a protocol that is the same for both. 


@EwoutH, one more thing, I guess it will take a few iterations to get all this stuff right. I would prefer moving forward with these PRs and start merging them so we can start iterating. We should not let the perfect be the enemy of the good. In particular since all this stuff for now still resides in experimental.

codebreaker32:
> The `track_*` methods already return the dataset, `.listen()` would attach schedule config to it. The Listener reads configs from the datasets rather than needing a separate dict. Datasets without `.listen()` aren't collected

I also really like the chaining proposal. It makes the declaration self-contained and much more readable. Regarding the naming, I agree `listen()` feels slightly off since the dataset is passive. I would prefer `record()` (or `collect()`), e.g., `track_agents(...).record(interval=1)`

> As a separate question, do we need a `Schedule` or `Pattern` object that can be reused?

I advise caution against strict coupling between Data Collection and Event Scheduling.
While I agree on standardizing the vocabulary (Protocol), sharing a concrete **class** introduces risks:

1. Data collection is typically uniform sampling (simple intervals). Event scheduling often requires complex control flow (conditional recurrence). A shared object risks being over-engineered for collection or restrictive for events or maybe so generic to be actually useful.
2. "Scheduling" implies active model intervention; "Collection" is passive observation.

We should enforce a shared Protocol/Signature (strictly aligning keywords like `start_at`, `interval`, `end_at`) to ensure consistency(as @quaquel noted), but keep the implementations decoupled. This avoids premature abstraction while solving the parameter drift issue.

quaquel:
I like `record()`. This might also offer a better name for the `CollectorListener`: `DataRecorder`. And then, depending on the backend you get. eg., JSONDataRecorder etc. 

Implementation-wise, however, I suggest we first merge at least the model, reactive, and dataset stuff, and then either merge this PR and add a record in a follow-up or include it as part of this PR. I can't really include it in the dataset PR since it is tied to this PR.

codebreaker32:
I have modified snapshot methods to properly handle incoming `agent_ids` from dataregistry. Also I have added the tests. I believe it is ready for review now

mesa/experimental/data_collection/basedatarecorder.py
```diff
@@ -0,0 +1,265 @@
...
+
+    def _subscribe_to_model(self) -> None:
+        """Subscribe to model.time for automatic collection."""
+        # Subscribe to time units observable
+        self.model.observe("time", ObservableSignals.CHANGED, self._on_time_change)
+
+        # Check for immediate collection (Initial State / Time 0)
+        current_time = self.model.time
+
+        for name, config in self.configs.items():
```
quaquel: This seems repetitive with `_on_time_change`. Why not just call that method instead

mesa/model.py
```diff
@@ -115,6 +115,7 @@ def __init__(
             you have to pass either seed or rng, but not both.
 
         """
+        HasObservables.__init__(self)
```
quaquel: why can't we call super here? In principle shoudl super also handle multiple inheritance.

mesa/experimental/data_collection/datarecorders.py
```diff
@@ -0,0 +1,664 @@
...
+            f"rows={sum(s.total_rows for s in self.storage.values())}, "
+            f"memory={memory})"
+        )
+
+
+class NumpyJSONEncoder(json.JSONEncoder):
+    """JSON Encoder that handles Numpy types."""
+
+    def default(self, obj):
+        """Convert Numpy types to native Python types."""
```
quaquel, line 272: is there no built in way to do this?

quaquel:
I really like the clarity of the code. I made a start with the review and hope to spend some more time on it tomorrow.

In the meantime. Any idea what's going on with the benchmarks? And can you improve code coverage?

mesa/experimental/data_collection/basedatarecorder.py
quaquel: Also, this means that you can only instantiate the recorder after having initialized the rest of the model. So it's fragile. As discussed [here](https://github.com/mesa/mesa/discussions/2921#discussioncomment-15748424), we might need signals for run_started and run_ended so we can use these to trigger the first and last data collection if desired.

codebreaker32: I agree with this new signals' approach. We can do this then
```python
def _subscribe_to_model(self) -> None:
        """Subscribe to model signals."""
        ...
        self.model.observe("time", ObservableSignals.CHANGED, self._on_time_change)
        self.model.observe("model", ModelSignals.RUN_STARTED, self._on_run_started)
        self.model.observe("model", ModelSignals.RUN_ENDED, self._on_run_ended)

    def _on_run_started(self, signal) -> None:
        """Handle run start (collect initial state)."""
        self._check_and_collect(self.model.time)

    def _on_run_ended(self, signal) -> None:
        """Handle run end."""
        self._check_and_collect(self.model.time)
   
```
`_check_and_collect()` will ensure we don't collect the same data twice even if RUN_ENDED fires on a step we just collected

mesa/model.py
codebreaker32: As far as I remember, I added it because I ran into some error but now I tested it and its running fine without it so yes its redundant

mesa/experimental/data_collection/datarecorders.py
codebreaker32: Sadly there isn't, I searched for some workarounds and liked this one

See: https://stackoverflow.com/questions/50916422/python-typeerror-object-of-type-int64-is-not-json-serializable
https://stackoverflow.com/questions/26646362/numpy-array-is-not-json-serializable

codebreaker32:
> Any idea what's going on with the benchmarks? And can you improve code coverage?

When I tested it previously(when you just completed NumpyAgentDataset) on BoltzmannWealth, it was comparatively better but yesterday when I modified all examples, Schelling showed huge regression, I will investigate it today and hopefully add tests coverage also

quaquel:
> When I tested it previously(when you just completed NumpyAgentDataset) on BoltzmannWealth, it was comparatively better but yesterday when I modified all examples, Schelling showed huge regression, I will investigate it today and hopefully add tests coverage also

I played around with the code yesterday locally. After removing the one line from the model class, I have the same performance across main and this PR. Adding in some datasets, and I see improved performance.

mesa/experimental/data_collection/basedatarecorder.py
quaquel: @EwoutH, shall I add this via a separate PR after #3266 has been merged? 

We still need to figure out the run_ended signal....

codebreaker32:
> I played around with the code yesterday locally. After removing the one line from the model class, I have the same performance across main and this PR. Adding in some datasets, and I see improved performance.

Could you elaborate a bit please?

quaquel:
> Could you elaborate a bit please?

First, I removed the redundant `HasObservables.__init__` line in model.py. Then I ran the default benchmark code. Performance was identical, as expected. Nothing in this PR should affect the default benchmarks. Then, I updated Boltzmann by replacing the data collector with datasets and a datarecorder. If I use an AgentDataSet, I see a small (5%) reduction in runtime. If I use a NumpyAgentDataSet and update gini to use a numpy implementation, I see a slight increase in runtime (about 10%). I haven't tested the other examples yet.

codebreaker32:
Thanks for the detailed clarification. The performance regression I was seeing might be due to that `HasObservables.__init__`.

EwoutH:
Thanks again for all the work on this.

With all the accumulated insights, I look a clean look at the API. I think I have a quite neat proposal: https://github.com/mesa/mesa/discussions/1944#discussioncomment-15766733. Very curious about your thoughts.

codebreaker32:
I have addressed all concerns under the scope of this PR. The only remaining issue is that the final data collection step is currently skipped if `current_time % interval != 0`; we are planning to resolve this by utilizing `RUN_STARTED` and `RUN_ENDED` signals to ensure that the last data point is captured.

Edit: Added explicit `finalise()` method to capture final state that internally calls `self.collect()`

pyproject.toml
```diff
@@ -69,6 +69,7 @@ dev = [
   "pytest-ipywidgets",
   "playwright",
   "ipython>=9.2.0",
+  "pyarrow",
```
quaquel, line 72: this adds an optional dependecy. I am fine with that, but would love to get the perspective of some other maintainers. @EwoutH, @tpike3, any thoughts?

mesa/experimental/data_collection/basedatarecorder.py
quaquel: I'll take a look and see how difficult a RUN_ENDED signal will be at the moment. My hunch is that it won't be trivial because we lack a RunConfiguration class. So, I suggest we add a close or finalize method for now, but with a fixme related to RUN_ENDED.

pyproject.toml
EwoutH: Why is it a dev dependency? Only used in tests?

codebreaker32: Yes, for testing parquet

mesa/experimental/data_collection/basedatarecorder.py
codebreaker32: Done.

pyproject.toml
EwoutH: For dev only it's fine

mesa/experimental/data_collection/basedatarecorder.py
quaquel: To me this is the only remaining issue we have to adress before we can merge this PR. See #3284 for one approach (which would require a minor change to this code).

EwoutH: Lets address this in a follow-up PR

codebreaker32: I am fine with #3284. I explored alternatives to fix this on the recorder side (to avoid modifying model.py), such as using a flag to track if the initial state was collected. However, that approach is flawed: if the first signal fires after the model has advanced (e.g., at t=1), we capture the current state and falsify it as an initial state rather than the true initial state (t=0). Since this is fundamentally an initialization order issue in the model, it must be solved there or via a dedicated signals approach mentioned above

EwoutH:
@codebreaker32 could you update the PR description based on the accumulated insights in the discussions and review process? Please take some time for this and don’t fully outsource it to an LLM, having clarity on the current direction and all the considerations behind it is extremely important.

codebreaker32:
> Please take some time for this and don’t fully outsource it to an LLM, having clarity on the current direction and all the considerations behind it is extremely important

Sure, I understand the importance of it.

mesa/experimental/data_collection/basedatarecorder.py
quaquel: > Lets address this in a follow-up PR

I would not be in favour of that. I suggest we merge #3284, open a seperate issue or discussion on RunConfiguration, update this PR and then merge this. 

With that everything is in place and we have a clean usable alternative to data collection that people can start experimenting with. 

In a follow up PR, we can then start adding the more refined API you suggested. But, for example `model.datasets.track().record` can only be implemented properly after we have resolved the intit data collection issue.

codebreaker32:
Hi @EwoutH 

I have updated the PR description. Please check and tell if you want me to modify anything

EwoutH:
Thanks, appreciated!

The discussed chaining is excluded from this PR right? What’s the plan for it, if any?

Could you add one minimal example of going from the old data collection to the new one?

codebreaker32:
> The discussed chaining is excluded from this PR right? What’s the plan for it, if any?

It might make more sense to clarify this in the PR description of #3156

>Could you add one minimal example of going from the old data collection to the new one?

Sure

quaquel:
> > The discussed chaining is excluded from this PR right? What’s the plan for it, if any?

> It might make more sense to clarify this in the PR description of https://github.com/mesa/mesa/pull/3156

I need to have #3284, and this one merged. Once both are in, I'll add a PR for the discussed chaining. It requires an update to the dataset protocol and implementations.

codebreaker32:
> Could you add one minimal example of going from the old data collection to the new one?

I have added a migration APIs in the PR Description.

I have one more suggestion, Should we move BoltzmannWealth to the experimental section to use these new APIs?

quaquel:
> I have one more suggestion, Should we move BoltzmannWealth to the experimental section to use these new APIs?

For the time being, we need to duplicate the data collection since solara does not use the new style yet. Not sure which examples would be good to move over. For benchmarking, boltzmann is usefull and it will illustrate the key components. It can also be done in a next PR.

quaquel:
I have merged #3284. As a last request for this PR, can you tie the initial collection to this signal?

codebreaker32:
> I have merged #3284. As a last request for this PR, can you tie the initial collection to this signal?

Sure

codebreaker32:
One more thing worth noting is:
1. Now first `model.step()` will emit two signals So calling `model.step()` initially, will populate data with twice entries
2. If user wants to collect initial data without calling `step()`, they'll have to manually call `recorder.collect()`

quaquel:
> Now first model.step() will emit two signals So calling model.step() initially, will populate data with twice entries

I am not sure I follow. If I track the signals, I nicely see the time incrementing. So the first signal has as `new` 0, then the second signal has as `new` 1. etc. So it's not entirely true to state that `model.step` emits twice. Rather, before `model.step` is called, `model.time` emits a change to 0.0. Then `model.step` is run, after which `model.time` emits a change to 1.0. Then model.step is run again, and `model.time` emits a change to 3.0. etc.

codebreaker32:
> If I track the signals, I nicely see the time incrementing

You are right that's why I didn't call it bug anywhere but the time=0.0 signal is now emitted inside the `_advance_time` method (which is triggered by `step()`), the very first time you write `model.step()`, Python executes both the $t=0$ emission and the $t=1$ emission.

I noticed this behavior while modifying the tests for new behavior. You can check yourself

MRE:
```python
from mesa import Model, Agent
from mesa.experimental.data_collection import DataRecorder

class SimpleAgent(Agent):
    def __init__(self, model, wealth):
        super().__init__(model)
        self.wealth = wealth

class SimpleModel(Model):
    def __init__(self):
        super().__init__()
    
        SimpleAgent.create_agents(self, 3, [10, 20, 30])
        self.data_registry.track_agents(self.agents, "agent_data", fields=["wealth"])
        

model = SimpleModel()
recorder = DataRecorder(model)
df = recorder.get_table_dataframe("agent_data")
print(f"No step called,\n {df.to_string()} \n")


model.step()
df = recorder.get_table_dataframe("agent_data")

print(f"Step called once:\n {df.to_string()}\n")

model.step()

df = recorder.get_table_dataframe("agent_data")

print(f"Step called twice:\n {df.to_string()}")

```

Output:

```plaintext
No step called,
 Empty DataFrame
Columns: []
Index: [] 

Step called once:
    unique_id  wealth  time
0          1      10   0.0
1          2      20   0.0
2          3      30   0.0
3          1      10   1.0
4          2      20   1.0
5          3      30   1.0

Step called twice:
    unique_id  wealth  time
0          1      10   0.0
1          2      20   0.0
2          3      30   0.0
3          1      10   1.0
4          2      20   1.0
5          3      30   1.0
6          1      10   2.0
7          2      20   2.0
8          3      30   2.0
```

quaquel:
This is the behaviour I would expect. Ideally, we would trigger the t=0 collect after completing the initialization of the model, but before returning the instantiated object. Unfortunately, there is no `__post_init__` on Python classes. There are ways to achieve the same via metaclasses or possibly via `__init_subclass__`. For now, however, I'd say that this is good enough. Finetuning this will become relevant once we start to use signals and potentially this new data collection in combination with the ui.

codebreaker32:
> Finetuning this will become relevant once we start to use signals and potentially this new data collection in combination with the ui.

Agree, just felt the need to explicitly state it out

PR #3147, opened 2026-01-16
Deprecate `seed` parameter in favor of `rng` in Model

quaquel:
This PR deprecates the `seed` parameter in `Model.__init__()` and `Model.reset_randomizer()` in favor of the SPEC-7 compliant `rng` parameter. Users will now see a `FutureWarning` when using `seed`, encouraging migration to `rng`.

### Migration
This is a simple parameter name change with no behavioral differences:

```python
# Old (now deprecated)
model = MyModel(seed=42)

# New
model = MyModel(rng=42)
```

The functionality remains identical - both approaches seed the random number generators in the same way. This change is part of aligning Mesa with [SPEC-7](https://scientific-python.org/specs/spec-0007/) (scientific Python community standards for random number generation).

### Notes
- `model.random` (Python's `random.Random`) continues to exist and function as before
- `model.rng` (NumPy's `Generator`) continues to exist and function as before
- Both are seeded identically whether you use `seed` or `rng` parameter
- The deprecation will be removed in Mesa 4.0

Ref: #2352 (original `rng` implementation)

### DIFF
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -256,7 +256,7 @@ class Model[A: Agent]:
             seed: A new seed for the RNG; if None, reset using the current seed
         """
         warnings.warn(
-            "the use of seed is deprecated, use rng instead",
+            "The use of the `seed` keyword argument is deprecated, use `rng` instead. No functional changes.",
             FutureWarning,
             stacklevel=2,
         )


mesa/model.py
```diff
@@ -120,6 +120,12 @@ def __init__(
                 self.random = random.Random(seed)
             self._seed = seed  # this allows for reproducing stdlib.random
         elif rng is None:
+            warnings.warn(
+                "the use of seed is deprecated, use rng instead",
```
EwoutH: Add link to migration guide

mesa/model.py
```diff
@@ -282,6 +288,12 @@ def reset_randomizer(self, seed: int | None = None) -> None:
         Args:
             seed: A new seed for the RNG; if None, reset using the current seed
         """
+        warnings.warn(
+            "the use of seed is deprecated, use rng instead",
```
EwoutH: link to migration guide

mesa/model.py
```diff
@@ -120,6 +120,12 @@ def __init__(
                 self.random = random.Random(seed)
             self._seed = seed  # this allows for reproducing stdlib.random
         elif rng is None:
+            warnings.warn(
+                "the use of seed is deprecated, use rng instead",
+                DeprecationWarning,
```
EwoutH: We now use `FutureWarning` following our [Deprecation policy](https://github.com/mesa/mesa/blob/main/CONTRIBUTING.md#deprecation-policy). `DeprecationWarning` isn't always shown to users if they run a sub-classed model.
```suggestion
                FutureWarning,
```

mesa/model.py
```diff
@@ -282,6 +288,12 @@ def reset_randomizer(self, seed: int | None = None) -> None:
         Args:
             seed: A new seed for the RNG; if None, reset using the current seed
         """
+        warnings.warn(
+            "the use of seed is deprecated, use rng instead",
+            DeprecationWarning,
```
EwoutH:
```suggestion
            FutureWarning,
```

tests/test_model.py
```diff
@@ -60,33 +60,33 @@ def step(self):
...
-    model.reset_randomizer(seed=newseed)
-    assert model._seed == newseed
+# def test_reset_randomizer(newseed=42):
+#     """Test resetting the random seed on the model."""
+#     model = Model()
+#     oldseed = model._seed
+#     model.reset_randomizer()
+#     assert model._seed == oldseed
+#     model.reset_randomizer(seed=newseed)
+#     assert model._seed == newseed
```
EwoutH: Is there a reason these tests can't keep running? Technically we still want to know when deprecated behavior breaks.

You can catch warnings with something like:

```Python
with pytest.warns(DeprecationWarning):
    model = Model(seed=42)
    ...
```

EwoutH (commented):
Thanks for this extensive effort, appreciated.

Overal looks good, a (small) note in the migration guide should be added and a few nitpicks.

mesa/model.py
quaquel: There is nothing in the migration guide on this. Also, I would argue that the message is all you need. Just use rng instead of seed.

tests/test_model.py
quaquel: I just needed to check if you could indeed do this with pytest and I forgot about it.

mesa/model.py
EwoutH: While the actual code change is indeed very small, that change is made for a reason and has implications. Therefor I do think an entry is warranted.

Completeness is also important here. The migration guide should contain everything you need to know when updating to a new Mesa version.

It's also a chance to educate our users.

I think something like this would be a good start (feel free to edit of course).

````md
## Mesa 3.5.0
### Deprecation of `seed` in favor of `rng`
The `seed` parameter on `Model.__init__` and the `reset_randomizer` method are deprecated in favor of [[SPEC-7](https://scientific-python.org/specs/spec-0007/)](https://scientific-python.org/specs/spec-0007/) compliant NumPy random number generation.

NumPy's `Generator` class provides higher-quality random number generation compared to Python's built-in `random` module or NumPy's legacy `RandomState`. The new `rng` parameter follows scientific Python community standards and offers better reproducibility controls.

```python
# Old
model = MyModel(seed=42)
model.reset_randomizer(seed=42)

# New
model = MyModel(rng=42)
model.reset_rng(rng=42)
```

Within your model, use `model.rng` (a `numpy.random.Generator`) instead of `model.random` for random number generation. The `rng` parameter accepts integers, sequences of integers, `numpy.random.SeedSequence`, `numpy.random.Generator`, or `numpy.random.BitGenerator`.

* Ref: [PR #2352](https://github.com/mesa/mesa/pull/2352)](https://github.com/mesa/mesa/pull/2352)
````

Comment (quaquel) [reply mesa/model.py]: But most of what is said in your suggested text is not what this PR does. `Model.random` still exists. If you used `seed` and now shift to `rng`, the model will even behave the same. Really, the only thing that has changed is that we prefer using the `rng` keyword over the `seed` keyword, without any change in underlying logic or behavior.

Comment (EwoutH) [reply mesa/model.py]: Ah right, I thought this was a bigger shift.

**Comment by EwoutH** [comment]
For my context, is this the first step in a series of deprecations towards SPEC-7 compliance, of is this (for now) the only one?

What's the future of `.random()`?

I feel somehow we should communicate to our users why we're changing this and how that impacts them.

**Comment by quaquel** [comment]
> For my context, is this the first step in a series of deprecations towards SPEC-7 compliance, of is this (for now) the only one?

This is all that is required to become compliant with spec-7. Spec-7 specifies the use of `rng` as keyword argument and the use of non-global random number generators. The latter we already have with model.rng and model.random

> What's the future of .random()?

To be decided. I prefer fully removing it inside Mesa, but this has profound performance consequences. I also figured out that `random.Random`, which we use in mesa, is a non-global instance of a seeded stdlib random number generator. So one of  my reasons for removing `model.random` is not valid.

> I feel somehow we should communicate to our users why we're changing this and how that impacts them.

Like I said, the impact at the moment is only a change in preferred keyword (to become spec-7 compliant). We could explain this in the release notes.

File: mesa/model.py [hunk 6]
```diff
@@ -118,6 +119,12 @@ def __init__(
                 self.random = random.Random(seed)
             self._seed = seed  # this allows for reproducing stdlib.random
         elif rng is None:
+            warnings.warn(
+                "the use of seed is deprecated, use rng instead",
```
EwoutH:
```suggestion
                "The use of the `seed` keyword argument is deprecated, use `rng` instead. No functional changes.",
```

mesa/model.py
```diff
@@ -248,6 +255,12 @@ def reset_randomizer(self, seed: int | None = None) -> None:
         Args:
             seed: A new seed for the RNG; if None, reset using the current seed
         """
+        warnings.warn(
+            "the use of seed is deprecated, use rng instead",
```
EwoutH:
```suggestion
            "The use of the `seed` keyword argument is deprecated, use `rng` instead. No functional changes.",
```

PR #3152, opened 2026-01-16
Add unified time and event scheduling API to Model

EwoutH:
### Summary
This PR implements the first phase of the unified time and event scheduling API discussed in #2921. It integrates event scheduling directly into the `Model` class, making it simple to use both traditional time-stepping and discrete event simulation without breaking any existing functionality.

Basically, this PR does two things:
1. Introduce a nicer API for DEVS
2. Letting `Model.step()` run the model for 1 timestep

### Motivation
Currently, users who want to use event scheduling must work with the experimental `Simulator` classes (`ABMSimulator`, `DEVSimulator`), which requires managing two separate objects and understanding the distinction between them. This creates unnecessary complexity for a feature that should be a natural part of Mesa models.

This PR makes event scheduling a first-class citizen by integrating it directly into `Model`, while maintaining full backward compatibility with existing step-based models.

### Usage patterns
This API enables classical ABM, pure event-driven, and hybrid stuff in between. I think 

### 1. Pure classical ABM
Everything happens in `step()`, time advances in integer increments:
```python
class WolfSheep(Model):
    def step(self):

model = WolfSheep()
for _ in range(100):
    model.step()  # t=1, 2, 3, ...
```
This is what most people who don't use a `simulator` do now.

### 2. Hybrid: step + events
Regular stepping with scheduled one-off or agent-triggered events:
```python
class WolfSheep(Model):
    def __init__(self):
        super().__init__()
        self.schedule_at(self.drought, time=50)  # One-off event
    
    def step(self):
    
    def drought(self):
        self.grass_layer.modify_cells(lambda g: g * 0.5)

model = WolfSheep()
for _ in range(100):
    model.step()  # step() at t=1,2,3... + drought fires at t=50
```
This is functionally equivalent to the ABMScheduler, but you still call model.step(). So instead of a scheduler scheduling the step, this works the other way around, the step run the scheduled events (if any).

### 3. Pure event-driven (no step)
No regular stepping, only scheduled events with continuous time:
```python
class QueueingModel(Model):
    def __init__(self, arrival_rate):
        super().__init__()
        self.arrival_rate = arrival_rate
        self.schedule_at(self.customer_arrival, time=0)
    
    def customer_arrival(self):
        Customer(self)
        next_time = self.time + self.random.expovariate(self.arrival_rate)
        self.schedule_at(self.customer_arrival, time=next_time)

model = QueueingModel(arrival_rate=2.0)
model.run_until(1000.0)  # Time jumps between events: t=0, 0.3, 0.8, 1.2, ...
```
This is functionally equivalent to the DEVScheduler.

<hr>

This PR is fully backward compatible. Users can opt-in to new features without modifying existing code.

Curious what everybody thinks of this approach. Being backwards compatible has a lot going for it, as well as the easy intro into scheduling events.

The only disadvantage might be that users could expect the `run()` methods to also perform the step. But I think we can explain that.

We could optionally add a `run_steps()` utility method:

```Python
def run_steps(self, n_steps):
    """Run n steps using the step() method."""
    for _ in range(n_steps):
        self.step()
```

If we agree on this direction, I will add/update tests, the simulator, example models and the tutorial.

Part of #2921.

### DIFF
--- a/mesa/experimental/devs/__init__.py
+++ b/mesa/experimental/devs/__init__.py
@@ -18,7 +18,14 @@ The implementation supports both pure discrete event simulation and hybrid appro
 combining agent-based modeling with event scheduling.
 """
 
-from .eventlist import Priority, SimulationEvent
+from .eventlist import EventList, Priority, SimulationEvent
 from .simulator import ABMSimulator, DEVSimulator, Simulator
 
-__all__ = ["ABMSimulator", "DEVSimulator", "Priority", "SimulationEvent", "Simulator"]
+__all__ = [
+    "ABMSimulator",
+    "DEVSimulator",
+    "EventList",
+    "Priority",
+    "SimulationEvent",
+    "Simulator",
+]
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -8,7 +8,7 @@ from __future__ import annotations
 
 import random
 import sys
-from collections.abc import Sequence
+from collections.abc import Callable, Sequence
 
 # mypy
 from typing import Any
@@ -19,6 +19,7 @@ from mesa.agent import Agent, AgentSet
 from mesa.experimental.devs import Simulator
 from mesa.experimental.scenarios import Scenario
 from mesa.mesa_logging import create_module_logger, method_logger
+from mesa.timeflow import Priority, RunControl, Scheduler, SimulationEvent
 
 SeedLike = int | np.integer | Sequence[int] | np.random.SeedSequence
 RNGLike = np.random.Generator | np.random.BitGenerator
@@ -95,6 +96,10 @@ class Model[A: Agent]:
         # Track if a simulator is controlling time
         self._simulator: Simulator | None = None
 
+        # Add timeflow components
+        self._scheduler = Scheduler(self)
+        self._run_control = RunControl(self, self._scheduler)
+
         # check if `scenario` is provided
         # and if so, whether rng is the same or not
         if scenario is not None:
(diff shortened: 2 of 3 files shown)


EwoutH:
> This is functionally equivalent to the ABMScheduler, but you still call model.step(). So instead of a scheduler scheduling the step, this works the other way around, the step run the scheduled events (if any).

The other approach is that we make `step()` a special method, that runs always on integer time. Then all run() function will schedule a step on each integer time. If you don't want to use the step, you just leave it empty.

EwoutH:
The latter would make for an extremely clean API:

### Pattern 1: Pure classical ABM
```python
class WolfSheep(Model):
    def step(self):

model = WolfSheep()
model.run_for(100)  # step() auto-scheduled at t=1,2,3...
```

### Pattern 2: Hybrid
```python
class WolfSheep(Model):
    def __init__(self):
        super().__init__()
        self.schedule_at(self.drought, time=50)
    
    def step(self):

model = WolfSheep()
model.run_for(100)  # step() at t=1,2,3... + drought at t=50
```

### Pattern 3: Pure event-driven (no step)
```python
class QueueingModel(Model):
    def __init__(self):
        super().__init__()
        # Don't override step(), so it doesn't get auto-scheduled
        self.schedule_at(self.customer_arrival, time=0)
```

quaquel:
I can only comment on the API at the moment; I have no time to review the code at this moment. 

1. I would expect that `model.run_for` etc. would just work out of the box, or would require something minor from my side to make it work. A simple option would be to add an annotation like `@scheduled`. Can you elaborate on what you mean by making `step` a special method? However, if we can avoid this, that would be even nicer.
2. I like the fact that by default, the current approach, where you write your own for loop, still works. 
3. In what way can the user specify the run setup? I would like to have a default run setup (e.g, `time_step=1`, `end_time=100`)

EwoutH:
Succeeded by production PR #3155.

PR #3153, opened 2026-01-16
Replace Computable Descriptor with @computed in mesa_signals

codebreaker32:
This PR modifies `mesa_signals` to move from a class-based descriptor pattern (`Computable`/`Computed`) to a standard Python property factory (`@computed`). 

Currently, accessing a derived property triggers the `Computable.__get__` descriptor, which then fetches a separate `Computed` object to execute the logic. This adds unnecessary function call depth to every attribute access.

I suggest two streamlined components:

1. **`@computed` Decorator:** A factory that returns a standard Python `property`. The "reactive" logic (dependency tracking, dirty checking) is captured in the property's closure, eliminating the descriptor lookup overhead.
2. **`ComputedState`:** A lightweight, internal-only class using `__slots__` to hold the value, dirty flag, and dependencies.
- **`ComputedProperty`:** A custom property class to identify computed properties.

### API Changes (Before vs. After)

**Before (Current)**

```python
class MyAgent(HasObservables):
    wealth = Computable()  # Descriptor

    def __init__(self):
        super().__init__()
        # Manual binding required
        self.wealth = Computed(self._calculate_wealth)

    def _calculate_wealth(self):
        return self.cash + self.assets

```

**After**

```python
class MyAgent(HasObservables):
    @computed
    def wealth(self):  # Clean, standard Python syntax
        return self.cash + self.assets

```
Addresses (2) from #3131

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -72,6 +72,74 @@ That's great! You can just start working on things, reach out to us. Skim to the
 ### I'm neither
 Start with creating your own models, for fun. Once you have some experience, move to the topics above.
 
+## Mesa development process
+Mesa is a library that aims to provide elegant, scalable, flexible, and powerful building blocks to a wide audience of agent-based modellers. When contributing, it helps to understand our development philosophy and process.
+
+### When to open what
+- **Small enhancements or bug fixes**: Opening an issue or directly a PR is often great.
+- **Larger features or uncertain scope**: Open a [Discussion](https://github.com/mesa/mesa/discussions) first. This helps align on the problem and approach before investing significant effort.
+- **When in doubt**: Start with a Discussion. It's always easier to move from discussion to implementation than to rework a large PR.
+
+### Our development process
+For significant features, we typically follow a three-stage process. While not always linear, this flow helps ensure we build the right thing the right way.
+
+#### Stage 1: Problem alignment
+Before jumping into solutions, we first align on the problem:
+
+- **Target audience**: Who is this feature for? What's their experience level and context?
+- **Use cases**: When and why would someone use this? What are they trying to accomplish?
+- **Existing solutions**: What do other libraries offer? What can we learn from or connect to?
+- **Gap analysis**: Given existing solutions (both in Mesa and elsewhere), where are the gaps?
+
+This stage answers the fundamental question: *Is this problem worth solving, and are we the right ones to solve it?*
+
+#### Stage 2: Conceptual/high-level design
+Once we agree on the problem, we design the solution at a high level:
+
+- **User API**: What will the interface look like from a user's perspective? How will modellers interact with this feature?
+- **Core architecture**: What data structures and patterns will we use internally?
+- **Integration**: How does this fit with existing Mesa components?
+
+The goal is to wrap our heads around the conceptual approach before diving into code.
+
+Sometimes the answer is that we just need better documentation. Python itself is a powerful tool and there are many existing libraries in the ecosystem to leverage.
+
+#### Stage 3: Implementation
+With alignment on both problem and design, implementation can proceed:
+
+- Write the code following Mesa's standards
+- Include/update tests and and check test coverage
+- Include/update documentation and check Readthedocs rendering
+- Open a PR for review
+
+And if needed:
+- Update relevant examples
+- Update/extend the tutorial
+- Write a section for the [migration guide](docs/migration_guide.md)
+
+### Pathfinding PRs
+For complex features where feasibility or proportionality is uncertain, we sometimes use *pathfinding PRs*. These are exploratory implementations meant to:
+
+- Test whether an approach is technically viable
+- Understand the true complexity and scope
+- Gather concrete data (e.g., micro-benchmarks) to inform design decisions
+- Spark discussion with working code rather than abstract ideas
+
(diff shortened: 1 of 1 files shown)


mesa/experimental/mesa_signals/mesa_signal.py
```diff
@@ -248,61 +187,71 @@ def _remove_parents(self):
         # we can unsubscribe from everything on each parent
         for parent in self.parents:
             parent.unobserve(All(), All(), self._set_dirty)
+        self.parents.clear()
```
codebreaker32, line 190: This PR also exposed a subtle but critical bug.

The issue occurs when the system **unsubscribes** from dependencies but fails to **forget** them. Without explicitly clearing the dependency list, a computed property retains stale dependencies it no longer uses.


Consider this example:

```python
@computed
def move(self):
    if self.is_injured:
        return self.crawl_speed
    else:
        return self.sprint_speed
```

**Healthy state**: Depends on: `is_injured`, `sprint_speed`

**Injured state**: Depends on: `is_injured`, `crawl_speed`


### Without `.clear()`

Old dependencies linger:

* `move` keeps listening to `sprint_speed` even when injured(if is_injured=False initially)
* Changes to `sprint_speed` trigger unnecessary recomputation
* CPU wasted, result unchanged

### With `.clear()`

Dependencies are rebuilt on each run:

* `sprint_speed` is dropped when unused
* Only relevant changes trigger recomputation

mesa/experimental/mesa_signals/mesa_signal.py
```diff
@@ -248,61 +187,71 @@ def _remove_parents(self):
...
-                            # trick for breaking cleanly out of nested for loops
-                            # see https://stackoverflow.com/questions/653509/breaking-out-of-nested-loops
-                            continue
-                        break
-                    else:
-                        # one of our parents no longer exists
+            else:
+                for parent, observations in state.parents.items():
+                    if parent is None:
                         changed = True
```
quaquel, line 285: Why did you remove this? Basically, this means you always recompute if `self._is_dirty` is true. This is probably fine if computing is cheap, but might be very wasteful if it is expensive.

mesa/experimental/mesa_signals/mesa_signal.py
quaquel: is this bug now covered by the tests?

mesa/experimental/mesa_signals/mesa_signal.py
```diff
@@ -144,86 +144,25 @@ def __set__(self, instance: HasObservables, value):  # noqa D103
         PROCESSING_SIGNALS.clear()  # we have notified our children, so we can clear this out
 
 
-class Computable(BaseObservable):
-    """A Computable that is depended on one or more Observables.
+class ComputedState:
```
quaquel, line 147: can't this be a simple dataclass?

mesa/experimental/mesa_signals/mesa_signal.py
codebreaker32: I haven’t removed it, I only refactored it. The logic that avoids re-computation (by checking whether values actually changed rather than just being marked dirty) is still present inside the new `@computed` wrapper.
In the old implementation, we iterated manually over `keyrefs()` and used a `for-else` loop trick to manage control flow.
In the new implementation, I use `state.parents.items()`, which automatically handles weak-reference liveness. The dirty-check logic is now:

```python
# Inside the new @computed wrapper:
if state.is_dirty:
    changed = False
    if not state.parents:
        changed = True
    else:
        # This loop performs the exact same "did it actually change?" optimization
        for parent, observations in state.parents.items():
            for attr, old_val in observations.items():
                current_val = getattr(parent, attr)
                if current_val != old_val:
                    changed = True  # Only set changed if values differ
                    break
            if changed:
                break
```

If `changed` remains `true` (for example, when the value flips true→ false → true), we still skip the expensive re-calculation

mesa/experimental/mesa_signals/mesa_signal.py
codebreaker32: It can't be mainly due to 2 reasons:-
1. Since `ComputedState` is instantiated for every single computed property on every agent, even a small per-instance overhead adds up quickly in models with 1k+ agents.
2. By default, `@dataclass(slots=True)` does not create a __weakref__ slot. To get one, you have to do something like adding a dummy field just to force a slot creation or rely on inheritance.

mesa/experimental/mesa_signals/mesa_signal.py
codebreaker32: Now it is

mesa/experimental/mesa_signals/mesa_signal.py
quaquel: thanks for the clarification

quaquel:
I ran some quick tests locally. It seems this implementation is much faster because it eliminates significant overhead. I also think the API is much cleaner. I'll try to review in more detail soon. In the meantime, can you check the code coverage report and see if you can add additional tests to ensure we have close to 100% coverage?

codebreaker32:
>can you check the code coverage report and see if you can add additional tests to ensure we have close to 100% coverage?

Done

quaquel:
Thanks again for this PR. A very nice improvement.

PR #3155, opened 2026-01-16
Add unified event scheduling and time progression API

EwoutH:
### Summary
This PR implements a unified time and event scheduling API for Mesa, making discrete event simulation a first-class feature alongside traditional time-stepping. The new API provides a clean, intuitive interface that supports everything from simple step-based models to advanced discrete event simulations—all within the same framework.

After problem and high-level design agreement in #2921, and many pathfinding PRs (#2928, #2932 and #3152), this one is ready for production.

### Motivation
Currently, Mesa supports two distinct approaches to time advancement. The traditional approach has users call `model.step()` repeatedly in a for-loop. For discrete event simulation, users must work with experimental `Simulator` classes (`ABMSimulator`, `DEVSimulator`), managing two separate objects (model and simulator). This creates a fundamental disconnect: discrete event simulation feels bolted-on rather than integrated into Mesa's core.

This separation creates problems. Users must understand the distinction between simulator types before they can use event scheduling. Combining regular stepping with event scheduling is awkward because they live in different parts of the API. Most importantly, the current design doesn't guide users through a natural progression from simple stepping to advanced discrete event patterns.

This PR unifies both approaches under a single, clean API that lives directly on the `Model` class. Time advancement and event scheduling become complementary features of the same system rather than separate mechanisms. Users can now start with the simplest pattern and progressively adopt more advanced features without rewriting their models.

#### New Module: `mesa/timeflow.py`
The new module introduces two core classes and a decorator:
- **`Scheduler`**: Handles event scheduling with `schedule_at()`, `schedule_after()`, and `cancel()` methods. Automatically discovers and schedules `@scheduled` methods.
- **`RunControl`**: Handles time advancement with `run_until()`, `run_for()`, `run_while()`, and `run_next_event()` methods.
- **`@scheduled` decorator**: Marks methods for automatic recurring execution with custom intervals and priorities.

#### Model Integration
The `Model` class instantiates `_scheduler` and `_run_control` in `__init__` and exposes their methods directly. It detects `@scheduled` methods and skips wrapping them with the traditional `_wrapped_step` mechanism. When `step()` is called in the old for-loop pattern, it issues a `PendingDeprecationWarning` explaining the new approach.

#### Experimental DEVS Updates
The `eventlist.py` gains a `RecurringEvent` class that automatically reschedules itself after execution, solving the weak reference issues that prevented lambda functions from working.

#### Visualization Updates
SolaraViz now detects whether models use `@scheduled` and calls the appropriate run method (`run_for(1)` for new-style models vs `step()` for old-style models), ensuring visualizations work with both patterns.

_Note: I did the bare minimum to get this working, More validation is needed._

### Usage Examples
```python
# Existing pattern, now deprecated (for reference)
class WolfSheep(Model):
    def step(self):

model = WolfSheep()
for _ in range(100):
    model.step()  # Shows PendingDeprecationWarning
```

```python
# Modern stepping with @scheduled
from mesa.timeflow import scheduled

class WolfSheep(Model):
    @scheduled
    def step(self):

model = WolfSheep()
model.run_for(100)  # Clean, integrated time advancement
```

```python
# Hybrid: stepping + events
class WolfSheep(Model):
    def __init__(self):
        super().__init__()
        self.schedule_at(self.drought, time=50)  # One-off event
    
    @scheduled
    def step(self):
    
    def drought(self):
        self.grass_layer.modify_cells(lambda g: g * 0.5)

model = WolfSheep()
model.run_for(100)  # step() at t=1,2,3... + drought at t=50
```

```python
# Agents can also schedule themselves
class Citizen(Agent):
    def get_arrested(self, sentence):
        self.in_jail = True
        self.model.schedule_after(self.release, delay=sentence)
    
    def release(self):
        self.in_jail = False
```

```python
# Pure event-driven (no stepping)
class QueueingModel(Model):
    def __init__(self, arrival_rate):
        super().__init__()
        self.arrival_rate = arrival_rate
        self.schedule_at(self.customer_arrival, time=0)
    
    def customer_arrival(self):
        Customer(self)
        next_time = self.time + self.random.expovariate(self.arrival_rate)
        self.schedule_at(self.customer_arrival, time=next_time)

model = QueueingModel(arrival_rate=2.0)
model.run_until(1000.0)  # Time jumps between events: 0, 0.3, 0.8, 1.2...
```
You can mix and match everything

### Notes
- The implementation maintains full backward compatibility - the old for-loop pattern issues a deprecation warning but continues to work. Most importantly, the design supports progressive enhancement: users can adopt features incrementally without rewriting their models.
- Many of the complexity (especially in `Model`) come from backwards support for both the wrapped step and the experimental simulators. Both can be removed in Mesa 4.0, hugely simplifying it.
- The PR includes new tests covering scheduler, run control, and decorator functionality.
- I updated a single example (Wolf-Sheep) to use the new API
- Migration guide section added

### Review questions
1. I'm in doubt about both `run_while()` and the role of `model.running`. Options:
  - Add a `condition` keyword argument to `run_until()` and `run_for()`
  - Always stop if `model.running` if set `False`.
  - Create a new method that stops any run method

References: #2921

### DIFF
--- a/benchmarks/global_benchmark.py
+++ b/benchmarks/global_benchmark.py
@@ -13,8 +13,6 @@ sys.path.insert(0, os.path.abspath(".."))
 
 from configurations import configurations
 
-from mesa.experimental.devs.simulator import ABMSimulator
-
 
 # Generic function to initialize and run a model
 def run_model(model_class, seed, parameters):
@@ -28,18 +26,14 @@ def run_model(model_class, seed, parameters):
     Returns:
         startup time and run time
     """
-    uses_simulator = ["WolfSheep"]
+    uses_scheduled = ["WolfSheep"]
     start_init = timeit.default_timer()
-    if model_class.__name__ in uses_simulator:
-        simulator = ABMSimulator()
-        model = model_class(simulator=simulator, seed=seed, **parameters)
-    else:
-        model = model_class(seed=seed, **parameters)
+    model = model_class(seed=seed, **parameters)
 
     end_init_start_run = timeit.default_timer()
 
-    if model_class.__name__ in uses_simulator:
-        simulator.run_for(config["steps"])
+    if model_class.__name__ in uses_scheduled:
+        model.run_for(config["steps"])
     else:
         for _ in range(config["steps"]):
             model.step()


EwoutH:
While not for this PR, in https://github.com/mesa/mesa/discussions/2921#discussioncomment-15524497 I look forward to an advanced scheduling API possible on this architecture.

mesa/timeflow.py
```diff
@@ -0,0 +1,333 @@
...
+
+    Can be used with or without parentheses:
+        @scheduled
+        def step(self): ...
+
+        @scheduled(interval=7)
+        def weekly(self): ...
+    """
+
+    def decorator(f):
```
quaquel, line 43:
```suggestion
    @functool.wraps(f)
    def decorator(f):
```

mesa/timeflow.py
```diff
@@ -0,0 +1,333 @@
...
+            scheduler=self,
+            interval=interval,
+            priority=priority,
+            function_args=[],
+            function_kwargs={},
+        )
+        self.event_list.add_event(event)
+
+
+class RunControl:
```
quaquel, line 223: more of a conceptual question, but I suggest we include in this also some additional information on e.g., time unit, and max run time.

EwoutH: I think that would be useful indeed. We have to think about how that's initialized.

Would it fit in a `Scenario`? Makes conceptual sense I think, that a Scenario has a defined max run time (and time unit)?

I want to look at datetime support at some point, this could be part of that.

Other PR though.

mesa/timeflow.py
EwoutH: Could you explain this suggestion to me?

The current `scheduled` decorator doesn't create a wrapper function. It just adds `_scheduled` attributes to the original function and returns it unchanged. So I think since we're not replacing `f` with a new callable, `functools.wraps` wouldn't be needed here.

quaquel: Whenever I see an annotation, I assume we need to use functool.wrap. Can you confirm locally if a docstring is still accessible?

EwoutH: I think the docstring is still accessible because we're returning `f` itself, not wrapping it.

Testing locally, it seems like everything is working. Anything else I should test?

image

quaquel: can you test it in a notebook as well just to be sure? IDE's tend to be more powerfull with this kind of stuff.

EwoutH: Yes, it seems like it also works:

image

EwoutH: So, to summarize (with a little help):

> ### When Do You Need `functools.wraps`?
> **The key distinction:**
> 
> - **Attribute-only decorators** (like `@scheduled`) just tag the original function with metadata and return it unchanged → **No `functools.wraps` needed**
> - **Wrapping decorators** that replace the function with a new callable → **Always use `functools.wraps`**
> 
> ### Attribute-Only Decorator (Current Case)
> ```python
> def scheduled(func=None, *, interval=1.0):
>     def decorator(f):
>         f._scheduled = True  # Just adding attributes
>         return f  # Return original function unchanged
>     return decorator
> ```
> ✅ Docstrings, `__name__`, etc. preserved automatically
> 
> ### Wrapping Decorator (Needs functools.wraps)
> ```python
> import functools
> 
> def log_calls(f):
>     @functools.wraps(f)  # ← CRITICAL: Preserves f's metadata
>     def wrapper(*args, **kwargs):
>         print(f"Calling {f.__name__}")
>         return f(*args, **kwargs)
>     return wrapper  # Return NEW function
> ```
> ❌ Without `@functools.wraps(f)`, the wrapper's `__name__` would be "wrapper", not the original function's name, and the docstring would be lost.
> 
> **Rule of thumb:** If your decorator creates a new function that calls the original (`def wrapper...`), use `functools.wraps`. If you just modify and return the original function, you don't need it.

I learned something thanks! And good to keep checking this stuff now and then.

mesa/timeflow.py
quaquel: I am going back and forth on this. I see good arguments for including it in the scenario, but also good arguments for including it inside the run control. I agree its a future PR.

EwoutH:
I would like to move this forward. @quaquel, anything needed from your side?

Other @mesa/maintainers if you haven't taken a look at PR, please do. Probably the biggest Mesa change since the AgentSet.

**Edit:** I rebased the PR on the latest `main`, but rebases are always tricky. Ideally I don't have to do that again.

quaquel:
> I would like to move this forward. @quaquel, anything needed from your side?

It's a lot of code to review. Can you indicate to me which parts are, in essence, copied from/renamed by devs, and what is new? Anything that is copied over/renamed, I won't need to review as closely.

EwoutH:
I think the following things are most important:

- Scheduler: `schedule_at` and `schedule_after` should be familiar. The `@scheduled` decorator is new.
- RunControl: Quite straight forward. Mostly review method names and API. Like if we need a separate `run_while` or we can add a `condition` to the other `run_` methods.
- RecurringEvent: New, please review
- Model: The intergration into the main Mesa model is worth reviewing. Especially the new API surfaces we're adding.

The functional code is maybe 200 lines, the rest is docs or tests.

tpike3:
@EwoutH This is very cool!

I am sorry I haven't had time to do a thorough review, I have been overwhelmed with life. However, at a glance a couple thoughts:
- With `.step` being deprecated could you update the intro tutorials as well (appreciating this is already a lot of work)  
- with regards to `model.running` I agree with your concern, I suppose it cant be removed without breaking backwards compatibility. My bias would to remove it at some point (Mesa 4) so all time is held in `timeflow`. Wit that I would add both options of a condition in the appropriate run functions and an explicit stop method to give the user options, but that is just my preference. I am not really tied to it.

mesa/experimental/devs/eventlist.py
```diff
@@ -123,6 +126,65 @@ def __lt__(self, other):  # noqa
         )
 
 
+class RecurringEvent(SimulationEvent):
```
quaquel, line 129: I think this is a nice solution to the step problem I had in the `ABMSimulator`

mesa/experimental/devs/eventlist.py
```diff
@@ -123,6 +126,65 @@ def __lt__(self, other):  # noqa
...
+        interval: Time between recurring executions
+        scheduler: Reference to the scheduler to reschedule with
+    """
+
+    def __init__(
+        self,
+        time: int | float,
+        function: Callable,
+        scheduler: Scheduler,
+        interval: int | float,
```
quaquel, line 142: Can `interval` also be a Callable?

mesa/visualization/solara_viz.py
```diff
@@ -509,9 +509,15 @@ def visualization_task():
     @function_logger(__name__)
     def do_step():
         """Advance the model by the number of steps specified by the render_interval slider."""
+        # Check if model uses @scheduled (new unified API) or traditional step
+        uses_scheduled = hasattr(model.value.__class__.step, "_scheduled")
+
         if playing.value:
             for _ in range(render_interval.value):
-                model.value.step()
+                if uses_scheduled:
```
quaquel, line 517: beyond this PR, but looking ahead to mesa 4. Do we want to make use_scheduled the required behavior for solara?

mesa/visualization/solara_viz.py
```diff
@@ -637,9 +646,15 @@ def visualization_task():
 
     def do_step():
         """Advance the model by the number of steps specified by the render_interval slider."""
+        # Check if model uses @scheduled (new unified API) or traditional step
+        uses_scheduled = hasattr(model.value.__class__.step, "_scheduled")
+
```
quaquel, line 651: for mesa 4, we need to find a cleaner solution for this. Basically, the coupling between models and solara is too tight

mesa/model.py
```diff
@@ -162,19 +159,55 @@ def __init__(
             [], random=self.random
         )  # an agenset with all agents
 
+        # Add timeflow components
+        self._scheduler = Scheduler(self)
+        self._run_control = RunControl(self, self._scheduler)
+
+        # Check the class definition, not the instance
+        if hasattr(self.__class__.step, "_scheduled"):
```
quaquel, line 167: is there not a way to identify all methods annotated with @scheduled? It can even be made part of the decorator that it registers it somewhere for easy retrieval.

mesa/model.py
```diff
@@ -162,19 +159,55 @@ def __init__(
...
         """Automatically increments time and steps after calling the user's step method."""
-        # Automatically increment time and step counters
+        # Warn about deprecated manual stepping
+        warnings.warn(
+            "Calling model.step() directly in a loop is deprecated and will be removed in Mesa 4.0."
+            "Use the @scheduled decorator and run methods instead. See: https://mesa.readthedocs.io/latest/migration_guide.html#unified-time-api",
+            PendingDeprecationWarning,
+            stacklevel=2,
+        )
+
```
quaquel, line 184: this answers one of my earlier comments. I agree

mesa/model.py
```diff
@@ -162,19 +159,55 @@ def __init__(
...
+                args=args,
+                kwargs=kwargs,
+            )
 
-        _mesa_logger.info(
-            f"calling model.step for step {self.steps} at time {self.time}"
-        )
-        # Call the original user-defined step method
-        self._user_step(*args, **kwargs)
+            _mesa_logger.info(
```
quaquel, line 199: if you use @method_logger on step this should not be needed to be done separately

mesa/model.py
```diff
@@ -162,19 +159,55 @@ def __init__(
...
+                f"calling model.step for step {self.steps} at time {self.time + 1}"
+            )
+
+            # Run until that scheduled event completes
+            self.run_for(1)
+        else:
+            # Simulator is controlling time, just call the method
+            _mesa_logger.info(
```
quaquel, line 207: Same here. Basically, annotate methods to be logged rather than prior to calling the method. THis keeps logging closer to where it belongs.

mesa/model.py
```diff
@@ -300,3 +333,45 @@ def remove_all_agents(self):
...
+
+    def run_until(self, end_time: int | float) -> None:
+        """Run the model until the specified time."""
+        self._run_control.run_until(end_time)
+
+    def run_for(self, duration: int | float) -> None:
+        """Run the model for a specific duration."""
+        self._run_control.run_for(duration)
+
+    def run_while(self, condition: Callable[[Model], bool]) -> None:
```
quaquel, line 371: I prefer having a seperate `run_while` over including condition in the other run methods. It keeps the implementation of the other methods simple. And running for 500 ticks unless a condition happens before that can easily be included as part of the condition. So it's simple to do via `run_while`.

mesa/timeflow.py
```diff
@@ -0,0 +1,333 @@
...
+    RecurringEvent,
+    SimulationEvent,
+)
+
+if TYPE_CHECKING:
+    from mesa.model import Model
+
+
+def scheduled(
+    func=None, *, interval: int | float = 1.0, priority: Priority = Priority.DEFAULT
```
quaquel, line 31: I agree with the idea of making interval also here a callable or a number

mesa/timeflow.py
```diff
@@ -0,0 +1,333 @@
...
+        self.event_list.remove(event)
+
+    def clear(self) -> None:
+        """Clear all scheduled events."""
+        self.event_list.clear()
+
+    def _setup_scheduled_methods(self) -> None:
+        """Find and schedule all methods decorated with @scheduled."""
+        # Iterate through the class, not the instance
+        for name in dir(self.model.__class__):
```
quaquel, line 181: I would make it part of the behavior of the annotation that it creates this overview, instead of having to figure it out after the fact

mesa/timeflow.py
```diff
@@ -0,0 +1,333 @@
...
+        """Initialize run control.
+
+        Args:
+            model: The Mesa model instance
+            scheduler: The scheduler managing events for this model
+        """
+        self.model = model
+        self.scheduler = scheduler
+
+    def run_until(self, end_time: int | float) -> None:
```
quaquel, line 244: This might be simplified if we had special simulator events. So we could just schedule a stop simulation event at the end time, and you don't need the while loop or the end-time check. 

A stop simulation event would be the lowest-priority event, so all other events at the end time would still be processed.

mesa/model.py
EwoutH: I agree that would be useful. Will look into it.

Note that this specific line is just for backwards compatibility, it will be removed with Mesa 4.

mesa/visualization/solara_viz.py
EwoutH: Agreed. We remove this for Mesa 4.

EwoutH:
> I am sorry I haven't had time to do a thorough review, I have been overwhelmed with life.

No worries, I know how it is, already appreciated!

> - With `.step` being deprecated could you update the intro tutorials as well (appreciating this is already a lot of work)  

Fully agree. I'm committed to doing so, once we get through this stage of design and implementation. 

> - with regards to `model.running` I agree with your concern, I suppose it cant be removed without breaking backwards compatibility. My bias would to remove it at some point (Mesa 4) so all time is held in `timeflow`. Wit that I would add both options of a condition in the appropriate run functions and an explicit stop method to give the user options, but that is just my preference. I am not really tied to it.

I see two options here:

1. `model.running` is a first-class citizen of Mesa, and as soon as it is set `False` it fires an event or something that immediately stops the simulation. We can also use this in the visualization and (an replacement of) `batch_run`. We can also just make it a function, like `.stop()`.
2. We leave this up to the users to define in the `run_()` functions with a custom condition or something.

I think it would make sense to have one central place to stop a model from running, that also might eliminate a lot of API surface. So I'm nudging towards some implementation of 1, but then there should be full support of it. Supporting it in some place and not in others is in my opinion the worst of two worlds.

<hr>

> I really like the idea, the api format you chose, and from what I've seen in your updates of the wolf sheep example, it simplifies the code and removes quite a bit more lines than it adds.

Thanks, appreciated!

> I'm just a bit worried about migration with mesa-llm that work based on step wrappers, which might be difficult to update (maybe it's also the case for other mesa elements), but it should be manageable. I'll try to review the code more line-by line as soon as I have a bit more time.

I can understand. This will require some rewiring on other parts of Mesa to. Two thoughts:
1. Mesa-LLM maybe doesn't need full support for every element of Mesa. We maybe see what a sensible subset is to support.
2. One thing we can consider is adding a "hearth beat", like an event that fires always every `x` time (1 by default). Things like data collection and visualization could then subscribe to this heart beat, to have a sensible default.

mesa/experimental/devs/eventlist.py
EwoutH: Thanks, I was a bit worried when creating it if it was the right place to solve this problem, but I think it turned out well. Glad you agree.

mesa/experimental/devs/eventlist.py
EwoutH: In this PR I haven't implemented that, but I do propose it in https://github.com/mesa/mesa/discussions/2921#discussioncomment-15524497. So in the final API, in my opinion, yes.

mesa/visualization/solara_viz.py
EwoutH: Yes, because there won't be anything else.

mesa/model.py
EwoutH: How would that work exactly? (haven't used it before)

mesa/model.py
EwoutH: See my comment below https://github.com/mesa/mesa/pull/3155#issuecomment-3769885458

mesa/timeflow.py
EwoutH: I see your point. There might be some problem with when this exactly happens, it might happen to late or too early in the model (class) init. Or not and I remember it wrong.

Will take a look at it.

mesa/timeflow.py
EwoutH: Good idea. I think something like that would be useful, see also https://github.com/mesa/mesa/pull/3155#issuecomment-3769885458.

mesa/model.py
quaquel: Just annotate the step method with `@method_logger(__name__)` but that would not include the step and time info, so perhaps less than ideal for this use case. As an aside, it might be an idea to expand the logging with time info once this PR is done. 

I would suggest going to debug level for this.

mesa/timeflow.py
quaquel: yes, I saw your comment there and immediately thought about this remark. 

We could even play with priority then. For run_until etc. you use the lowest, while for condition you might want to use highest. The main thing is whether you want to stop in the middle of a step method of a classic abm or not.

EwoutH: The fact that we now even can discus if we want to stop in the middle of a step (and have that level of control), is quite awesome in itself. We’ve come a long way since the days of `RandomActivation`

wang-boyu:
Thanks for working on this. I think it is awsome.

Based on the discussion (https://github.com/mesa/mesa/discussions/2921#discussioncomment-15524497) I thought there is a core `model.schedule()` method and there are convenience methods such as `model.schedule_at()`, `model.schedule_after()`, etc. Is it not part of this PR?

Similarly does it make sense to have a core `model.run()` function with `model.run_while()`, `model.run_until()` and others as convenient alternatives?

This is to allow users to combine

```python
model.schedule_at(self.condition_1)
model.schedule_after(self.condition_2)
```

into one single `model.schedule(at=self.condition_1, after=self.condition_2)`.

Similarly it would be possible to write `model.run(duration=10, condition=self.condition_3)`, whichever `10` or `self.condition_3` is reached earlier.

In addition, what's the difference between `model.run_for()` and `model.run_until()`?

EwoutH:
Thanks for your insights, some very useful stuff in there!

> Based on the discussion (https://github.com/mesa/mesa/discussions/2921#discussioncomment-15524497) I thought there is a core model.schedule() method and there are convenience methods such as model.schedule_at(), model.schedule_after(), etc. Is it not part of this PR?

This is indeed nog part of this PR. There is a bit of historical bias here in this PR, it was focussed on adding the framework for the new capabilities while keeping both compatibility with the current stepping as with the simulator.

However, my vision for the final API is to have one powerful and flexible `schedule()` method indeed, with convenience methods `model.schedule_at()`, `model.schedule_after()`.

> Similarly does it make sense to have a core model.run() function with model.run_while(), model.run_until() and others as convenient alternatives?

Jan and I have this discussion many times, sometime extensively, and also in #2921. Jan prefers more specialized functions, I prefer more general ones. For both are good arguments. I think we have to weigh this each time.

> This is to allow users to combine
> 
> ```python
> model.schedule_at(self.condition_1)
> model.schedule_after(self.condition_2)
> ```
> 
> into one single `model.schedule(at=self.condition_1, after=self.condition_2)`.
> 
> Similarly it would be possible to write `model.run(duration=10, condition=self.condition_3)`, whichever `10` or `self.condition_3` is reached earlier.

Conceptually, quite clean (with minor remarks), but implementation wise, very hard. How do you know when `condition_3` reached  True? How often do you check? Even if you check every millisecond, it still won't be exact. You need it to fire an event when changed, and thus make it observable. See https://github.com/mesa/mesa/discussions/2921#discussioncomment-15524512.

So, our current consensus is that if you want to do that kind of conditional stuff, just add a schedule() call where ever you check/determine that condition.

> In addition, what's the difference between `model.run_for()` and `model.run_until()`?

```Python
model.run_for(10)    # Runs for exactly 10 time units
model.run_until(10)  # Runs until t=10
```
So if you call this at t=5, the first runs to t=15 and the second to t=10.

pragam-m25:
This is a really nice step forward.

From a user / example-writer perspective, the biggest value I see is that
– agents can now naturally schedule long-running actions and interruptions
– examples no longer need ad-hoc counters or step wrappers

I’m planning to build a small Mesa example that stresses:
• agent actions with duration
• interruption / rescheduling (e.g. jail → release → re-arrest)
• no manual step() loop

I’ll report back if I run into any rough edges or migration pain points.

EwoutH:
For all reviewers: Don’t focus too much on the code right now, based on accumulated insights from the discussion in here and in https://github.com/mesa/mesa/discussions/2921#discussioncomment-15524497, I’m going to do a bit cleaner version based around a central `schedule()` method. There will be utility functions, but maybe not decorators yet, since I’ve come back a bit on the need for them, especially now we have the `RecurringEvent` object we can interact with.

API review and discussion is of course welcome. Think from the perspective of the user.

pragam-m25:
Thanks for the clarification, that makes sense.

From a user / example-writer perspective, a single expressive `schedule()` primitive feels like the right thing to validate first, before adding decorators or additional sugar.

I’ll focus on building a small example around:
– long-running agent actions
– interruption + rescheduling
– pause / resume / stop of recurring behavior

and try to express all of that using the proposed `schedule()` shape.
I’ll report back specifically on:
• what feels natural to write
• what feels awkward or easy to misuse
• where examples or docs would really help users.

EwoutH:
Thanks of all the input. Based on it, I continued development, leading to https://github.com/mesa/mesa/pull/3188

EwoutH:
Thanks for all the insights. This was useful pathfinding (as was #3188).

To be continued in:
- https://github.com/mesa/mesa/discussions/2921#discussioncomment-15562881

Issue #3128, opened 2026-01-12
Do we really need weakrefs in AgentSet?


codebreaker32:
The `AgentSet` class, which serves as the core container for `model.agents` and other groups, is backed by a `weakref.WeakKeyDictionary`. This design choice ensures that if an agent is deleted elsewhere in the code, it is automatically removed from the `AgentSet`, preventing memory leaks and "ghost" agents. However, this safety mechanism imposes a significant performance penalty .When iterating over a `WeakKeyDictionary`, Python must:

1. Iterate over the weak reference objects.
2. Call the weak reference to retrieve the actual object.
3. Check if the object is None (garbage collected).
4. Yield the object.

For the main `Model.agents` container, this overhead is redundant. The Model class explicitly manages the lifecycle of these agents via `register_agent` and `deregister_agent`, and it holds hard references in `self._agents`. Therefore, the agents in `Model.agents` will never be implicitly garbage collected while the model is running. We are effectively paying a performance penalty for a safety feature that is unnecessary in the simulations.

codebreaker32:
I replaced WeakKeyDictionary with dict locally and results are :-

| Model           | Size   | Init time [95% CI]         | Run time [95% CI]          |
|-----------------|--------|----------------------------|----------------------------|
| BoltzmannWealth | small  | 🟢 -15.8% [-21.3%, -9.7%]  | 🟢 -14.7% [-20.5%, -8.3%]  |
| BoltzmannWealth | large  | 🟢 -26.5% [-30.7%, -22.2%] | 🟢 -20.2% [-24.3%, -15.9%] |
| Schelling       | small  | 🟢 -22.4% [-24.5%, -20.4%] | 🟢 -17.8% [-19.9%, -15.7%] |
| Schelling       | large  | 🟢 -25.8% [-29.8%, -21.6%] | 🟢 -21.6% [-22.9%, -20.5%] |
| WolfSheep       | small  | 🟢 -31.1% [-35.8%, -26.8%] | 🟢 -27.3% [-29.0%, -25.6%] |
| WolfSheep       | large  | 🟢 -54.3% [-59.7%, -47.7%] | 🟢 -31.7% [-34.5%, -28.3%] |
| BoidFlockers    | small  | 🟢 -29.2% [-33.9%, -24.8%] | 🟢 -19.1% [-21.9%, -16.1%] |
[...]

PR #3156, opened 2026-01-17
Add data registry

quaquel:
# Summary
Expanding on [an idea](https://github.com/mesa/mesa/pull/3145#issuecomment-3761293330) from #3145 as well as past discussion on data collection, this PR adds a novel data-registry approach to Mesa. This new approach rests on the idea of a `DataSet`. A DataSet contains part of the state of a model at a given instant. 

The key point is that current data collection does too much. With this PR, we separate the getting of the state of part of the model at a given instant from the storage of these states over time. With explicit `DataSet` classes, it's now trivial to extend this if you need your own custom data collection. Another benefit of `DataSet` classes is that we get rid of the complex dict-style configuration of what to collect. Everything can be handled through args for attributes and kwargs for callables. 

This PR adds 
* `DataRegistry`: a dict-like collection of datasets. It is always available via model.data_registry.
* `ModelDataSet`: a dataset for gathering model-level data.
* `AgentDataSet`: a dataset for gathering agent data from an `AbstractAgentSet`.
* `TableDataSet`: a dataset for gathering miscellaneous data, works by adding rows to it.
* `NumpyAgentDataSet`: a Numpy array-based dataset containing agent data for a specified Agent class. 
* A `DataSet` protocol

Datasets gather data from fields. Fields are always strings and assumed to be accessible via attribute access. DataSet does not support lambda functions. If you want to do something like that, use properties or descriptors instead. Data is accessed via `DataSet.data`. `ModelDataSet` and `AgentDataSet` will at that moment gather the data and return it. `NumpyAgentDataSet` will return a view on the numpy array containing the data. This view is always in sync with the attribute values, so this data is not separately gathered on request. `TableDataSet` will return the current list of rows. 

This PR is a first draft, exploring the idea. Feedback is very much welcome. The focus is more on fleshing out the API than on optimizing the code itself.

**API**

```python
class MyModel(Model):

	def __init__(self, rng=None):
		super.__init(self, rng=rng)

		self.model_output = DataRegistry()
		self.model_output.track_agents("agent_data", self.model.agents, fields=["wealth", "health", "age"])
		self.model_output.track_model("model_data", self, fields="average_age")

		self.data_collector = CollectorListener()
```

### DIFF
--- a/mesa/experimental/data_collection/dataset.py
+++ b/mesa/experimental/data_collection/dataset.py
@@ -52,7 +52,12 @@ class BaseDataSet(abc.ABC):
 
     """
 
-    def __init__(self, name, *, fields: str|list[str]|None = None,):
+    def __init__(
+        self,
+        name,
+        *,
+        fields: str | list[str] | None = None,
+    ):
         """Initialize a base data set."""
         self.name = name
 
@@ -134,7 +139,7 @@ class ModelDataSet[M: Model](BaseDataSet):
 
     """
 
-    def __init__(self, name, model: M, fields: str| list[str] | None = None):
+    def __init__(self, name, model: M, fields: str | list[str] | None = None):
         """Init of ModelDataSet."""
         super().__init__(name, fields=fields)
         self.model = model
@@ -168,7 +173,7 @@ class TableDataSet:
 
     """
 
-    def __init__(self, name, fields: str | list[str]| None = None):
+    def __init__(self, name, fields: str | list[str] | None = None):
         """Init."""
         self.name = name
 
@@ -468,7 +473,9 @@ class DataRegistry:
         else:
             raise RuntimeError(f"Dataset '{dataset.name}' already registered")
 
-    def create_dataset(self, dataset_type, name, *args, fields: str|list[str]|None = None, **kwargs) -> DataSet:
+    def create_dataset(
+        self, dataset_type, name, *args, fields: str | list[str] | None = None, **kwargs
+    ) -> DataSet:
         """Create a dataset of the specified type and add it to the registry."""
         dataset = dataset_type(name, *args, fields=fields, **kwargs)
         self.datasets[name] = dataset
@@ -478,12 +485,17 @@ class DataRegistry:
         self,
         agents: AbstractAgentSet,
         name: str,
-        fields: str|list[str]|None = None,
+        fields: str | list[str] | None = None,
     ):
         """Track the specified fields for the agents in the AgentSet."""
         return self.create_dataset(AgentDataSet, name, agents, fields=fields)
 
-    def track_model(self, model: Model, name: str, fields: str|list[str]|None = None,):
+    def track_model(
+        self,
+        model: Model,
+        name: str,
+        fields: str | list[str] | None = None,
+    ):
         """Track the specified fields in the model."""
         return self.create_dataset(ModelDataSet, name, model, fields=fields)
 
@@ -491,7 +503,7 @@ class DataRegistry:
         self,
         agent_type: type[Agent],
         name: str,
-        fields: str|list[str]|None = None,
+        fields: str | list[str] | None = None,
         n: int = 100,
         dtype: np.dtype | type = np.float64,
     ) -> NumpyAgentDataSet:
--- a/tests/experimental/test_dataset.py
+++ b/tests/experimental/test_dataset.py
@@ -252,7 +252,9 @@ def test_numpy_agent_dataset_remove_agent():
     class MyModel(Model):
         def __init__(self):
             super().__init__()
-            self.dataset = NumpyAgentDataSet("test", MyAgent, fields="value", dtype=float)
+            self.dataset = NumpyAgentDataSet(
+                "test", MyAgent, fields="value", dtype=float
+            )
             self.data_registry.add_dataset(self.dataset)
 
             MyAgent.create_agents(self, 5, [0.0, 1.0, 2.0, 3.0, 4.0])
@@ -310,7 +312,9 @@ def test_numpy_agent_dataset_expand_storage():
         def __init__(self):
             super().__init__()
             # Start with small capacity
-            self.dataset = NumpyAgentDataSet("test", MyAgent, fields="value", n=5, dtype=float)
+            self.dataset = NumpyAgentDataSet(
+                "test", MyAgent, fields="value", n=5, dtype=float
+            )
             self.data_registry.add_dataset(self.dataset)
             # Add more agents than initial capacity
             for i in range(20):
@@ -335,7 +339,9 @@ def test_numpy_agent_dataset_property_cleanup_on_close():
     class MyModel(Model):
         def __init__(self):
             super().__init__()
-            self.dataset = NumpyAgentDataSet("test", MyAgent, fields="value", dtype=float)
+            self.dataset = NumpyAgentDataSet(
+                "test", MyAgent, fields="value", dtype=float
+            )
             self.data_registry.add_dataset(self.dataset)
 
     model = MyModel()


codebreaker32:
I did some research for (3) and I found that we technically can, but we probably shouldn't:-

1. `operator.attrgetter` does not support default values. Even If one agent is missing the attribute, entire simulation would crash with an AttributeError. ( if we want to wrap it in a try...except block, we will lose the performance gain we were trying to achieve)
2. attrgetter has a noticeable setup cost. `DataSet` pays this cost only once in `__init__` but AgentSet will have to pay it everytime we call `get`. (For agents < 200, it will be slower than a normal "for" loop)
3. The biggest overhead that will stop `AgentSet` to be vector-fast is because it stores "python objects" which are scattered in heap.

However it can be python-fast(For eg. Supporting both AgentSet and WeakAgentSet discussed in #3128 )

codebreaker32:
One more minor thing that I'd like to address is that `operation.attrgetter` would return scalar when `len(self._args) = 1` which would result in `TypeError` if we tried to `zip` it. So for this special case, we'll have to force the result into tuple

EwoutH:
Thanks a lot for this. Very useful pathfinding.

The separation of extraction (DataSet) from storage and timing/triggering (future work) is helpful. Always chop problems into smaller problems if you can.

From looking back at our original discussions, few questions on how flexible the extraction layer can be:

1. Dynamic agent selection: Right now `AgentDataSet` takes a fixed AgentSet at initialization. Could it accept a callable instead, so membership is re-evaluated each time `.data` is accessed? This would enable tracking things like "all starving agents" where the set changes each step:

```python
# Instead of evaluating once at init
registry.track_agents("starving", model.agents.select(lambda a: a.energy < 10))

# Evaluate fresh each time
registry.track_agents("starving", lambda: model.agents.select(lambda a: a.energy < 10))
```
Of course, this add some complexity that you're not working with a fixed set of agents anymore.

2. Aggregation: Can DataSets reference the data from other DataSets? For example, if I'm already collecting `wealth` from agents, can I create a `gini` DataSet that operates on that extracted wealth data rather than re-extracting it?

```python
wealth_data = AgentDataSet("wealth", model.agents, "wealth")

# Option A: Pass the wealth_data DataSet as the "model"?
gini_data = ModelDataSet("gini", wealth_data, gini=lambda ds: calculate_gini(ds.data))

# Option B: Keep model, but callable references wealth_data?
gini_data = ModelDataSet("gini", model, gini=lambda m: calculate_gini(wealth_data.data))

# Option C: Something else entirely?
```

3. DataSet composition: More generally, is there a pattern for DataSets to depend on other DataSets to avoid redundant extraction and enable data pipelines?

quaquel:
> Dynamic agent selection

This indeed currently does not work (nor does it for the existing data collector). I am actually thinking of making this possible in a custom DynamicAgentDataSet class that listens for agent_registered signals for inclusion and agent_deregistered signals for removal. So no need to filter on every call to `.data`, but make the dynamic AgentSet itself reactive instead.

>Aggregation: Can DataSets reference the data from other DataSets? 
> DataSet composition: More generally, is there a pattern for DataSets to depend on other DataSets to avoid redundant extraction and enable data pipelines

This is related to the `_is_dirty` idea in #3145. At present, we cannot do this yet, but this design has more potential to make it possible. The most promising idea I have is to take a cue from the property layers and continuous space position approach. Basically, the attributes you want to track in an `AgentDataSet`, or even a specialized `NumpyAgentDataSet`, are defined as properties on the agent, which internally provide a view into a numpy array. In this way, the dataset is always in sync with the agent state because they are just two different views on the same data.

codebreaker32:
Just asking this out of curiosity, Is it a good idea to make Dataset observable and then squeeze out a new Class from ModelDataset known as `ComputedDataset`(depends on some parent Dataset eg. gini depends on wealth) whereas ModelDataset will purely keep environmental variables(such as tax_rate, time in boltzzmann). Whenever parent Dataset will change, it will trigger a signal and accordingly ComputedDataset will do its calculation(zero extraction cost)

```python
class ComputedDataSet(DataSet):

    def __init__(
        self,
        name: str,
        parent: DataSet,
        compute_fn: Callable[[dict], Any],
    ):
       self.parent.observe(parent.name, SignalType.CHANGE, self.collect)
```

quaquel:
> Just asking this out of curiosity, Is it a good idea to make Dataset observable and then squeeze out a new Class from ModelDataset known as ComputedDataset

Something like this might be made to work. However, any dataset, by definition, is already a `Computed `because it depends on something else. For example, a `wealth` dataset depends on the `wealth` attributes in the agents. To know that `_is_dirty` is true, the `wealth` dataset would have to subscribe to all agents, and all agents would have to make `self.wealth` observable. The net result is an explosion of signals. In the simple case of the Boltzmann model, any given interaction between two agents results in two signals. So in any given step of the model, you might have to process 200 signals (assuming 100 agents). In contrast, a pull-based approach would involve a single loop over 100 agents.

codebreaker32:
You are right that a pure reactive approach would cause a massive Signal Explosion and a performance nightmare But that is not what I am proposing.
I am proposing a "Smart Pull" architecture that combines the efficiency of "Pull" approach with the modularity of Reactivity.

The Hybrid Flow:

1. Silent Updates: Agents update `self.wealth` normally. They do not emit signals. There is zero overhead here.
2. Trigger (1 Signal per Step): The `AgentDataSet` subscribes only to the global `model.step` signal (or sparse events). It assumes the data is "dirty" at the end of every step.
3. The "Smart Pull": When the step signal fires, `AgentDataSet` performs the standard Pull

Result: We have the memory buffer.
Only after the pull is complete, `AgentDataSet` emits a single CHANGE signal.

`ComputedDataSet` (e.g., Gini) listens to this signal.
It grabs the already-pulled buffer from `AgentDataSet` to calculate the metric.

quaquel:
> You are right that a pure reactive approach would cause a massive Signal Explosion and a performance nightmare But that is not what I am proposing.
I am proposing a "Smart Pull" architecture that combines the efficiency of "Pull" approach with the modularity of Reactivity.

That is a design that indeed makes more sense. However, it is also fragile. Basically, the dataset would become dirty on step. It is clean again after the first call to data. But there is nothing preventing the user to have updates to agents in the dataset afterwards:

```python

agent.shuffle_do("step_a")
self.update_stats() # --> calls .data on some of our DataSets
self.shuffle_do("step_b")
self.update_stats() #  assumes that the data is clean but not true.
```

So, that is why a numpy view style design, as we use for the property layers (i.e., `Cell.elevation`) and `ContinuousAgent.position`, is superior. Any call to `.data` just returns the numpy array. This numpy array, by construction, is always in sync with the relevant attributes in all the agents. No need to bother with a dirty flag or anything. 

However, I want to leave the numpy style agent data set for a future PR to avoid complicating this one too much.

quaquel:
> Dynamic agent selection: Right now AgentDataSet takes a fixed AgentSet at initialization. Could it accept a callable instead, so membership is re-evaluated each time .data is accessed? This would enable tracking things like "all starving agents" where the set changes each step:

@EwoutH, a quick update on this: I added an extra argument to `AgentDataSet` called `select_kwargs`, which, if used, is passed on to the `agentset.select` on every call to `.data`. This does not fully solve the problem and is not particularly efficient, since you iterate over the agents twice (in `agentset.select` and when getting the data). But it does show how it can be done. 

Clearly, a better solution would be to filter while looping for data gathering. But then you would not have the full support that `agentset.select` offers.

Another nice to have would be the ability to sort and select. Say, I want to only gather statistics on the 20 richest agents. This is basically an `agentset.sort().select(at_most=20)` style chaining.

quaquel:
I added a proof of principle for a NumPy array based dataset inspired by how it's done in `ContinuousSpace`. You can see in the Boltzman example how its used. There is 1 big caveat in the current implementation: it does not handle agent removal yet. Also, the code is rough and needs cleaning. 

The approach uses two classes: the poorly named `NumpyAgentDataSet` and `DataField`. `DataField` is a property used on an agent. It is given the name of the table to which it belongs and the name of the attribute (if we use a Descriptor instead of a property, we might not need the latter, but it might be less performant). `NumpyAgentDataSet` is, at its core, a NumPy array and a mapping of agents and attributes to the row and column in the NumPy array. If we were to replace this numpy array with e.g., a pandas dataframe or polars dataframe, we might not even need these mappings (making it perhaps also a bit faster). 

The net result is the following API

```python
class MyAgent(Agent):
	wealth = DataField("wealth_table", "wealth")

class MyModel(Model)

	def __init__(self. scenario=None):
		super.__init__(scenario=scenario)

		self.data_registry = DataRegistry() # this name should become part of the base model class
        self.data_registry.create_dataset(NumpyAgentDataSet, "wealth_table", "wealth")

```

When we now do `self.data_registry["wealth_table"]` returns the numpy array for the active agents, which is always in sync with the agents' actual wealth.

quaquel:
Quick update, I profiled my initial solution for Boltzmann, and it was quite slow (20% slower on run time for boltzman). So I have changed a few things around to make it faster. The updated version is as fast as the Boltzmann reference version.

The updated approach only relies on the still poorly named `NumpyAgentDataSet`. You now pass an agent type to it, and internally, this triggers the addition of a getter and setter property for each `arg in args`. So, there is no longer any need to specify in the agent that a given field is a DataField. Instead, whenever you want to collect data from an agent type, the data set does this for you. (As an aside, we might adopt this approach in the continuous space as well for a nice speed up). 

The resulting API:

```python
class MyAgent(Agent):
    def __init__(self):
        wealth = 1

class MyModel(Model)

    def __init__(self. scenario=None):
        super.__init__(scenario=scenario)

        self.data_registry = DataRegistry() # this name should become part of the base model class
        self.data_registry.create_dataset(NumpyAgentDataSet, "wealth_table", MyAgent, "wealth")

    def calculate_ginit(self):
        agent_wealth = self.data_registry["wealth_table].data. # this still feels clunky.
        ...
```

Since we are now automatically adding properties to an agent type, we might leverage this to achieve some additional speedups. The main challenge in `NumpyAgentDataSet` is keeping the agent indices in sync when agents are added and removed. However, we could update our base agent class to check whether dataset-related properties are defined (e.g., in `__new__`), then check with the data table to request an index. Likewise, we can do the same in `agent.remove()` to inform the table that the agent is being removed and that the index can be wiped. 

This new approach has some additional consequences. Since we are adding properties, all agents in the dataset will have the attribute. So, there is no need to worry about getting an `AttributeError`. This approach stores results in a numpy array. This means that all agent attributes you want to collect must have the same data type. Currently, it's hard-coded to floats, but this could be adapted if needed. However, as long as you rely on a numpy array for storage, the datatype must be the same for all attributes, and you want to limit yourself to numbers only to properly leverage numpy. This approach is designed to gather attributes for a given agent type. It could be extended to take a list of agent types, but I don't know how valuable that would be.

quaquel:
I think with this updated `NumpyAgentDataSet`, and once I work out the remaining wrinkles, there is little need for the original pull-based agent dataset. This, however, depends mainly on how important it is to store heterogeneously typed data in a single data set. If this is deemed essential, I can check the performance with either polars or pandas as the backend storage object. Regardless, the architecture used here resolves the entire `is_dirty` problem.

quaquel:
Open question: in the current data collector, we support attributes and callables. But do we need the callables? Any model callable can just be implemented as a property on the model. Any agent-callable function can likewise be implemented as a property on the agent. Not having to worry about callables might simplify and speed up the data set operations a lot. Any thoughs on this are welcome.

Also, at the moment, we don't have a clean solution yet for operations where one might transform one dataset into a new dataset. For example, we have agent-level data, but I also want to have the mean and variance of each attribute. Of course, this can be implemented via a ModelDataSet and appropriate callables/properties, but might there be a use case for expression this more directly at the data set level?

EwoutH:
> Any model callable can just be implemented as a property on the model. Any agent-callable function can likewise be implemented as a property on the agent. Not having to worry about callables might simplify and speed up the data set operations a lot.

Fully agree, I found throw-away callables / lambdas for collectors doubtful at best. Perfectly fine with asking users to define a function or property.

> Also, at the moment, we don't have a clean solution yet for operations where one might transform one dataset into a new dataset. For example, we have agent-level data, but I also want to have the mean and variance of each attribute.

I think this is very important and worth thinking about as well.

codebreaker32:
>in the current data collector, we support attributes and callables. But do we need the callables? Any model callable can just be implemented as a property on the model. Any agent-callable function can likewise be implemented as a property on the agent. Not having to worry about callables might simplify and speed up the data set operations a lot. Any thoughs on this are welcome.

I strongly agree and believe we should move towards a "strict schema" approach

>Also, at the moment, we don't have a clean solution yet for operations where one might transform one dataset into a new dataset. For example, we have agent-level data, but I also want to have the mean and variance of each attribute. Of course, this can be implemented via a ModelDataSet and appropriate callables/properties, but might there be a use case for expression this more directly at the data set level?

Also, If we ban callables, we **must** provide a structured way to do common math, or the user experience will suffer.
I believe we can introduce **`AggregatorDataSet`** (or `DerivedDataSet`) as you suggested.
Here is a rough design for an aggregation system.

```python
registry.create_dataset(NumpyAgentDataSet, "wealth", MoneyAgent, "wealth")

# Aggregation (Declarative)
# No functions. Just linking strings.
registry.create_aggregator(
    name="wealth_stats", 
    source="wealth", 
    operations=["mean", "std", "min", "max"]
)

```

This class adheres to "No Callables" rule by using **String Mappings** to compiled NumPy functions.

```python
class AggregatorDataSet(DataSet):
    """Calculates statistics from another dataset."""
    
    # Allowed operations map directly to optimized Numpy functions
    OPS = {
        "mean": np.mean,
        "std": np.std,
        "var": np.var,
        "sum": np.sum,
        "min": np.min,
        "max": np.max
    }

    def __init__(self, name, registry, source_name, operations):
        super().__init__(name)
        self.registry = registry
        self.source_name = source_name
        self.operations = operations

    @property
    def data(self):
        # Fetch Source (Zero overhead if using NumpyAgentDataSet)
        source_data = self.registry[self.source_name].data
        
        # Compute Stats (Vectorized)
        result = {}
        for op_name in self.operations:
            func = self.OPS[op_name]
            result[op_name] = func(source_data, axis=0)
            
        return result

```
Now
1. It is impossible to write a buggy aggregation loop. The logic is locked inside `np.mean`.
2. It guarantees the operation happens in C-level NumPy, not Python.

quaquel:
@codebreaker32, could you elaborate a bit on what you mean by a strict schema approach? It's not something I am particularly familiar with. 

Just to clarify, my callables point was on agent and model datasets. In the current collector, you can pass both attributes and callables. I want to simplify that to attributes only. Because of Python properties, you can still do callables, but the accessor is attribute-based. So the internal of the dataset becomes much simpler: its attributes only, so we can easily wrap everything into a single `operator.attrgetter`.

There is a lot I like about your aggregator idea. I think it captures most of what I had in mind. Also, the sketched api is quite clear with `source` and `operations`. I have to think a bit about what else we might need. The obvious stuff is simple numpy operations like mean, min, max, etc. However, we might also want to filter. For example, I might want to know the poorest 10 percent of agents in my model on climate change and poverty. So, at a minimum, we also have a filter or select type operation. I am personally not a fan of string-based operations. It becomes its own mini language that people have to remember, and it is hard to support via autocomplete. Still, your point on errors is well taken. If we implement an aggregator, it would be nice if an error in it didn't fully crash the simulation.

codebreaker32:
In the context of the DataRegistry, a "Strict Schema" means moving from Runtime Discovery to Definite-Time Certainty.

-> Loose Schema (Current DataCollector):
- Logic: "I will iterate through agents and figure out what data they have."
- Mechanism: Relies on `getattr(agent, attr)` inside a loop. Handles missing attributes dynamically.
- Cost: Python interpreter overhead per agent, per step. Impossible to pre-allocate memory because types are unknown.

-> Strict Schema (NumpyAgentDataSet):
- Logic: "I know upfront that MoneyAgent has a wealth column of type float64."
- Mechanism: We validate hasattr(MoneyAgent, 'wealth') once at initialization.
- Benefit: We can pre-allocate the exact contiguous memory block (NumPy array) at start time. 
 
Allowing callables (like lambdas) forces us to keep the "Slow Path" (Python Loop) open. By removing them, we enforce a 100% Vectorized Pipeline. The data flows from Object Property -> NumPy Array -> Aggregator without ever waking up the Python interpreter loop.
```python
registry.create_dataset(
    NumpyAgentDataSet, 
    "wealth", 
    MoneyAgent, 
    "wealth", 
    dtype=np.float64  # <--- The "Strict" definition
)
```
By enforcing a Strict Schema, `NumpyAgentDataSet` becomes a generalized, agent-centric version of `PropertyLayer`. It provides the same zero-copy NumPy storage and vectorized access, but removes the dependency on the Grid system. This might pave the way for deprecating PropertyLayer entirely

>Because of Python properties, you can still do callables, but the accessor is attribute-based

This will always be supported by `AgentDataSet`'s `attrgetter`

codebreaker32:
>I am personally not a fan of string-based operations.

```python
# Type-safe, Autocomplete-friendly
registry.create_aggregator(
    "wealth_stats",
    source="wealth",
    operations=[np.mean, np.std, np.min] 
)
# Internally: key = func.__name__ ("mean", "std")
```

>So, at a minimum, we also have a filter or select type operation


The Pipeline:
Source: Collect raw wealth (All agents).
Filter (View): Create a dynamic mask for "Poorest 10%".
Aggregate: Compute stats on the View.

```python
# The Source (Zero-Copy)
registry.create_dataset(NumpyAgentDataSet, "wealth", MoneyAgent, "wealth")

# The Filter Logic (Vectorized)
# Accepts the whole array, returns a filtered view
def poorest_10_percent(array):
    threshold = np.percentile(array, 10)
    return array[array <= threshold]

registry.create_filter(
    name="poorest_wealth",
    source="wealth",
    function=poorest_10_percent
)

# The Aggregator 
# Calculates stats ONLY on the filtered view
registry.create_aggregator(
    "poverty_stats",
    source="poorest_wealth",
    operations=[np.mean]
)
```
This approach keeps the core AgentDataSet incredibly simple and fast (attributes only), while moving complex logic (filtering/math) into specialized, vectorized components in the Registry.
I am not suggesting the APIs but the direction we might explore

quaquel:
I have done some more work on the numpy DataSet. It now uses a more sophisticated mechanism to assign agents to indices and dynamically resizes the underlying numpy array that stores the attributes. The new approach is a tombstone approach. If an agent is removed, its index is freed and becomes available for reuse. A Boolean index is maintained on the active indices. Also, if the array is too small, the array is expanded. If it is too large or becomes too fragmented (with available slots scattered), it compacts automatically and rebuilds the indices. Lastly, I refined the way the getters and setters work to make them even faster. 

Overall, I really like how everything is coming together. To get a sense of the resulting API in action, take a look at the Boltzmann and Wolf-Sheep model. Boltzmann shows a ModelDataSet being dependent on an Agent-level numpy dataset. Because the latter is always correct, we avoid looping over all agents in the Gini calculation. Wolf-Sheep shows how we can use ModelDataSets with properties (avoiding the callables of the current data collector) and demonstrates the dynamic resizing of the Numpy Agent DataSet in action. 

My next items of business are:
1. Make the numpy agent dataset the default AgentDataSet, and turn the pull-based AgentDataSet that mimics the current collector into a fallback option. This also means ensuring agent IDs are properly stored inside the dataset.
2. Add support for multiple agent classes to the NumpyAgentDataSet. In Wolf-Sheep, `energy` exists on both Wolf and Sheep agents, so it might be nice to be able to gather this in a single dataset if desired.
3. Explore the Aggregator idea further.

EwoutH:
I quickly scanned this PR. Interesting (and impressive) stuff. Especially keeping the `_is_active` in a single 1D boolean array, so you can batch agent addition and removal.

The property assignment also looks like some black magic.

Let me know if/when you're ready for an API review.

Thanks for working on this. Really interesting!

quaquel:
Thanks, at its core, it's what we already do for property layers and continuous space. The main improvement, and the one to backport to continuous space, is the shift to using the tombstone approach with `_is_active`. Before that, I reindexed everything on every agent addition or removal. With this tombstone approach, however, each agent index persists for the agent's lifetime (unless compact is called). This makes the getter and setter faster because I can avoid a dict lookup on every get and set operation. While implementing continuous space back then, I already considered this more sophisticated approach. Some profiling today suggested that it might be beneficial to actually do so.

The property assignment is inspired by some of what @codebreaker32 used for the property factory pattern in a recent PR. Having key references wrapped up in a closure is much faster than doing attribute lookups. If we port this to continuous space, we can even support individual agents across multiple continuous spaces (Something I've been wanting to figure out for a while).

EwoutH:
Really cool! Also glad to see the collaboration between @codebreaker32 and you

mesa/experimental/statistics.py
```diff
@@ -0,0 +1,430 @@
...
+
+        Returns a view (not a copy) when possible for efficiency.
+        """
+        if self._n_active == 0:
+            return np.empty((0, len(self._attributes)), dtype=self.dtype)
+
+        # Boolean indexing - returns only active rows
+        # self._n_slots is the number of slots that have been handed out
+        # some might be inactive now, but there can never be more
+        return self._agent_data[self._is_active[: self._n_slots]]
```
codebreaker32:
```suggestion
        return self._agent_data[:self._n_slots][self._is_active[: self._n_slots]]
```

Current implementation throws an exception
```plaintext
IndexError: boolean index did not match indexed array along axis 0; size of axis is 12800 but size of corresponding boolean axis is 8039
```

quaquel: I'll take a look later today. I don't think your suggestion is the most efficient fix

codebreaker32: You might add a check before this
```python
if self._n_active == self._n_slots:
        return self._agent_data[:self._n_slots]
```

quaquel: what are you testing? In my machine, boltzmann and wolf-sheep both work fine

codebreaker32: I was testing integration of listener and data_registry for schelling when I got this error

```text
IndexError: boolean index did not match indexed array along axis 0; size of axis is 12800 but size of corresponding boolean axis is 8039
Collection failed: dataset='agent_data', step=3: boolean index did not match indexed array along axis 0; size of axis is 12800 but size of corresponding boolean axis is 8039
  File "/home/aman/mesa/mesa/experimental/collector/collectorlistener.py", line 257, in _on_step_change
    data_snapshot = dataset.data
                    ^^^^^^^^^^^^

  File "/home/aman/mesa/mesa/experimental/statistics.py", line 369, in data
    return self._agent_data[self._is_active[: self._n_slots]]
```

quaquel: It's a rather strange error to get on Schelling. Schelling has a fixed number of agents, just like Boltzmann.

quaquel: Ok, I just committed a small update that should fix the problem. There is probably some microoptimization stuff that could be done here, but the fix I just pushed should be fine for now. 

I still don't understand why it would happen on Schelling.

codebreaker32: It happens on Schelling because of the Growth Factor, not because of agent removal.
When Schelling initializes ~8,000 agents, the NumpyAgentDataSet capacity doubles until it hits 12,800(2<sup>7</sup> *100).
Buffer Size: 12,800
Active Mask Size (_n_slots): ~8,000

The crash occurs because the code tries to apply the shorter mask (8k) directly to the longer buffer (12.8k): self._agent_data[mask]. Numpy requires the boolean mask dimension to match the array dimension exactly, or for the array to be sliced first.
It will most likely happen in other models also

When I switched to track_agents, the error was gone

quaquel: > When I switched to track_agents, the error was gone

That makes sense because that still defaults to the normal AgentDataSet.

codebreaker32: Ok I figured out, I didn't use `n` to initialise population. Sorry for that.

quaquel: If the stuff is well implemented, it should auto-resize, and everything should continue to work. So, you still ran into a real bug. And with my last commit, that should have been fixed.

Still, it also highlights the need for proper test coverage.

codebreaker32: Sorry but I couldn't see your latest commit

quaquel:
With the progress on #3212, I realized that the current TableDataSet can, in principle, be largely replaced with a more reactive version. In my view, the main use case for the current `Table` in the `DataCollector` is to gather time-independent statistics from agents. For example, you might want to track your agents' life expectancy. In the table, add a row with the agent's age at the time of death. Or you might want to track how long an agent was sick in a pandemic model. Again, you can write this to a table whenever the agent is cured. 

With the move towards making signals more central in Mesa, all of this can be expressed much more naturally. For example, you can have a reactive dataset that subscribes to `ModelSignals.agent_removed` and, upon removal, extracts relevant attributes from the agent. Likewise, you might make your Agents reactive by having a table subscribe to `AgentSignals.RECOVERED` and handle the relevant attribute retrieval when this signal is received.

The advantage of such a reactive table is that you can implement your model without adding code to your agents for statistics (i.e., `DataCollector.tables[my_table].add_row`). Instead, you only add the relevant signals to your agents, while keeping the entire data-gathering specification inside the dataset.  The net result is a much cleaner separation of concerns. And code related to statistics gathering does not pollute your agents' simulation logic.

codebreaker32:
Hi @quaquel,

I have reviewed the latest changes and testing the implementation locally. It looks really solid!
I specifically wanted to confirm that the integration with the listener mechanism is working perfectly. The data collection is syncing up with the signal updates exactly as expected, and the new architecture feels very robust.

Thanks for all the hard work on this!

codebreaker32:
I have a small question regarding `NumpyAgentDataSet`:

Since it monkey-patches the global `Agent` class (using `setattr` on `self.agent_type`), it breaks isolation between model instances. This creates conflicts if models run **sequentially** (old agents incorrectly read from the new model's array) or in **parallel** (models overwrite each other's class properties).

Are we looking to solve this, or is it an accepted trade-off for performance? I am concerned that relying on monkey-patching will create significant blockers for multiprocessing support. Do you have anything in mind?

quaquel:
> Since it monkey-patches the global Agent class (using setattr on self.agent_type), it breaks isolation between model instances. This creates conflicts if models run sequentially (old agents incorrectly read from the new model's array) or in parallel (models overwrite each other's class properties).

Its a bit more complicated. 

1. At the agent class level, we add the name of the dataset to a `_dataset` attribute
2. At the agent class level, we add a property for each field. It is true that the getters and setters of this property contain a reference to a specific dataset instance in their closure.
3. agent instances then do `self.model.data_registry[dataset].add_agent(self)`. So they register themselves with each of the items in `_dataset`. Since this runs through `self.model`, we don't run the risk you describe. Agents can only ever register themselves to a dataset that is part of the same model.
4. Whenever an agent instance calls `add_agent`, we add the index in the numpy dataset to the agent instance dict.
5. On `close()`, we remove the property set in 2 (and we should remove the dataset entry I now realize).

The bigger issue is properly closing datasets when running experiments. In fact, I am increasingly inclined to add a specific `model.close()` method that would internally call' remove_all_agents' and' data_registry.close() '. This would offer stronger guarantees of proper cleanup. 

I am not yet convinced the current design is ideal. At the same time, at least for the properties, I don't see another solution. Properties only exist at the class level, so there is no avoiding doing this. And if you run the same model sequentially, each new model instance will override the property, resolving the problem.

codebreaker32:
Thanks for the detailed explanation.

> 2. At the agent class level, we add a property for each field. It is true that the getters and setters of this property contain a reference to a specific dataset instance in their closure.
> 3. agent instances then do `self.model.data_registry[dataset].add_agent(self)`. So they register themselves with each of the items in `_dataset`. Since this runs through `self.model`, we don't run the risk you describe. Agents can only ever register themselves to a dataset that is part of the same model

Understood
You are right that the new model works perfectly because it overwrites the property.

But that overwrite is exactly the problem for past results. If I have a list results = [model_1, model_2], and I try to plot model_1's data after model_2 has started, model_1's agents will mistakenly read from model_2's array because the class property was updated.
We essentially break the previous models in memory the moment a new one starts.

I think we can solve this by making the property context-aware. Instead of hard-coding the array in the closure, the getter can look up the correct dataset via the agent's model instance.

**Current**

```python
# Captures 'data' (the specific array for this dataset) forever
data = self._agent_data 
def getter(agent):
    return data[agent.__dict__[index_attr], j]

```

Can we do something like this?

```python
# Captures only the dataset name
dataset_name = self.name 

def getter(agent):
    # Dynamically find the dataset belonging to THIS agent's model
    # Access the agent's model
    # Find the dataset by name in that model's registry
    dataset = agent.model.data_registry[dataset_name]
    
    # Read from the correct array
    return dataset._agent_data[agent.__dict__[index_attr], j]

```

This ensures `model_A` always reads from `model_A`, even if `model_B` is running. It adds a slight lookup overhead but guarantees safety for sequential run and interactive analysis.

quaquel:
> But that overwrite is exactly the problem for past results. If I have a list results = [model_1, model_2], and I try to plot model_1's data after model_2 has started, model_1's agents will mistakenly read from model_2's array because the class property was updated.

Yes, but only if you refer to the model's state at a given instant rather than a snapshot of it. In an experimentation context, I would run a sequence of scenarios. Each scenario is combined with a new model instance. And we gather data using the collector for each experiment. So if you want to plot afterwards, you are more likely to plot the data as collected by the Collector.  


> Can we do something like this?

Yes, but it adds quite a bit of overhead. It would involve doing something like `agent.model.data_registry[dataset_name]._agent_data` on each access. Another option might be to store the reference to `_agent_data` in the instance dict, as we already do with the index. However, even that would add dict lookup costs on each getter or setter call.

codebreaker32:
> Yes, but only if you refer to the model's state at a given instant rather than a snapshot of it. In an experimentation context, I would run a sequence of scenarios. Each scenario is combined with a new model instance. And we gather data using the collector for each experiment. So if you want to plot afterwards, you are more likely to plot the data as collected by the Collector.

Fair enough. Thanks

codebreaker32:
I ran into one more issue while testing NumpyAgentDataSet

Using `track_agents()` (without explicitly mentioning unique_id)
<br>
Screenshot (184)

Using `track_agents_numpy()`(without explicitly mentioning unique_id)
<br>
Screenshot (185)

After mentioning explicitly
```python
self.data_registry.track_agents_numpy(
            MoneyAgent,
            "agents_wealth",
            "wealth",
            "unique_id",
            n=self.num_agents,
        )
```
I got
```plaintext
  File "/home/aman/mesa/mesa/examples/basic/boltzmann_wealth_model/agents.py", line 20, in __init__
    super().__init__(model)
  File "/home/aman/mesa/mesa/agent.py", line 71, in __init__
    self.unique_id = None
    ^^^^^^^^^^^^^^
  File "/home/aman/mesa/mesa/experimental/data_collection/dataset.py", line 261, in setter
    data[agent.__dict__[index_attr], j] = value
         ~~~~~~~~~~~~~~^^^^^^^^^^^^
KeyError: '_index_datatable_agents_wealth'
```

codebreaker32:
I think this is because NumpyAgentDataSet installs a property setter on the Agent class. `Agent.__init__` triggers this setter (via self.unique_id = None) before the agent is registered with the dataset (i.e., before add_agent assigns it an index)

quaquel:
> I think this is because NumpyAgentDataSet installs a property setter on the Agent class. Agent.__init__ triggers this setter (via self.unique_id = None) before the agent is registered with the dataset (i.e., before add_agent assigns it an index)

Yes I still have a fixme for unique_id. It cannot, by default, be included in the numpy array because unique_id is an int but if your attributes are bool, you have an error. And the issue you report here is another issue I had not even realized but makes sense. The solution to both is probably to maintain a seperate agent_id vector and only put stuff in here via add_agent. It is quite simple to maintain the order of this vector by applying any reindexing on the core numpy array also to this vector.

quaquel:
This PR is ready for review although there are still a few minor loose ends that I willl clean up over the weekend.

1. ~I still need to include agent_id in the NumPyAgentDataSet~
2. There is some legacy code at the end of `core.py` used for testing that will be removed
3. ~Coverage is virtually complete, but I might add a few more tests where needed.~

quaquel:
I added the unique_ids of agents as a separate numpy array so this is now included. However, I have left `.data` the same for now. In many cases, you only want the data and are not interested in the unique_id's of the agents to which this data belongs. 

For the data recording, unique_ids do matter. Ideally, you want to store agent data by unique_id so you can trace agents over the simulation. @codebreaker32, I would love your perspective on this and how you think we could include this in the API. There is also a broader point on `DataSet.data` being heterogeneously typed which is unavoidable but might still benefit from some cleaning.

codebreaker32:
I recommend we keep `.data` pure (values only) to preserve the ease of mathematical operations (e.g., `np.mean(dataset.data)`), and instead expose IDs via a separate interface.

* **The Dataset:** We add a public property `dataset.ids` (or `dataset.index`) to `NumpyAgentDataSet` that returns the `_agent_ids` array corresponding to the active rows.
* **The Listener:** We update `CollectorListener` to handle this side-channel data. Since the listener has access to the registry, it can do something like:

```python
# inside CollectorListener._store_dataset_snapshot
dataset = self.registry.datasets[name]
data = dataset.data

if hasattr(dataset, "ids"):
    ids = dataset.ids
    # Stack IDs(using np.hstack) with Data for storage

```

This way, the `DataSet` remains a fast numerical container, and the `Listener` handles the complexity of joining IDs for storage.

>There is also a broader point on DataSet.data being heterogeneously typed which is unavoidable

Agree and rather I see "Listener" as the normalization layer.

quaquel:
> I recommend we keep .data pure (values only) to preserve the ease of mathematical operations (e.g., np.mean(dataset.data)), and instead expose IDs via a separate interface.

I agree with this. I'll add a `NumpyAgentDataSet.agent_ids` property.

> Agree and rather I see "Listener" as the normalization layer.

Ok, for now, let's keep it this way. We might revisit this depending on if and how we want to use the DataRegistry in the UI side of things.

EwoutH:
From the usage example in the PR description:
> 		self.model_output.track_agents("agent_data", self.model.agents, "wealth", "health", "age", )
> 		self.model_output.track_model("model_data", self, "average_age")

I find it a bit weird you can just pile one arguments. Can we make this a list or a set?

I think requiring keywords here might also help (including for future API changes).

quaquel:
> I find it a bit weird you can just pile one arguments. Can we make this a list or a set?
> I think requiring keywords here might also help (including for future API changes).

I actually thought it was very convenient. You are just passing the different attributes/properties/descriptors you want to collect. So yes, you could also do this via e.g., a single argument `fields: str|iterable[str]`.

quaquel:
So, before merging this, I would like to know if there is a preference regarding fields. We have three options

```python

# current implementations
registry.track_model("model_dataset", "attr1", "attr2", "attr3")

# have a single fields argument
registry.track_model("model_dataset",  ["attr1", "attr2", "attr3"])

# have a single fields keyword argument
registry.track_model("model_dataset",  fields=["attr1", "attr2", "attr3"])
```

@EwoutH, @codebreaker32 do either of you have a clear preference? I like the convenience of the current implementation, but I see @EwoutH's point of future extendability. We also separately indicated a desire to move towards a keyword-preferred design, which would favor option 3.

codebreaker32:
I am fine with option 3 as well. It is self-documenting, extensible, and aligns with the project's shift toward explicit keyword arguments

quaquel:
I added NumpyAgentDataSet.agent_ids and shifted to fields as a keyword argument. I am merging this so we can move on to finalizing #3145.

PR #3157, opened 2026-01-17
Fix select_cells into boolean/index methods and fix its tests

champ-byte:
### Summary
It resolves fixme `consider splitting into two separate functions
        #  select_cells_boolean
        #  select_cells_index` inside PropertyLayer and HasPropertyLayer
   
It also resolves fixme in tests: `fixme, add an agent and update the np.all test accordingly` , ` # fixme add extreme_values highest and lowest`, `fixme add pre-specified mask to any other option`

### Implementation

Changes:

- Split `PropertyLayer.select_cells` into `select_cells_boolean` (returns mask) and `select_cells_index` (returns coordinates).

- Split `HasPropertyLayers.select_cells` into `select_cells_boolean` and `select_cells_index`.

- Retained `select_cells` as a deprecated wrapper with a FutureWarning to ensure backward compatibility.

Test Fixes :

- Resolved FIXME for only_empty=True: Added a test case with an agent present to verify occupied cells are correctly excluded.

- Resolved FIXME for masks: Added a test case using a pre-specified boolean mask.

- extreme_values coverage .

### Testing

- I have modified the `test_select_cells` to use the new methods along with addressing the fixme
- Added `test_select_cells_deprecation` to check for warning
- Added `test_property_layer_select_cells_split` which tests the underlying PropertyLayer under isolation

### Additional Notes
I have just currently added a future warning to `select_cells`, in order to follow mesa`s deprecation policy,  and not completely removed it.
- I have not currently made any changes to the docs and overview , do you want me to make changes in them ?

Please let me know if there’s anything I missed or if further adjustments are needed, I would be happy to update it.


quaquel:
We are considering removing the `PropertyLayer` class altogether in Mesa 4. So I am holding off on reviewing this PR until we have made a decision on that.

quaquel:
I am closing this PR. It seems highly likely that we'll remove the property layer class in Mesa 4 in favor of directly working with numpy arrays. If this changes, we can always revisit this.

PR #3158, opened 2026-01-17
Add parameter access and copy support to Scenario

navneetkumaryadav207001:
### Summary
Adds two utility methods to the `Scenario` base class: `get_parameters()` for accessing scenario parameters without metadata, and `copy()` for creating modified scenario variants.

### Motive
Working with experiment sweeps, logging, and reproducibility often requires separating scenario parameters from internal metadata and creating slightly modified copies of scenarios. This change provides a simple, consistent API for both use cases.

### Implementation
- Added `get_parameters()` to return scenario attributes excluding `model` and `_scenario_id`.
- Added `copy(**updates)` to create a new `Scenario` instance with optional parameter updates.
- The copied scenario:
  - gets a new `_scenario_id`
  - does not retain the `model` reference
  - shares the same RNG object by reference (documented).

### Usage Examples
```python
params = scenario.get_parameters()

new_scenario = scenario.copy(alpha=0.5, beta=10)
```

solves #3159.

### DIFF
--- a/tests/experimental/test_scenarios.py
+++ b/tests/experimental/test_scenarios.py
@@ -78,6 +78,23 @@ def test_scenario_serialization():
     assert unpickled._scenario_id == scenario._scenario_id
 
 
+def test_copy_creates_new_scenario_with_updates():
+    """Test copy creates a new scenario with updated parameters."""
+    Scenario._reset_counter()
+
+    scenario = Scenario(a=1, b=2, rng=42)
+    copied = scenario.copy(b=3)
+
+    # Parameters
+    assert copied.a == 1
+    assert copied.b == 3
+    assert copied.rng is scenario.rng  # RNG shared by reference
+
+    # Metadata
+    assert copied.model is None
+    assert copied._scenario_id != scenario._scenario_id
+
+
 def test_agent_scenario_property():
     """Test that agents can access scenario via property."""
     scenario = Scenario(test_param=100, another_param="test", rng=42)


mesa/experimental/scenarios/scenario.py
```diff
@@ -80,6 +80,25 @@ def to_dict(self):
         """Return a dict representation of the scenario."""
         return {**self.__dict__, "model": self.model, "_scenario_id": self._scenario_id}
 
+    def get_parameters(self):
+        """Get scenario parameters excluding metadata."""
+        excluded = {"model", "_scenario_id"}
+        return {k: v for k, v in self.__dict__.items() if k not in excluded}
+
```
quaquel: redundant with the already existing `to_dict` method.

mesa/experimental/scenarios/scenario.py
```diff
@@ -80,6 +80,25 @@ def to_dict(self):
...
+        Note: Creates a new scenario with a unique _scenario_id.
+        Note: rng is copied by reference, not cloned. Both scenarios will share
+              the same RNG object and produce identical random sequences.
+        """
+        params = self.__dict__.copy()
+        params.update(updates)
+        params.pop("model", None)  # Don't copy model reference
+        params.pop("_scenario_id", None)  # Create new scenario with new ID
+
+        return self.__class__(**params)
```
quaquel: Not sure if this is needed either because you have `to_dict`.

Can you give me an example of when you would need this?

PR #3160, opened 2026-01-17
Introducing _StrongAgentSet to support strong references in agents.py

### LINKED ISSUE #3128: Do we really need weakrefs in AgentSet?

codebreaker32:
Problem:
1. Iteration Overhead: Profiling indicated that `AgentSet` iteration was unnecessarily expensive due to WeakKeyDictionary. Every iteration required a liveness check for every agent. Since `model.agents` (the central registry) relies on AgentSet, this overhead impacted every step of the simulation.

This PR solves them by:-
1. Introduced `_StrongAgentSet`, a subclass of `AgentSet` that uses a standard dict for storage instead of `WeakKeyDictionary`.
2. Implemented automatic downgrading so that copies or views (e.g., from .select()) return a standard `AgentSet` (weak references), preventing memory leaks in user space.

Backward Compatibility: No breaking changes to the public API signature.

Closes #3128

### DIFF
--- a/mesa/experimental/devs/eventlist.py
+++ b/mesa/experimental/devs/eventlist.py
@@ -125,6 +125,28 @@ class SimulationEvent:
             other.unique_id,
         )
 
+    def __getstate__(self):
+        """Prepare state for pickling."""
+        state = self.__dict__.copy()
+        # Convert weak reference back to strong reference for pickling
+        fn = self.fn() if self.fn is not None else None
+        state["_fn_strong"] = fn
+        state["fn"] = None  # Don't pickle the weak reference
+        return state
+
+    def __setstate__(self, state):
+        """Restore state after unpickling."""
+        fn = state.pop("_fn_strong")
+        self.__dict__.update(state)
+        # Recreate weak reference
+        if fn is not None:
+            if isinstance(fn, MethodType):
+                self.fn = WeakMethod(fn)
+            else:
+                self.fn = ref(fn)
+        else:
+            self.fn = None
+
 
 class EventGenerator:
     """A generator that creates recurring events at specified intervals.
--- a/tests/experimental/test_devs.py
+++ b/tests/experimental/test_devs.py
@@ -273,6 +273,45 @@ def test_simulation_event():
     assert event1 > event2
 
 
+def test_simulation_event_pickle():
+    """Test pickling and unpickling of SimulationEvent."""
+
+    # Test with regular function
+    def test_fn():
+        return "test"
+
+    event = SimulationEvent(
+        10.0,
+        test_fn,
+        priority=Priority.HIGH,
+        function_args=["arg1"],
+        function_kwargs={"key": "value"},
+    )
+
+    # Pickle and unpickle
+    state = event.__getstate__()
+    assert state["_fn_strong"] is test_fn
+    assert state["fn"] is None
+
+    new_event = SimulationEvent.__new__(SimulationEvent)
+    new_event.__setstate__(state)
+
+    assert new_event.time == 10.0
+    assert new_event.priority == Priority.HIGH.value
+    assert new_event.function_args == ["arg1"]
+    assert new_event.function_kwargs == {"key": "value"}
+    assert new_event.fn() is test_fn
+
+    # Test with canceled event
+    event.cancel()
+    state = event.__getstate__()
+    assert state["_fn_strong"] is None
+
+    new_event = SimulationEvent.__new__(SimulationEvent)
+    new_event.__setstate__(state)
+    assert new_event.fn is None
+
+
 def test_eventlist():
     """Tests for EventList."""
     event_list = EventList()


quaquel:
Would you mind separating this into 2 PRs? The agent creation one will be straightforward to merge. 

The agent set stuff is more difficult. I don't want to change the internal weakkeydict to a dict while still under mesa 3.X. It might cause all kinds of unexpected results in existing models that rely on agentset having a weakkeydict. I'm also currently thinking of keeping the existing AgentSet as a weakkeydict, adding a separate dict version used only internally for `model.agents` and `model.agents_by_type`, and removing model._agents (because it's redundant from there on out). But even that might not be fully backward compatible because people might use a copy operation on their agentset. What is also poorly defined at the moment is the behavior of all methods that return an agentset.

codebreaker32:
I have reverted the creation change.

>I'm also currently thinking of keeping the existing AgentSet as a weakkeydict, adding a separate dict version used only internally for model.agents and model.agents_by_type

What if we can keep AgentSet as it is and inherit a new subclass `StrongAgentSet`(or some better name) for dict-only purposes. It'd be more cleaner and all the methods of AgentSet would be still available to it

quaquel:
> What if we can keep AgentSet as it is and inherit a new subclass StrongAgentSet(or some better name) for dict-only purposes. It'd be more cleaner and all the methods of AgentSet would be still available to it

Yes that is a logical next step. 

The main issue however is how to properly integrate this into mesa for `model.agents` and `model.agents_by_type`. Clearly, for performance, we want to use normal dicts here. And they are controlled by `model.register_agent` and `model.deregister_agent`, which in turn is tied to `agent.remove`. So, memory leaks here are not a concern. 

However, what happens if someone makes a copy of these hard ref agent sets?. This can happen via any of the methods with an `inplace` keyword argument. If you say `inplace=False`, you'll get a copy. And now we create the potential for a memory leak because these copies are not controlled by `model.deregister_agent`. 

I see two solutions:
1. Don't allow `inplace` on any of the methods of the hard ref agent set. This avoids the implict copy and forces a user to use `copy.copy`. Next, we can make it clear in the docstring that making a copy is a bad idea due to the potential for memory leaks. 
2. Allow inplace, but always return the current AgentSet (so with weakkeydict). It keeps the API the same between both; however, it implicitly changes the datatype of the object you are dealing with.

codebreaker32:
>Allow inplace, but always return the current AgentSet (so with weakkeydict). It keeps the API the same between both; however, it implicitly changes the datatype of the object you are dealing with.

The implementation I suggested **already addresses this naturally**. We don't need to write any extra logic to enforce it; the inheritance structure handles it for us.

We define `_StrongAgentSet` to strictly override the storage mechanism:

```python
class _StrongAgentSet(AgentSet):
    """Internal optimization for Model.agents. Uses strong refs for speed."""
    
    def _storage(self, agents):
        return dict.fromkeys(agents)

```

In the base `AgentSet` class, methods like `select`, `shuffle`, and `sort` explicitly call the `AgentSet` constructor for their return values

```python
def select(self, ...):
    # ...
    return AgentSet(agents, self.random)  # <--- Explicitly returns a WEAK AgentSet

```

1. **Input:** You call `model.agents.select(...)` on the optimized `_StrongAgentSet` (Internal/Strong).
2. **Output:** The method runs, filters the agents, and returns a standard `AgentSet` (Public/Weak).

This automatically creates the safety boundary you described:

* **Internal Engine (`model.agents`):** Fast, Strong Refs.
* **External Views (return values):** Safe, Weak Refs.

For extra safety, we can explicitly override `__copy__`, to ensure that Python's standard copy operations also respect this "downgrade" to weak references.

quaquel:
Thanks for the clarification. This PR currently subclasses the weakkey version from the hard key version. I suggest changing this around. 

Moreover, there are many optimizations inside the current agent set when we iterate. These really need to stay in place because they have a profound impact on performance. From memory, the one in `shuffle` is the most critical, but the others also matter. Basically, we cannot just change the type of the dict used internally and keep everything else the same.

codebreaker32:
>Basically, we cannot just change the type of the dict used internally and keep everything else the same.

You are completely right, we cannot just swap the storage backend and expect it to work. The current AgentSet implementation is tightly coupled to WeakKeyDictionary specific APIs (like .keyrefs() and .data), which don't exist on a standard dict.

To make _StrongAgentSet work, We have to do more than just change _storage; Instead have to override every method that relies on those weakref internals.

codebreaker32:
I’ve implemented a working version where _StrongAgentSet passes all safety checks and correctly downgrades copies to weak sets. Happy to leave the final merge decision to you

quaquel:
Can you update the title to reflect what this PR does.

I'll try to review this tomorrow. A first quick look suggests most if fine.

What are your thoughts on updating the internals of Mesa to start using this?

codebreaker32:
>What are your thoughts on updating the internals of Mesa to start using this?

In my opinion, this change isn't introducing strong ownership; it is optimizing the `model.agents` to match the strong ownership that already exists, removing the redundant weak-reference overhead.

I think updating `model.agents` to use `_StrongAgentSet` is the right move because:
1. Conceptually, the Model is the "source of truth" and owner of the agents. In the current implementation, `model.py` already enforces this ownership via the internal `_agents` dict. However, the public `model.agents` (currently a standard `AgentSet`) wraps these in `WeakKeyDictionary`, adding unnecessary overhead to check for liveness that the Model already guarantees. Switching to `_StrongAgentSet` removes this redundancy and optimizes the hot loop.

2. Since the Model already holds strong references (in `_agents`), "Lazy Deletion" (relying on GC to remove agents) is already disabled in practice. Adopting `_StrongAgentSet` formalizes this: agents persist until `agent.remove()` is explicitly called.

My view: This is a positive standard. Relying on Python's Garbage Collector for simulation logic is non-deterministic. Explicit `remove()` calls ensure the simulation state is stable and predictable.

**Integration Plan**
I recommend we apply this specifically to `Model.agents` 

* **Change:** Initialize `self._all_agents = _StrongAgentSet([], self.random)` and `self._agents_by_type` inside `Model.__init__`.
* **Constraint:** Keep `_StrongAgentSet` strictly internal. It should never be instantiated by users directly.


I also believe this is be gonna a critical change as AgentSet directly interacts/will interact with AgentDataSet

mesa/agent.py
```diff
@@ -627,6 +636,84 @@ def groupby(self, by: Callable | str, result_type: str = "agentset") -> GroupBy:
...
+    def _update(self, agents):
+        """Update the internal strong dictionary."""
+        self._agents = self._storage(agents)
+        return self
+
+    def __copy__(self):
+        """Explicitly return a standard (weak) AgentSet when copied."""
+        return AgentSet(self._agents, self.random)
+
+    def do(self, method: str | Callable, *args, **kwargs) -> AgentSet:
```
EwoutH: Could you briefly reflect why some these methods need to be overwritten, others don't, and how we can detect which need to be overwritten if we add a new feature to the AgentSet?

mesa/agent.py
```diff
@@ -180,6 +180,14 @@ class AgentSet[A: Agent](MutableSet[A], Sequence[A]):
 
     """
 
+    def _storage(self, agents: Iterable[A]):
```
EwoutH: The name `_storage` implies it's a property, while this is a function you only call once, right?

EwoutH (commented):
Thanks for working on this, interesting performance benefits!

I will leave the full review to Jan, two minor points.

mesa/agent.py
codebreaker32: Great question.

We strictly override methods that depend on WeakKeyDictionary specific APIs, primarily .keyrefs().

Base AgentSet: Uses .keyrefs() in do, shuffle, and map to efficiently iterate and check liveness of weak references simultaneously.
_StrongAgentSet: Uses a standard dict, which lacks .keyrefs(). We override these to use standard iteration (for agent in self._agents) combined with an explicit liveness check (if agent in self._agents).

Why NOT others: Methods like select, sort, or `__len__` rely on the standard `__iter__` protocol. Since both dict and WeakKeyDictionary support `__iter__`, these methods work "out of the box" via polymorphism and don't need changes.

Detection: If a new method utilizes .keyrefs() or .data without an override, _StrongAgentSet will immediately raise an AttributeError during tests. Ideally, future features should prefer standard iteration (__iter__) to maximize code sharing.

mesa/agent.py
codebreaker32: It acts as a factory method, not a property. It is called during `__init__` and `_update` to generate the backing container.

I used _storage as a shorthand for "storage mechanism," but a verb like `_create_storage` or `_gen_storage` would definitely be more semantically accurate. I'm happy to rename it to `_create_storage` to make the intent clearer.

EwoutH: `_create_storage` is a lot clearer in my opinion!
```suggestion
    def _create_storage(self, agents: Iterable[A]):
```

codebreaker32:
I have updated mesa/model.py to utilize the new _StrongAgentSet implementation primarily to see benchmarks

Results on my local machine:
| BoltzmannWealth | small  | 🟢 -28.3% [-32.5%, -24.0%] | 🟢 -27.2% [-31.7%, -23.1%] |
| BoltzmannWealth | large  | 🔵 +7.2% [-15.0%, +42.1%]  | 🟢 -14.5% [-18.1%, -10.8%] |
| Schelling       | small  | 🟢 -7.0% [-9.5%, -4.6%]    | 🟢 -5.5% [-7.8%, -3.1%]    |
| Schelling       | large  | 🔵 -5.4% [-11.2%, +0.0%]   | 🔵 -6.2% [-16.5%, +3.7%]   |
| WolfSheep       | small  | 🟢 -16.3% [-18.4%, -14.0%] | 🟢 -12.8% [-15.4%, -10.3%] |
| WolfSheep       | large  | 🟢 -24.3% [-29.1%, -19.3%] | 🟢 -27.4% [-32.5%, -22.0%] |
| BoidFlockers    | small  | 🟢 -15.0% [-19.0%, -11.4%] | 🟢 -15.3% [-18.5%, -12.0%] |
| BoidFlockers    | large  | 🟢 -10.8% [-16.2%, -5.6%]  | 🟢 -13.8% [-19.3%, -8.3%]  |

edit:results updated

quaquel:
For Schelling, you should also update agents_by_type to use the dict based agentset. Since this one is already managed by `Model.deregister_agent` it should be a trivial change.

mesa/model.py
```diff
@@ -156,9 +156,9 @@ def __init__(
             A, None
         ] = {}  # the hard references to all agents in the model
```
quaquel, line 161: This can be removed now since self.all_agents already has the hard ref.

codebreaker32:
Some cons that I could think for this approach are:-

1. Copying Costs
AgentSet.copy() is usually fast (shallow copy of weakrefs). _StrongAgentSet.copy() forces the creation of a new AgentSet, which requires generating new weak references for every agent in the set. This makes operations like copy() or non-inplace shuffle() significantly heavier.

2. Polymorphism Issue
_StrongAgentSet shadows AgentSet. If a future contributor adds a new method to AgentSet that relies on internal weakref structures (like .data or .keyrefs()) and forgets to override it in _StrongAgentSet, it will cause runtime crashes (AttributeError).

mesa/agent.py
quaquel: The problem with relying on `__iter__` is that you forgo potential optimizations in case of weakkeydict.

However, I doubt there will be major changes to AgentSet. The API has proven very stable after its introduction and early optimizations.

mesa/model.py
codebreaker32: Removing it will cause cascading failures in tests. I think we can do this in another PR

mesa/agent.py
```diff
@@ -180,6 +180,14 @@ class AgentSet[A: Agent](MutableSet[A], Sequence[A]):
 
     """
 
+    def _create_storage(self, agents: Iterable[A]):
+        """Generate the storage container for the agents.
+
+        Can be overridden by subclasses to change storage behavior (e.g. WeakKeyDictionary).
+        """
+        # We use dict.fromkeys to maintain order and uniqueness efficiently
+        return weakref.WeakKeyDictionary(dict.fromkeys(agents))
```
quaquel, line 183: Why have this as a seperate method? 

I guess it's because of inheritance, but then I suggest moving the random determination into a seperate helper method instead, and simply not calling super, or adding an AbstractBaseClass agent set that contains the overlapping parts, and then have to seperate subclasses of it.

mesa/agent.py
```diff
@@ -205,15 +214,15 @@ def __init__(
...
+            self.random = next(iter(self._agents)).model.random
 
     def __len__(self) -> int:
         """Return the number of agents in the AgentSet."""
         return len(self._agents)
 
     def __iter__(self) -> Iterator[A]:
         """Provide an iterator over the agents in the AgentSet."""
-        return self._agents.keys()
+        return iter(self._agents)
```
quaquel, line 219: why did you change this?

mesa/agent.py
```diff
@@ -627,6 +636,84 @@ def groupby(self, by: Callable | str, result_type: str = "agentset") -> GroupBy:
...
+        return AgentSet(self._agents, self.random)
+
+    def do(self, method: str | Callable, *args, **kwargs) -> AgentSet:
+        """Execute method on agents."""
+        # Snapshot keys to avoid RuntimeError if agents are removed during iteration
+        agents = list(self._agents)
+
+        if isinstance(method, str):
+            for agent in agents:
+                if agent in self._agents:  # Liveness check
```
quaquel: is this if needed because of avoiding changing an iterator while iterator over it?

mesa/agent.py
```diff
@@ -627,6 +636,84 @@ def groupby(self, by: Callable | str, result_type: str = "agentset") -> GroupBy:
...
+        """Map method over agents."""
+        # Map usually implies return values, so we generally don't expect removals.
+        # However, for consistency with 'do', we snapshot.
+        agents = list(self._agents)
+
+        if isinstance(method, str):
+            return [
+                getattr(agent, method)(*args, **kwargs)
+                for agent in agents
+                if agent in self._agents
```
quaquel, line 728: why is this check needed here? We are iterating over our agents to get stuff, so no chance of modifying the iterator.

mesa/agent.py
```diff
@@ -627,6 +636,84 @@ def groupby(self, by: Callable | str, result_type: str = "agentset") -> GroupBy:
...
+    def shuffle(self, inplace: bool = False) -> AgentSet:
+        """Shuffle agents."""
+        agents = list(self._agents)
+        self.random.shuffle(agents)
+
+        if inplace:
+            self._agents = dict.fromkeys(agents)
+            return self
+        else:
+            # Downgrade to standard (weak) AgentSet for external usage
```
quaquel, line 758: it might be worth explicating this behavior in all relevant docstrings

mesa/agent.py
codebreaker32: The reasoning for _create_storage (Template Method pattern) was to reuse the entire `__init__` flow without duplicating code. `AgentSet.__init__ `handles three things:

1. Storage creation.
2. Random resolution (passed vs. derived from agents).
3. Warnings (if empty agents + no random).

If we don't call super(), _StrongAgentSet would need to duplicate the Random and Warning logic(or we would need to extract those into a separate _setup_state helper). I chose the _create_storage hook to keep the changes minimal and ensure both classes share the exact same initialization safety checks. Also, Making AgentSet an Abstract Base Class is tricky because AgentSet is the concrete class users currently instantiate. You would have to rename the current implementation to WeakAgentSet, make AgentSet the ABC, and then add a factory method, which is a much larger (and potentially breaking) change

mesa/agent.py
codebreaker32: `dict.keys()` returns a dictionary view (which is iterable but not an iterator), whereas the `__iter__` method is strictly required to return an iterator. Using `iter(self._agents)` ensures we return a valid key-iterator object that supports `next()`.

mesa/agent.py
codebreaker32: It is slightly more subtle than that.

You are right that we iterate over a snapshot (agents list) to avoid the RuntimeError of changing the dictionary during iteration
Since we iterate over a snapshot of strong references, agents removed during the loop are not garbage collected and remain "alive" in the list (unlike weak references which resolve to None). This check prevents executing methods on these "zombie" agents that have been removed from the model but persist in the iteration snapshot

mesa/agent.py
codebreaker32: While map is often used for retrieval, the method being called can technically have side effects (like removing another agent). This check ensures we don't execute logic on a 'zombie' agent that was removed by a previous agent earlier in the same sequence.

For eg, 
Imagine a tournament scenario where we map a `fight_and_score()` method.
If Agent A fights Agent B and wins, Agent B is removed immediately.

That said, if you want, I can revert it( because this scenario is rare)

mesa/agent.py
codebreaker32: Agreed

mesa/agent.py
quaquel: This is largely an implementation detail, so not key. However, I was not proposing to make the current AgentSet an ABC. Rather, I was suggesting adding a novel AbstractAgentSet(abc.ABC), which is then subclassed in both the existing AgentSet and the new _HardKeyAgentSet.

mesa/agent.py
quaquel: Yes, exactly what I thought. It might be helpful to slightly elaborate on this point in the comment. I know it because of past Mesa stuff, but people new to the repo might not realize how important that is.

mesa/agent.py
quaquel: As a safety, it might make sense, so fine with keeping it in. It's not a very expensive check.

mesa/agent.py
codebreaker32: Ah, I understand now, introducing a common AbstractAgentSet parent for both AgentSet and _StrongAgentSet to share the initialization logic.
That is definitely a cleaner separation of concerns and avoids the "override hook" pattern I used here.
However, since _StrongAgentSet is strictly an internal optimization and you mentioned this isn't a blocker, I am inclined to stick with the current approach to keep the PR focused and the class hierarchy flat. We might implement this pattern in future PRs as long as we don't add more `AgentSet` variants in the future

quaquel:
With 3.4.2 released, we can move towards merging this. Code-wise, it looks good to me. However, can you take a look at the code coverage to see if it is reasonably possible to increase this to above 95%? Also, the benchmark results differ significantly from the original numbers in #3128. Any reason why this might be the case?

codebreaker32:
>Also, the benchmark results differ significantly from the original numbers in https://github.com/mesa/mesa/issues/3128. Any reason why this might be the case?

I am also puzzled by this, even I ran this tests with latest changes rn and results are :-


Screenshot (183)

These tests are very inconsistent on my system :(

quaquel:
I have been running into a similar situation where local results don't bear out on GitHub. Which Python version are you on?

codebreaker32:
>Which python version are you on?

I am on `Python 3.12.3`. The discrepancy might be due to virtualization overhead on CI or something else also might be the case

quaquel:
> I am on Python 3.12.3. The discrepancy might be due to virtualization overhead on CI or something else also might be the case

1. I am also on 3.12.something. So, I'll test your code locally to see my results
2. Which OS are you on?
3. On CI, we use 3.14, which might have something to do with it. But to my knowledge, weakrefs have not changed between 3.12 and 3.14. 
4. There might be a memory issue. Although I fixed one critical leak, there was still another issue that I haven't resolved (but could be circumvented by using remove_agents).
5. It might be an idea to open a seperate issue on the discrepancy between local and CI benchmarks so at least we know to investigate this further.

codebreaker32:
>Which OS are you on?

I am on Ubuntu 22.04

>On CI, we use 3.14, which might have something to do with it. But to my knowledge, weakrefs have not changed between 3.12 and 3.14

Fair enough.

>It might be an idea to open a seperate issue on the discrepancy between local and CI benchmarks so at least we know to investigate this further.

Seems like a good idea. I'll also try to investigate about it online.

quaquel:
No idea what's going on with the benchmarks. I am on Python 3.12.2, on Mac OS 26.1. These are my results if I run main against this PR.

| Model           | Size   | Init time [95% CI]         | Run time [95% CI]       |
|-----------------|--------|----------------------------|-------------------------|
| BoltzmannWealth | small  | 🔵 -4.2% [-6.3%, -2.0%]    | 🔵 -3.0% [-3.4%, -2.6%] |
| BoltzmannWealth | large  | 🟢 -35.7% [-39.7%, -27.7%] | 🔵 +3.4% [-0.3%, +8.3%] |
| Schelling       | small  | 🔵 -2.8% [-3.6%, -1.9%]    | 🔵 -2.6% [-3.8%, -1.4%] |
| Schelling       | large  | 🔵 -3.3% [-5.4%, -0.9%]    | 🔵 -1.0% [-4.7%, +3.7%] |
| WolfSheep       | small  | 🟢 -6.2% [-7.3%, -5.1%]    | 🔵 -3.5% [-4.4%, -2.6%] |
| WolfSheep       | large  | 🟢 -31.5% [-44.5%, -17.6%] | 🔵 -1.7% [-5.0%, +1.4%] |
| BoidFlockers    | small  | 🟢 -5.3% [-7.2%, -3.5%]    | 🔵 -1.6% [-2.3%, -0.8%] |
| BoidFlockers    | large  | 🔵 -3.6% [-5.6%, -1.6%]    | 🔵 -0.1% [-1.1%, +0.9%] |

EwoutH:
On the GitHub side, it’s running on shared, virtual runners. Performance is influenced by other workloads. Not the most stable platform for benchmarks (but free).

> With 3.4.2 released, we can move towards merging this.

Small note: We have maintenance branches (or can create ones), which to we can backport specific fixes. `main` always points to at least a new minor version, sometimes a major, but never a patch. So don’t let that slow you down :)

codebreaker32:
>it’s running on shared, virtual runners. Performance is influenced by other workloads

Agree. But I found some points that maybe we could fix from our side
1. We currently measure from the very first run. This captures the cold start penalty. Standard practice(On netlogo and ASV) is to run the model 3+ times to "warm up" the system before starting the timer.
2.  `gc.enable()` is currently on, meaning the Garbage Collector fires randomly during runs. If GC hits during a specific run, that run looks artificially slow. The standard approach is `gc.disable()` during the timed loop and `gc.collect()` manually between runs.
3. We currently track the minimum(`fastest_init`). While common, this is extremely noisy on shared CI runners. The industry standard (used by NetLogo and ASV) is to calculate the median and Interquartile Range (IQR), which filters out outliers caused by system spikes.
4. We use `timeit.default_timer()`. We should use `time.perf_counter()`, which provides nanosecond resolution and is monotonic (unaffected by system clock updates).

quaquel:
and

5. We don't properly clean up after a run. As indicated in #3179, there is still some memory issue. It would be good to call `model.remove_agents()` at the end of `run_model`.

@codebreaker32, If you have time, a PR would be welcome. Otherwise, I'll take a look hopefully tonight.

codebreaker32:
>@codebreaker32, If you have time, a PR would be welcome

Sure, I am on it

EwoutH:
> 3\. We currently track the minimum(`fastest_init`). While common, this is extremely noisy on shared CI runners. The industry standard (used by NetLogo and ASV) is to calculate the median and Interquartile Range (IQR), which filters out outliers caused by system spikes.

To give some context, there are good arguments for this: While compute execution is often slower, is rarely faster. The distribution tends to be heavily skewed. When originally developing these benchmarks, using the minimum was a conscious decision with more stable results.

That said, improvements to our benchmarking are certainly welcome.

> Polymorphism Issue
_StrongAgentSet shadows AgentSet. If a future contributor adds a new method to AgentSet that relies on internal weakref structures (like .data or .keyrefs()) and forgets to override it in _StrongAgentSet, it will cause runtime crashes (AttributeError).

Did we address this yet?

codebreaker32:
> To give some context, there are good arguments for this: While compute execution is often slower, is rarely faster. The distribution tends to be heavily skewed. When originally developing these benchmarks, using the minimum was a conscious decision with more stable results.

After testing both approaches(results shared on #3203), I strongly agree with you :)



> > Polymorphism Issue
> > _StrongAgentSet shadows AgentSet. If a future contributor adds a new method to AgentSet that relies on internal weakref structures (like .data or .keyrefs()) and forgets to override it in _StrongAgentSet, it will cause runtime crashes (AttributeError).
> 
> Did we address this yet?

I thought it is okay for now after what @quaquel said
>However, I doubt there will be major changes to AgentSet. The API has proven very stable after its introduction and early optimizations.

EwoutH:
Could you check if there's a simple test we could add to check for this?

quaquel:
> Could you check if there's a simple test we could add to check for this?

I think you are overthinking the issue. First, the API has been very stable. Second, if new stuff is added to agentset, we are likely to remember this and test accordingly. Third, I expect to go in and clean this up in a follow-up PR by adding an ABC AgentSet class that both weakkeydict and normal dict inherit from.

codebreaker32:
> Could you check if there's a simple test we could add to check for this?

Also, DW There are enough tests for `_StrongAgentSet` to save us from unsafe method definition or overrides :)

codebreaker32:
Above benchmarks ran using old `global_benchmarks.py` on this PR and updated on main, Now I have merged the main into this branch. You can run again for the updated behavior

quaquel:
I am reviewing the code locally at the moment. I have a few minor comments still:

1. The typing on model for `model.agents` and `model.agents_by_type` is not entirely correct anymore since these return the new _StrongAgentSet instead of AgentSet. I guess because of inheritance it still works. 
2. It might also be good to slightly expand the documentation on `model.agents` and `model.agents_by_type`, warning people to use this with care.
3. It would be good to open an issue indicating that model._agent and model._all_agents can be removed. I know this will be a messy job.

codebreaker32:
You are right since _StrongAgentSet is a subclass, the typing holds up via inheritance, but it is subtle. Also, I acknowledge that the current structure (with _agent, _all_agents, and the inheritance mix) is due for a cleanup.

I see this complete process as a 3-step roadmap to maintain backward compatibility and easy to review:

1. Refactor AgentSet: Introduce an AbstractAgentSet and make the standard AgentSet inherit from it.
2. Formalize Strong Storage: Introduce HardKeyAgentSet (inheriting from AbstractAgentSet) and formally swap it into the Model.
3. Cleanup: Remove redundant instances like model._all_agents once the new hierarchy is stable.

We can open a tracking issue for this roadmap. If you are fine with this approach, I will proceed

quaquel:
Sounds like a plan. Does this mean that this PR will be superseded and so does not need to be merged, or do you want to start from this PR?

codebreaker32:
>Does this mean that this PR will be superseded and so does not need to be merged, or do you want to start from this PR?

I am fine either way!

If you want to get an ABC in immediately, we can merge this PR by tomorrow( I will finalize it by tonight).
Alternatively, if you prefer to have the clean architecture in place first, we can consider this PR superseded. I would then close this and start the new PR series (beginning with `AbstractAgentSet`) from scratch and tracking issue.

Which approach do you prefer?

quaquel:
Let's start from the ABC instead of merging this. The git history will be cleaner and easier to reconstruct in the future, if needed, what changed and why.

PR #3163, opened 2026-01-17
Optimise create_agents by replacing 'ListLike' approach with itertools

codebreaker32:
**Description**
This PR optimizes the `Agent.create_agents` factory method by replacing the custom `ListLike` helper class with standard `itertools` functions. This reduces Python-level overhead during the creation of large batches of agents.

**Motivation**
The previous implementation used a custom `ListLike` class to handle the distinction between scalar arguments (shared by all agents) and list arguments (distributed among agents). This introduced unnecessary attribute lookups (`self.value`) and method calls (`__getitem__`) inside the creation loop.

**Changes**

1. **Removed `ListLike` Class:** The helper class has been removed entirely.
2. **Adopted `itertools.repeat`:** Scalar arguments are now wrapped in `itertools.repeat()`, allowing the loop to handle all arguments uniformly using `zip` without manual indexing or Python-side conditional logic.

**Performance Impact**
Moves the iteration logic to C-level iterators, reducing the per-agent overhead during initialization.

**Backwards Compatibility**
This is a purely internal refactor of the method logic. The public API signature and behavior remain identical.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -151,6 +151,11 @@ class Agent[M: Model]:
         """Return a seeded np.random rng."""
         return self.model.rng
 
+    @property
+    def scenario(self):
+        """Return the scenario associated with the model."""
+        return self.model.scenario
+
 
 class AgentSet[A: Agent](MutableSet[A], Sequence[A]):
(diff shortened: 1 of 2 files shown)


codebreaker32:
Please rerun the benchmarks

Issue #3159, opened 2026-01-17
Add get_parameters() and copy() utilities to Scenario for easier experiment management


navneetkumaryadav207001:
**What's the problem this feature will solve?**
Currently, when working with `Scenario` instances, there is no simple way to:
- retrieve only the scenario parameters while excluding internal metadata (`model`, `_scenario_id`)
- create a modified copy of an existing scenario with updated parameters

This makes experiment sweeps, logging, and reproducibility cumbersome and error-prone.

**Describe the solution you'd like**
Add two small utility methods to the `Scenario` class:

1. `get_parameters()` – returns scenario parameters excluding metadata.
2. `copy(**updates)` – creates a new scenario instance with optional parameter updates:
   - assigns a new `_scenario_id`
   - does not copy the `model` reference
   - shares the same RNG object by reference

These additions would simplify scenario management and improve usability for experiments.

**Additional context**
- A PR implementing this already exists: #3158 
- The `copy()` method is useful for creating slightly modified scenarios for sweeps or testing different parameters without affecting the original instance.
- The RNG sharing behavior is documented to prevent confusion.

quaquel:
resolved via #3168

PR #3166, opened 2026-01-17
Add copy support to Scenario

navneetkumaryadav207001:
# Scenario Copy Utility

## Motivation

When running experiment sweeps, ablations, or parameter tuning, it is common to create multiple variants of a base scenario with small parameter changes.  

Manually reconstructing scenarios is:

- **Error-prone**: Forgetting to update `_scenario_id` or other hidden attributes can cause subtle bugs.  
- **Verbose**: Loops to create multiple scenario variants become messy and repetitive.  

This utility introduces a simple, explicit API for cloning scenarios while preserving reproducibility guarantees.

---

## Implementation

- Added a `copy(**updates)` method to the `Scenario` class.
- **Behavior of copied scenario**:
  - Receives a **new `_scenario_id`**
  - Does **not retain the `model` reference**
  - Shares the same **RNG object by reference** (explicitly documented)
  - Allows **parameter overrides** at creation

---

### Without Copy 
```python
# This won't run
base = Scenario(a=1, b=2, rng=42)
scenarios2 = []
for b in [3, 4, 5]:
    params = base.to_dict() 
    scenarios2.append(Scenario(**params, b= b)) 
```
It has three problems
1. TypeError: experimental.scenarios.scenario.Scenario() got multiple values for keyword argument 'b'
2.  _scenario_id of all scenerios would be same

```python
# Solution
base = Scenario(a=1, b=2, rng=42)
scenarios2 = []
for b in [3, 4, 5]:
    params = base.to_dict()
    params.pop('_scenario_id', None) # fixes 2 
    params.update({'b': b})
    scenarios2.append(Scenario(**params)) 
```
3. Too Verbose

### With Copy

```python
base = Scenario(a=1, b=2, rng=42)
scenarios1 = [base.copy(b=b) for b in [3, 4, 5]]
```

## Additional Context

- Dropped get_parameters as it is redundant on [quaquel](https://github.com/quaquel) suggestion.
- I am not sure why are we copying _scenerio_id from *kwargs in scenerio class initialization
```python
self._scenario_id: int = (
            next(self._ids[self.__class__])
            if "_scenario_id" not in kwargs
            else kwargs.pop("_scenario_id")
        )
```

solves #3159.

### DIFF
--- a/mesa/examples/advanced/alliance_formation/app.py
+++ b/mesa/examples/advanced/alliance_formation/app.py
@@ -8,7 +8,7 @@ from mesa.visualization import SolaraViz
 from mesa.visualization.utils import update_counter
 
 model_params = {
-    "seed": {
+    "rng": {
         "type": "InputText",
         "value": 42,
         "label": "Random Seed",
--- a/mesa/examples/advanced/alliance_formation/model.py
+++ b/mesa/examples/advanced/alliance_formation/model.py
@@ -15,7 +15,7 @@ class MultiLevelAllianceModel(mesa.Model):
     Model for simulating multi-level alliances among agents.
     """
 
-    def __init__(self, n=50, mean=0.5, std_dev=0.1, seed=42):
+    def __init__(self, n=50, mean=0.5, std_dev=0.1, rng=42):
         """
         Initialize the model.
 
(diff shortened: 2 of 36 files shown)


quaquel:
With #3168 merged, I see even less reason for this. Because you can just subclass a scenario with default values and then create new instances from this with only the relevant parameters changed. So again, can you give me a clear practical example of why this would be needed.

I disagree with the naming. A copy should return a copy, not contain support for also changing the content. 

In fact, I am considering making Scenario instances fully frozen. This removes the need for the internal model attribute and `model.running` check, simplifying internal logic and avoiding subtle circular references.

navneetkumaryadav207001:
Thanks for the discussion.

With the subclassing approach from #3168, I agree that there is no strong need for this anymore: when parameters are known ahead of time, subclassing provides a clean solution, and for fully dynamic cases the existing explicit construction patterns remain available, even if more verbose.

Given that, I’m happy to close this PR and focus on follow-up contributions aligned with the new Scenario design.

PR #3167, opened 2026-01-17
Update Epstein Civil Violence model to use the new experimental `Scenario` class

EwoutH:
This PR updates the Epstein Civil Violence model to use the new experimental `Scenario` class for parameter management.

The goal for me personally was to get some feel how our new `Scenario` class works in practice. My first impression is quite positive, it removes so much of the boilerplate of just passing variables. 

It also led me to a few open questions:
1. We now have default parameters. How do we want to handle that? I now chose a `default_scenario`, does that make sense?
2. Is using a `default_scenario`, where whould it be defined? At module level, as a class attribute, a separate `scenario.py`, or somewhere else?
3. Documentation patters: Is `# Parameters accessed via model.scenario: x, y, z` sufficient in docstrings, or should we be more explicit about each parameter's purpose?
4. With the current "just create an instance", type hinting doesn't work:
    Screenshot_778
    Should we support/recommend subclassing for this? That might also make a more elegant way to define default values.

    Or could we make `Scenario` be a `@dataclass` to enable type-hinted subclasses without boilerplate?
    ```python
       class EpsteinScenario(Scenario):
           citizen_density: float = 0.7
           cop_vision: int = 7
           # ... full IDE autocomplete and type checking
    ```

### DIFF
--- a/mesa/examples/advanced/epstein_civil_violence/model.py
+++ b/mesa/examples/advanced/epstein_civil_violence/model.py
@@ -1,3 +1,5 @@
+from typing import Literal
+
 import mesa
 from mesa.discrete_space import OrthogonalMooreGrid, OrthogonalVonNeumannGrid
 from mesa.examples.advanced.epstein_civil_violence.agents import (


EwoutH:
One issue (and why the tests are failing) is that when using a module-level `default_scenario` the same scenario instance gets reused across multiple model instantiations, causing failures when creating a second model.

**The problem:**
```python
# model.py (at module level)
default_scenario = Scenario(...)  # Created once

# Later:
model1 = EpsteinCivilViolence(seed=42)  # scenario.model = model1, model1.running = True
model2 = EpsteinCivilViolence(seed=43)  # Tries to set scenario.model = model2
# ERROR: "Cannot change scenario parameters during model run"
# because scenario.model still points to model1 where running=True
```

**Current workaround:** Create a fresh scenario instance inside `__init__` if none is provided:
```python
def __init__(self, ..., scenario=None):
    if scenario is None:
        scenario = Scenario(rng=seed, citizen_density=0.7, ...)
    super().__init__(seed=seed, scenario=scenario)
```

Subclassing might also help with this. Then we could have:
```python
class EpsteinScenario(Scenario):
    citizen_density: float = 0.7
    cop_vision: int = 7
    # ... etc
```

And then either:
- Users instantiate: `EpsteinCivilViolence(scenario=EpsteinScenario(rng=42))`
- Or model creates internally: `scenario = scenario or EpsteinScenario(rng=seed)`

This gives us type hints, default values, and avoids the shared-instance problem.

(but maybe I'm using all this wrong)

quaquel:
> One issue (and why the tests are failing) is that when using a module-level default_scenario the same scenario instance gets reused across multiple model instantiations, causing failures when creating a second model.

This is useful for stress-testing the design. In part, this interacts with #3166. 
1. The current design is not easy to combine with defaults. The simplest option would be to wrap the defaults in their own dict and then do `Scenario(**defaults.copy().update(rng=43))`. However, the use of defaults is so common that one might as well just have defined a scenario dataclass instead. Basically, this example is a strong counterargument to my dislike of forcing users to subclass.
2. We might just remove the write protection.
3. We might add a factory method, where you can do something like `new.scenario = defaults.from(rng=42)` which would return a copy of defaults with the fields that are passed updated.

EwoutH:
Thanks. I'm curious what you think about https://github.com/mesa/mesa/pull/3168.

quaquel:
I assume this can be updated with #3168 merged. 

While thinking about it some more, I am inclined to make scenarios frozen. That is, they simply cannot be changed once created. We can then remove the `model` attribute and the `model.running` check. This avoids subtle circular reference bugs.

quaquel:
I have made a bunch of updates. The bigger issue is that the UI side of things is not yet updated to use `Scenario`.

EwoutH:
Ah, I also had some changes ready on my branch (part to develop and test #3168), I will check how they compare.

EwoutH:
I stored your work on [`agent_scenario_property_quaquel`](https://github.com/EwoutH/mesa/commits/agent_scenario_property_quaquel/) and pushed my branch. Do you want to compare them?

quaquel:
No, feel free to continue what you were doing. I was mainly curious to see how much #3168 helped.

EwoutH:
In that case I think this PR is ready.

Viz indeed might need to be updated.

One question: I put the `rng` just in the scenario, do we regard that best practice from now on?

quaquel:
> One question: I put the rng just in the scenario, do we regard that best practice from now on?

I don't know yet. It basically means that people are forced to use `Scenario`. But for Mesa 4, I think it is a good idea to do that anyway.

mesa/examples/advanced/epstein_civil_violence/model.py
```diff
@@ -1,10 +1,30 @@
...
+    citizen_vision: int = 7
+    cop_vision: int = 7
+    legitimacy: float = 0.8
+    max_jail_term: int = 1000
+    active_threshold: float = 0.1
+    arrest_prob_constant: float = 2.3
+    movement: bool = True
+    max_iters: int = 1000
+    activation_order: str = "Random"
+    grid_type: str = "Von Neumann"
```
quaquel:
```suggestion
    grid_type: Literal["Von Neumann", "Moore"] = "Von Neumann"
```

mesa/examples/advanced/epstein_civil_violence/model.py
```diff
@@ -1,10 +1,30 @@
...
+    cop_density: float = 0.074
+    citizen_vision: int = 7
+    cop_vision: int = 7
+    legitimacy: float = 0.8
+    max_jail_term: int = 1000
+    active_threshold: float = 0.1
+    arrest_prob_constant: float = 2.3
+    movement: bool = True
+    max_iters: int = 1000
+    activation_order: str = "Random"
```
quaquel:
```suggestion
    activation_order: Literal["Random", "Sequential"] = "Random"
```

tests/examples/test_examples.py
```diff
@@ -70,7 +70,7 @@ def test_epstein():  # noqa: D103
 
     app.page  # noqa: B018
 
-    model = EpsteinCivilViolence(rng=42)
+    model = EpsteinCivilViolence()
```
quaquel, line 123: it now defaults to rng=None?

mesa/examples/advanced/epstein_civil_violence/model.py
```diff
@@ -83,27 +78,22 @@ def __init__(
...
-                    self,
-                    regime_legitimacy=legitimacy,
-                    threshold=active_threshold,
-                    vision=citizen_vision,
-                    arrest_prob_constant=arrest_prob_constant,
-                )
-                citizen.move_to(cell)
+            if klass is not None:
+                agent = klass(self)  # Either Citizen or Cop
+                agent.move_to(cell)
```
quaquel, line 98: this change higlights how `Scenario` can really clean up stuff.

mesa/examples/advanced/epstein_civil_violence/model.py
EwoutH: Thanks, don't often use `Literal`, but it looks like I should!

tests/examples/test_examples.py
EwoutH: No, `rng = 42` is defined in `EpsteinScenario`, and if no `Scenario` is inputted, a default one is created:

```Python
if scenario is None:
    scenario = EpsteinScenario()
```
That should pass the `rng=42` to the right place, right?

mesa/examples/advanced/epstein_civil_violence/model.py
EwoutH: Yes, I was very happy when I realized this could be simplified!

quaquel:
It might be good to test this against #3178 to see if it works with solara. If I have time today, I'll try that and report back here.

EwoutH:
Awesome, thanks!

mesa/examples/advanced/epstein_civil_violence/model.py
```diff
@@ -1,10 +1,30 @@
 import mesa
-from mesa.discrete_space import OrthogonalMooreGrid, OrthogonalVonNeumannGrid  #
+from mesa.discrete_space import OrthogonalMooreGrid, OrthogonalVonNeumannGrid
```
quaquel, line 4:
```suggestion
from mesa.discrete_space import OrthogonalMooreGrid, OrthogonalVonNeumannGrid
from typing import Literal
```

mesa/examples/advanced/epstein_civil_violence/model.py
```diff
@@ -1,10 +1,30 @@
...
 from mesa.examples.advanced.epstein_civil_violence.agents import (
     Citizen,
     CitizenState,
     Cop,
 )
+from mesa.experimental.scenarios import Scenario
+
+
+# Define a typed scenario subclass with defaults
+class EpsteinScenario(Scenario):
```
quaquel, line 14: this should be used correctly in app.py

EwoutH:
> It might be good to test this against #3178 to see if it works with solara. If I have time today, I'll try that and report back here.

Considering this and the interaction with #3178, what would your preference be that I did with this PR?

quaquel:
I'll try to take a look and see what's going on with #3178 or, otherwise, open a new PR myself to fix it.

quaquel:
I can also confirm that this example now runs locally.

PR #3174, opened 2026-01-18
Improved typing support for Scenario

quaquel:
Small follow up for #3168, which ensures that `model.scenario` is properly typed. Also see #3167.


EwoutH:
This now just removes a comment?

quaquel:
don't know what happpend but the changes this made are already present on main

Issue #3176, opened 2026-01-18
Support Scenarios in Visualization


EwoutH:
### Problem
Mesa now has first-class support for Scenarios (#3103, #3168, and updated example #3167), but `SolaraViz` doesn't support them. When users try to visualize a model that uses a Scenario, they have no clear way to specify which parameters belong to the Scenario vs the Model.

**Current situation:**
```python
class MyScenario(Scenario):
    density: float = 0.7
    vision: int = 7

class MyModel(Model):
    def __init__(self, height=40, width=40, scenario: MyScenario | None = None):
        super().__init__(scenario=scenario)
```

Users need to create sliders for `density` and `vision`, but these aren't accepted by `MyModel.__init__`, since they belong to the Scenario. When the user clicks "Reset", `SolaraViz` tries to instantiate `MyModel(density=..., vision=...)`, which fails.

#### Option 1: Add `scenario_params` argument
```python
page = SolaraViz(
    model,
    renderer,
    model_params={"height": 40, "width": 40},
    scenario_params={
        "density": Slider(...),
        "vision": Slider(...),
    }
)
```

**Mesa changes needed:**
- Add `scenario_params` parameter to `SolaraViz()`
- Update `ModelController.do_reset()` to reconstruct Scenario from parameters
- Render scenario params in separate UI section

**Pros:** Clear, explicit, easy to understand  
**Cons:** More verbose, requires manual specification

#### Option 2: Auto-detect from type hints
```python
# SolaraViz introspects model.scenario and auto-generates UI
page = SolaraViz(model, renderer)

# Optional customization
page = SolaraViz(model, renderer, scenario_params={"density": Slider(...)})
```

**Mesa changes needed:**
- Introspect `model.scenario` to get Scenario class
- Use `get_type_hints()` and `_scenario_defaults` to auto-generate controls
- Allow overrides via `scenario_params`

**Pros:** Minimal user code, leverages existing type hints  
**Cons:** Less explicit, auto-generated ranges might not fit

### Option 3: Something else?
Other nice ideas for a clean API?

### Discussion Questions
1. How would the user API looks?
2. Should scenario parameters appear in a separate UI section?
3. How do we handle the `rng` parameter that scenarios own?

### Implementation areas
Regardless of option chosen:
- `solara_viz.py`: Update `SolaraViz`, `ModelCreator`, `ModelController`
- Documentation: Add examples and migration guide

ShreyasN707:
[...]

PR #3178, opened 2026-01-19
Support Scenarios in SolaraViz visualization

falloficaruss:
### Summary
This PR adds first-class support for `Scenarios` in `SolaraViz`. It allows users to visualize and interact with models that utilize the new Mesa `Scenario` system, ensuring that parameters are correctly routed to either the Model or the Scenario during instantiation and resets.

### Motive
With the introduction of `Scenarios` in Mesa (#3103, #3168), there was a disconnect in the visualization layer. 
`SolaraViz` previously tried to pass all user-defined sliders and parameters directly into `Model.__init__`. This caused a `TypeError` when a model used a Scenario, because parameters intended for the Scenario were rejected by the Model's initialization function. This PR solves the routing problem and provides a clear UI separation between Model and Scenario parameters.

### Implementation
The implementation follows the "Option 1" (Explicit API) and "Option 2" (Auto-detection) suggestions from the issue:

- `SolaraViz` Parameter Routing: Enhanced `SolaraViz` to automatically introspect the model's `scenario` attribute and intelligently split parameters between Model and Scenario based on their respective signatures.
- Scenario Reconstruction on Reset: Updated `ModelController` and `SimulatorController`. When the "Reset" button is clicked, it now captures the current values of scenario sliders, reconstructs the specific `Scenario` subclass instance, and passes it into the `Model` constructor via the `scenario` keyword argument.
- UI Enhancements: Updated `ModelCreator` to render Scenario-bound sliders in a dedicated "Scenario Parameters" section beneath the main Model parameters, providing visual clarity on parameter ownership.
- Tests: Added comprehensive test coverage in both `tests/visualization/test_solara_viz.py` (testing core functionality like parameter splitting logic, `ModelCreator` with scenarios, and `SolaraViz` integration) and `tests/visualization/test_solara_viz_scenarios.py` (testing scenario-specific edge cases and type hints).

### Usage Examples
Auto Detection:
If your sliders match field names in your `Scenario` class, `SolaraViz` handles the routing automatically:

```python
class MyScenario(Scenario):
    density: float = 0.7

class MyModel(Model):
    def __init__(self, height=40, width=40, scenario=None):

# 'density' is auto-detected as a Scenario parameter
page = SolaraViz(
    MyModel(),
    model_params={
        "height": 40,
        "density": Slider("Density", 0.7, 0, 1, 0.1)
    }
)
```

### Additional Notes
This change is fully backward compatible. Models not using Scenarios will function exactly as before.
Fixes: #3176

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -10,15 +10,15 @@ changelog:
     - title: ⏳ Deprecations
       labels:
         - deprecation
-    - title: 🧪 Experimental features
-      labels:
-        - experimental
     - title: 🎉 New features added
       labels:
         - feature
     - title: 🛠 Enhancements made
       labels:
         - enhancement
+    - title: 🧪 Experimental
+      labels:
+        - experimental
     - title: 🐛 Bugs fixed
       labels:
         - bug
--- a/mesa/discrete_space/cell.py
+++ b/mesa/discrete_space/cell.py
@@ -18,6 +18,8 @@ from functools import cache
 from random import Random
 from typing import TYPE_CHECKING
 
+import numpy as np
+
 from mesa.discrete_space.cell_agent import CellAgent
 from mesa.discrete_space.cell_collection import CellCollection
 
@@ -31,7 +33,8 @@ class Cell:
     """The cell represents a position in a discrete space.
 
     Attributes:
-        coordinate (Tuple[int, int]) : the position of the cell in the discrete space
+        coordinate (Coordinate) : the logical position(or index) of the cell in the discrete space
+        position (np.ndarray | None): the physical position of the cell in the discrete space
         agents (List[Agent]): the agents occupying the cell
         capacity (int): the maximum number of agents that can simultaneously occupy the cell
         random (Random): the random number generator
@@ -41,9 +44,10 @@ class Cell:
     __slots__ = [
         "_agents",
         "_empty",
+        "_position",  # physical position
         "capacity",
         "connections",
-        "coordinate",
+        "coordinate",  # Logical index
         "properties",
         "random",
     ]
@@ -56,9 +60,26 @@ class Cell:
     def empty(self, value: bool) -> None:
         self._empty = value
 
+    @property
+    def position(self) -> np.ndarray:
+        """Get the physical position of the cell.
+
+        Returns:
+            np.ndarray: Physical position of the cell
+        """
+        if self._position is not None:
+            return self._position
+        # Default for implicit grids
+        return np.asarray(self.coordinate, dtype=float)
+
+    @position.setter
+    def position(self, value: np.ndarray | None) -> None:
+        self._position = value
+
     def __init__(
         self,
         coordinate: Coordinate,
+        position: np.ndarray | None = None,
         capacity: int | None = None,
         random: Random | None = None,
     ) -> None:
@@ -66,12 +87,14 @@ class Cell:
 
         Args:
             coordinate: coordinates of the cell
+            position: physical coordinates of the cell
             capacity (int) : the capacity of the cell. If None, the capacity is infinite
             random (Random) : the random number generator to use
 
         """
         super().__init__()
-        self.coordinate = coordinate
+        self.coordinate = coordinate  # Logical index
+        self._position = position  # Physical position
         self.connections: dict[Coordinate, Cell] = {}
         self._agents: list[
             CellAgent
--- a/mesa/discrete_space/discrete_space.py
+++ b/mesa/discrete_space/discrete_space.py
@@ -20,6 +20,8 @@ from functools import cached_property
 from random import Random
 from typing import TypeVar
 
+import numpy as np
+
 from mesa.agent import AgentSet
 from mesa.discrete_space.cell import Cell
 from mesa.discrete_space.cell_collection import CellCollection
@@ -84,6 +86,13 @@ class DiscreteSpace[T: Cell]:
 
     def _connect_cells(self): ...
     def _connect_single_cell(self, cell: T): ...
+
+    def find_nearest_cell(self, position: np.ndarray) -> T:
+        """Find the cell at or nearest to the given position."""
+        raise NotImplementedError(
+            f"{type(self).__name__} does not implement find_nearest_cell(). "
+            "This space may be purely topological."
+        )
 
     def add_cell(self, cell: T):
         """Add a cell to the space.
--- a/mesa/discrete_space/grid.py
+++ b/mesa/discrete_space/grid.py
@@ -14,12 +14,14 @@ or Hex for more uniform distances.
 from __future__ import annotations
 
 import copyreg
+import math
 from collections.abc import Sequence
 from itertools import product
 from random import Random
 from typing import Any, TypeVar
 
 import numpy as np
+from scipy.spatial import KDTree
 
 from mesa.discrete_space import Cell, DiscreteSpace
 from mesa.discrete_space.property_layer import (
@@ -123,6 +125,34 @@ class Grid(DiscreteSpace[T], HasPropertyLayers):
         self._celllist = list(self._cells.values())
         self._connect_cells()
         self.create_property_layer("empty", default_value=True, dtype=bool)
+
+    def find_nearest_cell(self, position: np.ndarray) -> T:
+        """Find the cell containing the given position.
+
+        Args:
+            position: Physical position [x, y]
+
+        Returns:
+            Cell: The cell containing the position
+
+        Raises:
+            KeyError: If position is outside grid bounds and not a torus
+        """
+        # Floor to get cell coordinate
+        coord = tuple(np.floor(position).astype(int))
+
+        # Handle torus wrapping
+        if self.torus:
+            coord = tuple(c % d for c, d in zip(coord, self.dimensions))
+
+        # Check bounds for non-torus grids
+        elif not all(0 <= c < d for c, d in zip(coord, self.dimensions)):
+            raise ValueError(
+                f"Position {position} is outside grid bounds. "
+                f"Dimensions: {self.dimensions}"
+            )
+
+        return self._cells[coord]
 
     def _connect_cells(self) -> None:
         if self._ndims == 2:
@@ -286,6 +316,68 @@ class HexGrid(Grid[T]):
     Raises:
         ValueError: If torus=True and either width or height is odd.
     """
+
+    def __init__(
+        self,
+        dimensions: Sequence[int],
+        torus: bool = False,
+        capacity: float | None = None,
+        random: Random | None = None,
+        cell_klass: type[T] = Cell,
+    ) -> None:
+        """Initialize the hex grid.
+
+        Args:
+            dimensions: the dimensions of the space
+            torus: whether the space wraps
+            capacity: capacity of the grid cell
+            random: a random number generator
+            cell_klass: the base class to use for the cells
+        """
+        super().__init__(
+            dimensions=dimensions,
+            torus=torus,
+            capacity=capacity,
+            random=random,
+            cell_klass=cell_klass,
+        )
+        self._init_hex_geometry()
+
+    def _init_hex_geometry(self) -> None:
+        """Calculate physical positions for all cells and build KD-Tree.
+
+        Refer https://www.redblobgames.com/grids/hexagons/#hex-to-pixel for more detail
+        """
+        positions = []
+        self._kdtree_coords = []
+
+        size = 1.0
+        for coord, cell in self._cells.items():
+            col, row = coord
+            x = size * math.sqrt(3) * (col + 0.5 * (row % 2))
+            y = size * 1.5 * row
+            position = np.array([x, y])
+
+            cell.position = position
+            positions.append(position)
+            self._kdtree_coords.append(coord)
+
+        self._kdtree = KDTree(np.array(positions))
+
+    def find_nearest_cell(self, position: np.ndarray) -> T:
+        """Find the hex cell at the given position."""
+        position = np.asarray(position)
+
+        if self.torus:
+            width_pixels = self.dimensions[0] * math.sqrt(3)
+            height_pixels = self.dimensions[1] * 1.5
+            position = np.array(
+                [position[0] % width_pixels, position[1] % height_pixels]
+            )
+
+        _, index = self._kdtree.query(position)
+        coord = self._kdtree_coords[index]
+        return self._cells[coord]
 
     def _connect_cells_2d(self) -> None:
         # fmt: off
--- a/mesa/discrete_space/network.py
+++ b/mesa/discrete_space/network.py
@@ -12,8 +12,13 @@ Useful for modeling systems like social networks, transportation systems,
 or any environment where connectivity matters more than physical location.
 """
 
+from collections.abc import Callable, Mapping
 from random import Random
 from typing import Any
+
+import networkx as nx
+import numpy as np
+from scipy.spatial import KDTree
 
 from mesa.discrete_space.cell import Cell
 from mesa.discrete_space.discrete_space import DiscreteSpace
@@ -28,6 +33,7 @@ class Network(DiscreteSpace[Cell]):
         capacity: int | None = None,
         random: Random | None = None,
         cell_klass: type[Cell] = Cell,
+        layout: Mapping | Callable | None = nx.spring_layout,
     ) -> None:
         """A Networked grid.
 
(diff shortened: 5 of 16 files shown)


EwoutH:
Thanks for the PR. However, I feel we didn’t completere finish the discussion in https://github.com/mesa/mesa/issues/3176 on how we would integrate scenario variables exactly. Let’s get full consensus on the design first.

API design is important, since once we commit to it, we have to support it for a long time.

See also https://github.com/mesa/mesa/blob/main/CONTRIBUTING.md#mesa-development-process

quaquel:
I suggest simplifying this by only solving the current scenario problem. That is, keep the reflection stuff to figure out what is in the scenario object and what is on the model init, and use that for the ModelController. Leave the slider stuff untouched.

falloficaruss:
@quaquel Any thoughts here?

quaquel:
Please include test code in the existing test files. @Sahil-Chhoker, I would appreciate your input here as well. 

Basically, for now, I just want to ensure that solara-viz is compatible with scenarios.

mesa/visualization/solara_viz.py
```diff
@@ -535,7 +537,35 @@ def do_reset():
             10,
             f"creating new {model.value.__class__} instance with {model_parameters.value}",
         )
-        model.value = model.value = model.value.__class__(**model_parameters.value)
+        kwargs = {}
+        scenario_kwargs = {}
+
+        # Check if the model has a scenario and split parameters accordingly
+        if hasattr(model.value, "scenario") and model.value.scenario:
```
quaquel:
```suggestion
        if getattr(model.value, "scenario", None):
```

mesa/visualization/solara_viz.py
```diff
@@ -535,7 +537,35 @@ def do_reset():
...
+            scenario_class = model.value.scenario.__class__
```
quaquel:
```suggestion
            scenario_class = type(model.value.scenario)
```

quaquel:
I tested it with #3167. It does not yet work correctly because rng is not handled properly. 

My suggestion is that you port Boltzmann to use a new scenario, modeled on #3167, and use that to have a proper integration test.

mesa/examples/advanced/epstein_civil_violence/agents.py
quaquel, line 1: Why did you add in Epstein instead of e.g.,. Boltzmann? this now classes with #3176.

falloficaruss: I applied the scenario changes to Epstein as a quick integration target.
I’ll move the scenario‑based integration work to Boltzmann (modeled on #3167) and adjust the tests accordingly.

quaquel:
It won't run for me locally: 
```
raise ValueError(
ValueError: Mesa's visualization requires the use of keyword arguments to ensure the parameters are passed to Solara correctly. Please ensure all model parameters are of form param=value
```

mesa/examples/basic/boltzmann_wealth_model/model.py
```diff
@@ -27,19 +36,42 @@ class BoltzmannWealth(Model):
...
-            rng (int, optional): Random rng. Defaults to None.
+            scenario: BoltzmannScenario object containing model parameters.
         """
-        super().__init__(rng=rng)
+        if args:
+            if len(args) > 3:
+                raise ValueError("Expected at most 3 positional args: n, width, height")
+            arg_keys = ("n", "width", "height")
+            for key, value in zip(arg_keys, args):
+                kwargs.setdefault(key, value)
```
quaquel: why, please just keep the example simple and as close as possible to the original version.e

mesa/examples/basic/boltzmann_wealth_model/model.py
```diff
@@ -27,19 +36,42 @@ class BoltzmannWealth(Model):
...
+            }
+            scenario = BoltzmannScenario(rng=rng, **scenario_kwargs)
+        else:
+            scenario_kwargs = {
+                key: kwargs.pop(key)
+                for key in ("n", "width", "height")
+                if key in kwargs
+            }
+            for key, value in scenario_kwargs.items():
+                setattr(scenario, key, value)
```
quaquel: please just use 

```python
self.my_param = scenario.param
```

Also, remove the old kwargs stuff. in the if clause. It should work only with the scenario object, and if that is None, you instantiate a Scenario object with its default values.

mesa/visualization/solara_viz.py
```diff
@@ -533,9 +535,45 @@ def do_reset():
...
+                if k in scenario_defaults and k not in model_init_params:
+                    scenario_kwargs[k] = v
+                else:
+                    kwargs[k] = v
+
+            # Handle seeding continuity and ensure scenario instantiation
+            if "rng" in model_parameters.value:
+                scenario_kwargs["rng"] = model_parameters.value["rng"]
+            elif "seed" in model_parameters.value:
+                scenario_kwargs["rng"] = model_parameters.value["seed"]
```
quaquel: in my view, scenario contains rng not the other way around.

also seed is deprecated anyway.

falloficaruss:
> It won't run for me locally:
> 
> ```
> raise ValueError(
> ValueError: Mesa's visualization requires the use of keyword arguments to ensure the parameters are passed to Solara correctly. Please ensure all model parameters are of form param=value
> ```

Can u check it again?

mesa/examples/basic/boltzmann_wealth_model/model.py
```diff
@@ -27,19 +36,22 @@ class BoltzmannWealth(Model):
         datacollector (DataCollector): Collects and stores model data
     """
 
-    def __init__(self, n=100, width=10, height=10, rng=None):
+    def __init__(self, scenario=None, rng=None):
```
quaquel:
```suggestion
    def __init__(self, scenario=None):
```

rng is contained in scenario so not needed anymore in the signature

mesa/examples/basic/boltzmann_wealth_model/model.py
```diff
@@ -27,19 +36,22 @@ class BoltzmannWealth(Model):
...
             rng (int, optional): Random rng. Defaults to None.
+            scenario: BoltzmannScenario object containing model parameters.
         """
-        super().__init__(rng=rng)
+        if scenario is None:
+            scenario = BoltzmannScenario(rng=rng)
 
-        self.num_agents = n
-        self.grid = OrthogonalMooreGrid((width, height), random=self.random)
+        super().__init__(scenario=scenario, rng=rng)
```
quaquel:
```suggestion
        super().__init__(scenario=scenario)
```

mesa/examples/basic/boltzmann_wealth_model/model.py
```diff
@@ -27,19 +36,22 @@ class BoltzmannWealth(Model):
...
         Args:
-            n (int, optional): Number of agents. Defaults to 100.
-            width (int, optional): Grid width. Defaults to 10.
-            height (int, optional): Grid height. Defaults to 10.
             rng (int, optional): Random rng. Defaults to None.
+            scenario: BoltzmannScenario object containing model parameters.
         """
-        super().__init__(rng=rng)
+        if scenario is None:
+            scenario = BoltzmannScenario(rng=rng)
```
quaquel:
```suggestion
            scenario = BoltzmannScenario()
```

mesa/examples/basic/boltzmann_wealth_model/model.py
```diff
@@ -11,6 +11,15 @@
...
 from mesa.examples.basic.boltzmann_wealth_model.agents import MoneyAgent
+from mesa.experimental.scenarios import Scenario
+
+
+class BoltzmannScenario(Scenario):
+    """Scenario parameters for the Boltzmann Wealth model."""
+
+    n: int = 100
+    width: int = 10
+    height: int = 10
```
quaquel, line 22:
```suggestion
    height: int = 10
    rng: SeedLike|None = None
```

mesa/examples/basic/boltzmann_wealth_model/st_app.py
```diff
@@ -19,7 +19,9 @@
 )
 height = st.slider("Select Grid Height", min_value=10, max_value=100, step=10, value=15)
 width = st.slider("Select Grid Width", min_value=10, max_value=100, step=10, value=20)
-model = BoltzmannWealth(num_agents, height, width)
+model = BoltzmannWealth(
+    scenario=BoltzmannScenario(n=num_agents, width=width, height=height)
```
quaquel:
```suggestion
    scenario=BoltzmannScenario(n=num_agents, width=width, height=height, rng=42)
```

falloficaruss:
> Member

Applied the changes but I think this PR needs further fixing.

falloficaruss:
- Aligned the Boltzmann example with the scenario-first design. `BoltzmannWealth` now only takes scenario (no rng in the model signature). If scenario is `None`, it instantiates `BoltzmannScenario()` with defaults and uses parameters via `scenario.*`

- Updated all call sites/tests to pass `rng` through `BoltzmannScenario(rng=...)` instead of `BoltzmannWealth(rng=...)`. This keeps RNG ownership in the scenario as requested.

- Also removed `SolaraViz’s` override that injected `rng/seed` into scenarios from model `params` to avoid reversing the ownership.

This keeps the example close to the original while making the scenario the single source of truth for parameters (including RNG).

EwoutH:
Thanks for working on this.

I think the goal of scenarios (in the long term) is also to decomplicate handling model parameters, like in the visualisation. Right now it seems it mainly complicates it, at least on the source code side. Any reflections on how this will develop in the future?

quaquel:
In my opinion, the coupling between the UI and models is too tight and that is causing a lot of the problems.

falloficaruss:
I agree the current integration adds complexity, especially in the viz layer. The intent is that this is a transitional cost.

As dicussed in https://github.com/mesa/mesa/discussions/2972 in the longer term the goal is to make the UI only deal with a single surface: `scenario`. When models consistently expose parameters via scenarios, the visualization doesn’t need to inspect model signatures or split params manually. It just binds controls to scenario fields and recreates the model with a scenario instance. That should reduce coupling between UI and model internals.

Right now we’re in a hybrid state (models with and without scenarios), so the UI has to support both and it looks more complex than the steady state. Once scenarios are the default pattern, we can remove the fallback logic and the visualization code should get simpler.

mesa/examples/basic/boltzmann_wealth_model/app.py
```diff
@@ -34,6 +32,11 @@ def agent_portrayal(agent):
     },
     "width": 10,
     "height": 10,
+    "rng": {
+        "type": "InputText",
+        "value": 42,
+        "label": "Random Seed",
+    },
```
quaquel: can you put this back at the top. It is is misleading in the diff.

mesa/visualization/solara_viz.py
```diff
@@ -205,7 +204,9 @@ def set_reactive_use_threads(value):
                 )
         with solara.Card("Model Parameters"):
             ModelCreator(
-                model, model_params, model_parameters=reactive_model_parameters
+                model,
+                model_params,
+                model_parameters=reactive_model_parameters,
```
quaquel: these are also not really changes. Please check the diff and revert anything that is not a real change (unless this is due to ruff or so, but I doubt it)

mesa/visualization/solara_viz.py
```diff
@@ -469,6 +470,7 @@ def ModelController(
     if model_parameters is None:
         model_parameters = {}
     model_parameters = solara.use_reactive(model_parameters)
+
```
quaquel: same here

mesa/visualization/solara_viz.py
```diff
@@ -533,9 +535,41 @@ def do_reset():
...
+        # Check if the model has a scenario and split parameters accordingly
+        if getattr(model.value, "scenario", None):
+            scenario_class = type(model.value.scenario)
+            # We assume scenario parameters are those in scenario class but NOT in model init
+            # Check for _scenario_defaults first (Mesa's Scenario mechanism) or fallback to signature
+            scenario_defaults = getattr(model.value.scenario, "_scenario_defaults", {})
+            if not scenario_defaults:
+                # Fallback to signature inspection if _scenario_defaults is missing
+                scenario_defaults = inspect.signature(scenario_class).parameters
+
```
quaquel: I am not sure we need this. Just assume or check if the scenario is a subclass of mesa's Scenario and raise a ValueError otherwise.

mesa/visualization/solara_viz.py
```diff
@@ -533,9 +535,41 @@ def do_reset():
...
+            model_init_params = inspect.signature(type(model.value).__init__).parameters
+
+            for k, v in model_parameters.value.items():
+                if k in scenario_defaults and k not in model_init_params:
+                    scenario_kwargs[k] = v
```
quaquel: why this check?

mesa/visualization/solara_viz.py
```diff
@@ -660,9 +696,33 @@ def do_reset():
         simulator.reset()
         visualization_pause_event.clear()
         pause_step_event.clear()
-        model.value = model.value = model.value.__class__(
-            simulator=simulator, **model_parameters.value
-        )
+
```
quaquel, line 720: how much of the code below is redundant with the code above and if so can it be handled in a separate function to avoid code duplication.

mesa/visualization/solara_viz.py
```diff
@@ -766,7 +826,7 @@ def ModelCreator(
     model_parameters = solara.use_reactive(model_parameters)
 
     solara.use_effect(
-        lambda: _check_model_params(model.value.__class__.__init__, user_params),
+        lambda: _check_model_params(type(model.value).__init__, user_params),
```
quaquel: please remove this. You can put in a separate PR for just the `__class__` to `type` changes and we can quickly merge that because it is a good idea.

mesa/visualization/solara_viz.py
falloficaruss: This check is to avoid ambiguous routing in hybrid models.
We only send a key to `scenario_kwargs` when it is scenario-defined and not explicitly declared in model `__init__`.
If a name exists in both places, we currently treat it as a model kwarg to preserve backward compatibility.
If we want stricter semantics, I can switch to raising on overlap (or always prefer scenario) and remove this compatibility behavior.

quaquel:
I still get the error shown below when I try to run this locally. 

```
  File "/opt/anaconda3/lib/python3.13/site-packages/reacton/core.py", line 1900, in _reconsolidate
    effect()
    ~~~~~~^^
  File "/opt/anaconda3/lib/python3.13/site-packages/reacton/core.py", line 1131, in __call__
    self._cleanup = self.callable()
                    ~~~~~~~~~~~~~^^
  File "/Users/user_name/Documents/GitHub/mesa/mesa/visualization/solara_viz.py", line 831, in <lambda>
    lambda: _check_model_params(type(model.value).__init__, user_params),
            ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/user_name/Documents/GitHub/mesa/mesa/visualization/solara_viz.py", line 887, in _check_model_params
    raise ValueError(f"Invalid model parameter: {name}")
ValueError: Invalid model parameter: rng
```

falloficaruss:
> I still get the error shown below when I try to run this locally.
> 
> ```
> Traceback (most recent call last):
>   File "/opt/anaconda3/lib/python3.13/site-packages/reacton/core.py", line 1900, in _reconsolidate
>     effect()
>     ~~~~~~^^
>   File "/opt/anaconda3/lib/python3.13/site-packages/reacton/core.py", line 1131, in __call__
>     self._cleanup = self.callable()
>                     ~~~~~~~~~~~~~^^
>   File "/Users/user_name/Documents/GitHub/mesa/mesa/visualization/solara_viz.py", line 831, in <lambda>
>     lambda: _check_model_params(type(model.value).__init__, user_params),
>             ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/user_name/Documents/GitHub/mesa/mesa/visualization/solara_viz.py", line 887, in _check_model_params
>     raise ValueError(f"Invalid model parameter: {name}")
> ValueError: Invalid model parameter: rng
> ```

Will you take a look again?

quaquel:
I now get `asyncio.exceptions.CancelledError`. Does it run locally for you, and, if so, can you elaborate on how you are testing it?

falloficaruss:
> I now get `asyncio.exceptions.CancelledError`. Does it run locally for you, and, if so, can you elaborate on how you are testing it?

I am currently testing the visualization using `pytest` combined with `solara.render`.
In Solara and Reacton, `asyncio.CancelledError` should actually be normal and expected behavior. When a user clicks "Reset" or "Pause," Solara cancels the previous simulation task and (if resetting) starts a new one. I'm not totally sure about this and I would like to get some help on this.
For now I will explicitly catch and handle these silently.

quaquel:
I had a closer look at this and I know what is going on. First, revert your last commit. Errors should never be past over in silence. Also, it is not the root cause of the problem

My problem now happens when I hit reset. This trigger indirectly `solara_viz._get_scenario_defaults`. Here , we try to reconstruct the Scenario instance partly based on `model_parameters`.  This dict is populated based on the UI sliders. And so includes the rng slider. However, `rng` is not included in `_scenario_defaults`, so it is not included when creating the scenario instance and instead is kept as part of model_parameters. 

The solution is to change how we recreate the scenario instance in `_build_model_init_kwargs`. 

for example

```python
        for key, value in model_parameters.items():
            if key == 'rng':
                scenario_kwargs['key'] = value
            elif key in scenario_defaults and key not in model_init_params:
                scenario_kwargs[key] = value
            else:
                kwargs[key] = value
```

fixes the problem. This is clearly hacky, but it shows that my diagnosis is correct. A slightly more elabore solution is to change `_get_scenario_defaults`:

```python
def _get_scenario_defaults(scenario: Scenario) -> dict[str, Any]:
    """Return defaults for a scenario instance, validating scenario type."""
    scenario_class = type(scenario)
    if not issubclass(scenario_class, Scenario):
        raise ValueError(
            "Expected model.scenario to be a subclass instance of mesa.experimental.scenarios.Scenario."
        )
    defaults = getattr(scenario, "_scenario_defaults", {})
    defaults['rng'] = None

    return defaults
```

tests/visualization/test_solara_viz_scenarios.py
quaquel, line 1: We are moving our tests over to pytest, so please rewrite this using pytest instead of unittest. Should be a five minute job or even faster with an llm.

tests/examples/test_examples_viz.py
```diff
@@ -16,13 +17,73 @@
...
 )
+from mesa.examples.basic.boltzmann_wealth_model.model import BoltzmannScenario
 from mesa.visualization.components import AgentPortrayalStyle
 from mesa.visualization.components.matplotlib_components import (
     PlotMatplotlib,
     SpaceMatplotlib,
 )
 
 
+def _screenshot_first_image(page, attempts=3, wait_ms=100):
```
quaquel: What is going on with these new functions? How is this relevant to the scenario issue? If it is not relevant, please consider submitting it as a separate pr

falloficaruss: sorry these were not meant to be pushed (Might have messed up while checking out with uncommitted changes)

quaquel:
It works now on my end. I will give the code another quick check, but it seems we are almost there.

tests/visualization/test_solara_viz_scenarios.py
```diff
@@ -0,0 +1,148 @@
...
+
+    assert scenario.density == 0.8
+    assert scenario.vision == 10
+    assert scenario.speed == 1.0  # Still default.
+
+
+@pytest.mark.skip(
+    reason="Requires interactive Solara state handling; covered indirectly."
+)
+def test_reset_with_scenario():
```
quaquel, line 128: I understand that you skip this, but you might want to test `_build_model_init_kwargs`. It would have helped in finding the bug.

falloficaruss: Okay I'll replace the skipped test with a concrete `_build_model_init_kwargs` test

falloficaruss:
> It works now on my end. I will give the code another quick check, but it seems we are almost there.

Thank you for the help!

EwoutH:
Great to see the progress!

I would like to also have @Sahil-Chhoker’s take on it.

quaquel (approved):

quaquel:
I have tested this locally and can confirm it works. I am merging this PR, and we'll see if #3167 also works. @Sahil-Chhoker, I would still appreciate your thoughts even post-merge.

EwoutH:
Thanks for this effort, glad to see it merged!

Issue #3179, opened 2026-01-19
Memory leak in models


Juillermo:
Not sure if this is a bug or I'm missing something, but I couldn't find any solution in the documentation or after attempting some debugging. Not sure if this has anything to do and could be addressed with the _StrongAgentSet_ change.

**Describe the bug**
When I run a few instance of a model (encapsulated within a function), objects don't get cleaned and they accumulate in RAM.

**Expected behavior**
If a model is run within a function, I would expect no trace in memory of the model once the function has been exited.

**To Reproduce**
Basic model from https://mesa.readthedocs.io/latest/overview.html.
```python
import mesa

class MyAgent(mesa.Agent):
    def __init__(self, model, age):
        super().__init__(model)
        self.age = age

    def step(self):
        self.age += 1
        # print(f"Agent {self.unique_id} now is {self.age} years old")
        # Whatever else the agent does when activated

class MyModel(mesa.Model):
    def __init__(self, n_agents):
        super().__init__()
        self.grid = mesa.space.MultiGrid(10, 10, torus=True)
        for _ in range(n_agents):
            initial_age = self.random.randint(0, 80)
            a = MyAgent(self, initial_age)
            coords = (self.random.randrange(0, 10), self.random.randrange(0, 10))
            self.grid.place_agent(a, coords)

    def step(self):
        self.agents.shuffle_do("step")


def main():
    n_agents=1000
    model = MyModel(n_agents=n_agents) 
    for i in range(20 * n_agents):
        model.step()


if __name__ == '__main__':
    print(f'Mesa version: {mesa.__version__}')
    import objgraph

    print(objgraph.show_growth())
    main()
    print(objgraph.show_growth())
    main()
    print(objgraph.show_growth())
```

Output:
```
Mesa version: 3.3.1
function              22515    +22515
dict                  11268    +11268
tuple                  9130     +9130
cell                   4286     +4286
list                   4265     +4265
ReferenceType          3169     +3169
wrapper_descriptor     3112     +3112
set                    2918     +2918
getset_descriptor      2894     +2894
method_descriptor      2249     +2249
None
ReferenceType         5171     +2002
MyAgent               1000     +1000
list                  4378      +113
dict                 11273        +5
function             22517        +2
set                   2920        +2
method                 143        +2
WeakKeyDictionary        6        +2
AgentSet                 2        +2
Random                   2        +1
None
ReferenceType         7173     +2002
MyAgent               2000     +1000
list                  4491      +113
[...]

PR #3180, opened 2026-01-19
Fix for Memory leak

quaquel:
This is a bugfix for the memory leak identified in #3179.

The problem is that `Agent._ids` is a defaultdict that stores references to model instances. This was done to ensure that `uniqiue_id` is unique relative to a given model. However, since it is a class attribute, this reference persists across the Python process, preventing the entire model blob from being garbage-collected. 

There are various solutions
1. Use a weakref of the model
2. Use the hash of the model
3. Move the assignment of `unique_id` into `register_agent`
4. Add some method to `Model` to clean up (not yet explored), including removing the ref in `Agent._ids`

Here, I implement option 3 because, among the options tested, it was the fastest locally. I also moved away from `itertool.count` and instead just use an index that is being incremented. The main reason is that itertools.count will not be pickleable in Python 3.14, and count is overkill for the simple integer increments needed here anyway.

For reasons that escape me at present, it is still necessary to remove all agents from the model before it can be garbage-collected, at least in the updated test_examples. But when I try a minimal version of Boltzmann, this seems unnecessary. So, there might be some other memory issue remaining.


quaquel:
Ok, these results are better than when testing locally, so this seems a reasonable solution.

Now we just need to add support for pickling the model because `itertools.count` won't be pickleable anymore (not that difficult to achieve I think), and we need to figure out why the alliance formation model breaks.

codebreaker32:
Hi @quaquel 

I was also working on it to fix the example error and modifying register_agent passed all the test
```python
# In model.py

def register_agent(self, agent: Agent) -> None:
    # Check if the agent already has a valid ID
    if agent.unique_id is not None:
        # It's already registered! Don't touch the ID.
        # Just ensure it's in the internal list if needed.
        self._agents[agent] = None 
        return

    # Only generate a NEW ID if it completely lacks one
    agent.unique_id = next(self.agent_id_counter)
    self._agents[agent] = None
```

The reason I could've think of is(I might be wrong):

```python
# meta_agent.py
def add_constituting_agents(self, new_agents: set[Agent]):
    for agent in new_agents:
        self._constituting_set.add(agent)
        agent.meta_agent = self
        self.model.register_agent(agent)  # <--- Culprit
```

Suppose Agent A's id 5 is used as a key to store data in various dictionaries. Suddenly, the agent's unique_id attribute is overwritten to 105. The Python dictionaries are not corrupted, but they are now out of sync: the dictionary still holds the data under the old key (5), but the code is now trying to retrieve it using the new key (105). This mismatch leads to a KeyError or logic errors because the system looks for an ID that isn't there

quaquel:
@codebreaker32, yes, this is indeed the source of the problem. But in my view, agents should never call register_agent twice. In fact, a Mesa user should never have to call it if they use super properly in their custom agents. So, I see this as a bug in `create_meta_agents`, not in `register_agent` (See #3183 for my proposed fix.)

quaquel:
The benchmarks remain strange. For example, Boltman Wealth has no additional agents during the run, so nothing in this PR would change the runtime.

EwoutH (approved):
I'm going to pre-approve this to unblock you.

Please document the problem, the solution, the rejected alternatives and the reasoning behind all those well. If this ever bites us back we can trace this back.

quaquel:
> Please document the problem, the solution, the rejected alternatives and the reasoning behind all those well. If this ever bites us back we can trace this back.

It was included in the start post. In short, moving the assignment of `agent.unique_id` into the model ensures it remains unique within a model instance, while being faster than the alternatives I tested. 

The main issue encountered next is that ideally `model.register_agent` and `modelderegister_agent` should be never called by the user directly but allways via super on the agent. It might be good to document this (and hard refs in general) for Mesa 4.

PR #3183, opened 2026-01-20
Make create_meta_agent deterministic

### LINKED ISSUE #3182: `create_meta_agent` is non-deterministic
`create_meta_agent` uses a set to ensure that all agents are unique. However, this renders the ordering of agents non deterministic, and therefore the model as well.

quaquel:
This is a simple bug fix to ensure deterministic behavior of `create_meta_agent`. Instead of a set, we use a dict which is ordered, to ensure we have only unique agents. We do the same inside path_1 to also make this deterministic. Finally, we remove the call to `model.register_agent` in `add_constituting_agents`.

I am not sure about `remove_constituting_agents`, this does a deregister call, but I am fairly sure this should be agent.remove((). @tpike3, can you clarify?

Closes #3182 and #3184

### DIFF
--- a/mesa/experimental/meta_agents/meta_agent.py
+++ b/mesa/experimental/meta_agents/meta_agent.py
@@ -242,7 +242,9 @@ def create_meta_agent(
 
         else:
             constituting_agent = model.random.choice(constituting_agents)
-            agents = list( (dict.fromkeys(agents) | dict.fromkeys(constituting_agents)).keys() )
+            agents = list(
+                (dict.fromkeys(agents) | dict.fromkeys(constituting_agents)).keys()
+            )
             add_attributes(constituting_agent.meta_agent, agents, meta_attributes)
             add_methods(constituting_agent.meta_agent, agents, meta_methods)
             constituting_agent.meta_agent.add_constituting_agents(agents)


EwoutH:
Since dicts are ordered in modern Python, can we just use dict keys? I don’t know if there are any performance implications here.

quaquel:
> Since dicts are ordered in modern Python, can we just use dict keys? I don’t know if there are any performance implications here.

Which is exactly what I did :)

quaquel:
@tpike3, @EwoutH, For the memory leak bugfix to pass, this PR needs to be merged first.

edgeofchaos42:
@quaquel Thanks for crushing on this to fix the issue. 

Generally speaking, the difference between `remove_constituting_agents` and `agent.remove` is remove consituting agents removes the agent from the meta agent (e.g., they leave the alliance) but the agent is still a part of the model, agent.remove would remove them from the model entirely if I am not mistaken.

quaquel:
No, `model.deregister_agent` is what actually removes an agent from the model. `Agent.remove()` calls this. And, if you have a `CellAgent` or `ContinuousSpaceAgent`, `agent.remove()` also removes the agent from the space.

edgeofchaos42:
> No, `model.deregister_agent` is what actually removes an agent from the model. `Agent.remove()` calls this. And, if you have a `CellAgent` or `ContinuousSpaceAgent`, `agent.remove()` also removes the agent from the space.

Hmmm I still thing there is still a bug in meta-agents/ the logic needs to be more transparent. This is currently not how it is written and was not touched by this PR. My goal with this is to write down my understanding until i/someone else has time to address. Document for the group. 

User goal - remove agent from meta-agent but keep it in the model: 
`remove_constituting_agents` should only remove form the meta-agent not the model 

- line 384 in function `remove_constituting_agents` deregisters agent which will remove from the model not remove it from the meta-agent.
- Needs to be updated to just remove it from the meta-agent
- Needs to validate that when agent removed from model, it is also removed from meta-agent

quaquel:
@edgeofchaos42, I agree. It is why I pinged the original maker, @tpike3, to have a look as well, and this is why I did not change `remove_constituting_agent`. 

`model.deregister_agent` is the method that is used to remove agents from a model. But the proper way to do this is to call `agent.remove()`, since there might be more cleanup needed than just removing the model from the central agent registration system in Mesa. So, indeed, what is the intent of `remove_constituting_agent`? Is it just to remove a single agent from a meta agent, or is it to fully remove the constituting agent from the model?

PR #3188, opened 2026-01-20
Add event scheduling and time progression API v2

EwoutH:
Based on the discussions and accumulating insights in #3155 and https://github.com/mesa/mesa/discussions/2921#discussioncomment-15524497, a new implementation that implements the single `schedule()` method.

Builds on #3155. Can either be directly merged or after #3155 (I recommend former).

What's quite cool is that it schedules a `model.step()` event every 1 time. So like a ABMSimulator by default. Like a hearth beat. Can easily be turned off for full DEVS.

That also makes it super super backwards compatible.

### Design decisions
I started from the following design decisions:
* We want a single, powerful schedule function. Utility functions come second and syntactic suger like decorators third.
* Same goes for run: Powerful run function, then utility functions.
* The RecurringEvent is a very useful concept. It should have methods to pause, resume and stop.
* The `step` is going to be the hearth beat of Mesa. This keeps backwards compatibility and a familiar way for basic models, with the only difference for them being to run a model with run() instead of repeated step calls. Internally, it's just a scheduled function with interval 1, and can be overwritten.
* The `model.schedule()` method returns the `Event` or `RecurringEvent` object directly, which in case of Recurring has methods like `pause()`, `resume()`, and `cancel()` for user control.
* Users who need to modify or cancel scheduled events simply keep a reference to the returned object (e.g., `self.trading = model.schedule(self.trade, interval=1)`), while one-off events that don't need later control can ignore the return value.
* The default `step` method is automatically scheduled as a `RecurringEvent` with `interval=1` and stored in `self.step_event`, making it easy to introspect or modify the model's heartbeat.
* We're not implementing a central registry of central events for now, users manage their own event references as needed.

### Methods
Some methods
```Python
# Schedule method
model.schedule(
    callback,
    start_at=None,      # Absolute time to start
    start_after=None,   # Delay before starting  
    interval=None,      # Time between recurrences (None = one-off, number or callable)
    count=None,         # Max executions (None = infinite for recurring, 1 for one-off)
    end_at=None,        # Absolute time to stop
    end_after=None,     # Duration to run (from first execution)
    priority=Priority.DEFAULT,
    args=None,
    kwargs=None,
) -> SimulationEvent | RecurringEvent
```

Return Values
- One-off events return `SimulationEvent`
- Recurring events return `RecurringEvent` with `pause()`, `resume()`, `stop()` methods

Default Step Behavior
- `Model.__init__` automatically schedules `self.step` as a `RecurringEvent` with `interval=1`
- Stored in `self.step_event` for user access/modification
- Users override `step()` method, the scheduling is automatic

Convenience Methods
- `model.schedule_at(callback, time, ...)` → wraps `schedule(start_at=time)`
- `model.schedule_after(callback, delay, ...)` → wraps `schedule(start_after=delay)`

Run Methods
- `model.run_for(duration)` - run for N time units
- `model.run_until(end_time)` - run until absolute time
- `model.run_while(condition)` - run while condition is true
- `model.run_next_event()` - single event execution (debugging)

### TODO
- [ ] Fix recurring event pausing
- [ ] Update PR description
- [ ] Update migration guide

Follow-up PRs:
- Deprecate simulators
- Update example models which use simulators
- Update tutorial

_(please do not update, rebase or touch my branch in any way)_

### DIFF
--- a/tests/test_timeflow.py
+++ b/tests/test_timeflow.py
@@ -1,9 +1,11 @@
 """Tests for unified time and event scheduling API."""
 
+import random
+
 import pytest
 
 from mesa import Agent, Model
-from mesa.experimental.devs.eventlist import Priority
+from mesa.experimental.devs.eventlist import Priority, RecurringEvent
 from mesa.timeflow import RunControl, Scheduler
 
 
@@ -24,6 +26,7 @@ class TestScheduler:
 
         assert event.time == 10.0
         assert not event.CANCELED
+        # Note: Model.__init__ schedules step, so we use a fresh scheduler here
         assert len(scheduler.event_list) == 1
 
     def test_schedule_at_past_raises_error(self):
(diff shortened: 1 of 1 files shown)


EwoutH:
If you think, +1800 lines of code, what are you smoking?
- over half are tests
- another few hunderd docs
- a huge slew of backwards compatibility stuff (both for increasing time with step() and the simulators)
- Functional code is about 500 lines (with a lot of inline docs, maybe too much), which is about the same as simulators we're going to remove.

quaquel:
I am sorry, but I don't like this API. Notions such as `start_at` and `end_at` only make sense for recurring events. I find them confusing for scheduling individual events where `at` or `after` make sense, and the other arguments don't matter. So, to me, it would make sense to at least separate `schedule` and `schedule _repeatedly`. Also, picking up on the idea that recurring events are processes, would it not make more sense to have `start_at` and `start_from`. This is also better aligned with `end_at` and `end_after`.

EwoutH:
Thanks for your perspective. It may indeed make sense to split into `schedule` and `schedule_recurring`, especially since they return different objects.

EwoutH:
Something else: I’m a bit in doubt if `RecurringEvent` is the right abstraction, or it should be something like a `EventGenerator` that just schedules normal events.

quaquel:
> Something else: I’m a bit in doubt if RecurringEvent is the right abstraction, or it should be something like a EventGenerator that just schedules normal events.

Good question. The name is clear, but adding methods like start and stop makes it more than just a recurring event. Generator does not really capture this either. So, I don't know yet.

EwoutH:
Thanks for the insights. To be continued in:
- https://github.com/mesa/mesa/discussions/2921#discussioncomment-15562881

PR #3189, opened 2026-01-21
Fix: Replace 'assert ValueError' with proper return None

codebyNJ:
### Summary
Fixed a no-op assertion in the DataCollector test that masked intended behavior for invalid reporter input; ensures the reporter returns `None` for invalid input instead of silently doing nothing.

### Bug / Issue
The `MockModel.test_model_calc_comp` test used `assert ValueError` in an else branch (a no-op since is truthy). This meant the failure case (division by zero) was not handled as intended and could hide problems. Expected: the reporter returns `None` for invalid input; Actual: the code did nothing meaningful.

### Implementation
**Modified test_datacollector.py (MockModel.test_model_calc_comp)**
- Replaced `assert ValueError` with a comment.
- Explicitly return `None`.


quaquel:
I am sorry, but I don't understand the test code itself. This changes the return in a MockModel for one of the model reporters. But if a value is invalid, should that not always raise an exception and thus be handled inside the data collector? 

To be clear, the current code makes no sense to me. Jus trying to understand the wider context within which this PR is being used.

codebyNJ:
Let me clarify:

The original code has a bug: assert ValueError is a no-op. It doesn't raise an exception - it just checks that the ValueError class is truthy (which it always is). The code was always returning None, the assert line did nothing.

The test expects None: `Line 165-166` explicitly asserts that `model_calc_fail` returns None:

```python
for element in data_collector.model_vars["model_calc_fail"]:
    assert element is None
```

DataCollector doesn't handle exceptions: Looking at `datacollection.py:378`, there's no try/except around reporter calls. If a reporter raises an exception, it would crash the simulation.

My PR just removes the no-op: I'm not changing the design - just removing the useless assert ValueError and making the intent clear with a comment.

However, you raise a valid design question: Should DataCollector catch reporter exceptions and store None (or some error value) instead? That would be a separate enhancement. My PR just fixes the misleading no-op assertion in the existing test.

quaquel:
> However, you raise a valid design question: Should DataCollector catch reporter exceptions and store None (or some error value) instead? That would be a separate enhancement. My PR just fixes the misleading no-op assertion in the existing test.

Thanks for the clarification. On this, let's not touch the current data collector and instead focus on the emerging ideas arround dataset and collector. I am going to merge this pr in the meantime.

PR #3191, opened 2026-01-21
Replace `Sequence` with `to_list` for `AgentSet`

### LINKED ISSUE #3185: Replace sequence with `to_list`
AgentSet currently subclasses `collections.abc.Sequence`. This is very slow, and typically not needed. **For Mesa 4**, I suggest we remove the sequence behavior from AgentSet and replace it with a `to_list` helper method. This reduces the potential for users to use the very slow sequence behavior.

Sonu0305:
### Summary
Removes `collections.abc.Sequence` inheritance from `AgentSet` and introduces a `to_list()` helper method. `AgentSet` is no longer subscriptable, preventing unintended slow indexing and slicing operations.
Closes #3185

### Motive
`AgentSet` currently subclasses `collections.abc.Sequence` by default. This behavior is often slow.

### Additional Notes
This is a breaking change for users relying on indexing or slicing `AgentSet` directly. It is intended for the Mesa 4 release cycle.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -13,7 +13,7 @@ import operator
 import warnings
 import weakref
 from collections import defaultdict
-from collections.abc import Callable, Hashable, Iterable, Iterator, MutableSet, Sequence
+from collections.abc import Callable, Hashable, Iterable, Iterator, MutableSet
 from random import Random
 
 # mypy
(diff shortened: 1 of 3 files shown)


quaquel:
Thanks for this PR. As clearly indicated, it is something for Mesa 4. We haven't yet opened a Mesa 4 branch. So I am closing this PR for now.

PR #3192, opened 2026-01-21
Fix seed logic to ensure reproducibility

### LINKED ISSUE #3190: Reproducibilty Trap in model.py
This code block is really confusing me
https://github.com/mesa/mesa/blob/b6e96d1351c756c6ba7ef5de35a380f63d3700df/mesa/model.py#L122-L127

It uses a `try-except` block to handle compatibility issues but the logic fails to capture the seed value in the successful cases.

* **If `rng` is `None`:** `random.Random(None)` succeeds, skipping the `except` block. `seed` remains `None`.
* **If `rng` is an `int`:** `random.Random(int)` succeeds, skipping the `except` block. `seed` remains `None`.

As a result, `self._seed` is set to `None` at the end of the block (`self._seed = seed`). This breaks reproducibility and causes `model.scenario.rng` to be `None`

Please correct me if I am wrong!

codebreaker32:
### Summary

This PR fixes a bug in `Model.__init__` where `self._seed` was incorrectly set to `None` even when a valid `rng` (integer) was provided. This ensures that the `Scenario` instance is initialized with the correct seed and that `reset_randomizer()` works as intended.

### Bug / Issue

Closes #3190 

### Implementation

I refactored the initialization logic in `mesa/model.py` to use a `try-except-else` pattern, strictly separating seed generation from validation.

* **Case A (`rng=None`):** Explicitly generates and stores an integer seed. This ensures **post-hoc reproducibility** (users can check `model.scenario.rng` to debug a random run that wasn't explicitly seeded).
* **Case B (`rng=42`):** The `else` block now correctly captures `seed = rng` when `random.Random(rng)` succeeds.
* **Case C (`rng=Generator`):** The `except` block handles Numpy generators by extracting a valid integer seed.

### DIFF
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -86,7 +86,7 @@ def test_seed():
     model_explicit = Model(rng=42)
     assert model_explicit._seed == 42
     assert model_explicit.scenario.rng == 42
-    assert model_explicit.seed == 42
+    assert model_explicit._seed == 42
 
     # Case 2: None
     # The model should generate a integer seed


mesa/model.py
```diff
@@ -119,11 +119,24 @@ def __init__(
                 self.rng.bit_generator.state
             )  # this allows for reproducing the rng
 
-            try:
-                self.random = random.Random(rng)
-            except TypeError:
+            # We must ensure 'seed' captures the actual integer used,
+            if rng is None:
+                # Case A: No rng provided. Generate a random integer seed.
```
quaquel:
```suggestion
                # Case A: No seed provided. Generate a random integer seed.
```

mesa/model.py
```diff
@@ -119,11 +119,24 @@ def __init__(
...
                 self.random = random.Random(seed)
+            else:
+                try:
+                    # Case B: Try to use the provided rng (e.g., int) directly
+                    self.random = random.Random(rng)
+                except TypeError:
+                    # Case C: rng is a Generator/BitGenerator (not valid for random.Random)
+                    # Extract an integer seed from it
+                    seed = int(self.rng.integers(np.iinfo(np.int32).max))
+                    self.random = random.Random(seed)
```
quaquel: This is the same as the None case. So it feels strange to repeat the code.

Might a match case statement not work better here

match seed:
    case None | GeneratorLike : # this is psuedocode, so bear with e
        > do the self.rng thing
    case SeedLike:
        >  use is so random.Random(rng)
     case _
         > raise value error.

quaquel: it can probebly be simlified further by just checking seedlike only and otherwise falling back on the rng.integer thing

codebreaker32: Did other way round

quaquel:
can you add tests to ensure the original bug is now discovered (and the fixed with this PR)?

tests/test_model.py
```diff
@@ -80,6 +80,29 @@ def test_rng(rng=23):
     )
 
 
+def test_seed():
+    """Test seed persistence and generation for various rng inputs."""
+    # Case 1: Explicit Integer
+    model_explicit = Model(rng=42)
+    assert model_explicit._seed == 42
+    assert model_explicit.scenario.rng == 42
```
quaquel, line 88:
```suggestion
    assert model_explicit.scenario.rng == 42
    assert model_explicit.seed == 42
```

tests/test_model.py
```diff
@@ -80,6 +80,30 @@ def test_rng(rng=23):
...
+    assert model_explicit.seed == 42
```
quaquel:
```suggestion
    assert model_explicit._seed == 42
```

PR #3194, opened 2026-01-22
Fix: Redundant agent registration and meta-agent determinism (#3184)

### LINKED ISSUE #3184: additional bugs in meta_agents
#3182 identified 1 source of indeterminacy in the meta_agents code.

However there are 2 more problems with it

1. In `create_meta_agent`, path 1 there are two more set operations
2. In `add_constituting_agents`, model.register_agent is called. This is redundand because if you call the `__init__` of Agent, this allways ends with calling `model.register_agent`.  Moreover, because oft he indended bugfix for #3179, this breaks because unique_ids on already existing agents are changed.

Nithurshen:
### Summary

This PR addresses bugs reported in #3184 regarding agent identity stability and non-deterministic behavior in the experimental meta-agent feature. It ensures `register_agent` is idempotent and that meta agent method resolution is deterministic.

### Bug / Issue

Fixes #3184

- Previously, calling `model.register_agent(agent)` on an agent that was already registered would overwrite the agent's `unique_id` with a new incremented value. This broke agent identity and tracking.
- The `create_meta_agent` function used a Python set to deduplicate agent types when aggregating methods (`{type(agent) for agent in agents}`). Because sets are unordered, the method resolution order was non-deterministic if multiple constituting agents defined methods with the same name.

### Implementation

- `mesa/model.py`: Modified `register_agent` and added an early return (`if agent in self._agents: return`) to make the method idempotent. This prevents the `unique_id` from being overwritten if the method is called multiple times for the same agent.
- `mesa/experimental/meta_agents/meta_agent.py`: In the `add_methods` helper function, replaced the set comprehension with list(`dict.fromkeys(type(agent) for agent in agents)`). This change ensures that agent types are processed in a deterministic order, so the resulting meta agent always inherits methods in a predictable sequence.

### Testing

Test:
```python
import unittest
from mesa.model import Model
from mesa.agent import Agent


class MockAgentA(Agent):
    def get_tag(self):
        return "Tag_A"


class MockAgentB(Agent):
    def get_tag(self):
        return "Tag_B"


class TestIssue3184(unittest.TestCase):
    def test_redundant_registration_preserves_id(self):
        model = Model()
        agent = Agent(model)
        original_id = agent.unique_id

        print(f"\n[Test Registration] Created Agent with ID: {original_id}")

        model.register_agent(agent)

        self.assertEqual(
            agent.unique_id,
            original_id,
            f"FAIL: Agent unique_id changed from {original_id} to {agent.unique_id} after redundant registration.",
        )
        print(f"[Test Registration] PASS: Agent ID preserved as {agent.unique_id}")

        self.assertEqual(
            len(model._agents), 1, "FAIL: Agent duplicated in model._agents"
        )
        self.assertEqual(
            len(model.agents), 1, "FAIL: Agent duplicated in model.agents AgentSet"
        )
        print("[Test Registration] PASS: No duplicate agents found in registry.")

    def test_meta_agent_method_determinism(self):
        model = Model()
        agent_a = MockAgentA(model)
        agent_b = MockAgentB(model)

        print("\n[Test Determinism] Checking method resolution order...")

        meta_agent_1 = create_meta_agent(
            model,
            "MetaAgent1",
            [agent_a, agent_b],
            Agent,
            assume_constituting_agent_methods=True,
        )

        tag_1 = meta_agent_1.get_tag()
        self.assertEqual(
            tag_1,
            "Tag_B",
            f"FAIL: Expected 'Tag_B' (last in list) but got '{tag_1}'. Order not preserved for [A, B].",
        )
        print(f"  [Case 1] Input [A, B] -> Result '{tag_1}' (Matches B: Correct)")

        meta_agent_2 = create_meta_agent(
            model,
            "MetaAgent2",
            [agent_b, agent_a],
            Agent,
            assume_constituting_agent_methods=True,
        )

        tag_2 = meta_agent_2.get_tag()
        self.assertEqual(
            tag_2,
            "Tag_A",
            f"FAIL: Expected 'Tag_A' (last in list) but got '{tag_2}'. Order not preserved for [B, A].",
        )
        print(f"  [Case 2] Input [B, A] -> Result '{tag_2}' (Matches A: Correct)")


if __name__ == "__main__":
    unittest.main()
```

Result:
```bash
============================================================================== test session starts ===============================================================================
platform darwin -- Python 3.13.8, pytest-9.0.2, pluggy-1.6.0
rootdir: /Users/nithurshen/Projects/mesa
configfile: pyproject.toml
plugins: mock-3.15.1, playwright-0.7.2, anyio-4.12.0, base-url-2.1.0, ipywidgets-1.55.1, cov-7.0.0
collected 2 items                                                                                                                                                                

test.py ..                                                                                                                                                                 [100%]

=============================================================================== 2 passed in 0.35s ================================================================================
```

### Additional Notes

- I am also open to add this test case

### DIFF
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -214,6 +214,9 @@ class Model[A: Agent, S: Scenario]:
             is no need to use this if you are subclassing Agent and calling its
             super in the ``__init__`` method.
         """
+        if agent in self._agents:
+            return
+
         self._agents[agent] = None
         agent.unique_id = self.agent_id_counter
         self.agent_id_counter += 1
--- a/mesa/experimental/meta_agents/meta_agent.py
+++ b/mesa/experimental/meta_agents/meta_agent.py
@@ -189,7 +189,7 @@ def create_meta_agent(
         meta_methods (Dict[str, Callable]): methods to be added to the meta-agent.
         """
         if assume_constituting_agent_methods:
-            agent_classes = {type(agent) for agent in agents}
+            agent_classes = list(dict.fromkeys(type(agent) for agent in agents))
             if meta_methods is None:
                 # Initialize meta_methods if not provided
                 meta_methods = {}


mesa/model.py
```diff
@@ -214,6 +214,9 @@ def register_agent(self, agent: A):
             is no need to use this if you are subclassing Agent and calling its
             super in the ``__init__`` method.
         """
+        if agent in self._agents:
```
quaquel, line 217: no, we are not going to change agent registration for an experimental feature. This most be resolved inside meta_agents not here.

quaquel:
I am closing this PR. I agree the issue needs to be addressed. But I don't yet know how. One hard requirement is that it must be fixed inside meta_agents. We are not going to change the agent registration and deregistration. This is at the core of how Mesa manages agents.

PR #3197, opened 2026-01-23
POC: Exception Hierarchy for Mesa 4.0

Nithurshen:
### Summary

This PR introduces a standardized exception hierarchy for Mesa 4.0, acting as a Proof of Concept (POC) for the architectural proposal discussed recently. It adds a new `mesa.errors` module containing a base `MesaError` class and refactors `SingleGrid` to use a specific `CellNotEmptyError` instead of a generic `Exception`.

### Motive

Currently, Mesa relies heavily on generic Python exceptions (like `Exception` or `ValueError`) for simulation logic failures. This makes it difficult for users to programmatically handle simulation errors without risking the suppression of unrelated system errors (like `KeyboardInterrupt` or `MemoryError`).

By introducing a dedicated hierarchy rooted in `MesaError`, we allow users to write cleaner, safer error-handling code. This PR demonstrates this pattern by fixing the "Cell not empty" error in `SingleGrid`, which was previously raising a raw `Exception`.

### Implementation

1. **New Module:** Created `mesa/exceptions.py`.
2. **Base Class:** Defined `MesaException(Exception)` as the root for all Mesa-specific exceptions.

### Additional Notes

* This is a **Proof of Concept**. If approved, subsequent PRs can migrate other parts of the codebase to this hierarchy incrementally.


quaquel:
Thanks for this PR. I think its still a bit premature. I don't agree with `mesa.errors.py`, and `mesa.space` is a bad choice to start with because it's about to be removed. 

Also `CellNotEmptyError` is not a good name. I prefer Exception over Error. Also, the issue for at least discrete_space is not that the cell is not empty. Rather, It is that the cell is at capacity. So CellFullException might be more on point.

quaquel:
Thanks for the update. I think this is moving in a good direction. 

What other generic exceptions should be replaced in mesa.discrete_space.

Nithurshen:
Thanks.
 I have three additional candidates for replacement in mind within `mesa.discrete_space` (`AgentMissingException`, `CellMissingException`, and `ConnectionMissingException`).

I have posted the detailed breakdown and rationale for these additions in the main Discussion [thread](https://github.com/mesa/mesa/discussions/3196#discussioncomment-15585752).

quaquel:
> I have three additional candidates for replacement in mind within mesa.discrete_space (AgentMissingException, CellMissingException, and ConnectionMissingException).

Sounds good to me. Where possible, consider using the `raise from` syntax. For example, the key error cases could use this. Basically, this wraps the source (key error) inside our more informative custom exception. Please feel free to add these to this PR.

Nithurshen:
> > I have three additional candidates for replacement in mind within mesa.discrete_space (AgentMissingException, CellMissingException, and ConnectionMissingException).
> 
> Sounds good to me. Where possible, consider using the `raise from` syntax. For example, the key error cases could use this. Basically, this wraps the source (key error) inside our more informative custom exception. Please feel free to add these to this PR.

Done

Nithurshen:
@EwoutH,
I am thinking of `PropertyLayerDimensionException`, `DuplicatePropertyLayerException`, and `InvalidRadiusException` again for `mesa.discrete_space`.
I have updated the discussion thread in detail. Kindly let me know your opinion.

quaquel:
Just a kind request: please refrain from pinging maintainers. We get the updates and try to respond as soon as we can. Sometimes, we as maintainers might ping each other because we are aware of our respective expertise across different parts of the codebase.

Nithurshen:
As discussed in  the [thread](https://github.com/mesa/mesa/discussions/3196#discussioncomment-15596602),
`MesaSpaceException`, and `DimensionException` has been implemented.

quaquel:
I like the direction this is moving in. Can you next take a look at the experimental continouous space and see what custom exception needs there might be there and how those might be integrated cleanly into this hierarchy?

Nithurshen:
If we establish `DimensionException` as a shared/generic error, `ContinuousSpace` fits in cleanly. I propose adding a `ContinuousSpaceException` base class and one specific logic error:

### Proposed Hierarchy
```text
MesaException
├── DimensionException (Shared: used for invalid width/height or grid shape mismatches)
├── DiscreteSpaceException
│   └── ... (CellFull, etc.)
└── ContinuousSpaceException (New Module Base)
    └── OutOfBoundsException
```

Rationale for `OutOfBoundsException`:
In a continuous space with `torus=False`, agents often hit the edge. Currently, we might rely on clamping or generic errors.
• Use Case: A user wants custom boundary logic (e.g., "if agent hits wall, reverse velocity" or "if agent hits wall, remove agent").
• Benefit: They can catch `OutOfBoundsException` to trigger this logic cleanly, rather than checking coordinates manually after every move step.

quaquel:
Do we need the distinction between ContinuousSpaceException and DiscreteSpaceException, or can we simplify this to just a generic SpaceException?

I'll get back to you on the`OutOfBoundsException`.

Nithurshen:
> Do we need the distinction between ContinuousSpaceException and DiscreteSpaceException, or can we simplify this to just a generic SpaceException?
> 
How about, as of now we implement all the exceptions we want under `ContinousSpaceException` and `DiscreteSpaceException`, then judging the number of exceptions under each branch, we decide if we want a unified `SpaceException`.
Or another option is, we implement `SpaceException`, and make `ContinousSpaceException` and `DiscreteSpaceException` inherit from it. But I think this option will make the hierarchy more complicated.
> I'll get back to you on the`OutOfBoundsException`.

Sure.

quaquel:
Do we need a separate `ContinuousSpaceException` and `DiscreteSpaceException`?

Nithurshen:
> Do we need a separate `ContinuousSpaceException` and `DiscreteSpaceException`?

I don't think we will need separate exceptions. I have implemented a unified `SpaceException` to take care of this.

mesa/exceptions.py
```diff
@@ -0,0 +1,79 @@
...
+        """Initialize the exception.
+
+        Args:
+            coordinate: The coordinate tuple of the full cell.
+        """
+        self.coordinate = coordinate
+        super().__init__(f"Cell at coordinate {coordinate} is full.")
+
+
+class AgentMissingException(SpaceException):
```
quaquel: there are other places that this might be used (e.g., model.deregister_agent), so this might subclass mesaexception, rather than space exception.

Nithurshen: Done

quaquel:
@EwoutH, I am happy with how this is turning out and would be in favor of merging this as part of 4.0 while expanding the use of custom exceptions where appropriate. Would like to hear your thoughts as well before merging this.

EwoutH:
Sorry, missed it last week. But it looks good and clean, thanks both!

PR #3198, opened 2026-01-23
Optimise mesa_signals by skipping signals for empty subscribers to reduce subsequent overheads

codebreaker32:
### Summary

This PR optimizes the `mesa_signal` module by flattening the internal subscriber storage and introducing a "fast path" that bypasses the signal emission pipeline for `Observable` properties with no active subscribers.

### Motive

When an agent updates an observable (e.g., `agent.wealth = 10`), the following sequence currently occurs regardless of subscriber status:

1. **Lookup overhead:** `getattr(self, _wealth)` is called to fetch the `old_value`. This is wasteful if no one is listening.
2. **Allocation overhead:** The `notify()` method creates a new `Message` object. This allocates memory for a signal payload that will immediately be discarded.
3. **Function Call Overhead:** `self.notify(...)` and `self._mesa_notify(message)` are called.
4. **The Check:** Finally, inside `_mesa_notify`, the system checks the dictionary.

See [https://github.com/mesa/mesa/issues/3131#issuecomment-3785576047](https://github.com/mesa/mesa/issues/3131#issuecomment-3785576047) also.

### Implementation

The optimization relies on flattening the data structure to make the subscriber check as cheap as possible 

1. We changed the internal `subscribers` structure from a nested dictionary (`dict[str, dict[str, list]]`) to a flat dictionary using tuple keys (`dict[tuple[str, str], list]`). This reduces pointer chasing and simplifies the lookup logic.

2. Added a helper method `_has_subscribers` that utilizes the new flat structure to check for listeners in a single step:

```python
def _has_subscribers(self, name: str, signal_type: str | SignalType) -> bool:
    # Single tuple key lookup
    key = (name, signal_type)
    if key not in self.subscribers:
        return False
    return len(self.subscribers[key]) > 0

```

3. Updated `BaseObservable.__set__` to check `_has_subscribers` immediately. If it returns `False`, the method exits early, skipping `getattr`, `Message` allocation, and notification calls.

```python
def __set__(self, instance: HasObservables, value):
    # Early exit if no one is listening
    if not instance._has_subscribers(self.public_name, SignalType.CHANGE):
        return

    instance.notify(
        self.public_name,
        getattr(instance, self.private_name, self.fallback_value),
        value,
        SignalType.CHANGE,
    )

```

### DIFF
--- a/mesa/experimental/mesa_signals/mesa_signal.py
+++ b/mesa/experimental/mesa_signals/mesa_signal.py
@@ -276,7 +276,9 @@ class HasObservables:
 
     # we can't use a weakset here because it does not handle bound methods correctly
     # also, a list is faster for our use case
-    subscribers: dict[tuple[str, str], list]   # (observable_name, signal_type) -> list of weakref subscribers
+    subscribers: dict[
+        tuple[str, str], list
+    ]  # (observable_name, signal_type) -> list of weakref subscribers
     observables: dict[str, set[str]]
 
     def __init__(self, *args, **kwargs) -> None:


mesa/experimental/mesa_signals/mesa_signal.py
```diff
@@ -294,6 +299,21 @@ def _register_signal_emitter(self, name: str, signal_types: set[str]):
...
+        """Check if there are any subscribers for a given observable and signal type.
+
+        Args:
+            name: name of the Observable
+            signal_type: the type of signal on the Observable to check for
+
+        Returns:
+            bool: True if there are subscribers, False otherwise
+        """
+        if name not in self.subscribers or signal_type not in self.subscribers[name]:
```
quaquel: might we not further improve this by changing how we store subscribers?

Subscribers is currently `dict[str, dict[str, list]]`. We might just flatten this by doing
`dict[tupe[str, str], list]`. So we make a tuple of name and event and store subscribers in that way. 

The main drawback is when you want to clear subscriptions for a given observable, you have to iterate over the name and signal type but that is a small cost to pay in my view.

codebreaker32: great suggestion. Flattening the dictionary reduces the pointer chasing in the "hot path" (the `_has_subscribers` check inside `__set__`) to a single lookup. Since clearing subscriptions is a rare operation compared to variable assignment, the trade-off is definitely worth it.

mesa/experimental/mesa_signals/mesa_signal.py
```diff
@@ -272,13 +276,13 @@ class HasObservables:
 
     # we can't use a weakset here because it does not handle bound methods correctly
     # also, a list is faster for our use case
-    subscribers: dict[str, dict[str, list]]
+    subscribers: dict[tuple[str, str], list]
```
quaquel: can you add a comment above this explaining the content: so (name, signal_type)

mesa/experimental/mesa_signals/mesa_signal.py
```diff
@@ -294,6 +298,13 @@ def _register_signal_emitter(self, name: str, signal_types: set[str]):
         """
         self.observables[name] = signal_types
 
+    def _has_subscribers(self, name: str, signal_type: str | SignalType) -> bool:
+        """Check if there are any subscribers for a given observable and signal type."""
+        key = (name, signal_type)
+        if key not in self.subscribers:
```
quaquel, line 306: we are using a defaultdict, so this check is not needed. We allways get a list back

quaquel:
I think this is basically ready to go.

mesa/experimental/mesa_signals/mesa_signal.py
codebreaker32: But If I removed the check and just write if `len(self.subscribers[key]) > 0`, the defaultdict will automatically insert an empty list [] for every single property(mutate the subscriber) we'll ever check. That will result in unnecessary memory bloat if we'll have large number of keys

quaquel: It's a micro optimization. I doubt it matters in everyday use, where keys probably are. quite small. But fine with me to do it this way.

PR #3199, opened 2026-01-24
Add DataFrame support to Agent creation

### LINKED ISSUE #3186: Better DataFrame support for `Agent.create_agents()`
I think we can improve the ergonomics of creating agents from pandas DataFrames. Currently it requires verbose `.tolist()` conversions for each column.

### Background
Users may initialize agents from tabular data (CSV, Parquet, database queries).

```python
# Example synthetic population data
df = pd.DataFrame({
    'age': [25, 34, 45, 67, 29],
    'bmi': [22.5, 28.1, 31.2, 24.8, 26.3],
    'condition_status': ['healthy', 'at_risk', 'chronic', 'healthy', 'at_risk'],
    'income': [45000, 62000, 51000, 38000, 71000]
})
```
The current `create_agents()` API requires converting each DataFrame column to a list:

```Python
class HealthAgent(Agent):
    def __init__(self, model, age, bmi, condition_status, income):
        super().__init__(model)
        self.age = age
        self.bmi = bmi
        self.condition_status = condition_status
        self.income = income

# Current approach - verbose and inefficient
agents = HealthAgent.create_agents(
    model=model,
    n=len(df),
    age=df['age'].tolist(),        # Manual conversion
    bmi=df['bmi'].tolist(),        # Manual conversion
    condition_status=df['condition_status'].tolist(),  # Manual conversion
    income=df['income'].tolist()   # Manual conversion
)
```

### Potential solutions
We're considering two approaches (not mutually exclusive):

### Option 1: Accept DataFrame columns directly
Allow pandas Series as arguments without manual conversion:

```python
# Proposed - cleaner API
agents = HealthAgent.create_agents(
    model=model,
    n=len(df),
    age=df['age'],           # No .tolist() needed
    bmi=df['bmi'],
    condition_status=df['condition_status'],
    income=df['income']
)
```

### Option 2: Add `df` parameter for direct DataFrame input
Add a dedicated parameter that accepts a DataFrame:

```python
# Most concise - auto-map all columns
agents = HealthAgent.create_agents(
    model=model,
    df=df
)

# If you only want certain columns, just filter them yourself before input:
agents = HealthAgent.create_agents(
    model=model,
    df=df[['age', 'bmi', 'condition_status', 'income']]
)

# Mix DataFrame with additional parameters
agents = HealthAgent.create_agents(
    model=model,
    df=df,
    initial_energy=100  # Same value for all agents
)

# Use DataFrame subset with overrides
agents = HealthAgent.create_agents(
    model=model,
    df=df[['age', 'bmi']],
    condition_status='healthy',  # Override for all
    income=df['adjusted_income']  # Mix with Series
)
```
### Questions for discussion
1. Which option should we implement? Both? Start with Option 1, add Option 2 later?
2. For Option 2, how should we handle conflicts if df contains a column 'age' AND the user passes age=... explicitly?
3. Should we also support other tabular formats like Polars DataFrames or NumPy structured arrays?

### Related
- Discussion #3181

falloficaruss:
### Summary
This PR improves the integration between pandas DataFrames and agent creation in Mesa. It enables `Agent.create_agents` to accept pandas Series directly and introduces a dedicated `Agent.from_dataframe` factory method.

### Motive
Currently, initializing agents from tabular data (e.g., CSVs, Parquet) is verbose because users must manually convert every column to a list using `.tolist()`. As discussed in #3186, improving this functionality aligns Mesa with other data-centric libraries ([like Hugging Face Datasets or Ray](https://github.com/mesa/mesa/issues/3186#issuecomment-3782215523)) and makes working with synthetic populations or large datasets much more ergonomic.

### Implementation
1. Direct Series Support: Updated the argument parsing logic in `Agent.create_agents` to recognize `pandas.Series` as a valid sequence type. This ensures that Series passed as arguments are distributed across agents row-by-row, rather than being treated as a single atomic value.
2. Optimized Factory Method: Added `Agent.from_dataframe(model, df, **kwargs)`. This method uses a row-based loop (`df.to_dict(orient="records")`) for efficiency, as suggested during review.
3. Conflict Handling: The `from_dataframe` method automatically filters out model and n columns from records to avoid clashing with the `Agent` constructor's positional arguments.
4.Performance: Implemented a lazy-check for the pandas module in `create_agents` to ensure zero performance overhead for users who do not have pandas imported.

### Usage Examples
```python
import pandas as pd
from mesa.agent import Agent

# Sample Data
df = pd.DataFrame({
    'age': [25, 34, 45],
    'income': [45000, 62000, 51000],
    'model': ['unused', 'unused', 'unused'] # Will be filtered out
})

# --- create_agents (Now supports Series directly) ---
agents = MyAgent.create_agents(
    model=model,
    n=len(df),
    age=df['age'],
    income=df['income']
)

# --- from_dataframe (New Factory Method) ---
# Automatically maps 'age' and 'income' columns to __init__ arguments
agents = MyAgent.from_dataframe(model, df)

# With additional constant arguments (applied to all agents)
agents = MyAgent.from_dataframe(
    model, 
    df, 
    agent_type='citizen'
)
```

### Additional Notes
- Resolves #3186.
- pandas is already a core dependency of Mesa, so no new dependencies are introduced.

### DIFF
--- a/tests/test_agent.py
+++ b/tests/test_agent.py
@@ -318,6 +318,8 @@ def test_agent_from_dataframe():
     df_with_index.index = range(100, 100 + n)
     agents = TestAgent.from_dataframe(model, df_with_index, extra_attr=5)
     assert [a.value for a in agents] == list(range(n))
+
+
 def test_agent_add_remove_discard():
     """Test adding, removing and discarding agents from AgentSet."""
     model = Model()


quaquel:
Thanks for this PR. Can you ensure you get to 100% test code coverage, or at least as close as is reasonable? You can check the coverage report to see which parts are not yet covered by tests.

@EwoutH,  you started the original issue. What do you think? Do you prefer overloading `create_agent`, or do you prefer a separate method `from_dataframe` as done here and as argued for by several people, including me, in the original issue.

falloficaruss:
@quaquel Got the 100% test code coverage. Also, interesting performance benchmark results.
Any thoughts here?

quaquel:
The benchmark results make no sense. Nothing here touches code that is used in the benchmarks. But we have been seeing funky benchmark results a lot lately. I'll try to take a closer look at the PR itself soonish.

mesa/agent.py
```diff
@@ -135,6 +134,24 @@ def create_agents[T: Agent](
...
+            model: The model instance.
+            df: The pandas DataFrame. Each row represents an agent.
+            kwargs: Additional keyword arguments to pass to the agent constructor.
+
+        Returns:
+            AgentSet containing the agents created.
+        """
+        new_kwargs = {col: df[col] for col in df.columns if col not in ("model", "n")}
+        new_kwargs.update(kwargs)
+        return cls.create_agents(model, len(df), **new_kwargs)
```
quaquel: Why run in through create_agents? In my view, this defeats the purpose of having a separate factory method: you could go straight to looping over the rows while instantiating agents.

mesa/agent.py
```diff
@@ -135,6 +134,29 @@ def create_agents[T: Agent](
...
+            df: The pandas DataFrame. Each row represents an agent.
+            kwargs: Additional keyword arguments to pass to the agent constructor.
+
+        Returns:
+            AgentSet containing the agents created.
+        """
+        agents = []
+        for record in df.to_dict(orient="records"):
+            # Filtering out "model" and "n" to avoid conflicts with Agent constructor
+            for key in ("model", "n"):
```
quaquel: errors should never be passed over in silence and when in doubt refuse the temptation to guess.

So, if `model` is amongst the columns, you should raise an exception. Not sure which one is appropriate, but I would guess a value error. I am even wondering whether it should be dissallowed, because everything here is passed as keyword arguments, while `model` is an argument.

Also, why is n not allowed?

falloficaruss: I'll update this to raise a `ValueError` if a column named model is found. Since model is already passed positionally to `Agent.__init__` a collision would result in a `TypeError` during instantiation; an explicit error at the start is better than letting it fail deep in the loop or silently dropping it.

The restriction on n was an oversight. It was blocked because it clashing with the n parameter in 
`create_agents`, but if we move to a direct row-loop that avoids that method, n will perfectly be valid as an agent attribute. I'll remove that check

mesa/agent.py
```diff
@@ -135,6 +134,32 @@ def create_agents[T: Agent](
...
+            df: The pandas DataFrame. Each row represents an agent.
+            kwargs: Additional keyword arguments to pass to the agent constructor.
+
+        Returns:
+            AgentSet containing the agents created.
+        """
+        if "model" in df.columns:
+            raise ValueError(
+                "The column name 'model' is reserved for the Mesa Model instance and cannot be used in the DataFrame."
+            )
```
quaquel: at the moment `model` is a positional argument, so this check is technically not needed.

mesa/agent.py
```diff
@@ -135,6 +134,32 @@ def create_agents[T: Agent](
 
         return AgentSet(agents, random=model.random)
 
+    @classmethod
+    def from_dataframe[T: Agent](
+        cls: type[T], model: Model, df: pd.DataFrame, **kwargs
```
quaquel, line 140: Why allow additional kwargs here?

falloficaruss: The `**kwargs` are included to allow users to pass constant values or shared objects to all agents without needing to add them as redundant columns in the `DataFrame` (e.g., `from_dataframe(model, df, status='active')`). This provides parity with the `create_agents` API, where positional/keyword arguments can be mixed with sequences

quaquel: not fully since you don't allow args, only kwargs.

falloficaruss: I'll add support for `*args` so that positional arguments can be mixed with the DataFrame data, mirroring the 
`create_agents` API. I'll implement this while maintaining the direct row-based iteration to keep it efficient.

mesa/agent.py
```diff
@@ -135,6 +134,52 @@ def create_agents[T: Agent](
...
+            if isinstance(v, (list, np.ndarray, tuple, pd.Series)) and len(v) == n:
+                kw_val_iters.append(v)
+            else:
+                kw_val_iters.append(itertools.repeat(v, n))
+
+        pos_iter = zip(*arg_iters) if arg_iters else itertools.repeat(())
+        kw_iter = zip(*kw_val_iters) if kw_val_iters else itertools.repeat(())
+
+        records = df.to_dict(orient="records")
+        agents = []
```
quaquel: looking at this, it starts to overlap again a lot with `create_agents`. 

I am wondering whether we should not simply drop args and kwargs. If people want kwargs, they can just add this to the dataframe prior to calling it. And given the conversation around prefering kwargs in mesa 4, we might simply not allow args at all here.

falloficaruss: I'll simplify the method to accept only the model and `df`, and remove the `*args` and `**kwargs` support. If users need to pass additional data, they can leverage pandas to pre-populate the `DataFrame`

EwoutH: > I am wondering whether we should not simply drop args and kwargs.

I’m wondering this at well. I don’t have a clear view yet, I can see the arguments for both.

I think this first needs to be discussed (in #3186).

tests/test_agent.py
```diff
@@ -225,6 +226,152 @@ def __init__(self, model, attr, def_attr, a=0, b=0):
...
+    for i, agent in enumerate(agents):
+        assert agent.value == i
+        assert agent.list_attr == [i]
+        assert agent.df_value == f"df_{i}"
+        assert agent.tuple_attr == (1, 2)
+
+    # Test from_dataframe with reserved column name 'model' should raise TypeError
+    df_model = pd.DataFrame({"model": ["some_model_string"] * n})
+    with pytest.raises(TypeError, match="got multiple values for argument 'model'"):
+        TestAgent.from_dataframe(model, df_model)
```
quaquel: this can be removed

tests/test_agent.py
```diff
@@ -225,6 +226,152 @@ def __init__(self, model, attr, def_attr, a=0, b=0):
...
+    class TestAgentWithN(Agent):
+        def __init__(self, model, n=None):
+            super().__init__(model)
+            self.n_val = n
+
+    df_n = pd.DataFrame({"n": [999] * n})
+    agents = TestAgentWithN.from_dataframe(model, df_n)
+    for agent in agents:
+        assert agent.n_val == 999
+
```
quaquel: this can be removed

mesa/agent.py
```diff
@@ -135,6 +134,30 @@ def create_agents[T: Agent](
...
+        cls: type[T], model: Model, df: pd.DataFrame
+    ) -> AgentSet[T]:
+        """Create agents from a pandas DataFrame.
+
+        Each row of the DataFrame represents one agent. The DataFrame columns are
+        mapped to the agent's constructor as keyword arguments.
+
+        Note:
+            If you need to pass constant values or sequences not in your original data,
+            add them as columns to the DataFrame before calling this method.
```
quaquel: move this after the args and return fields

tests/test_agent.py
```diff
@@ -225,6 +226,152 @@ def __init__(self, model, attr, def_attr, a=0, b=0):
         assert agent.b == 7
 
 
+def test_agent_create_edge_cases():
```
quaquel: Why are these added? They are not relevant for this PR

EwoutH:
I doo agree with dropping **args but keeping **kwargs

quaquel:
If we allow series as per #3186, allowing kwargs here is redundant.

falloficaruss:
I see the point about redundancy with a Series-capable `create_agents`, but I think keeping `**kwargs` in `from_dataframe` is worth it for Data Integrity and Ergonomics:

- The DataFrame often represents the 'Raw Population' (observed data like age, income), while the extra `**kwargs` represent 'Simulation State' (initial model parameters like `is_infected=False, energy=100`). Forcing users to inject simulation constants into their raw dataframes just to instantiate agents feels like 'polluting' the data layer.
- Adding a column of constant scalars to a large DataFrame (e.g., 1M rows) is an $O(N)$ operation that consumes extra memory. Passing a single scalar via `**kwargs` is $O(1)$ and avoids making pandas allocate memory for a column that would be immediately discarded after iteration.
- Users coming from data-heavy workflows (HuggingFace, Ray, etc.) will look for `.from_dataframe()` specifically because they have a table and want to map it. If that method lacks the ability to add a simple constant override, it feels 'broken' compared to other Pythonic factory methods.

Essentially, `create_agents` is for when you are building a population by columns, and `from_dataframe` is for when you are building a population from a table. Both should be first-class citizens with simple override support.

Also, as stated by @EwoutH in https://github.com/mesa/mesa/issues/3186#issuecomment-3809557834 should I split this into 2 seperate PRs?

EwoutH:
I agree.

While I would prefer 2 PRs, keeping it in one is also fine since the create agent changes are small.

quaquel:
Ok, I am fine with adding kwargs back to from_dataframe, but let's then restrict it to non-sequence data only.

EwoutH:
Agreed.

mesa/agent.py
```diff
@@ -16,11 +16,10 @@
 from collections import defaultdict
 from collections.abc import Callable, Hashable, Iterable, Iterator, MutableSet, Sequence
 from random import Random
-
-# mypy
 from typing import TYPE_CHECKING, Any, Literal, overload
 
 import numpy as np
+import pandas as pd
```
EwoutH, line 22: This will make pandas a hard runtime dependency. Can we do the import in from_dataframe instead?

quaquel: it already is a hard dependency via the toml and the datacollector.py already imports it.  So I fail to see the point

tests/test_agent.py
```diff
@@ -225,6 +226,84 @@ def __init__(self, model, attr, def_attr, a=0, b=0):
...
+    for i, agent in enumerate(agents):
+        assert agent.value == i
+        assert agent.list_attr == [i]
+        assert agent.df_value == f"df_{i}"
+        assert agent.tuple_attr == (1, 2)
+        assert agent.extra_attr == 5
+
+    # Test that passing a sequence in kwargs raises TypeError
+    with pytest.raises(TypeError, match="does not support sequence data in kwargs"):
+        TestAgent.from_dataframe(model, df, list_attr=[1, 2, 3])
```
EwoutH: Test everything
```suggestion
    # Test that passing a sequence in kwargs raises TypeError
    for bad in ([1, 2, 3], (1, 2, 3), np.array([1, 2, 3]), pd.Series([1, 2, 3])):
        with pytest.raises(TypeError, match="does not support sequence data in kwargs"):
            TestAgent.from_dataframe(model, df, list_attr=bad)
```

tests/test_agent.py
```diff
@@ -225,6 +226,84 @@ def __init__(self, model, attr, def_attr, a=0, b=0):
...
+
+
```
EwoutH:
```suggestion
    # kwargs should override DataFrame columns on key collision
    agents = TestAgent.from_dataframe(model, df, value=999)
    assert all(a.value == 999 for a in agents)

    # empty DataFrame should create an empty AgentSet
    empty_df = pd.DataFrame(columns=list(df.columns))
    agents = TestAgent.from_dataframe(model, empty_df, extra_attr=5)
    assert len(agents) == 0

    # DataFrame index should be ignored
    df_with_index = df.copy()
    df_with_index.index = range(100, 100 + n)
    agents = TestAgent.from_dataframe(model, df_with_index, extra_attr=5)
    assert [a.value for a in agents] == list(range(n))
```

EwoutH:
Do we have any idea how this will handle non-Pandas tables? Like NumPy arrays or Polars dataframes?

quaquel:
> Do we have any idea how this will handle non-Pandas tables? Like NumPy arrays or Polars dataframes?

Numpy does not have to_dict, so it will fail. Polars has a to_dict, but with different keyword arguments. Again it will fail.

mesa/agent.py
falloficaruss: I've tried it before but `ruff` doesn't allow it.

falloficaruss:
I can add Duck-Typing for Polars support.

Something like:
```python
# In from_dataframe
if hasattr(df, "to_dict"): # Pandas
    records = df.to_dict(orient="records")
elif hasattr(df, "to_dicts"): # Polars
    records = df.to_dicts()
else:
    raise TypeError("Object must be a Pandas or Polars DataFrame.")
```

But this might require slightly more complex error handling if `to_dict` exists but doesn't support the orient kwarg in some other library.

quaquel:
I suggest we just handle pandas for now and move the polars conversation more broadly to the mesa 4 discussion

EwoutH (approved):
Yeah that’s fine

PR #3200, opened 2026-01-24
AgentSet.__getitem__ Creates Full List

codebyNJ:
### Summary
Every single index access (e.g., agents[0]) creates a complete list of all agents just to return one element. For an AgentSet with 10,000 agents, accessing agents[0] allocates and populates a list of 10,000 items.

### Implementation
Before implmentation:
```python
def __getitem__(self, item):
    return list(self._agents.keys())[item]
```

After implementation:
```python
import itertools

def __getitem__(self, item):
    # Handle negative indices and slices
    if isinstance(item, slice):
        return list(self._agents.keys())[item]
    
    # For single index, use islice to avoid creating full list
    if item < 0:
        # Negative index - need full list
        return list(self._agents.keys())[item]
    
    try:
        return next(itertools.islice(self._agents.keys(), item, item + 1))
    except StopIteration:
        raise IndexError("AgentSet index out of range")
```

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -522,15 +522,12 @@ class AgentSet[A: Agent](MutableSet[A], Sequence[A]):
         Returns:
             Agent | list[Agent]: The selected agent or list of agents based on the index or slice provided.
         """
-        # For slices, we need the full list
         if isinstance(item, slice):
             return list(self._agents.keys())[item]
 
-        # For negative indices, we need the full list
         if item < 0:
             return list(self._agents.keys())[item]
 
-        # For positive indices, use islice to avoid creating full list
         try:
             return next(itertools.islice(self._agents.keys(), item, item + 1))
         except StopIteration:


quaquel:
As indicated in #3185, I am suggesting removing sequence support for agentset in mesa 4.

codebyNJ:
Understood. If #3185 is the planned direction, we can close this PR in favor of that work. Would it make sense to prioritize #3185 for the next release, or should we merge this as a stop-gap improvement?

quaquel:
> Understood. If https://github.com/mesa/mesa/issues/3185 is the planned direction, we can close this PR in favor of that work. Would it make sense to prioritize https://github.com/mesa/mesa/issues/3185 for the next release, or should we merge this as a stop-gap improvement?

Fully addressing #3185 can only be done in Mesa 4. We follow [semver](https://semver.org). Removing sequence is a backward incompatibility change. So it can only be done once we move to Mesa 4. However, what can be done is add `to_list` and raise a `PendingDepracationWarning` on `__getitem__` which points users to using the new `to_list`.

quaquel:
I am closing this PR, but a focussed one on deprecating sequence behavior is still welcome.

codebyNJ:
Thanks @quaquel. I'll open a new PR focused specifically on the deprecation path:
  - Add to_list() method to AgentSet
  - Add PendingDeprecationWarning on __getitem__ directing users to to_list()

PR #3201, opened 2026-01-24
Add EventGenerator class for recurring event patterns

EwoutH:
I was doing too much stuff at the same time. Let's break it down in a few smaller PRs.

This is (in my opinion) a crucial and obvious building block: A `EventGenerator` class that generates events in a certain pattern.

### Summary
Adds an `EventGenerator` class to the discrete event simulation module, enabling recurring event patterns as a first-class concept separate from one-off events.

### Motive
The current event scheduling system only supports single events via `SimulationEvent`. For recurring behaviors (like data collection every N time units, or Poisson arrivals), users must manually reschedule events after each execution. This PR introduces `EventGenerator` to handle recurring patterns cleanly.

### Implementation
- Added `EventGenerator` class to `eventlist.py`
- Supports fixed intervals (`interval=5.0`) or callable intervals (`interval=lambda m: m.random.expovariate(2.0)`)
- `start(at, after)` controls when the pattern begins
- `stop(at, after, count)` controls when the pattern ends
- Tracks execution count and active state
- Uses weak references internally through `SimulationEvent`
- All methods return `self` for chaining: `generator.start(at=0).stop(count=10)`

### Usage Examples
```python
# Fixed interval pattern
gen = EventGenerator(model, callback, interval=5.0)
gen.start(at=0.0).stop(count=100)

# Stochastic arrivals
gen = EventGenerator(model, self.customer_arrival, interval=lambda m: m.random.expovariate(2.0))
gen.start()

# Including arguments and keyword arguments
from functools import partial
EventGenerator(model, partial(fn, "a", k="v"), interval=1.0)
```

### Additional Notes
This is part of a larger effort to improve Mesa's scheduling API (#2921). Future work will integrate `EventGenerator` further, potentially into the `Model` class via `model.create_pattern()` and/or auto-schedule `model.step()` as a pattern.

### DIFF
--- a/mesa/experimental/devs/eventlist.py
+++ b/mesa/experimental/devs/eventlist.py
@@ -131,15 +131,13 @@ class EventGenerator:
 
     EventGenerator represents a pattern for when things should happen repeatedly.
     Unlike a single SimulationEvent, an EventGenerator is persistent and can be
-    paused, resumed, or stopped.
+    stopped or configured with stop conditions.
 
     Attributes:
         model: The model this generator belongs to
         function: The callable to execute for each generated event
         interval: Time between events (fixed value or callable returning value)
         priority: Priority level for generated events
-        function_args: Arguments passed to the function
-        function_kwargs: Keyword arguments passed to the function
 
     """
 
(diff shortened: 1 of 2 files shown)


mesa/experimental/devs/eventlist.py
```diff
@@ -123,6 +126,189 @@ def __lt__(self, other):  # noqa
...
+        function_args: Arguments passed to the function
+        function_kwargs: Keyword arguments passed to the function
+
+    """
+
+    def __init__(
+        self,
+        model: Model,
+        function: Callable,
+        interval: float | int | Callable[[Model], float | int],
```
quaquel, line 148: I don't understand the typing of the callable here. Why would it need a number next to model?

mesa/experimental/devs/eventlist.py
```diff
@@ -123,6 +126,189 @@ def __lt__(self, other):  # noqa
...
+        priority: Priority = Priority.DEFAULT,
+        function_args: list[Any] | None = None,
+        function_kwargs: dict[str, Any] | None = None,
```
quaquel: I understand doing this, although with `functools.partial` we can also ofload this to a python library.

EwoutH: Thanks for the swift review.

Would the API them become like this?
```Python
EventGenerator(model, partial(fn, "a", k="v"), interval=1.0)
```

mesa/experimental/devs/eventlist.py
EwoutH: Sharp eye. This is intended, the Model is passed (and typed) so you can access the random number generator, like for:
```Python
# Poisson arrival
EventGenerator(model, fn, interval=lambda m: m.random.expovariate(2.0))
```

PR #3204, opened 2026-01-24
Unify event list between Model and Simulator

EwoutH:
### Summary
Refactors the internal architecture so that `Model` owns a single event list (`_event_list`) that simulators use via a property. This lays the groundwork for a unified event-based API while maintaining full backward compatibility.

### Motive
We're moving toward an event-based API where both `model.step()` and simulator-driven execution use the same underlying mechanism. Previously, the `Simulator` owned its own `EventList`, creating two separate systems. This PR unifies them so there's one event list with two ways to drive it:
1. `model.step()` → advances 1 time unit via `_advance_time()`
2. `simulator.run_until()/run_for()` → processes events from the same list

### Implementation
- **Model**: Added `_event_list` attribute and `_advance_time(until)` method that processes events up to a given time
- **Simulator**: Changed `event_list` from instance variable to property that returns `model._event_list`
- **ABMSimulator**: Simplified by adding `_schedule_step()` and `_do_step()` methods, removed duplicate `run_until()` override
- **SimulationEvent**: Added `__getstate__`/`__setstate__` for pickling support with weak references
- Cleaned up redundant code and unified the event processing logic

### Usage Examples
No public API changes. Existing code works exactly as before:

```python
# Direct stepping (unchanged)
model = Model()
model.step()  # Internally uses _advance_time()

# Simulator-driven (unchanged)  
simulator = ABMSimulator()
simulator.setup(model)
simulator.run_for(10)  # Uses model._event_list
```

### Additional Notes
- All existing tests pass with minimal modifications (model isn't removed on a reset)
- Added one test to verify simulator uses model's event list
- This is purely internal refactoring with no breaking changes
- Prepares for future public event scheduling API on Model

### DIFF
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -186,18 +186,18 @@ class Model[A: Agent, S: Scenario]:
             until: The time to advance to
 
         """
-        while not self._event_list.is_empty():
+        while True:
             try:
-                event = self._event_list.peek_ahead(1)[0]
+                event = self._event_list.pop_event()
             except IndexError:
                 break
 
-            if event.time > until:
+            if event.time <= until:
+                self.time = event.time
+                event.execute()
+            else:
+                self._event_list.add_event(event)
                 break
-
-            self._event_list.pop_event()
-            self.time = event.time
-            event.execute()
 
         self.time = until
 


mesa/experimental/devs/simulator.py
```diff
@@ -92,19 +98,18 @@ def setup(self, model: Model) -> None:
                 f"Model time ({model.time}) does not match simulator start_time ({self.start_time}). "
                 "Has the model already been run?"
             )
-        if not self.event_list.is_empty():
+        if not model._event_list.is_empty():
```
quaquel, line 101: sincy you have properties for time and eventlist, why not use those?

mesa/experimental/devs/simulator.py
```diff
@@ -92,19 +98,18 @@ def setup(self, model: Model) -> None:
...
             raise ValueError("Events already scheduled. Call setup before scheduling.")
 
         self.model = model
         model._simulator = self  # Register simulator with model
 
     def reset(self):
         """Reset the simulator."""
-        self.event_list.clear()
         if self.model is not None:
+            self.model._event_list.clear()
```
quaquel: this can all be rerouted via the eventlist property I guess

mesa/experimental/devs/simulator.py
```diff
@@ -316,7 +317,18 @@ def setup(self, model):
...
+    def _schedule_step(self, time: int) -> None:
+        """Schedule the model step at the given time."""
+        event = SimulationEvent(time, self._do_step, priority=Priority.HIGH)
+        self.event_list.add_event(event)
+
+    def _do_step(self) -> None:
+        """Execute one step: call user step, increment steps, schedule next."""
+        self.model.steps += 1
+        self.model._user_step()
+        self._schedule_step(int(self.model.time) + 1)
```
quaquel: what is the idea behind this and should this not use the EventGenerator?

mesa/model.py
```diff
@@ -168,17 +174,33 @@ def __init__(
 
     def _wrapped_step(self, *args: Any, **kwargs: Any) -> None:
         """Automatically increments time and steps after calling the user's step method."""
-        # Automatically increment time and step counters
+        self._advance_time(self.time + 1)
+        self._user_step(*args, **kwargs)
+
+    def _advance_time(self, until: float) -> None:
```
quaquel, line 182: not sure I would put this on the model directly. See the earlier conversation on encapsulating the different parts instead of having a blob model class.

mesa/experimental/devs/simulator.py
EwoutH: The model is only assigned 3 lines later. So here it can't.

mesa/experimental/devs/simulator.py
EwoutH: Here it can indeed

mesa/experimental/devs/simulator.py
EwoutH: The idea is that no matter if you progress time with `model.step()` or `simulator.run_for(1)`  you get the exact same behavior. I hope this will make it easier to deprecate old behavior and add public APIs for new ones.

It might have used EventGenerator. For now this was the minimal-change version to get this working.

mesa/model.py
EwoutH: We should have a bigger discussion where we put the public API. This is a private method that bridges the gap between the step() time progression in Mesa 3 and whatever we will have in Mesa 4. It will be removed once processing time with step() is deactivated.

EwoutH:
I accidentally touched the internal event scheduling loop, which of course was highly optimized. Reverted to the old loop, performance restored.

Coverage failure is unrelated.

mesa/model.py
quaquel: > We should have a bigger discussion where we put the public API. 

Yes, I agree on this. I see a few options. One of which is what I think you did in an earlier PR: have a bunch of "passthrough" methods on the model which internally call the appropriate method on the underlying class. Another option would be to use multiple inheritance, although that can become tricky quickly.

EwoutH: I’m indeed exploring a `Scheduling` Mixin. It is quite clean, full IDE and typing support, and can be added to both the Model and Agent.

This is still some essential foundation to be able to properly deprecate old behavior.

quaquel:
Can you check why coverage is only 80%?

EwoutH:
I didn’t add test for the setstate and get state. I already split it of in https://github.com/mesa/mesa/pull/3205, I will address it there.

EwoutH:
While convenient, I don't like the "Update branch from main" applied on my development branches. The merge commits makes undoing a commit, cherry picking and rebasing more difficult.

I can rebase on request :)

EwoutH:
> Can you check why coverage is only 80%?

Now all green, including coverage

mesa/model.py
```diff
@@ -12,12 +12,15 @@
...
+from typing import TYPE_CHECKING, Any
 
 import numpy as np
 
+if TYPE_CHECKING:
+    from mesa.experimental.devs import Simulator
+
 from mesa.agent import Agent, AgentSet
-from mesa.experimental.devs import Simulator
+from mesa.experimental.devs.eventlist import EventList, Priority, SimulationEvent
```
quaquel, line 23: since you are now integrating everything into model, when do you plan to move devs out of experimental?

EwoutH: Good question. I think as soon as we have a stable public API might make sense.

PR #3205, opened 2026-01-24
Fix pickling for scheduled events by serializing weak-referenced callbacks safely

EwoutH:
Small cherry-pick from #3204, to keep that PR atomic.

This PR adds `__getstate__`/`__setstate__` to the event object in `mesa.experimental.devs.eventlist` to make it pickle-friendly when the callback is stored as a weak reference. During pickling we materialize the weak reference into a temporary strong reference (`_fn_strong`) and omit the weakref itself, and during unpickling we restore the callback by recreating the appropriate weak reference (`WeakMethod` for bound methods, `ref` for functions) or leaving it as `None` when absent.

### DIFF
--- a/tests/experimental/test_devs.py
+++ b/tests/experimental/test_devs.py
@@ -273,6 +273,45 @@ def test_simulation_event():
     assert event1 > event2
 
 
+def test_simulation_event_pickle():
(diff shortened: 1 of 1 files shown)


EwoutH:
@quaquel rebased and added tests.

PR #3208, opened 2026-01-25
Deprecate AgentSet sequence behavior and add `to_list()` method

codebyNJ:
### Summary

- Add to_list() method to AgentSet for explicit list conversion
- Add PendingDeprecationWarning to __getitem__ directing users to use to_list()

This PR implements the deprecation path discussed in #3185, preparing users for the removal of sequence behavior in Mesa 4.0.

### Motivation

As discussed in #3185, sequence support (indexing/slicing via __getitem__) on AgentSet will be removed in Mesa 4.0. This PR provides:

1. A migration path for users via the new to_list() method
2. Clear deprecation warnings guiding users to update their code

### Changes
**`mesa/agent.py`**

- Added to_list() method that returns a list of all agents in the AgentSet
- Added PendingDeprecationWarning to __getitem__ with guidance to use to_list()[index]

**`tests/test_agent.py`**

- Added test_agentset_to_list() to test the new method
- Updated test_agentset_get_item() to verify the deprecation warning is raised
- Updated existing tests to use to_list() instead of direct indexing


mesa/agent.py
```diff
@@ -521,7 +537,19 @@ def __getitem__(self, item):
 
         Returns:
             Agent | list[Agent]: The selected agent or list of agents based on the index or slice provided.
+
+        .. deprecated::
+            Sequence behavior (indexing/slicing) is deprecated and will be removed in Mesa 4.0.
+            Use :meth:`to_list` instead: ``agentset.to_list()[index]`` or ``agentset.to_list()[start:stop]``.
         """
+        if not AgentSet._getitem_warning_emitted:
```
quaquel: i dont;t think this is needed. warnings.warn should filter repeated warnings if I am not mistaken (but please test and check).

codebyNJ: You're partially correct! Python's `warnings.warn` with `'default'` filter does suppress repeated warnings, but only per unique call site (due to stacklevel=2). Without the flag, if a user has agentset[0] in multiple locations, they'd see the warning once per location.

  The class-level flag ensures the warning appears only once globally per session, which is less noisy but may hide additional call sites that need updating.

`Options:`
  1. Keep the flag - Shows warning exactly once globally (current implementation)
  2. Remove the flag - Shows warning once per call site (Python's default behavior, helps users find all locations to update)

  Note: `PendingDeprecationWarning` is ignored by default in Python, so most users won't see this warning unless they enable it explicitly (e.g., running with `-W default::PendingDeprecationWarning`).

  Which behavior do you prefer?

quaquel: I suggest removing the if and relying on Python's warning behavior. Let's keep this as simple as possible.

codebyNJ: Sure. I have updated the PR by relying on python's warning behavior.

EwoutH:
Thanks of the PR.

Can you add a small section to our migration guide?

codebyNJ:
> Thanks of the PR.
> 
> Can you add a small section to our migration guide?

Sure I will.

codebyNJ:
@EwoutH would you like to add this section under mesa 3.4.0 version or new mesa 3.5.0? And also could you verify the Content in it.

### AgentSet Sequence Behavior Deprecation
The `Sequence` behavior (indexing and slicing) on `AgentSet` is deprecated and will be removed in Mesa 4.0. Use the new `to_list()` method instead.

```python
# Old (deprecated)
first_agent = model.agents[0]
some_agents = model.agents[1:5]
last_agent = model.agents[-1]

# New
first_agent = model.agents.to_list()[0]
some_agents = model.agents.to_list()[1:5]
last_agent = model.agents.to_list()[-1]

```

For multiple list operations, convert once and reuse:
```python
agent_list = model.agents.to_list()
first = agent_list[0]
last = agent_list[-1]
subset = agent_list[2:8]
```

Ref: [PR #3208](https://github.com/projectmesa/mesa/pull/3208)

EwoutH:
On first glance the content looks good. I can add it, but if you like, you can also open a PR yourself. Our migration guide is available/editable here: https://github.com/mesa/mesa/blob/main/docs/migration_guide.md

codebyNJ:
@EwoutH I have created PR #3218 regarding the migration guide.

Issue #3209, opened 2026-01-25
Tracking Issue: AgentSet Refactoring and _HardKeyAgentSet Implementation using AbstractAgentSet


codebreaker32:
*This is a tracking issue(See #3128 and #3160 for more)*

The current `AgentSet` implementation relies on `WeakKeyDictionary` for storage. While memory-efficient, this approach introduces performance overheads. To solve this fundamentally, we propose refactoring the `AgentSet` hierarchy to use an Abstract Base Class (ABC) pattern(credits to @quaquel) . This ensures a clean separation between the interface and the storage mechanism (Weak vs. Strong).

### Proposed Roadmap

This refactor will be executed in three stages to maintain backward compatibility and easy to review

#### Phase 1: Introduce `AbstractAgentSet`

* Define a new `abc.ABC` named `AbstractAgentSet`.
* Move common logic (iteration, selection, etc) into the abstract base.
* Refactor the existing `AgentSet` to inherit from `AbstractAgentSet`, preserving its current `WeakKeyDictionary` behavior for backward compatibility.

*Merged PR: #3210* 

#### Phase 2: Implement `StrongAgentSet` (Hard Refs)

* Create a new class `StrongAgentSet` (or `HardKeyAgentSet`) inheriting from `AbstractAgentSet`.
* Implement storage using a standard Python `dict` to guarantee deterministic behavior.

*Merged PR: #3219* 

#### Phase 3: Integration & Cleanup

* Update `mesa.Model` to use `StrongAgentSet` as the internal storage for `model._agents_by_type` and `model_all_agents`.
* Remove redundant internal attributes such as `model._agents`.

*Merged PR: #3224*

### Definition of Done

* [x] `AbstractAgentSet` implemented and tested.
* [x] `StrongAgentSet` implemented with full test coverage for liveness/removal scenarios.
* [x] `mesa.Model` updated to use `StrongAgentSet`
* [x] Legacy internal attributes removed from `Model`.

Srinath0916:
Hi @EwoutH, @codebreaker32 , and @quaquel ,

I’ve been following this discussion closely and the proposed roadmap aligns well with some local experimentation I’ve been doing around an AbstractAgentSet + strong/weak-backed implementations.

The ABC-based split cleanly separates interface from storage and seems to address both:

the performance concerns around WeakKeyDictionary, and

the safety/use-case flexibility of keeping a weak-ref option available.

[...]

PR #3210, opened 2026-01-26
Introduce AbstractAgentSet to agent.py and refactor AgentSet to inherit from it

codebreaker32:
### Summary
This PR introduces AbstractAgentSet as an abstract base class for all agent collection implementations in Mesa. This establishes a formal interface and sets the foundation for introducing alternative storage mechanisms (strong vs weak references) while maintaining backward compatibility. Further, It refactors `AgentSet` class to inherit from it

### Motive
For complete discussion: see #3128 and #3160 

### Changes
1. New: AbstractAgentSet class: Defines the minimal interface for agent collections via abstract methods.

2. Refactored: AgentSet class
- Now inherits from AbstractAgentSet
- All existing functionality preserved
- Implements all abstract methods with weakref-optimized versions
- Retains performance optimizations:

Fully backward compatible - No breaking changes

Ticks (1) from #3209

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -232,7 +232,7 @@ class AbstractAgentSet[A: Agent](ABC, MutableSet[A]):
         # Use type(self) to ensure we return the correct subclass (AgentSet vs StrongAgentSet)
         return self._update(agents) if inplace else type(self)(agents, self.random)
 
-        def agg(
+    def agg(
         self, attribute: str, func: Callable | Iterable[Callable]
     ) -> Any | list[Any]:
         """Aggregate an attribute of all agents in the AgentSet using one or more functions.


codebreaker32:
I also have a refactoring suggestion:

In `groupby`:
- Instead of passing a string `result_type="list" or "agentset"`, can we pass a boolean flag `as_list`.
- If `as_list=True`, list will be returned otherwise agentset(or strongagentset)

I have not introduced it here to keep it atomic

mesa/agent.py
```diff
@@ -151,75 +152,44 @@ def scenario(self):
         return self.model.scenario
 
 
-class AgentSet[A: Agent](MutableSet[A], Sequence[A]):
-    """A collection class that represents an ordered set of agents within an agent-based model (ABM).
+class AbstractAgentSet[A: Agent](ABC, MutableSet[A], Sequence[A]):
```
quaquel: You have probably seen #3208. So, I suggest removing Sequence here and in StrongAgentSet in favor of a `to_list` method. WeakAgentSet can than still keep Sequence, while emitting a deprecation warning.

mesa/agent.py
```diff
@@ -151,75 +152,44 @@ def scenario(self):
...
+        ...
 
     def select(
         self,
         filter_func: Callable[[A], bool] | None = None,
         at_most: int | float = float("inf"),
         inplace: bool = False,
         agent_type: type[A] | None = None,
-    ) -> AgentSet[A]:
+    ) -> AbstractAgentSet[A]:
```
quaquel, line 192: Not sure about the typing here. We know that we never return an abstract agent set, so it's a bit strange to say that we do. One option is to define an AgentSet protocol and use that instead. But typing is not my core area of expertise, so thoughts and inputs are welcome.

mesa/agent.py
```diff
@@ -259,7 +229,293 @@ def agent_generator(filter_func, agent_type, at_most):
...
+    def __getitem__(self, item):
+        """Retrieve an agent or a slice of agents from the AgentSet.
+
+        Args:
+            item (int | slice): The index or slice for selecting agents.
+
+        Returns:
+            Agent | list[Agent]: The selected agent or list of agents based on the index or slice provided.
+        """
+        return list(self)[item]
```
quaquel: see my first comment on the deprecation of this.

mesa/agent.py
```diff
@@ -259,7 +229,293 @@ def agent_generator(filter_func, agent_type, at_most):
...
+
+    def groupby(self, by: Callable | str, result_type: str = "agentset") -> GroupBy:
```
quaquel:
```suggestion
    def groupby(self, by: Callable | str, result_type: Literal["agentset", "list"] = "agentset") -> GroupBy:
```

mesa/agent.py
```diff
@@ -259,7 +229,293 @@ def agent_generator(filter_func, agent_type, at_most):
...
+            return GroupBy(
+                {k: AgentSet(v, random=self.random) for k, v in groups.items()}
+            )
+        else:
+            return GroupBy(groups)
+
+    # Performance-critical methods are left abstract
+
+    @abstractmethod
+    def shuffle(self, inplace: bool = False) -> AbstractAgentSet[A]:
```
quaquel, line 428: see my earlier remark on using AbstractAgentSet as a return.

mesa/agent.py
```diff
@@ -259,7 +229,293 @@ def agent_generator(filter_func, agent_type, at_most):
...
+        """Randomly shuffle the order of agents in the AgentSet."""
+        ...
+
+    @abstractmethod
+    def sort(
+        self,
+        key: Callable[[A], Any] | str,
+        ascending: bool = False,
+        inplace: bool = False,
+    ) -> AbstractAgentSet[A]:
```
quaquel, line 438: see my earlier remark on using AbstractAgentSet as a return.

quaquel:
> I also have a refactoring suggestion:
> 
> In `groupby`:
> 
> * Instead of passing a string `result_type="list" or "agentset"`, can we pass a boolean flag `as_list`.
> * If `as_list=True`, list will be returned otherwise agentset(or strongagentset)
> 
> I have not introduced it here to keep it atomic

It would also be a backward-incompatibility change, so we cannot do that in Mesa 3.x because of semver.

mesa/agent.py
codebreaker32: Actually, type hinting the Abstract Base Class is the idiomatic way to handle this (following the LSP).
Also, This is crucial for our roadmap: it allows functions to return either AgentSet or StrongAgentSet indiscriminately, without confusing IDEs or static analysis tools. If we forced a concrete type here, we'd lose that flexibility and I think we need it if we want to explicitly return an AgentSet(weakrefs) during StrongAgentSet overrides. And defining a Protocol would result in huge code duplication so I think this is the right way to go

quaquel: not sure what you mean with LSP, but fair enough otherwise.

codebreaker32: I meant Liskov Substitution Principle

codebyNJ:
few observations @codebreaker32:

- **Typo:** "ABstractAgentSet" in `_update` docstring
- **Question:** `select()` uses `type(self)` while `groupby()` hardcodes AgentSet - is this intentional? (For Phase 2 #3209, should StrongAgentSet return weak refs from all methods, or should the base class use `type(self)` and StrongAgentSet override?)

codebreaker32:
>select() uses type(self) while groupby() hardcodes AgentSet

You are mistaken, both are using `type(self)`

>should StrongAgentSet return weak refs from all methods, or should the base class use type(self) and StrongAgentSet override?)

See The main agenda for StrongAgentSet is performance and determinism, but we know that handling strong references manually creates a high risk of memory leaks for end-users.

The goal is that operations creating a "view" or "copy" of the data (like shuffle(inplace=False) or copy()) should default to returning a standard AgentSet (weak refs). This ensures that while the Model uses strong refs internally for stability, objects handed back to the user are "safe" by default and won't keep dead agents alive.

So yes StrongAgentSet will override them

quaquel:
with the update to agent.py, there is now a conflict. I guess it has to do with deprecating sequence behavior on AgentSet. Can you take a look at the conflict and resolve it? I think this will be ready to be merged afterwards.

codebreaker32:
> Can you take a look at the conflict and resolve it?

Done.

PR #3212, opened 2026-01-26
Making Model reactive

### LINKED ISSUE #3211: Making Model reactive
As part of various ongoing conversations, it has been suggested to make the `Model` reactive. Specifically, the idea is that `model.time` becomes observable and that whenever agents are registered or deregistered, also a signal is emitted. 

I started some pathfinding for this.

1. Making time observable is trivial. Just let Model extend `HasObservables`, and declare `time` as observable is all that is needed.
2. Making agent registration and deregistration is not straightforward. Mesa signals are built around class-level declarations that stuff is observable. But for `agents`, this is currently not easy to do. It is possible to hack around this by just adding "agents" to `self.observables` with 2 possible signals (`AGENT_ADDED` and `AGENT_REMOVED`). However, a cleaner solution would be nice. 
3. We recently shifted to using enums for declaring signal types. But the typing for this is tricky. For example, `self.notify` specifies signal_type to be either a `SignalType` or a `str`, but for the `model.agent` stuff, I declare a new model-specific ModelSignals enum, which breaks typing. 
4. The performance overhead is not bad. Testing locally, I get the numbers below. The init times, I wouldn't worry about much. The main conversation will focus on the wolf-sheep runtime. Here, we see a clear performance loss at runtime. This is due to agents dying and being born, which call `register_agent` and `deregister_agent`, which in turn trigger the emit logic. Some further profiling is probably needed because the overhead is larger than I expected. Remember, in these benchmarks, nothing is listening to these signals.

| BoltzmannWealth | small  | 🔴 +14.5% [+13.0%, +15.9%] | 🔴 +3.6% [+3.1%, +4.2%]    |
| BoltzmannWealth | large  | 🔴 +17.8% [+16.0%, +19.3%] | 🔵 +1.8% [-2.9%, +6.3%]    |
| Schelling       | small  | 🔴 +7.8% [+7.0%, +8.5%]    | 🔵 +0.1% [-0.8%, +1.1%]    |
| Schelling       | large  | 🔴 +7.5% [+4.9%, +9.8%]    | 🔵 +1.2% [-3.8%, +6.1%]    |
| WolfSheep       | small  | 🔴 +21.5% [+20.8%, +22.2%] | 🔴 +4.3% [+3.6%, +5.0%]    |
| WolfSheep       | large  | 🔴 +27.1% [+25.1%, +29.1%] | 🔴 +16.3% [+13.3%, +19.1%] |
| BoidFlockers    | small  | 🔴 +19.0% [+18.0%, +19.9%] | 🔵 +2.4% [+1.6%, +3.2%]    |
| BoidFlockers    | large  | 🔴 +18.5% [+17.7%, +19.2%] | 🔵 +1.8% [+1.0%, +2.7%]    |

quaquel:
This PR makes `model.time` observable, and a signal is emitted when an agent is registered or deregistered.

This draft pr does the following
1. `Model` extends `HasObservables`
2. `Model.time` is declared as being `Observable`
3. Add a new `@emit` decorator that can be used on `HasObservable` methods
4. Add `@emit` to `model.register_agents` and `model.deregister_agents`, which emits a signal from the new `ModelSignals` enum
5. Renames `@computed` to `@computed_property`
6. Refactors signal enums for better typing support. There is now a root SignalType enum that all relevant enums for signals subclass. So, we have `ObservableSignals`, `ListSignals`, and `ModelSignals`. Under conditions, subclassing of enums is allowed, see [the python docs](https://docs.python.org/3/howto/enum.html).
7. It moves the check for subscribers to the start of `HasObservables.notify` to fully remove any redundancy if there are no subscribers to a signal. The only overhead in that case is a single method call and an if check.
8. Extensive internal refactoring for performance and code readability reasons
9. Simplifies the `Message` dataclass so it can be used generically instead of being tied to `ObservableSignals.CHANGE`.

With this pr, the following things have become possible

```python
model.observe("time", ObservableSignals.CHANGED, my_handler)
model.observe("agents", ModelSignals.AGENT_ADDED, my_handler)
model.observe("agents", ModelSignals.AGEND_REMOVED, my_handler)

@emit(name_of_obervable, MyCustomSignals.CALLED, when='before')
def my_method(self):
	...

@emit(name_of_obervable, MyCustomSignals.CALLED, when='after')
def my_method(self):
	...
```

The handler will be called with a `Message` instance. This is a dataclass with `name`, `owner`, `signal_type`, and `additional_kwargs` as fields. `name` is the name of the observable (e.g., time, agents, name_of_observable). `owner` is the `HasObservable` instance emitting the signal. `signal_type` is the type of signal (e.g., `ObservableSignals.CHANGED`). `additional_kwargs` contains the payload and this will differ across signal types. 

The table below lists the fields for all signal types. Note here that anything sent by `@emit` will just include all arguments (`args`) and all keyword arguments of the method call in the payload. 

| signal type                | fields                |
|----------------------------|-----------------------|
| ObservableSignals.CHANGED  | `old`, `new`          |
| ModelSignals.AGENT_ADDED   | `args`                |
| ModelSignals.AGENT_REMOVED | `args`                |
| ListSignals.SET            | `old`, `new`          |
| ListSignals.INSERTED       | `index`, `new`        |
| ListSignals.APPENDED       | `index`, `new`        |
| ListSignals.REMOVED        | `index`, `old`        |
| ListSignals.REPLACED       | `index`, `old`, `new` |

closes #3211


quaquel:
On open issue number 2, I am exploring an idea and would appreciate some feedback. I am thinking of adding an  `@observable(name_of_observable, signal_type, when)` decorator that allows the user to emit a signal before or after calling the method:

```python
@observable("agents", ModelSignals.AGENT_ADDED, when="after")
def register_agent(self, agent):
    ...
```

In this case, an AGENT_ADDED signal will be emitted after register_agent completes. This signal is part of the `model.agents` observable. So you would subscribe to it via `model.observe("agents", ModelSignals.AGENT_ADDED, my_handler)`.

What is nice about this approach is that signals are no longer limited to attributes and attribute changes, but can be used more broadly.

However, there are also various drawbacks and unresolved issues. 

First, `HasObservables`populates `self.observables` by checking all fields in the dicts of all classes in the MRO. So I still need to figure out how to add "agents" and the associated signals to the class. Something like what is used for `@computed` might work here, but I am not sure, since we might want to associate several signals with the same observable (as with `AGENT_ADDED` and `AGENT_REMOVED`). 

Second, the observable computed stuff is tricky as it is. I don't yet know how these kinds of additional signals would be implemented in it. The `ObservableList` is kept out of the computed. So that would be the simplest option. But in that case, we might use a different name for the decorator. Something like `@emit_signal`  instead of `@observable` might make more sense. So, we keep observables and computed confined to simple states and functions of states. While also using signals for wider reactive stuff, but just using `emit_signal` for this.

mesa/model.py
```diff
@@ -103,6 +115,10 @@ def __init__(
         self.time: float = 0.0
         self.agent_id_counter: int = 1
 
+        # hacky fixme
+        for element in ModelSignals:
+            self.observables[("agents", element)] = set()
```
codebreaker32:
```suggestion
        self.observables["agents"] = ModelSignals
```

It will raise ValueError, Maybe you confused it with subscribers.

Edited: Removed casting to set()

quaquel: It should be the available observers, and your fix should work. Although I don't think the casting to set is needed (or should not be needed) if we can fully handle declaring the available signals via subclassing SignalType.

EwoutH:
Thanks for this effort. I will try to dive in later this week(end).

mesa/model.py
codebreaker32: The only reason I added set() is that self.observables is currently type-hinted as dict[str, set[SignalType]] in HasObservables, so passing the raw Enum class would technically violate the type hint (though run fine). In future maybe we could relax that type hint to Iterable[SignalType] to make this cleaner

quaquel: It seems the proper typing should be `observables: dict[str, type[SignalType]]`.  And this shows that the subclassing from SignalType as root enum works and solves a lot of typing headaches.

codebreaker32:
I’ve been exploring the decorator approach and I think we are moving to the right direction but some minor suggestions from my side

>First, HasObservablespopulates self.observables by checking all fields in the dicts of all classes in the MRO. So I still need to figure out how to add "agents" and the associated signals to the class

We don't need manual registration. We can extend `descriptor_generator` to scan the MRO for decorated methods, exactly as it currently scans for `Observable` descriptors and `@computed` properties

If we have the decorator tag the method with an attribute (e.g., `_emits_signal`), the generator can pick it up:

```python
def descriptor_generator(obj):
    seen_observables = {} 
    
    for base in type(obj).__mro__:
        for name, entry in base_dict.items()
            # existing checks

            # Check for @emits decorated methods
            elif callable(entry) and hasattr(entry, "_emits_signal"):
                obs_name, signal_type = entry._emits_signal
                if obs_name not in seen_observables:
                    seen_observables[obs_name] = set()
                # Accumulate signals
                seen_observables[obs_name].add(signal_type)
    
    for observable_name, signal_types in seen_observables.items():
        yield observable_name, signal_types

```

This naturally handles the "multiple signals per observable" case (like `AGENT_ADDED` and `AGENT_REMOVED` both belonging to `"agents"`) by accumulating them in the `seen_observables` dict before yielding. To implement this, we will need to have `set()` structure back instead of Enums

>Something like @emit_signal instead of @observable might make more sense

I agree with separating the naming. `@observable` implies state (like `time`), whereas these methods represent discrete events. `@emits` works well because it functions as a verb, mirroring the distinction between state monitoring and event logging.

> the observable computed stuff is tricky as it is. I don't yet know how these kinds of additional signals would be implemented in it.

I would advise keeping `@emits` and `@computed` completely separate. They represent two different reactive paradigms:

`@computed` is for **State Derivation** (Dataflow: A changes  B updates).
`@emits` is for **Event Notification** (Control flow: Action happens  Signal fires).

Trying to make a computed property depend on an event signal (like `AGENT_ADDED`) gets messy because events don't persist as state. If a user needs to track state changes based on events (e.g., counting agents), they should explicitly `observe` the event and update a counter, rather than trying to derive it inside a computed property.

Example of why separation matters:
```python
class Model(HasObservables):
    time = Observable()
    
    # This makes sense - computed depends on Observable
    @computed
    def agent_count(self):
        return len(self._agents)
    
    # This does NOT make sense - computed depending on events
    @computed
    def something(self):
        # What would it mean to depend on "agent added" event?
        # Events are point-in-time, not continuous state
        pass
```

If you need to react to events, use observation:

```python
def __init__(self):
    super().__init__()
    self.agent_count = 0
    
    # React to events to update state
    self.observe("agents", ModelSignals.AGENT_ADDED,
                lambda s: setattr(self, 'agent_count', self.agent_count + 1))
```

`Observable` + `@computed` = Dataflow programming (reactive state)
`@emits` + `observe()` = Event-driven programming (reactive events)

quaquel:
Thanks a lot for the write-up. It cleanly and clearly summarizes what I have been figuring out this afternoon. I'll update this PR hopefully later today along these lines.

Also, @codebreaker32, I was thinking of changing `@computable` to `@computable_property` to highlight that it is still a property, implying attribute-style access.

codebreaker32:
This is somewhat the rough sketch I was working on, Might help.
```python
class emits:
    def __init__(
        self,
        observable_name: str,
        signal_type: SignalType,
        when: Literal["before", "after"] = "after",
        include_result: bool = False,
    ):
        self.observable_name = observable_name
        self.signal_type = signal_type
        self.when = when
        self.include_result = include_result

    def __call__(self, func: Callable) -> Callable:
        @functools.wraps(func)
        def wrapper(instance: HasObservables, *args, **kwargs):
            # If no one is listening, skip ALL signal logic.
            if not instance._has_subscribers(self.observable_name, self.signal_type):
                return func(instance, *args, **kwargs)

            # ASSUMPTION: First arg is the payload (e.g. agent) It works for now, might optimise later
            # We capture this for signal.new convenience.
            primary_arg = args[0] if args else None
            
            # Emit "before" if requested
            if self.when in ("before"):
                instance.notify(
                    self.observable_name,
                    old_value=None,
                    new_value=primary_arg,
                    signal_type=self.signal_type,
                )
            
            # Execute actual method
            result = func(instance, *args, **kwargs)
            
            # Emit "after" if requested
            if self.when in ("after"):
                signal_kwargs = {}
                if self.include_result:
                    signal_kwargs["result"] = result
                    
                instance.notify(
                    self.observable_name,
                    old_value=None,
                    new_value=primary_arg,
                    signal_type=self.signal_type,
                    **signal_kwargs,
                )
            
            return result
        
        # Tag for auto-discovery by descriptor_generator
        wrapper._emits_signal = (self.observable_name, self.signal_type)
        return wrapper
```



>I was thinking of changing @computable to @computable_property to highlight that it is still a property, implying attribute-style access.

You mean `computed`? Yes its a great name, pythonic and unambiguous

codebreaker32:
I initially added a "both" literal also in the `when` clause, which is why I was checking with `in`

codebreaker32:
> Thanks a lot for the write-up

Glad it helped :)

quaquel:
Ok, I pushed an update that builds on all the input. Again, thanks a lot. 

1. I added an `@emit` decorator that can be used to annotate functions or methods so they emit a signal before or after having been called.
2. I renamed `@computed` to `@computed_property`
3. I updated the inner `descriptor_generator` to handle the `@emit` decorator. 
4. I updated `model.Model` to use `@emit` on `Model.register_agent` and `Model.deregister_agent`. 

There is still some work left to be done:
1. ~We have to decide on where to put the message payload for `@emit`. `Observable` and `@computed_property` are simple: they change an old value to a new value. `@emit`, and `ObservableList`, have a richer payload. However, the current  `Message` dataclass is tailored to the first, and has an `additional_kwargs` field for richer payloads. Now that we are expanding signals. Do we want to revisit the `Message` class design? One option would be to have two message types, but I don't like that because it makes writing the generic handlers more difficult. Another option would be to simply drop the old and new fields, and, for `Observable` and `@computed_property`, put this info in the additional_kwargs dict.~ 
2. There is some more cleanup and refactoring I would like to do. The distinction between BaseObservable and Observable serves little purpose with the introduction of `@computed_property`.
3. ~`@emit` specifies the name of the observable and the signal that will be emitted when the decorated method is called.  This makes the `descriptor_generator` quite straightforward. For `Observable` and `@computed_property`, however, we hard-code the signal that will be emitted inside the decorator. I suggest moving the signals that will be emitted into the property itself (i.e., `Observable` or `ComputedProperty`), so they are properly encapsulated where they belong.~ this is now addressed

codebreaker32:
>Another option would be to simply drop the old and new fields, and, for Observable and @computed_property, put this info in the additional_kwargs dict

I agree, we might do something like this to maintain backward compatibility
```python
class Message:
    name: str
    owner: Any
    signal_type: SignalType | ListSignals
    payload: dict[str, Any]
    
    @property
    def old(self) -> Any:
        return self.payload.get('old')
    
    @property
    def new(self) -> Any:
        return self.payload.get('new')
```

quaquel:
> I agree, we might do something like this to maintain backward compatibility

Thanks. Since this is experimental and not yet used in Mesa internals or our examples, I am not overly worried about backward compatibility. Still, your suggestion offers a simple way to achieve this.

codebreaker32:
>Since this is experimental and not yet used in Mesa internals

Fair enough, I was using it in `baseCollectorListener.py` that's why I said so

quaquel:
Another update: I have done some more refactoring, various minor performance improvements, and removed `old` and `new` from Message. I am still considering getting rid of BaseObservable. This will make `@computed_property` only interact with `Observables`, while `@emit` and `SignallingList` will both also emit signals, but this won't affect the derived state of an `@computed_property`. If you want to depend on a `SignallingLists` (or any future collection that emits signals on changes), you can just subscribe directly to those signals. Likewise with `@emit`. 

This pathfinding has gotten quite large. I am not sure whether I can break it down into a list of clean, separate, more atomic PRs.

quaquel:
> Instead of monkey-patching, you could implement proper agent observables like:

This is what I did early on, but the problem is that `HasObservables` holds a dict, `cls.observables` that contains all observables, and per observable, the signals it emits. It is this part that I was monkey-patching, but if you read subsequent posts, I have now solved it much more generally by adding a new `@emit` decorator.

quaquel:
This PR is ready for review. The pre-commit check fails because `BaseObservable` is declared as ABC, but it has no abstract methods. See my previous comment on whether to keep BaseObservable or not. @codebreaker32, your thoughts on this would be welcome. 

I want to do a separate PR further down the line with more extensive documentation on signals, messages, and payloads. But I first want to flesh out the details as needed for the emerging new experimental approach to data collection. It is plausible that some more updates on signals will follow.  

Also, even though this PR makes Model emit signals, I want to keep signals experimental for the time being. Look at this PR with that in mind.

codebreaker32:
> whether to keep BaseObservable or not

I agree that we should get rid of `BaseObservable`. I propose splitting the hierarchy into two distinct paths clearing the separation between Events and State

### ->The Issue

Currently, `BaseObservable` conflates two behaviors:

1. **Signal Emission:** Notifying listeners that something happened.
2. **Dependency Tracking:** registering with `CURRENT_COMPUTED` inside `__get__` to trigger re-calculations.

`ObservableList` inherits the dependency tracking behavior from `BaseObservable`. This creates ambiguity: if a list is accessed inside a `@computed_property`, should the property recompute on every list mutation (append/remove)? Usually, the answer is no (or it requires explicit handling), but the current inheritance forces the dependency.

### ->The Solution: Split Hierarchy

We should remove `BaseObservable` and use two separate base classes.

1. **`Observable` (Standalone)**: Retains the current `__get__` logic with dependency tracking. This is for values where access implies a dependency.
2. **`SignalEmitter` (New Base)**: A clean base for descriptors that emit signals but do not automatically register as dependencies in `@computed_property`.

We also need to update `descriptor_generator` to recognize both classes.

```python
class SignalEmitter:
    """Base for descriptors that emit signals but don't track dependencies."""
    signal_types = SignalType

    def __get__(self, instance, owner):
        # ...
        # Just return the value, NO dependency tracking
        return getattr(instance, self.private_name, self.fallback_value)

class Observable:
    """Descriptor that tracks dependencies."""
    signal_types = ObservableSignals

    def __get__(self, instance, owner):
        # Existing Logic
```

```python
# Inherits from SignalEmitter, NOT Observable/BaseObservable
class ObservableList(SignalEmitter): 
    signal_types = ListSignals
    # .....

```

> This will make @computed_property only interact with Observables, while @emit and SignallingList will both also emit signals, but this won't affect the derived state.

This architecture enforces that separation at the class level. It prevents `ObservableList` from accidentally triggering expensive computed property re-runs just because the list was accessed. If a user wants a computed property to depend on a list event, they should subscribe to the specific signal (e.g., `ListSignals.APPEND`) explicitly.

So I'd suggest to Remove BaseObservable, split into Observable (with dependency tracking) and SignalEmitter (signals only). This enforces the design goal that `@computed_property` only interacts with Observables, while collections and `@emit` remain reactive but require explicit subscription

quaquel:
@codebreaker32, thanks for the detailed response. It seems we are broadly thinking in the same direction, even down to the naming of the classes. I already had a `SignalEmitter` as root on a local branch, exactly as you suggested. 

>  If a user wants a computed property to depend on a list event, they should subscribe to the specific signal (e.g., ListSignals.APPEND) explicitly.

This is one of the reasons I am not yet fully decided: this won't work. Currently, on every loop of `@computed` it disposes all of its parents. If one of those parents is a signal emitter, but not an observable, this parent won't automatically be added back in via `BaseObservable.__get__` -> `CURRENT_COMPUTED._add_parent`. This points to a wider architectural issue that I have to make up my mind about. Namely, do we need to fully dispose of and rebuild the parents on each loop? It is clear that it adds significant complexity and overhead, though I am not entirely sure about the exact benefits. 

At the same time, having autodiscovery as it is is great because, as a user, you don't have to manually specify the dependency graph. It's automatically created when accessing observables.

codebreaker32:
Can we add a `.track()` method directly to `SignalingList`?

```python
class SignalingList(MutableSequence):
    # .....

    def track(self):
        if CURRENT_COMPUTED is not None:
             # We use the existing owner/name stored on the list
             CURRENT_COMPUTED._add_parent(self.owner, self.name, self)
        return self

```

1. `SignalingList` already knows its `owner` and `name`.
2. `.track()` checks `CURRENT_COMPUTED` and calls `_add_parent()`, just like `Observable.__get__` does.
3. Because `.track()` runs inside the function, it re-registers the dependency on every recompute.




```python
class Model(HasObservables):
    wealth = Observable()      # Auto-tracked 
    agents = ObservableList()  # Not auto-tracked by default
    
    @computed_property
    def total_wealth(self):
        # Observable: automatically tracked 
        base = self.wealth
        
        # SignalEmitter: explicitly tracked via .track()
        return base + sum(a.wealth for a in self.agents.track())

```
<br>


Am I missing something?

quaquel:
I am thinking in roughly the same direction, but we have a few critical differences. First, dependencies should be declared by a `@computed_property`, not in what is being tracked. Second, it should work for any non-Observable that emits, including the `@emit` decorator.

From an API point of view, there are two obvious solutions
1. Dependency injection at the decorator level:
```python
@computed_property(dependencies=("agents", ALL))
def gini(self):
    ...
```

Here, gini is dependent on agents, and gini will subscribe to all signals from the "agents" emitter. What is nice about this API is that everything is declared within the property closure and needs to be handled only once prior to the first access to gini (so agents is resolved relative to self inside gini), not on every call. 

2. Dependency injection inside the annotated function

```python

@computed_property
def gini(self):
    track(self.agents, ALL)
    ...
```
Syntax is simpler, but it will run on each call to gini. 

A related choice is how to handle the update of @computed_property. Currently is removes all parents every time and will rebuild the dependency graph from scratch. This is simple but not efficient and wont work well with either option. My current idea is to diff the dependency graph on recompute and only change subscriptions that are in the diff.

codebreaker32:
I broadly agree with what you said and I think neither option is optimal but if I have to choose one, I'll align to Option 1.

quaquel:
I have decided to hold off on a choice right now. I want to get this merged so we can move on with the collector listener and the dataset. I added it to the list of issues in #3227.

EwoutH:
Thanks for working on this. I haven't reviewed at implementation level, but a few design and broader questions:
1. Model now extends `HasObservables`.
   a) advantages/disadvantages and are there alternatives
   b) is the plan to make Agent also extend `HasObservables`? Do other things (spaces) need this also?
2. `@emit()` decorator API looks convenient. How would it compare to just putting a `notify()` in the method? Bit zoomed out, maybe at a broader level we need a vision on the role of decorators in Mesa.
3. I think ObservableSignals, ListSignals and ModelSignals make sense. AgentSignals (and space etc.) needed in the future?

Maybe at some point we could use a `MesaObject` class, that has functionality like `HasObservables` and other shared functionality in it. Than we avoid multiple inherentence at multiple places.

If there are specific questions or things you would like me to look into, please let me know.

quaquel:
Thanks for working on this. I haven't reviewed at implementation level, but a few design and broader questions:

>Model now extends HasObservables.
> a) advantages/disadvantages and are there alternatives
> b) is the plan to make Agent also extend HasObservables? Do other things (spaces) need this also?

It relies on multiple inheritance. The only alternative is to use composition instead of inheritance but then you get the passthrough problem we also have with the time API. Moreover, the autodetection of observables is quite bit more complicated in that case. Since HasObservables is mixin-like (it has its own well encapsulated attributes) it is trival to add it to agent if needed. I don't see a use case for it yet in the core of mesa itself. 

> @emit() decorator API looks convenient. How would it compare to just putting a notify() in the method? Bit zoomed out, maybe at a broader level we need a vision on the role of decorators in Mesa.

The only difference is that you now can add it easily and also handle the registration of the emitter in a single location. Remember, HasObservables has a dict with all things that can emit and the signals that the thing can emit. If you were to just put a notify inside a method you would still need to seperately add this notify and associated signal type to the observables dict. That is, you need to make changes in at least two disconnected places in your code. With the decorator, this is all conveniently done in a single place.

> I think ObservableSignals, ListSignals and ModelSignals make sense. AgentSignals (and space etc.) needed in the future?

I don't know, but with the SignalType enum base class this becomes much easier to do if desired.

> Maybe at some point we could use a MesaObject class, that has functionality like HasObservables and other shared functionality in it. Than we avoid multiple inherentence at multiple places.

I agree.

EwoutH:
Thanks, seems like a well-considered design.

> If you were to just put a notify inside a method you would still need to separately add this notify and associated signal type to the observables dict.

If I understand it correctly, hat's a very valid reason to use a decorator I think.

I have a few more API design questions:

1. `ListSignals` uses verbs (`INSERT`, `APPEND`) while `ModelSignals` uses past tense (`AGENT_ADDED`). Should there be a consistent convention?
2. The `@emit` decorator emits after the *entire* method completes. What if users need to emit mid-method, or emit conditionally based on what happened? Is manual `notify()` still the recommended approach for complex cases?
3. Can subscribers filter by agent type for `AGENT_ADDED`/`AGENT_REMOVED`? Or would they need to filter in their handler? This could matter for models with many agent types.
4. Handlers that previously accessed `signal.old` and `signal.new` directly now need `signal.additional_kwargs["old"]` (if I understand correctly). I think we should consider whether `Message` should have optional `old`/`new` properties that pull from `additional_kwargs` for backward compatibility. If we don't want to support this behavior in the long term, we can add a deprecation warning to it.

Further:
5. a simple test showing `Model.time` and agent registration signals working might be useful
6. a brief note in the PR description about migration from old API would be appreciated
7. and maybe add a small example in the experimental docs or PR description

mesa/examples/basic/boltzmann_wealth_model/model.py
```diff
@@ -72,3 +72,9 @@ def compute_gini(self):
         # Calculate using the standard formula for Gini coefficient
         b = sum(xi * (n - i) for i, xi in enumerate(x)) / (n * sum(x))
         return 1 + (1 / n) - 2 * b
+
+
+if __name__ == "__main__":
+    model = BoltzmannWealth(1000)
+    for _ in range(100):
+        model.step()
```
EwoutH: Are these changes only for testing or intended to stay here?

mesa/model.py
```diff
@@ -31,8 +38,15 @@
 _mesa_logger = create_module_logger()
 
 
+class ModelSignals(SignalType):
+    """Signal types for model-level events."""
+
+    AGENT_ADDED = "agent_added"
+    AGENT_REMOVED = "agent_removed"
```
EwoutH: Can this class stay in the Experimental space?

mesa/model.py
```diff
@@ -236,6 +255,7 @@ def agents_by_type(self) -> dict[type[A], AgentSet[A]]:
         """A dictionary where the keys are agent types and the values are the corresponding AgentSets."""
         return self._agents_by_type
 
+    @emit("agents", ModelSignals.AGENT_ADDED)
```
EwoutH, line 288: Would it be possible (and useful) to also emit some metadata about the agent added? Like the agent Class or unique_id. Or maybe even the whole instance?

mesa/examples/basic/boltzmann_wealth_model/model.py
quaquel: just for testing

mesa/model.py
quaquel: yes that is possible

mesa/model.py
quaquel: The message that is emitted contains the agent instance that is added, so the requested data is all there in the message.

quaquel:
> ListSignals uses verbs (INSERT, APPEND) while ModelSignals uses past tense (AGENT_ADDED). Should there be a consistent convention?

Good question, I don't have a strong opinion about this. Note also that Observables emit CHANGED. We could change the list signals to be past tense for consistency. 

> The @emit decorator emits after the entire method completes. What if users need to emit mid-method, or emit conditionally based on what happened? Is manual notify() still the recommended approach for complex cases?

First, emit allows controllign whether to emit before or after `func`. I think one is stretching the reactive formalism if one is doing any more complicated stuff. In such cases it might be easier to add am Observable flag and let depedends subscribe to the flag itself instead. 

> Can subscribers filter by agent type for AGENT_ADDED/AGENT_REMOVED? Or would they need to filter in their handler? This could matter for models with many agent types.

At the moment everything is tied to AGENT_ADDED/AGENT_REMOVED, so you would have to filter in the handler, which is relatively cheap to do. In #3227, I list ideas for future extensions to mesa_signals that might allow more fine grained approaches at the Agent class level.


> Handlers that previously accessed signal.old and signal.new directly now need signal.additional_kwargs["old"] (if I understand correctly). I think we should consider whether Message should have optional old/new properties that pull from additional_kwargs for backward compatibility. If we don't want to support this behavior in the long term, we can add a deprecation warning to it.

Mesa_signals is experimental and was not used in any of our own examples. I doubt it is a feature users are using extensively in real models at the moment, so I don't see the need for this.

mesa/model.py
EwoutH: Perfect!

EwoutH:
> We could change the list signals to be past tense for consistency.

Sounds good.

> I think one is stretching the reactive formalism if one is doing any more complicated stuff. 

Fair, if it turns out to be a real need we can always deal with it later.

> I list ideas for future extensions to mesa_signals that might allow more fine grained approaches at the Agent class level.

Cool, let's discuss after this PR.

> Handlers that previously accessed signal.old and signal.new directly now need signal.additional_kwargs["old"]

Aside from the backwards compatibility, `signal.old` (and `signal.new`) is of course a way more elegant API than `signal.additional_kwargs["old"]`. Are those important, if so would properties be nice, and are there any other kwargs that would benefit from properties?

quaquel:
> Aside from the backwards compatibility, signal.old (and signal.new) is of course a way more elegant API than signal.additional_kwargs["old"]. Are those important, if so would properties be nice, and are there any other kwargs that would benefit from properties?

The problem is that old and new are only meaningful for `ObservableSignals.CHANGED`. For other signals, they don't make sense. Hence, I decided, in conversation with @codebreaker32, that it makes more sense to have a single generic kwargs dict that contains the message's payload. This is valid across all messages. 

The other possible design is to build an entire hierarchy of messages with their own unique fields. But this becomes very unwieldy because now in notify you have to figure out which message type to send and you need a complex registration system for user defined signals and message types.

EwoutH:
That gives some context, thanks. I'm good with it.

quaquel:
Small update in light of the feedback

1. All signal types now use a past tense
2. I moved all signal types into their module
3. I removed the testing code from the example

EwoutH:
I think these aren't yet addressed, right?

> 5. a simple test showing `Model.time` and agent registration signals working might be useful
> 6. a brief note in the PR description about migration from old API would be appreciated
> 7. and maybe add a small example in the experimental docs or PR description

quaquel:
> a simple test showing Model.time and agent registration signals working might be useful
> a brief note in the PR description about migration from old API would be appreciated
> and maybe add a small example in the experimental docs or PR description

`@emit`  and `Observable` are both fully tested, so it feels redundant to add tests for that. But if you insist I can do it. 

The api changes are already noted, so not sure what more would be needed. I want to open a separate issue on adding an overview of signals and message payloads to the docs somewhere. 

What do you mean with an example? How to use the new decorator?

EwoutH:
With experimental features we often don't have tutorials or example models, so the PR and its description is the only part where you can see more than the source code. We want power users to use and test our experimental features, and we often link to the PR in the release notes/highlights. Therefore I always like our PR description to be up to spec, like following the [PR feature template](https://github.com/mesa/mesa/blob/main/.github/PULL_REQUEST_TEMPLATE/feature.md) when applicable. It helps users to get started and us to remind why and how we did things.

The way the PR templates are currently structured, they have dual-use for developers and early adopters (as a quick-start guide).

Practically, LLMs are great in helping with this. I often write the motivation myself and then throw in the diff (just add `.diff` to the URL, like `https://github.com/mesa/mesa/pull/3212.diff`), the relevant [PR template](https://github.com/mesa/mesa/tree/main/.github/PULL_REQUEST_TEMPLATE), and other relevant parts of discussion, original PR description, etc. In Claude I have a "Mesa development" project in which the PR feature template is added by default, alongside most important source code (model.py, agent.py, etc.). It gives 90% of the results (which I edit if needed, most often only lightly) for not even a minute of work.

If you don't agree on the usefulness of PR descriptions (for experimental features) or the PR templates itself, let's discuss that separately.

quaquel:
I agree on the PR stuff, but this is part of a series of PRs that, once they come together, must be illustrated by examples and in the PR descriptions.

EwoutH:
That's a fair point. I would approach it this way: Write up the PR description of what this PR does now, and if it (significantly) changes in the feature, just add a line on top:

_This PR is outdated, see #XXXX for the current approach_

EwoutH:
So we can now subscribe to `model.time`. What if I only want to do something if:
- `model.time` increases by at least 1 since I last checked?
- `model.time` passes the next integer?

quaquel:
You have to handle that inside the handler. As it should be in my view.

EwoutH:
Could you give a minimal example?

quaquel:
You can look at th #3145, where in the handler it is checked. but a minimum example would be:

```python

def my_handler(message:Message):
    delta = message.kwargs["new"] - message.kwargs["old"]
    if delta < 1:
        return
    # my handler code

# if you want memory for your last action
class MyHandler:
    def __init__(self):
        self.last_action = 0
    
    def __call__(message:Message):
        new = message.kwargs["new"]
        delta = new - self._last_action
        if delta < 1:
            return
        self.last_action = new
        # my action

```

EwoutH:
Looks complex.

But maybe I'm overthinking this. If you want to do sometime every 1 timestep anyways, just schedule an recurring event.

quaquel:
> But maybe I'm overthinking this. If you want to do something every 1 timestep anyway, just schedule a recurring event.

It really depends on what you want to achieve. I would, in general, use event scheduling for simulation stuff, while using signals for looser coupling to e.g., a UI or a data recorder.

EwoutH:
I think time is important enough to warrant some default handlers. Especially if they can take a Schedule.

quaquel:
> I think time is important enough to warrant some default handlers. Especially if they can take a Schedule.

It might be quite easy to have a base handler class that takes a schedule, or even instantiate a class with a schedule and a callable.

quaquel:
> I think time is important enough to warrant some default handlers. Especially if they can take a Schedule.

Thinking about this a bit more. This conversation focussed mostly on time being observable. In that context, there might be a use case for a default handler. However, I don't think that this generalizes to observables/emitters as such.

PR #3213, opened 2026-01-26
support for custom data input types

usuallycwdillon:
### Summary
Added support for custom parameter types in Mesa's visualization system through a parameter extractor registry. This allows users to define complex parameter structures (nested dicts, matrices, file loaders, etc.) beyond the built-in UI types.

### Motive
We found a need to keep some user inputs grouped in dict-like structures to make input a more intuitive (for example 'alpha' and 'beta' parameters to configure a beta distribution), to allow a checkbox (allow personnel turnover) to make subsequent inputs visible (how many, how often, replacement time, ...), and to create range sliders to define min/max input values. In our implementation, we made these dict input types collapsible so that we could minimize the  vertical space the user inputs take up. Mesa's visualization previously only supported simple parameter types (sliders, checkboxes, text inputs). We also wanted users to be able to upload data files or configuration files when the number of user parameters in our model became too cumbersome for drop-downs and sliders. Even when we over-wrote the 'UserInputs' we realilzed that Mesa still needed to know how to process the data when the 'Reset' button reloaded the current user params into the model. Setting up a custom data extractor registry in 'solara_viz.py' seemed the least invasive way to make our customizations generalizable for any use case. 

### Implementation
- Added `register_param_extractor()` function to register custom parameter type handlers
- Modified `extract_default_value()` to check registered extractors before falling back to standard "value" key extraction
- Updated `UserInputs` to skip rendering parameters with custom extractors that don't have standard UI components
- Added tests in `tests/visualization/test_solara_viz.py`  and made certain we didn't break anything that was already working

### Usage Examples
Added an example demonstrating the feature:
1.  **`mesa/examples/basic/boltzmann_wealth_model/app_with_dict_params.py`**: Modified Boltzmann Wealth Model using a Dict parameter to group related configuration options (**note** this also includes a `custom_params.py` file that tells Solara how to render the inputs.)

Example usage:
```python
from mesa.visualization.solara_viz import register_param_extractor

def extract_dict_param(spec):
    entries = spec.get("entries", {})
    result = {}
    for key, value in entries.items():
        if isinstance(value, dict) and "value" in value:
            result[key] = value["value"]
        elif isinstance(value, dict):
            result[key] = extract_dict_param({"entries": value})
        else:
            result[key] = value
    return result

register_param_extractor("Dict", extract_dict_param)

model_params = {
    "config": {
        "type": "Dict",
        "entries": {
            "width": {"value": 10},
            "height": {"value": 10}
        }
    }
}


<img width="564" height="537" alt="Dict param - expanded" src="https://github.com/user-attachments/assets/f88b20e0-b1ff-4415-bc78-3a6f21b0afde" />

<img width="544" height="340" alt="Dict param - collapsed" src="https://github.com/user-attachments/assets/79dd185f-506e-4545-bdeb-e96eec7c756b" />

### DIFF
--- a/docs/apis/visualization.md
+++ b/docs/apis/visualization.md
@@ -102,9 +102,9 @@ model_params = {
 }
 
 # The model will receive config as:
-# {"learning_rate": 0.01, "batch_size": 32, 
+# {"learning_rate": 0.01, "batch_size": 32,
 #  "optimizer": {"name": "adam", "momentum": 0.9}}
-``` 
+```
 
 ### How It Works
 
--- a/docs/apis/visualization.md
+++ b/docs/apis/visualization.md
@@ -59,6 +59,79 @@ For detailed tutorials, please refer to:
    :show-inheritance:
 ```
 
+## Custom Parameter Types
+
+Mesa's visualization system supports custom parameter types through an extractor registry. This allows you to define complex parameter structures beyond the built-in types (Slider, Checkbox, etc.).
+
+### Basic Usage
+
+Register a custom parameter extractor before creating your `SolaraViz` instance:
+
+```python
+from mesa.visualization.solara_viz import register_param_extractor
+
+def extract_dict_param(spec):
+    """Extract default values from a nested dictionary parameter."""
+    entries = spec.get("entries", {})
+    result = {}
+    for key, value in entries.items():
+        if isinstance(value, dict) and "value" in value:
+            result[key] = value["value"]
+        elif isinstance(value, dict):
+            result[key] = extract_dict_param({"entries": value})
+        else:
+            result[key] = value
+    return result
+
+# Register the extractor for "Dict" type parameters
+register_param_extractor("Dict", extract_dict_param)
+
+# Now you can use Dict type in model_params
+model_params = {
+    "config": {
+        "type": "Dict",
+        "entries": {
+            "learning_rate": {"value": 0.01},
+            "batch_size": {"value": 32},
+            "optimizer": {
+                "name": {"value": "adam"},
+                "momentum": {"value": 0.9}
+            }
+        }
+    }
+}
+
+# The model will receive config as:
+# {"learning_rate": 0.01, "batch_size": 32, 
+#  "optimizer": {"name": "adam", "momentum": 0.9}}
+``` 
+
+### How It Works
+
+When SolaraViz initializes model parameters:
+1. It checks if a parameter has a registered extractor for its "type"
+1. If found, it calls the extractor function with the parameter specification
+1. If not found, it falls back to extracting the "value" key (default behavior)
+1. This allows you to create arbitrarily complex parameter types while keeping Mesa's core simple and extensible.
+
+For a complete working example, see `mesa/examples/basic/boltzmann_wealth_model/app_with_dict_params.py` and note the way we overwrite Mesa's UserInputs with our own CustomUserInputs:
+
+```python
+# Import custom parameter support
+from custom_params import CustomUserInputs
+import mesa.visualization.solara_viz as solara_viz
+
+# Override UserInputs with CustomUserInputs
+solara_viz.UserInputs = CustomUserInputs
+```
+
+Common Use Cases
+* Nested configurations: Complex model settings organized hierarchically
+* Matrix/array parameters: Initial states for grid-based models
+* Graph structures: Network topologies as parameters
+* Custom data structures: Any domain-specific parameter format
+* Data file uploads (GeoJSON, for example) in a way that keeps the parsing logic separate from your model code
+
 
 ## Matplotlib-based visualizations
 
--- a/mesa/examples/basic/boltzmann_wealth_model/app_with_dict_params.py
+++ b/mesa/examples/basic/boltzmann_wealth_model/app_with_dict_params.py
@@ -1,0 +1,115 @@
+import altair as alt
+
+# Import custom parameter support
+from custom_params import CustomUserInputs
+
+import mesa.visualization.solara_viz as solara_viz
+from mesa.examples.basic.boltzmann_wealth_model.model import BoltzmannWealth
+from mesa.mesa_logging import INFO, log_to_stderr
+from mesa.visualization import (
+    SolaraViz,
+    SpaceRenderer,
+    make_plot_component,
+)
+from mesa.visualization.components import AgentPortrayalStyle
+
+# Override UserInputs with CustomUserInputs
+solara_viz.UserInputs = CustomUserInputs
+
+log_to_stderr(INFO)
+
+
+def agent_portrayal(agent):
+    return AgentPortrayalStyle(color=agent.wealth)
+
+
+# Modified model_params to use Dict type for grid dimensions
+model_params = {
+    "rng": {
+        "type": "InputText",
+        "value": 42,
+        "label": "Random Seed",
+    },
+    "n": {
+        "type": "SliderInt",
+        "value": 50,
+        "label": "Number of agents:",
+        "min": 10,
+        "max": 100,
+        "step": 1,
+    },
+    "grid_dimensions": {
+        "type": "Dict",
+        "label": "Grid Dimensions",
+        "entries": {
+            "width": {
+                "value": 10,
+                "label": "Width",
+                "type": "SliderInt",
+                "min": 5,
+                "max": 50,
+                "step": 1,
+            },
+            "height": {
+                "value": 10,
+                "label": "Height",
+                "type": "SliderInt",
+                "min": 5,
+                "max": 50,
+                "step": 1,
+            },
+        },
+    },
+}
+
+
+# Wrapper model to accept grid_dimensions dict
+class BoltzmannWealthWithDictParams(BoltzmannWealth):
+    def __init__(self, n=100, grid_dimensions=None, rng=None):
+        if grid_dimensions is None:
+            grid_dimensions = {"width": 10, "height": 10}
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by EwoutH** [comment]
Thanks for this PR! I appreciate the thought and effort that went into it, but I think we're solving this problem in an overly specific way when Python already gives us what we need.

Looking at your use case, the real problem is wanting custom UI rendering (collapsible sections, grouped inputs, etc.), not parameter extraction. But even with this PR's extractor registry, you still had to override `UserInputs` anyway - so the registry doesn't actually eliminate any work.

We're a Python library, which means almost al data pumping can just be done in Python, Pandas, or something else.

Mesa models are just Python classes. You can already pass complex parameters today. Like:

```python
# Fixed nested config - no extraction needed
model_params = {
    "n": {"type": "SliderInt", "value": 50, "min": 10, "max": 100},
    "grid_config": {"width": 10, "height": 10},  # Fixed param, no UI
}

class MyModel(Model):
    def __init__(self, n, grid_config, rng=None):
        super().__init__(rng=rng)
        # Just use grid_config directly
        self.grid = MultiGrid(grid_config["width"], grid_config["height"])
```

Mesa's `split_model_params()` already separates user-adjustable params (with `"type"` keys) from fixed params.

For custom UI, just override `UserInputs`:

```python
@solara.component
def MyUserInputs(user_params, on_change):
    # Render whatever custom UI you need
    # You have full Solara/Python power here
```

So I'm afraid this PR adds quite some complexity for marginal benefit when most users can just use fixed parameters, creates multiple ways to do the same thing (confusing API surface), fights against the composability (that's the whole point of using Solara), and your own example proves it doesn't help since you still needed to override UserInputs anyway.

I think we could improve the documentation showing these patterns:
- Using fixed parameters for complex configs (works today)
- Overriding `UserInputs` for custom UI (works today)
- Building fully custom Solara apps when needed (works today)

The whole point of using Python and Solara is that users have full power to extend and customize. What we offer with Mesa are building blocks to get started that are useful to a broad audience.

Thanks for the PR and thought that went into it, that's appreciated! If you tell more about your actual issues and frustrations, we can see what would be a fitting solution.

I would also like to hear @Sahil-Chhoker takes, especially on if there's low-hanging fruit on improving this (either though code or documentation).

usuallycwdillon:
Hi @EwoutH, Thanks for the quick review. 

Respectfully, I would not have gone to all this effort if overriding `UserInputs` for custom UI worked. For example, take a simple dict pattern, like the one I added for the Boltzmann wealth model. It's possible to initialize the model with default values as in the original, `model = BoltzmannWealth(n=50, grid_dimension=extract_dict_values({'width': 10, 'height':10}) )` and the model will run a single time, because the function to extract 'width' and 'height' can be called outside of Solara to initialize the model. 

However, if you click 'Reset' in the interface, the `do_reset()` function in `solara_viz.py` does **NOT** know how to use the information contained in the 'grid_dimension' from `model_params` spec. It won't even pass it to model to be handled there. It passes a `None` value. 

The issue we had to solve was not how custom UI controls can be made, but how `solara_viz.py` handles the information we give it. That cannot be done with Mesa as it is currently. This seems the least invasive way to do it.

EwoutH:
Thanks for your perspective. I would like to hear @Sahil-Chhoker's take on this.

EwoutH:
Was the high-level design or user API for this feature discussed anywhere before this PR?

usuallycwdillon:
Not at all, aside from promising Jackie and Tom (two years ago, when I
consulted with them before proposing to use Mesa for our project) that I
would share anything useful we created. ...that's hardly specific.

Honestly, I'm surprised by your hesitation. Without counting the code we
added for the examples, testing, and the docstrings, we're suggesting that
Mesa add 9 lines of real code and change 1. This is the least intrusive way
to add a new capability.

quaquel:
@usuallycwdillon, could you check what is going on with the tests and ensure that these are passing. I understand the use case and need for this functionality. Although Boltzmann might not be the best example to showcase it. The $\alpha$ and $\beta$ example is however quite clear. The main reason for being hesitant is that any code we add needs to be maintained as well. We like to iterate a bit over API design and implementation before committing to a new feature. Often, we first discuss the API and implementation before putting in a PR. So, please bear with us. 

@Sahil-Chhoker, please take a look and give us your perspective. I'll try to find time to take a look myself as well at the Boltzmann example. 

implementation wise, @usuallycwdillon, you might want to check [`functools.singledispatch`](https://docs.python.org/3/library/functools.html#functools.singledispatch). At face value, this offers a stdlib solution on which to build this which would further simplify the mesa side of things. Your example would simplify to

```python
from mesa.visualization.solara_viz import param_extractor

@param_extractor.register(dict)
def extract_dict_param(spec):
    """Extract default values from a nested dictionary parameter."""
    entries = spec.get("entries", {})
    result = {}
    for key, value in entries.items():
        if isinstance(value, dict) and "value" in value:
            result[key] = value["value"]
        elif isinstance(value, dict):
            result[key] = extract_dict_param({"entries": value})
        else:
            result[key] = value
    return result
```

The main ambiguity in your example is how "Dict" should be understood in your example. `functools.singledispatch` dispatches on the real python type of the first argument. So if you want to use multiple different "types" that are in fact all dicts, it won't work. A simple solution then might be to use data classes instead of the dictionary based configuration system. The advantage of using a dataclasses is that you can use defaults and rely on autocomplete instead of having to type out the entire dictionary. Dataclasses will work here because spec becomes a specific dataclass type and we register these to singledispatch by type. So each data class has its own handler function.

Sahil-Chhoker:
> However, if you click 'Reset' in the interface, the `do_reset()` function in `solara_viz.py` does **NOT** know how to use the information contained in the 'grid_dimension' from `model_params` spec. It won't even pass it to model to be handled there. It passes a `None` value.

Yes, this is the current behavior of `solara_viz`, and it does make it hard to reliably define and reuse custom inputs. That’s why introducing a data extractor feels like a reasonable step.

While it’s true that this adds some complexity and that the use case may not be very common, it still seems like a necessary change. Right now, model resets and model initialization are handled separately, which makes the system easy to break. Small differences between the two can lead to failures in the visualization. 

In my opinion having a data extractor would help keep initialization and reset behavior aligned, making state handling more predictable and less error-prone.

usuallycwdillon:
I've fixed those tests and updated the multiple `with` statements in the `custom_params.py`. (I had to look that up, I didn't even know I _could_ do that.) The final call to `page` in `app_with_dict_params.py` is standard practice, but it looks like quashing the code check for that is also standard.

That final failing test should be unrelated to this change, unless I misunderstand what's going on there. 

@Sahil-Chhoker It took me a little while to figure out what was going on. We were setting default values in our `model.py` and couldn't figure out why the model behavior didn't change when we moved the sliders around.  I initially set out to completely rewrite the `Page()` class, but eventually realized that passing custom extractor functions to the existing `do_reset()` was not only easier, but it worked more reliably. 

@quaquel Yes, we're using much more complicated structures than dictionaries. In our use case, we have quite a few parameters that can change the initial conditions and the strategies agents use in the simulation, so we anticipate users will primarily run the model using YAML configuration files and we've added command line args to our `run.py` to allow users to specify a 'my_config.cfg' file. 

But that approach turned out to be incompatible with the UI controls. In many cases, we have nested dicts where the values are lists. We also created "range sliders" for int and float types. Some top-level dicts are collapsible (to keep the length of the controls manageable) so that it's possible to contain several values in a set. For example:

```
"rq_distribution": {
        "type": "Dict",
        "collapsible": True,
        "label": "Task Requirements Distribution Parameters",
        "entries": {
            "cognitive": {
                "alpha": {
                    "type": "SliderRangeInt",
                    "label": "Min/Max alpha",
                    "value": sorted([1, 2]),
                    "min": 1.0,
                    "max": 20,
                    "step": 1
                },
                "beta": {
                    "type": "SliderRangeInt",
                    "label": "Min/Max beta",
                    "value": sorted([1, 2]),
                    "min": 1.0,
                    "max": 20,
                    "step": 1
                },
            },
            "physical": {
                "alpha": {
                    "type": "SliderRangeInt",
                    "label": "Min/Max alpha",
                    "value": sorted([1, 2]),
                    "min": 1,
                    "max": 20,
                    "step": 1
                },
                "beta": {
                    "type": "SliderRangeInt",
                    "label": "Min/Max beta",
                    "value": sorted([1, 2]),
                    "min": 1.0,
                    "max": 20,
                    "step": 1
                },
            }
        }
    },
```

We made it possible to hide elements, too. For example, we have a sub-model for reassignment of team members between teams, which requires the user to add information about the extent, frequency, replacement rules, etc. But, if the 'allow_reassignment' boolean is false, none of those controls are presented to the user (and the unused defaults are sent to `model.py`, to prevent `__init__` from complaining). 

We're happy to share the whole `custom_params.py` if it's useful, but I don't think there's enough user demand for these things to expand the `Controls` card with it.

quaquel:
Thanks a lot for the extensive explanation and updates to the tests. I hope to get to this soon because we plan a 3.5 release somewhere this month. @Sahil-Chhoker, if you have time to take a second look as well, that would be much appreciated.

Sahil-Chhoker:
> @Sahil-Chhoker, if you have time to take a second look as well, that would be much appreciated.

I'll definitely try, but with how busy this year is looking, I can’t promise anything.

quaquel:
This PR has stalled and now has conflicts. @usuallycwdillon, do you have time to resolve these conflicts?

usuallycwdillon:
I _think_ I'll be able to work on it today. If not, tomorrow at the latest.

PR #3214, opened 2026-01-27
Micro optimization of grid.select_random_empty_cell

quaquel:
This is a micro optimization of `grid.select_random_empty_cell`. Currently, this calls cellcollection.select_random_cell, which in turn does `random.choice(cells)`. This indirection can be avoided by just having a `_celllist` inside `grid` and using `random.choice` directly on it. Then, to top it off, we assign `self.random` and `self._celllist` to local variables in the method to further improve performance. This trick avoids the Python calling overhead of attribute lookup. Since this for loop runs 50 times in the worst case, this micro stuff adds up a lot.

### DIFF
--- a/mesa/discrete_space/grid.py
+++ b/mesa/discrete_space/grid.py
@@ -120,6 +120,7 @@ class Grid(DiscreteSpace[T], HasPropertyLayers):
             coord: self.cell_klass(coord, capacity, random=self.random)
             for coord in coordinates
         }
+        self._celllist = list(self._cells.values())
         self._connect_cells()
         self.create_property_layer("empty", default_value=True, dtype=bool)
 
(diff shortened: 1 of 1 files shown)


quaquel:
If I use timeit with Schelling, on a 50X50 grid, a fixed seed, and a density of 0.9, the results are for just `grid.select_random_empty_cell`.

**3.4.2**: 3.12 μs ± 64 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
**with this PR merged**: 2.42 μs ± 21.6 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

quaquel:
It was a no-brainer that I came across while microbenchmarking stuff for mesa-rust.

EwoutH:
20% in Schelling is quite a lot

quaquel:
Large schelling is in essence just a test for how fast `get_random_empty_cell` is. Which is what I figured out while working on the mesa-rust stuff.

PR #3217, opened 2026-01-27
Bugfix for pickling dynamically modified grids

### LINKED ISSUE #3215: Grid connectivity changes are lost after pickle / unpickle
I ran into an issue when using pickle to save and restore a model that modifies grid connectivity at runtime (for example, creating walls by disconnecting neighboring cells).

After unpickling, the grid always falls back to its default neighbor structure (Moore / Von Neumann / Hex), even if cells were explicitly disconnected before saving. This means runtime topology changes are silently lost, which makes checkpointing unreliable for models with dynamic grids.

### Reproduction:
```python 
import pickle

grid = OrthogonalMooreGrid((3, 3))

a = grid[(1, 1)]
b = grid[(1, 2)]

a.disconnect(b)
b.disconnect(a)

restored = pickle.loads(pickle.dumps(grid))

ra = restored[(1, 1)]
rb = restored[(1, 2)]

print(rb in ra.connections.values())  # True (but should be False)
```

During pickling, cell connections are cleared:
```python
# grid_cell.py
state[1]["connections"] = {}
```

Then during unpickling, the grid rebuilds the default topology:
```python
# grid.py
self._connect_cells()
```

So any runtime changes to connectivity are overwritten.

This matters because many models rely on changing spatial connectivity (walls, obstacles, damaged roads, etc.). After unpickling, the model is no longer running on the same world it was saved from, which breaks reproducibility and checkpointing.

### Expected behavior
After pickle / unpickle, the grid should preserve whatever connectivity existed at the time it was saved. If this isn’t supported by design, it would at least be good to document it clearly.

If this is indeed a bug, I’d be happy to work on a fix or propose a PR.

quaquel:
This bugfix updates `Cell.__getstate__` and `Grid.__setstate__` to ensure that the actual structure of the grid is properly preserved when pickling a grid. 

closes #3215

### DIFF
--- a/mesa/discrete_space/discrete_space.py
+++ b/mesa/discrete_space/discrete_space.py
@@ -175,7 +175,6 @@ class DiscreteSpace[T: Cell]:
 
         for cell in self._cells.values():
             connections = {
-                coordinate: self._cells[coordinate]
-                for coordinate in cell.connections
+                coordinate: self._cells[coordinate] for coordinate in cell.connections
             }
             cell.connections = connections


mesa/discrete_space/discrete_space.py
```diff
@@ -172,4 +172,10 @@ def select_random_empty_cell(self) -> T:
     def __setstate__(self, state):
         """Set the state of the discrete space and rebuild the connections."""
         self.__dict__ = state
-        self._connect_cells()
+
+        for cell in self._cells.values():
+            connections = {
+                coordinate: self._cells[coordinate]
+                for coordinate in cell.connections.keys()
```
EwoutH: To fix ruff / pre-commit
```suggestion
                for coordinate in cell.connections
```

quaquel: Grid pickling was covered by tests, but not for dynamically modified grids. Theoretically, the grid data structure allows for dynamic modifications. However, because of the caching, which is essential for performance, we did not develop that further into a real API. #3086 was just taking advantage of a version bump on the minimum Python side, so not a bug.

ShreyasN707:
> Thanks.
> 
> Since pickling has been more often an issue (#2378, #3086, and most recently #3205), should we create a follow-up issue to increase test coverage on pickling follow-up? And maybe pickle and unpickle our example models in a test?

@quaquel Thanks for jumping on this so quickly — it’s great to see this getting fixed. I’d be happy to help add some tests around pickling/unpickling dynamic grids or example models if that would be useful.

PR #3219, opened 2026-01-28
Introduce `_HardKeyAgentSet` in agents.py

codebreaker32:
## Summary
This PR introduces `_HardKeyAgentSet`, that inherits from `AbstractAgentSet` using strong references (standard dict) instead of weak references. This eliminates WeakKeyDictionary overhead for Model-managed collections where lifecycle is explicitly controlled.
## Motive
See #3128 and #3160

## Implementation:
- Inherits from `AbstractAgentSet`
- Uses `dict[A, None]` for storage (vs `WeakKeyDictionary`)
- All abstract methods implemented with dict-based optimizations
- copying strong-ref sets creates memory leak risk. By downgrading to `AgentSet`, we get best of both worlds

### Changes
1. Introduced `_HardKeyAgentSet`
2. Moved `to_list()` from AgentSet to AbstractAgentSet

✅ Fully backward compatible - no changes to existing APIs. This is a new internal class.
Ticks (2) from #3209

### DIFF
--- a/tests/test_agent.py
+++ b/tests/test_agent.py
@@ -857,7 +857,8 @@ def test_hardkeyagentset_init():
     agents = [AgentTest(model) for _ in range(5)]
 
     with pytest.warns(UserWarning):
-        _ = _HardKeyAgentSet([])
+        _ = _HardKeyAgentSet([], random=None)
+        _ = _HardKeyAgentSet(agents, random=None)
     hard_set = _HardKeyAgentSet(agents, model.random)
 
     assert len(hard_set) == 5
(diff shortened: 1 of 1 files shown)


codebreaker32:
As `_HardKeyAgent` doesn't inherit from `Sequence`, DeffuantWeisbuchModel is raising `TypeError`

```plaintext
        if not isinstance(population, _Sequence):
>           raise TypeError("Population must be a sequence.  "
                            "For dicts or sets, use sorted(d).")
E           TypeError: Population must be a sequence.  For dicts or sets, use sorted(d).
```

quaquel:
> As _HardKeyAgent doesn't inherit from Sequence, DeffuantWeisbuchModel is raising TypeError

We have to fix the model anyway because of the deprecation. The source seems to be `self.random.sample(self.agents, 2)`. Which is easily replaced with `self.random.sample(self.agents.tolist(), 2)`.

codebreaker32:
I think `to_list()` is quite good improvement
  For eg. In this DeffuantWeisbuchModel
  ```python
# Looks like O(N), but is actually O(N²)
for _ in range(self.n):
    # random.sample calls __getitem__ repeatedly.
    # Since our __getitem__ rebuilds the list every time (O(N)), this line is expensive.
    agent_a, agent_b = self.random.sample(self.agents, 2)
  ```
  Initially user was doing this, unaware of what was going under the hood
  Now this API explicitly tells them what they are doing giving them chance to optimise
  ```python
  agents_list = self.agents.to_list()
  for _ in range(self.n):
          agent_a, agent_b = self.random.sample(agents_list, 2)

# Truly O(N)
  ```
  However this was only possible because loop didn't involve any addition or deletion but I think it is still
  cool

quaquel:
> Initially user was doing this, unaware of what was going under the hood

And this was exactly the reason why I wanted to deprecate it. I saw other models doing exactly this without realizing the overhead.

quaquel:
Coverage sits at 87%. Can you check this and see if you can add relevant tests to bring this up? I'll try to review later today.

codebreaker32:
> Can you check this and see if you can add relevant tests to bring this up?

Done

mesa/model.py
```diff
@@ -214,7 +214,7 @@ def _do_step(self) -> None:
         self._schedule_step(self.time + 1)
 
     @property
-    def agents(self) -> AgentSet[A]:
+    def agents(self) -> _HardKeyAgentSet[A]:
```
quaquel: Might it be a good idea to update the docstrings for agents and agents_by_type to highlight that they return an agentset with hard refs and should not be messed with, since they are the key agent registration system in Mesa?

quaquel (approved):
One remaining question regarding how to properly document this. Codewise, I am fine with merging this.

codebreaker32:
> One remaining question regarding how to properly document this. Codewise, I am fine with merging this.

I see two options:
1. Update it in this PR only, 
2. We can update documentation in next PR where I am planning to remove the `self._agents` and updating documentation 

If you prefer option 2, I will revert the changes in `model.py`

codebreaker32:
I prefer option 2, we can finalize the storage refactor there

quaquel:
Option 2 is fine with me as well. I'll take a last look at this and merge it hopefully this afternoon.

codebreaker32:
PR title and description updated to reflect reverted `model.py` changes

PR #3220, opened 2026-01-29
check whether model reporter is a partial function

wang-boyu:
fixes https://github.com/mesa/mesa/issues/1871

If we use a partial function as a data collector, for example

```python
from functools import partial


def count_agents(model, multiplier):
    return len(model.agents) * multiplier

dc = DataCollector(
    model_reporters={"AgentsTimesTwo": partial(count_agents, multiplier=2)}
)
dc.collect(model)
```

then we will get a RuntimeError:

```bash
# Type 2: Method of class/instance (bound methods are callable)
if callable(reporter) and not isinstance(reporter, types.LambdaType):
    try:
>        reporter()  # Call without args for bound methods
          ^^^^^^^^^^
E               TypeError: TestPartialReporterValidation.test_partial_model_reporter.<locals>.count_agents() missing 1 required positional argument: 'model'

mesa/datacollection.py:172: TypeError
```

A similar problem was previously fixed in https://github.com/mesa/mesa/pull/1872 by checking whether the reporter is a partial function:

```python
# Check if lambda or partial function
if isinstance(reporter, (types.LambdaType, partial)):
```

However in https://github.com/mesa/mesa/pull/2605 where we extracted a separate function to validate reporters, this check was not performed:

```python
# Type 1: Lambda function
    ...
```

This PR essentially added this check into the model reporter validation function:

```python
# Type 1: Lambda function or partial
if isinstance(reporter, (types.LambdaType, partial)):
    ...
```

and added a test case for when a model reporter is a partial function.

### DIFF
--- a/mesa/datacollection.py
+++ b/mesa/datacollection.py
@@ -64,9 +64,10 @@ class DataCollector:
 
         Both model_reporters, agent_reporters, and agenttype_reporters accept a
         dictionary mapping a variable name to either an attribute name, a function,
-        a method of a class/instance, or a function with parameters placed in a list.
+        a method of a class/instance, a partial function, or a function with
+        parameters placed in a list.
 
-        Model reporters can take four types of arguments:
+        Model reporters can take five types of arguments:
         1. Lambda function:
            {"agent_count": lambda m: len(m.agents)}
         2. Method of a class/instance:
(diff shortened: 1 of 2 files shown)


quaquel:
Can you ensure the starting post is independently readable instead of only a link? Otherwise, this looks fine to me.

wang-boyu:
Sorry about the trouble. I've edited the description.

PR #3222, opened 2026-01-29
Prevent RecursionError and data loss in Cell/Grid deepcopy

### LINKED ISSUE #3221: Data Loss and Recursion Errors in Cell/Grid Deepcopy/Pickle
**Description**
The `discrete_space` system currently fails when attempting to copy or pickle models, particularly those with large grids. There are two distinct failure modes:

- Data Loss: Standalone Cell objects explicitly set their neighbor connections to None during serialization in `getstate.` This makes it impossible to inspect or use connections on a copied cell.
- Recursion Errors: For large grids (e.g., 100x100), the highly interconnected graph of cells triggers a `RecursionError` in Python's `deepcopy` and `pickle` modules because the default recursion limit is exceeded while traversing circular references.

**Expected behavior**
- Performing a `deepcopy` on a Cell should preserve its connection metadata (keys and neighbor mapping).
- Performing a `deepcopy` on a Grid (or any `DiscreteSpace`) should handle any size supported by Mesa without crashing, while ensuring that all circular references between cells are correctly reconstructed in the copy.

**To Reproduce**
```python
import copy
from mesa.discrete_space import Cell, OrthogonalMooreGrid

# 1. Standalone Data Loss
c1 = Cell(coordinate=(0, 0))
c2 = Cell(coordinate=(0, 1))
c1.connect(c2)
c1_copy = copy.deepcopy(c1)
print(f"Neighbor in copy: {c1_copy.connections[(0, 1)]}") 
# Expected: (0, 1) or Cell object
# Actual: None

# 2. Recursion Error in Large Grids
grid = OrthogonalMooreGrid((100, 100))
# This triggers: RecursionError: maximum recursion depth exceeded
grid_copy = copy.deepcopy(grid) 
```

**Additional context**
The issue stems from the current `Cell.__getstate__` implementation attempting to prevent recursion by simply nullifying objects. A more robust solution is to use a coordinate-based serialization strategy where `Cell` objects store neighbor coordinates during state extraction, and the parent `DiscreteSpace` container handles the relinking of these coordinates back into object references during state injection (`setstate`).

falloficaruss:
### Description
This PR addresses critical bugs in the new `discrete_space` system when copying or pickling model states. This is essential for model branching, batch runs, and saving/loading model progress.

### Problem
- RecursionError: In large or dense grids (e.g., 100x100), the default `deepcopy` algorithm exceeds Python's maximum recursion depth due to the high number of interconnected circular cell references.
- Data Loss: The previous `Cell.__getstate__` implementation attempted to avoid recursion by nullifying neighbor references. This made standalone copied cells unusable (neighbors became None) and broke relational integrity within the grid.
- Overwritten Modifications: `Grid.__setstate__` was re-calling `_connect_cells()`, which would blindly recalculate the default grid pattern and overwrite any custom connections or removals made to the grid topology before serialization.

### Solution
- Coordinate-based Serialization: `Cell.__getstate__` now extracts the neighbor's coordinate instead of the object reference during state collection. This "breaks" the circular object chain for the pickler, preventing `RecursionError`.
- Structured Relinking: `DiscreteSpace` and `Grid` handle the "relinking" phase in `setstate`. Once all cells are restored, the parent space iterates through them and replaces the coordinate placeholders with actual `Cell` object pointers from its internal mapping.
- Topology Preservation: By relinking based on the serialized state rather than re-running the generator pattern, any manual changes to the grid (like removing specific walls or adding non-standard links) are correctly preserved across the copy/pickle cycle.

### Changes
- `mesa/discrete_space/cell.py`: Updated `getstate` to use coordinate mapping.
- `mesa/discrete_space/discrete_space.py`: Added coordinate-based relinking in `setstate`.
- `mesa/discrete_space/grid.py`: Updated `setstate` to use the new relinking logic and removed the redundant `_connect_cells()` call.
- `tests/discrete_space/test_discrete_space.py`: Added a new test test_cell_deepcopy to verify both standalone and grid-based relational integrity.

Closes #3221

### DIFF
--- a/mesa/discrete_space/grid.py
+++ b/mesa/discrete_space/grid.py
@@ -197,9 +197,6 @@ class Grid(DiscreteSpace[T], HasPropertyLayers):
         """Custom __setstate__ for handling dynamic GridCell class and PropertyDescriptors."""
         super().__setstate__(state)
 
-        self.cell_klass = type(
-            next(iter(self._cells.values()))
-        )  # the __reduce__ function handles this for us nicely
         for layer in self._mesa_property_layers.values():
             setattr(
                 self.cell_klass,
--- a/mesa/discrete_space/discrete_space.py
+++ b/mesa/discrete_space/discrete_space.py
@@ -173,6 +173,9 @@ class DiscreteSpace[T: Cell]:
         """Set the state of the discrete space and rebuild the connections."""
         self.__dict__ = state
 
+        if self._cells:
+            self.cell_klass = type(next(iter(self._cells.values())))
+
         for cell in self._cells.values():
             cell.connections = {
                 key: self._cells[coord] for key, coord in cell.connections.items()
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -19,7 +19,7 @@ import numpy as np
 if TYPE_CHECKING:
     from mesa.experimental.devs import Simulator
 
-from mesa.agent import Agent, AgentSet
+from mesa.agent import Agent, _HardKeyAgentSet
 from mesa.experimental.devs.eventlist import EventList, Priority, SimulationEvent
 from mesa.experimental.scenarios import Scenario
 from mesa.mesa_logging import create_module_logger, method_logger
@@ -162,13 +162,10 @@ class Model[A: Agent, S: Scenario]:
         self.step = self._wrapped_step
 
         # setup agent registration data structures
-        self._agents: dict[
-            A, None
-        ] = {}  # the hard references to all agents in the model
         self._agents_by_type: dict[
-            type[A], AgentSet[A]
+            type[A], _HardKeyAgentSet[A]
         ] = {}  # a dict with an agentset for each class of agents
-        self._all_agents: AgentSet[A] = AgentSet(
+        self._all_agents: _HardKeyAgentSet[A] = _HardKeyAgentSet(
             [], random=self.random
         )  # an agenset with all agents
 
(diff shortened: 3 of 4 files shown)


mesa/discrete_space/cell.py
```diff
@@ -241,13 +241,15 @@ def _neighborhood(
         return neighborhood
 
     def __getstate__(self):
-        """Return state of the Cell with connections set to empty."""
+        """Return state of the Cell, replacing neighbor objects with coordinates to break recursion."""
```
quaquel: I am fine with extending the docstring, but the motivation should not be in the first line. The first line is just what the method does.

mesa/discrete_space/cell.py
```diff
@@ -241,13 +241,15 @@ def _neighborhood(
...
         state = super().__getstate__()
-        # Replace connections with empty dict to avoid infinite recursion error in pickle/deepcopy
-        state[1]["connections"] = {
-            cell.coordinate: None for cell in self.connections.values()
+        # Replace neighbor objects with their coordinates to avoid deep recursion
+        # while preserving the connection keys.
+        slots = state[1].copy()
```
quaquel: why?

mesa/discrete_space/grid.py
```diff
@@ -196,10 +196,15 @@ def __getstate__(self) -> dict[str, Any]:
     def __setstate__(self, state: dict[str, Any]) -> None:
         """Custom __setstate__ for handling dynamic GridCell class and PropertyDescriptors."""
         self.__dict__ = state
-        self._connect_cells()  # using super fails for this for some reason, so we repeat ourselves
+
```
quaquel: can you check if calling super here now works. It would simplify the implementation and avoid repetition.

mesa/discrete_space/cell.py
falloficaruss: The `.copy()` is a defensive measure to ensure we aren't mutating the dictionary returned by `super().__getstate__()` in-place, which can sometimes lead to side effects during complex `deepcopy` operations.

mesa/discrete_space/grid.py
falloficaruss: Sure. I'll get back to you with it.

mesa/discrete_space/cell.py
quaquel: But in that case, you are doing a double copy, and this will be expensive. So, I still don't see the exact reason. Do you have an example of how this is causing a bug or a link that points to such bugs more generally?

falloficaruss: The initial reasoning was purely defensive—to ensure that modifying the state dictionary didn't inadvertently side-effect the source object's internal state mapping during complex `deepcopies`.
But maybe `.copy()` is indeed redundant here so I'll get back to you with an update.

mesa/discrete_space/grid.py
```diff
@@ -195,11 +195,10 @@ def __getstate__(self) -> dict[str, Any]:
...
-        self.__dict__ = state
-        self._connect_cells()  # using super fails for this for some reason, so we repeat ourselves
+        super().__setstate__(state)
 
         self.cell_klass = type(
-            self._cells[(0, 0)]
+            next(iter(self._cells.values()))
         )  # the __reduce__ function handles this for us nicely
         for layer in self._mesa_property_layers.values():
             setattr(
```
quaquel: After reviewing your code, I realize this should be moved to the DiscreteSpace class.

quaquel (approved):

PR #3224, opened 2026-01-29
Update `model.py` to replace `AgentSet` with `_HardKeyAgentSet`

### LINKED ISSUE #3128: Do we really need weakrefs in AgentSet?

codebreaker32:
This PR finalizes the agent storage refactor by fully integrating `_HardKeyAgentSet` as the primary storage mechanism in `Model`.
Previously, `Model` maintained two parallel structures: `self._agents`(for lookups and maintaining Hard References of agents) and `self._all_agents` for iteration. This duality created memory redundancy.

This PR resolves it by:
1. Removing `self._agents` (the raw dictionary).
2. Updating `self._all_agents` to be an instance of `_HardKeyAgentSet`.
3. Updating `self._agents_by_type` to use `_HardKeyAgentSet` for type-specific groups.
4. Updating docstrings to reflect the changes and warnings accordingly.

See #3128 and #3160 for complete discussion
Ticks (3) from #3209 
Resolves #3128 

*This is a follow-up PR of #3210 and #3219*

### DIFF
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -19,7 +19,7 @@ import numpy as np
 if TYPE_CHECKING:
     from mesa.experimental.devs import Simulator
 
-from mesa.agent import Agent, AgentSet
+from mesa.agent import Agent, _HardKeyAgentSet
 from mesa.experimental.devs.eventlist import EventList, Priority, SimulationEvent
 from mesa.experimental.scenarios import Scenario
 from mesa.mesa_logging import create_module_logger, method_logger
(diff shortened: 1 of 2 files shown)


quaquel (approved):
This looks good to me. I particularly appreciate the very clear docstrings.

PR #3226, opened 2026-01-30
Refactor `PropertyLayer.select_cells` into explicit mask/index methods

### LINKED ISSUE #3225: Refactor `PropertyLayer.select_cells` into explicit mask and index methods
The current implementation of `select_cells` in `PropertyLayer` determines its return type based on the boolean flag `return_list`. If `True`, it returns a list of coordinate tuples; if `False`, it returns a boolean NumPy mask. This behavior is implicit and inconsistent, making type hinting difficult and the API less intuitive. A `FIXME` comment in the code specifically suggests splitting this into two separate functions to improve clarity.

I would like to refactor `PropertyLayer` (and the corresponding wrapper in `HasPropertyLayers`) to use explicit methods for each return type:
1. Create `select_cells_mask(self, condition)`: Returns the raw boolean `np.ndarray` mask.
2. Create `select_cells_indices(self, condition)`: Returns the list of coordinate tuples (e.g., `[(x, y), ...]`).
3. Deprecate `select_cells`: Modify the existing method to wrap the two new methods. It should emit a `FutureWarning` advising users to switch to the new explicit methods, preserving backward compatibility for now.

Relevant file: `mesa/discrete_space/property_layer.py`

```python
    def select_cells(self, condition: Callable, return_list=True):
        # ...
        # fixme: consider splitting into two separate functions
        #  select_cells_boolean
        #  select_cells_index

```

Nithurshen:
### Summary

This PR refactors the `select_cells` method in both `PropertyLayer` and the `HasPropertyLayers` mixin. It introduces two new explicit methods (`select_cells_mask` and `select_cells_indices`) to replace the ambiguous behavior of the existing method. The original `select_cells` is retained for backward compatibility but now emits a `FutureWarning`.

Fixes #3225 

### Motive

The previous implementation of `select_cells` relied on a boolean flag (`return_list`) to determine whether to return a list of coordinate tuples or a boolean NumPy mask. This implicit return type made type inference difficult and the API less intuitive. A `FIXME` comment in the source code specifically suggested splitting this functionality into separate, explicit functions to improve code clarity and maintainability.

### Implementation

1. **New Methods**:
* Added `select_cells_mask(self, condition)`: Returns the raw boolean `np.ndarray` mask.
* Added `select_cells_indices(self, condition)`: Returns a list of coordinate tuples (e.g., `[(x, y), ...]`).

2. **Deprecation**:
* Modified the existing `select_cells` method to wrap the two new methods.
* Added a `FutureWarning` to `select_cells` advising users to switch to the new explicit methods.

3. **Refactoring**: Applied these changes to both the `PropertyLayer` class and the `HasPropertyLayers` mixin in `mesa/discrete_space/property_layer.py`.
4. **Testing**: Updated `tests/discrete_space/test_discrete_space.py` to verify the new methods work as expected and that the deprecated method correctly raises a warning.

### Usage Examples

**Before:**

```python
# To get a list of coordinates
indices = grid.select_cells(condition, return_list=True)

# To get a boolean mask
mask = grid.select_cells(condition, return_list=False)

```

**After:**

```python
# To get a list of coordinates
indices = grid.select_cells_indices(condition)

# To get a boolean mask
mask = grid.select_cells_mask(condition)

```

### Additional Notes

This change preserves backward compatibility. However, calls to `select_cells` will now trigger a warning, encouraging users to update their code before the method is eventually removed in a future release.

### DIFF
--- a/mesa/discrete_space/property_layer.py
+++ b/mesa/discrete_space/property_layer.py
@@ -156,24 +156,49 @@ class PropertyLayer:
             self.data[mask] = vectorized_operation(target_data)
 
     def select_cells(self, condition: Callable, return_list=True):
-        """Find cells that meet a specified condition using NumPy's boolean indexing, in-place.
+        """Find cells that meet a specified condition (Deprecated).
+
+        .. warning::
+            This method is deprecated. Please use `select_cells_mask` or
+            `select_cells_indices` instead.
 
         Args:
             condition: A callable that returns a boolean array when applied to the data.
             return_list: (Optional) If True, return a list of (x, y) tuples. Otherwise, return a boolean array.
+        """
+        warnings.warn(
+            "select_cells is deprecated and will be removed in a future version. "
+            "Use select_cells_indices or select_cells_mask instead.",
+            FutureWarning,
+            stacklevel=2,
+        )
+        if return_list:
+            return self.select_cells_indices(condition)
+        else:
+            return self.select_cells_mask(condition)
+
+    def select_cells_mask(self, condition: Callable) -> np.ndarray:
+        """Find cells that meet a specified condition and return a boolean mask.
+
+        Args:
+            condition: A callable that returns a boolean array when applied to the data.
 
         Returns:
-            A list of (x, y) tuples or a boolean array.
+            A boolean numpy array where True indicates the condition is met.
         """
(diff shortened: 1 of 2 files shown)


quaquel:
See my response in #3225.

Issue #3227, opened 2026-01-30
mesa signal improvements


quaquel:
While working on making `Model` reactive, I have been refactoring mesa_signals. Below is a list of potential improvements. 

mesa_signals is under active development, so PRs at this stage on any of these issues that are not in sync with the development I am doing, e.g., #3212, will not be accepted. Ideas and other suggestions are clearly welcome.

- [ ] Add a `batch` context manager; accumulate all signals until the context manager exists, for `Observable.CHANGE`, only keep the last signal, and also otherwise filter out any redundant signals.
- [ ] Add a `suppress` context manager; fully discard all signals until the context manager exists.
- [ ] Add `HasObservable.subscribe`; make it possible to subscribe at the class level such that the subscription is inherited by all instances. This would make it trivial to observe agents without relying on `model.register_agent`. The main challenge is that, while running experiments, it is critical that the observer (and by extension, the model instance) is properly garbage-collected.

Satyamgupta2365:
Hi @quaquel, I’m interested in working on this issue and can align with the ongoing mesa_signals refactor. If it works for you, please feel free to assign this task to me.

codebreaker32:
While working on some behavioral models, I found a critical gap where we need `track()`.

Static dependencies (`dependencies=[]`) work great for Events on `self`, and Auto-Discovery works great for State (Observables) on any object. But we currently have no way to listen to Events on Runtime Objects.

When you need to listen to a specific object determined at runtime (e.g., a neighbor or target) emitting an **Event** (not a state change).

```python
class Predator(Agent):
    @computed_property
    def reaction(self):
        # target changes every step (Agent A, then Agent B...)
        target = self.find_closest_prey() 
        
        # Auto-discovery and static deps fail: We can't write dependencies=[("target", ...)] 
        # because 'target' is a local variable, not a fixed attribute.
        
        # We must track the specific object instance dynamically
        track(target, "shouted")
        
        return f"Target {target.unique_id} shouted!"

```

The issue: Static dependencies resolve by name on `self`. When the dependency is **an Event on a Runtime Reference**, both mechanisms fail. We need `track(object, signal)` to subscribe by object reference during execution.

I believe a hybrid approach combining both `track()` and static deps would be a better solution

quaquel:
Thanks for this example. I am trying to get my head around it, but I am not sure I fully follow. If on every access of computed_property, target is different, should it even be implemented as a computed_property? 

[...]

PR #3230, opened 2026-01-30
feat: add batch and suppress context managers to mesa_signals

Nithurshen:
### **Summary**

This PR introduces two new context managers to the `mesa_signals` system: `batch` and `suppress` mentioned in #3227. These tools provide fine-grained control over signal emission, allowing for significant performance optimizations during bulk operations and minimizing redundant updates in reactive workflows.

### **Motive**

Currently, every change to an `Observable` or `ObservableList` triggers an immediate signal. While this ensures reactivity, it can lead to "signal storms" during bulk initializations or complex state updates.

1. Multiple updates to the same attribute in a single logical step cause multiple UI redraws or downstream re-computations.
2. In scenarios like model setup, intermediate signals are often unnecessary and computationally expensive.

### **Implementation**

The features were integrated directly into the `HasObservables` mixin:

* `suppress()`: Uses a `_silenced` boolean flag. When active, the `notify()` method returns immediately, discarding all signals.
* `batch()`: Implements a `_signal_buffer` list. When active, signals are collected rather than emitted.
* Signal Deduplication: Upon exiting a `batch` context, the `_flush_buffer()` method processes the queue. For `SignalType.CHANGE`, it performs a "last-one-wins" optimization, ensuring only the final state of an attribute is emitted. Other signal types (like list appends) are preserved in their original order.
* Nesting Support: The `batch` manager supports nesting; flushing only occurs when the outermost context exits.

### **Usage Examples**

**Using `suppress` to silence bulk initialization:**

```python
model = MyModel()
with model.suppress():
    # None of these changes will emit signals to observers
    for i in range(1000):
        model.large_array[i] = initial_value 

```

**Using `batch` to prevent redundant UI updates:**

```python
# Instead of 3 separate 'change' signals, only 1 signal (value=3) is emitted at the end
with model.batch():
    model.value = 1
    model.value = 2
    model.value = 3

```

### **Additional Notes**

* This PR includes comprehensive unit tests in `tests/experimental/test_mesa_signals.py` covering deduplication, signal ordering, and nested contexts.
* These changes are foundational for making the `Model` class reactive without regressing performance during high-frequency updates.

### DIFF
--- a/mesa/experimental/mesa_signals/mesa_signal.py
+++ b/mesa/experimental/mesa_signals/mesa_signal.py
@@ -18,6 +18,7 @@ clean separation of concerns.
 
 from __future__ import annotations
 
+import contextlib
 import functools
 import weakref
 from abc import ABC, abstractmethod
@@ -287,6 +288,11 @@ class HasObservables:
         self.subscribers = defaultdict(list)
         self.observables = dict(descriptor_generator(self))
 
+        # New state for signal management
+        self._silenced = False
+        self._batching = False
+        self._signal_buffer = []
+
     def _register_signal_emitter(self, name: str, signal_types: set[str]):
         """Helper function to register an Observable.
 
@@ -449,6 +455,9 @@ class HasObservables:
             kwargs: additional keyword arguments to include in the signal
 
         """
+        if self._silenced:
+            return
+
         signal = Message(
             name=observable,
             old=old_value,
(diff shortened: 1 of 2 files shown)


mesa/experimental/mesa_signals/mesa_signal.py
```diff
@@ -18,6 +18,7 @@
 
 from __future__ import annotations
 
+import contextlib
```
falloficaruss, line 21: This is significant and should be documented in the PR description.

quaquel:
as indicated in #3227:

> mesa_signals is under active development, so PRs at this stage on any of these issues that are not in sync with the development I am doing, e.g., https://github.com/mesa/mesa/pull/3212, will not be accepted.

Please hold off until 3212 has been merged.

PR #3234, opened 2026-01-31
fix: update Altair tooltip type inference for compatibility

### LINKED ISSUE #3233: Altair visualization fails with Altair v5+ due to removed `infer_vegalite_type_for_pandas`
**Describe the bug**

The Altair-based visualization component uses
`alt.utils.infer_vegalite_type_for_pandas`, which was removed in Altair v5+.
As a result, visualization-related tests fail with an `AttributeError` on
current Altair versions.

**Expected behavior**

Altair-based visualizations and related tests should work correctly with
current Altair versions (v5+), without raising errors due to deprecated or
removed APIs.

**To Reproduce**

1. Install Mesa with a recent version of Altair (v5+)
2. Run the visualization test suite: pytest tests/visualization/
3. The test `test_call_space_drawer_full` fails with: `AttributeError: module 'altair.utils' has no attribute 'infer_vegalite_type_for_pandas'`

This error originates from the Altair-based space visualization component when
constructing tooltips.

**Additional context**

This appears to be caused by dependency drift after upgrading to Altair v5+, where
  `alt.utils.infer_vegalite_type_for_pandas` was removed.
The affected code path is exercised by Altair-based space visualizations.

souro26:
Fixes #3233 

### Summary
Fixes a visualization failure with newer versions of Altair caused by the use of a removed utility function. This restores compatibility and allows visualization tests to pass on Altair v5+.

### Bug / Issue

The Altair-based space visualization used
`alt.utils.infer_vegalite_type_for_pandas`, which was removed in Altair v5+.
This caused visualization-related tests to fail with an `AttributeError`,
notably `test_call_space_drawer_full`.

### Implementation

- Removed the call to the deprecated `infer_vegalite_type_for_pandas` utility.
- Updated tooltip construction to explicitly specify the Vega-Lite type:
  - `quantitative` for numeric values
  - `nominal` for non-numeric values

This aligns with Altair v5+ requirements when data is not provided as a pandas DataFrame

### Testing

- Ran: pytest tests/visualization/test_solara_viz_legacy.py::test_call_space_drawer_full
- Ran: pytest tests/visualization/

All visualization tests pass after the change.

### Additional Notes

This change is limited to visualization components and is intended solely to
address dependency compatibility with newer Altair versions.


Sahil-Chhoker:
@souro26 thanks for the PR! Does the same API works for versions below Altair v5?

sanika-n:
From the looks of it I think @souro26 used pure python logic and is not using an Altair API to make the type inference right? So I guess it might work for all versions

souro26:
Yes, the change avoids relying on Altair utility APIs and instead does explicit type assignment in python. I also verified it locally against Altair v4, and it behaves the same there, so it should work for all versions i think.

PR #3218, opened 2026-01-28
docs: Add Mesa migration guide for AgentSet sequence behavior

codebyNJ:
This PR is regarding the migration guide for AgentSet addition.

PR #3235, opened 2026-01-31
fix: update alliance_formation example to use AgentSet.to_list()

souro26:
### Summary

Updates the `alliance_formation` example model to use the new `AgentSet.to_list()` method instead of deprecated direct indexing.

### Bug / Issue

The alliance_formation model uses deprecated `AgentSet.__getitem__` (e.g., `agents[0]`), which was deprecated in #3208 and will be removed in Mesa 4.0.
This triggers PendingDeprecationWarning when running the example, and since this is a reference example that users often copy, it should demonstrate current best practices.

### Implementation

Replaced direct indexing:
```python
agent_0 = agents[0]
agent_1 = agents[1]
```
With the recommended pattern from the migration guide (#3218):

```python
agent_list = agents.to_list()
agent_0 = agent_list[0]
agent_1 = agent_list[1]
```
### Testing

Ran the alliance_formation example test:
```bash
pytest tests/examples/test_examples.py::test_alliance_formation_model -v
```
### Additional Notes

Follow-up to #3208 and #3218.
This change eliminates deprecation warnings in the alliance_formation example model.

### DIFF
--- a/mesa/examples/advanced/alliance_formation/model.py
+++ b/mesa/examples/advanced/alliance_formation/model.py
@@ -3,7 +3,6 @@ import numpy as np
 
 import mesa
 from mesa import Agent
-from mesa.agent import AgentSet
 from mesa.examples.advanced.alliance_formation.agents import AllianceAgent
 from mesa.experimental.meta_agents.meta_agent import (
     create_meta_agent,
@@ -65,7 +64,10 @@ class MultiLevelAllianceModel(mesa.Model):
         """
         agent_0, agent_1 = agents
 
-        new_position = 1 - (max(agent_0.position, agent_1.position) - min(agent_0.position, agent_1.position))
+        new_position = 1 - (
+            max(agent_0.position, agent_1.position)
+            - min(agent_0.position, agent_1.position)
+        )
         potential_utility = (agent_0.power + agent_1.power) * 1.2 * new_position
 
         value_0 = 0.5 * agent_0.power + 0.5 * (potential_utility - agent_1.power)


mesa/examples/advanced/alliance_formation/model.py
```diff
@@ -62,8 +62,9 @@ def calculate_shapley_value(self, agents):
         Returns:
             tuple: Potential utility, new position, and level.
         """
-        agent_0 = agents[0]
-        agent_1 = agents[1]
+        agent_list = agents.to_list()
+        agent_0 = agent_list[0]
+        agent_1 = agent_list[1]
```
quaquel: I suggest fixing this `find_combinations` instead.

souro26: Thanks for the review, i'll update the fix to address this in the `find_combinations` function directly

souro26:
Thanks for the quick review @quaquel.

I double-checked `find_combinations`, and it doesn’t use deprecated indexing. The deprecated access was in `calculate_shapley_value`, which is now resolved by converting the `AgentSet` once via `to_list()`. I avoided changing `find_combinations` since converting the `AgentSet` there would alter its iteration semantics. Please let me know if you’d prefer a different approach.

quaquel:
That is not entirely true. Look at the typing of `find_combinations`. `group` is an `AgentSet`. This should be changed to a list, or, looking at the code, an iterable. Next, `evaluate_combination` explicitly puts the group (which is typed as a tuple) into an agent set. This line should be removed. Also, the typing of `evaluation_func` should be updated across all of meta_agents.py.

souro26:
Thanks for the clarification — I’ve updated the implementation accordingly.

Changes made:

- Updated `evaluation_func` typing across meta_agents.py to operate on tuple[Agent, ...]

- Removed internal `AgentSet` construction from `evaluate_combination`

- Updated `find_combinations` to accept `AgentSet | Iterable`, converting once for iteration

- Updated the `alliance_formation` example to explicitly convert the tuple to `AgentSet` only where AgentSet-specific methods are required

I’ve also noticed a couple of now-redundant lines in `calculate_shapley_value` (duplicate agent extraction / AgentSet usage) and can clean those up if you’d like me to push a small follow-up commit.

All relevant tests are passing:

- `tests/experimental/test_meta_agents.py`

- `tests/examples/test_examples.py::test_alliance_formation_model`

I'm happy to adjust further based on your preference.

mesa/experimental/meta_agents/meta_agent.py
```diff
@@ -49,61 +49,68 @@
...
-        return group_set, value
+        value = evaluation_func(candidate_group)
+        return candidate_group, value
     return None
 
 
 def find_combinations(
     model,
-    group: AgentSet,
+    group: AgentSet | list | Iterable,
```
quaquel:
```suggestion
    group: Iterable,
```

iterable is enough here. AgentSet and list are already iterable so no need to add them here.

mesa/experimental/meta_agents/meta_agent.py
```diff
@@ -49,61 +49,68 @@
...
+    ]
     | None = None,
-) -> list[tuple[AgentSet, float]]:
+) -> list[tuple[tuple[Agent, ...], float]]:
     """Find valuable combinations of agents in this set.
 
     Args:
         model: The model instance.
-        group (AgentSet): The set of agents to find combinations in.
+        group (AgentSet | list | Iterable): The set of agents to find combinations in.
```
quaquel: please remove all typing from docstring. Should not be there if typing is used.

mesa/experimental/meta_agents/meta_agent.py
```diff
@@ -49,61 +49,68 @@
...
+        List[Tuple[Tuple[Agent, ...], float]]: The list of valuable combinations, in
         a tuple first agentset of valuable combination  and then the value of
         the combination.
     """
+    # Convert AgentSet to list for iteration
+    if isinstance(group, AgentSet):
+        agent_list = group.to_list()
+    else:
+        agent_list = list(group) if not isinstance(group, list) else group
+
```
quaquel: not needed at all.

mesa/examples/advanced/alliance_formation/model.py
```diff
@@ -95,7 +97,8 @@ def only_best_combination(self, combinations):
         best = {}
         # Determine best option for EACH agent
         for group, value in combinations:
-            agent_ids = sorted(group.get("unique_id"))  # by default is bilateral
+            group_set = AgentSet(group, random=self.random)
+            agent_ids = sorted(group_set.get("unique_id"))  # by default is bilateral
```
quaquel: you don't need an agentset here. You can just use operator.itemgetter instead as key in sorted.

quaquel:
> I’ve also noticed a couple of now-redundant lines in calculate_shapley_value (duplicate agent extraction / AgentSet usage) and can clean those up if you’d like me to push a small follow-up commit.

Please do.

mesa/examples/advanced/alliance_formation/model.py
```diff
@@ -57,17 +58,17 @@ def calculate_shapley_value(self, agents):
...
-            agents (list): List of agents.
+            agents: Pair of agents.
 
         Returns:
             tuple: Potential utility, new position, and level.
         """
-        agent_0 = agents[0]
-        agent_1 = agents[1]
+        agent_0, agent_1 = agents
+        agent_set = AgentSet(agents, random=self.random)
```
quaquel: this should not be needed at all.

mesa/examples/advanced/alliance_formation/model.py
```diff
@@ -57,17 +58,17 @@ def calculate_shapley_value(self, agents):
...
 
-        positions = agents.get("position")
+        positions = agent_set.get("position")
         new_position = 1 - (max(positions) - min(positions))
-        potential_utility = agents.agg("power", sum) * 1.2 * new_position
+        potential_utility = agent_set.agg("power", sum) * 1.2 * new_position
```
quaquel: You only have 2 agents, so why use an aggregation here? Just write it up explicitly for 2 agents.

souro26:
Thanks for the feedback, it makes sense I'll simplify it to use the attributes directly since it's just 2 agents.

quaquel (approved):

PR #3241, opened 2026-02-04
fix: update meta_agents extract_class to use to_list()

### LINKED ISSUE #3240: Deprecated AgentSet indexing in meta_agents.extract_class function
**Describe the bug**

The `extract_class` function in `mesa/experimental/meta_agents/meta_agent.py` (line 133) still uses deprecated `AgentSet.__getitem__` syntax, which was deprecated in #3208 and will be removed in Mesa 4.0.

### Current Code (Line 133)
```python
return type(agents_by_type[agent_type_names[new_agent_class]][0])
```

**Expected behavior**

Should use `AgentSet.to_list()[index]` as recommended in the migration guide (#3218).

### Expected Code
```python
return type(agents_by_type[agent_type_names[new_agent_class]].to_list()[0])
```

**To Reproduce**

The issue can be observed by examining the code at line 133 of `mesa/experimental/meta_agents/meta_agent.py`:

```python
return type(agents_by_type[agent_type_names[new_agent_class]][0])
```

This line uses `AgentSet[0]` syntax, which is deprecated. While it currently works because `model.agents_by_type` returns `_HardKeyAgentSet` instances (which don't emit the warning), this code violates the Mesa 4.0 migration guidelines and should be updated to use the recommended `to_list()` method for consistency and future compatibility.

The function is called internally when creating meta-agents:

```python
from mesa import Model, Agent

class CustomAgent(Agent):
    pass

model = Model()
agents = [CustomAgent(model) for _ in range(3)]

# First call creates the MetaAgentClass
meta1 = create_meta_agent(model, "MetaAgentClass", agents[:2], Agent)

# Second call uses extract_class to find existing MetaAgentClass
# This is where line 133 executes with the deprecated syntax
meta2 = create_meta_agent(model, "MetaAgentClass", agents[2:], Agent)
```

**Additional context**

- This issue was partially addressed in PR #3235, which fixed similar deprecation warnings in the `evaluate_combination` and `find_combinations` functions within the same file
- However, the `extract_class` function on line 133 was missed and still uses the deprecated syntax
- This needs to be fixed to ensure compatibility with Mesa 4.0 where `AgentSet.__getitem__` will be removed
- The fix is straightforward: replace `agents_by_type[...][0]` with `agents_by_type[...].to_list()[0]`

**I am working on this issue and will submit a PR to fix it.**

Jayantparashar10:
### Summary

Updates the `extract_class` function in `mesa/experimental/meta_agents/meta_agent.py` to use `AgentSet.to_list()[index]` instead of deprecated `AgentSet[index]` syntax.

### Bug / Issue

Fixes #3240

The `extract_class` function (line 133) uses deprecated `AgentSet.__getitem__` syntax, which was deprecated in #3208 and will be removed in Mesa 4.0. This violates Mesa 4.0 migration guidelines established in #3218.

While the code currently works (since `model.agents_by_type` returns `_HardKeyAgentSet` instances), it should follow the recommended `to_list()` pattern for consistency and future compatibility.

### Implementation

Changed line 133 in `mesa/experimental/meta_agents/meta_agent.py`:

```python
# Before
return type(agents_by_type[agent_type_names[new_agent_class]][0])

# After
return type(agents_by_type[agent_type_names[new_agent_class]].to_list()[0])
```

### Testing

Ran all meta_agents tests to verify the fix:
```bash
pytest tests/experimental/test_meta_agents.py -v
# Result: 16 passed in 0.71s
```

All tests pass successfully with the updated code.

### Additional Notes

- This is a follow-up to PR #3235, which fixed similar deprecation issues in the `evaluate_combination` and `find_combinations` functions within the same file, but missed the `extract_class` function
- Ensures consistency with Mesa 4.0 API migration guidelines
- No breaking changes or side effects

### DIFF
--- a/mesa/experimental/meta_agents/meta_agent.py
+++ b/mesa/experimental/meta_agents/meta_agent.py
@@ -130,7 +130,7 @@ def extract_class(agents_by_type: dict, new_agent_class: object) -> type[Agent] 
         agent_type_names[agent.__name__] = agent
 
     if new_agent_class in agent_type_names:
-        return type(agents_by_type[agent_type_names[new_agent_class]].to_list()[0])
+        return type(next(iter(agents_by_type[agent_type_names[new_agent_class]])))
     return None
 
 


mesa/experimental/meta_agents/meta_agent.py
```diff
@@ -130,7 +130,7 @@ def extract_class(agents_by_type: dict, new_agent_class: object) -> type[Agent]
         agent_type_names[agent.__name__] = agent
 
     if new_agent_class in agent_type_names:
-        return type(agents_by_type[agent_type_names[new_agent_class]][0])
+        return type(agents_by_type[agent_type_names[new_agent_class]].to_list()[0])
```
quaquel: the aim is to get the class, so you can better use the iter protocal here:

next(iter(agents_by_type[agent_type_names[new_agent_class]]))

Jayantparashar10: @quaquel Thank you for the feedback! You're absolutely right - using the iterator protocol is more efficient. I've updated the code to use `next(iter(...))` instead of `.to_list()[0]`.

All tests still pass:
```bash
pytest [test_meta_agents.py](http://_vscodecontentref_/2) -v 
```
### 16 passed in 0.72s

quaquel (approved):

PR #3244, opened 2026-02-05
Fix Solara Altair dependency updates

### LINKED ISSUE #3243: Altair SolaraViz ignores dependencies / update_counter, causing stale renders
## Description
`SpaceRendererComponent` accepts a `dependencies` parameter but only wires it into `Matplotlib` rendering. The Altair path does not use `dependencies` (or `update_counter`) at all, so updates that should trigger a redraw can be missed, leading to stale charts.

There’s a `# FIXME: Manage dependencies properly` comment at the `SpaceRendererComponent` signature, which appears to refer to this gap.

## Steps to Reproduce

1. Create a model that updates visible state each step.
2. Use Altair backend in `SolaraViz`.
3. Trigger updates via `update_counter` or by passing dependencies.
4. Observe that the Altair chart does not re-render consistently.

## Expected
Altair renders should respond to the same dependency signals as `Matplotlib` renders, including `update_counter` and any explicit dependencies passed into `SpaceRendererComponent`.

## Actual
Altair rendering ignores `dependencies`, so updates may not trigger a re-render.

## Proposed Fix
Compute `viz_dependencies = [update_counter.value] + dependencies` for both backends, and use these dependencies to drive Altair chart recomputation (e.g., via `solara.use_memo` or a supported dependencies argument to `FigureAltair`).

## Possible Test
Add a test that renders `SolaraViz` with Altair, triggers a dependency change, and asserts that the draw methods are called again or that a new chart is produced.

falloficaruss:
### Summary
Refactors dependency handling in `SpaceRendererComponent` so `viz_dependencies` is initialized once and shared across backends.

### Bug / Issue
`viz_dependencies` was built inside the Matplotlib branch only. This made dependency handling backend-specific and inconsistent in `SpaceRendererComponent`.
This PR keeps behavior unchanged while moving dependency setup to a common location.

### Implementation
- Moved:
    - `viz_dependencies = [update_counter.value]`
    - `if dependencies`: `viz_dependencies.extend(dependencies)`
- This now runs before backend branching and is used by Matplotlib exactly as before.
- No Altair rendering logic changes.

### Testing
`test_solara_viz.py`

### Additional Notes
Closes #3243

### DIFF
--- a/mesa/batchrunner.py
+++ b/mesa/batchrunner.py
@@ -84,7 +84,8 @@ def batch_run(
         )
     if iterations is not None:
         warnings.warn(
-            "The `iterations` keyword argument is deprecated, please use `rng` instead."
+            "The `iterations` keyword argument is deprecated. "
+            "Use `rng` instead (e.g. `iterations=5` is equivalent to `rng=[None] * 5`). "
             "See https://mesa.readthedocs.io/latest/migration_guide.html#batch-run",
             DeprecationWarning,
             stacklevel=2,
--- a/mesa/experimental/devs/eventlist.py
+++ b/mesa/experimental/devs/eventlist.py
@@ -22,6 +22,7 @@ from __future__ import annotations
 
 import itertools
 from collections.abc import Callable
+from dataclasses import dataclass
 from enum import IntEnum
 from heapq import heapify, heappop, heappush, nsmallest
 from types import MethodType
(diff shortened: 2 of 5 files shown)


Sahil-Chhoker:
I do see the issue, but except this all the changes are unnecessary and out of the scope of this PR:
```python
viz_dependencies = [update_counter.value]
if dependencies:
    viz_dependencies.extend(dependencies)
```

Please remove the unnecessary changes so we can move forward.

falloficaruss:
@Sahil-Chhoker can you take a look now?

PR #3250, opened 2026-02-06
Add Schedule dataclass and refactor EventGenerator

EwoutH:
### Summary
This PR introduces a `Schedule` dataclass to centralize recurring event configuration and refactors `EventGenerator` to use it. The change simplifies the event-generation API, consolidates stop/start behavior, and improves clarity around recurring execution rules.

A new stable API for creating events and running will still be introduced in another PR, this PR just builds the foundation for that. See https://github.com/EwoutH/mesa/commit/440fdbed5fe51057b0535f5adfba0430e1a5d73e for an initial concept.

### Motive
Previously, `EventGenerator` handled interval logic, start timing, and stop conditions (`at`, `after`, `count`) internally. This created a more complex API surface and scattered scheduling logic across multiple methods.

### Implementation
A new `Schedule` dataclass was introduced to define recurring execution behavior (`interval`, `start`, `end`, and `count`). The `interval` field supports both fixed numeric values and callables evaluated against the model at runtime.

`EventGenerator` was refactored to depend on `Schedule` instead of storing interval and stop-condition state internally. The generator now reads execution limits directly from the schedule when determining whether to reschedule events. The `start()` method was simplified to use `schedule.start` or default to `model.time + interval`, and `stop()` now performs only immediate cancellation.

Tests were updated to reflect the new API, and a new test suite for `Schedule` was added. Existing behavior such as callable intervals, execution counting, and priority ordering remains covered.

### Usage Examples
```Python
# Basic recurring generator
gen = EventGenerator(model, fn, Schedule(interval=1.0, start=0.0))
gen.start()

# schedules
schedule = Schedule(interval=1.0, start=0.0, count=5)
schedule = Schedule(interval=1.0, start=0.0, end=10.0)
schedule = Schedule(interval=lambda m: m.time * 0.5, start=0.0)

# using schedule
EventGenerator(model, fn, schedule).start()
```

### Additional Notes
* The `Schedule` object can be used for other things too, like data collection schedules (see https://github.com/mesa/mesa/pull/3145#issuecomment-3860278930, cc @codebreaker32).
* This PR introduces a small API break in `EventGenerator` by replacing the `interval` argument with `Schedule`.
* Stop conditions are now declarative via `Schedule` instead of configured at runtime.
* The generator lifecycle is now simpler: `start()` activates scheduling, `stop()` cancels immediately.
* Test coverage for DEVS scheduling behavior has been expanded.

### DIFF
--- a/mesa/experimental/devs/eventlist.py
+++ b/mesa/experimental/devs/eventlist.py
@@ -149,7 +149,7 @@ class SimulationEvent:
             self.fn = None
 
 
-@dataclass
+@dataclass(frozen=True, slots=True)
 class Schedule:
     """Defines when something should happen repeatedly.
 


quaquel:
As indicated also in #3145, I am not convinced by this for data collection. But I agree that this might be cleaner to use over a long list of keyword arguments. I'll try to review later this weekend.

EwoutH:
Even if not useful for anything other than event scheduling, from this refactor a `Scheduling` object (of which instances can be reused) seems to clean up both source and user code nicely.

quaquel:
> Even if not useful for anything other than event scheduling, from this refactor a Scheduling object (of which instances can be reused) seems to clean up both source and user code nicely.

Similar to Scenario, so this is for me the prime reason why I still like this PR.

mesa/experimental/devs/eventlist.py
```diff
@@ -148,26 +149,41 @@ def __setstate__(self, state):
             self.fn = None
 
 
+@dataclass
```
quaquel: should this not be frozen and using slots?

quaquel (approved):
one clarifying question, but looks fine to me otherwise.

mesa/experimental/devs/eventlist.py
EwoutH: Fair point
```suggestion
@dataclass(frozen=True, slots=True)
```

EwoutH:
@quaquel conceptually, do you think `priority` should be part of the `Schedule`? Because it basically defines in which order you do things at a certain time, and thus is part of the ordering of events.

quaquel:
> @quaquel conceptually, do you think priority should be part of the Schedule? Because it basically defines in which order you do things at a certain time, and thus is part of the ordering of events.

I can see arguments for this but also against this. It depends a bit on how we wan tot use Schedule. At the moment, it is purely used for repeated scheduling of events on the event list. If you want to reuse a schedule object, having the priority frozen in it might be a bit restrictive. At the same time, your argument also holds and creating a Schedule instance is not a big deal, removing the need to extensively reuse them. So, I am fine either way, with a slight preference for including it.

EwoutH:
You could essentially do stuff like:
```Python
step_start = Schedule(interval=1.0, priority=10)
step_end = Schedule(interval=1.0, priority=0)
```

PR #3251, opened 2026-02-07
Add batch and suppress context managers for mesa_signals

falloficaruss:
### Summary
This PR adds two new context managers to `mesa_signals` via `HasObservables`:

- `batch_signals()`: buffers signals and flushes at context exit.
- `suppress_signals()`: suppresses signal emission while active.
It is designed to reduce signal storms during bulk updates while keeping current behavior unchanged outside the contexts.

### Motive
While making model behavior more reactive, frequent updates can emit many intermediate signals that are not useful to observers (especially during bulk state changes or data collection phases).
This PR introduces explicit signal-flow control so users can:

- coalesce noisy intermediate updates into final meaningful notifications (`batch_signals`)
- temporarily disable notifications during setup/mutation phases (`suppress_signals`)

### Implementation
Changes were made in `core.py` and tests in `test_mesa_signals.py`

Key implementation details:

- Added internal per-instance state in `HasObservables`:
    - `_batch_depth`
    - `_suppress_depth`
    - `_signal_buffer`
- Updated `notify()` to route through `_buffer_or_dispatch()`:
    - If suppress is active: drop signal.
    - Else if batch is active: buffer signal.
    - Else: dispatch immediately via `_mesa_notify()`.
- Added `_flush_signal_buffer()`:
    - For `ObservableSignals.CHANGED`, keep only the last signal per observable name in the batch.
    - Keep `non-CHANGED` signals in original order.
- Added context managers on `HasObservables`:
    - `batch_signals()` supports nesting and flushes only on outermost exit.
    - `suppress_signals()` supports nesting.
- Interaction rule:
    - suppress has precedence over batch.
### Usage Examples
```python
from mesa.experimental.mesa_signals import ObservableSignals

# batch updates: only last CHANGED is emitted for each observable
agent.observe("wealth", ObservableSignals.CHANGED, handler)

with agent.batch_signals():
    agent.wealth = 10
    agent.wealth = 20
    agent.wealth = 30
# handler called once with new=30
```

```python
# suppress all signals during initialization/setup
with model.suppress_signals():
    for a in model.agents:
        a.energy = 100
        a.wealth = 0
# no signals emitted inside the context
```

```python
# nested behavior
with obj.batch_signals():
    obj.value = 1
    with obj.suppress_signals():
        obj.value = 2   # suppressed
    obj.value = 3
# flush emits only final non-suppressed CHANGED signal (new=3)
```

### Additional Notes
- Scope intentionally limited to signal flow control; no changes to computed-property dependency tracking in this PR.
- Tests added for:
    - CHANGED deduplication in batch
    - non-CHANGED ordering
    - nested batch/suppress behavior
    - batch+suppress interaction
    - buffer cleanup after flush
- Related Issue: #3227

### DIFF
--- a/mesa/experimental/mesa_signals/core.py
+++ b/mesa/experimental/mesa_signals/core.py
@@ -8,8 +8,6 @@ functionality:
 - computed_property: Decorator for creating properties that automatically update based on dependencies
 - HasObservables: Mixin class that enables an object to contain and manage observables
 - emit: Decorator for methods that emit signals
-- All: Helper class for subscribing to all signals from an observable
-- SignalType: Enum defining the types of signals that can be emitted
 
 The module implements a robust reactive system where changes to observable properties
 automatically trigger updates to dependent computed values and notify subscribed
@@ -85,7 +83,6 @@ class BaseObservable:
     def __set_name__(self, owner: HasObservables, name: str):  # noqa: D105
         self.public_name = name
         self.private_name = f"_{name}"
-        # owner.register_observable(self)
 
     def __set__(self, instance: HasObservables, value):  # noqa: D105
         # If no one is listening, Avoid overhead of fetching old value and
@@ -101,8 +98,8 @@ class BaseObservable:
         instance.notify(
             self.public_name,
             change_signal,
-            old=getattr(instance, self.private_name, self.fallback_value),
-            new=value,
+            old=value,
+            new=getattr(instance, self.private_name, self.fallback_value),
         )
 
     def __str__(self):  # noqa: D105
@@ -110,7 +107,11 @@ class BaseObservable:
 
 
 class Observable(BaseObservable):
-    """Observable class."""
+    """Observable descriptor.
+
+    An observable is an attribute that emits ObservableSignals.CHANGED whenever it is changed.
+
+    """
 
     # fixme: when we go to 3.13, we might explore changing this into a property
     #    instead of descriptor, which is likely to be more performant
@@ -123,13 +124,24 @@ class Observable(BaseObservable):
         ):
             raise ValueError(
                 f"cyclical dependency detected: Computed({CURRENT_COMPUTED.name}) tries to change "
-                f"{instance.__class__.__name__}.{self.public_name} while also being dependent it"
+                f"{instance.__class__.__name__}.{self.public_name} while also being dependent on it"
             )
 
-        super().__set__(instance, value)  # send the notify
+        send_notify = False
+        old_value = None
+        if instance._has_subscribers(self.public_name, ObservableSignals.CHANGED):
+            old_value = getattr(instance, self.private_name, None)
+            send_notify = True
         setattr(instance, self.private_name, value)
 
-        PROCESSING_SIGNALS.clear()  # we have notified our children, so we can clear this out
+        if send_notify:
+            instance.notify(
+                self.public_name,
+                ObservableSignals.CHANGED,
+                old=old_value,
+                new=value,
+            )
+            PROCESSING_SIGNALS.clear()  # we have notified our children, so we can clear this out
 
 
 class ComputedState:
@@ -363,7 +375,7 @@ class HasObservables:
     def clear_all_subscriptions(self, name: ObservableName):
         """Clears all subscriptions for the observable <name>.
 
-        if name is All, all subscriptions are removed
+        if name is ALL, all subscriptions are removed
 
         Args:
             name: name of the Observable to unsubscribe for all signal types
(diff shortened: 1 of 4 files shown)


quaquel:
Thanks for this PR. I am not sure about this PR. It makes a number of choices that  might have to be discussed first. For example, the context managers now work at the instance level rather than being global. What are the pro's and con's of this design? Also, I want to see a more elegant design for how signals are aggregated in the batch context manager. Most importantly, this must be easily user extendable for custom signals that a user might define.

mesa/experimental/mesa_signals/core.py
```diff
@@ -402,8 +406,58 @@ def notify(
...
+        """Flush buffered signals with CHANGED deduplication."""
+        if not self._signal_buffer:
+            return
+
+        last_changed_by_name: dict[str, int] = {}
+        for i, signal in enumerate(self._signal_buffer):
+            if signal.signal_type == ObservableSignals.CHANGED:
+                last_changed_by_name[signal.name] = i
+
+        changed_indexes = set(last_changed_by_name.values())
```
quaquel, line 464: we cannot use set here because it makes a model indeterminate.

falloficaruss: The `set` is only used for membership lookup (`i in changed_indexes`), not for iteration/order.
The output order still comes from iterating `self._signal_buffer`, so behavior remains deterministic.

falloficaruss:
> Thanks for this PR. I am not sure about this PR. It makes a number of choices that might have to be discussed first. For example, the context managers now work at the instance level rather than being global. What are the pro's and con's of this design? Also, I want to see a more elegant design for how signals are aggregated in the batch context manager. Most importantly, this must be easily user extendable for custom signals that a user might define.

I think the key design tradeoff is instance-local vs global batching/suppression.

### Instance-local context managers
Pros:

- Strong isolation: one object’s batching does not silently affect others.
- Better determinism in multi-model/experiment settings.
- Cleaner lifecycle/GC boundaries (state stays with the instance).
- Easier debugging because behavior is scoped and explicit.

Cons:

- Less convenient when users want to batch a whole graph of objects at once.
- Requires explicit coordination if multiple related objects should flush together.

### Global context managers
Pros:

- Very convenient “single switch” to batch/suppress everything.
- Easy for broad bulk operations.

Cons:

- Hidden coupling across otherwise unrelated objects.
- Higher risk of surprising behavior and order sensitivity.
- Harder to reason about with concurrent/parallel runs.
- More complex lifecycle management and leak risk.

My preference is instance-local as the safe default, and if needed we can add an explicit opt-in “global coordinator” later rather than making global behavior implicit by default.

I can make current signal aggregation more extensible by replacing hardcoded `CHANGED` handling with a pluggable aggregation policy API so custom/user-defined signals are first-class.

quaquel:
Thanks for the overview of pro's and cons. 

> Better determinism in multi-model/experiment settings.

This one makes no sense to me. When done well signals should always stay within a single model unless the user does something very strange

> I can make current signal aggregation more extensible by replacing hardcoded CHANGED handling with a pluggable aggregation policy API so custom/user-defined signals are first-class.

Too me this is critical, so I would like to see this developed at least at the API level before moving forward with this PR. For the record, I have been playing around with this and have some directions I might explore myself.

falloficaruss:
1. Made batch aggregation policy-driven and user-extensible in `core.py`:
- Added `BatchSignalAggregator` type alias.
- Added instance policy list: `self._batch_aggregators`.
- Reworked `_flush_signal_buffer()` to run configured aggregators in sequence.
- Added default built-in policy as API method: `HasObservables.aggregate_last_changed(...)`
- Added API methods:
    - `add_batch_aggregator(...)`
    - `clear_batch_aggregators(...)`
- Kept existing behavior by default via `self._batch_aggregators = [self.aggregate_last_changed]`.
2. Added `@emit` dependency tracking for `@computed_property` in `core.py`:
- Introduced `_EMIT_DEPENDENCY` sentinel.
- In emit(...) wrappers, when a computed is currently evaluating:
    - register dependency via `CURRENT_COMPUTED._add_parent(self, observable_name, _EMIT_DEPENDENCY)`
    - register processing signal.
- Updated computed change detection to treat `_EMIT_DEPENDENCY` as a signal-driven dependency that forces recomputation when dirty.
3. Added tests in `test_mesa_signals.py`:
- `test_custom_batch_aggregator_for_user_defined_signal`
- `test_computed_property_tracks_emit_dependency`

quaquel:
Thanks for the update. 

1. I am going to take a look at this myself as well. I am not yet convinced by the resulting API and want to explore some other ideas. 
2. Please revert the emit stuff to keep this PR focussed on just the context managers. Also, we are still discussing the dependency problem in the issue so putting in a PR for this is premature.

falloficaruss:
Closed this in favour of #3261

PR #3256, opened 2026-02-08
Fix a class-level signal registry to respect subclass shadowing

falloficaruss:
### Summary
Fixes a class-level signal registration bug in `mesa_signals` where subclass overrides were not fully respected during observable/emitter registry construction. This could expose stale inherited signal names as subscribable on subclasses.

### Additional Notes


quaquel:
I am closing this PR. Either make it atomic and independent of your batch surpress PR or if they are intimately linked address it there.

PR #3260, opened 2026-02-08
Refactor step scheduling to use `EventGenerator` internally

EwoutH:
_See the "Does Mesa need a "heartbeat"?" discussion in https://github.com/mesa/mesa/discussions/2921#discussioncomment-15591609._

### Summary
Replaces the hand-rolled step rescheduling loop in `Model` with an internal `EventGenerator`, simplifying the code while keeping all behavior and API identical.

### Motive
The current step scheduling uses a manual pattern: `_do_step()` calls `_schedule_step()`, which creates a `SimulationEvent` that calls `_do_step()` again. This is functionally an `EventGenerator` — but reimplemented ad-hoc. Now that `EventGenerator` exists (#3201), we should use it.

This also removes `_schedule_step()` entirely and simplifies both `_do_step()` and `_wrapped_step()`, reducing the amount of scheduling logic in `Model`.

### Implementation
The self-rescheduling loop (`_do_step()` → `_schedule_step()` → `SimulationEvent` → `_do_step()`) is replaced by a single `_default_schedule` `EventGenerator` created and started in `Model.__init__()`. This simplifies `_wrapped_step()` to just `self._advance_time(self.time + 1)`, reduces `_do_step()` to incrementing steps and calling the user's step method, and removes `_schedule_step()` entirely.

On the simulator side, `Simulator.setup()` now checks `model._simulator is not None` instead of `event_list.is_empty()` (since the event list has events from `_default_schedule` at init), `ABMSimulator.setup()` no longer needs to manually schedule the first step, and `DEVSSimulator.setup()` explicitly stops `_default_schedule` before clearing the event list.

### Additional Notes
- No public API changes: this is purely an internal refactor
- All existing `model.step()`, `run_for()`, `run_until()` behavior is unchanged
- `_default_schedule` is private (`_`-prefixed). Exposing it as a public API with a property setter is left for a follow-up PR
- The `Simulator.setup()` guard change from checking `event_list.is_empty()` to `model._simulator is not None` is necessary because the event list now contains events from `_default_schedule` immediately after init

### DIFF
--- a/mesa/visualization/solara_viz.py
+++ b/mesa/visualization/solara_viz.py
@@ -151,7 +151,7 @@ def SolaraViz(
         display_components.insert(0, (create_space_component(renderer.value), 0))
 
     with solara.AppBar():
-        solara.AppBarTitle(name if name else model.value.__class__.__name__)
+        solara.AppBarTitle(name if name else type(model.value).__name__)
         solara.lab.ThemeToggle()
 
     with solara.Sidebar(), solara.Column():
@@ -532,9 +532,9 @@ def ModelController(
         visualization_pause_event.clear()
         _mesa_logger.log(
             10,
-            f"creating new {model.value.__class__} instance with {model_parameters.value}",
+            f"creating new {type(model.value)} instance with {model_parameters.value}",
         )
-        model.value = model.value = model.value.__class__(**model_parameters.value)
+        model.value = model.value = type(model.value)(**model_parameters.value)
         if renderer is not None:
             renderer.value = copy_renderer(renderer.value, model.value)
             force_update()
@@ -659,7 +659,7 @@ def SimulatorController(
         simulator.reset()
         visualization_pause_event.clear()
         pause_step_event.clear()
-        model.value = model.value = model.value.__class__(
+        model.value = model.value = type(model.value)(
             simulator=simulator, **model_parameters.value
         )
         if renderer is not None:
@@ -765,7 +765,7 @@ def ModelCreator(
     model_parameters = solara.use_reactive(model_parameters)
 
     solara.use_effect(
-        lambda: _check_model_params(model.value.__class__.__init__, user_params),
+        lambda: _check_model_params(type(model.value).__init__, user_params),
         [model.value],
     )
     user_adjust_params, fixed_params = split_model_params(user_params)
@@ -922,7 +922,7 @@ def make_initial_grid_layout(num_components):
 
 def copy_renderer(renderer: SpaceRenderer, model: Model):
     """Create a new renderer instance with the same configuration as the original."""
-    new_renderer = renderer.__class__(model=model, backend=renderer.backend)
+    new_renderer = type(renderer)(model=model, backend=renderer.backend)
 
     attributes_to_copy = [
         "agent_portrayal",


quaquel:
Thanks for this PR. I agree with the basic premise of using EvenGenerator for this, but I am not sure about the current implementation. I'll try to take a closer look asap.

> This is potential groundwork for a "heartbeat" concept: a default recurring event that Mesa components (data collection, visualization, stop conditions) can (by default) attach to. By exposing it as default_schedule, we make this pattern explicit and user-configurable without committing to a larger API yet.

With `model.time` being reactive, there is simply no need for a heartbeat. Anything that is triggered by time changing can just observe `model.time` and take if from there.

> Disable stepping for pure event-driven models:

I don't really like this I prefer opt-in over opt-out.

> Currently model.step() is overloaded to mean two things:
> 
> the user's model logic ("what happens each step")
> time advancement ("advance the clock by 1").

I agree wtih you on this completely and decoupling the two is the way to go. Time then becomes the purview of the evenlists and the run control. And this is also why I wanted time to be reactive.

EwoutH:
> I don't really like this I prefer opt-in over opt-out.

I think we almost go to a philosophical level of "are ticks core to ABM or were they just the easy implementation and everybody stuck with it".

Anyway, we can't change that in Mesa 3.x, but it's something to discuss for Mesa 4. In that case it just will be as easy as setting `default_schedule` to `None` by default.

But discussing the merrit of a `default_schedule` is important.

quaquel:
> But discussing the merrit of a default_schedule is important.

As indicated, I am not convinced by it, but without diving into the code I am also not sure how you could otherwise achieve the things that are clearly valuable here. Namely, replacing "the internal hand-rolled step rescheduling mechanism with an exposed `EventGenerator`", and more cleanly separating the step method from the time increment.

What I would like to see is an easy way for the user to specify that e.g. `model.step` is to be run on a given schedule with meaningful defaults (i.e., interval=1, start=0, stop=None, count=None). Where I guess I make a difference is that would prefer (even if only in MESA 4) to not make an assumption about this and leave it to the user to specify it. This was also behind the earlier idea of just adding a `@scheduled` to `model.step`.

EwoutH:
If we do see value in this refactor but not in exposing it (currently) to the public, we could of course make `default_schedule` private (`_default_schedule`).

@mesa/maintainers I would like some more opinions and perspectives if we need a default scheduled method ("heart beat") or not. You don't need to dive in to the code, just conceptually if you see use/need for it. See also: https://github.com/mesa/mesa/discussions/2921#discussioncomment-15591609.

mesa/model.py
```diff
@@ -169,8 +174,18 @@ def __init__(
...
-        # Wrap the user-defined step method
+        # Store user's step method and create the default schedule
         self._user_step = self.step
+
+        # The default recurring event: calls _do_step every 1.0 time units
+        # Users can modify, replace, or stop this via the property setter.
+        self.default_schedule = EventGenerator(
+            self,
+            self._do_step,
+            Schedule(interval=1.0, start=1.0),
```
quaquel, line 183: why start at 1 instead of at zero? zero is just the state of the system prior to any events I guess (so the initial condition)?

mesa/model.py
```diff
@@ -210,17 +222,11 @@ def _advance_time(self, until: float) -> None:
...
 
-    def _schedule_step(self, time: float) -> None:
-        """Schedule a step event at the given time."""
-        event = SimulationEvent(time, self._do_step, priority=Priority.HIGH)
-        self._event_list.add_event(event)
-
     def _do_step(self) -> None:
-        """Execute one step and schedule the next."""
+        """Execute one step. Rescheduling is handled by the EventGenerator."""
         self.steps += 1
```
quaquel, line 224: I assume we are going to deprecate and remove self.steps in Mesa 4?

quaquel:
I reviewed the code. I don't have major issues with this. I think it is helpful to separate mesa 3 from mesa 4 going forward. We currently have a lot of baggage with `wrapped_step` and `model.steps` etc. This makes finding an elegant backward compatible solution clearly non trivial. So, I am fine with this solution in Mesa 3, but would prefer to revisit this for mesa 4.

Also, coverage is a bit low, so I assume you'll update some tests after which I am fine with approving this.

mesa/model.py
EwoutH: Basically because it’s the current convention in Mesa. 0 is setup, 1 the first step. Original discussion: https://github.com/mesa/mesa/discussions/2229#discussioncomment-10394866

(thank our excellent migration guide to help me find this back).

quaquel: I recall that discussion, so yes this makes perfect sense.

I quick semi related question: do we allow `Schedule(start=0)`?  Or should start always be larger than 0?

EwoutH: In my view we explicitly allow that. The default is just to start after one `interval`, but that can be any non-negative number.

EwoutH:
I realized this PR was doing two things: An internal refactor and exposing a new user API surface. That doesn't make it atomic, and so, let's split the two.

I refactored this PR to only do the former. Then we can later discuss the latter.

Since we don't expose a new API or enable new behavior, no additional tests are needed.

quaquel:
> Since we don't expose a new API or enable new behavior, no additional tests are needed.

I don't agree with this claim. The relevant question is how the coverage has changed due to refactoring and ensuring that the refactored code is still properly covered. That is the case here judged by codecov, but it also makes me wonder how the percentage is so low. I guess because the number of lines changed is quite small.

EwoutH:
Fair point. I'm planning on fully revising the test suite once we have done the post branching to Mesa 4 cleanup.

PR #3261, opened 2026-02-08
Adding batch and suppress to mesa_signals

quaquel:
**Summary**
This PR adds a `batch()` and `suppress()` context manager to `HasObservables`, as well as an easy to use way of controlling how batched signals are to be aggregated, even for custom user defined signals. The resulting API is straightforward:

```python

with self.batch():
    # allsModelSignals.agent_added signals are aggregated
    MyAgent.create_agents(self, n=500, value=self.rng.random(500, ))

with self.suppress():
    # no ModelSignals.agent_added event will be emitted
    MyAgent.create_agents(self, n=500, value=self.rng.random(500, ))

@aggregate.register(MyCustomSignalType)
def _custom_aggregate(signal_type:SignalType, signals:list[Message], value:Any|None=None):
    # add custom aggregation logic for MyCustomSignalType
    ....
```

**`batch()`**
  Buffers all signals emitted within the context. On exit, signals are aggregated per observable name:
  - **Observable**: collapses multiple CHANGED signals into one (old from first, new from last). No signal if no net change.
  - **ObservableList**: collapses all list operations (append, insert, remove, replace, set) into a single SET signal with correct old/new lists. No signal if no net change.

Nesting is supported — inner batches merge into the outer batch, only the outermost dispatches.

**`suppress()`**
Drops all signals. Nesting is supported — only the outermost exit restores normal dispatch.

**Design choices**
  - Intercepts at notify, not _mesa_notify, so list signals are captured regardless of current subscriber state (important for aggregating e.g., APPENDED into SET)
  - Suppress takes priority over batch — suppress() inside batch() drops signals; batch() inside suppress() also drops signals
  - Exceptions discard the buffer — no signals dispatched on error
  - Computed properties may return stale values during a batch (documented trade-off)

**Comparison to #3251**
The PR's resulting API is basically the same as per #3251, but the implementation details are quite different. The key differences are

1. most logic is contained in batching.py and the helper classes `_BatchContext` and `_SuppressContext`. 
2. aggregation is handled by a function, not a class. I am relying on `functools.singledispatch` for registering custom aggregation handlers. The resulting API for custom signal aggregation is straightforward as shown above. Each of the arguments here is critical. `signal_type` is used by `functools.singledispatch` to pick the correct aggregate function. `signals` is the list of batched Message instances. `value`, if provided, is the value of the observable before the first signal. You can store a value via `_BatchContext.capture_original_value_once`. So this functionality is also available to the user to use with custom signals.

TODO
- [ ] add an aggregation function for ModelSignals.

@falloficarus22, I appreciate your original PR and would love to read your feedback on this alternative take.


falloficaruss:
I think this approach is better as an upstream baseline if the goal is semantic batching and first-class API.

mesa/experimental/mesa_signals/batching.py
```diff
@@ -0,0 +1,272 @@
...
+
+    # reverse engineer the original value of list before first signal
+    idx = kwargs.get("index")
+    match first.signal_type:
+        case ListSignals.APPENDED:
+            old = current[:idx]
+        case ListSignals.INSERTED:
+            old = current[:idx] + current[idx + 1 :]
+        case ListSignals.REMOVED:
+            old = [*current[:idx], kwargs["old"], *current[idx:]]
```
falloficaruss: `idx` should be normalized first when negative (and bounded) before slicing, otherwise reverse reconstruction is wrong for negative-index operations

`batching.py` reconstructs pre-batch list state using raw index, but negative indices are not normalized before reverse reconstruction. This makes emitted `ListSignals.SET` history wrong for operations like `pop()` / `pop(-1)` / `insert(-1, x)`

falloficarus22 (commented):
`aggregate` uses module-level `functools.singledispatch`, so registrations are global to the Python process. That means one test/plugin/library can affect aggregation behavior in unrelated code paths.
Not a blocker, but worth considering whether aggregation should be instance-scoped or at least have a reset/isolation mechanism for tests and integrations.

falloficaruss:
What I think works well in #3251 :

- per-instance, pluggable aggregation pipeline (`add_batch_aggregator`) gives very flexible composition
- linear signal-buffer model is simple to reason about for cross-signal ordering/policies

What I think is stronger in this PR:

- clearer user-facing API (`batch()` / `suppress()`)
- cleaner default semantics for observables (first old, last new, no-net-change suppression)
- signal-type-specific aggregation via `singledispatch` is a good long-term direction for maintainability and extension

Overall, I prefer this PR’s direction.

quaquel:
Thanks for the detailed review. This is very usefull. Below are responses to some key points.

> idx should be normalized first when negative (and bounded) before slicing, otherwise reverse reconstruction is wrong for negative-index operations

Good catch. I'll add tests and modify the code. 

> `aggregate` uses module-level functools.singledispatch, so registrations are global to the Python process. That means one test/plugin/library can affect aggregation behavior in unrelated code paths.

There is, unfortunately, no built-in way yet to clean the registration of singledispatch. However, it does use weakrefs internally so when a registered function goes out of scope it will be removed. So tests for example are unlikely to cause problems. I highly doubt this will become a problem. 

> The docstring warns about it, but it would be safer to have an explicit test proving expected behavior and a note in user-facing docs.

Ok, I'll add a test for this. 

> What I think works well in #3251 :
> per-instance, pluggable aggregation pipeline (add_batch_aggregator) gives very flexible composition

It is true that in #3251, you can handle the aggregation of signals differently across HasObservable instances. I am not sure whether I like that. In my view it can be highly confusing. In the design here, The aggregation policy is direclty tied to the specific SignalType and thus easier to understand and reason about. Conceptually, it makes more sense to me that the strategy for handling aggregation is determined by the SignalType. For example, ObservableSignals can just take the old value from the first signal and the new value from the last signal. If they are different, you emit otherwise you don't. For lists, I currently just aggregate all changes into a single SET signal. I am not entirely sure about this and at least it requires careful documentation. ModelSignalTypes are currently not aggregated, but I am thinking of adding an aggregation where agents that are added and subsequently removed or vice versa are filtered out. So you keep the unique agents that are added and that are removed. But yes designing aggregations at the SignalType level foregoes some flexibility which might or might not be a problem. 

> linear signal-buffer model is simple to reason about for cross-signal ordering/policies 

Not sure I fully understand what you mean with this. In this design, nested batching will just aggregate signals in the outermost batch so it will also be linear.

falloficaruss:
> There is, unfortunately, no built-in way yet to clean the registration of singledispatch. However, it does use weakrefs internally so when a registered function goes out of scope it will be removed. So tests for example are unlikely to cause problems. I highly doubt this will become a problem.
> 
I still think the global-registration concern is real. My understanding is that registered handlers are kept in the dispatcher registry (strongly), so they are not automatically removed just because a local function goes out of scope. So cross-test/plugin/process-wide contamination is still possible once something registers on `aggregate`. I’m fine treating this as non-blocking, but I’d suggest at least documenting this behavior clearly (and maybe adding a test/reset strategy for isolation).
 
> It is true that in #3251, you can handle the aggregation of signals differently across HasObservable instances. I am not sure whether I like that. In my view it can be highly confusing. In the design here, The aggregation policy is direclty tied to the specific SignalType and thus easier to understand and reason about. Conceptually, it makes more sense to me that the strategy for handling aggregation is determined by the SignalType. For example, ObservableSignals can just take the old value from the first signal and the new value from the last signal. If they are different, you emit otherwise you don't. For lists, I currently just aggregate all changes into a single SET signal. I am not entirely sure about this and at least it requires careful documentation. ModelSignalTypes are currently not aggregated, but I am thinking of adding an aggregation where agents that are added and subsequently removed or vice versa are filtered out. So you keep the unique agents that are added and that are removed. But yes designing aggregations at the SignalType level foregoes some flexibility which might or might not be a problem.
> 
In general, this SignalType-scoped policy is cleaner and easier to reason about for most users, and I agree that’s a strong advantage of this PR.

> Not sure I fully understand what you mean with this. In this design, nested batching will just aggregate signals in the outermost batch so it will also be linear.
>
 My earlier “linear buffer” point was specifically about expressiveness for cross-observable/global policies: in #3251 aggregators run over one ordered signal stream, while here buffering is grouped per observable before aggregation. I agree that nested batching remains outermost/linear in both designs; my point was about aggregation scope rather than nesting mechanics.

quaquel:
> I still think the global-registration concern is real. 

I digged into the code a bit more. You are right that the registry uses hard refs not weakrefs. I found one [hacky way](https://stackoverflow.com/questions/25951651/unregister-for-singledispatch) to still do it. So if this becomes a relevant concern, we can add support for this:  the base `aggregate` function has the register as a attribute, so you can easily clear this: `aggregate.registry.clear()`.

> My earlier “linear buffer” point was specifically about expressiveness for cross-observable/global policies: in #3251 aggregators run over one ordered signal stream, while here buffering is grouped per observable before aggregation.

Ok, I now get your point.

falloficaruss:
> I digged into the code a bit more. You are right that the registry uses hard refs not weakrefs. I found one [hacky way](https://stackoverflow.com/questions/25951651/unregister-for-singledispatch) to still do it. So if this becomes a relevant concern, we can add support for this: the base `aggregate` function has the register as a attribute, so you can easily clear this: `aggregate.registry.clear()`.
> 
Thanks, and agreed on the hard-ref point.

`aggregate.registry.clear()` won’t work because registry is a `mappingproxy` (read-only).
The StackOverflow workaround mutates private `singledispatch` closure state, which is possible but pretty fragile.
I’m okay treating this as non-blocking for now, but I still think we should document global-registration behavior clearly and consider an isolation strategy for tests/plugins.

mesa/experimental/mesa_signals/batching.py
quaquel: I fixed it inside SignalingList instead of in the aggregator. I added additional tests in both places. So everything should work fine now. Again good catch.

falloficaruss:
image

I verified one regression introduced by the negative-index normalization change in `SignalingList`:
`ObservableList` slice operations now raise `TypeError` because `__setitem__` / `__delitem__` compare `index < 0` unconditionally, but for slice ops `index` is a `slice`.

Repro:
```python
from mesa.experimental.mesa_signals import HasObservables, ObservableList

class O(HasObservables):
    items = ObservableList()
    def __init__(self):
        super().__init__()
        self.items = [1,2,3,4]

o = O()
o.items[1:3] = [9,9]   # TypeError on PR branch
del o.items[1:3]       # TypeError on PR branch
```

quaquel:
> I verified one regression introduced by the negative-index normalization change in SignalingList:
ObservableList slice operations now raise TypeError because __setitem__ / __delitem__ compare index < 0 unconditionally, but for slice ops index is a slice.

I did some more work on this. I explicitly handle slices now, including slices with negative numbers and updated various tests. I will give it a once over hopefully later today but we are getting there.

mesa/experimental/mesa_signals/observable_collections.py
```diff
@@ -74,30 +74,58 @@ def __init__(self, iterable: Iterable, owner: HasObservables, name: str):
...
+            # this resolves negative numbers in slice
+            index = slice(*index.indices(len(self.data)))
+            old_value = self.data[index]
+            self.data[index] = value
+            self.owner.notify(
+                self.name,
+                ListSignals.REPLACED,
+                index=index,
+                old=old_value,
+                new=list(value),
```
falloficaruss: Slice assignment emits wrong new payload for one-shot iterables (generators/iterators).
- In `SignalingList.__setitem__` slice path, code does `self.data[index] = value` and then `new=list(value)`.
- If `value` is a generator, it is consumed by assignment first, so emitted signal has `new=[]` even though list content changed correctly.

Repro:
```python
from unittest.mock import Mock

from mesa.experimental.mesa_signals import HasObservables, ListSignals, ObservableList


class O(HasObservables):
    items = ObservableList()

    def __init__(self):
        super().__init__()
        self.items = [1, 2, 3, 4]


o = O()
handler = Mock()
o.observe("items", ListSignals.REPLACED, handler)

g = (x for x in [8, 9])
o.items[1:3] = g

signal = handler.call_args.args[0]
print("actual list:", list(o.items))
print("signal old:", signal.additional_kwargs["old"])
print("signal new:", signal.additional_kwargs["new"])
```

quaquel: there are so many edge cases around signaling list that I am starting to wonder about changing the design a bit. A lot of the trouble stems from trying to reverse engineer old from the first signal. Should we actually just aggregate all signals into a single set with old and new? or can we capture old before the signal?

falloficaruss: I think reverse-engineering from first signal keeps creating edge-case debt so I would recommend:

- For `ListSignals` in batch, aggregate to a single `SET(old,new)`.
- Capture `old` before first mutation (once per list per batch), then read `new` at flush.
- Keep granular per-op list signals for non-batch mode.

This is much simpler and avoids reconstructing old state from operation metadata entirely.

quaquel: Unfortunately, it is very hard to capture `old`. The current implementation batches signals. Signals are emitted after the change. So, to capture it before the first change, a SignalingList has to detect that a context manager is active on its owning HasObservables instance, and set `old` in some way. I'll take a look to see how difficult it would be to implement this, and if it is a design that can be encapsulated so it can be used downstream if needed as well. 


A `SignallingList` knows its owner. So it can check for `self.owner._batchcontext is not None`. If this is true, we are in a batch context and we then only need to add old to some data structure on `_BatchContex`, and ensure that this is properly flushed to the outermost context manager on each exit.

mesa/experimental/mesa_signals/batching.py
```diff
@@ -0,0 +1,254 @@
...
+if TYPE_CHECKING:
+    from .core import HasObservables
+
+__all__ = ["aggregate"]
+
+
+@functools.singledispatch
+def aggregate(
+    signal_type: SignalType, signals: list[Message], value: Any = None
+) -> list[Message]:
```
falloficaruss, line 49: `aggregate` remains process-global via `singledispatch`, so custom registrations are global side effects. Non-blocking, but still worth documenting.

quaquel: I added a note on this to the docstring of batching.py.

quaquel:
> ObservableList writes directly to _BatchContext internals (ctx._captured_values) in observable_collections.py. A small API like ctx.capture_old_once(name, value) would reduce coupling and make future refactors safer.

I added a `capture_original_value_once` convenience method to `_BatchContext`

quaquel:
@falloficarus22 Thanks a lot for your original PR and your contributions to improving this alternative take. I think it came together nicely.

EwoutH:
@codebreaker32 could you give this a look?

PR #3266, opened 2026-02-09
Add public event scheduling and time advancement methods

EwoutH:
### Summary
Exposes event scheduling and time advancement methods directly on `Model`, giving users a public API for scheduling events and advancing simulation time without needing to interact with the `Simulator` class.

Basically, these methods allow us:
- to replace all existing behavior (in examples, tutorials, etc.)
- start deprecation all current scheduling and stepping APIs
- remove them in Mesa 4.

### Motive
Currently, users advance their models by calling `model.step()`, which is tightly coupled to the concept of a single discrete step. With the event-based architecture taking shape (#3201, #3204), we need public methods that let users think in terms of *time* rather than *steps*.

The key method here is `model.run_for(duration)`. For existing ABM users, `model.run_for(1)` is functionally equivalent to `model.step()`: it advances time by 1 unit, which triggers the scheduled step event. But unlike `step()`, it generalizes naturally: `run_for(0.5)` advances half a time unit, `run_for(100)` advances 100 units processing all scheduled events along the way. This makes `run_for` the foundation for both traditional ABMs and event-driven models.

This is also a stepping stone toward deprecating `model.step()` as the primary way to advance a model. Instead of calling a method that *is* a step, users call a method that *advances time*, and steps happen as scheduled events within that time. The mental model shifts from "execute step N" to "advance time, and whatever is scheduled will run."

### Implementation
Four new public methods on `Model`:
- **`schedule_event(function, *, at, after, priority)`**: schedule a one-off event at an absolute or relative time
- **`schedule_recurring(function, schedule, priority)`**: schedule a recurring event using a `Schedule` object, returns an `EventGenerator`
- **`run_for(duration)`**: advance time by the given duration, processing all events
- **`run_until(end_time)`**: advance time to an absolute point, processing all events

All four are thin wrappers around existing internal machinery (`_event_list`, `_advance_time`, `EventGenerator`).

`Model` also stores strong references to active `EventGenerator`s in `_event_generators` to prevent garbage collection — `SimulationEvent` wraps bound methods in `WeakMethod`, so without this a fire-and-forget `schedule_recurring()` call would silently stop working.

### Testing
Tests cover `run_for`, `run_until`, `schedule_event` (at/after/cancel/validation), and `schedule_recurring` (fixed interval/stop/count). Includes an explicit `gc.collect()` regression test for the `WeakMethod` lifetime issue.

### Usage Examples

**Drop-in replacement for `model.step()`:**
```python
# Old
for _ in range(100):
    model.step()

# New — functionally identical
for _ in range(100):
    model.run_for(1)

# Or simply
model.run_for(100)
```

**Schedule custom events alongside steps:**
```python
class MyModel(mesa.Model):
    def __init__(self):
        super().__init__()
        # Schedule a one-off event
        self.schedule_event(self.market_crash, at=50.0)

        # Schedule recurring data export every 10 time units
        self.schedule_recurring(
            self.export_data,
            Schedule(interval=10.0, start=10.0),
        )

    def step(self):
```

**Run to a specific time:**
```python
model = MyModel()
model.run_until(500.0)  # run until t=500, executing all scheduled events
```

### Additional Notes
- No breaking changes, `model.step()` continues to work exactly as before
- `run_for(1)` and `step()` produce identical results for standard ABMs, since both advance time by 1 unit and trigger the same scheduled step event
- `schedule_event` and `schedule_recurring` return their event/generator objects, allowing cancellation via `.cancel()` / `.stop()`
- These methods work both with and without a `Simulator` attached

One we agree on the API I will add tests. In separate PRs we can update examples and tutorials and deprecate old API surfaces.

### DIFF
--- a/tests/test_schedule_run.py
+++ b/tests/test_schedule_run.py
@@ -1,0 +1,197 @@
+"""Tests for Model public event scheduling and time advancement API."""
+# ruff: noqa: D101 D102 D107
+
+import gc
+
+import pytest
+
+from mesa import Agent, Model
+from mesa.experimental.devs.eventlist import Schedule
+
+
+class StepAgent(Agent):
+    def __init__(self, model):
+        super().__init__(model)
+        self.steps_taken = 0
+
+    def step(self):
+        self.steps_taken += 1
+
+
+class SimpleModel(Model):
+    def __init__(self, n=3):
+        super().__init__()
+        StepAgent.create_agents(self, n)
+
+    def step(self):
+        self.agents.shuffle_do("step")
+
+
+# --- run_for / run_until ---
+class TestRunFor:
+    def test_single_unit(self):
+        model = SimpleModel()
+        model.run_for(1)
+        assert model.steps == 1
+        assert model.time == 1.0
+
+    def test_multiple_units(self):
+        model = SimpleModel()
+        model.run_for(10)
+        assert model.steps == 10
+        assert model.time == 10.0
+
+    def test_agents_activated(self):
+        model = SimpleModel(n=5)
+        model.run_for(3)
+        for agent in model.agents:
+            assert agent.steps_taken == 3
+
+    def test_equivalent_to_step(self):
+        """run_for(1) should produce the same result as step()."""
+        m1, m2 = SimpleModel(n=3), SimpleModel(n=3)
+        for _ in range(5):
+            m1.step()
+            m2.run_for(1)
+        assert m1.steps == m2.steps == 5
+        assert m1.time == m2.time == 5.0
+
+    def test_sequential_calls(self):
+        model = SimpleModel()
+        model.run_for(5)
+        model.run_for(5)
+        assert model.steps == 10
+        assert model.time == 10.0
+
+
+class TestRunUntil:
+    def test_basic(self):
+        model = SimpleModel()
+        model.run_until(5.0)
+        assert model.steps == 5
+        assert model.time == 5.0
+
+    def test_already_past(self):
+        model = SimpleModel()
+        model.run_for(10)
+        model.run_until(5.0)  # already past t=5
+        assert model.steps == 10  # no additional steps
+
+    def test_sequential(self):
+        model = SimpleModel()
+        model.run_until(3.0)
+        model.run_until(7.0)
+        assert model.steps == 7
+        assert model.time == 7.0
+
+
+# --- schedule_event ---
+class TestScheduleEvent:
+    def test_at(self):
+        model = SimpleModel()
+        log = []
+
+        def fire():
+            log.append("fired")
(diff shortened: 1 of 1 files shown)


mesa/model.py
```diff
@@ -388,3 +394,70 @@ def remove_all_agents(self):
         # we need to wrap keys in a list to avoid a RunTimeError: dictionary changed size during iteration
         for agent in list(self._all_agents):
             agent.remove()
+
+    ### Event scheduling and time progression methods ###
+    def schedule_event(
```
quaquel, line 404: I can accept this design, although you know my preference to separate them into schedule_event_at and schedule_event_after.

EwoutH: The coin fall to your liking with `run_for` and `run_until` ;)

On a bit more serious note, I went for this way because `schedule_event(something, at=2.5, after=1)` can make some conceptual sense (in this example firing two events), while `run(for=1, untill=2.5)` is by definition mutually exclusive (this regardless from if we want to enable it or not).

But I'm happy to follow the majority here.

EwoutH:
Found and fixed a subtle bug while writing these tests: `schedule_recurring` returns an `EventGenerator`, but if the user doesn't save it (which is completely natural, fire-and-forget recurring events), it gets garbage collected. This happens because `SimulationEvent` wraps bound methods in `WeakMethod`, and once the generator is GC'd the weak reference dies silently.

Fix is two lines: `Model` stores generators in `self._event_generators`. Added a test that forces `gc.collect()` after a fire-and-forget call to prevent regression.

EwoutH:
@mesa/maintainers please critically review the API and functionality. This is an API we're going to use a lot!

EwoutH:
If no objections, I would like to merge this tomorrow and move forward.

mesa/model.py
jackiekazil: This is a tough one. I like schedule_event ever so slightly better, because it feels more self contained.

jackiekazil:
I like these updates.

jackiekazil:
I will add one thing -- I think we need a more formal process for proposing such changes -- like a write up. 
This worked out well, but I don't think everyone tracks all progress all the time -- a formal process would allow for a period of feedback.

EwoutH:
@jackiekazil we do have a process: https://github.com/mesa/mesa/blob/main/CONTRIBUTING.md#mesa-development-process (#3135).

In this case there was:
- extensive, structured discussion: https://github.com/mesa/mesa/discussions/2921
- Pathfinding PRs: #3152, #3155, #3188
- A tracking issue: https://github.com/mesa/mesa/issues/3132

I think the current process was followed quite well. Of course, we could sharpen it. What do you find lacking or how would you improve it?

Issue #3247, opened 2026-02-06
Allow user-defined selection logic when multiple existing meta-agents are eligible


souro26:
## Summary
Add optional `meta_agent_selector` parameter to `create_meta_agent` to let users control which meta-agent receives new agents when multiple candidates of the same class exist.

## Current Behavior
When multiple meta-agents of the same class already exist, `create_meta_agent` selects one deterministically by lowest `unique_id` (lines 268-275):
````python
# TODO: Add way for user to specify how agents join meta-agent
# instead of random choice if there are multiple meta-agents of the same class
meta_agent = (
    sorted(existing_meta_agents, key=lambda x: x.unique_id)[0]
    if len(existing_meta_agents) > 1
    else existing_meta_agents[0]
)
````
## Proposed Solution
Add an optional callback parameter to `create_meta_agent`:
````python
meta_agent_selector: Callable[[list[MetaAgent], list[Agent]], MetaAgent] | None = None
````
If provided, it receives the candidate meta-agents and agents being added, and returns which meta-agent should receive them.
## Implementation
Replace the current selection logic with:
````python
if meta_agent_selector:
    meta_agent = meta_agent_selector(existing_meta_agents, agents)
elif len(existing_meta_agents) > 1:
    meta_agent = sorted(existing_meta_agents, key=lambda x: x.unique_id)[0]
else:
    meta_agent = existing_meta_agents[0]
````
## Example Usage
````python
# Select meta-agent with most members
def select_largest(candidates, new_agents):
    return max(candidates, key=len)

meta = create_meta_agent(
    model, "Team", [agent1, agent2], Agent,
    meta_agent_selector=select_largest
)

# Select based on aggregate power
def select_strongest(candidates, new_agents):
    return max(candidates, key=lambda ma: sum(a.power for a in ma.agents))

meta = create_meta_agent(
    model, "Alliance", [agent3, agent4], Agent,
    meta_agent_selector=select_strongest
)
````
## Benefits

- Fully backward compatible (default behavior unchanged)
- Clean extension point for model-specific selection logic
- No additional policy baked into Mesa

Would this be a useful addition to `meta_agents`?

Krishsharma179:
Hey @souro26,

[...]

PR #3267, opened 2026-02-09
Allow user-defined selection when multiple meta-agents exist

souro26:
### Summary
This PR adds an optional selection hook to `create_meta_agent` that allows users to control how an existing meta-agent is chosen when multiple eligible meta-agents of the same class already exist. The default behavior remains unchanged unless the selector is explicitly provided.

This addresses the design limitation discussed in #3247 .

### Motive
Currently, when multiple meta-agents of the same class exist, `create_meta_agent` deterministically selects the one with the lowest `unique_id`. While predictable, this choice is arbitrary and cannot express common modeling needs such as joining the largest existing group, selecting the  most-fit meta-agent and applying domain-specific aggregation logic. The lack of an extension point forces users to reimplement or work around meta-agent creation logic externally. This PR introduces a fully opt-in extension point that allows users to supply their own resolution logic without changing any default behavior.

### Implementation

Added an optional `meta_agent_selector` parameter to `create_meta_agent`. The selector is invoked only when there are more than one existing meta-agent of the same class is eligible.

The selector receives:
- The list of candidate meta-agents
- The agents being added

The selector must return exactly one of the candidate meta-agents

Error handling is explicit:
- If the selector raises an exception, a `ValueError` is raised
- If the selector returns an invalid meta-agent, a `ValueError` is raised

Fallback behavior:
- If no selector is provided, the existing deterministic behavior (lowest `unique_id`) is preserved

No existing code paths are modified unless the selector is explicitly passed.

### Usage Examples

````python
# Select the meta-agent with the most members
def select_largest(candidates, new_agents):
    return max(candidates, key=len)

meta = create_meta_agent(
    model,
    "Team",
    agents,
    Agent,
    meta_agent_selector=select_largest,
)
````
````python
# Select based on domain-specific logic
def select_strongest(candidates, new_agents):
    return max(
        candidates,
        key=lambda ma: sum(a.power for a in ma.agents),
    )
````
### Additional Notes

- It is fully backward compatible
- No behavioral changes unless `meta_agent_selector` is provided

I can add tests in a follow up commit if the semantics and failure-mode expectations introduced here are okay. 
Related discussion: #3247


Krishsharma179:
hii @souro26  see the codecov/patch failure (54.54%). Since you mentioned tests could come in a follow-up commit, I'd like to contribute a test file covering: default behavior (lowest unique_id), custom selectors (largest group/strongest), argument validation, and error cases. Would this help unblock the PR?

souro26:
Yes i noticed that the codecov failed, i was waiting for some insight into my design before i pushed the tests in a follow up commit. I have the tests ready so I'll push them right away.

souro26:
Added comprehensive tests for the selector semantics.

souro26:
hi @tpike3 and @EwoutH, please let me know if you would like any changes made to this PR from my side, thanks.

quaquel:
@tpike3, I guess this can be closed given the overhaul of meta agents?

PR #3268, opened 2026-02-09
Distinguish Logical Index from Physical Position

codebreaker32:
### Summary

This PR adds the ability to convert between logical coordinates (cell indices) and physical positions (spatial coordinates) across all discrete space types. This lays the foundation for stacked spaces as discussed in https://github.com/mesa/mesa/discussions/2585#discussioncomment-15715409

### Changes

1. **`Cell.position`**: A dedicated property for physical coordinates, distinct from the logical `coordinate`
2. **`find_nearest_cell(position)`**: High-performance lookup of the cell containing or nearest to a physical position(using KD-Trees for Voronoi, Network and HexGrid).
3. **HexGrid, VoronoiGrid & Network**: Implemented `scipy.spatial.KDTree` for  spatial lookups, to improve performance over brute-force distance checks.

### Breaking Changes

**VoronoiGrid only**: `cell.coordinate` is now an integer index instead of a float tuple.
**Migration:**

```python
# Before
centroid = cell.coordinate  # Was (55.2, 12.1)

# After  
centroid = cell.position    # np.array([55.2, 12.1])
cell_id = cell.coordinate   # 5 (Integer index)

```

This makes `VoronoiGrid` consistent with other spaces, improves performance by using integer keys for dictionary lookups, and decouples physical location from logical identity.

### DIFF
--- a/mesa/discrete_space/cell.py
+++ b/mesa/discrete_space/cell.py
@@ -33,7 +33,7 @@ class Cell:
     """The cell represents a position in a discrete space.
 
     Attributes:
-        coordinate (Tuple[int, int]) : the logical position(or index) of the cell in the discrete space
+        coordinate (Coordinate) : the logical position(or index) of the cell in the discrete space
         position (np.ndarray | None): the physical position of the cell in the discrete space
         agents (List[Agent]): the agents occupying the cell
         capacity (int): the maximum number of agents that can simultaneously occupy the cell
--- a/mesa/discrete_space/network.py
+++ b/mesa/discrete_space/network.py
@@ -91,7 +91,7 @@ class Network(DiscreteSpace[Cell]):
     def _rebuild_kdtree(self) -> None:
         """Rebuild the KD-Tree."""
         if self._kdtree_cells:
-            positions = np.array([c.position for c in self._kdtree_cells])
+            positions = np.array([c._position for c in self._kdtree_cells])
             self._kdtree = KDTree(positions)
         else:
             self._kdtree = None
--- a/mesa/discrete_space/voronoi.py
+++ b/mesa/discrete_space/voronoi.py
@@ -208,11 +208,8 @@ class VoronoiGrid(DiscreteSpace):
         self.centroids_coordinates = centroids_coordinates
         self._validate_parameters()
 
-        # Store centroids as numpy array for KD-tree
-        self.centroids = np.array(centroids_coordinates)
-
         # Build KD-tree for fast nearest-centroid lookup
-        self._kdtree = KDTree(np.array(self.centroids_coordinates))
+        self._kdtree = KDTree(np.array(centroids_coordinates))
 
         # Create cells with Integer coordinates and physical positions
         # coordinate is now an integer index
@@ -221,9 +218,9 @@ class VoronoiGrid(DiscreteSpace):
                 coordinate=i,  # Integer index
                 capacity=capacity,
                 random=self.random,
-                position=self.centroids[i],  # Physical centroid position
+                position=self.centroids_coordinates[i],  # Physical centroid position
             )
-            for i in range(len(self.centroids))
+            for i in range(len(self.centroids_coordinates))
         }
 
         self.regions = None


mesa/discrete_space/discrete_space.py
```diff
@@ -85,6 +87,54 @@ def agents(self) -> AgentSet:
...
+
+        Raises:
+            NotImplementedError: If space type doesn't support position lookup
+        """
+        raise NotImplementedError(
+            f"{self.__class__.__name__} does not implement pos_to_cell(). "
+            "This space may be purely topological (no physical coordinates)."
+        )
+
+    def cell_to_pos(self, cell: T) -> np.ndarray | None:
```
quaquel: why did you implement these two methods on the space instead of the cell?

mesa/discrete_space/network.py
```diff
@@ -56,6 +70,53 @@ def _connect_single_cell(self, cell: Cell):
...
+        Args:
+            position: Physical coordinates [x, y]
+
+        Returns:
+            Cell: The node closest to the position
+
+        Raises:
+            ValueError: If network is not spatial
+        """
+        if not self.spatial:
```
quaquel: I would suggest using the networkx layout algorithm here instead. You have the network object itself already, so it's trivial to use that here. The key challenge is to make it easy for users to control the layout or supply their own if so desired.

mesa/discrete_space/discrete_space.py
codebreaker32: A Cell usually only knows its own coordinate and its contents. It does not know about the boundaries of the world, whether the world wraps around (torus), or the dimensions of the grid.
To convert a physical position (x, y) to a cell, you need to know the Space Dimensions.

Looking at the logic you will see, It relies heavily on self.torus and self.dimensions, which exist on the Grid, not the Cell.

```python
def pos_to_cell(self, position):
    # Needs Grid Dimensions to know if pos is out of bounds
    if not 0 <= position[0] < self.width:
        # Needs Torus setting to know if we should wrap or raise error
        if self.torus:
            coord = position % self.dimensions 
        else:
            raise ValueError("Out of bounds")
            
    return self._cells[coord]
```
Hypothetical Alternative: If we moved this to Cell, we would have to inject the Grid into every function call, which is clumsy and circular.

```Python
# Method on the Cell class
class Cell:
    def pos_to_cell(self, position, grid_instance): # Must pass grid!
        # The cell itself doesn't know the grid's width/height
        if grid_instance.torus:
            ...
```

mesa/discrete_space/network.py
codebreaker32: Agreed. I have introduced a flexible layout argument in `__init__`. It addresses:

1. Existing Data: If the graph nodes already have "pos" attributes, we use them (zero touch).
2. User Control: If missing, we use the layout function provided by the user.
3. Opt-Out: Passing layout=None disables spatial features entirely (pure topology)

codebreaker32:
> I am wondering whether pos_to_cell and cell_to_pos might not be better of on the Cell itself and be renamed to pos_to_index and index_to_pos.

I partially agree with the naming but I will strongly advise to keep it on space itself

mesa/discrete_space/discrete_space.py
codebreaker32: I think we also did something like this in https://github.com/mesa/mesa-geo/pull/299, where RasterLayer(eq. to Grid) owns the global context(transform etc.) and Cell only holds their positions

quaquel: Thanks for the clarification. It makes sense now. But then naming matters as stated below.

mesa/discrete_space/discrete_space.py
```diff
@@ -85,6 +87,54 @@ def agents(self) -> AgentSet:
     def _connect_cells(self): ...
     def _connect_single_cell(self, cell: T): ...
 
+    def pos_to_cell(self, position: np.ndarray) -> T:
```
quaquel:
```suggestion
    def position_to_cell_index(self, position: np.ndarray) -> T:
```

mesa/discrete_space/discrete_space.py
quaquel:
```suggestion
    def cell_index_to_pos(self, cell: T) -> np.ndarray | None:
```

Should this ever return None? 

Also, if you have the cell already, what's the point of this method? You can just check `cell.position` yourself. 

In my view, when instantiating a discrete space, we generate the indices and the positions a give these to the cell. Afterwards, there is no need to interact much with a space at all.

In case of stacked spaces, we still might need find nearest cell which returns the cell for a given position.

mesa/discrete_space/network.py
```diff
@@ -56,6 +81,53 @@ def _connect_single_cell(self, cell: Cell):
...
+            raise ValueError(
+                "Cannot convert position to cell in non-spatial network. "
+                "Set spatial=True when creating the Network."
+            )
+
+        position = np.asarray(position)
+
+        # Find nearest node by brute force   FIXME: Could optimize with KD-tree if many nodes
```
quaquel: As indicated above, I do think we need at some point a nearest_cell _for_position style method so a KD tree might make sense for that.

mesa/discrete_space/discrete_space.py
codebreaker32: >if you have the cell already, what's the point of this method? You can just check cell.position yourself

This is true for Explicit Geometry (Voronoi/Network) but false for Implicit Geometry (Grids especially HexGrid).

1. Explicit (Voronoi/Network): Every cell has a unique, irregular location, so we store it in `cell.position`.
2. Implicit (Grid): We do **not** store `position` on Grid cells to save memory.
* A 1,000 x 1,000 grid has 1 million cells.
* Storing a numpy.array on every cell would consume ~100 MB of RAM unnecessarily.
* Instead, `Grid` calculates it on-the-fly using the index and spacing.


In `VoronoiGrid` (Explicit), `cell.position` is populated:

```python
self._cells = {
    i: cell_klass(..., position=self.centroids[i]) # Stored!
}

```

In `Grid` (Implicit), `cell.position` is `None`:

```python
# We do NOT pass position here to save memory.
self._cells = {
    coord: self.cell_klass(coord, ...) 
}

# The Space acts as the calculator:
def _calculate_position(self, cell) -> np.ndarray:
    # Calculated on demand!
    return np.array(cell.coordinate) + 0.5 

```

`space.cell_to_pos(cell)` provides a unified API that handles both cases (Lookup vs. Calculation) without the user needing to know the implementation detail.

>Should this ever return None?

Yes. For Purely Topological Networks.
If a user creates a Network without a layout (as discussed above), the cells exist logically but have no physical form.


Also, I prefer `cell_to_pos(cell)` because the Agent holds a reference to the Cell Object (`self.cell`), not just the index.
However what you are suggesting might be better formulated as `space.cell_index_to_pos(agent.cell.coordinate)` 

Since `pos_to_cell(pos)` returns a Cell Object, the inverse `cell_to_pos(cell)` should take a Cell Object.

mesa/discrete_space/network.py
codebreaker32: Do you prefer it in this PR only or should I leave it for now?

quaquel: it can wait

mesa/discrete_space/discrete_space.py
quaquel: 1. I doubt people with do 1000x1000 grids in Mesa. But fair enough
2. The position differs across space types. For orthogonal grids, we can simply use the index (rather than +.5). For networks, it is up to the user. For Voronoi, we have the position, and we create a logical index. For hexgrids, it's a bit trickier. Regardless, I think that for none of these, the dimensions of the space or the torus property matter, and so it can be handled in the cell or at the instantiation of the Cell. Am I missing something?

codebreaker32: I agree with both of your points and must say they are strong arguments for keeping `cell_to_pos` at Cell level but I still would like to keep `pos_to_cell` at Space level because it needs the space knowledge(about torus etc.)
Would `coordinate_to_position` and `position_to_coordinate` be fine or Do you prefer some other names?

quaquel: Would it be possible to handle index to position translation in a `position` property?  Something along the lines of:

```
@property
def position(self):
    if self._position is not None:
        return self._position
    return np.asarray(self.index)
```

And then in a discrete space, if we have the position, e.g., for networks, we set self._position explicitly.

I agree on coordinate_to_position, but this might be collapsed into `find_nearest_cell(position:np.ndarray)` if I am not mistaken.

codebreaker32: Thanks @quaquel, this structure makes a lot of sense. I will proceed with implementing `Cell.position` as a property and refactoring the space API to use `find_nearest_cell(position)` instead of the conversion methods, which offers a much cleaner abstraction for Voronoi and Networks. 

Regarding the HexGrid implementation, I will opt for the approach where we calculate and inject the explicit pixel positions during initialization, as the Cell itself shouldn't need awareness of the Grid's hex_size

>For orthogonal grids, we can simply use the index (rather than +.5

Honestly, I don't see it as an ideal choice for obvious reasons

quaquel: > Honestly, I don't see it as an ideal choice for obvious reasons

Not sure what you mean here. To me, it makes more sense that the `cell.coordinate` is the `cell.position` in the case of OrthogonalGrids. And then, for finding the nearest cell for a coordinate in case of orthogonal grids, you just round to the nearest integer.

EwoutH:
@codebreaker32 can we set properties on the cell that calls the methods from the space?

mesa/discrete_space/discrete_space.py
codebreaker32: >Because of this, on the visualization side, we have to map these coordinates to x,y positions.

1. Looking at the long term, this function will essentially act as the coordinate provider for our visualization modules. If get_position returns the raw index (the corner), agents will be rendered sitting on the grid intersections (vertices) rather than inside the cells, which visually implies a physics bug. If we attempt to fix this 'downstream' by having the visualization layer manually normalize the coordinates (e.g., adding 0.5), we force the renderer to understand the specific geometry rules of the grid. That reintroduces the exact tight coupling we are trying to eliminate

>Given a way of declaring the relationship between spaces, the next thing is to automatically make an agent show up in all "stacked" spaces. This is an implementation detail. But, whenever agent.position, this should be propagated to all relevant spaces. We also need an API for an agent or perhaps even an agent type to "opt out" of any given space.

2. I interpret this as:

- Continuous agent has agent.position (e.g., [5.7, 3.2])
- Agent registers in stacked Grid space
- Grid automatically determines which cell: find_nearest_cell([5.7, 3.2]) → cell (5, 3)
- Agent appears in cell (5, 3) for grid-based operations

But: When the agent needs to reference the cell's position (for physics, movement, attraction), what should cell.position return?

Another weak argument is if we look at it through lens of pure physics, agents at center of cell looks more intuitive to me...

Curious to know your thoughts or please correct if I am assuming anything wrong.

codebreaker32:
> @codebreaker32 can we set properties on the cell that calls the methods from the space?

Yes, we can do something like this

```python
class Cell:
    """Cell with position property that delegates to space."""
    
    __slots__ = [
        "_agents",
        "_empty",
        "capacity",
        "connections",
        "coordinate",
        "properties",
        "random",
        "_space",  # NEW
    ]
    
    def __init__(
        self,
        coordinate: Coordinate,
        capacity: int | None = None,
        random: Random | None = None,
        space: DiscreteSpace | None = None,  # NEW
    ):
        self.coordinate = coordinate
        self._space = space  # Store reference to parent space
        ...
    
    @property
    def position(self) -> np.ndarray:
        """Physical position of the cell."""
        if self._space is None:
            # Fallback: use coordinate as position
            return np.asarray(self.coordinate, dtype=float)
        
        # Delegate to space (space knows its geometry)
        return self._space._get_cell_position(self)
```

@quaquel What do you think of this approach?

mesa/discrete_space/discrete_space.py
quaquel: > If get_position returns the raw index (the corner), agents will be rendered sitting on the grid intersections (vertices) rather than inside the cells, which visually implies a physics bug.

I don't follow. The grid line drawing is an implementation detail that should be derived from how we handle the position of the cell. If we put the cell on 1,1, then grid lines should be drawn on 0.5 and 1.5. So what is considered the corner is a choice we can make while implementing this. 


> I interpret this as:

Good questions, and part of implementing all this is to flesh out the answers to these questions.

1. The agent position is leading in my view.
2. Once it is determined that a given agent is in cell 5.3, its `agent.cell` attribute can be set to that cell, and so it is added to `cell.agents`. This means, if other agents want to use cell-based lookups, they can easily find them. For example, what are the agents that live in the same postcode area as me? Here, postcode areas could be implemented using a DiscreteSpace. 
3. I don't know of a use case where the physical position of the cell should matter inside a discrete space itself. Through the connections, we already specify the topology of a discrete space. If you want to know neighboring cells, you can just do a `cell.neighborhood`. 
4. Of course, in the case of the postcode example, an agent might want to know the nearest neighboring postcode to itself. But then, it could just ask the postcode space for k nearest neighbors via `agent.position`.

quaquel:
I still don't understand why this would be needed. In my view, we either pass the position at cell construction to the cell. This is needed for networks (if provided) for Voronoi (where we already have it), and for hexgrids (I guess). Or we don't pass it, and the cell constructs it on the fly to the best of its ability. So if `cell.position` is None, it will try to create a 2D np array using its coordinate. 

Discrete space level properties like `torus` don't matter as far as I can see in this. Only for finding the nearest cell for a given x,y position might the torus property matter. But of that, I am not sure. For orthogonal grids, you can just round or floor the x,y positions (depending on how we handle the coordinate position translation). Again, for hexgrids, I would need to take a closer look.

mesa/discrete_space/discrete_space.py
codebreaker32: > I don't follow. The grid line drawing is an implementation detail that should be derived from how we handle the position of the cell. If we put the cell on 1,1, then grid lines should be drawn on 0.5 and 1.5. So what is considered the corner is a choice we can make while implementing this.

Understood.

>I don't know of a use case where the physical position of the cell should matter inside a discrete space itself. Through the connections, we already specify the topology of a discrete space. If you want to know neighboring cells, you can just do a cell.neighborhood

**Should cells have SPATIAL EXTENT (position in space) or only TOPOLOGICAL EXTENT (connections)?**

### If cells are purely topological, then:

1. Cells only matter for lookups ("who's near my cell?")
2. All physics happens agent-to-agent via agent.position
3. Cell.position is rendering-only


### If cells have spatial extent, then:

1. Cells can have spatial properties (temperature at center, resource density)
2. Physics can be agent-to-agent AND agent-to-cell
3. Cell.position is needed for physics calculations


For now, I think we can agree over cell to be purely topological but as we move closer to implementation of stacked spaces, it might be worth exploring the second alternative and its use-cases. Also, Thanks for detailed clarification

quaquel: > Should cells have SPATIAL EXTENT (position in space) or only TOPOLOGICAL EXTENT (connections)?

That is a very good question. I don't know. For simple spaces like hexes and orthogonal grids, the spatial extent is simple (assuming a given size for a cell). For Voronoi, we can generate it by construction. For networks, there is no natural answer to the question.

codebreaker32:
I have updated the PR to align with our discussion, 

1. Implemented `space.find_nearest_cell(position)` and the `Cell.position` property.
2. I completely refactored `HexGrid` to cleanly inherit from `Grid.__init__`, intitalising its geometry using a dedicated method `_init_hex_geometry` and using `scipy.spatial.KDTree` indexing. 
3. This same  KD-Tree optimization has been used across all non-orthogonal spaces, including spatial `Network`s (which now dynamically rebuild their trees upon node changes) and `VoronoiGrid`. 
4. Finally, Test Suit updated

.github/release.yml
```diff
@@ -10,15 +10,15 @@ changelog:
...
-    - title: 🧪 Experimental features
-      labels:
-        - experimental
     - title: 🎉 New features added
       labels:
         - feature
     - title: 🛠 Enhancements made
       labels:
         - enhancement
+    - title: 🧪 Experimental
```
quaquel, line 19: Would you mind moving this back to the original place. It's one less file in the diff.

mesa/discrete_space/cell.py
```diff
@@ -56,22 +60,41 @@ def empty(self) -> bool:  # noqa: D102
...
+    def position(self) -> np.ndarray:
+        """Get the physical position of the cell.
+
+        Returns:
+            np.ndarray: Physical position of the cell
+        """
+        if self._position is not None:
+            return self._position
+        # Default for implicit grids
+        return np.asarray(self.coordinate, dtype=float)
```
quaquel, line 73: Is `coordinate` always 2D, and, if not, should we check for this?

mesa/discrete_space/discrete_space.py
```diff
@@ -85,6 +87,13 @@ def agents(self) -> AgentSet:
     def _connect_cells(self): ...
     def _connect_single_cell(self, cell: T): ...
 
+    def find_nearest_cell(self, position: np.ndarray) -> T:
```
quaquel, line 90: Should we not turn DiscreteSpace into an ABC and then handle this and the methods above as abstract methods?

mesa/discrete_space/grid.py
```diff
@@ -124,6 +126,34 @@ def __init__(
...
+            position: Physical position [x, y]
+
+        Returns:
+            Cell: The cell containing the position
+
+        Raises:
+            KeyError: If position is outside grid bounds and not a torus
+        """
+        # Floor to get cell coordinate
+        coord = tuple(np.floor(position).astype(int))
```
quaquel, line 142: This is also related to your remark on spatial extent. 

In my thinking, I see the position of the cell at the center. So I would have used round instead of floor. 

Also, either approach, of course, implicitly assumes that the cell size is 1. Which is an assumption I can live with for now.

mesa/discrete_space/grid.py
```diff
@@ -287,6 +317,68 @@ class HexGrid(Grid[T]):
...
+            col, row = coord
+            x = size * math.sqrt(3) * (col + 0.5 * (row % 2))
+            y = size * 1.5 * row
+            position = np.array([x, y])
+
+            cell.position = position
+            positions.append(position)
+            self._kdtree_coords.append(coord)
+
+        self._kdtree = KDTree(np.array(positions))
```
quaquel, line 365: have you done any performance benchmarking on the KDTree?

mesa/discrete_space/grid.py
```diff
@@ -287,6 +317,68 @@ class HexGrid(Grid[T]):
...
+
+    def _init_hex_geometry(self) -> None:
+        """Calculate physical positions for all cells and build KD-Tree.
+
+        Refer https://www.redblobgames.com/grids/hexagons/#hex-to-pixel for more detail
+        """
+        positions = []
+        self._kdtree_coords = []
+
+        size = 1.0
```
quaquel, line 354: This is again related to spatial extent. What about turning this into a class variable on Grid? So its one for both orthogonal grids and hexgrids, but also becomes user controllable if desired?

mesa/discrete_space/network.py
```diff
@@ -36,17 +42,50 @@ def __init__(
...
+                nodes if they are missing (e.g. nx.spring_layout).
+                Set to None to force a purely topological network (no positions).
+                Defaults to nx.spring_layout.
 
         """
         super().__init__(capacity=capacity, random=random, cell_klass=cell_klass)
         self.G = G
 
+        # Check for existing positions to determine if this is a spatial network
+        has_existing_pos = any("pos" in data for _, data in self.G.nodes(data=True))
```
quaquel: You implicitly assume that "pos" on the node is used for position. I am not sure about going that route. Instead, I would consider overloading layout to optionally also take a numpy array of (len(nodes) X 2).

mesa/discrete_space/network.py
```diff
@@ -36,17 +42,50 @@ def __init__(
...
+
+        # If positions are missing and a layout is provided, generate them.
+        if not has_existing_pos and layout is not None:
+            positions = layout(self.G)
+            nx.set_node_attributes(self.G, positions, "pos")
```
quaquel: is this needed?

mesa/discrete_space/network.py
```diff
@@ -36,17 +42,50 @@ def __init__(
...
+                random=self.random,
+                position=pos,  # None for topological networks
             )
 
         self._connect_cells()
+        self._build_kdtree()
+
+    def _build_kdtree(self) -> None:
+        """Build the KD-Tree for fast nearest-cell lookups."""
+        self._kdtree_cells = [
```
quaquel: it seems a bit strange to loop over positions again since we loop over this also in the init

mesa/discrete_space/network.py
```diff
@@ -36,17 +42,50 @@ def __init__(
...
+            cell for cell in self._cells.values() if cell._position is not None
+        ]
+
+        if self._kdtree_cells:
+            positions = np.array([c.position for c in self._kdtree_cells])
```
quaquel: and this is the third loop over all cells. With larger spaces, this seems really redundant and a bit of waste of compute time.

mesa/visualization/mpl_space_drawing.py
```diff
@@ -71,6 +71,10 @@ def get_agent_pos(agent, space):
             agent_x, agent_y = agent.pos, agent.pos
         elif isinstance(space, Network):
             agent_x, agent_y = agent.cell.coordinate, agent.cell.coordinate
+        elif isinstance(space, VoronoiGrid):
```
quaquel, line 74: should we not consider updating all space drawing for discrete spaces in a follow up PR?

.github/release.yml
codebreaker32: I haven't commited this change, See https://github.com/mesa/mesa/pull/3268/changes/2593d73e587475045dc024d250f04fa6b7154fa0

mesa/discrete_space/cell.py
codebreaker32: `np.asarray` will handle it so I don't think we need a check

mesa/discrete_space/discrete_space.py
codebreaker32: Agree. If you advice, I can do this in follow-up PR

mesa/discrete_space/grid.py
codebreaker32: I haven't done any benchmarking in Mesa but In some of my previous projects, I have used it and it significantly reduces time due to its better complexity. I will do it soon and write it in PR note.

mesa/discrete_space/grid.py
codebreaker32: Agree.

mesa/discrete_space/network.py
codebreaker32: Fair enough. I will refactor it.

mesa/discrete_space/network.py
codebreaker32: if we move away from explicitly relying on the nx "pos" attribute, we can safely remove this

mesa/visualization/mpl_space_drawing.py
codebreaker32: Agree.

mesa/discrete_space/grid.py
codebreaker32: Wouldn't it mean that first cell will stretch from -0.5 to 0.5?

mesa/discrete_space/network.py
codebreaker32: You are right. I will try to optimise it.

mesa/discrete_space/network.py
codebreaker32: I think `Mapping` would be better instead of a numpy array

mesa/discrete_space/grid.py
quaquel: yes that is a fair point

mesa/discrete_space/grid.py
codebreaker32: So I ran some tests only for spatial lookups(`find_nearest_cell()`) and KD-Tree is ~200x faster than Brute-Force for about 10,000 nodes and 1,000 queries which clearly surpasses the `_build_tree` overhead

codebreaker32:
Thanks for the great review as always :)
I have addressed all your major suggestions, particularly eliminating the redundant graph traversals in `network.py` by fusing the initialization and KD-Tree array construction into a single pass. I also refactored the layout argument to accept `Mapping` or `Callable`

For the remaining structural ideas (making DiscreteSpace an ABC and introducing a configurable size attribute on Grid and modify space_drawer), I think those would be best suited for a follow-up PR to keep this one atomic.

codebreaker32:
Minor tweaks done. PR description updated.
Good to go from my end

PR #3270, opened 2026-02-10
Migrate tutorials and examples to model.run_for()

falloficaruss:
### Summary
This PR migrates Mesa's tutorials, examples, and benchmarks to use the new `Model.run_for()` method. It replaces manual `for _ in range(N): model.step()` loops with a single, unified call to `model.run_for(N)`.

### Motive
With the introduction of event-based scheduling, running a model for a specific duration should be handled by the model's internal time-advancement mechanism rather than external loops. Manual loops calling `.step()` can lead to inconsistencies when models use discrete-event simulators. `model.run_for()` provides a declarative, robust, and unified way to advance a simulation regardless of the underlying scheduling mechanism.

### Implementation
1. Tutorials: Updated all Jupyter notebooks in `docs/tutorials/` to use `run_for()` which was introduced in #3266 . This involved updating both the code cells and the surrounding markdown explanation to reflect the recommended usage.
2. Benchmarks: Refactored `benchmarks/global_benchmark.py` to use `model.run_for()`. This significantly simplified the code by removing the need to check if a model was using a `Simulator` or a standard loop.
3. Tests:
  - Refactored `tests/examples/test_examples.py` to replace repetitive test loops with concise `run_for()` calls.
  - Added a dedicated unit test in tests/test_model.py to verify that `run_for()` correctly increments both model steps and simulation time.

### Usage Examples
Legacy Pattern:
```python
model = MoneyModel(10)
for _ in range(30):
    model.step()
```
New Pattern:
```python
model = MoneyModel(10)
model.run_for(30)
```
### Additional Notes
- Related Issue: #3132

### DIFF
--- a/benchmarks/global_benchmark.py
+++ b/benchmarks/global_benchmark.py
@@ -12,8 +12,6 @@ sys.path.insert(0, os.path.abspath(".."))
 
 from configurations import configurations
 
-from mesa.experimental.devs.simulator import ABMSimulator
-
 
 # Generic function to initialize and run a model
 def run_model(model_class, seed, parameters):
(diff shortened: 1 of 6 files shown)


EwoutH:
Thanks for working on this. I will look into it after #3266 is merged.

EwoutH:
#3266 is merged. Go ahead and update/rebase this PR!

tests/test_model.py
EwoutH, line 1: I have tests covered in #3266, so this file can be reverted.

mesa/model.py
```diff
@@ -349,6 +349,14 @@ def run_model(self) -> None:
         while self.running:
             self.step()
 
+    def run_for(self, steps: float) -> None:
```
EwoutH: This is implemented in #3266, so this file can be reverted.

tests/test_model.py
EwoutH: These still need to be reverted

falloficaruss: Done

PR #3271, opened 2026-02-10
Add warning for `mesa.batch_run` top-level API

souro26:
### Summary
Add a `DeprecationWarning` to the `mesa.batch_run` entry point at the top-level API. This provides a runtime signal to users that this specific access pattern is scheduled for removal in Mesa 4.0.

### Motive
`batch_run` is planned for removal as part of the Mesa 4.0 transition. While the deprecation is documented, many users still access it via `mesa.batch_run`. Emitting a warning at this entry point ensures users receive immediate feedback when relying on this access pattern, improving migration visibility and reducing surprise when the API is removed.

### Implementation
- Wrapped the `mesa.batch_run` entry point in `mesa/__init__.py` to emit a `DeprecationWarning` before delegating to the existing implementation.
- Added a test to assert that calling `mesa.batch_run` triggers the warning.
- No functional behavior changes were made to `batch_run` itself.

### Usage Examples
When a user runs an existing model using the top-level import

```python

import mesa
# ... model setup ...
results = mesa.batch_run(MyModel, params)
````

They will now see: `DeprecationWarning: mesa.batch_run is deprecated and will be removed in Mesa 4.0. Use the new Scenario / DataSet APIs instead.`

### Additional Notes
- The warning only applies to access via `mesa.batch_run`, internal and direct imports remain unchanged.
- Aligns with the Mesa 4.0 deprecation and migration plan.

### DIFF
--- a/tests/test_batch_run.py
+++ b/tests/test_batch_run.py
@@ -815,5 +815,8 @@ def test_batch_run_agenttype_and_agent_reporters():
 
 def test_batch_run_top_level_deprecation_warning():
     """Calling mesa.batch_run should emit a DeprecationWarning."""
-    with pytest.warns(DeprecationWarning, match="mesa.batch_run is deprecated"), contextlib.suppress(Exception):
+    with (
+        pytest.warns(DeprecationWarning, match="mesa.batch_run is deprecated"),
+        contextlib.suppress(Exception),
+    ):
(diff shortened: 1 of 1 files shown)


quaquel:
This is premature. I suggest we wait until mesa 4 takes a clear shape and than reverse engineer what we will deprecate and create a 3.6 release with all relevant warnings.

Issue #3273, opened 2026-02-11
experimental/continuous_space: add optional movement helpers (mixin/helper module)


falloficaruss:
**What's the problem this feature will solve?**
In `experimental/continuous_space`, common movement operations (translate by vector, move toward target, move along direction) are repeatedly implemented in model code. This creates duplication and inconsistent handling of edge cases, especially in torus spaces where shortest-path movement is easy to implement incorrectly.

**Describe the solution you'd like**
Add optional movement helpers for continuous-space agents as an opt-in layer (mixin and/or helper module), rather than expanding the base `ContinuousSpaceAgent` API.

Proposed helpers:

- `move_by(delta)`
- `move_towards(target, distance, clamp=True) -> float`
- `move_in_direction(direction, distance)`

Design expectations:

- Reuse `ContinuousSpace.calculate_difference_vector(...)` for torus-aware direction calculation.
- Keep position assignment as the single source of truth for bounds/torus correction.
- Define explicit behavior for overshoot (clamp), zero-distance targets, and zero direction vectors.
- Keep backward compatibility by leaving existing `ContinuousSpaceAgent` behavior unchanged.

**Additional context**
- Related discussion: https://github.com/mesa/mesa/discussions/2967
- This is an ergonomic enhancement for the experimental API.
- Goal is to improve usability while avoiding base-class API sprawl.

Ishwarpatra:
Request to assign it to me,I will try solve the issue @falloficarus22

falloficaruss:
A related PR is already raised for this issue

tpike3:
Closing to make issues more impactful and relevant to current Mesa issues (anything before Mesa 3.0 release Nov 9, 2024) per May 26, 2026 dev meeting.

PR #3274, opened 2026-02-11
Add opt-in movement helpers and MovingContinuousSpaceAgent

falloficaruss:
### Summary
Adds an opt-in movement API for experimental continuous space via a new `BasicContinuousMovement` mixin and `MovingContinuousSpaceAgent` class. This provides common movement operations without expanding the base `ContinuousSpaceAgent` interface.

### Motive
Continuous-space model code frequently reimplements the same movement logic (`translate`, `move toward target`, `move along direction`). This leads to duplication and inconsistent handling of torus edge cases. The goal is to improve ergonomics while keeping the core agent API minimal, aligned with discussion #2967.

### Implementation
1. Added `movement.py`:
  - `BasicContinuousMovement.move_by(delta)`
  - `BasicContinuousMovement.move_towards(target, distance, clamp=True) -> float`
  - `BasicContinuousMovement.move_in_direction(direction, distance)`
2. Added `MovingContinuousSpaceAgent(BasicContinuousMovement, ContinuousSpaceAgent)` in:
  - `continuous_space_agents.py`
3. Updated exports in:
  - `__init__.py`
4. Added tests in:
  - `test_continuous_space.py`
5. Key behavior decisions:
  - Torus-aware target direction uses existing `ContinuousSpace.calculate_difference_vector(...)`.
  - `move_towards(..., clamp=True)` prevents overshoot.
  - Zero-distance targets and zero direction vectors are no-ops.
  - Negative `distance` raises `ValueError` for movement methods.

### Usage Examples
```python
import numpy as np
from mesa import Model
from mesa.experimental.continuous_space import ContinuousSpace, MovingContinuousSpaceAgent

model = Model(rng=42)
space = ContinuousSpace(np.asarray([[0, 1], [0, 1]]), torus=True, random=model.random)

agent = MovingContinuousSpaceAgent(space, model)
agent.position = [0.9, 0.5]

# Translate by vector
agent.move_by([0.05, -0.1])

# Move toward target by up to distance (torus-aware shortest path)
moved = agent.move_towards([0.1, 0.5], distance=0.15, clamp=True)

# Move in normalized direction
agent.move_in_direction([3, 4], distance=0.2)
```

### Additional Notes
- This is an additive change; existing `ContinuousSpaceAgent` behavior remains unchanged.

- Related discussion: https://github.com/mesa/mesa/discussions/2967
- Related Issue: #3273


EwoutH:
Thanks for the PR. There is consensus among maintainers that we really want this in Mesa eventually. However, I question if the time is now. We aligned on a process of integrating and stabilizing the spaces in https://github.com/mesa/mesa/discussions/2585#discussioncomment-15496687 and I think it would be wise to do that first before adding new API surfaces.

Secondly, this movement API has a big surface: methods, keywords, and all their possible names. So I like to have a separate discussion about API design and naming, following our [development process](https://github.com/mesa/mesa/blob/main/CONTRIBUTING.md#mesa-development-process).

@quaquel is currently the lead on spaces. Curious on his perspective, especially on the envisioned order of things.

mesa/experimental/continuous_space/movement.py
```diff
@@ -0,0 +1,61 @@
...
+
+from __future__ import annotations
+
+from typing import Protocol
+
+import numpy as np
+from numpy.typing import ArrayLike
+
+
+class SupportsContinuousMovement(Protocol):
```
quaquel, line 11: should we make this runtime checkable?

mesa/experimental/continuous_space/movement.py
```diff
@@ -0,0 +1,61 @@
...
+    """Protocol for objects that can use continuous movement helpers."""
+
+    space: object
```
quaquel, line 14: the typing is not correct here

mesa/experimental/continuous_space/movement.py
```diff
@@ -0,0 +1,61 @@
...
+    position: np.ndarray
+
+
+class BasicContinuousMovement:
```
quaquel, line 18: There is an [open discussion](https://github.com/mesa/mesa/discussions/2967) on this. i am not sure the methods here  align well with the ideas put forward in that discussion

quaquel:
I agree on the need for this functionality. However, I don't think we aligned completely yet, and exactly what we wanted. See [this discussion](https://github.com/mesa/mesa/discussions/2967). Also, from a development process, for MESA 4 we want to offer support for "stacked spaces". This will require some rerigging of the internals of the various space classes. I believe it might be better to first do that and then return to adding this functionality.

falloficaruss:
Thanks, this makes sense. I agree the stacked-spaces work for Mesa 4 should come first, since it may change the right API surface for movement helpers.
I’ll pause this PR and keep it as a reference implementation

mesa/experimental/continuous_space/movement.py
falloficaruss: I’d keep this protocol static-only for now. We’re not using runtime `isinstance` checks against it, so `@runtime_checkable` wouldn’t provide practical value yet. If we later introduce runtime validation or plugin-style checks, we can add `@runtime_checkable` at that point.

mesa/experimental/continuous_space/movement.py
falloficaruss: I used `object` initially to keep it generic, but this should be tightened.

falloficaruss:
@tpike Closing this as the stacked-space foundation is till in the flux. Better to redo in the future.

PR #3275, opened 2026-02-11
replace generic `Exception` with `TypeError` in `eventlist.py`

souro26:
### Summary
Use `TypeError` instead of a generic `Exception` when a non-callable is passed to `SimulationEvent`.

### Bug / Issue
Currently, `SimulationEvent.__init__` raises a bare `Exception()` when the provided argument is not callable. Using a generic `Exception` makes the error less precise and harder to catch programmatically. Since the failure is caused by an invalid argument type, `TypeError` is more appropriate and aligns with standard python error semantics.

### Implementation
Replaced `raise Exception()` with
````python
raise TypeError("SimulationEvent function must be callable")
````
Updated the corresponding test in `test_simulation_event` to expect `TypeError` instead of a generic `Exception`.

### Testing
- Updated the existing test case that verifies non-callable input handling to assert `TypeError`.
- Ran the full experimental test suite to ensure no regressions.

### Additional Notes
This change is intentionally scoped only to `SimulationEvent` to avoid broader semantic changes across other experimental modules.

### DIFF
--- a/tests/experimental/test_devs.py
+++ b/tests/experimental/test_devs.py
@@ -181,7 +181,7 @@ def test_simulation_event():
     event.execute()
     some_test_function.assert_called_once()
 
-    with pytest.raises(TypeError, match = "function must be callable"):
+    with pytest.raises(TypeError, match="function must be callable"):
         SimulationEvent(
             time, None, priority=Priority.DEFAULT, function_args=[], function_kwargs={}
         )


quaquel:
We are reorganizing the code. I am closing this PR, but feel free to open a new one once those PRs have been merged. Please also check the discussion on Mesa 4 and the exception hierarchy.

PR #3276, opened 2026-02-11
Stabilize event scheduling system from experimental to mesa.time

EwoutH:
### Summary
Move the event scheduling functionality from `mesa.experimental.devs.eventlist` to a new stable `mesa.time` module. This stabilizes the event-based simulation system that was introduced experimentally and makes it a first-class feature of Mesa.

Also renames `SimulationEvent` to `Event`.

### Motive
The event scheduling system (EventList, SimulationEvent, EventGenerator, Schedule, Priority) has been in experimental status since its introduction. After extensive testing and usage, it's ready to be stabilized and made part of Mesa's core API. This promotes better time management in agent-based models and supports both discrete event simulation and hybrid approaches.

### Usage Examples
The event scheduling system can now be imported from the stable namespace:

```python
from mesa.time import EventList, SimulationEvent, Priority, Schedule, EventGenerator

# Schedule a one-time event
model.schedule_event(my_function, at=5.0, priority=Priority.HIGH)

# Schedule recurring events
schedule = Schedule(interval=1.0, start=0.0)
model.schedule_recurring(my_function, schedule, priority=Priority.DEFAULT)

# Use event generators for complex recurring patterns
generator = EventGenerator(model, my_function, schedule).start()
```

All existing functionality remains unchanged, this is purely a reorganization to stabilize the API.

### Additional Notes
- This is a non-breaking change for users already using the experimental API (imports from `mesa.experimental.devs.eventlist` still work via the simulator module's imports)
- The Simulator classes remain in experimental status (and will be deprecated and removed in follow-up PRs)
- All tests pass with updated imports
- Documentation has been updated to reflect the new stable location
- Follows Mesa's deprecation policy by providing the alternative before any deprecation warnings

Follows #3201, #3204 and #3266, discussed in #2921, part of #3132.

### DIFF
--- a/mesa/experimental/devs/simulator.py
+++ b/mesa/experimental/devs/simulator.py
@@ -27,7 +27,7 @@ import warnings
 from collections.abc import Callable
 from typing import TYPE_CHECKING, Any
 
-from mesa.time import EventList, Priority, SimulationEvent
+from mesa.time import Event, EventList, Priority
 
 if TYPE_CHECKING:
     from mesa import Model
@@ -171,7 +171,7 @@ class Simulator:
         priority: Priority = Priority.DEFAULT,
         function_args: list[Any] | None = None,
         function_kwargs: dict[str, Any] | None = None,
-    ) -> SimulationEvent:
+    ) -> Event:
         """Schedule event for the current time instant.
 
         Args:
@@ -181,7 +181,7 @@ class Simulator:
             function_kwargs (Dict[str, Any]):  dict of keyword arguments for function
 
         Returns:
-            SimulationEvent: the simulation event that is scheduled
+            Event: the simulation event that is scheduled
 
         """
         return self.schedule_event_relative(
@@ -199,7 +199,7 @@ class Simulator:
         priority: Priority = Priority.DEFAULT,
         function_args: list[Any] | None = None,
         function_kwargs: dict[str, Any] | None = None,
-    ) -> SimulationEvent:
+    ) -> Event:
         """Schedule event for the specified time instant.
 
         Args:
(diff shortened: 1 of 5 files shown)


mesa/experimental/devs/__init__.py
```diff
@@ -1,24 +1,8 @@
...
-The implementation supports both pure discrete event simulation and hybrid approaches
-combining agent-based modeling with event scheduling.
+Stabilized in the Mesa Model and in mesa.time.events. Simulators are deprecated.
 """
 
-from .eventlist import Priority, SimulationEvent
 from .simulator import ABMSimulator, DEVSimulator, Simulator
 
-__all__ = ["ABMSimulator", "DEVSimulator", "Priority", "SimulationEvent", "Simulator"]
+__all__ = ["ABMSimulator", "DEVSimulator", "Simulator"]
```
quaquel, line 8: should SimulationEvent not still be here and also raise a deprecation warning? 

Of course, since it is experimental, we can technically delete it.

quaquel (approved):
One minor comment, but looks fine to me otherwise.

mesa/experimental/devs/__init__.py
EwoutH: Certainly true, but like I said in #3277:

> We could move more aggressive and directly remove it, which is allowed for experimental features in our deprecation policy, but this feature has been with us for a long time, so let's deprecate it gently.

We can remove it once we move from 3.5 to 4.0.

EwoutH:
Just realized I didn't update the test file, that's still in `tests/experimental/test_devs.py`.

Once we remove the simulators, I will clean it up and move/rename to `tests/time/test_events.py`.

PR #3278, opened 2026-02-11
Update wolf-sheep example to use new event scheduling API

EwoutH:
### Summary
Updates the wolf-sheep example to use the new public event scheduling API (`model.schedule_event()`) instead of the deprecated `ABMSimulator`, and refactors `GrassPatch` for cleaner, more Pythonic code.

### Implementation
The main change is migrating from `simulator.schedule_event_relative()` to `model.schedule_event(after=...)`, and removing the old simulator objects.

`agents.py` also received some refactoring. The old `GrassPatch` implementation used a property setter with side effects—setting `fully_grown = False` would automatically schedule a regrowth event using `setattr` as a callback. This was clever but opaque. The new implementation adds explicit `regrow()` and `eat()` methods that make the grass lifecycle clear: when eaten, grass schedules its own regrowth. The `Sheep.feed()` method now calls `grass_patch.eat()` instead of directly assigning to the property.

Grass is now regrowing:
Screenshot_787

### Additional Notes
- This is the only included Mesa example using the experimental Simulator functionality. After this change, all core examples use only stable, public APIs.
- There is something weird going on with Events, lambda's, weakrefs, and garbage collection. I thought I fixed it in https://github.com/mesa/mesa/pull/3266#issuecomment-3876089720 / https://github.com/mesa/mesa/pull/3266/commits/3929158961591df2016506bb5a03e9c4eb15f982, but it might not be fully fixed.

### DIFF
--- a/benchmarks/global_benchmark.py
+++ b/benchmarks/global_benchmark.py
@@ -12,8 +12,6 @@ sys.path.insert(0, os.path.abspath(".."))
 
 from configurations import configurations
 
-from mesa.experimental.devs.simulator import ABMSimulator
-
 
 # Generic function to initialize and run a model
 def run_model(model_class, seed, parameters):
@@ -27,26 +25,17 @@ def run_model(model_class, seed, parameters):
     Returns:
         startup time and run time
     """
-    uses_simulator = ["WolfSheep"]
     # Explicitly collect garbage before the run to ensure a clean memory state
     gc.collect()
 
     # Disable GC during timed runs to avoid random slowdowns
     gc.disable()
     start_init = time.perf_counter()
-    if model_class.__name__ in uses_simulator:
-        simulator = ABMSimulator()
-        model = model_class(simulator=simulator, rng=seed, **parameters)
-    else:
-        model = model_class(rng=seed, **parameters)
+    model = model_class(rng=seed, **parameters)
 
     end_init_start_run = time.perf_counter()
 
-    if model_class.__name__ in uses_simulator:
-        simulator.run_for(config["steps"])
-    else:
-        for _ in range(config["steps"]):
-            model.step()
+    model.run_for(config["steps"])
 
     end_run = time.perf_counter()
     gc.enable()  # Re-enable GC after benchmarking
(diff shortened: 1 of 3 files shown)


mesa/examples/advanced/wolf_sheep/agents.py
```diff
@@ -139,12 +122,19 @@ def __init__(self, model, countdown, grass_regrowth_time, cell):
...
-            self.model.simulator.schedule_event_relative(
-                setattr, countdown, function_args=[self, "fully_grown", True]
-            )
+            self.model.schedule_event(self.regrow, after=countdown)
+
+    def regrow(self):
+        """Regrow the grass."""
+        self.fully_grown = True
+
+    def eat(self):
```
quaquel: A pit of a nitpick, but this naming is not ideal. it suggest the grass eats, rather than the grass being eaten.

EwoutH: Shall I rename to `get_eaten()`?

EwoutH:
Not for this PR, but I don't know if the current example displays desired behavior. There's a grass growth spike around t=30 caused by synchronization: while initial grass patches get random regrowth countdowns (0-30), all grass eaten by sheep regrows after exactly 30 time units.

This creates synchronized waves: the initial random grass all finishes around t=30, sheep eat it, then it all regrows together 30 steps later at t=60, creating cascading population spikes. If smoother dynamics are desired, the `get_eaten()` method should use `self.model.rng.integers(1, self.grass_regrowth_time + 1)` instead of the fixed `self.grass_regrowth_time` to maintain randomized regrowth throughout the simulation.

Screenshot_787

quaquel:
> This creates synchronized waves: the initial random grass all finishes around t=30, sheep eat it, then it all regrows together 30 steps later at t=60, creating cascading population spikes. If smoother dynamics are desired, the get_eaten() method should use self.model.rng.integers(1, self.grass_regrowth_time + 1) instead of the fixed self.grass_regrowth_time to maintain randomized regrowth throughout the simulation.

It makes sense to use some distribution for this. I would not use integers, because that's uniform. But a triangular distribution or so, and flooring the number to an integer should work and create nicer dynamics.

EwoutH:
The exact same happens on main:

image

The random initialization spreads out the *first* regrowth, but doesn't prevent synchronization from emerging through the deterministic 30-step regrowth after eating.

EwoutH:
Really nice how this PR also cleans up our test and benchmark code.

quaquel:
> The random initialization spreads out the first regrowth, but doesn't prevent synchronization from emerging through the deterministic 30-step regrowth after eating.

I initially misread your comment and then updated it afterwards.

PR #3284, opened 2026-02-11
Add a signal at start of run

quaquel:
This PR ensures that time is set to 0.0 on the first call to advance_time. This means that anything that subscribes to changes of time can handle the t=0 situation after the model has been fully initialized. 

This is one way of handling the problem. It is, in my view, for now, the simplest way to make the new data collection work. More sophisticated approaches require either additional SignalTypes (e.g., RunSignals.RUN_STARTEDk, RunSignals.RUN_ENDED), but these in turn require a RunConfiguration object. Feedback on this approach is welcome. 

The way it works is that I set `self._time` to 0.0 in the init. This ensures that any get of time will return 0.0 (required in many tests). Then, in `_advance_time` we check this attribute and explicitly set self.time to 0.0. This, in turn, triggers a signal that time is changed to 0.0. And this signal can be used to trigger, e.g., data recording.

If we agree, I'll add additional tests.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,6 +3,132 @@
 ---
 title: Release History
 ---
+# 3.5.0 beta 0 (2026-02-11)
+## Highlights
+Mesa 3.5.0 will be a major feature release that introduces a public event scheduling API, stabilizes the event system, and lays the groundwork for Mesa 4.0 by deprecating several legacy patterns.
+
+Because of the size of this release, we're now first releasing beta release `v3.5.0b0`. Feedback and bug reports are welcome and appreciated!
+
+### Public event scheduling and time advancement
+The headline feature of 3.5.0 is a new public API for event scheduling and time advancement directly on `Model` (#3266). Instead of interacting with experimental `Simulator` classes, users can now schedule events and advance time with simple, expressive methods:
+
+```python
+# Run the model for a duration
+model.run_for(10)       # Advance 10 time units
+model.run_until(50.0)   # Run until absolute time 50
+
+# Schedule one-off events
+model.schedule_event(callback, at=25.0)    # At absolute time
+model.schedule_event(callback, after=5.0)  # Relative to now
+
+# Schedule recurring events
+from mesa.time import Schedule
+model.schedule_recurring(func, Schedule(interval=10, start=0))
+```
+
+For traditional ABMs, `model.run_for(1)` is functionally equivalent to `model.step()`, but this generalizes naturally to event-driven and hybrid models. The mental model shifts from "execute step N" to "advance time, and whatever is scheduled will run."
+
+### Stabilized event scheduling system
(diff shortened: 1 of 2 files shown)


EwoutH:
One thing that I’ve been encountering while actually building stuff with both this and events, is that control over the order becomes more difficult. You need to keep carefully keep track of priorities with events. For signals, how can you currently determine the order of execution if multiple things subscribe to the same signal?

The previous data collection stuff we just setup at the end of the init. I assume here is the problem that there’s user init code executed after that? Can/need we to wrap the user init? Or add a post init?

edit: I thought we discussed something like this at some point, just found it:
- https://github.com/mesa/mesa/discussions/2354

quaquel:
> One thing that I’ve been encountering while actually building stuff with both this and events, is that control over the order becomes more difficult. You need to keep carefully keep track of priorities with events. For signals, how can you currently determine the order of execution if multiple things subscribe to the same signal?

Both events and signals are first-in in first-out. Of course, with events, you have priority to further control this. For signals, you don't have priorities. 

> The previous data collection stuff we just setup at the end of the init. I assume here is the problem that there’s user init code executed after that? Can/need we to wrap the user init? Or add a post init?

The problem is that with the old collection, the user explicitly has to call collect. With the new stuff, we aim for a fully reactive design. So, in #3145, after initializing the DataRecorder, it wants to collect immediately if the first collect is on t=0. But of course, this forces the user to initialize the recorder at the end of the `__init__`. Which is not a nice design. 

There is no `__post_init__` for normal Python classes. Yes, you can do various things using `__init_subclass__`, but in my view, this solution is simpler. It also is a bit of a stopgap solution. For Mesa 4, we really should add a RunConfiguration class that at a minimum specifies start_time and end_time, but which might be extended to include datetime and calendar information.

EwoutH:
So what if somebody wants to check `self.time` before the first run is called?

quaquel:
> So what if somebody wants to check self.time before the first run is called?

It just returns 0.0, because in the `__init__` I set the private variable of the observable to 0.0. 

Just to clarify. An `Observable` is a descriptor. A common pattern in descriptors is to have the `__get__` tied to a name (e.g., `time`) and use an internal private variable (typically `_{public_name}`, so `_time`) for storage. I exploit that here by setting the private variable explicitly in `model.__init__`. Any `__get__` on `model.time` afterwards will just return the value of the private variable (so. 0.0). Once we start running, we explicitly do `model.time = 0.0`, which emits an `ObservableSignals.CHANGED` signal and sets the private variable (again) to 0.0. 

This PR is a somewhat convoluted way to emit a signal indicating that the run has started without having to add new signals. For mesa 4, I want to add RunControlSignals like `RUN_STARTED`, `RUN_STOPPED`, and `RUN_ENDED`. But this requires the TIME api to be further fleshed out on the condition side, likely a RunConfiguration class, and perhaps some other things. For 3.5, and the further development of the new data collection, this solution is good enough.

EwoutH:
Thanks for explaining. Are our observables build in such a way that they always (and automatically) have a private variable with the exact same name but with an underscore? If not, where does that assignment happen?

quaquel:
> Are our observables build in such a way that they always (and automatically) have a private variable with the exact same name but with an underscore? I

yes.

PR #3288, opened 2026-02-12
Fix boid flocking cohesion to use local centroid

falloficaruss:
### Summary
Fixes the `Boids` example so cohesion is explicitly computed toward the **average position of local flockmates** (centroid), matching the canonical Reynolds behavior.

### Bug / Issue
Related issue: https://github.com/mesa/mesa/issues/3286

The `Boids` example claimed canonical behavior, but its steering update combined summed vectors and divided once at the end. That made cohesion less explicitly tied to local centroid steering and could be interpreted as inconsistent with the expected “move toward average neighbor position” rule.

Expected behavior:
- Cohesion steers toward the centroid of nearby flockmates.

Actual behavior before fix:
- Cohesion/alignment/separation were sum-based with a shared final normalization by neighbor count, rather than each rule being directly expressed as an average-driven steering component.

### Implementation
Updated `agents.py`:

- Set self.neighbors directly from `get_neighbors_in_radius(...)` (already excludes self).
- Changed cohesion to:
  - `cohere_vector = delta.mean(axis=0) * self.cohere_factor`
- Changed separation to use mean of only close neighbors (or zero vector if none).
- Changed alignment to use mean neighbor heading.
- Removed shared / len(neighbors) scaling in the final steering update.
- Added safe direction normalization guard to avoid division by zero.
- Added regression test in `test_boid_flockers.py`:

  - `test_boid_cohesion_steers_to_local_centroid`
- Controlled 3-agent setup verifies boid direction moves toward the centroid of two neighbors.

### Testing
`test_examples.py -k boid_flockers`

### Additional Notes
- No visualization/UI changes.
- Behavior now aligns more clearly with canonical centroid-based cohesion semantics for `Boids`.

### DIFF
--- a/mesa/examples/basic/boid_flockers/agents.py
+++ b/mesa/examples/basic/boid_flockers/agents.py
@@ -82,7 +82,8 @@ class Boid(ContinuousSpaceAgent):
             separation_vector = np.zeros(self.space.ndims)
 
         match_vector = (
-            np.asarray([n.direction for n in neighbors]).mean(axis=0) * self.match_factor
+            np.asarray([n.direction for n in neighbors]).mean(axis=0)
+            * self.match_factor
         )
 
         # Update direction based on the three behaviors


EwoutH:
Looking at the diff, maybe I filed the issue wrong. It seems the original model is already steering to the group centroid?

Weird why I remember it being different.

falloficaruss:
> Looking at the diff, maybe I filed the issue wrong. It seems the original model is already steering to the group centroid?
> 
> Weird why I remember it being different.

I think I missed something. Maybe we should keep the regression test and drop the behavior changes.
Also, this implementation is:

- Clearer: each Boids rule is computed the canonical way (own local mean).
- More tunable: weights (cohere, separate, match) have more direct, predictable effects.
- Less hidden coupling: no shared final / n scaling across all rules.
- Easier to test and explain.

So, maybe I can create a seperate PR regarding this refactor?

PR #3289, opened 2026-02-12
Prevent infinite loop from non-positive Schedule intervals

souro26:
### Summary
Adds validation to `Schedule` and `EventGenerator` to prevent infinite loops caused by non-positive intervals or counts. Previously, creating a `Schedule(interval=0)` or `Schedule(interval=-1)` would cause the event scheduler to repeatedly reschedule events at the same simulation time, hanging the model execution.

### Bug / Issue
The `mesa.time` event scheduling system allowed invalid `Schedule` configurations:
```python
Schedule(interval=0)   # Causes infinite loop
Schedule(interval=-1)  # Causes infinite loop
Schedule(count=0)      # Meaningless configuration
```

When used with `schedule_recurring()`, non-positive intervals result in events being rescheduled at the same time:
```python
next_time = model.time + 0  # Time never advances!
```

This causes `Model._advance_time()` to enter an infinite loop, continuously executing the same event without advancing simulation time.

### Implementation
Added validation in `Schedule.__post_init__()`:
- Fixed numeric intervals must be > 0
- `count` must be > 0 if provided

The validation ensures invalid schedules fail immediately at creation rather than causing model hangs during execution. No changes were made to event execution logic or time advancement behavior.

### Testing
Added comprehensive tests in `tests/test_schedule_run.py`:
- `Schedule(interval=0)` raises `ValueError`
- `Schedule(interval=-1)` raises `ValueError`  
- `Schedule(count=0)` raises `ValueError`
- `Schedule(count=-5)` raises `ValueError`

All existing scheduling tests continue to pass, confirming backward compatibility for valid configurations.

### Additional Notes
This change provides early failure detection for configuration errors that would otherwise manifest as hard-to-debug infinite loops during model execution.

### DIFF
--- a/mesa/time/events.py
+++ b/mesa/time/events.py
@@ -230,7 +230,6 @@ class EventGenerator:
             return self.schedule.interval(self.model)
         return self.schedule.interval
 
-
     def _should_stop(self, next_time: float) -> bool:
         """Check if the generator should stop before scheduling the next event."""
         return (


mesa/time/events.py
```diff
@@ -217,7 +227,12 @@ def execution_count(self) -> int:
     def _get_interval(self) -> float | int:
         """Get the next interval value."""
         if callable(self.schedule.interval):
-            return self.schedule.interval(self.model)
+            value = self.schedule.interval(self.model)
+            if value <= 0:
```
quaquel: I am not sure that this is an error for all callables. If you are drawing times between arivals, at least theoretically it can be the case that the arival time is on the same time step.

souro26: You’re right that zero delay events can make sense in discrete event simulation. The check I added is specifically for recurring generators. If a recurring interval evaluates to 0, `_execute_and_reschedule()` keeps scheduling the next event at the same timestamp, so `model.time` never advances and the event loop can run forever. So the guard is meant to prevent that case.

If you think zero-interval recurring schedules should be allowed, we’d probably need an extra safeguard in the generator to avoid infinite same-time rescheduling. i can adjust the code in that direction if that is preferred.

EwoutH:
This seems vast overkill. Can just be two lines of code, or am I missing something here?

mesa/time/events.py
quaquel: > If you think zero-interval recurring schedules should be allowed, 

In my view, it should be allowed. And I fail to see the need to protect against everything. It is Python after all.

quaquel:
> This seems vast overkill. Can just be two lines of code, or am I missing something here?

the post init validation is in essence two lines, so you are not missing anything. I just would remove the protection in the generator and call it fine.

souro26:
I removed the callable interval guard in the generator and kept validation in `Schedule.__post_init__` only. Is there anything else i should change?

PR #3291, opened 2026-02-13
Deprecate model.step() and run_model() with FutureWarning

falloficaruss:
### Summary
Adds deprecation warnings for `model.step()` and `model.run_model()` to support the Mesa 4 API transition toward event/time-based execution (`run_for`, `run_until`).

### Motive
Issue #3132: Add deprecation warnings to `model.step()` and `run_model()`
Without warnings, users continue using `step()`/`run_model()` without clear migration guidance, which makes future removal in Mesa 4 more disruptive

### Implementation
- Added `FutureWarning` in `Model._wrapped_step` when `model.step()` is called.
- Added `FutureWarning` in `Model.run_model()` when `model.run_model()` is called.
- Introduced internal helper `Model._step_once()` to advance one tick without emitting deprecation warnings.
  - `model.step()` warns, then calls` _step_once()`.
  - `run_model()` warns once, then loops via `_step_once()` to avoid warning spam each iteration.
- Updated tests in `test_model.py`:
  - Assert warning on `model.step()`.
  - Assert warning on `model.run_model()`.
  - Added dedicated deprecation-warning tests.
  - Added warning filtering for the looped time-increment test to keep intent focused on time behavior.

### Usage Examples
```python
from mesa.model import Model

model = Model()

# Deprecated in Mesa 3.x, will be removed in Mesa 4.0
model.step()        # emits FutureWarning
model.run_model()   # emits FutureWarning
```

Preferred Replacements:
```python
from mesa.model import Model

model = Model()

# Recommended API
model.run_for(1)
model.run_until(10)
```

### Additional Notes
- Warning category used: `FutureWarning` (consistent with other Mesa deprecations).
- Behavior is preserved; this change is additive (warning-only).
- `test_model.py` passes locally after updates.

### DIFF
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -202,6 +202,15 @@ class Model[A: Agent, S: Scenario](HasObservables):
 
     def _wrapped_step(self) -> None:
         """Advance time by one unit, processing any scheduled events."""
+        warnings.warn(
+            "model.step() is deprecated and will be removed in Mesa 4.0. Use model.run_for(1) instead.",
+            FutureWarning,
+            stacklevel=2,
+        )
+        self._step_once()
+
+    def _step_once(self) -> None:
+        """Advance time by one unit without deprecation warnings."""
         self._advance_time(self.time + 1)
 
     def _advance_time(self, until: float) -> None:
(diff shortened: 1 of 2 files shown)


falloficaruss:
The CI failures are expected from this deprecation change: the repository still has internal/test paths that call `model.step()` directly, and with warnings treated as errors those usages now fail.

I’ll open a follow-up PR to migrate internal usages to `run_for(1)` (and update tests accordingly). That migration PR should be reviewed/merged first, then this warning PR can land cleanly.

mesa/model.py
```diff
@@ -202,6 +202,15 @@ def __init__(
 
     def _wrapped_step(self) -> None:
         """Advance time by one unit, processing any scheduled events."""
+        warnings.warn(
```
quaquel, line 205: we have not decided yet on removing model.step, so this is premature.

The only thing is that we prefer calling model.run_for instead of having a for loop calling model.step.

PR #3292, opened 2026-02-13
Migrate internal step calls to run for

falloficaruss:
### Summary
Migrate internal framework/test call sites from deprecated `model.step()` loops to `model.run_for(1)` so the codebase no longer fails under warning-as-error after introducing deprecation warnings.

### Motive
#3291 added `FutureWarning` for `model.step()` / `run_model()` as part of Mesa 4 migration. CI then failed because Mesa internals and tests still relied on `model.step()` directly.
This follow-up PR removes those internal dependencies on deprecated APIs and unblocks the deprecation rollout.

### Implementation
- Updated batch execution loop in `batchrunner.py`:
  - while `model.running` and `model.steps < max_steps: model.step()`
changed to `model.run_for(1)` to use the non-deprecated progression API.
- Updated affected tests to use `run_for(1)` instead of direct `step()` calls where deprecation warnings were not the test subject:
  - `test_meta_agents.py`
  - `test_datacollector.py`
  - `test_lifespan.py`
  - `test_schedule_run.py`
 
Kept behavior equivalent (one tick advancement per iteration), only changing API entrypoints used internally.

### Usage Examples
Before:
```python
while model.running and model.steps < max_steps:
    model.step()
```

After:
```python
while model.running and model.steps < max_steps:
    model.run_for(1)
```

Before:
```python
for _ in range(5):
    model.step()
```

After:
```python
for _ in range(5):
    model.run_for(1)
```

### Additional Notes
- This PR is intended as a prerequisite/follow-up to #3291 
- Focus is migration, not behavior change.

### DIFF
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -16,13 +16,11 @@ def test_model_set_up():
     assert model.time == 0.0
     assert model._simulator is None
 
-    with pytest.warns(FutureWarning, match=r"model\.step\(\) is deprecated"):
-        model.step()
+    model.step()
     assert model.steps == 1
     assert model.time == 1.0
 
 
-@pytest.mark.filterwarnings("ignore::FutureWarning")
 def test_model_time_increment():
     """Test that time increments correctly with steps."""
     model = Model()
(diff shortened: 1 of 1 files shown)


quaquel:
I think this is largely premature. For mes 3.5, using a for loop with model.step is still fine and thus should be tested. Once we have a more stable via on the design of Mesa 4, we can reverse engineer what needs to be deprecated. Also, we still need to flesh out the conditional running.

PR #3300, opened 2026-02-13
Correct restart logic to properly reset execution count

souro26:
### Summary
This PR addresses and resolves an issue in the `EventGenerator` lifecycle, where event counts were not being reset properly after restarting a generator. The event handling logic was also improved to ensure events are handled consistently when restarted.

### Bug / Issue
The primary issue was with the `EventGenerator` where the execution count was not being reset upon restart, which caused the generator to behave inconsistently after it exhausted its count. Also, calling start() on an already active generator did not function as expected.

### Implementation
- The `execution_count` is now reset when the generator is restarted, ensuring that the generator fires events from the beginning again.
- Corrected the behavior of the start() method in the `EventGenerator` to properly handle the reset of the execution count and to avoid redundant event scheduling.

### Testing
- Added and updated tests to verify that the event generator restart behavior works correctly
- Ran all relevant tests to confirm that the changes did not break existing functionality. All tests passed successfully.

### DIFF
--- a/mesa/time/events.py
+++ b/mesa/time/events.py
@@ -278,7 +278,7 @@ class EventGenerator:
         self._active = True
         self._schedule_next(start_time)
         return self
-    
+
     def restart(self) -> EventGenerator:
         """Restart the event generator.
 


quaquel:
I don't agree with the basic premise of this PR. You are turning `start` into a restart in case of an exhausted generator. But that is, in my view, a behaviour change. And it is questionable whether that is desirable behaviour. Moreover, if you call stop and then start, you also reset the activation count. Again, that seems strange behavior to me.

souro26:
The goal of the change was to ensure the generator resets the `execution_count` after exhaustion, so it behaves predictably when restarted. Right now, after exhaustion, there's no way to reuse a generator without creating a new instance.

I see your point about stop() and start() resetting the count. I will add a separate restart() method instead that explicitly resets `execution_count` and begins fresh and keep start() unchanged.

mesa/time/events.py
```diff
@@ -353,7 +374,7 @@ def pop_event(self) -> Event:
 
     def is_empty(self) -> bool:
         """Return whether the event list is empty."""
-        return len(self) == 0
+        return not any(not e.CANCELED for e in self._events)
```
quaquel, line 377: this is unrelated to this PR. Please remove.

quaquel:
I fail to see the need for a restart. The generator is a lightweight class that is easy to create. Moreover, if the generator has a `schedule.start` or `schedule.end`, having a restart is utterly pointless and thus a major source of bugs and weird behavior. I am closing this PR.

souro26:
Thanks for the review. I understand where i went wrong with my thinking. I agree the generator could easily be created again since its lightweight. I apologize for the unnecessary PR.

PR #3303, opened 2026-02-14
Data collector for batchrunner

### LINKED ISSUE #3296: BatchRunner and Visualization Require DataCollector, Blocking DataRegistry Usage
In Mesa’s newer data collection APIs, I noticed that core components still hard-depend on the legacy `DataCollector`, which makes it difficult to use `DataRegistry` on its own.

For example, `BatchRunner` explicitly fails if `model.datacollector` is missing:
```python
# mesa/batchrunner.py
if not hasattr(model, "datacollector"):
    raise AttributeError("The model does not have a datacollector attribute.")
```
Core visualization components make the same assumption:
```python
# mesa/visualization/components/altair_components.py
df = model.datacollector.get_model_vars_dataframe()
At the same time, the model already initializes the newer system:
```
```python
# mesa/model.py
self.data_registry = DataRegistry()
```
This currently forces users running batch experiments or long simulations to rely on the in-memory `DataCollector`, even though more scalable backends (Parquet, SQL, etc.) already exist under `mesa.experimental.data_collection`.

Ideally,` BatchRunner` and core visualizations should be able to use `DataRegistry `when it’s available, while still supporting `DataCollector` for backward compatibility. A simple success case would be that a model using only `DataRegistry ` can run batch experiments and standard visualizations without errors.

@quaquel I would like to work on this!! if it aligns with the project’s direction.

ShreyasN707:
This PR Prevents crashes in `BatchRunner` and visualization components when a model uses `DataRegistry` instead of `DataCollector`. Ensures graceful handling while preserving legacy behavior.

Fixes #3296.

### Bug / Issue
Core components assumed `model.datacollector` always exists.  
For `DataRegistry`-only models, this caused `AttributeError` in:

- `BatchRunner._collect_data`
- `PlotAltair`
- `PlotMatplotlib`

### Implementation

**BatchRunner**

```python
if hasattr(model, "datacollector"):
    dc = model.datacollector
elif hasattr(model, "data_registry"):
    return {}, []
else:
    raise AttributeError(...)
Allows batch runs to proceed without crashing (no-op collection for registry-only models).

Visualization (Altair & Matplotlib)

if not hasattr(model, "datacollector"):
    if hasattr(model, "data_registry"):
        raise NotImplementedError(
            "Plotting with DataRegistry is not yet supported."
        )
    raise AttributeError(...)

```
No schema inference or fallback logic added.

Testing
- Updated batchrunner coverage tests.

- Full test suite passes (531 passed, 0 failed).

### DIFF
--- a/docs/tutorials/7_visualization_propertylayer_visualization.ipynb
+++ b/docs/tutorials/7_visualization_propertylayer_visualization.ipynb
@@ -66,7 +66,7 @@
     "# Check Mesa version for visualization compatibility\n",
     "if mesa.__version__.startswith((\"3.0\", \"3.1\", \"3.2\")):\n",
     "    print(\n",
-    "        f\"⚠️: Mesa {mesa.__version__} detected. Visualization features require Mesa 3.3+\"\n",
+    "        f\"⚠️  Mesa {mesa.__version__} detected. Visualization features require Mesa 3.3+\"\n",
     "    )\n",
     "    print(\"To upgrade: pip install --upgrade mesa\")\n",
     "\n",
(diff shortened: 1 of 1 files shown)


quaquel:
The data registry is still in experimental, and in mesa 4, we plan to fully remove the batch runner. I don't see a lot of value in this PR. Last, what matters is the recorder, not the registry.

ShreyasN707:
I raised this mainly to prevent unexpected crashes when DataRegistry is present, but I understand the limited value if BatchRunner is being phased out.
Please let me know if there’s a more appropriate area to contribute around the recorder.

PR #3312, opened 2026-02-15
Only emit signal of new value of observable is different from old value

### LINKED ISSUE #3302: `DataRecorder` Captures Stale Data Due to Redundant Observable Triggers in `_advance_time` of `model.py`
### Describe the bug

The `DataRecorder` captures incorrect intermediate states during event execution instead of the final state at each timestep. This occurs because `model.time` triggers the Observable signal **before every event** even when time hasn't changed.

When multiple events occur at the same simulation time, the `DataRecorder` captures data multiple times - once before each event executes. The final capture overwrites earlier ones, resulting in data that represents the state **after some events but before others** at that timestep.

### Expected behavior

For the MRE below:
- Time 0.0: wealth should be 0 (initial state)
- Time 1.0: wealth should be 11 (after step: +1, then trade: +10)
- Time 2.0: wealth should be 22 (after step: +1, then trade: +10)
- Time 3.0: wealth should be 43 (after step: +1, then scheduled trade: +10, then recurring trade: +10)
- Time 4.0: wealth should be 54 (after step: +1, then trade: +10)

### Actual behavior

```plaintext
   unique_id  wealth  time
0          1       0   0.0
1          1       1   1.0
2          1      12   2.0
3          1      23   3.0
4          1      44   4.0
```

The DataRecorder shows values that suggest it's capturing **after `step()` but before `trade()`**.

## To Reproduce

```python
from mesa import Model, Agent
from mesa.time import Schedule
from mesa.experimental.data_collection import DataRecorder

class MyAgent(Agent):
    def __init__(self, model):
        super().__init__(model)
        self.wealth = 0
    
    def step(self):
        self.wealth += 1 

class MyModel(Model):
    def __init__(self):
        super().__init__()
        MyAgent.create_agents(self,1)
        self.data_registry.track_agents(self.agents, "Wealth", "wealth")
            agent_reporters={"Wealth": "wealth"},
        )
        self.recorder = DataRecorder(self)
        self.schedule_event(self.trade, at=3.0)
        self.schedule_recurring(
            self.trade,
            Schedule(interval=1)
        )
    
    def step(self):

    def trade(self):
        for agent in self.agents:
            agent.wealth += 10

model = MyModel()

model.run_until(4)

print(f"New DataRecorder: \n{model.recorder.get_table_dataframe('Wealth').to_string()}\n")
print(f"Old DataCollector: \n{model.datacollector.get_agent_vars_dataframe().to_string()}")
```

_Note: Using either `schedule_recurring()` or `schedule_event()` will also reproduce the bug_ 

**MRE Output:**
```plaintext
New DataRecorder: 
   unique_id  wealth  time
0          1       0   0.0
1          1       1   1.0
2          1      12   2.0
3          1      23   3.0
4          1      44   4.0

Old DataCollector: 
              Wealth
Step AgentID        
0.0  1             0
1.0  1            11
2.0  1            22
3.0  1            43
4.0  1            54
```

### Root Cause Analysis

The bug is in `model.py` in the `_advance_time()` method (line 220):

https://github.com/mesa/mesa/blob/744df2a75b1df62c8a5dc8924012d35033155c53/mesa/model.py#L213-L224

### Proposed Fix

```python
    if event.time <= until: 
         if self.time != event.time
             self.time = event.time 
         event.execute()
```

quaquel:
This PR makes a small change to Observable such that it only emits a signal if the new value is different from the old value. This is the expected behavior of an observable. It also closes #3302.

### DIFF
--- a/tests/experimental/test_datarecorder.py
+++ b/tests/experimental/test_datarecorder.py
@@ -779,7 +779,7 @@ def test_parquet_recorder_list_data_storage():
         recorder.clear()
 
         model.step()
-        recorder.collect() # force collect after step
+        recorder.collect()  # force collect after step
 
         # Check list was stored correctly
         df = recorder.get_table_dataframe("agent_data")
@@ -1030,7 +1030,7 @@ def test_recorder_start_time_behavior():
     # Step to start_time
     model.step()  # t=1 (should not collect)
     model.step()  # t=2 (should not collect)
-    model.step() # the change to t=3 triggers a single collect for t=2
+    model.step()  # the change to t=3 triggers a single collect for t=2
 
     df = recorder.get_table_dataframe("model_data")
     times = df["time"].unique()
--- a/tests/experimental/test_mesa_signals.py
+++ b/tests/experimental/test_mesa_signals.py
@@ -36,11 +36,10 @@ def test_observables():
     agent.observe("some_attribute", ObservableSignals.CHANGED, handler)
 
     agent.some_attribute = 10
-    handler.assert_not_called() # we change it to the same value so no signal
+    handler.assert_not_called()  # we change it to the same value so no signal
 
     agent.some_attribute = 20
     handler.assert_called_once()
-
 
 
 def test_HasObservables():


EwoutH:
Thanks, I assume you want this in 3.5? In that case, please merge ASAP.

PR #3314, opened 2026-02-15
Use scenario for all examples in benchmarks

### LINKED ISSUE #3307: Broken BoltzmannWealth in benchmarks
Recent merged PR #3178 has modified the model signature of `BoltzmannWealth` Model which is failing the `global_benhcmarks.py`
```plaintext
07:05:03 starting benchmarks.
  File "/home/aman/mesa/benchmarks/global_benchmark.py", line 90, in <module>
    results = run_experiments(model, config)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/aman/mesa/benchmarks/global_benchmark.py", line 71, in run_experiments
    run_model(model_class, seed, config["parameters"])
  File "/home/aman/mesa/benchmarks/global_benchmark.py", line 34, in run_model
    model = model_class(rng=seed, **parameters)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I see two solution:
1. Revert the BoltzmannWealth to match old signature
2. Modify all benchmarks examples to use `Scenario` and then update `global_benchmarks.py`

codebreaker32:
This PR refactors the core examples used in the global performance benchmarks to utilize the `Scenario`.

### Changes Made
* Updated the `Schelling`, `BoidFlockers`, and `WolfSheep` models to accept a `Scenario` object upon initialization rather than individual parameters.
* Implemented specific scenario classes (e.g., `SchellingScenario`, `BoidsScenario`) for each model.
* Modified `global_benchmarks.py` to dynamically unpack configuration parameters into `Scenario` objects during the run.

Fixes #3307

### DIFF
--- a/tests/examples/test_examples.py
+++ b/tests/examples/test_examples.py
@@ -75,6 +75,7 @@ def test_schelling_model():  # noqa: D103
 
     app.page  # noqa: B018
 
+    _model = Schelling(scenario=None)
     model = Schelling(scenario=SchellingScenario(rng=42))
     ref = weakref.ref(model)
 
@@ -106,6 +107,8 @@ def test_boid_flockers():  # noqa: D103
     from mesa.examples.basic.boid_flockers import app  # noqa: PLC0415
 
     app.page  # noqa: B018
+
+    _model = BoidFlockers(scenario=None)
 
     model = BoidFlockers(scenario=BoidsScenario(rng=42))
     ref = weakref.ref(model)
@@ -170,6 +173,8 @@ def test_wolf_sheep():  # noqa: D103
     from mesa.examples.advanced.wolf_sheep import app  # noqa: PLC0415
 
     app.page  # noqa: B018
+
+    _model = WolfSheep(scenario=None)
 
     model = WolfSheep(scenario=WolfSheepScenario(rng=42))
     ref = weakref.ref(model)


EwoutH:
Thanks for the PR.

As described (on a weird location) in our [examples Readme](https://github.com/mesa/mesa/blob/main/mesa/examples/README.md), we make a distinction between basic and advanced examples:

> The examples are categorized into two groups:
> 1. **Basic Examples** - Simpler models that use only stable Mesa features; ideal for beginners
> 2. **Advanced Examples** - More complex models that demonstrate additional concepts and may use some experimental features

Since `Scenario` is still an experimental feature, it's too soon to update the basic examples. So ideally we would update the benchmark script to be able to handle them.

quaquel:
In #3307, I suggested fixing it in this way. The main reason for this is that in Mesa 4, we want `scenario` to become the default anyway, so we'd better start benchmarking it accordingly.

EwoutH:
Sure, we're targeting 4.0.0 now anyways.

benchmarks/global_benchmark.py
```diff
@@ -61,18 +63,23 @@ def run_experiments(model_class, config):
...
     init_times = []
     run_times = []
+
+    steps = config["steps"]
+    base_params = config["parameters"]
+
     for seed in range(1, config["seeds"] + 1):
         fastest_init = float("inf")
         fastest_run = float("inf")
+        run_kwargs = {**base_params, "seed": seed}
```
quaquel:
```suggestion
        run_kwargs = {**base_params, "rng": seed}
```

this is probably all that is needed to make it run correctly

codebreaker32:
> benchmarks don't yet run, see my suggested/suspected fix.

Its expected, because main branch still has broken code. However I am trying to fix the other errors

codebreaker32:
Fixed the `configurations.py`. Now it is running fine locally

quaquel (approved):
Looks good to me

PR #3317, opened 2026-02-15
Fix unbounded growth of `Model._event_generators`

### LINKED ISSUE #3313: `Model._event_generators` grows unbounded
**Describe the bug**
`Model._event_generators` continues to grow while running a model, and also will contain generators that are exhausted. It is a minor bug, but worth addressing at some point. But since a generator also holds a hard ref to the callable, it might, for example, prevent agents from being garbage collected. 

**Expected behavior**
Whenever a generator is exhausted (or stopped?), it should remove itself from `model._event_generators`. 

**proposed direction for fix**
There are a couple of directions:

1. Turn `._event_generators` into a set, and whenever a generator is exhausted, it removes itself from this set. The only function of `_event_generators` is to serve as a hard ref to generators, so ordering does not matter. This fix would also imply that the responsibility of adding a generator to `schedule_recurring` should move from `schedule_recurring` into the `EventGenerator` class, so all instances are in the set.
2. Make it the responsibility of the caller of `schedule_recurring` to maintain a hard ref to the generator. This is tricky for users, so less ideal.
3. To reduce the severity of the leak, we should consider turning self.function into a weakref similar to how it is done for `Event`, and if the ref becomes None, this should automagically stop the generator.

EwoutH:
### Summary
`Model._event_generators` grows unbounded because exhausted and stopped `EventGenerator`s are never removed.

### Bug / Issue
`_event_generators` exists solely to hold strong references to active generators, preventing garbage collection. However, generators were never removed when they finished or were stopped, causing the list to grow indefinitely. Since generators hold hard references to their callables (which may be bound agent methods), this could also prevent agents from being garbage collected.

Closes #3313.

### Implementation
- Changed `_event_generators` from `list` to `set` (ordering is irrelevant for a GC-prevention container).
- Moved registration responsibility into `EventGenerator`: `start()` adds itself to the set, `stop()` and exhaustion remove it via `discard()`.
- Removed the `_event_generators.append()` call from `Model.schedule_recurring`, since `start()` now handles it.

### Testing
Existing tests pass. Could add a test verifying that a generator with `count=n` is removed from `_event_generators` after completing its executions.

### Additional Notes
No public API changes, `_event_generators` is a private attribute.

### DIFF
--- a/mesa/model.py
+++ b/mesa/model.py
@@ -129,7 +129,7 @@ class Model[A: Agent, S: Scenario](HasObservables):
         # Event list for event-based execution
         self._event_list: EventList = EventList()
         # Strong references to active EventGenerators (prevent GC)
-        self._event_generators: list[EventGenerator] = []
+        self._event_generators: set[EventGenerator] = set()
 
         # check if `scenario` is provided
         # and if so, whether rng is the same or not
@@ -450,7 +450,6 @@ class Model[A: Agent, S: Scenario](HasObservables):
         """
         generator = EventGenerator(self, function, schedule, priority)
         generator.start()
-        self._event_generators.append(generator)
         return generator
 
     def run_for(self, duration: float | int) -> None:
--- a/mesa/time/events.py
+++ b/mesa/time/events.py
@@ -252,6 +252,7 @@ class EventGenerator:
         else:
             self._active = False
             self._current_event = None
+            self.model._event_generators.discard(self)
 
     def _schedule_next(self, time: float) -> None:
         """Schedule the next event at the given time."""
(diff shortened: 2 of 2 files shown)


EwoutH:
When designing this part I thought this would be a problem, because now an EventGenerator can't be restarted anymore. But that's not true, because:
- If you create a reference you can use that to restart it
- If you didn't create a reference yourself you couldn't apply start to it anyways

So this nicely solves itself.

quaquel:
> If you create a reference you can use that to restart it

What do you mean by restarting?

EwoutH:
Calling .start() after you have called .stop()

quaquel:
Fair enough.

PR #3318, opened 2026-02-15
Remove seed kwarg

quaquel:
Seed was deprecated in Mesa 3. This PR removes it as an argument to `model.Model()`, cleans up the inner workings of `model.Model` related to `seed`, and modifies `reset_rng` to also reset `model.random`. Last, it updates all tutorials to use `rng`.

**Note**
I focused on just removing the deprecated seed argument. There is still a need for a separate decision on whether to keep rng as a separate argument or force users to always use Scenario or a Scenario subclass.


mesa/model.py
```diff
@@ -353,38 +323,38 @@ def run_model(self) -> None:
...
+        try:
+            self.random = random.Random(rng)
+        except TypeError:
+            seed = int(self.rng.integers(np.iinfo(np.int32).max))
+            self.random = random.Random(seed)
         else:
-            self.rng = np.random.default_rng(rng)
-            self._rng = self.rng.bit_generator.state
+            seed = rng
+        self._seed = seed  # this allows for reproducing stdlib.random
```
EwoutH, line 357: Is it correct this is still called `seed`?

mesa/model.py
```diff
@@ -353,38 +323,38 @@ def run_model(self) -> None:
...
+        if failed:
+            self.rng: np.random.Generator = np.random.default_rng(rng)
+
+        self._rng = self.rng.bit_generator.state  # this allows for reproducing the rng
+
+        try:
+            self.random = random.Random(rng)
+        except TypeError:
+            seed = int(self.rng.integers(np.iinfo(np.int32).max))
+            self.random = random.Random(seed)
```
EwoutH, line 354: This whole dance seems quite complex, do we still need all this?

mesa/model.py
quaquel: It is the seed value that was used to initialize the random.Random instance. This might well be different from the value that was stored in `_rng`, so yes I think it makes sense to keep it like that.

mesa/model.py
quaquel: The current behavior of this method is to restore the original generator if `rng is None`. If we want to change the behavior of this method, we can simplify this code. If we want to keep the current behavior, this complexity is unavoidable.

EwoutH: Let's keep that in mind, for a separate discussion/PR.

EwoutH:
I wrote something up on how to handle this.

- https://github.com/mesa/mesa-examples/pull/328

quaquel:
I opened a pr on mesa-examples for fixing this. I am merging this PP in the meantime.

PR #3320, opened 2026-02-16
Fail fast on unsupported lambda/partial callbacks in schedule_event

### LINKED ISSUE #3319: `schedule_event` silently drops inline lambda callbacks before execution
**Describe the bug**
silent acceptance + later no-op due to weakref.

**Expected behavior**
`schedule_event(lambda ...)` should raise immediately with a clear error.

**To Reproduce**
```python
from mesa import Model

m = Model()
state = {"x": 0}

m.schedule_event(lambda: state.__setitem__("x", state["x"] + 1), at=1.0)
m.run_for(2.0)

print(state["x"])  # observed: 0, expected: 1
```
**Reproduction steps:**
1. Create a Model.
2. Schedule an event with an inline lambda using `at=...` (or `after=...)`.
3. Advance time with `run_for(...)`.
4. Observe that side effects from the lambda do not occur.

**Additional context**
- This appears related to event callbacks being stored via weak references in mesa.time.events.Event.
- Named functions with a strong reference execute as expected.

falloficaruss:
### Summary
Aligns `Event` callback handling with the weakref-based design: callbacks are accepted based on weakref capability (not callable type), and callback lifetime semantics are explicitly documented.

### Bug / Issue
Closes #3319 
The final agreed behavior is to preserve weakref semantics: callbacks may no-op if the target is garbage-collected before execution.
This PR updates implementation/tests/docs to match that contract and avoids type-based rejection of valid callables like `functools.partial`.

### Implementation
Updated `events.py`:

  - Added a shared module-level helper to create callback references (`_create_callable_reference`).
  - Centralized callback reference creation in one place and reused it in `__setstate__`.
  - Preserved weakref behavior for callbacks (including bound methods via `WeakMethod`).
  - Removed type-based lambda/partial rejection.
  - Added explicit error path for callable objects that cannot be weak-referenced.
  - Updated Event notes/docstring to reflect callback lifetime semantics.

Updated tests:

  - `test_schedule_run.py`
    - Replaced rejection tests with coverage showing lambda/partial callbacks execute when a strong reference is retained.
  - `test_devs.py`
    - Added execution coverage for lambda/partial callbacks with strong refs.

### Testing
- `test_schedule_run.py`: 22 passed
- `test_dev.py`: 23 passed

### Additional Notes
- This PR does **not** introduce strong-reference retention for callbacks.
- Runtime behavior remains weakref-driven by design; if the callback target is collected, execution is skipped silently.
- Main change is contract alignment + capability-based validation.


quaquel:
It was a deliberate design choice not to allow lambdas, and this behavior is documented. I would need strong reasons to change my mind on this.

falloficaruss:
> It was a deliberate design choice not to allow lambdas, and this behavior is documented. I would need strong reasons to change my mind on this.

My concern is only that `schedule_event(lambda ...)` is currently accepted and then fails silently later. I suggest failing fast at scheduling time with a clear error, while keeping the existing weakref behavior for supported callables.

quaquel:
> My concern is only that schedule_event(lambda ...) is currently accepted and then fails silently later. I suggest failing fast at scheduling time with a clear error, while keeping the existing weakref behavior for supported callables.

But that is not what this PR does. It creates an internal hard ref for weak references. Also, the same issue as lambdas also holds potentially for partial functions. 

I am potentially open to accepting that lambdas and partials raise a value error when passed to an event.

falloficaruss:
> 
> I am potentially open to accepting that lambdas and partials raise a value error when passed to an event.

I suggest fail-fast validation instead: reject unsupported ephemeral callables at event creation (at least lambdas and `functools.partial`) with a clear `ValueError`, while preserving the existing weakref behavior for supported callables.

What are your thoughts?

quaquel:
>  suggest fail-fast validation instead: reject unsupported ephemeral callables at event creation (at least lambdas and functools.partial) with a clear ValueError, while preserving the existing weakref behavior for supported callables.

This is exactly what I was trying to suggest.

falloficaruss:
Should I refactor this PR then?

quaquel:
Thanks for the update. I have been looking at this issue myself as well for some other reasons.

1. I would move the check into a separate function rather than a static method. We need the validation in other places as well.
2. I am still wondering about partials. At the moment, we don't allow args and kwargs in Event, so if we also don't allow partials, we severely limit the functions and methods that can be scheduled. Is there a way to check if a partial function has been assigned to a variable? The other option is to allow args and kwargs on Event.

falloficaruss:
There is no way to check whether a partial was “assigned to a variable” in a way that guarantees safe lifetime.
So the reliable way might be to  “allow args/kwargs on Event usage path” (or expose them via `schedule_event`), not trying to detect variable assignment for partials.

quaquel:
The other option is to issue a warning if the weakref resolves to None. Warnings in general are issued only once, but it would allow users to more quickly diagnose that something might be wrong.

quaquel:
I would suggest approaching this slightly differently:

```python
        if isinstance(function, MethodType):
            function = WeakMethod(function)
        else:
            function = ref(function)

        if function() is None:
            # error message needs work
            raise TypeError(function is not weakly referencable ...)

        self.fn = function

```

So rather than checking the type and everything, we just check if we can take a weakref to the function at creation. If we can do that, we accept it. Otherwise, we raise an error. In this way, partials assigned to a variable are still accepted.

falloficaruss:
I like the direction, but I want to flag one concern: validating weakref resolution at creation time is only a point-in-time check. A callback can still be GC’d before execution, so we may still get nondeterministic no-ops.
Would you be comfortable making that an explicit contract (“caller must keep a strong reference alive”), or should we prefer stricter fail-fast rules to avoid that class of runtime ambiguity?
If we go with the weakref-capability approach, I’d suggest also keeping a runtime warning when `fn()` resolves to `None` so failures are diagnosable.

quaquel:
> A callback can still be GC’d before execution, so we may still get nondeterministic no-ops.

Yes, but that is desirable behavior. Imagine you schedule events for an agent. Somewhere along the way, this agent is removed from the simulation. All events that call this agent now must either be removed or fail silently. With weakrefs, we get silent failures. 

Not sure about the warning. I think it might be more confusing than helpful.

mesa/time/events.py
```diff
@@ -33,6 +33,22 @@
...
 
 
+def _create_callable_reference(function: Callable):
+    """Create a weak reference wrapper for an event callback."""
+    if isinstance(function, MethodType):
+        function_ref = WeakMethod(function)
+    else:
+        try:
+            function_ref = ref(function)
+        except TypeError as exc:
```
quaquel, line 50: when can this happen?

quaquel:
Can you update the starting post to align with the final implementation?

mesa/time/events.py
falloficaruss: This can happen when the callback is callable but not weak-referenceable (e.g., `len`, or a callable class using `__slots__` without `__weakref__`). It won’t happen for normal Python functions/bound methods, but the guard keeps the error explicit for those edge callables.

mesa/time/events.py
```diff
@@ -94,12 +106,7 @@ def __init__(
         self.priority = priority.value
         self._canceled = False
 
-        if isinstance(function, MethodType):
-            function = WeakMethod(function)
-        else:
-            function = ref(function)
-
-        self.fn = function
+        self.fn = _create_callable_reference(function)
```
quaquel: This is still not the behavior I suggested. I want to check if the return of `_create_callable_reference` returns None in the init, and if so, raise a ValueError. So if we know at the init of the Event that the function is not a valid weakref, we fail immediately. If it fails later, we pass silently. 


```suggestion
        weak_ref_fn = _create_callable_reference(function)
        if weak_ref_fn is None:
            raise ValueError("function must be weak referenceable at Event creation.")
        self.fn = weak_ref_fn
        
        
```

quaquel:
The behavior is still not exactly what I have in mind. Let me try to explain it

- [x]  If function is not a callable --> raise a TypeError
- [x]  If function is not weakly referenceable --> raise a TypeError
- [ ] If weakref to function resolves to None when Event is instantiated, raise a ValueError

This last one is the sticky one and also in my own testing I can't seem to get it right. Basically, `Event(5, lambda:5)` or `Event(5, partial(my_func, 1, 2,3))` should raise a ValueError. While

```python

a = partial(my_func, 1,2,3)
Event(5, a)

```

should run fine.

falloficaruss:
Thanks, this helps clarify the intended behavior.
I think we can implement the “tricky” case by doing the liveness check after dropping the local strong reference in `Event.__init__`:

if not callable -> `TypeError`
if not weakrefable -> `TypeError`
create weakref, drop local strong ref, then check `wr() is None -> ValueError`
That should make `Event(5, lambda: 5)` and inline `Event(5, partial(...))` fail, while `a = partial(...); Event(5, a)` still works.

falloficaruss:
We hit cross-platform instability with the current inline-`partial` check.  
The logic depends on `sys.getrefcount`, and that value changes across CPython builds/OS/test environments, so a variable-bound `partial` is sometimes misclassified and rejected.

So the check is not deterministic in CI.  
To keep behavior stable, I propose we remove refcount-based `partial` detection and keep only deterministic guards:
- callable required
- weak-referenceable required
- lambda rejected at creation
- dead weakref at creation rejected

This keeps fail-fast behavior without flaky platform-dependent failures.

mesa/time/events.py
```diff
@@ -95,12 +116,17 @@ def __init__(
         self.priority = priority.value
         self._canceled = False
 
-        if isinstance(function, MethodType):
-            function = WeakMethod(function)
-        else:
-            function = ref(function)
+        weak_ref_fn = _create_callable_reference(function)
+        if weak_ref_fn() is None or _is_lambda_callback(function):
```
quaquel: the first if clause will never fail because function still holds the hard ref

mesa/time/events.py
```diff
@@ -33,6 +34,24 @@
...
+        except TypeError as exc:
+            raise TypeError("function must be weak referenceable") from exc
+
+    return function_ref
+
+
+def _is_lambda_callback(function: Callable) -> bool:
+    """Return True when callback is a Python lambda function."""
+    return isinstance(function, types.FunctionType) and function.__name__ == "<lambda>"
+
```
quaquel: can't this be integrated into a single function?

mesa/time/events.py
falloficaruss: Thanks! Good Catch.

mesa/time/events.py
falloficaruss: Yes. I'll update the PR

quaquel:
code wise, it seems fine now. I want to find some time to test a few things manually on this brach before approving it.

mesa/time/events.py
```diff
@@ -88,19 +110,19 @@ def __init__(
...
-            function = WeakMethod(function)
-        else:
-            function = ref(function)
+        weak_ref_fn = _create_callable_reference(function)
+
+        # Drop Event.__init__'s local strong reference and verify the callback
+        # still has an external owner at creation time.
+        function = None
+        if weak_ref_fn() is None:
+            raise ValueError("function must be alive at Event creation.")
```
quaquel: This code is currently not tested. I assume this is because of differences of how gc behaves across OS's. 

But then, might it not be an idea to move this check out of Event and into model.schedule_event and see if that behaves more reliably across OSs.

falloficaruss: I’d prefer to keep validation in `Event` (single source of truth) rather than split behavior between `Event` and `schedule_event`. If the post-drop weakref check is not reliably testable, we can remove that branch and keep only deterministic creation-time validation

quaquel:
the docs are breaking because the event scheduling tutorial is breaking because it uses lambda functions. Can you take a look at fix this?

falloficaruss:
Fixed it.

quaquel (approved):

EwoutH:
@quaquel backport?

PR #3323, opened 2026-02-16
Modify Space Drawers to use explicit Cell positions

codebreaker32:
This PR updates the `mesa.visualization.space_drawers` to leverage the newly introduced `cell.position` attribute for the DiscreteSpace API.

Previously, Space Drawers had to independently guess or mathematically recalculate the physical geometry of spaces (e.g., computing hex offsets or running network layout algorithms).

### Key Changes

1. `HexSpaceDrawer`

* `viz_xmin`, `viz_xmax`, `viz_ymin`, and `viz_ymax` are now dynamically calculated directly from the absolute minimum and maximum values of the cells' physical positions.
* `_get_hexmesh` now anchors hexagons exactly at `cell.position` instead of relying on a hardcoded row/col math algorithm.
* Corrected the return type of `_get_hexmesh` from `list[tuple[float, float]]` to the factually accurate `list[list[tuple[float, float]]]` (a list of hexagons).

2. `NetworkSpaceDrawer`

* The drawer now intercepts pre-calculated network node coordinates directly from the cells (`cell._position`). This prevents the visualizer from redundantly executing heavy NetworkX layout algorithms (like `nx.spring_layout`) if the model has already calculated them during setup.

3. `VoronoiSpaceDrawer`

* Uses `cell.position` instead of centroid coordinates

_Note: Full backward compatibility has been maintained for `mesa.space` objects via fallbacks_

### DIFF
--- a/mesa/examples/advanced/pd_grid/agents.py
+++ b/mesa/examples/advanced/pd_grid/agents.py
@@ -35,9 +35,6 @@ class PDAgent(CellAgent):
         best_neighbor = max(neighbors, key=lambda a: a.score)
         self.next_move = best_neighbor.move
 
-        if self.model.activation_order != "Simultaneous":
-            self.advance()
-
     def advance(self):
         self.move = self.next_move
         self.score += self.increment_score()
--- a/mesa/examples/advanced/pd_grid/model.py
+++ b/mesa/examples/advanced/pd_grid/model.py
@@ -62,12 +62,11 @@ class PdGrid(mesa.Model):
         # Activate all agents, based on the activation regime
         match self.activation_order:
             case "Sequential":
-                self.agents.do("step")
+                self.agents.do(lambda a: (a.step(), a.advance()))
             case "Random":
-                self.agents.shuffle_do("step")
+                self.agents.shuffle_do(lambda a: (a.step(), a.advance()))
             case "Simultaneous":
-                self.agents.do("step")
-                self.agents.do("advance")
+                self.agents.do("step").do("advance")
             case _:
                 raise ValueError(f"Unknown activation order: {self.activation_order}")
 


codebreaker32:
Do we need fallback if we are removing `mesa.space`?

quaquel:
> Do we need fallback if we are removing mesa.space?

No, I'll try to open a pr later this week for removing mesa.space.

quaquel:
In #3337, I removed mesa.space. Can you update this PR accordingly. Code wise, I have given it a quick check already and did not see anything strange.

codebreaker32:
All Set! I have updated the PR

mesa/visualization/space_drawers.py
```diff
@@ -193,29 +192,27 @@ def __init__(self, space: HexGrid):
...
+
         self.x_spacing = np.sqrt(3) * size
         self.y_spacing = 1.5 * size
 
-        x_max = self.space.width * self.x_spacing + (self.space.height % 2) * (
-            self.x_spacing / 2
-        )
-        y_max = self.space.height * self.y_spacing
+        positions = np.array([cell.position for cell in self.space.all_cells])
+        x_min, y_min = positions.min(axis=0)
```
quaquel: space has dimensions, so you you just use those instead of deriving it from the position of the cells.

codebreaker32:
> Have you visually checked everything?

Yes, On `virus_on_network`

codebreaker32:
I ran `virus_on_network` and I think something is wrong However its not due to this PR because its persistent on `main` as well


Screenshot (210)


See "infected count isn't reflected well in viz"

codebreaker32:
Remaining examples are working fine.

quaquel:
There is more weird stuff going on with networks. If you reset the model, the visual style is also changed. Initially, it is a grey circle with a white center. On reset, it becomes a grey center. @Sahil-Chhoker, any thoughts?

quaquel:
I have started to dig into this. The problem is in `SpaceRenderer._map_coordinates`. In the if clause for networks, stuff is mapped from node ids to positions, but of course we don't need to do that anymore if position is already stored in the cell.

codebreaker32:
> I have started to dig into this. The problem is in `SpaceRenderer._map_coordinates`. In the if clause for networks, stuff is mapped from node ids to positions, but of course we don't need to do that anymore if position is already stored in the cell.

This doesn't seem to solve the issue

quaquel:
If I do 

```python
        elif isinstance(self.space, Network):
            # no need to map network node IDs to positions 
            mapped_arguments["loc"] = arguments["loc"].astype(float)
```

the issue is gone

codebreaker32:
Screenshot (215)

codebreaker32:
Ok, My bad. I was running it on `main` branch. Running fine on this branch

codebreaker32:
I have opened a PR to fix it. That can be merged after this one.

mesa/visualization/space_drawers.py
```diff
@@ -204,18 +204,16 @@ def __init__(self, space: HexGrid):
...
         y_padding = size
 
-        self.hexagons = self._get_hexmesh(self.space.width, self.space.height, size)
+        self.hexagons = self._get_hexmesh(size)
 
         # Parameters for visualization limits
         self.viz_xmin = -1.8 * x_padding
         self.viz_xmax = x_max
         self.viz_ymin = -1.8 * y_padding
         self.viz_ymax = y_max
```
quaquel, line 213: is all of this still needed of can some of this be simplified?

mesa/visualization/space_drawers.py
```diff
@@ -375,7 +368,12 @@ def __init__(
 
         # gather locations for nodes in network
         self.graph = self.space.G
-        self.pos = self.layout_alg(self.graph, **self.layout_kwargs)
+
+        self.pos = {}
+
+        for node_id, cell in self.space._cells.items():
+            if getattr(cell, "_position", None) is not None:
+                self.pos[node_id] = cell.position
```
quaquel, line 376: can't this be simplified to a numpy array. We don't need the dictionary given the other PR to be merged after this one.

mesa/visualization/space_drawers.py
codebreaker32: Yes, it is still strictly needed to draw the space properly because `viz_xmin`, `viz_xmax`, `viz_ymin`, and `viz_ymax` are required by the `BaseSpaceDrawer` as well as by `draw_x` methods to define the viewport
However, we can simplify this if we use the `cell.position` approach.
Previously in this PR, I used this approach to find the bounds. see https://github.com/mesa/mesa/pull/3323/changes/0bc8babd438f1e20cce9a294fe8d83fe987a2ccd.
If we revert to that approach, we can delete manual spacing logic.

mesa/visualization/space_drawers.py
codebreaker32: We need the dictionary because of `NetworkX` here. In `draw_matplotlib`, we pass `self.pos` directly into `nx.draw_networkx_nodes` and `nx.draw_networkx_edges`, which strictly require a mapping

mesa/visualization/space_drawers.py
quaquel: I am fine either way, so if you prefer the current version, that's fine with me.

codebreaker32: I reverted it to the current version after your suggestion(see https://github.com/mesa/mesa/pull/3323#discussion_r2822191001). I think using positions is more intuitive but current version is faster. What do you prefer?

quaquel: speed is important so fine with me

PR #3328, opened 2026-02-16
remove model.steps

quaquel:
This PR removes `model.steps` and updates all usages elsewhere to `model.time`, in line with the time-centric nature of Mesa 4. It also simplifies some of the step wrapping that was needed to advance `model.steps`.


quaquel:
The fact that mesa-examples just runs on this PR shows that we can indeed just remove `model.steps`.

Issue #3327, opened 2026-02-16
Bug in model.run_until


quaquel:
**Describe the bug**
There is a bug with model.run_until and model.run_for. Specifically, if you first run_for(10) and then do run_until(5.0), model.time after the second call should be 10, but is set to 5.0.

**Expected behavior**
Model.time should not go backward.

**To Reproduce**
Just add an extra check to `test_schedule_run.test_already_past`, asserting `model.time==10`

**fix**
The problem is that in `model._advance_time`, after the while loop, time is updated to `until`. But if until is smaller than the current value of time, we go backwards. The simplest solution is to check before the while loop if until is smaller than time and just return if true.

Krishsharma179:
Ok I will work on this

Krishsharma179:
Hey @quaquel, I've created a PR to fix this bug.

Added an early `return` in `_advance_time()` to prevent time rewind when `until <= self.time`.

✅ Verified:  
- `run_for(10)` → `time = 10`  
- `run_until(5.0)` → `time` remains **10** (no regression)  
- All CI checks pass

Would appreciate your review when you have a moment!

PR #3329, opened 2026-02-17
fix: prevent time rewind in _advance_time by adding early return

Krishsharma179:
### Summary
Fixes critical bug where model.run_until() incorrectly rewinds simulation time when target time is less than current time. Time must be monotonic in simulations (never decrease) to maintain causality and data integrity.

### Bug / Issue
Related Issue: #3327 
Context: When calling model.run_until(target) where target < model.time, the method overwrites model.time = target, causing time to jump backward.
Expected: Time should remain unchanged (no-op) when target is in the past.
Actual: Time rewinds (e.g., run_for(10) → time=10, then run_until(5.0) → time=5.0 (wrong)).

### Implementation
Modified mesa/model.py in _advance_time() method:
```
def _advance_time(self, until: float) -> None:
    """Advance time to the given point, processing events along the way."""
    # PREVENT TIME REWIND: exit immediately if target is in the past
    if until <= self.time:  # ← ADDED THIS GUARD CLAUSE
        return
    
    # ... (existing event processing logic) ...
    self.time = until
```

### Testing
```
m = Model()
m.time = 10.0
m._advance_time(5.0)  # Attempt to rewind
assert m.time == 10.0  # ✅ PASS: Time unchanged
m._advance_time(15.0)  # Forward advancement
assert m.time == 15.0  # ✅ PASS: Still works
````
output:
✅ Time remains 10.0 after run_until(5.0) (no rewind)
✅ Forward advancement to 15.0 works correctly
✅ All existing Mesa tests pass (verified locally)

### Additional Notes
Minimal change: Only 2 lines added, no logic modifications elsewhere
Critical for: Time-series data integrity, agent state consistency, reproducibility
No side effects: Guard clause exits early before any event processing
Standards compliance: Aligns with universal simulation principle: "Time never decreases"
Future-proof: Prevents subtle bugs in complex models using event scheduling

### DIFF
--- a/tests/test_schedule_run.py
+++ b/tests/test_schedule_run.py
@@ -75,10 +75,6 @@ class TestRunUntil:
         model = SimpleModel()
         model.run_for(10)
 
-
-with pytest.warns(RuntimeWarning):
-    model.run_until(5.0)  # already past t=5
-    assert model.steps == 10  # no additional steps
         with pytest.warns(RuntimeWarning):
             model.run_until(5.0)  # already past t=5
             assert model.steps == 10  # no additional steps

### CODE REVIEW

**Comment by quaquel** [comment]
1. I would move the check to `run_until`.
2. Adding a warning would be a good idea
3. Yes, we need an extra test. This is trivial, see my issue where I point to the test that needs to be updated. 

Note that I am working on a bigger PR with a larger set of bug fixes and other improvements for `model.time`. I am including this one in that as well. But I am happy to merge this one if it is in line with my suggestions.

**Comment by Krishsharma179** [comment]
Thanks for the review @EwoutH and @quaquel ! 

I'll implement both suggestions:
1.  Add a warning when time rewind is attempted
2.  Add a test case to prevent regression

I'll push the updates to this branch  today.

**Comment by Krishsharma179** [comment]
hey @EwoutH and @quaquel  please check the implementation I've made according to your suggestions

File: mesa/model.py [hunk 1]
```diff
@@ -181,6 +182,13 @@ def _advance_time(self, until: float) -> None:
             until: The time to advance to
 
         """
+        if until <= self.time:
+            warnings.warn(
+                f"Attempted to advance time to {until}, but current time is {self.time}. Time did not change.",
+                UserWarning,
+                stacklevel=2,
+            )
```
quaquel, line 190:
```suggestion
            warnings.warn(f"end_time {end_time} is larger than time {self.time}", RuntimeWarning, stacklevel=2)
            )
```

mesa/model.py
```diff
@@ -181,6 +182,10 @@ def _advance_time(self, until: float) -> None:
             until: The time to advance to
 
         """
+        if until <= self.time:
+            warnings.warn(f"end_time {end_time} is larger than time {self.time}", RuntimeWarning, stacklevel=2)
+            )
```
quaquel, line 190:
```suggestion
            
```

mesa/model.py
```diff
@@ -181,6 +182,14 @@ def _advance_time(self, until: float) -> None:
             until: The time to advance to
 
         """
+        if until <= self.time:
+            warnings.warn(
+                f"end_time {end_time} is larger than time {self.time}",
+                RuntimeWarning,
+                stacklevel=2,
+            )
```
quaquel, line 190:
```suggestion
            warnings.warn(
                f"end time {until} is larger than time {self.time}",
                RuntimeWarning,
                stacklevel=2,
            )
```

tests/test_schedule_run.py
```diff
@@ -76,6 +76,7 @@ def test_already_past(self):
         model.run_for(10)
         model.run_until(5.0)  # already past t=5
         assert model.steps == 10  # no additional steps
+        assert model.time == 10
```
quaquel:
```suggestion
with pytest.warns(RuntimeWarning):
    model.run_until(5.0)  # already past t=5
    assert model.steps == 10  # no additional steps
    assert model.time == 10
```

tests/test_schedule_run.py
```diff
@@ -74,8 +74,10 @@ def test_basic(self):
     def test_already_past(self):
         model = SimpleModel()
         model.run_for(10)
-        model.run_until(5.0)  # already past t=5
-        assert model.steps == 10  # no additional steps
+with pytest.warns(RuntimeWarning):
+    model.run_until(5.0)  # already past t=5
+    assert model.steps == 10  # no additional steps
+    assert model.time == 10
```
quaquel:
```suggestion
        with pytest.warns(RuntimeWarning):
            model.run_until(5.0)  # already past t=5
            assert model.steps == 10  # no additional steps
            assert model.time == 10
```

PR #3330, opened 2026-02-17
Add deprecation warnings to model.step() and run_model()

codebyNJ:
## Summary
- Add `FutureWarning` to `model.step()` and `model.run_model()`, guiding users toward `model.run_for()` / `model.run_until()`
- Update internal Mesa code (`batchrunner.py`) to use `run_for(1)` instead of `step()`
- Migrate tests from `model.step()` to `model.run_for(1)` and add deprecation warning tests

Addresses the tracking item in #3132:
> - [ ] Add deprecation warnings to `model.step()` and `run_model()`

## Changes
| File | Change |
|------|--------|
| `mesa/model.py` | `FutureWarning` in `_wrapped_step()` and `run_model()` |
| `mesa/batchrunner.py` | `model.step()` → `model.run_for(1)` internally |
| `tests/test_model.py` | Migrated to `run_for(1)`, added 2 deprecation warning tests |
| `tests/test_schedule_run.py` | `m1.step()` → `m1.run_for(1)` |
| `tests/test_lifespan.py` | `self.step()` → `self.run_for(1)` |
| `tests/experimental/test_meta_agents.py` | `model.step()` → `model.run_for(1)` |
| `tests/test_datacollector.py` | Added `pytestmark` to filter `FutureWarning` |
| `tests/experimental/test_datarecorder.py` | Added `pytestmark` to filter `FutureWarning` |

## Design decisions
- **`FutureWarning`** (not `DeprecationWarning`): Visible to end users by default, consistent with the `Simulator` deprecation pattern in Mesa
- **`run_model()` bypasses `self.step()`**: Calls `self._advance_time()` directly to avoid emitting a double warning (one from `run_model` + one from `step`)
- **Tests use `filterwarnings`** where `model.step()` is not the focus of the test (e.g., DataCollector tests), and **migrated to `run_for(1)`** where feasible

## Test plan
- [x] All 437 core tests pass with `-Werror` (matching CI configuration)
- [x] `test_step_deprecation_warning` verifies `model.step()` emits `FutureWarning`
- [x] `test_run_model_deprecation_warning` verifies `model.run_model()` emits `FutureWarning`

### DIFF
--- a/tests/experimental/test_datarecorder.py
+++ b/tests/experimental/test_datarecorder.py
@@ -6,10 +6,6 @@ import tempfile
 from pathlib import Path
 from unittest.mock import Mock
 
-import pytest
-
-pytestmark = pytest.mark.filterwarnings("ignore::FutureWarning")
-
 import numpy as np
 import pandas as pd
 import pytest
@@ -26,6 +22,8 @@ from mesa.experimental.data_collection import (
 )
 from mesa.experimental.data_collection.datarecorders import NumpyJSONEncoder
 from mesa.model import Model
+
+pytestmark = pytest.mark.filterwarnings("ignore::FutureWarning")
 
 
 class MockAgent(Agent):
--- a/tests/test_datacollector.py
+++ b/tests/test_datacollector.py
@@ -3,14 +3,13 @@
 import unittest
 from functools import partial
 
-import pytest
-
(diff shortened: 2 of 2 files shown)


quaquel:
main is already on 4.0, so these warnings are not useful here. Rather, we want to first develop 4.0, and then once we have a fair sense of everything that is changing, we can work on the 3.x branch and add in warnings there.

PR #3331, opened 2026-02-17
Event fixes

quaquel:
This PR fixes several minor issues and bugs in `mesa.time` and `mesa.model` related to event scheduling and run control.

1. It moves the relevant tests from `mesa.experimental.test_devs` into a new folder (time) and test_events. This ensures the test structure mimics the layout of the repo.
2. It improves code coverage for Schedule. Specifically, it adds additional tests for the post-init validation.
3. It adds a check that `interval` in the EventGenerator raises an exception if it is negative (in case of callable) and adds a test for this.
4. It removes the return from EventGenerator.stop. The idea of stop is that it stops and deposes the generator. There is no need for a return. 
5. The eventlist `__len__` and `__contains__` now ignore canceled events. This is also explicitly tested.

### DIFF
--- a/tests/time/test_events.py
+++ b/tests/time/test_events.py
@@ -442,7 +442,6 @@ class TestEventGenerator:
 
     def test_callable_interval_raises(self, setup):
         """Test if callable interval raises exception if return is negative."""
-
         with pytest.raises(ValueError):
             model, fn = setup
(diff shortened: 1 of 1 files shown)


EwoutH (approved):
Thanks. Indeed backport-worthy.

PR #3337, opened 2026-02-17
Remove mesa.space

quaquel:
This PR removes `mesa.space` and `agent.pos` and all uses of it throughout mesa. 

It interacts with #3323 on the visualization side.


quaquel:
I opened [#322](https://github.com/mesa/mesa-examples/issues/332) on mesa-examples listing the examples that must be updated in line with this PR.

EwoutH (commented):
Looks good. Thanks for the Mesa-examples issue.

What’s your plan for the core examples in this repo?

At some point we should also write migration guide, update tutorial, etc.

quaquel:
> What’s your plan for the core examples in this repo?

They already moved away from `mesa.space`when we added the deprecation warning. So nothing needs to be done. 

> At some point we should also write migration guide, update tutorial, etc.

I suggest we try to make simple changes where appropriate (see, for example, the removal of model steps). However, there will be an accumulation of changes coming, so it might be better to rewrite those completely once we are preparing the beta release of Mesa 4.

Issue #3332, opened 2026-02-17
EventGenerator.function is a hard ref


quaquel:
EventGenerator.function is a hard reference rather than a weak ref as used in Event. This can lead to memory leaks if a generator is used in combination with, e.g., an Agent. If the agent is removed from the model but remains part of an EventGenerator, it cannot be garbage-collected.

The best solution is to turn `function` into a weakref, identical to how this is done in `Event`. However, this leaves a `Model` instance in an unpickleable state because the step method is currently scheduled via an EventGenerator.

There is also a link to #3319

Jayantparashar10:
Hi @quaquel,

I'd like to work on this issue. I've confirmed the memory leak with a reproduction test:

**Test results**: Population of 500 agents -> 1,300 agent objects in memory after turnover cycles (800 leaked agents).

**Proposed solution**: Implement weak references with pickle support in `EventGenerator`, mirroring the existing `Event` class implementation:
- Use `WeakMethod`/`ref` for `function` attribute
- Add `__getstate__`/`__setstate__` for pickle compatibility
- Auto-stop generator when function is garbage collected

This fixes the leak while maintaining `batch_run()` compatibility.

Could you assign this to me? I'm ready to implement and create a PR. Thanks!

souro26:
Before directly mirroring `Event` and turning `EventGenerator.function` into a weakref, I think it might be worth reconsidering how the generator stores callable state more broadly. Unlike `Event`, an `EventGenerator` is persistent and self-rescheduling. Simply switching to a weak reference adds additional lifecycle complexity like handling disappearing callables mid-run, ensuring correct stop behavior, and maintaining pickle compatibility.

An alternative approach could be to avoid storing a direct callable reference entirely. For bound methods, the generator could store a weak reference to the owning object and the method name. The callable can then be reconstructed via `getattr` at execution time. If the object no longer exists, the generator can stop cleanly.

This avoids strong references, keeps pickling simpler, and makes lifecycle behavior explicit rather than relying purely on weakref mechanics.

Krishsharma179:
Hello @Jayantparashar10 !! are you actively working on this issue If so, could you please share an update on your progress?

Krishsharma179:
@Jayantparashar10, I haven't heard back from you regarding your progress on this memory leak fix. Since this is a critical issue affecting simulation performance, I've gone ahead and implemented the fix based on @quaquel's suggestion to use weak references (identical to the Event class implementation).

[...]

PR #3339, opened 2026-02-18
fix: use weakref in EventGenerator to prevent memory leak (#3332)

Krishsharma179:
### Summary
Fixes memory leak in EventGenerator by using weak references instead of hard references, identical to the Event class implementation. #3332 

### Bug / Issue
EventGenerator held hard references to functions, preventing garbage collection of agents even after they were removed from the model. This caused significant memory leaks in long-running simulations.

### Implementation
- Changed `EventGenerator.function` from hard reference to weak reference
- Added `@property` getter to dereference the weakref
- Added pickle support (`__getstate__` and `__setstate__`) for batch_run compatibility
- Pattern matches existing implementation in Event class

### Testing
Added tests to `tests/test_lifespan.py`:

```
class TestEventGeneratorMemoryLeak(unittest.TestCase):
    """Test that EventGenerator doesn't cause memory leaks."""

    def test_event_generator_weak_reference(self):
        """Test that EventGenerator uses weak references."""
        model = Model()
        agent = FiniteLifeAgent(10, model)
        
        schedule = Schedule(interval=1.0)
        generator = EventGenerator(model, agent.step, schedule)
        
        # CRITICAL: Remove agent from model
        agent.remove()
        
        # Delete the agent
        del agent
        gc.collect()
        
        # Accessing function should raise RuntimeError
        with self.assertRaises(RuntimeError):
            _ = generator.function

    def test_event_generator_memory_leak(self):
        """Test that agents can be garbage collected (reproduces issue scenario)."""
        model = Model()
        schedule = Schedule(interval=1.0)
        
        agents = []
        weak_refs = []
        
        # Create 50 agents
        for _ in range(50):
            agent = FiniteLifeAgent(10, model)
            agents.append(agent)
            weak_refs.append(weakref.ref(agent))
            
            EventGenerator(model, agent.step, schedule)
        
        # CRITICAL: Remove ALL agents from model
        for agent in agents:
            agent.remove()
        
        # Delete all agents
        del agents
        gc.collect()
        
        # Now they can be garbage collected
        collected_count = sum(1 for ref in weak_refs if ref() is None)
        self.assertGreaterEqual(
            collected_count, 45,
            f"Expected at least 45 agents collected, got {collected_count}"
        )
  ```
  output
  ```
  tests/test_lifespan.py::TestEventGeneratorMemoryLeak::test_event_generator_weak_reference PASSED
  tests/test_lifespan.py::TestEventGeneratorMemoryLeak::test_event_generator_memory_leak PASSED
```      

### Additional Notes
1.)Backward compatibility: The change is backward compatible existing code will continue to work, but now agents can be properly garbage collected
2.)This issue is also linked to #3319

### DIFF
--- a/mesa/experimental/mesa_signals/__init__.py
+++ b/mesa/experimental/mesa_signals/__init__.py
@@ -10,21 +10,26 @@ to attributes, computing derived values, and managing collections that emit sign
 when modified.
 """
 
-from .mesa_signal import (
-    All,
+from .core import (
     HasObservables,
     Observable,
-    SignalType,
-    computed,
+    computed_property,
+    emit,
 )
-from .observable_collections import ListSignalType, ObservableList
+from .observable_collections import ObservableList
+from .signal_types import ListSignals, ModelSignals, ObservableSignals
+from .signals_util import ALL, Message, SignalType
 
 __all__ = [
-    "All",
+    "ALL",
     "HasObservables",
-    "ListSignalType",
+    "ListSignals",
+    "Message",
+    "ModelSignals",
     "Observable",
     "ObservableList",
+    "ObservableSignals",
     "SignalType",
-    "computed",
+    "computed_property",
+    "emit",
 ]
--- a/mesa/experimental/mesa_signals/core.py
+++ b/mesa/experimental/mesa_signals/core.py
@@ -20,71 +20,54 @@ from __future__ import annotations
 
 import functools
 import weakref
-from abc import ABC, abstractmethod
 from collections import defaultdict, namedtuple
-from collections.abc import Callable
-from enum import Enum
-from typing import Any
+from collections.abc import Callable, Generator, Iterable
+from typing import Any, Literal
 
-from mesa.experimental.mesa_signals.signals_util import Message, create_weakref
+from mesa.experimental.mesa_signals.signals_util import (
+    ALL,
+    Message,
+    SignalType,
+    create_weakref,
+)
 
-__all__ = ["All", "HasObservables", "Observable", "SignalType", "computed"]
+from .signal_types import ObservableSignals
+
+__all__ = [
+    "BaseObservable",
+    "HasObservables",
+    "Observable",
+    "computed_property",
+    "emit",
+]
 
 
-class SignalType(str, Enum):
-    """Enumeration of signal types that observables can emit.
-
-    This enum provides type-safe signal type definitions with IDE autocomplete support.
-    Inherits from str for backward compatibility with existing string-based code.
-
-    Attributes:
-        CHANGE: Emitted when an observable's value changes.
-
-    Examples:
-        >>> from mesa.experimental.mesa_signals import Observable, HasObservables, SignalType
-        >>> class MyModel(HasObservables):
-        ...     value = Observable()
-        ...     def __init__(self):
-        ...         super().__init__()
-        ...         self._value = 0
-        >>> model = MyModel()
-        >>> model.observe("value", SignalType.CHANGE, lambda s: print(s.new))
-        >>> model.value = 10
-        10
-
-    Note:
-        String-based signal types are still supported for backward compatibility:
-        >>> model.observe("value", "change", handler)  # Still works
-    """
-
-    CHANGE = "change"
-
-    def __str__(self):
-        """Return the string value of the signal type."""
-        return self.value
+ObservableName = str | type(ALL) | Iterable[str]
+SignalSpec = str | SignalType | type(ALL) | Iterable[str | SignalType]
 
 
 _hashable_signal = namedtuple("_HashableSignal", "instance name")
 
 CURRENT_COMPUTED: ComputedState | None = None  # the current Computed that is evaluating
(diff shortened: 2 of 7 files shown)


Krishsharma179:
@quaquel , I've verified this fix locally with passing tests. The failing CI checks appear to be unrelated to this change (common Windows/doc issues in Mesa's CI). Would appreciate your review of the memory leak fix itself.

codebreaker32:
Hi @Krishsharma179, Thank you for this PR!
It seems @quaquel has already addressed this issue in #3331. See https://github.com/mesa/mesa/pull/3331/commits/253de9e39e181e7ae8fe658d995d3610c9aa72c3

quaquel:
> It seems @quaquel has already addressed this issue in https://github.com/mesa/mesa/pull/3331. See https://github.com/mesa/mesa/commit/253de9e39e181e7ae8fe658d995d3610c9aa72c3

I rolled back that change in a subsequent commit because of the pickling problem.

Krishsharma179:
> > It seems @quaquel has already addressed this issue in #3331. See [253de9e](https://github.com/mesa/mesa/commit/253de9e39e181e7ae8fe658d995d3610c9aa72c3)
> 
> I rolled back that change in a subsequent commit because of the pickling problem.
hey @quaquel ! I understand you rolled back because of pickling issue can you please let me know what should i do close this pr or wait for you to review it

Jayantparashar10:
Thanks again for the PR @Krishsharma179 — the weakref + pickle fix works great (leak gone in my 500-agent turnover test).

One small usability suggestion for the `function` property:

Instead of raising `RuntimeError` when the weakref is dead, we could auto-stop the generator and return a no-op:

```python
@property
def function(self) -> Callable:
    func = self._function_ref() 
    if func is None:
        self.stop()
        return lambda: None
    return func
```
This way:

- Orphaned generators become harmless (no more calls)
- No uncaught exceptions mid-step
- Still prevents the leak
- Aligns with "fail-safe" behavior in simulations

The current `RuntimeError` is clean and explicit, but might require extra try/except in the scheduler or user code. 
What do you think?

souro26:
I’d lean toward raising an error instead of returning a no-op. If the function reference is gone, that usually indicates a lifecycle issue. Silently returning a no-op could hide that and make debugging harder. Failing clearly seems safer to me for simulation behavior.

quaquel:
Please see the conversation in #3320. The behavior of both Event and EventGenerator should be the same. So fail fast on initialization, but if a weakref resolves to None later, just stop the generator.

Krishsharma179:
@quaquel ok got it! I'll update the PR to match Event's behavior:

Krishsharma179:
@quaquel i have implemented the suggestion please review it

falloficaruss:
I don't see how this PR follows our discussion in #3320. It introduces extra behavior changes (like stopping inside a property getter), which goes beyond the preferred direction. Also there are unresolved conflicts in the diff.

Krishsharma179:
> I don't see how this PR follows our discussion in #3320. It introduces extra behavior changes (like stopping inside a property getter), which goes beyond the preferred direction. Also there are unresolved conflicts in the diff.

Ok i will resolve the conflict and read the discussion again thanks for the feedback

quaquel:
I will merge #3320 first anyway so this PR can use the function added there.

Krishsharma179:
@quaquel I just read the conversation in the issue #3320 
 1.)If function is not a callable --> raise a TypeError
 2.)If function is not weakly referenceable --> raise a TypeError
 3.)If weakref to function resolves to None when Event is instantiated, raise a ValueError
 
Ig these were the behavior you had in mind so i implemented these
 1.)If function is not a callable --> raise a TypeError
```
#If the function is not callable raise Typeerror
        if not callable(function):
            raise TypeError(f"function must be callable, got {type(function).__name__}") 
```
2.)if the function is not weakly referencable
```
 # Here it will raise a Typeerror if the function is not weakly referncable
        
        try:
            if isinstance(function, MethodType):
                function = WeakMethod(function)
            else:
                function = ref(function)
        except TypeError:
            raise TypeError(
                f"function {function} is not weakly referenceable. "
                f"Use a user-defined function or method instead."
            )            
```
 3.)If weakref to function resolves to None when Event is instantiated, raise a ValueError
````
if function() is None:
            raise ValueError("Cannot create EventGenerator with a function that is already garbage collected")   
 ````
 Silent no-op if weakref dies (no RuntimeError)
Generator stops during _execute_and_reschedule() (not in property getter)           
If there is any changes or suggestion please let me know

Krishsharma179:
> I don't see how this PR follows our discussion in #3320. It introduces extra behavior changes (like stopping inside a property getter), which goes beyond the preferred direction. Also there are unresolved conflicts in the diff.

Hi @falloficarus22,
Thank you for taking the time to review my PR! 
I've implemented all the requested changes from issue #3320:
1.) Added TypeError check for non-callable functions
2.)Added TypeError check for non-weakly-referenceable functions
3.)Added ValueError check for already garbage-collected functions
4.)Added tests covering all three error cases
5.)Fixed the generator to stop when weakref dies during execution
Any suggestions on improving the implementation?

PR #3340, opened 2026-02-18
Remove PropertyLayer and HasPropertyLayers mixin 

codebreaker32:
**This PR removes `PropertyLayer` and `HasPropertyLayers` from discrete spaces**

Replaces `PropertyLayer` with raw NumPy arrays stored directly on the grid as `_property_layers: dict[str, np.ndarray]`. Property access on cells is handled via native `property` closures attached to the dynamic `GridCell` class.

**Changes:**
- `create_property_layer(name, ...)` - allocates and attaches a new array
- `add_property_layer(name, array)` - attaches an existing array
- `remove_property_layer(name)` - removes array and cell accessor
- `select_cells()` removed - use numpy directly on `property_layers`
- Visualization backends updated to match new APIs
- Sugarscape and Termites examples updated

**Example Usage**

```python
grid.create_property_layer("sugar", default_value=0.0)

# grid-level: full array
grid.sugar[:] = 0              
grid.sugar[grid.sugar > 4] = 4 # numpy ops directly

 # cell-level: scalar
cell.sugar = 2.0              

# replaces select_cells type ops
coords = list(zip(*np.where(grid.sugar > 2)))  
```

### DIFF
--- a/mesa/examples/advanced/sugarscape_g1mt/model.py
+++ b/mesa/examples/advanced/sugarscape_g1mt/model.py
@@ -112,8 +112,8 @@ class SugarscapeG1mt(mesa.Model):
         Unique step function that does staged activation of sugar and spice
         and then randomly activates traders
         """
-        np.minimum(self.grid.sugar + 1, self.sugar_distribution, out=self.grid.sugar)
-        np.minimum(self.grid.spice + 1, self.spice_distribution, out=self.grid.spice)
+        self.grid.sugar[:] = np.minimum(self.grid.sugar + 1, self.sugar_distribution)
+        self.grid.spice[:] = np.minimum(self.grid.spice + 1, self.spice_distribution)
 
         # step trader agents
         # to account for agent death and removal we need a separate data structure to


codebreaker32:
Hi @quaquel 
The `PropertyLayer` is deeply embedded in Visualisation as well, so I would love to get a heads up before working on it further if it aligns with your vision. I realize the diff is fairly large, but the core changes are concentrated in `grid.py` only

quaquel:
I'll try to take a look somewhere today.

mesa/discrete_space/grid.py
```diff
@@ -124,7 +121,129 @@ def __init__(
...
+        conditions: dict | None = None,
+        extreme_values: dict | None = None,
+        masks=None,
+        only_empty: bool = False,
+        return_list: bool = True,
+    ):
+        """Select cells using vectorised NumPy operations on property arrays.
+
+        Args:
+            conditions: ``{prop_name: callable}`` — callable receives the array, returns boolean mask.
```
quaquel: I am not sure about this signature. `conditions` can become quite big and difficult to read. It also seems redundant with masks.

I would leave out return_list. Just always return a mask. With np.where, this can easily be turned into coordinates. So just document this possibility.

mesa/examples/advanced/sugarscape_g1mt/model.py
```diff
@@ -84,12 +83,8 @@ def __init__(
         self.sugar_distribution = np.genfromtxt(Path(__file__).parent / "sugar-map.txt")
         self.spice_distribution = np.flip(self.sugar_distribution, 1)
 
-        self.grid.add_property_layer(
-            PropertyLayer.from_data("sugar", self.sugar_distribution)
-        )
-        self.grid.add_property_layer(
-            PropertyLayer.from_data("spice", self.spice_distribution)
-        )
+        self.grid.add_property("sugar", self.sugar_distribution.copy())
```
quaquel: it's important to document this on the add and create method on Grid.

mesa/discrete_space/grid.py
codebreaker32: I very well agree with you. This function was never my idea, also it is very much bloated in my opinion. I have only copy-pasted it from old implementation. I will look to further optimise it if needed and will remove if redundant

mesa/examples/advanced/sugarscape_g1mt/model.py
codebreaker32: Sure, I have not added any docstrings for now. I will add it once complete

mesa/discrete_space/grid.py
quaquel: yes feel free to cleanup and simplify. No need to worry about mesa 3 api.

codebreaker32:
This PR is ready for review.

During refactoring, I noticed that using `property` feels a bit awkward since it’s also a built-in Python decorator. Should we consider a clearer alternative name?

codebreaker32:
I profiled the Sugarscape model comparing main against this PR and results are:

| Run    | main | This PR | Change(in %) |
|--------|-------------------|-------------------|--------------------------|
| Run 1  | 3.81              | 3.69              | 96.90%                   |
| Run 2  | 3.97              | 3.56              | 89.89%                   |
| Run 3  | 3.25              | 3.08              | 94.50%                   |
| Run 4  | 3.39              | 3.14              | 92.82%                   |
| Run 5  | 3.44              | 3.16              | 91.67%                   |
| Run 6  | 3.43              | 3.49              | 101.64%                  |
| Run 7  | 3.23              | 3.15              | 97.58%                   |
| Run 8  | 3.05              | 3.02              | 98.93%                   |
| Run 9  | 3.50              | 3.62              | 103.60%                  |
| Run 10 | 3.91              | 3.52              | 90.03%                   |

1000 steps in each Run with different rng

codebreaker32:
On termites model from mesa-examples

| Run    | main | This PR | Change (%) |
|--------|----------|---------------|------------|
| Run 1  | 6.90     | 6.93          | 100.39%    |
| Run 2  | 6.72     | 7.36          | 109.63%    |
| Run 3  | 6.33     | 7.06          | 111.46%    |
| Run 4  | 7.01     | 6.49          | 92.64%     |
| Run 5  | 6.57     | 7.30          | 111.23%    |
| Run 6  | 6.49     | 6.21          | 95.63%     |
| Run 7  | 6.56     | 6.78          | 103.34%    |
| Run 8  | 6.25     | 6.59          | 105.54%    |
| Run 9  | 6.57     | 6.85          | 104.18%    |
| Run 10 | 6.37     | 6.37          | 100.06%    |

1000 steps in each run with different rng

quaquel:
Can you take a look at the pre-commit and fix the errors?

mesa/discrete_space/discrete_space.py
```diff
@@ -38,7 +38,7 @@ class DiscreteSpace[T: Cell]:
         random (Random): The random number generator
         cell_klass (Type) : the type of cell class
         empties (CellCollection) : collection of all cells that are empty
-        property_layers (dict[str, PropertyLayer]): the property layers of the discrete space
+        _properties (dict[str, np.ndarray]): property arrays of the discrete space
```
quaquel: also given your comment, I think we should keep calling it a property layer.

mesa/discrete_space/grid.py
```diff
@@ -124,7 +121,135 @@ def __init__(
         }
         self._celllist = list(self._cells.values())
         self._connect_cells()
-        self.create_property_layer("empty", default_value=True, dtype=bool)
+        self.create_property("empty", default_value=True, dtype=bool)
+
+    def create_property(
```
quaquel: Here to, I would suggest keeping the property_layer naming.

mesa/discrete_space/grid.py
```diff
@@ -124,7 +121,135 @@ def __init__(
...
+
+        accessor = (
+            property(getter, doc=f"Property '{name}'")
+            if read_only
+            else property(getter, setter, doc=f"Property '{name}'")
+        )
+        setattr(self.cell_klass, name, accessor)
+        self.cell_klass._properties.add(name)
+
+    def select_cells(
```
quaquel: I suggest just removing this entire method. I don't know what the use case for it is now that the numpy array is directly available.

codebreaker32:
In some methods/arguments naming, we were using `propertylayer*`( such as for `propertylayer_portrayal`) and somewhere else `property_layer*`. While renaming back, I was finding it difficult to sync where we have used what so I have replaced all `propertylayer*` namings with `property_layer*` to maintain uniformity

docs/tutorials/9_visualization_propertylayer_visualization.ipynb
```diff
@@ -15,17 +15,14 @@
...
     "\n",
     "Due to conflict with Colab and Solara there are no colab links for this tutorial\n",
     "\n",
     "*If you are running locally, please ensure you have the latest Mesa version installed.*\n",
     "\n",
     "## Tutorial Description\n",
-    "This tutorial builds upon the [Visualization Rendering with SpaceRenderer](8_visualization_rendering_with_space_renderer.ipynb) tutorial. We will explore more advanced features of the SpaceRenderer to create property layers and their visualization.\n",
-    "\n",
-    "> **NOTE:**\n",
-    "> This is not a tutorial on property layers themselves, but rather on their visualization. To explore the functionalities of property layers, please refer to this tutorial.\n",
```
codebreaker32, line 28: I removed this Note because I couldn't find any tuorial link here.

codebreaker32:
I think this is good to go from my end. I have also opened a PR in mesa-examples for failing example https://github.com/mesa/mesa-examples/pull/334

mesa/examples/advanced/sugarscape_g1mt/model.py
```diff
@@ -117,13 +112,11 @@ def step(self):
...
         and then randomly activates traders
         """
-        # step Resource agents
-        self.grid.sugar.data = np.minimum(
-            self.grid.sugar.data + 1, self.sugar_distribution
-        )
-        self.grid.spice.data = np.minimum(
-            self.grid.spice.data + 1, self.spice_distribution
-        )
+        sugar = self.grid._property_layers["sugar"]
```
quaquel: I am not sure I like this from an API point of view. To access a property layer at the grid level, we have to go through a private attribute. What would be needed to keep either the current attribute access to sugar, or offer an other clean way to offer grid level access to the entire layer?

codebreaker32: How is `grid.sugar`? similar to `cell.sugar` that already works. The only difference is `grid.sugar` will return `np.ndarray` and `cell.sugar` will return scalar

We'd be needing `__getattr__` on Grid to look up `_property_layers`

quaquel: Yes I would be fine with that. It will be important to document this and check if an other attribute with that name already exists.

mesa/discrete_space/grid.py
```diff
@@ -126,6 +123,119 @@ def __init__(
...
+
+        Raises:
+            AttributeError: If no property layer with that name exists.
+        """
+        if name == "_property_layers":  # guard against recursion during __init__
+            raise AttributeError(name)
+        try:
+            return self._property_layers[name]
+        except KeyError as e:
+            raise AttributeError(f"No property_layer named '{name}' exists") from e
```
quaquel: I realize that this is in line with how it was done in `HasPropertyLayers`, however, I am wondering whether we should not find a nicer solution for this. Messing with `__getattr__`  can be tricky. What about just adding dynamically a property via `property(getter)`?

codebreaker32: Wouldn't that mean maintaining two dynamically subclassed classes per grid? Or am I missing something?

codebreaker32: Some other options that I could think of are implementing `__getitem__` instead of `__getattr__` or a dedicated public property

```python
@property
def properties(self):
    return self._property_layers
```

quaquel: > Wouldn't that mean maintaining two dynamically subclassed classes per grid? Or am I missing something?

Not sure how you get to two but you make an important point that I had forgotten. We can only add properties to the class not the instance. And so, handling it via `__getattr__` is the better route. 

Separately, I do think that exposing self.property_layers directly is a good idea, but then I would suggest simply removing the underscore, rather than wrapping it in a `@property` accessor.

codebreaker32: > Not sure how you get to two

I may be overcomplicated it. To isolate per-instance, I thought a second dynamic subclass would be needed. 
```python
self.cell_klass = type(
        "GridCell",
        (cell_klass,),
        {"_property_layers": set(), "__slots__": ()},
    )

    self.__class__ = type(
        self.__class__.__name__,
        (self.__class__,),
        {},
    )
```
Could you clarify what approach you had in mind?

quaquel: I had not thought about it extensively yet. Exactly how to do it depends on what we are trying to achieve. I agree with this PR that it should be handled at the instance level rather than at the class level. We don't want to handle it at the class level because we might want to stack multiple grids.

So, there are 3 options left:
1. This PR, do it via `__getattr__`
2. Handle it via `__getattribute__`, basically the same therefor as this PR
3. Just directly assign the numpy array to the attribute via `setattr()`, potentially with a bit of protection to first check if no attribute with that name exists. So in, `_attach_property_layer` simply do:

```python

self.property_layers[name] = array 
setattr(self, name) = array

```

I don't see an immediate drawback to doing this. We have the check on the attribute name already in this method.

codebreaker32: (3) is a good solution and it looks intuitive as well and no messing with dunder methods. The only issue that I have is `grid.sugar` will now be a mutable reference so user can accidentally mutate it, And if we want to protect against it, we'll have to override `__setattr__` which is again what we want to avoid.

quaquel: As long as we carefully document the behavior in both add_property_layer and create_property_layer, I am inclined to prefer the simplicity of option 3 over the other two options.

mesa/examples/advanced/sugarscape_g1mt/model.py
```diff
@@ -117,13 +112,11 @@ def step(self):
...
-            self.grid.sugar.data + 1, self.sugar_distribution
-        )
-        self.grid.spice.data = np.minimum(
-            self.grid.spice.data + 1, self.spice_distribution
-        )
+        sugar = self.grid.property_layers["sugar"]
+        spice = self.grid.property_layers["spice"]
+
+        np.minimum(sugar + 1, self.sugar_distribution, out=sugar)
+        np.minimum(spice + 1, self.spice_distribution, out=spice)
```
quaquel:
```suggestion
        np.minimum(self.sugar + 1, self.sugar_distribution, out=self.sugar)
        np.minimum(self.spice + 1, self.spice_distribution, out=self.spice)
```

or 

```suggestion
        self.sugar  = np.minimum(self.sugar+1, self.sugar_distribution)
        self.spice = np.minimum(self.spice+1, self.spice_distribution)
```

quaquel: sorry, one last realy minor suggestion. i promise this is the last one.

codebreaker32: Not an issue at all. I really appreciate the suggestions. Please feel free to share any other thoughts if something comes to mind.

codebreaker32: I have applied your 1st suggestion with a minor tweak and 2nd suggestion is dangerous as we have warned about the same in docstrings of `create_property_layer` and `add_property_layer`

quaquel: You are right about the second one. The correct way to do it would be `self.spice[:] = np.minimum(self.spice+1, self.spice_distribution)`

codebreaker32: `self.grid` rather than `self`, as `self` refers to the model object here

PR #3345, opened 2026-02-18
Update `Network` to use `Cell.position` and `layout` for Visualisation

codebreaker32:
This PR updates `space_renderer.py` to treat Network spaces as direct position-based systems, aligning their rendering logic with OrthogonalGrid, VoronoiGrid, and ContinuousSpace.

### Changes:-

1. It decouples the visualization backends from NetworkX layout calculations. Because `cell.position` is now guaranteed, all dictionary lookups, `np.unique` ID mapping, and fallback layout algorithms were removed from the visualization layer. `Network` spaces now use the same highly optimized, direct coordinate pass-through as `OrthogonalGrid` and `ContinuousSpace`.

2. Deleted obsolete tests that were checking the legacy dictionary lookup and missing-node warning logic, and updated coordinate mapping assertions.

### DIFF
--- a/mesa/experimental/data_collection/dataset.py
+++ b/mesa/experimental/data_collection/dataset.py
@@ -112,13 +112,16 @@ class AgentDataSet[A: Agent](BaseDataSet):
         name: the name of the data set
         agents: the agents to collect data from
         fields: fields to collect
-
+        use_dirty_flag: if True, enables manual dirty-flag caching optimization.
+                        Default is False (always recompute snapshot).
     """
 
     def __init__(
         self,
         name: str,
         agents: AbstractAgentSet[A],
+        *,
+        use_dirty_flag: bool = False,
         fields: str | list[str] | None = None,
     ):
         """Init. of AgentDataSet."""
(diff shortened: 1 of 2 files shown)


mesa/visualization/space_renderer.py
```diff
@@ -128,7 +127,9 @@ def _map_coordinates(self, arguments):
         """
         mapped_arguments = arguments.copy()
 
-        if isinstance(self.space, OrthogonalGrid | VoronoiGrid | ContinuousSpace):
+        if isinstance(
+            self.space, OrthogonalGrid | VoronoiGrid | ContinuousSpace | Network
```
quaquel, line 131: how does this work if we don't assign positions to network inside the model? Is it still possible to handle this in the visualization or not?

codebreaker32: >how does this work if we don't assign positions to network inside the model?

It won't

>Is it still possible to handle this in the visualization or not?

Honestly I need to explore the visualization side a bit more, once #3323 is merged, I will explore it bit further if that's okay.

quaquel: I am also fine with requiring this at the model level across the board. In my view, this would be in line with the idea of a Locatable protocol.

quaquel:
The tests are breaking on the network change. Can you take a look?

codebreaker32:
Something is definitely going wrong with Networks

<br>

https://github.com/user-attachments/assets/2f6a2c30-950d-450e-aedf-537c52526d94


I am working on it but any inputs will be highly appreciated

codebreaker32:
Hi @quaquel,

After digging into these edge cases, they all trace back to a single root cause: handling topological networks where `Cell._position` is empty.

Currently, the visualization layer is forced to do the heavy lifting to translate `node_ids` into `(x, y)` coordinates and manage fallback layouts. If the long-term architectural vision is for the visualization to natively consume `cell._position`, I propose we enforce that `Cell._position` is never empty at the model level for `Network` spaces.

I see two options for when a user initializes a Network with `layout=None`:

1. Automatically generate a default layout (e.g., `nx.spring_layout`) under the hood.
2. Raise an exception

What are your thoughts?

quaquel:
>  I propose we enforce that Cell._position is never empty at the model level for Network spaces.

I agree and I am inclined to use a meaningful default if no position is provided. Not sure what that default should be. In my view it should be very cheap to calculate for large networks. So something like a circle layout might be sufficient. As long as this behavior is well documented, I think this is the best way to go.

codebreaker32:
> In my view it should be very cheap to calculate for large networks. So something like a circle layout might be sufficient

Agree. circular_layout is the perfect choice

codebreaker32:
This PR is ready for review and can be merged after #3355 as discussed

Issue #3341, opened 2026-02-18
Improvements to DataSet and DataRecorder


quaquel:
In #3156, we added a new `DataSet` for tracking the state of part of the model at any given instance. This `DataSet` forms the basis of the new approach to data collection. In #3145, we added `DataRecorder` for takings snapshots of DataSets. However there is ample room for improvements

- [x] add DataSet.set_dirty_flag(); A major potential benefit of datasets is that we can reuse them elsewhere in the model. However, to fully benefit from this, we need a way of indicating whether the data is clean or dirty. For example, in Boltzmann, we collect wealth data from agents and use this in gini. Ideally, we gather the agent wealth data only once per step. If we have a way of setting, manually, a flag indicating whether the data is clean or needs to be collected again, that would save runtime
- [ ] Handle run_ended; At the moment, DataRecorder subscribes to `model.time` `ObservableSignals.CHANGED`. So whenever time is changed, we collect data. This effectively means that data is collected before the event of the new time is executed. In turn this means that when a run is completed, we do not collect the data after that last event. 
- [ ] Add an `ObservableAgentDataSet`; 
- [ ] Align `AgentDataSet` and `NumpyDataSet` to both use an Agent class instead of the agents for the first and the class for the second.
- [ ] Agree on and implement an API along the lines of the [conversation here](add DataSet.set_dirty_flag()
handle run_ended
add a dataset for observables). `DataSet.record` has already been added. We also agree on the idea of a `Snapshot` data class. The other ideas are still under consideration.

som-or:
> * Handle run_ended; At the moment, DataRecorder subscribes to `model.time` `ObservableSignals.CHANGED`. So whenever time is changed, we collect data. This effectively means that data is collected before the event of the new time is executed. In turn this means that when a run is completed, we do not collect the data after that last event.

Hi @quaquel ,

I'd like to take this on.

I see that there are two ways to solve this.

1. Add a `STEP_COMPLETED` signal
Have `DataRecorder` subscribe to a lifecycle event emitted after `model.step()` finishes. This would align snapshots with completed steps and naturally capture the final state. Architecturally this feels better.
2. Add a `RUN_ENDED` signal
Keeping the current behavior but emit a `RUN_ENDED` event and collect one final snapshot when the simulation finishes.

let me know which direction you’d prefer and I can start working on it.

souro26:
[...]

PR #3346, opened 2026-02-19
Add manual dirty flag optimization to AgentDataSet

souro26:
### Summary
Adds a manual dirty-flag optimization to `AgentDataSet` to avoid repeated data extraction within the same simulation instant. This is implemented as a pull-based caching mechanism that recomputes only when explicitly invalidated.

Adds the dirty flag function mentioned in #3341 

### Motive
`AgentDataSet.data` currently recomputes the snapshot on every access. When datasets are reused within the same simulation instant, this results in redundant extraction work. This change introduces a manual `set_dirty_flag()` method so model code can explicitly mark the dataset as stale when needed. 

### Implementation
Added two internal attributes to `AgentDataSet`, `_is_dirty` and `_cache` and modified `.data` to recompute only if `_is_dirty` is True or `_cache` is None, otherwise return cached snapshot

Added a function,
````python
def set_dirty_flag(self) -> None
````
which marks the dataset as dirty for the next access.

### Usage Examples
````python
dataset = AgentDataSet("wealth", model.agents, fields="wealth")

# First access computes snapshot
snapshot = dataset.data

# Reuse within same instant (cached)
snapshot_again = dataset.data

# After model mutation
dataset.set_dirty_flag()
updated_snapshot = dataset.data
````

### Additional Notes
- This change alters identity behavior so that repeated `.data` access now returns the same list object until invalidated.
- Snapshot values and timing semantics remain unchanged and returned snapshot remains mutable.
- Added tests to cover the new changes

### DIFF
--- a/mesa/discrete_space/network.py
+++ b/mesa/discrete_space/network.py
@@ -33,7 +33,7 @@ class Network(DiscreteSpace[Cell]):
         capacity: int | None = None,
         random: Random | None = None,
         cell_klass: type[Cell] = Cell,
-        layout: Mapping | Callable | None = nx.spring_layout,
+        layout: Mapping | Callable | None = nx.circular_layout,
     ) -> None:
         """A Networked grid.
 
@@ -44,7 +44,9 @@ class Network(DiscreteSpace[Cell]):
             cell_klass (type[Cell]): The base Cell class to use in the Network
             layout: A dictionary mapping node IDs to physical positions (x, y),
                 or a callable that generates them (e.g. nx.spring_layout).
-                Set to None to force a purely topological network (no physical positions).
+                It defaults to nx.circular_layout
+                This ensures all nodes possess physical (x, y) positions for visualization and
+                spatial queries without introducing performance bottlenecks on large graphs
         """
         super().__init__(capacity=capacity, random=random, cell_klass=cell_klass)
         self.G = G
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by quaquel** [comment]
The current implementation in this PR forces the user to always handle the dirty flag. This is, in my view, less than ideal. Rather it should be an opt in behavior. So, probably in the `__init__` we need an extra keyword argument that specifies whether the user wants to opt in or not.

**Comment by souro26** [comment]
I’ve updated the implementation so the dirty-flag caching is now opt-in via a keyword argument in `AgentDataSet`. By default, behavior remains unchanged and .data recomputes on every access.

File: mesa/experimental/data_collection/dataset.py [hunk 1]
```diff
@@ -112,14 +112,17 @@ class AgentDataSet[A: Agent](BaseDataSet):
...
+        use_dirty_flag: if True, enables manual dirty-flag caching optimization.
+                        Default is False (always recompute snapshot).
     """
 
     def __init__(
         self,
         name: str,
         agents: AbstractAgentSet[A],
         fields: str | list[str] | None = None,
+        *,
```
quaquel, line 123: this is in the wrong location, should be before fields

mesa/experimental/data_collection/dataset.py
```diff
@@ -130,18 +133,49 @@ def __init__(
...
+                for agent in self.agents
+            ]
+
+        # Dirty-flag caching behavior (opt-in)
+        if self._is_dirty or self._cache is None:
+            self._cache = [
+                dict(zip(self._attributes, self._collector(agent)))
+                for agent in self.agents
+            ]
+            self._is_dirty = False
```
quaquel, line 156: this can be simplified to avoid duplicate code.

souro26:
I’ve made the suggested changes, thanks for the review. Let me know if anything else should be adjusted.

mesa/experimental/data_collection/dataset.py
quaquel, line 552: please update this as well to match the new signature of AgentDataSet

souro26: I've updated the track_agents signature to match AgentDataSet

PR #3348, opened 2026-02-19
Add sugarscape for benchmarking

quaquel:
This PR adds Sugarscape to the set of benchmark models., specifically for property layer performance. To do this, this PR updates the model to take advantage of various developments over the course of mesa 3.

### DIFF
--- a/mesa/examples/advanced/sugarscape_g1mt/model.py
+++ b/mesa/examples/advanced/sugarscape_g1mt/model.py
@@ -88,7 +88,7 @@ class SugarscapeG1mt(mesa.Model):
             self.random.choices(self.grid.all_cells.cells, k=n),
             sugar=self.rng.integers(
                 self.scenario.endowment_min,
-                self.scenario.max,
+                self.scenario.endowment_max,
                 (n,),
                 endpoint=True,
             ),


mesa/examples/advanced/sugarscape_g1mt/model.py
```diff
@@ -91,24 +83,41 @@ def __init__(
...
+
         Trader.create_agents(
             self,
-            initial_population,
-            self.random.choices(self.grid.all_cells.cells, k=initial_population),
+            SugarScapeScenario.initial_population,
+            self.random.choices(self.grid.all_cells.cells, k=n),
             sugar=self.rng.integers(
-                endowment_min, endowment_max, (initial_population,), endpoint=True
+                SugarScapeScenario.endowment_min,
```
EwoutH: Is it best practice we use the class here, and not an instance of the class?

I would expect this being `self.scenario.endowment_min` or something.

quaquel: you are right, I should use self.scenario.

EwoutH: Just checked, in the `__init__` you can simply use `scenario`, and `self.scenario` can be used everywhere in the Model and Agent (sub)classes.

quaquel: Thanks. I don't know what I was thinking when I used the class instead of the instance. Regardless, it's fixed now.

EwoutH (approved):

mesa/examples/advanced/sugarscape_g1mt/model.py
```diff
@@ -20,47 +21,41 @@ def geometric_mean(list_of_prices):
...
-    ):
-        super().__init__(rng=rng)
+    def __init__(self, scenario=None):
+        if scenario is None:
+            scenario = SugarScapeScenario()
+
+        super().__init__(scenario=scenario)
         # Initiate width and height of sugarscape
-        self.width = width
-        self.height = height
```
tpike3, line 60: I think you need to remove the height and width parameter in app.py, the solara viz is erroring out with this change.

tpike3 (commented):
Besides the one error for the solara aspect everything looks good.

mesa/examples/advanced/sugarscape_g1mt/model.py
```diff
@@ -86,24 +81,40 @@ def __init__(
...
         Trader.create_agents(
             self,
-            initial_population,
-            self.random.choices(self.grid.all_cells.cells, k=initial_population),
+            self.scenario.initial_population,
+            self.random.choices(self.grid.all_cells.cells, k=n),
             sugar=self.rng.integers(
-                endowment_min, endowment_max, (initial_population,), endpoint=True
+                self.scenario.endowment_min,
+                self.scenario.endowment_min,
```
jackiekazil: Should this be -- `self.scenario.endowment_max`?

mesa/examples/advanced/sugarscape_g1mt/model.py
```diff
@@ -86,24 +81,40 @@ def __init__(
...
+                (n,),
+                endpoint=True,
             ),
             spice=self.rng.integers(
-                endowment_min, endowment_max, (initial_population,), endpoint=True
+                self.scenario.endowment_min,
+                self.scenario.endowment_min,
```
jackiekazil: Should this be -- `self.scenario.endowment_max`?

mesa/examples/advanced/sugarscape_g1mt/model.py
```diff
@@ -20,47 +21,41 @@ def geometric_mean(list_of_prices):
     """
     find the geometric mean of a list of prices
     """
+    # protects against an invalid value if no prices
+    if len(list_of_prices) == 0:
+        return -1
```
jackiekazil, line 26: Why -1?  Wouldn't you use `np.nan`? I wonder if -1 may impact the analysis/output of where it returned in an unintended way.

mesa/examples/advanced/sugarscape_g1mt/agents.py
```diff
@@ -274,6 +278,14 @@ def maybe_die(self):
         if self.is_starved():
             self.remove()
 
+    def step(self):
+        """Agent step method."""
+        self.prices = []
+        self.trade_partners = []
+        self.move()
+        self.eat()
+        self.maybe_die()
```
jackiekazil, line 287: `maybe_die` used to before iteration -- but now it is mid-iteration.... this is handled by AgentSet's concurrent removal, yes? Is this tested in someway?

jackiekazil (changes requested):
See comments.

mesa/examples/advanced/sugarscape_g1mt/agents.py
quaquel: yes and yes.

mesa/examples/advanced/sugarscape_g1mt/model.py
quaquel: We store prices in lists and we validate that they are normal python numbers. So I cannot use None or numpy dtypes here. Also, a fix of this kind was needed anyway because the model won't run if `Scenario.enable_trade = False`.

quaquel: > I  wonder if -1 may impact the analysis/output of where it returned in an unintended way.

Also, there is no need to worry about this. There is no analysis with this example. This function is used for a model reporter. So at worst, you will get -1 for certain points in time, which are clearly wrong numbers for an average price. If I return `None` or `np.nan`, I create downstream problems in, for example, the line plot used in app.py.  Evidently, a cleaner solution would be to revamp the model to not track the average price, if `scenario.enable_trade = False`, but then you still have a problem in the UI.

PR #3350, opened 2026-02-19
Use model-specific scenario classes in benchmarks

EwoutH:
Follow-up to #3314. Now that the example models use `Scenario` subclasses, the benchmark configurations can be simplified:

- **Model-specific scenario classes**: Each configuration now references the appropriate scenario class (`BoltzmannScenario`, `SchellingScenario`, etc.) instead of the generic base `Scenario`. This means `global_benchmark.py` no longer needs to import `Scenario` directly.
- **Removed redundant default parameters**: Parameters that match the scenario class defaults are omitted from the configurations, reducing duplication and making it clear which values are actually being customized for the benchmark.
- **Fresh scenario per run**: A new scenario instance is created for each `run_model` call, avoiding the `ValueError` from the scenario's `__setattr__` guard when reassigning `model` on an already-bound scenario.

### DIFF
--- a/benchmarks/configurations.py
+++ b/benchmarks/configurations.py
@@ -1,11 +1,17 @@
 """configurations for benchmarks."""
 
-from mesa.examples import BoidFlockers, BoltzmannWealth, Schelling, WolfSheep, SugarscapeG1mt
+from mesa.examples import (
+    BoidFlockers,
+    BoltzmannWealth,
+    Schelling,
+    SugarscapeG1mt,
+    WolfSheep,
+)
+from mesa.examples.advanced.sugarscape_g1mt.model import SugarScapeScenario
 from mesa.examples.advanced.wolf_sheep.model import WolfSheepScenario
 from mesa.examples.basic.boid_flockers.model import BoidsScenario
 from mesa.examples.basic.boltzmann_wealth_model.model import BoltzmannScenario
 from mesa.examples.basic.schelling.model import SchellingScenario
-from mesa.examples.advanced.sugarscape_g1mt.model import SugarScapeScenario
 
 configurations = {
     # BoltzmannWealth Model Configurations


quaquel:
I was working on exactly the same thing last night :). There was one difference: I instantiated the scenario with the params in configuration. But then I ran into some issues with not being able to change scenarios anymore. I'll open up a pr with my code for comparison. It reveals some issues regarding Scenario that we have to think about with eye towards model experimentation.

EwoutH:
Great minds think alike!

If I understand it correctly, you pre-instantiate the Scenario in the config dict and then `copy.copy()` it for each run, while this PR stores the scenario class + parameter dict separately and creates a fresh instance per call. Both work, but I think the fresh-instantiation approach is a bit cleaner since it sidesteps the copy gymnastics entirely and makes it obvious that each run gets its own independent scenario. It also let me drop parameters that match the scenario class defaults, which reduces the noise in the config.

That said, your PR nicely surfaces the design tension around Scenario mutability that we need to resolve. I think we should tackle that separately (maybe as a focused discussion on whether `scenario.model` is even needed), and then whichever benchmark approach lands can benefit from the fix.

For now, this PR avoids the problem by just never reusing instances, but that's a workaround, not a solution. So I would propose landing this first and then improving on it.

quaquel:
> If I understand it correctly, you pre-instantiate the Scenario in the config dict and then copy.copy() it for each run, while this PR stores the scenario class + parameter dict separately and creates a fresh instance per call. Both work, but I think the fresh-instantiation approach is a bit cleaner since it sidesteps the copy gymnastics entirely and makes it obvious that each run gets its own independent scenario. 

I partially agree. The `copy.copy` is indeed a workaround and we need a clean solution. I want scenario to be effectively frozen after instantiation. I want this to ensure experimental conditions are not changed mid-run. However, I am unsure how to handle replications in this. Mesa is designed around a workflow where for each run, a fresh model is instantiated. This implies that a Scenario instance needs to have its rng. But, for replications, all we chance is the rng. One option might be to mimic numpy and create the possiblity to spawn scenario instances from a "parent" Scenario. The nice thing is that we can then also rely on the spawn functionality inside numpy and we only need a single seed for the initial generator. Moreover, we can move the entire rng logic into the scenario class and away from model. For details on numpy generator spawn, see #3195. 

```python

scenario = MyCustomScenarios(a=1, b=2, rng=42)
replications = scenario.spawn(replications=10)

for replication in replications:
    model = MyModel(scenario=replication)

```

> It also let me drop parameters that match the scenario class defaults, which reduces the noise in the config.

This can be done in both cases and is a benefit of using custom scenario subclasses.

EwoutH:
Thanks. Anything needed to change for this PR?

quaquel:
no this seems fine for now. Once we address the scenario rng stuff, we can simplify this further.

PR #3352, opened 2026-02-20
Simplify benchmarks

quaquel:
This PR is my take, developed in parallel to #3350, to using custom Scenario classes in the benchmark. 

The main difference between #3350 and this PR is that I instantiate the Scenario already in configurations. But, since scenarios cannot be changed while a model is running, I have to use a copy for every run. 

This does reveal at least one design flaw in Scenario: it holds a hard ref to the model and so might cause GC problems. Conceptually, I still think that treating Scenario instances as effectively frozen once instantiated is a good idea. But, at the same time, it would be rather convenient if we could easily update just the rng field when running a set of experiments:

```python
# code below won't work because scenario cannot be modified if
# it holds a reference to a model and model.running is true
for scenario in scenarios:
    for seed in seeds:
        scenario.rng = seed
        model = MyModel(scenario=scenario)
        model.run_until(500)

# current work around option 1
for s in scenarios:
    for seed in seeds:
        scenario = copy.copy(s)        
        scenario.rng = seed
        model = MyModel(scenario=scenario)
        model.run_until(500)

# current work around option 2
for param_set in param_sets:
    for seed in seeds:
        scenario = MyScenario(**param_set, rng=seed)
        model = MyModel(scenario=scenario)
        model.run_until(500)
```

### DIFF
--- a/benchmarks/global_benchmark.py
+++ b/benchmarks/global_benchmark.py
@@ -13,8 +13,6 @@ sys.path.insert(0, os.path.abspath(".."))
 
 from configurations import configurations
 
-from mesa.experimental.scenarios import Scenario
-
 
 # Generic function to initialize and run a model
 def run_model(model_class, steps, scenario):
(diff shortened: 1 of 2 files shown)


quaquel:
closing in favor of #3350

PR #3353, opened 2026-02-20
Add tests for execution ordering and cancellation invariants for EventList

souro26:
This PR adds explicit tests covering `EventList` execution semantics:

- FIFO ordering for events with identical timestamp and priority
- Execution of events scheduled at the same timestamp during execution
- Skipping of canceled events during popping/execution

These tests document and protect behavior already relied upon by the event engine

### DIFF
--- a/mesa/discrete_space/network.py
+++ b/mesa/discrete_space/network.py
@@ -33,7 +33,7 @@ class Network(DiscreteSpace[Cell]):
         capacity: int | None = None,
         random: Random | None = None,
         cell_klass: type[Cell] = Cell,
-        layout: Mapping | Callable | None = nx.spring_layout,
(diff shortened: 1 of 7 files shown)


tests/time/test_events.py
```diff
@@ -315,6 +315,76 @@ def test_eventlist():
     assert len(event_list) == 0
 
 
+def test_eventlist_fifo_same_time_same_priority_execution():
+    """Events with identical time and priority execute in insertion order."""
```
quaquel: that's not true, they execute on event id order

souro26: I'll change the function name and docstring to reflect that, thanks.

quaquel (approved):

PR #3359, opened 2026-02-21
Add compaction method to Eventlist to reduce performance degradation under heavy cancellation

### LINKED ISSUE #3358: EventList performance degradation under heavy cancellation due to lazy deletion
**Summary**
Currently EventList uses lazy deletion, so canceled events remain in the heap and are skipped during `pop_event()`. Current implementation:
````python
def pop_event(self) -> Event:
    while self._events:
        event = heappop(self._events)
        if not event.CANCELED:
            return event
    raise IndexError("Event list is empty")
````
When a large fraction of events are canceled, `pop_event()` may repeatedly discard canceled entries before reaching live ones. This makes pop cost proportional to total heap size rather than active size.

**Reproducible Benchmark**
````python
import time
from heapq import heapify
from mesa.time import Event, EventList, Priority

def dummy():
    pass

N = 200_000
CANCEL_RATIO = 0.9

event_list = EventList()
events = []

# Insert events
for i in range(N):
    e = Event(i, dummy, priority=Priority.DEFAULT)
    events.append(e)
    event_list.add_event(e)

# Cancel 90%
for e in events[:int(N * CANCEL_RATIO)]:
    e.cancel()

print("Heap size:", len(event_list._events))
print("Active events:", len(event_list))

# Pop without compaction
start = time.perf_counter()
try:
    while True:
        event_list.pop_event()
except IndexError:
    pass
print("Pop time (no compaction):", time.perf_counter() - start)
````
The observed result is,
````
Heap size: 200000
Active events: 20000
Pop time (no compaction): ~0.70s
````
If we manually compact the heap before popping,
````python
event_list._events = [
    e for e in event_list._events if not e.CANCELED
]
heapify(event_list._events)
````
Pop time becomes,
````
Heap size: 20000
Pop time (after compaction): ~0.05s
````
The slowdown is caused by repeatedly popping and discarding canceled entries.

**Proposed Direction**

Introducing adaptive heap compaction when canceled entries are large in number. For example,

Track `_canceled_count` . If `_canceled_count` > `len(self._events)` // 2, rebuild the heap,
````python
self._events = [
    e for e in self._events if not e.CANCELED
]
heapify(self._events)
self._canceled_count = 0
````
This preserves current semantics and ordering while preventing pathological heap bloat under heavy cancellation.

### PR DESCRIPTION
**Pull request by souro26**
### Summary
This PR improves `EventList` performance under heavy event cancellation by introducing adaptive heap compaction. When a large proportion of scheduled events are canceled, the internal heap can become dominated by canceled entries, significantly degrading `pop_event()` performance. The adaptive compaction mechanism rebuilds the heap when canceled events exceed a threshold.

Fixes #3358 

### Motive
`EventList` currently uses lazy deletion for cancellations. Under heavy cancellation like 90%, the heap can grow large while most entries are canceled. This leads to repeated skipping during `pop_event()`, causing significant performance degradation.

Benchmark (N=200,000, 90% canceled):
````
Without compaction: ~0.50–0.98s pop time
With compaction:    ~0.047–0.07s pop time
````
### Implementation

- Added an internal `_canceled_count` counter to track canceled events.
- Incremented the counter in `remove()` when an event is canceled.
- Decremented it during `pop_event()` when canceled events are skipped.
- Introduced a private _compact() method that filters out canceled events, puts the remaining events in a heap and then resets `_canceled_count`

Compaction is triggered adaptively inside `pop_event()` when
````
_canceled_count > len(_events) // 2
````
This keeps the normal case of low cancellation unchanged while preventing severe degradation when cancellations dominate. I considered manual compaction , but chose adaptive compaction to preserve lazy deletion semantics and avoid requiring explicit API calls from users.

### Additional Notes
- Existing lazy deletion semantics are kept the same.
- Added a behavioral test covering heavy cancellation to ensure correctness under stress.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature-request.md
+++ b/.github/ISSUE_TEMPLATE/feature-request.md
@@ -10,5 +10,8 @@ about: Suggest a new feature for Mesa
 **Describe the solution you'd like**
 <!-- A clear and concise description of what you want to happen. -->
 
+**Are you planning to open a PR for this?**
+<!-- For enhancements/new features, please wait for maintainer approval on this issue/discussion before opening a PR. -->
+
 **Additional context**
 <!-- Add any other context, links, etc. about the feature here. -->
--- a/.github/PULL_REQUEST_TEMPLATE/bug.md
+++ b/.github/PULL_REQUEST_TEMPLATE/bug.md
@@ -1,3 +1,9 @@
+> [!NOTE]
+> Use this template for bug fixes only. For enhancements/new features, use the feature template and get maintainer approval in an issue/discussion before opening a PR.
+
+### Pre-PR Checklist
+- [ ] This PR is a bug fix, not a new feature or enhancement.
+
 ### Summary
 <!-- Provide a brief summary of the bug and its impact. -->
 
@@ -14,3 +20,4 @@ If you're fixing the visualisation, add before/after screenshots. -->
 
 ### Additional Notes
 <!-- Add any additional information that may be relevant for the reviewers, such as potential side effects, dependencies, or related work.
+-->
--- a/.github/PULL_REQUEST_TEMPLATE/feature.md
+++ b/.github/PULL_REQUEST_TEMPLATE/feature.md
@@ -1,3 +1,12 @@
+> [!IMPORTANT]
+> For enhancements and new features, include the issue/discussion where a maintainer approved this work before opening a PR. Unapproved feature/enhancement PRs may be closed.
+
+### Pre-PR Checklist
+- [ ] I linked an issue/discussion where a maintainer approved this enhancement/feature for implementation.
+
+### Approval Link
+<!-- Link to the issue/discussion comment where a maintainer confirmed this is ready for a PR. -->
+
 ### Summary
 <!-- Provide a concise summary of the feature and its purpose. -->
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,3 +2,5 @@ Thanks for opening a PR! Please click the `Preview` tab and select a PR template
 
 - [🐛 Bug fix](?expand=1&template=bug.md)
 - [🛠 Feature/enhancement](?expand=1&template=feature.md)
+
+Feature/enhancement PRs require prior maintainer approval in an issue or discussion before they are accepted.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,8 @@ discuss via [Matrix] OR via [an issue].
 
 **To submit a contribution**
 
-- Create a ticket for the item that you are working on.
+- For enhancements or new features, open an [issue](https://github.com/mesa/mesa/issues) or [discussion](https://github.com/mesa/mesa/discussions) first and wait for maintainer approval before opening a PR.
+- For clear bug fixes, a direct PR is generally acceptable (opening an issue first is still encouraged for larger or unclear bugs).
 - Fork the Mesa repository.
 - [Clone your repository] from Github to your machine.
 - Create a new branch in your fork: `git checkout -b BRANCH_NAME`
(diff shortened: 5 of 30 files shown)

### CODE REVIEW

**Comment by SamipSGz** [comment]
Hi @souro26, thanks for the PR.

I ran the issue repro benchmark locally (`N=200000`, `cancel_ratio=0.9`, `runs=3`) and saw a clear improvement from ~0.619s avg (main) to ~0.330s avg (~46.6% faster, ~1.87x) with adaptive compaction.

One question on trigger accounting: this issue often cancels via direct `event.cancel()` (not only `EventList.remove(...)`). If       `_canceled_count` is only incremented in `remove()`, compaction might not trigger reliably for that path.                              
                                                                                                                                         
Would you be open to tracking compaction pressure based on what `pop_event()` actually discards (or otherwise accounting for direct  `event.cancel()` calls)?

File: mesa/time/events.py [hunk 1]
```diff
@@ -361,10 +362,29 @@ def pop_event(self) -> Event:
...
-                return event
+
+            if event.CANCELED:
+                if self._canceled_count > 0:
+                    self._canceled_count -= 1
+                continue
+
+            # Adaptive compaction check
+            if (
+                self._canceled_count > 0
```
quaquel: is this first check even necessary?

mesa/time/events.py
```diff
@@ -361,10 +362,29 @@ def pop_event(self) -> Event:
...
+
+            return event
+
         raise IndexError("Event list is empty")
 
+    def _compact(self) -> None:
+        """Remove canceled events from the heap when they dominate."""
+        self._events = [e for e in self._events if not e.CANCELED]
+        heapify(self._events)
+        self._canceled_count = 0
```
quaquel: The stress test you ran is really extreme. Can you do an analysis of how the cost of heapify/compact and popping canceled events affect each other?

mesa/time/events.py
souro26: yes that check is redundant, i'll remove it and simplify the code.

mesa/time/events.py
souro26: i ran a sweep across cancellation ratios from 10%–90%, n=200k, and 3 runs each to understand the interaction between skipping canceled entries and compaction cost.

approximate results (no compact vs compact + pop) :
10%: 0.70 vs 0.69
30%: 0.66 vs 0.51
50%: 0.67 vs 0.39
70%: 0.67 vs 0.21
90%: 0.68 vs 0.07


without compaction, pop time is fairly flat, between 0.66 and 0.70 seconds. heapify cost decreases as cancellation increases and the heap shrinks. compaction becomes clearly beneficial at around 20–30% cancellation. the 50% threshold is a bit on the higher side since it avoids rebuilds in moderate workloads while preventing degradation when cancellations dominate. i can adjust the trigger to be at around 30% since that's where a more significant difference can be seen.

quaquel: but what is the cost of the compacting itself?

souro26: the time complexity of compaction is linear. it consists of,
- filtering the heap list which is O(N)
- heapify on the remaining active elements which is O(A)

so total rebuild cost is O(N + A), which is linear in the heap size. in contrast, skipping canceled entries costs O(log N) per discarded element, so with C canceled entries the skip cost is roughly C * log N.

quaquel:
I have been thinking about this PR for the last few days. In my experience with event scheduling, cancelations are infrequent. So, I am not convinced the added complexity of this PR is worth it.

souro26:
In the normal case with few cancellations, this doesn’t change behavior or performance in any measurable way. It’s just a counter and one conditional in pop_event(), and compaction never triggers.

The goal was to prevent severe degradation when cancellations accumulate. Since lazy deletion keeps canceled entries in the heap, pop performance can degrade significantly once they dominate. The adaptive check just prevents that situation without changing semantics. I chose the 50% threshold to avoid frequent rebuilds but i can adjust it to 30% where the benefit becomes noticeable.

I’m open to simplifying the logic further. I can remove the persistent counter and instead trigger compaction based only on how many canceled entries are skipped during popping, which would reduce internal state further.

souro26:
I tried a few different approaches and ended up going with a simple `compact()` method instead. `pop_event()` works the same as before and nothing happens automatically.

If a model cancels a large number of events, `compact()` can be called to rebuild the heap and remove canceled entries. This keeps the core logic simple while still providing an option for heavy cancellation cases. I updated the tests and everything passes locally. Let me know if this approach works @quaquel.

mesa/time/events.py
```diff
@@ -407,10 +407,17 @@ def pop_event(self) -> Event:
...
         while self._events:
             event = heappop(self._events)
+
             if not event.CANCELED:
                 return event
+
         raise IndexError("Event list is empty")
 
+    def compact(self) -> None:
+        """Remove all canceled events from the heap and rebuild it."""
```
```suggestion
        """Remove all canceled events from the heap and rebuild it.
        
        If there are many canceled events, compaction can speed up performance substantially. """
```

PR #3360, opened 2026-02-21
fix: use weakref in EventGenerator to prevent memory leak

Krishsharma179:
### Summary
This PR fixes the memory leak by:
- Using weak references instead of hard references (like Event class)
- Adding error handling for invalid callables
- Implementing silent no-op when references become invalid
- Adding comprehensive tests

### Bug / Issue
 Pr for #3332  
EventGenerator.function is a hard reference rather than a weak ref as used in Event. This can lead to memory leaks if a generator is used in combination with, e.g., an Agent. If the agent is removed from the model but remains part of an EventGenerator, it cannot be garbage-collected.
### Implementation
As per the coversation in #3320   
The behavior of both Event and EventGenerator should be the same
so,
- if the function is not callable -> TypeError
- if the function is not weakrefable -> TypeError
- create weakref, drop local strong ref, then check wr() is None -> ValueError
- so we are failing fast during the initialization but if the weakref resolve to None we are just stoping the     generator(no-op)
- made a `__getstate__`  and `__setstate__` for  the pickling issue
### Testing
-  Error cases (non-callable, non-weakly-referenceable)
    It is work as expected `Event(5, lambda: 5)`  fail, while `a = lambada:10; ` `Event(5, a)` still     works.
- Weak reference behavior and garbage collection
-  Pickling/unpickling support
-  No-op behavior when weakref dies during execution

### Additional Notes
- Matches `Event` class weak reference pattern for consistency
- Maintains backward compatibility
- Comprehensive tests for weakrefs, garbage collection, and pickling
- Proper error handling for invalid callables

### DIFF
--- a/mesa/time/events.py
+++ b/mesa/time/events.py
@@ -333,7 +333,7 @@ class EventGenerator:
         state = self.__dict__.copy()
         fn = self.function() if self.function is not None else None
         state["_fn_strong"] = fn
-        state["_function"] = None
+        state["function"] = None
         return state
 
     def __setstate__(self, state):
--- a/tests/time/test_events.py
+++ b/tests/time/test_events.py
@@ -616,8 +616,8 @@ class TestEventGeneratorMemoryLeak(unittest.TestCase):
 
         # Verify state contains expected keys
         self.assertIn("_fn_strong", state)
-        self.assertIn("_function", state)
-        self.assertIsNone(state["_function"])
+        self.assertIn("function", state)
+        self.assertIsNone(state["function"])
 
         # Verify _fn_strong is the actual function
         self.assertEqual(state["_fn_strong"](), "hello")
@@ -637,7 +637,7 @@ class TestEventGeneratorMemoryLeak(unittest.TestCase):
         # 3. Test __setstate__ with None function (edge case)
         state_with_none = {
             "_fn_strong": None,
-            "_function": None,
+            "function": None,
             "schedule": schedule,
             "priority": Priority.DEFAULT,
             "_active": False,


Krishsharma179:
@quaquel This is the updated PR for the issue #3332 It passes all the checks just required your review please let me know if there is any changes required this PR follows the conversation #3320

quaquel:
Thanks for this cleaning and PR. I just merged #3320, which adds a new function for creating the weakref and do some tests on it. Can you update this PR by merging with upstream main and evaluate whether the new function is useful for this PR?

mesa/time/events.py
```diff
@@ -219,14 +219,48 @@ def __init__(
...
+            if isinstance(function, MethodType):
+                weak_fun = WeakMethod(function)
+            else:
+                weak_fun = ref(function)
+        except TypeError as e:
+            raise TypeError(
+                f"function {function} is not weakly referenceable. "
+                f"Use a user-defined function or method instead."
+            ) from e
+        # Here it will raise a value error if the function returns none
```
quaquel: all of this is replaceable by the new function.

mesa/time/events.py
```diff
@@ -219,14 +219,48 @@ def __init__(
...
+
+        del function  # dropping the strong ref
+
+        if weak_fun() is None:
+            raise ValueError(
+                "Cannot create EventGenerator with a function that is already garbage collected"
+            )
```
quaquel: we left this out of #3320, because it is very OS specific whether this adds value. I am still considering adding a check like this, but I need to check its behavior across different operating systems.

mesa/time/events.py
```diff
@@ -219,14 +219,48 @@ def __init__(
...
         self._active: bool = False
         self._current_event: Event | None = None
         self._execution_count: int = 0
 
+    @property
+    def function(self) -> Callable:
+        """Get the function. If garbage collected, returns no-op."""
+        func = self._function()
+
+        return func if func is not None else lambda: None
```
quaquel: why did you add this property?

mesa/time/events.py
```diff
@@ -258,7 +292,16 @@ def _execute_and_reschedule(self) -> None:
...
+        # FIXED: Check weakref HERE (execution time), not in property getter
+        # This matches Event class behavior - weakref check during execution
+        func = self._function()
+        if func is None:
+            # Stop the generator if weakref is dead
+            self.stop()
+            return  # Silent no-op (no error raised)
+
+        # Execute the function
+        func()
```
quaquel: I suggest not calling this variable func. This is a protected term in Python.

tests/test_lifespan.py
```diff
@@ -90,5 +95,137 @@ def test_agent_lifetime(self):  # noqa: D102
         assert lifetimes.Step.max() == 2
 
 
+class TestEventGeneratorMemoryLeak(unittest.TestCase):
```
quaquel: I suggest moving these tests into test_events.py, rather than here.

mesa/time/events.py
Krishsharma179: This property checks if the weak reference is dead (garbage collected) and then it will return none

mesa/time/events.py
Krishsharma179: ok i will change it

tests/test_lifespan.py
Krishsharma179: ok I will move it

mesa/time/events.py
quaquel: I know what the code does, I am asking why you deemed it necessary to put this into a property.

Krishsharma179:
> Thanks for this cleaning and PR. I just merged #3320, which adds a new function for creating the weakref and do some tests on it. Can you update this PR by merging with upstream main and evaluate whether the new function is useful for this PR?

Hello  @quaquel , I was going through the merged #3320 code and in the function 
````
if isinstance(function, types.FunctionType) and function.__name__ == "<lambda>":
        raise ValueError("function must be alive at Event creation.")
````
this line will raise the value error on both the cases
```
my_func = lambda x: x + 10
event = Event(my_func)  

# This is unsafe (inline), also blocked:
event = Event(lambda x: x + 10)  
````
Question: Is blocking Case 1 intentional? Since my_func holds a reference, it shouldn't be garbage collected. Should we allow stored lambdas, or is the strict blocking deliberate to avoid edge cases?

On the new function: Yes,  function is definitely useful for this PR - it centralizes the weak reference logic and makes the code cleaner. I'll update my implementation to use it.
Thanks!

**Comment by quaquel** [comment]
Assigning a a lambda function to a variable defeats the purpose of a lambda as an anonymous function. Most linters will also flag it as a problem and it's considered an anti pattern. So, I am fine with just blocking lambda's in general.

File: mesa/time/events.py (hunk 3 above)
Comment (Krishsharma179) [reply mesa/time/events.py]: Upon closer look, I realize the `@property` might be redundant.

We already check the weakref safely in `_execute_and_reschedule`:

```python
func = self._function()
if func is None:
    self.stop()
    return
````
 Since the safety check happens at execution time anyway, I'm okay to remove this @property completely and rely on the check in _execute_and_reschedule.

File: tests/test_lifespan.py (hunk 5 above)
Comment (quaquel) [reply tests/test_lifespan.py]: I meant all tests you added to test_lifespan.py

**Comment by Krishsharma179** [comment]
@quaquel  i have made the changes please check

File: mesa/time/events.py [hunk 6]
```diff
@@ -231,7 +231,10 @@ def __init__(
             priority: Priority level for generated events
         """
         self.model = model
-        self.function = function
+
+        weak_fun = _create_callable_reference(function)
+
+        self._function = weak_fun
```
```suggestion
        self._function = _create_callable_reference(function)
```

mesa/time/events.py
```diff
@@ -270,7 +273,16 @@ def _execute_and_reschedule(self) -> None:
         if not self._active:
             return
 
-        self.function()
+        # Check weakref HERE (execution time), not in property getter
+        # This matches Event class behavior - weakref check during execution
+        fn = self._function()
```
```suggestion
        # Check weakref HERE (execution time), not in property getter
        # This matches Event class behavior - weakref check during execution
        fn = self.function()
```

quaquel (approved):

PR #3363, opened 2026-02-23
Update mesa.examples models to use scenario

quaquel:
This PR updates the examples to use a custom scenario class. It also updates app.py and the tests. 

I left out Sugarscape since there is a PR open already that makes the change.

### DIFF
--- a/mesa/examples/advanced/alliance_formation/model.py
+++ b/mesa/examples/advanced/alliance_formation/model.py
@@ -16,7 +16,7 @@ class AllianceScenario(Scenario):
 
     n: int = 50
     mean: float = 0.5
-    std_dev: float = 0.2
+    std_dev: float = 0.1
 
 
 class MultiLevelAllianceModel(mesa.Model):


mesa/examples/advanced/alliance_formation/model.py
```diff
@@ -8,14 +8,23 @@
...
 )
+from mesa.experimental.scenarios import Scenario
+
+
+class AllianceScenario(Scenario):
+    """Scenario for the Alliance model."""
+
+    n: int = 50
+    mean: float = 0.5
+    std_dev: float = 0.2
```
EwoutH: You changed `std_dev` from 0.1 to 0.2, was that a conscious change?

mesa/examples/advanced/pd_grid/model.py
```diff
@@ -1,8 +1,20 @@
...
+from mesa.experimental.scenarios import Scenario
+
+
+class PrisonersDilemmaScenario(Scenario):
+    """Scenario for Prisoner's Dilemma model."""
+
+    width: int = 50
+    height: int = 50
+    activation_order: Literal["Sequential", "Random", "Simultaneous"] = "Random"
+    payoff: None | dict[tuple[str, str], float] = None
```
EwoutH, line 16: You changed `payoffs` to `payoff`, was this on purpose?

quaquel: It was not deliberate. The object in question is known as the payoff matrix in most literature so I just went with that without checking the original code

PR #3367, opened 2026-02-24
Refactor: Rename HasObservables to HasEmitters

Nithurshen:
### Summary
This PR refactors the core reactive programming mixin in the experimental `mesa_signals` module by renaming `HasObservables` to `HasEmitters`. This terminology change makes the framework more precise, as the base class manages generic event emitters (via the `@emit` decorator) rather than just observable states. 

### Bug / Issue
Resolves the renaming task outlined in the Mesa Signals tracking issue: #3227 

### Implementation
* `mesa/experimental/mesa_signals/core.py`: Renamed the `HasObservables` class to `HasEmitters`. Updated internal docstrings, type hints, and the `__all__` export list.
* `mesa/experimental/mesa_signals/`: Updated all internal imports and type hints across `batching.py`, `observable_collections.py`, and `__init__.py`.
* `mesa/model.py`: Updated the core `Model` class definition to inherit from the newly named `HasEmitters` mixin.
* `tests/`: Performed a global replacement of `HasObservables` to `HasEmitters` across the test suite to ensure all mock models and signal tests use the correct class.

### Testing
Ran the entire test suite locally using `pytest tests/`. 
* All 402 tests passed successfully, including the specific `test_mesa_signals.py` and `test_batching.py` suites. 
* Verified that the renaming caused no import errors or broken dependencies.

### Additional Notes
This PR addresses only the renaming bullet point from tracking issue #3227 to keep the diff clean and focused.

PR #3368, opened 2026-02-24
FEAT: Add class-level subscribe to HasEmitters

Nithurshen:
### Summary
This PR introduces the ability to subscribe to observable events at the class level via a new `@classmethod subscribe`. Subscriptions made this way are inherited by all instances of that class, meaning any instance emitting the specified signal will trigger the registered handler.

### Motive
This addresses a specific enhancement requested in the Mesa Signals tracking issue #3227. Previously, observing agents globally required relying on `model.register_agent` or manually binding `observe` to every instantiated object at runtime. This feature makes global data collection, UI updates, and logging significantly easier and more intuitive by allowing a one time class level binding.

### Implementation
* Modified `HasObservables.__init_subclass__` to initialize a distinct `_class_subscribers` dictionary for every subclass. This ensures that subscribing to a specific Agent subclass doesn't accidentally bind handlers to all other reactive classes.
* Added `@classmethod subscribe(...)` which mirrors the signature of the instance-level `observe` method but registers the handler globally for the class.
* Strictly utilized `create_weakref(handler)` for class-level subscribers. This directly addresses the concern raised in the tracking issue by ensuring that the global observer doesn't prevent model instances from being properly garbage-collected during large experiments or parameter sweeps.
* Updated `_has_subscribers` and `_mesa_notify` to seamlessly aggregate and dispatch messages to both instance-level and class-level listeners.
* Added `test_class_level_subscribe` to `tests/experimental/test_mesa_signals.py` to verify functionality across multiple instances.

### Usage Examples
You can now easily observe all instances of an agent class without needing to register handlers on them individually:

```python
from mesa.experimental.mesa_signals import HasObservables, Observable
from mesa.experimental.mesa_signals.signal_types import ObservableSignals

class MyAgent(HasObservables):
    state = Observable()

def state_change_handler(msg):
    agent = msg.owner
    old_val = msg.additional_kwargs.get("old")
    new_val = msg.additional_kwargs.get("new")
    print(f"Agent {agent} changed state from {old_val} to {new_val}")

MyAgent.subscribe("state", ObservableSignals.CHANGED, state_change_handler)

agent1 = MyAgent()
agent2 = MyAgent()

agent1.state = "active"   # Prints: Agent <...> changed state from None to active
agent2.state = "waiting"  # Prints: Agent <...> changed state from None to waiting
```

### Additional Notes
* This fulfills the third bullet point of tracking issue #3227.
* This PR is currently built using the existing HasObservables name. If the concurrent PR renaming this to HasEmitters is merged first (#3367), I am happy to rebase and resolve the naming conflict.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,7 +1,7 @@
 name: Performance benchmarks
 
 on:
-  pull_request_target:
+  pull_request:
     types: [opened, ready_for_review, labeled]
     branches:
       - main
@@ -9,17 +9,17 @@ on:
       - '**.py'
       - '.github/workflows/benchmarks.yml'
 
-permissions:
-  issues: write
-  pull-requests: write
-
 jobs:
   run-benchmarks:
+    permissions:
+      contents: read
     if: >
       github.event.action == 'labeled' && contains(github.event.pull_request.labels.*.name, 'trigger-benchmarks') ||
       github.event.action == 'opened' ||
       github.event.action == 'ready_for_review'
     runs-on: ubuntu-latest
+    outputs:
+      output: ${{steps.compare_timings.outputs.timing_comparison}}
     steps:
       # Python and dependency setup
       - name: Set up Python
@@ -51,7 +51,7 @@ jobs:
           path: benchmarks/timings_1.pickle
       - name: Checkout PR branch
         uses: actions/checkout@v6
-        if: github.event_name == 'pull_request_target'
+        if: github.event_name == 'pull_request'
         with:
           ref: ${{ github.event.pull_request.head.ref }}
           repository: ${{ github.event.pull_request.head.repo.full_name }}
@@ -72,9 +72,19 @@ jobs:
       # Run compare script and create comment
       - name: Run compare timings and encode output
         working-directory: benchmarks
+        id: compare_timings
         run: |
           TIMING_COMPARISON=$(python compare_timings.py | base64 -w 0)  # Base64 encode the output
-          echo "TIMING_COMPARISON=$TIMING_COMPARISON" >> $GITHUB_ENV
+          echo "timing_comparison=$TIMING_COMPARISON" >> $GITHUB_OUTPUT
+  comment_pr:
+    permissions:
+      pull-requests: write
+      issues: write
+    runs-on: ubuntu-latest
+    needs: run-benchmarks
+    env:
+      TIMING_COMPARISON: ${{needs.run-benchmarks.outputs.output}}
+    steps:
       - name: Comment PR
         uses: actions/github-script@v8
         with:


mesa/experimental/mesa_signals/core.py
```diff
@@ -288,9 +289,17 @@ def __init__(self, *args, **kwargs) -> None:
...
         key = (name, signal_type)
-        if key not in self.subscribers:
-            return False
-        return len(self.subscribers[key]) > 0
+
+        has_instance_subscribers = (
+            key in self.subscribers and len(self.subscribers[key]) > 0
+        )
+        has_class_subscribers = (
+            hasattr(self, "_class_subscribers")
```
quaquel: this check should not be needed if `_class_subscribers` is implemented correctly.

mesa/experimental/mesa_signals/core.py
```diff
@@ -449,18 +456,31 @@ def _mesa_notify(self, signal: Message):
...
         key = (observable, signal_type)
 
-        observers = self.subscribers[key]
-        active_observers = []
-        for observer in observers:
-            if active_observer := observer():
-                active_observer(signal)
-                active_observers.append(observer)
-            # use iteration to also remove inactive observers
+        if key in self.subscribers:
```
quaquel: this check seems redundant.

mesa/experimental/mesa_signals/core.py
```diff
@@ -449,18 +456,31 @@ def _mesa_notify(self, signal: Message):
...
+            active_class_observers = []
+            for observer in class_observers:
+                if active_class_observer := observer():
+                    active_class_observer(signal)
+                    active_class_observers.append(observer)
+
+            if active_class_observers:
+                self._class_subscribers[key] = active_class_observers
+            else:
+                del self._class_subscribers[key]
```
quaquel: This seems like code repetition. I am wondering whether this could not be implemented in a way that would avoid the apparent code duplication.

mesa/experimental/mesa_signals/core.py
```diff
@@ -511,6 +531,57 @@ def _process_signal_type(
         else:
             return signal_type
 
+    @classmethod
+    def subscribe(
```
quaquel: not sure about the naming here. We now have `Class.subcribe` and `Instance.observe`. What about e.g., `Class.observe_class` or something along those lines?

mesa/experimental/mesa_signals/core.py
```diff
@@ -511,6 +531,57 @@ def _process_signal_type(
...
+        Args:
+            observable_name: name of the Observable to subscribe to
+            signal_type: the type of signal on the Observable to subscribe to
+            handler: the handler to call
+        """
+        names = (
+            [observable_name]
+            if isinstance(observable_name, str)
+            else (cls.observables.keys() if observable_name is ALL else observable_name)
+        )
```
quaquel: please refactor this into something more readable.

mesa/experimental/mesa_signals/core.py
```diff
@@ -511,6 +531,57 @@ def _process_signal_type(
...
+                if st not in cls.observables[name]:
+                    raise ValueError(
+                        f"you are trying to subscribe to a signal of {st} "
+                        f"on Observable {name}, which does not emit this signal_type"
+                    )
+
+            ref = create_weakref(handler)
+            for st in signal_types:
+                cls._class_subscribers[(name, st)].append(ref)
+
```
quaquel, line 603: there seems to be overlap with `observe` code, so can this be simplified in a way that avoids code duplications?

quaquel:
1. What about `unobserve_class`? At a minimum, this class method should be there to mimic the instance level observe and unobserve.
2. What about clear_all_subscriptions? This currently clears everything at the instance level. How do we want to handle class level subscriptions in this? 

This second question points to a wider architectural question. At the moment, this PR maintains two lists of subscribers: instance-level subscriptions and class-level subscriptions. To me this is a sensible design. However, an alternative would be to let the instance at creation "inherit" all class-level subscriptions by adding these to the instance subscription dict. The advantage of this is that `clear_all_subscriptions` then does wat is says. However, it also means that any class level new subscriptions that happen after instantiation are not well defined. Thoughs on this are welcome.

Nithurshen:
Regarding the architectural question, I definitely think keeping the two list design is the most robust approach. If we were to let instances inherit class-level subscriptions by copying them into the instance dictionary during `__init__`, we would run into a few major issues:

1. If a user calls `observe_class` *after* some instances are already created, those older instances wouldn't receive the new subscription since their `__init__` has already run.
2. If a user later calls `unobserve_class`, we would have no way to remove that observer from already-instantiated objects without tracking every single instance in memory.
3. If a user calls `agent1.clear_all_subscriptions()`, it shouldn't unexpectedly wipe out a global class-level observer (like a DataCollector or UI renderer) that the rest of the model relies on. 

To resolve this and address the rest of your feedback, I have made the following changes:

* Added `unobserve_class` and `clear_all_class_subscriptions` to ensure class-level observers have the same lifecycle controls as instance-level ones.
* Refactored the internal logic into shared `_register_observer` and `_unregister_observer` helpers to keep the code DRY and eliminate the overlap between the class and instance methods.
* Merged the dispatch logic in `_mesa_notify` into a single loop over `(self.subscribers, self._class_subscribers)`, utilizing `.get()` to remove the redundant `if key in...` checks.
* Typed `_class_subscribers` as a `ClassVar` and initialized it on the base class, completely removing the need for the `hasattr` check overhead. 
* Renamed the method to `observe_class` for consistency, and refactored `_process_name` and `_process_signal_type` into `@classmethod`s to remove the nested ternary operators.

Let me know if this implementation aligns with your vision.

mesa/experimental/mesa_signals/core.py
```diff
@@ -344,49 +364,64 @@ def unobserve(
...
+        cls._unregister_observer(
+            cls._class_subscribers, observable_name, signal_type, handler
+        )
 
-                    if remaining:
-                        self.subscribers[key] = remaining
-                    else:
-                        del self.subscribers[key]
+    @classmethod
+    def _clear_all_subscriptions_from_dict(cls, subs_dict: dict, name: ObservableName):
```
quaquel:
```suggestion
    def _clear_all_subscriptions(cls, subs_dict: dict, name: ObservableName):
```

mesa/experimental/mesa_signals/core.py
```diff
@@ -344,49 +364,64 @@ def unobserve(
...
-                del self.subscribers[k]
-        else:
-            for n in name:
-                keys_to_remove = [k for k in self.subscribers if k[0] == n]
-                for k in keys_to_remove:
-                    del self.subscribers[k]
-                # ignore when unsubscribing to Observables that have no subscription
+        self._clear_all_subscriptions_from_dict(self.subscribers, name)
+
+    @classmethod
```
quaquel, line 417: The indentation seems off here?

mesa/experimental/mesa_signals/core.py
```diff
@@ -489,17 +525,19 @@ def suppress(self):
 
         return _SuppressContext(self)
 
-    def _process_name(self, name: ObservableName) -> Iterable[str]:
+    @classmethod
+    def _process_name(cls, name: ObservableName) -> Iterable[str]:
         """Convert name to an iterable of observable names."""
         if name is ALL:
-            return self.observables.keys()
+            return cls.observables.keys()
```
quaquel, line 534: I don't think this is correct, unless I am missing something.

Nithurshen: Since `_process_name` is a class method, we can only use `cls` instead of `self`.

mesa/experimental/mesa_signals/core.py
Nithurshen: I don't think so.

mesa/experimental/mesa_signals/core.py
quaquel: ~Yes but that is why process_name cannot be a class method. observables is an instance-level dict that contains the instance-level subscriptions.~

nevermind, I am mixing observables and subscribers.

Nithurshen:
@quaquel, Is there still anything you wish to be changed?

mesa/experimental/mesa_signals/core.py
```diff
@@ -308,27 +319,36 @@ def observe(
...
+        signal_type: SignalSpec,
+        handler: Callable,
+    ):
+        """Shared logic to validate and unregister an observer from a dictionary."""
+        names = cls._process_name(observable_name)
+        target_signals = cls._process_signal_type(signal_type)
 
-            signal_types = target_signals or self.observables[name]
+        for name in names:
+            signal_types = target_signals or cls.observables[name]
```
quaquel: You might want to clarify the logic here. If signal_type is ALL, `target_signals` returns None and that is handled nex in this statement.

quaquel:
Sorry took me a while to review the code. It looks fine with one remaining request.

PR #3376, opened 2026-02-24
Add RUN_ENDED signal for automatic final state capture in DataRecorder

yashhzd:
## Summary

Addresses the `run_ended` handling sub-task from #3341.

`DataRecorder` subscribes to `model.time` CHANGED, which records at the **old** time value. This means the final simulation state after the last step is never automatically captured — users had to manually call `recorder.finalise()`.

This PR:
- Adds `ModelSignals.RUN_ENDED` signal type
- Decorates `Model.run_model()` with `@emit("model", ModelSignals.RUN_ENDED)` so the signal fires after the run loop completes
- `BaseDataRecorder` subscribes to `RUN_ENDED` and auto-calls `finalise()`
- `finalise()` now includes per-dataset time-based deduplication, making it safe to call multiple times (e.g. both via signal and manually)

### Design decisions

- **Time-based recording preserved**: Following @quaquel's guidance in #3341, the existing time-change-driven collection is untouched. `RUN_ENDED` is additive — it only captures the final snapshot.
- **Lightweight deduplication**: `_last_finalized_times` dict tracks the last finalized time per dataset. Calling `finalise()` twice at the same model time is a no-op. Dedup resets naturally when time advances.
- **Works across all recorder backends**: `_on_run_ended` and dedup logic live in `BaseDataRecorder`, so `DataRecorder`, `JSONDataRecorder`, `ParquetDataRecorder`, and `SQLDataRecorder` all benefit automatically.
- **No signal on manual `step()`**: Only `run_model()` emits `RUN_ENDED`. Manual stepping does not trigger it, keeping the signal semantically correct.

### Files changed

| File | Change |
|------|--------|
| `mesa/experimental/mesa_signals/signal_types.py` | Added `RUN_ENDED` to `ModelSignals` |
| `mesa/model.py` | `@emit("model", ModelSignals.RUN_ENDED)` on `run_model()` |
| `mesa/experimental/data_collection/basedatarecorder.py` | Subscribe to `RUN_ENDED`, dedup-aware `finalise()` |
| `tests/experimental/test_datarecorder.py` | 14 new tests |

## Test plan

- [x] `test_run_ended_signal_exists` — signal type defined
- [x] `test_run_ended_signal_is_observable` — "model" registered as observable
- [x] `test_run_model_emits_run_ended` — signal actually fires
- [x] `test_run_ended_auto_finalise_captures_final_state` — final model + agent state recorded
- [x] `test_finalise_deduplication` — no duplicates on repeated calls
- [x] `test_finalise_deduplication_resets_on_new_time` — dedup allows new times
- [x] `test_run_ended_with_manual_stepping` — works with 1-step runs
- [x] `test_run_ended_disabled_dataset_not_collected` — respects enabled/disabled
- [x] `test_run_ended_with_json_recorder` — JSON backend captures final state
- [x] `test_run_ended_with_sql_recorder` — SQL backend captures final state
- [x] `test_run_ended_data_completeness` — complete timeline 0..N captured
- [x] `test_run_ended_model_val_progression` — values match expected progression
- [x] `test_step_without_run_model_no_run_ended` — manual step() does NOT emit
- [x] `test_multiple_recorders_both_receive_run_ended` — multiple recorders all get signal
- [x] All 316 existing tests pass (0 regressions)
- [x] ruff check passes with 0 issues

### DIFF
--- a/mesa/experimental/data_collection/basedatarecorder.py
+++ b/mesa/experimental/data_collection/basedatarecorder.py
@@ -28,7 +28,7 @@ from typing import TYPE_CHECKING, Any
 
 import pandas as pd
 
-from mesa.experimental.mesa_signals import ObservableSignals
+from mesa.experimental.mesa_signals import ModelSignals, ObservableSignals
 
 if TYPE_CHECKING:
     from mesa import Model
@@ -141,6 +141,9 @@ class BaseDataRecorder(ABC):
         if self.registry is None:
             raise AttributeError("Model must have a DataRegistry (model.data_registry)")
 
+        # Track last finalized time per dataset to prevent duplicate snapshots
+        self._last_finalized_times: dict[str, float] = {}
+
         if config is None:
             config = {}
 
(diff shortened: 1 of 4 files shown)


quaquel:
There is an ongoing discussion in #3347 and we have not yet agreed on a solution. Moreover, run_model will be removed in favor of run_for, run_until, and a to be implemented run_while or something equivalent.

yashhzd:
Thanks @quaquel, appreciate the quick feedback. I wasn't aware `run_model` was being replaced — that makes sense. I'll follow along in #3347 and once the `run_for`/`run_until`/`run_while` API is finalized, I'd be happy to adapt the signal work to target those methods instead. Sorry for the premature PR.

PR #3380, opened 2026-02-26
Tighten exception types and wrap missing-cell lookups

falloficaruss:
### Summary
This PR applies a targeted exception cleanup in `discrete_space`:
- It keeps domain-specific errors only where they add API value:
   - `DiscreteSpace.__getitem__` now raises `CellMissingException` instead of leaking internal `KeyError`.
- It uses built-in Python exceptions for normal input validation:
   - `TypeError` for wrong argument types (`Grid.torus`, `Grid.capacity`, `Network.layout`).
   - `ValueError` for invalid-but-well-typed values (e.g., out-of-bounds non-torus grid lookups).

### What changed
- `DiscreteSpace.__getitem__` now wraps internal `KeyError` as `CellMissingException`.
  - This hides internal dict details and exposes a domain-specific error to users.
- Type-validation checks now use `TypeError` (instead of `ValueError`) where appropriate:
  - `Grid._validate_parameters`:
    - `torus` must be `bool`
    - `capacity` must be `int | float | None`
  - `Network.__init__`:
    - `layout` must be `Mapping` or `Callable`
- Updated docstring in `Grid.find_nearest_cell` to match actual raised exception (`ValueError` for out-of-bounds non-torus positions).

### Tests
- Added coverage for `CellMissingException` on invalid coordinate access.
- Added/updated validation tests to assert `TypeError` for type mismatches.
- Updated spatial lookup test to expect `TypeError` for invalid `Network(layout=...)`.

### DIFF
--- a/mesa/discrete_space/discrete_space.py
+++ b/mesa/discrete_space/discrete_space.py
@@ -3,7 +3,7 @@
 DiscreteSpace provides the core functionality needed by all cell-based spaces:
 - Cell creation and tracking
 - Agent-cell relationship management
-- Property layer support
+- Property Layer support
 - Random selection capabilities
 - Capacity management
 
@@ -40,7 +40,7 @@ class DiscreteSpace[T: Cell](ABC):
         random (Random): The random number generator
         cell_klass (Type) : the type of cell class
         empties (CellCollection) : collection of all cells that are empty
-        property_layers (dict[str, PropertyLayer]): the property layers of the discrete space
+        property_layers (dict[str, np.ndarray]): property_layer of the discrete space
 
     Notes:
         A `UserWarning` is issued if `random=None`. You can resolve this warning by explicitly
--- a/mesa/discrete_space/grid.py
+++ b/mesa/discrete_space/grid.py
@@ -16,7 +16,7 @@ from __future__ import annotations
 import copyreg
 import math
 from collections.abc import Sequence
-from itertools import product
+from itertools import chain, product
 from random import Random
 from typing import Any, TypeVar
 
@@ -24,10 +24,6 @@ import numpy as np
 from scipy.spatial import KDTree
 
 from mesa.discrete_space import Cell, DiscreteSpace
-from mesa.discrete_space.property_layer import (
-    HasPropertyLayers,
-    create_property_accessors,
-)
 
 T = TypeVar("T", bound=Cell)
 
@@ -45,7 +41,7 @@ def unpickle_gridcell(parent, fields):
     cell_klass = type(
         "GridCell",
         (parent,),
-        {"_mesa_properties": set(), "__slots__": ()},
+        {"property_layers": set(), "__slots__": ()},
     )
     instance = cell_klass(
         (0, 0)
@@ -59,7 +55,7 @@ def unpickle_gridcell(parent, fields):
     return instance
 
 
-class Grid(DiscreteSpace[T], HasPropertyLayers):
+class Grid(DiscreteSpace[T]):
     """Base class for all grid classes.
 
     Attributes:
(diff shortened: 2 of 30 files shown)


mesa/discrete_space/cell.py
```diff
@@ -231,7 +232,7 @@ def _neighborhood(
         of recursion to avoid RecursionError on large radius values.
         """
         if radius < 1:
-            raise ValueError("radius must be larger than one")
```
quaquel, line 234: This is an overarching comment: where normal python errors are clearly appropriate, I prefer to use those. So here, value error is fully appropriate and I fail to see a need to replace those with custom mesa errors.

mesa/discrete_space/discrete_space.py
```diff
@@ -173,7 +173,10 @@ def __iter__(self):  # noqa
         return iter(self._cells.values())
 
     def __getitem__(self, key: tuple[int, ...]) -> T:  # noqa: D105
-        return self._cells[key]
+        try:
+            return self._cells[key]
+        except KeyError as e:
+            raise CellMissingException(key) from e
```
quaquel, line 177: Here, I am fine with wrapping the key error. It is not obvious to the user that DiscretSpace uses a dictionary internally.

mesa/discrete_space/cell.py
quaquel: As far as I can tell, you only changed this line. Can you carefully review the PR and if you are replacing a value or type error with a custom error, indicate why you think that is appropriate?

falloficaruss: I've refined the exception migration based on your feedback. I've focused the custom MesaException hierarchy on cases where it provides domain-specific context or hides internal implementation details.

For standard argument validation (type checking, value ranges), I've reverted to using built-in ValueError and TypeError as they are more than sufficient for those tasks and follow standard Python conventions.

mesa/discrete_space/grid.py
```diff
@@ -168,9 +168,9 @@ def _validate_parameters(self):
         if not all(isinstance(dim, int) and dim > 0 for dim in self.dimensions):
             raise ValueError("Dimensions must be a list of positive integers.")
         if not isinstance(self.torus, bool):
-            raise ValueError("Torus must be a boolean.")
+            raise TypeError("Torus must be a boolean.")
         if self.capacity is not None and not isinstance(self.capacity, float | int):
-            raise ValueError("Capacity must be a number or None.")
+            raise TypeError("Capacity must be a number or None.")
```
falloficaruss, line 279: Changed to TypeError because this is type validation, not value-range validation.

mesa/discrete_space/network.py
```diff
@@ -58,7 +58,7 @@ def __init__(
         elif isinstance(layout, Mapping):
             node_positions = layout
         else:
-            raise ValueError(
+            raise TypeError(
```
falloficaruss, line 61: this is validating the type/form of layout (Mapping or Callable expected). Invalid type should be TypeError

tests/discrete_space/test_spatial_lookups.py
```diff
@@ -93,7 +93,7 @@ def test_network_lookups():
     G.add_node(0)
     G.add_node(1)
 
-    with pytest.raises(ValueError):
+    with pytest.raises(TypeError):
```
falloficaruss, line 96: test must match production behavior after Network(..., layout=[]), changed to raise TypeError.

falloficaruss:
I narrowed the scope of this PR:
Instead of broadly replacing `ValueError`/`TypeError` with Mesa-specific exceptions across `discrete_space`, this PR now does two focused things:

Keep built-in exceptions for standard argument validation (`TypeError` for wrong types, `ValueError` for invalid values).
Use a Mesa-specific exception only where it improves the public API (`CellMissingException` in `DiscreteSpace.__getitem__`, instead of exposing internal `KeyError` behavior).

quaquel:
@falloficarus22 thanks for the update. I'll try to review later today. Might it be an idea to write up something in the contributers guidlines sumarizing how to handle exceptions in Mesa?

falloficaruss:
> Might it be an idea to write up something in the contributers guidlines sumarizing how to handle exceptions in Mesa?

This is indeed important. We can add a subsection under testing and code standards in CONTRIBUTING.md.
I'll address that in a seperate PR

PR #3165, opened 2026-01-17
Add Solara visualization feature

Saranyadharani:
## Summary
This PR introduces an **experimental Solara-based visualization utility** for Mesa models.

It provides a lightweight wrapper that allows Mesa models to be visualized using Solara
when the optional dependency is available, while remaining safe to import and use when
Solara is not installed.

Key points:
- Adds an optional Solara-based visualization entry point
- Ensures Mesa continues to work without Solara installed
- Improves test stability for visualization-related code

## Testing
- Ran `pytest tests/test_solara_viz.py` → all tests pass ✅
- Verified that the visualization function can be imported and called without errors
- Confirmed behavior when Solara is installed and when it is absent

## Scope & Limitations
- This PR does **not** introduce Solara as a required dependency
- The visualization is intentionally minimal and experimental
- UI richness and interactivity will be expanded in follow-up work

## Checklist
- [x] Code follows existing Mesa style and conventions
- [x] Tests updated and passing
- [x] No breaking changes introduced

## Notes for Mentors
This PR is an initial step toward improving Mesa’s visualization ecosystem and is aligned
with my proposed GSoC 2026 focus on modern, optional visualization backends.

PR #3381, opened 2026-02-26
Add check for positional-only parameters in _check_model_params

### LINKED ISSUE #2494: SolaraViz docs

Harshini2411:
**Summary**

Added a validation check in `_check_model_params` to detect positional-only parameters in model `__init__` signatures and raise a clear, descriptive error instead of a confusing crash.

**Bug / Issue**

Fixes #2494

`SolaraViz` recreates model instances using `model_parameters`, meaning all parameters must accept keyword arguments. If a model's `__init__` has positional-only parameters (defined before `/` in the signature), SolaraViz would crash with a confusing error message giving no hint about the real cause.

**Implementation**

Added a `has_positional_only` check in `_check_model_params` in `mesa/visualization/solara_viz.py`. It uses `inspect.Parameter.POSITIONAL_ONLY` to detect parameters defined before `/` and raises a `ValueError` with a clear message explaining the issue and how to fix it.

**Testing**

Added `test_check_model_params_with_positional_only` in `tests/visualization/test_solara_viz.py`. The test creates a model with a positional-only parameter and verifies that `_check_model_params` raises a `ValueError` with the expected message. All 92 existing tests continue to pass.

**Additional Notes**

A previous PR (#3165) was opened for this issue but was closed without merging. This PR addresses the same gap.


quaquel:
The plan for Mesa 4 is to disallow positional only arguments.

PR #3386, opened 2026-02-26
Fix: Enforce strict layout validation in Network space

### LINKED ISSUE #3385: Network initialization crashes when layout dictionary is missing node positions
**Describe the bug**
Network space initialization incorrectly handles nodes that are missing from the layout mapping. When a node ID is not found in the layout, `node_positions.get(node_id)` returns None. However, the code immediately wraps this in a NumPy array: `pos = np.array(None)`.

Because a NumPy array (even one wrapping None) is not the Python None singleton, the subsequent guard if pos is not None: evaluates to True. This causes the invalid position to be passed to `scipy.spatial.KDTree`, which crashes the initialization with a `ValueError`.

**Expected behavior**
The Network space should identify nodes without specified positions and skip them when building the spatial `KDTree`. If the resulting list of positions is empty, it should gracefully set `self._kdtree = None`, allowing the space to function in a purely topological mode (as intended and documented).

**To Reproduce**
```python
import networkx as nx
from mesa.discrete_space import Network

# 1. Create a graph with nodes
G = nx.Graph()
G.add_nodes_from([1, 2, 3])

# 2. Attempt to initialize Network with an empty layout dictionary
# This raises: ValueError: data must be of shape (n, m) (from SciPy KDTree)
net = Network(G, layout={})
```

**Additional context**
Suggested Fix: Check for None before wrapping the position in a NumPy array

Nithurshen:
### Summary

This PR fixes a bug where initializing a `Network` space with an incomplete or empty `layout` dictionary resulted in a cryptic SciPy `ValueError`. It introduces strict validation to explicitly catch missing node positions before spatial processing begins.

### Bug / Issue

Fixes #3385

When a `layout` dictionary was provided but was missing entries for certain node IDs, `node_positions.get(node_id)` returned `None`. This `None` value was incorrectly wrapped into a NumPy array (`np.array(None)`), which bypassed the subsequent `pos is not None` guard. This invalid position was then passed to `scipy.spatial.KDTree`, causing a crash with the cryptic error.

### Implementation

* Modified the `__init__` method in `mesa/discrete_space/network.py`. Extracted the position lookup (`raw_pos = node_positions.get(node_id)`) and added an explicit `None` check *before* NumPy conversion. If a node is missing from the layout, the initialization now immediately halts and raises a `SpaceException` with a clear, Mesa-specific error message detailing the exact missing Node ID.
* Removed the redundant `if pos is not None:` guard later in the loop since positions are now strictly validated.

### Testing

* Added a new test `test_network_missing_layout_node` in `tests/discrete_space/test_discrete_space.py`.
* Verified that a completely empty layout (`layout={}`) and a partially missing layout both correctly raise the expected `SpaceException`.

### Additional Notes

As discussed in the issue thread with the maintainers, this strict validation approach was chosen over silently falling back to a purely topological network mode. This prevents masking potential bugs in the user's data generation code where they might mistakenly expect spatial queries to work for dropped nodes.

### DIFF
--- a/mesa/discrete_space/network.py
+++ b/mesa/discrete_space/network.py
@@ -59,7 +59,7 @@ class Network(DiscreteSpace[Cell]):
         elif isinstance(layout, Mapping):
             node_positions = layout
         else:
-            raise ValueError(
+            raise TypeError(
                 "Incorrect Layout Argument.\nShould be either `Mapping` or `Callable`"
             )
 
--- a/mesa/discrete_space/discrete_space.py
+++ b/mesa/discrete_space/discrete_space.py
@@ -171,7 +171,10 @@ class DiscreteSpace[T: Cell](ABC):
         return iter(self._cells.values())
 
     def __getitem__(self, key: tuple[int, ...]) -> T:  # noqa: D105
-        return self._cells[key]
+        try:
+            return self._cells[key]
+        except KeyError as e:
+            raise CellMissingException(key) from e
 
     @property
     def empties(self) -> CellCollection[T]:
--- a/mesa/discrete_space/grid.py
+++ b/mesa/discrete_space/grid.py
@@ -236,7 +236,7 @@ class Grid(DiscreteSpace[T]):
             Cell: The cell containing the position
 
         Raises:
-            KeyError: If position is outside grid bounds and not a torus
+            ValueError: If position is outside grid bounds and not a torus
         """
         # Floor to get cell coordinate
         coord = tuple(np.floor(position).astype(int))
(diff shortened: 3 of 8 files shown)


mesa/discrete_space/network.py
```diff
@@ -67,7 +68,14 @@ def __init__(
 
         # Create cells and gather KD-Tree data simultaneously
         for node_id in self.G.nodes:
-            pos = np.array(node_positions.get(node_id))
+            raw_pos = node_positions.get(node_id)
```
quaquel: would it not make more sense to use: 


```suggestion
        try:
            raw_pos = node_positions[node_id]
        except KeyError as e:
            raise SpaceException() from e
```

Nithurshen: Yes

Issue #3379, opened 2026-02-25
Performance bottlenecks in mesa.discrete_space


souro26:
**Summary**

While profiling mesa.discrete_space under a high-movement workload, I observed significant time spent in cell movement and occupancy checks. This issue summarizes the hotspots and proposes a few improvements.

**Reproduction**
````python
import random
import cProfile
from mesa import Model
from mesa.discrete_space.grid import OrthogonalVonNeumannGrid
from mesa.discrete_space.cell_agent import CellAgent


class DummyModel(Model):
    pass


def run_benchmark():
    model = DummyModel()
    grid = OrthogonalVonNeumannGrid((100, 100), capacity=1)

    agents = [CellAgent(model) for _ in range(5000)]

    # initial placement
    for agent in agents:
        agent.cell = grid.select_random_empty_cell()

    # random movement loop
    for _ in range(500):
        for agent in agents:
            neighbors = agent.cell.neighborhood
            target = random.choice(list(neighbors))
            if target.is_empty:
                agent.cell = target


if __name__ == "__main__":
    cProfile.run("run_benchmark()", sort="cumtime")
````
**Observations**

- `Cell.is_empty` is called very frequently and currently computes `len(self._agents) == 0` every time.
- `CellAgent.cell` triggers add or remove operations that account for a large portion of total runtime.
- Neighborhood access wraps results in `CellCollection` and constructs intermediate mappings, contributing additional allocation overhead.

Total runtime is around 63 seconds
````
CellAgent.cell            ~14.7s
Cell.is_empty             ~11.1s
Cell.neighborhood         ~6.8s
Cell.remove_agent         ~6.1s
Cell.add_agent            ~4.8s
````
**Proposed Improvements**

- Make `_empty` the single source of truth for emptiness and avoid recomputing `len(self._agents)` on every `is_empty` call.
- Reduce overhead in the movement pathnto avoid unnecessary repeated checks and work.
- Simplify how neighborhoods are constructed and returned to reduce intermediate allocations during frequent movement.

I plan to address these in separate PRs to keep changes focused and measurable.

PR #3388, opened 2026-02-26
Optimize CellCollection internals and neighborhood access

souro26:
### Summary
This PR improves the performance of `CellCollection` and related neighborhood access by reducing temporary allocations and avoiding unnecessary list construction in hot paths.

Part of the optimization effort mentioned in #3379 

### Motive
`CellCollection` and neighborhood access are performance-critical components used heavily in discrete space models.Profiling showed repeated list construction and temporary allocations in `select_random_cell` , neighborhood handling and cell collection iteration paths These allocations add overhead in tight simulation loops, especially when called millions of times.

The goal of this PR is to reduce overhead without changing public behavior or API semantics.

### Implementation 
- Avoided unnecessary list construction in `select_random_cell` by using direct indexing with `randrange`.
- Reduced temporary allocations in `CellCollection` internal storage.
- Preserved mapping-style access `(collection[cell])` and `_capacity` attribute for backward compatibility.

### Usage Examples
All existing usage patterns remain identical:
````python
collection = cell.neighborhood
cell = collection.select_random_cell()
agents = collection[cell]
````
Behavior is unchanged.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -45,7 +45,7 @@ jobs:
         run: python global_benchmark.py
       # Upload benchmarks, checkout PR branch, download benchmarks
       - name: Upload benchmark results
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: timings-main
           path: benchmarks/timings_1.pickle
@@ -61,7 +61,7 @@ jobs:
       - name: Install Mesa of the PR branch
         run: pip install --no-deps .
       - name: Download benchmark results
-        uses: actions/download-artifact@v7
+        uses: actions/download-artifact@v8
         with:
           name: timings-main
           path: benchmarks
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,7 +41,7 @@ jobs:
         run: hatch build
       - name: Upload package as artifact to GitHub
         if: github.repository == 'mesa/mesa' && startsWith(github.ref, 'refs/tags')
-        uses: actions/upload-artifact@v6
+        uses: actions/upload-artifact@v7
         with:
           name: package
           path: dist/
--- a/benchmarks/configurations.py
+++ b/benchmarks/configurations.py
@@ -1,6 +1,12 @@
 """configurations for benchmarks."""
 
-from mesa.examples import BoidFlockers, BoltzmannWealth, Schelling, WolfSheep
+from mesa.examples import (
+    BoidFlockers,
+    BoltzmannWealth,
+    Schelling,
+    SugarscapeG1mt,
+    WolfSheep,
+)
 
 configurations = {
     # BoltzmannWealth Model Configurations
@@ -88,6 +94,38 @@ configurations = {
                 "grass": True,
                 "grass_regrowth_time": 10,
                 "sheep_gain_from_food": 4.0,
+            },
+        },
+    },
+    SugarscapeG1mt: {
+        "small": {
+            "seeds": 50,
+            "replications": 5,
+            "steps": 50,
+            "parameters": {
+                "initial_population": 100,
+                "enable_trade": False,
+                "vision_min": 1,
+                "vision_max": 2,
+                "endowment_min": 25,
+                "endowment_max": 50,
+                "metabolism_min": 1,
+                "metabolism_max": 5,
+            },
+        },
+        "large": {
+            "seeds": 10,
+            "replications": 3,
+            "steps": 50,
+            "parameters": {
+                "initial_population": 250,
+                "enable_trade": True,
+                "vision_min": 4,
+                "vision_max": 5,
+                "endowment_min": 25,
+                "endowment_max": 50,
+                "metabolism_min": 1,
+                "metabolism_max": 5,
             },
         },
     },
--- a/mesa/examples/advanced/sugarscape_g1mt/agents.py
+++ b/mesa/examples/advanced/sugarscape_g1mt/agents.py
@@ -225,6 +225,10 @@ class Trader(CellAgent):
             if cell.is_empty
         ]
 
+        if not neighboring_cells:
+            # all neighboring cells are occupied
+            return
+
         # 2. determine which move maximizes welfare
 
         welfares = [
@@ -273,6 +277,14 @@ class Trader(CellAgent):
 
         if self.is_starved():
             self.remove()
+
+    def step(self):
+        """Agent step method."""
+        self.prices = []
+        self.trade_partners = []
+        self.move()
+        self.eat()
+        self.maybe_die()
 
     def trade_with_neighbors(self):
         """
--- a/mesa/examples/advanced/sugarscape_g1mt/model.py
+++ b/mesa/examples/advanced/sugarscape_g1mt/model.py
@@ -5,6 +5,7 @@ import numpy as np
 import mesa
 from mesa.discrete_space import OrthogonalVonNeumannGrid
 from mesa.examples.advanced.sugarscape_g1mt.agents import Trader
+from mesa.experimental.scenarios import Scenario
 
 
 # Helper Functions
@@ -20,19 +21,23 @@ def geometric_mean(list_of_prices):
     """
     find the geometric mean of a list of prices
     """
+    # protects against an invalid value if no prices
+    if len(list_of_prices) == 0:
+        return -1
     return np.exp(np.log(list_of_prices).mean())
 
 
-def get_trade(agent):
-    """
-    For agent reporters in data collector
+class SugarScapeScenario(Scenario):
+    """Sugarscape scenario class."""
 
-    return list of trade partners and None for other agents
-    """
-    if isinstance(agent, Trader):
-        return agent.trade_partners
-    else:
-        return None
+    initial_population: int = 200
+    endowment_min: int = 25
+    endowment_max: int = 50
+    metabolism_min: int = 1
+    metabolism_max: int = 5
+    vision_min: int = 1
+    vision_max: int = 5
+    enable_trade: bool = True
 
 
 class SugarscapeG1mt(mesa.Model):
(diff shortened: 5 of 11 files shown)

### CODE REVIEW

File: mesa/discrete_space/cell.py [hunk 1]
```diff
@@ -215,7 +216,7 @@ def get_neighborhood(
             a list of all neighboring cells
 
         """
-        return CellCollection[Cell](
+        return CellCollection(
```
quaquel: why did you remove `[Cell]`

mesa/discrete_space/cell.py
```diff
@@ -224,7 +225,7 @@ def get_neighborhood(
     @cache  # noqa: B019
     def _neighborhood(
         self, radius: int = 1, include_center: bool = False
-    ) -> dict[Cell, list[Agent]]:
+    ) -> Mapping[Cell, list[Agent]] | tuple[Cell, ...]:
```
quaquel: I really don't like the idea of changing the return type here. Why did you do it and what are alternative designs?

At least one option would be to move getting `cell.agent` into the constructor of the `CellCollection`. And simplify everything inside Cell.

mesa/discrete_space/cell_collection.py
```diff
@@ -58,30 +58,46 @@ def __init__(
             cells: cells to add to the collection
             random: a seeded random number generator.
         """
-        if isinstance(cells, dict):
+        if isinstance(cells, Mapping):
```
quaquel: why?

mesa/discrete_space/cell_collection.py
```diff
@@ -58,30 +58,46 @@ def __init__(
...
+            self._capacity = next(iter(self._cells.keys())).capacity
+        elif not self._is_mapping and self._cells:
+            self._capacity = self._cells[0].capacity
+        else:
+            self._capacity = None
 
         if random is None:
             warnings.warn(
-                "Random number generator not specified, this can make models non-reproducible. Please pass a random number generator explicitly",
+                "Random number generator not specified, this can make models non-reproducible. "
```
quaquel: how did this layout change happen?

mesa/discrete_space/cell_collection.py
```diff
@@ -58,30 +58,46 @@ def __init__(
...
             self._cells = cells
+            self._is_mapping = True
         else:
-            self._cells = {cell: cell._agents for cell in cells}
-
-        # Get capacity from first cell if collection is not empty
-        self._capacity: int | None = (
-            next(iter(self._cells.keys())).capacity if self._cells else None
-        )
+            self._cells = tuple(cells)
```
quaquel: Why did you change this. 

If I understand the change correctly, getting `cell.agents` is done lazily in some cases now.

quaquel:
I left several comments. I appreciate the idea of trying to enhance the performance of the code, but I am not convinced by the changes themselves. They make the implementation more difficult to understand and the performance benefits are small or even negative from the looks of it.

mesa/discrete_space/cell.py
souro26: I removed it while simplifying internal typing during refactor, but there is no strong reason to drop the specialization here. I’ll restore [Cell] for clarity.

mesa/discrete_space/cell.py
souro26: I changed the return type to avoid temporary dictionary construction in the radius == 1 fast path. In that case, _neighborhood was previously building a {cell: cell._agents} mapping even though `CellCollection` immediately uses it. Returning a tuple of cells avoids constructing that intermediate dictionary. That said, i understand the concern about changing the return type since changing the structure of _neighborhood for performance reasons might make the abstraction less clear.

I was thinking of a few other alternative designs we could consider :

1. Keeping _neighborhood purely spatial and always returning cells such as tuple[Cell, ...]. `CellCollection` would then consistently handle mapping cells to agents internally.

2. Keeping the current dictionary return type and optimizing it how it is constructed without changing its return type.

3. Introducing an internal lazy view to avoid immediate dictionary creation, though I think this will likely increase complexity beyond what the performance gain justifies.

From a design perspective , option 1 seems the cleanest since _neighborhood determines membership and `CellCollection` handles grouping and agent access. That keeps a single return type and keeps performance optimizations within the collection layer. I can refactor accordingly and rerun the benchmarks, particularly focusing on the Schelling regression.

mesa/discrete_space/cell_collection.py
souro26: I changed dict to Mapping to make the constructor slightly more general, since _neighborhood may conceptually return mapping-like objects. There isn’t a strong requirement for that generalization here though. If keeping dict makes it simpler, I can revert my change.

mesa/discrete_space/cell_collection.py
souro26: The change wasn't intentional, it might have been automatically reformatted by ruff. I'll revert the change.

mesa/discrete_space/cell_collection.py
souro26: My intention to change to `self._cells = tuple(cells)` was to avoid constructing a dict when the neighborhood is already represented as a sequence of cells specially in the radius == 1 case. Previously, _neighborhood always built a dictionary and accessed `neighbor._agents`, even if the caller only needed to iterate over cells or select a random one. The tuple-based approach avoids that temporary dictionary allocation and defers agent access until explicitly requested.

Introducing two internal representations does add branching and increases complexity in `CellCollection` so this may not be the cleanest trade-off. I was thinking of a cleaner alternative where `CellCollection` always stores cells as a tuple and lets it resolve agents lazily in a single, consistent way. That would remove `_is_mapping`, eliminate branching logic, preserve behavior, and still avoid unnecessary dictionary construction. 

I'll test the  simplified version and rerun benchmarks to see if we can keep the performance benefit without increasing complexity.

souro26:
Thanks for the detailed feedback. I'm going to try out a few alternative approaches to reduce complexity and re run Schelling benchmarks to reduce the performance loss.

quaquel:
Don't focus on just Schelling. This benchmark is mainly about finding random empty cells but has less effect on the other changes this PR is making.

mesa/discrete_space/cell_collection.py
```diff
@@ -58,30 +58,46 @@ def __init__(
...
                 stacklevel=2,
             )
             random = Random()
+
         self.random = random
 
     def __iter__(self):  # noqa
+        if self._is_mapping:
+            return iter(self._cells)
         return iter(self._cells)
```
codebreaker32: Ain't both same?

mesa/discrete_space/cell_collection.py
```diff
@@ -58,30 +58,46 @@ def __init__(
...
-    def __getitem__(self, key: T) -> Iterable[CellAgent]:  # noqa
-        return self._cells[key]
+    def __getitem__(self, key: T) -> Iterable[CellAgent]:
+        """Return the agents associated with the given cell key."""
+        if self._is_mapping:
+            return self._cells[key]
+        else:
+            # emulate mapping behavior
+            if key in self._cells:
+                return key.agents
```
codebreaker32: Under the hood in the `Cell` class, the `.agents` explicitly returns a copy().
The original dictionary implementation returned a direct reference to the internal `_agents` list. Returning a copy instead means a new list is allocated in memory every single time a user accesses a cell's agents through the collection.

mesa/discrete_space/cell_collection.py
```diff
@@ -58,30 +58,46 @@ def __init__(
...
 
-    def __getitem__(self, key: T) -> Iterable[CellAgent]:  # noqa
-        return self._cells[key]
+    def __getitem__(self, key: T) -> Iterable[CellAgent]:
+        """Return the agents associated with the given cell key."""
+        if self._is_mapping:
+            return self._cells[key]
+        else:
+            # emulate mapping behavior
+            if key in self._cells:
```
codebreaker32: This forces a O(n) search. Since spatial lookups happen quite often per step in most ABMs, it may cause severe performance degradation for any model that relies heavily on spatial queries

souro26:
Thanks for the detailed feedback @quaquel and @codebreaker32 . I’ve simplified the implementation so `_neighborhood` now always returns `tuple[Cell, ...]` and is purely spatial. It no longer builds a `{cell: cell._agents}` mapping. `CellCollection` now stores a tuple of cells, and I removed the dual mapping logic and branching entirely. `__getitem__` returns `cell._agents` directly, so there is no .agents copy and no additional list allocation.

Regarding the O(n) membership check in `__getitem__` this only applies when `collection[cell]` is used. The previous implementation constructed a dictionary for every neighborhood, which faced allocation cost on every call. The current version avoids that direct allocation and only performs a linear check if indexing is explicitly used. If needed, the membership check can be removed, but I kept it for clearer error semantics.

Let me know if you’d prefer dropping the membership check or adjusting anything further.

mesa/discrete_space/cell_collection.py
```diff
@@ -92,15 +90,15 @@ def __repr__(self):  # noqa
...
 
     @property
     def agents(self) -> Iterable[CellAgent]:  # noqa
-        return itertools.chain.from_iterable(self._cells.values())
+        return itertools.chain.from_iterable(cell._agents for cell in self._cells)
 
     def select_random_cell(self) -> T:
         """Select a random cell."""
-        return self.random.choice(self.cells)
+        return self._cells[self.random.randrange(len(self._cells))]
```
quaquel: In my `%%timeit` testing, this is slower than using choice. Please refert this change

souro26: i have reverted the change back to random.choice

quaquel:
I am going to run some tests locally, but the benchmarks suggest mixed results. So, I am not yet convinced we should proceed in this direction.

souro26:
Thanks for taking a closer look. Given the mixed benchmark results and the added complexity, I’ll think about simplifying further or other alternative approaches. I’ll wait for your local test results before deciding the next step.

quaquel:
I have run it locally, and on both boltzmann and schelling it is notably slower. I am testing on mac os, python 3.13, against main. In short, I don't see the claimed performance improvements in the benchmark cases. I am inclined to close this PR and suggest a more piecemeal approach with detailed micro benchmarking first.

souro26:
Thanks for running the benchmarks locally and taking the time to review this in detail. This PR got too messy, i think it's better to close this one. I'll explore smaller, more focused optimizations with micro-benchmarks so that the impact of each change can be evaluated independently.

Issue #3378, opened 2026-02-25
ModelDataSet window_size eviction causes unpacking crash


ShreyasN707:
Hi,
While testing `ModelDataSet` with a sliding window (`DatasetConfig(window_size=N)`), I ran into a crash once the window limit is reached.

In `DataRecorder._store_dataset_snapshot`, eviction assumes blocks are stored as `(time, data)` tuples:

```python
_old_time, old_data = storage.blocks[0]
```
However, for modeldataset, blocks are stored as a single merged dict:
```
row = {**data, "time": time}
storage.blocks.append(row)
```

So when eviction tries to unpack storage.blocks[0], it ends up unpacking a dict and raises:
```
ValueError: too many values to unpack
```
This only happens when:

- Using ModelDataSet

- Setting window_size

- Running long enough to trigger eviction

It looks like an internal inconsistency — other dataset types store (time, data), but modeldataset stores a dict directly.
A minimal fix would be to store `modeldataset `blocks as (time, data) like the others, and handle merging "time" during DataFrame conversion instead.

I can open a small PR if this approach is correct.

souro26:
I think your diagnosis is right. Right now `_store_dataset_snapshot` assumes that blocks are stored as (time, data) and that’s what the eviction logic is unpacking. But `modeldataset` is the only case where we don’t follow that pattern. It stores a merged dict with time already inside it. So once the window fills up, eviction tries to unpack a dict and crashes.

This also means `modeldataset` is the only dataset type that doesn’t follow the same internal storage shape, which is probably not intentional. Storing model blocks as (time, data) like the others and merging time during `_convert_modelDataSet` would make the storage format consistent again and fix the crash without changing external behavior.

PR #3389, opened 2026-02-26
Fix sliding window eviction crash in ModelDataSet

ShreyasN707:
Fixes a crash in `ModelDataSet` when `window_size` is enabled by standardizing how blocks are stored internally.

###  Issue: #3378 
When the sliding window eviction runs, `_evict_blocks` expects blocks in the format `(time, data)`.  
However, dict-based data was being stored as a merged dictionary (`{"time": time, **data}`), which caused a `ValueError` during unpacking and crashed the data collection process.

### Implementation
Unified the storage format so all blocks (including dict data) are stored as `(time, data)` tuples.  
The merge into `{"time": time, **data}` is now handled only when building the final DataFrame.

Core change:

```python
# datarecorders.py

case dict():
    storage.blocks.append((time, data))
    storage.total_rows += 1
rows = [{**data, "time": time} for time, data in storage.blocks]
return pd.DataFrame(rows)

### DIFF
--- a/benchmarks/configurations.py
+++ b/benchmarks/configurations.py
@@ -1,6 +1,12 @@
 """configurations for benchmarks."""
 
-from mesa.examples import BoidFlockers, BoltzmannWealth, Schelling, WolfSheep
+from mesa.examples import (
+    BoidFlockers,
+    BoltzmannWealth,
+    Schelling,
+    SugarscapeG1mt,
+    WolfSheep,
+)
 
 configurations = {
     # BoltzmannWealth Model Configurations
@@ -88,6 +94,38 @@ configurations = {
                 "grass": True,
                 "grass_regrowth_time": 10,
                 "sheep_gain_from_food": 4.0,
+            },
+        },
+    },
+    SugarscapeG1mt: {
+        "small": {
+            "seeds": 50,
+            "replications": 5,
+            "steps": 50,
+            "parameters": {
+                "initial_population": 100,
+                "enable_trade": False,
+                "vision_min": 1,
+                "vision_max": 2,
+                "endowment_min": 25,
+                "endowment_max": 50,
+                "metabolism_min": 1,
+                "metabolism_max": 5,
+            },
+        },
+        "large": {
+            "seeds": 10,
+            "replications": 3,
+            "steps": 50,
+            "parameters": {
+                "initial_population": 250,
+                "enable_trade": True,
+                "vision_min": 4,
+                "vision_max": 5,
+                "endowment_min": 25,
+                "endowment_max": 50,
+                "metabolism_min": 1,
+                "metabolism_max": 5,
             },
         },
     },
--- a/mesa/examples/advanced/sugarscape_g1mt/agents.py
+++ b/mesa/examples/advanced/sugarscape_g1mt/agents.py
@@ -225,6 +225,10 @@ class Trader(CellAgent):
             if cell.is_empty
         ]
 
+        if not neighboring_cells:
+            # all neighboring cells are occupied
+            return
+
         # 2. determine which move maximizes welfare
 
         welfares = [
@@ -273,6 +277,14 @@ class Trader(CellAgent):
 
         if self.is_starved():
             self.remove()
+
+    def step(self):
+        """Agent step method."""
+        self.prices = []
+        self.trade_partners = []
+        self.move()
+        self.eat()
+        self.maybe_die()
 
     def trade_with_neighbors(self):
         """
--- a/mesa/examples/advanced/sugarscape_g1mt/model.py
+++ b/mesa/examples/advanced/sugarscape_g1mt/model.py
@@ -5,6 +5,7 @@ import numpy as np
 import mesa
 from mesa.discrete_space import OrthogonalVonNeumannGrid
 from mesa.examples.advanced.sugarscape_g1mt.agents import Trader
+from mesa.experimental.scenarios import Scenario
 
 
 # Helper Functions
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

File: tests/experimental/test_modeldataset_window.py
Comment (Nithurshen) [inline tests/experimental/test_modeldataset_window.py:1]: My suggestion would be to add this test to existing file `tests/experimental/test_datarecorder.py` instead of a separate file.

Comment (ShreyasN707) [reply tests/experimental/test_modeldataset_window.py:1]: Yes I have already done it will push it tonight.

**Comment by ShreyasN707** [comment]
@EwoutH Tests are passing now — updated the dict storage to (time, data).

**Comment by quaquel** [comment]
@codebreaker32 could you take a look at this?

**Review by codebreaker32 — COMMENTED** [review]
Good catch and thanks for this PR @ShreyasN707 However I have some minor requests:

1. Please remove "time" from the storage.metadata["columns"] as well, keeping it in the metadata columns will cause empty DataFrames to have an inconsistent schema.
2. Could you add a specific test that explicitly guards against the ValueError crash?

I will give it a final look once you push the changes

File: mesa/experimental/data_collection/datarecorders.py [hunk 1]
```diff
@@ -127,7 +124,7 @@ def _store_dataset_snapshot(
                 if "type" not in storage.metadata:
                     storage.metadata["type"] = "custom"
 
-        # Update bookkeeping for evicted data
+        """Update bookkeeping for evicted data"""
```
codebreaker32: Please don't use triple quotes for inline comments

mesa/experimental/data_collection/datarecorders.py
```diff
@@ -169,13 +166,16 @@ def get_table_dataframe(self, name: str) -> pd.DataFrame:
         storage = self.storage[name]
 
         if not storage.blocks:
-            # Empty DataFrame with correct columns
+            """ Empty DataFrame with correct columns"""
```
codebreaker32: Please don't use triple quotes for inline comments

mesa/experimental/data_collection/datarecorders.py
```diff
@@ -184,7 +184,7 @@ def get_table_dataframe(self, name: str) -> pd.DataFrame:
             case "modeldataset":
                 return self._convert_modelDataSet(storage)
             case _:
-                # Fallback
+                """Fallback"""
```
codebreaker32: Same as above

tests/experimental/test_datarecorder.py
```diff
@@ -286,29 +285,41 @@ def test_data_recorder_window_eviction_list():
...
+    model.step()
+    model.step()
 
     storage = recorder.storage["agent_data"]
     assert len(storage.blocks) == 2
 
 
-def test_data_recorder_window_eviction_dict():
-    """Test window eviction bookkeeping for dict data."""
+def test_modeldataset_window_eviction_crash():
```
codebreaker32: Why did you change the name?

tests/experimental/test_datarecorder.py
```diff
@@ -286,29 +285,41 @@ def test_data_recorder_window_eviction_list():
...
+    """Test that window eviction on ModelDataSet does not crash and produces a correct DataFrame."""
     model = MockModel(n=5)
     recorder = DataRecorder(model, config={"model_data": DatasetConfig(window_size=2)})
     recorder.clear()
 
-    # Fill window
-    model.step()  # 1
-    model.step()  # 2
-    model.step()  # 3 - should evict first
```
codebreaker32, line 307: Why have you deleted comments?

tests/experimental/test_datarecorder.py
```diff
@@ -323,10 +334,9 @@ def test_data_recorder_window_eviction_custom():
     recorder = DataRecorder(model, config={"custom_data": DatasetConfig(window_size=2)})
     recorder.clear()
 
-    # Fill window
     recorder.collect()
     recorder.collect()
-    recorder.collect()  # Should evict first
```
codebreaker32, line 329: Same as above

codebreaker32:
I still couldn't see the test that explicitly guards against the `ValueError` mentioned in issue

tests/experimental/test_datarecorder.py
ShreyasN707: Looks like I removed a few comments unintentionally during conflict resolution. My bad I’ll correct that and update the PR.

ShreyasN707:
> I still couldn't see the test that explicitly guards against the `ValueError` mentioned in issue

I’ll recheck the branch it’s possible the test didn’t get pushed properly. I’ll confirm and fix it right away.

mesa/experimental/data_collection/datarecorders.py
```diff
@@ -171,11 +168,14 @@ def get_table_dataframe(self, name: str) -> pd.DataFrame:
         if not storage.blocks:
             # Empty DataFrame with correct columns
             columns = storage.metadata.get("columns", [])
+
+            if "time" not in columns:
+                columns = [*columns, "time"]
```
codebreaker32: Why?

mesa/experimental/data_collection/datarecorders.py
```diff
@@ -226,9 +226,10 @@ def _convert_agentDataSet(self, storage: DatasetStorage) -> pd.DataFrame:
     def _convert_modelDataSet(self, storage: DatasetStorage) -> pd.DataFrame:
         """Convert model dict blocks to DataFrame."""
         if not storage.blocks:
-            return pd.DataFrame(columns=storage.metadata.get("columns", []))
+            return pd.DataFrame(columns=[*storage.metadata.get("columns", []), "time"])
```
codebreaker32: See above

tests/experimental/test_datarecorder.py
```diff
@@ -307,8 +307,48 @@ def test_data_recorder_window_eviction_dict():
...
+    assert isinstance(first_block, tuple)
+    assert len(first_block) == 2
+
+    _time, data = first_block
+    assert isinstance(data, dict)
+    assert "model_val" in data
+
+
+def test_modeldataset_window_eviction_no_valueerror():
+    """Explicit regression test for ValueError during sliding window eviction (#3378)."""
```
codebreaker32: It seems repetitive to me. Isn't it similar to above test? I am fine with above tests only

ShreyasN707: It overlapped with the existing dict eviction test. The only difference was an explicit try/except for the ValueError, but pytest already fails on uncaught exceptions, so it was redundant.

mesa/experimental/data_collection/datarecorders.py
ShreyasN707: Yes adding "time" there was unnecessary; I'll remove it to keep the schema aligned  with metadata-defined columns.

quaquel:
tests are failing, so that needs to be addressed before merging this PR.

mesa/experimental/data_collection/datarecorders.py
```diff
@@ -226,7 +223,8 @@ def _convert_modelDataSet(self, storage: DatasetStorage) -> pd.DataFrame:
         if not storage.blocks:
             return pd.DataFrame(columns=storage.metadata.get("columns", []))
```
codebreaker32:
```suggestion
            columns = storage.metadata.get("columns", [])
            return pd.DataFrame(columns=[*columns, "time"])
```

PR #3391, opened 2026-02-27
Add `first`, `last`, `any`, and `count` query methods to AgentSet

EwoutH:
### Summary
Adds four lightweight query methods to `AbstractAgentSet` that allow finding and counting agents without materializing intermediate collections. These are broadly useful across ABM and DES models and follow familiar patterns from Python builtins and pandas.

### Motive
When building models, particularly event-driven ones, a common need is to find a single agent matching a condition (e.g., an idle server, an empty cell, an available vehicle). Currently this requires `select()` which allocates a full `AgentSet`, or manual iteration:

```python
# Current — allocates a new AgentSet just to get one agent
server = model.agents_by_type[Server].select(lambda s: s.is_idle).to_list()[0]

# Or manual iteration
server = None
for s in model.agents_by_type[Server]:
    if s.is_idle:
        server = s
        break
```
Now with this PR, you can just do:
```python
server = model.agents_by_type[Server].first(lambda s: s.is_idle)
```
Similarly, checking existence or counting matches requires `select()` plus `len()`, which is wasteful when you only need a boolean or an integer.

This need was identified while building an M/M/c queue example for Mesa's new DES capabilities.

### Implementation
Four methods are added to `AbstractAgentSet`, so both `AgentSet` (weak refs) and `_HardKeyAgentSet` (strong refs) inherit them:

- **`first(filter_func, *, no_match)`**: Returns the first matching agent. Short-circuits on first match.
- **`last(filter_func, *, no_match)`**: Returns the last matching agent. Requires full iteration (no random access guarantee on `AbstractAgentSet`).
- **`any(filter_func)`**: Returns `True` if any agent matches. Short-circuits.
- **`count(filter_func)`**: Returns the number of matching agents. Without a filter, delegates to `len()`.

The `no_match` parameter on `first`/`last` controls behavior when no agent matches:
- `"none"` (default): returns `None` and emits a `UserWarning` (consistent with existing AgentSet warning patterns)
- `"error"`: raises `ValueError`

`any` and `count` don't need `no_match` since `False`/`0` are always valid returns.

### Usage Examples

```python
# Find the first idle server
server = model.agents_by_type[Server].first(lambda s: s.is_idle)
if server:
    server.start_service(customer)

# Check if any agent needs help
if model.agents.any(lambda a: a.health <= 0):
    dispatch_medic()

# Count agents matching a condition
n_idle = model.agents_by_type[Server].count(lambda s: s.is_idle)

# Get the last agent added (no filter)
newest = model.agents.last()

# Strict mode — raise if no match found
leader = model.agents.first(lambda a: a.is_leader, no_match="error")
```

### Additional Notes
- All methods are O(n) worst case, with `first` and `any` short-circuiting on match.
- `last` could be optimized for `_HardKeyAgentSet` with `reversed()` in the future, but correctness-first for now.
- No changes to existing methods or behavior.
- This was motivated by the M/M/c queue DES example but is broadly useful for all Mesa models.

For review, the handeling of zero agents available in first and last is probably the biggest design decision that should be discussed.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature-request.md
+++ b/.github/ISSUE_TEMPLATE/feature-request.md
@@ -10,5 +10,8 @@ about: Suggest a new feature for Mesa
 **Describe the solution you'd like**
 <!-- A clear and concise description of what you want to happen. -->
 
+**Are you planning to open a PR for this?**
+<!-- For enhancements/new features, please wait for maintainer approval on this issue/discussion before opening a PR. -->
+
 **Additional context**
 <!-- Add any other context, links, etc. about the feature here. -->
--- a/.github/PULL_REQUEST_TEMPLATE/bug.md
+++ b/.github/PULL_REQUEST_TEMPLATE/bug.md
@@ -1,3 +1,9 @@
+> [!NOTE]
+> Use this template for bug fixes only. For enhancements/new features, use the feature template and get maintainer approval in an issue/discussion before opening a PR.
+
+### Pre-PR Checklist
+- [ ] This PR is a bug fix, not a new feature or enhancement.
+
 ### Summary
 <!-- Provide a brief summary of the bug and its impact. -->
 
@@ -14,3 +20,4 @@ If you're fixing the visualisation, add before/after screenshots. -->
 
 ### Additional Notes
 <!-- Add any additional information that may be relevant for the reviewers, such as potential side effects, dependencies, or related work.
+-->
--- a/.github/PULL_REQUEST_TEMPLATE/feature.md
+++ b/.github/PULL_REQUEST_TEMPLATE/feature.md
@@ -1,3 +1,12 @@
+> [!IMPORTANT]
+> For enhancements and new features, include the issue/discussion where a maintainer approved this work before opening a PR. Unapproved feature/enhancement PRs may be closed.
+
+### Pre-PR Checklist
+- [ ] I linked an issue/discussion where a maintainer approved this enhancement/feature for implementation.
+
+### Approval Link
+<!-- Link to the issue/discussion comment where a maintainer confirmed this is ready for a PR. -->
+
 ### Summary
 <!-- Provide a concise summary of the feature and its purpose. -->
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,3 +2,5 @@ Thanks for opening a PR! Please click the `Preview` tab and select a PR template
 
 - [🐛 Bug fix](?expand=1&template=bug.md)
 - [🛠 Feature/enhancement](?expand=1&template=feature.md)
+
+Feature/enhancement PRs require prior maintainer approval in an issue or discussion before they are accepted.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,8 @@ discuss via [Matrix] OR via [an issue].
 
 **To submit a contribution**
 
-- Create a ticket for the item that you are working on.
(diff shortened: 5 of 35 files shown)


mesa/agentset.py
```diff
@@ -225,6 +225,112 @@ def to_list(self) -> list[A]:
         """
         return list(self._agents.keys())
 
+    def first(
+        self,
+        filter_func: Callable[[A], bool] | None = None,
+        *,
+        no_match: Literal["none", "error"] = "none",
```
quaquel, line 232: I am not in favor of this keyword argument. Just always raise an exception and let the caller decide how to handle it.

mesa/agentset.py
```diff
@@ -225,6 +225,112 @@ def to_list(self) -> list[A]:
...
+
+        Returns:
+            True if at least one agent matches, False otherwise.
+        """
+        if filter_func is None:
+            return len(self) > 0
+
+        return any(filter_func(agent) for agent in self)
+
+    def count(self, filter_func: Callable[[A], bool] | None = None) -> int:
```
quaquel, line 319: is this needed given that we have select or already?

mesa/agentset.py
```diff
@@ -225,6 +225,112 @@ def to_list(self) -> list[A]:
...
+
+            warnings.warn(
+                "No matching agent found in AgentSet, returning None.",
+                UserWarning,
+                stacklevel=2,
+            )
+
+        return result
+
+    def any(
```
quaquel, line 299: what is a use case for this?

mesa/agentset.py
EwoutH: Thanks for the swift review!

I agree this is a hard part of the AgentSet API. Don't have a clear view on what's best.

mesa/agentset.py
EwoutH: The current approach would be `len(agentset.select(condition))` I think? So than the question is, if there's value in `agentset.count(condition)` as shortcut.

EwoutH:
Okay I explored some more options, and I agree that always raising an error is the cleanest approach.

Now that we have custom error types, lets use them `AgentLookupError` could subclass `LookupError` (Python's base for `KeyError`/`IndexError`).

In model code, there are two situations:

```python
# 1. No match is a bug — just call it, let it raise
home = agents.first(lambda c: c.is_home)

# 2. No match is normal control flow — try/except/else
try:
    server = agents_by_type[Server].first(lambda s: s.is_idle)
except AgentLookupError:
    self.queue.append(customer)
else:
    server.start_service(customer)
```

This keeps the API surface minimal (no `any` needed, no default parameters), has zero allocations and short-circuits, and the `try/except/else` pattern is idiomatic Python (EAFP). The well-named exception makes intent clear at every call site.

If we ever want to add more (like `one_of`), we can adopt the same pattern.

mesa/agentset.py
quaquel: we have len(agentset) and len(agentset.select(condition)). I don't see a direct use case for `count`. To me, it violates the idea that there should be one obvious way to do something.

EwoutH:
I will drop `count` for now.

What's your stance on a `reverse` method to reverse the current order of the agentset?

quaquel:
> What's your stance on a reverse method to reverse the current order of the agentset?

AgentSet is described as being an ordered set, so reversing it can make sense. Not sure about the use case for it?

quaquel:
@EwoutH, any plans to pick this up again soonish?

EwoutH:
For now I don’t have a clean approach to handle empty sets. So closing for now.

PR #3392, opened 2026-02-27
Add tooltip support to Altair backend

### LINKED ISSUE #2795: Display info when clicking agents
Build an easy way to display in the front end some info about the agent (health, status, position, etc.) when the agent is clicked on the simulation

**Additional context**
It would be a very useful feature to have for simulations where agents carry a lot of information, which define their behavior, either to easily follow a precise agent throughout a simulation to study behavior or simply for debugging.

Special mention to @Sahil-Chhoker : given that you're currently working with GSoC to encapsulate visual styling through your `AgentPortrayalStyle` module, would it make sense to extend this class (or add a related interface) to include metadata or callbacks for click events?

That would also be particularly handy for mesa-llm to display the reasoning and the current discussions of an agent.

annapurna-gupta:
### Summary
Add  tooltip support to the Altair visualization backend so additional agent data can be shown on hover. 

### Motive
 Enable showing extra agent information on hover in Altair visualizations. Earlier, there was no simple way to add this hover data.

### Implementation
 1. Added an optional tooltip field to AgentPortrayalStyle so users can pass extra info to show on hover.

 2. Updated the Altair backend to collect this data, add it to the DataFrame, and display it using Altair tooltips.

3. Added a check in the Matplotlib backend to raise an error if tooltip is used (since Matplotlib is not interactive).

4. Updated an example and tests to show how the feature works. 

### Usage Examples
`def agent_portrayal(agent):
    return AgentPortrayalStyle(
        color=agent.wealth,
        tooltip={"Agent ID": agent.unique_id, "Wealth": agent.wealth},
    )`

### DIFF
--- a/mesa/visualization/backends/altair_backend.py
+++ b/mesa/visualization/backends/altair_backend.py
@@ -263,8 +263,7 @@ class AltairBackend(AbstractRenderer):
         ylabel = kwargs.pop("ylabel", "")
 
         # Tooltip list for interactivity
-        # FIXME: Add more fields to tooltip (preferably from agent_portrayal)
-        tooltip_list = ["x", "y"]
+        tooltip_list = []
 
         # Find ALL unique keys
         all_tooltips_key = set()


sakshi-thigale:
Thanks for the implementation — this is a nice addition to the Altair backend.

I had a small observation regarding the tooltip handling in draw_agents. We currently iterate over arguments["tooltip"] multiple times and perform per-row df.loc updates inside nested loops.

Since the Schelling benchmarks show a slight runtime increase, I’m wondering if the row-wise DataFrame mutation might be contributing.

It might be worth considering pre-aggregating tooltip values column-wise and extending the DataFrame in a more vectorized way instead of mutating individual cells.

Not blocking — just a performance consideration.

souro26:
The tooltip handling in `draw_agents` currently does nested loops with per-cell `df.loc[i, key] = value` writes. With n agents and k tooltip fields per agent, the algorithm is O(n·k). The main issue is the repeated scalar .loc writes, since those are expensive in pandas and add a large constant cost. That likely contributes to the Schelling regression in the benchmarks.

A cleaner approach would be to compute the full set of tooltip keys once, pre-allocate python lists of length n for each key., fill them in one O(n·k) pass and assign each column to the df once.

This avoids repeated df mutation and should reduce the overhead. Given the measurable runtime increase, I think this part is worth restructuring.

souro26:
There’s also an issue in how tooltip columns are determined.

Currently, the keys are taken from the first non-None tooltip only. If a later agent provides additional tooltip keys, those columns will still be created in the df via df.loc, but they won’t be included in `tooltip_list`. This can result in a mismatch between the df columns and the configured Altair tooltips.

A safer approach would be to compute the union of tooltip keys across all agents before constructing `tooltip_list`, for example by collecting all keys from non-None tooltips first and then extending `tooltip_list` with that full set.

annapurna-gupta:
Thanks @souro26  and @SAKSHI-DOT693 for the helpful feedback!

I'll fix:
1. Collect ALL unique tooltip keys (not just the first).
2. Pre-build columns vectorized instead of using df.loc in loops

This will fix both the performance regression and ensure all tooltip fields show up. Updating the PR now.

quaquel:
> Since the Schelling benchmarks show a slight runtime increase, I’m wondering if the row-wise DataFrame mutation might be contributing.

Just to clarify: the benchmarks only run the models without the visualization being involved. There is some variability simply due to using github actions.

mesa/visualization/backends/altair_backend.py
```diff
@@ -264,6 +267,30 @@ def draw_agents(
...
+
+        for i, tooltip in enumerate(arguments["tooltip"]):
+            if tooltip:
+                for key, value in tooltip.items():
+                    column_data[key][i] = value
+
+        for key, values in column_data.items():
+            df[key] = values
+
+        tooltip_list.extend(all_tooltips_key)
```
souro26: Since `all_tooltips_key` is a set, the tooltip column order may change between runs. Sorting the keys before extending `tooltip_list` would keep the order consistent.

annapurna-gupta: ok, got you.

mesa/visualization/backends/matplotlib_backend.py
```diff
@@ -144,6 +144,12 @@ def collect_agent_data(self, space, agent_portrayal, default_size=None):
                 # Set defaults if not provided
                 if aps.x is None and aps.y is None:
                     aps.x, aps.y = self._get_agent_pos(agent, space)
+                if aps.tooltip is not None:
+                    raise ValueError(
```
quaquel: Should this be an exception or just a warning? I would probably go with a warning personally.

annapurna-gupta: Yes, a warning would be better than stopping the execution. I’ll update this accordingly.

quaquel:
Thanks for this PR. I tested it locally, and it seems to be working with 1 caveat: why are x,y included? I guess this is some default behavior somewhere. However, with the explicit control this PR adds, I suggest we remove the coordinates so that only what the user specifies is included in the tooltip.

annapurna-gupta:
> Thanks for this PR. I tested it locally, and it seems to be working with 1 caveat: why are x,y included? I guess this is some default behavior somewhere. However, with the explicit control this PR adds, I suggest we remove the coordinates so that only what the user specifies is included in the tooltip.

Thanks and  I’ll remove the default `x` and `y` so only specified fields are included.

PR #3400, opened 2026-02-28
Fix ContinuousSpaceAgent.pos setter to alias position

### LINKED ISSUE #3399: `ContinuousSpaceAgent.pos` assignment does not update position
**Describe the bug**
`ContinuousSpaceAgent` exposes a `pos` property. The getter returns `self.position`, but the setter is currently a no-op (`pass`). As a result, assigning `agent.pos = ...` silently does nothing, which can lead to confusing behavior and unnecessary debugging

**Expected behavior**
`pos` should behave as a true alias for `position`: setting `agent.pos = value` should update `agent.position` and preserve existing torus correction.

**To Reproduce**
```
import numpy as np
from mesa import Model

m = Model(rng=42)
space = ContinuousSpace(np.asarray([[0, 1], [0, 1]]), torus=False, random=m.random)
a = ContinuousSpaceAgent(space, m)

a.position = [0.2, 0.2]
a.pos = [0.7, 0.7]  # currently a silent no-op

assert np.allclose(a.position, [0.7, 0.7])  # fails

```

**Additional context**
Low severity (workaround is using agent.position directly), but the silent no-op is surprising and easy to miss.
Added regression coverage in `tests/experimental/test_continuous_space.py`

MrAnayDongre:
### Summary
`ContinuousSpaceAgent.pos` is intended as a compatibility alias for position, but its setter was a silent no-op. This made `agent.pos = ...` do nothing without warning, which can lead to confusing behavior and wasted debugging time (especially in interactive/visualization workflows).

### Bug / Issue
Fixes #3399

`ContinuousSpaceAgent.pos` getter returns `self.position`, but the setter previously contained pass, so assignments like `agent.pos = [x, y]` did not update the agent’s position. Expected behavior is for `pos` to behave as a true alias of position, preserving the existing torus correction.

### Implementation
* Updated `ContinuousSpaceAgent.pos` setter to delegate to the existing position setter `(self.position = value)`, ensuring identical behavior (including torus correction and bounds checks).

* Added regression assertions to ensure `pos` matches `position` behavior.

### Testing
```
pytest -q tests/experimental/test_continuous_space.py::test_continuous_agent
pytest -q tests/experimental/test_continuous_space.py
```
Pre-commit hooks ran on commit (ruff format/check, etc.)

### Additional Notes
This is a low-severity behavioral fix with minimal scope. it only makes the compatibility property behave consistently with its getter and does not change position semantics.

### DIFF
--- a/mesa/experimental/continuous_space/continuous_space_agents.py
+++ b/mesa/experimental/continuous_space/continuous_space_agents.py
@@ -49,9 +49,9 @@ class ContinuousSpaceAgent(Agent):
         return self.position
 
     @pos.setter
-    def pos(self, value):
+    def pos(self, value: np.ndarray):
         # just here for compatibility solara_viz.
-        pass
+        self.position = value
 
     def __init__(self, space: ContinuousSpace, model):
         """Initialize a continuous space agent.
--- a/tests/experimental/test_continuous_space.py
+++ b/tests/experimental/test_continuous_space.py
@@ -106,11 +106,25 @@ def test_continuous_agent():
     agent.position = [1.1, 1.1]
     assert np.allclose(agent.position, [0.1, 0.1])
 
+    agent = ContinuousSpaceAgent(space, model)
+    agent.pos = [1.1, 1.1]
+    assert np.allclose(agent.position, [0.1, 0.1])
+    assert np.allclose(agent.pos, [0.1, 0.1])
+
     dimensions = np.asarray([[0, 1], [0, 1]])
     space = ContinuousSpace(dimensions, torus=False, random=model.random)
+
+    agent = ContinuousSpaceAgent(space, model)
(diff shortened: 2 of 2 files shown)


souro26:
This change looks solid enough. I had a few improvements in mind though,

1. Since position returns a direct reference into `space.agent_positions[...]`, in-place edits like `agent.pos[0] = x` bypass validation. This already applies to position, but it would be good to document that validated updates should go through the setter rather than in-place mutation.

2. The setter is annotated as `np.ndarray`, but tests use lists. The type hint should either be widened to an array-like type or the value should be normalized via `np.asarray(value)` before validation for consistency.

quaquel:
Please update this in line with me feedback in the issue. Just delete agent.pos. It should be fully removed.

quaquel:
I am closing this PR as stale.

PR #3412, opened 2026-03-02
Add and refine type annotations in mesa.time.event

Tushar1733:
This PR improves type annotations in the `mesa.time.event` module.

Changes include:
- Adding explicit return types for `__getstate__`, `__setstate__`, and other methods
- Refining `Callable` annotations to `Callable[..., Any]`
- Ensuring consistency across `Event` and `EventGenerator`

No functional changes were made.

### DIFF
--- a/mesa/time/events.py
+++ b/mesa/time/events.py
@@ -28,8 +28,7 @@ from enum import IntEnum
 from heapq import heapify, heappop, heappush, nsmallest
 from types import MethodType
 from typing import TYPE_CHECKING, Any
-from weakref import WeakMethod, ref
-from weakref import ReferenceType
+from weakref import ReferenceType, WeakMethod, ref
 
 if TYPE_CHECKING:
     from mesa import Model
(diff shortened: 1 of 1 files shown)


mesa/time/events.py
```diff
@@ -93,7 +95,7 @@ def CANCELED(self) -> bool:  # noqa: D102
     def __init__(
         self,
         time: int | float,
-        function: Callable,
+        function: Callable[..., Any],
```
quaquel, line 98: can you explain how to read this typing information.

mesa/agentset.py
quaquel, line 52: Please remove the changes to agentset.py. They are not part of this PR given your description.

mesa/time/events.py
Tushar1733: Callable[..., Any] means the function can take any arguments and return any type. Since event callbacks may have different signatures, this keeps the typing flexible while being more explicit than a plain Callable.

mesa/agentset.py
Tushar1733: Thanks for the feedback.
I noticed that the PR was created from main, which included unintended changes (e.g., in agentset.py).
I’ll create a clean branch with only the event module changes and open a new PR shortly.

PR #3415, opened 2026-03-02
fix: FixedCell setter atomically adds agent before updating reference (fixes #3411)

### LINKED ISSUE #3411: [BUG] FixedAgent keeps a reference to a cell that may reject it
**Describe the bug**

When a `FixedAgent` is placed into a cell that is already at capacity, the operation partially completes and leaves the agent in a broken state. The agent's internal reference gets set to the cell before the cell has actually accepted it. If the cell rejects the agent (because it's full), the reference is never cleaned up so agent permanently believes it occupies a cell that has no record of it.

**Expected behavior**

If placing an agent into a cell fails, nothing should change. The agent's cell reference should remain `None`, exactly as it was before the attempt.

**To Reproduce**

```python
from mesa import Model
from mesa.discrete_space import OrthogonalMooreGrid, FixedAgent

model = Model()
grid = OrthogonalMooreGrid((2, 2), random=model.random)
cell = list(grid.all_cells)[0]
cell.capacity = 1

a1 = FixedAgent(model)
a1.cell = cell  # works fine, cell is now full

a2 = FixedAgent(model)
try:
    a2.cell = cell  # raises — cell is full
except Exception:
    pass

print(a2.cell is cell)    # True  — agent thinks it was placed
print(a2 in cell.agents)  # False — cell has no record of it
```

The agent and cell are now permanently out of sync. Any logic that relies on `agent.cell` will silently misbehave for this agent.

**Additional context**

I can work on a PR to fix this if required

Discovered while building a simulation using `FixedAgent` and `OrthogonalMooreGrid` with per-cell capacity limits.

Rishav23av:
## Problem
FixedAgent kept a dangling reference to a cell that rejected it due to capacity.
After a failed cell.add_agent(), agent.cell was set but the cell had no record of the agent.

## Fix
Swap the order in FixedCell.cell.setter — call cell.add_agent(self) FIRST, 
then update self._mesa_cell only if it succeeds. This is the same atomic 
pattern already used in HasCell (PR #3374).

Fixes #3411

### DIFF
--- a/mesa/discrete_space/cell_agent.py
+++ b/mesa/discrete_space/cell_agent.py
@@ -84,6 +84,7 @@ class FixedCell(HasCell):
 
     @property
     def cell(self) -> Cell | None:
+        """The cell the agent is fixed to."""
         return self._mesa_cell
 
     @cell.setter


mesa/discrete_space/cell_agent.py
```diff
@@ -75,19 +75,30 @@ def move_relative(self: HasCellProtocol, direction: tuple[int, ...]):
...
         return self._mesa_cell
 
     @cell.setter
-    def cell(self, cell: Cell) -> None:
-        if self.cell is not None:
-            raise ValueError("Cannot move agent in FixedCell")
-        self._mesa_cell = cell
-
+    def cell(self, cell: Cell | None) -> None:
+        if self._mesa_cell is not None and cell is not self._mesa_cell:
```
quaquel: This still allows reassigning the same cell. I would just block any assignment after self._mesa_cell is no longer None.

Rishav23av: simplified it  just blocks any assignment once _mesa_cell is set

quaquel (approved):

quaquel:
There is an indentation error. Can you fix that? As a general rule, I suggest running all tests locally before committing changes.

Rishav23av:
Sorry about that Fixed the indentation and ran tests locally before pushing this time

Rishav23av:
@quaquel all done

PR #3419, opened 2026-03-03
Improve type annotations in mesa.time.event

Tushar1733:
This PR refines type annotations in the `mesa.time.event` module.

Changes include:
- Adding explicit return types for __getstate__ ,  __setstate__ and other methods
- Refining `Callable` annotations to `Callable[..., Any]`
- Ensuring consistency across `Event` and `EventGenerator`

No functional changes were made.

This replaces #3412 with a clean branch containing only the intended changes.

### DIFF
--- a/mesa/time/events.py
+++ b/mesa/time/events.py
@@ -35,7 +35,7 @@ if TYPE_CHECKING:
 
 
 def _create_callable_reference(
-    function: Callable[..., Any],
+    function: Callable[..., None],
 ) -> ReferenceType[Any] | WeakMethod:
     """Validate and create a weak-reference wrapper for an event callback."""
     if not callable(function):
@@ -95,7 +95,7 @@ class Event:
     def __init__(
         self,
         time: int | float,
-        function: Callable[..., Any],
+        function: Callable[..., None],
         priority: Priority = Priority.DEFAULT,
         function_args: list[Any] | None = None,
         function_kwargs: dict[str, Any] | None = None,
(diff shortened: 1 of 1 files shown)


mesa/time/events.py
```diff
@@ -400,7 +402,8 @@ def peek_ahead(self, n: int = 1) -> list[Event]:
             raise IndexError("event list is empty")
 
         # Filter out canceled events and get n smallest in correct chronological order
-        return nsmallest(n, (e for e in self._events if not e.CANCELED))
+        valid_events = [e for e in self._events if not e.CANCELED]
+        return nsmallest(n, valid_events)
```
souro26: This change reintroduces the intermediate list in `peek_ahead`. In #3413 we switched to a generator expression to avoid the extra allocation and improve performance. We should keep that version here.

Tushar1733: Thanks for pointing that out — you're right.
I’ll restore the generator expression to preserve the optimization from #3413.

quaquel:
> This replaces https://github.com/mesa/mesa/pull/3412 with a clean branch containing only the intended changes.

We typically use squash and merge, so there is no need to create a clean PR.

mesa/time/events.py
```diff
@@ -93,7 +95,7 @@ def CANCELED(self) -> bool:  # noqa: D102
     def __init__(
         self,
         time: int | float,
-        function: Callable,
+        function: Callable[..., Any],
```
quaquel: The return of the callable is ignored, so there in principle should not be a return from the callable.

Tushar1733: okay, since the return value is ignored, using Callable[..., None] is more appropriate. I've updated the annotation accordingly.

Tushar1733:
> > This replaces #3412 with a clean branch containing only the intended changes.
> 
> We typically use squash and merge, so there is no need to create a clean PR.

Understood, thanks for clarifying.

Issue #3420, opened 2026-03-03
Improve EventGenerator lifecycle clarity and control


souro26:
Right now EventGenerator keeps track of its state internally (`_active`, `_current_event`, execution count), but that state isn’t clearly exposed or separated in the public API.

**Current Limitations**

1. There’s no clean way to see when the next execution will happen.
2. The only lifecycle controls are start() and stop(), which makes temporary suspension awkward.

In more complex models, this makes the generator harder to reason about and sometimes pushes users to inspect private attributes.

**Proposed direction**

1. Add simple introspection properties like `next_scheduled_time`.
2. Based on that, add explicit lifecycle control such as `pause()` and `resume()`.

The goal is to make the generator’s state easier to understand and work with.

PR #3423, opened 2026-03-03
add next_scheduled_time property to EventGenerator

souro26:
### Summary
This PR adds a property `next_scheduled_time` to EventGenerator. It exposes the time of the next scheduled execution, or None if no event is currently scheduled.

Part of the event generator improvements in #3420.

### Motive
Currently there is no simple way to inspect when an EventGenerator will execute next without accessing internal attributes such as `_current_event`. For debugging, monitoring, or higher-level scheduling logic it is useful to know when the next execution is scheduled. This change exposes that information through a small explicit public API.

### Implementation
A new read-only property `next_scheduled_time` was added to EventGenerator. It returns the time of the currently scheduled event and None when no event is scheduled. It relies on the existing internal `_current_event` state.

### Usage Examples
````python
gen = EventGenerator(model, fn, Schedule(interval=2.0))
gen.start()
print(gen.next_scheduled_time)  # for example 2.0
gen.stop()
print(gen.next_scheduled_time)  # None
````

### DIFF
--- a/mesa/time/events.py
+++ b/mesa/time/events.py
@@ -248,7 +248,7 @@ class EventGenerator:
     def execution_count(self) -> int:
         """Return the number of times this generator has executed."""
         return self._execution_count
-    
+
     @property
     def next_scheduled_time(self) -> float | None:
         """Return the time of the next scheduled execution, or None if not scheduled."""
--- a/tests/time/test_events.py
+++ b/tests/time/test_events.py
@@ -584,7 +584,7 @@ class TestEventGenerator:
 
         model.run_for(1.5)
         assert order == ["H", "L"]
-        
+
     def test_introspection_properties(self, setup):
         """Test next_scheduled_time property."""
         model, fn = setup

### CODE REVIEW

File: mesa/time/events.py [hunk 1]
```diff
@@ -251,6 +249,18 @@ def execution_count(self) -> int:
...
 
+    @property
+    def next_time(self) -> float | None:
+        """Return the time of the next scheduled execution, or None if not scheduled."""
+        if self._current_event is None:
+            return None
+        return self._current_event.time
+
+    @property
+    def is_scheduled(self) -> bool:
```
quaquel: is this not redundant with `is_active`?

mesa/time/events.py
```diff
@@ -251,6 +249,18 @@ def execution_count(self) -> int:
         """Return the number of times this generator has executed."""
         return self._execution_count
 
+    @property
+    def next_time(self) -> float | None:
```
```suggestion
    def next_scheduled_time(self) -> float | None:
```

mesa/time/events.py
souro26: you're right it slipped my mind, they are redundant. i'll remove `is_scheduled`

quaquel:
1. Please update your tests. They are failing because you still test for `is_scheduled`
2. Update the pr description so it reflects the actual thing done in this PR.

souro26:
thanks for the review, i've updated the PR title, description and tests.

EwoutH:
@quaquel feel free to (also) tag me as a reviewer on time / event PRs.

PR #3424, opened 2026-03-03
Add explicit `RUN_ENDED` signal for terminal data handling in `DataRecorder`

codebreaker32:
This PR fixes the data collection/recorder architecture by ensuring the final state of a model is always captured at the end of a run, and introduces an "overwrite" mechanism to prevent duplicate data rows when the model is paused, modified, and resumed at the same timestep.

Previously, users had to manually call `finalise()` or `collect()` to get the last timestep if it didn't perfectly align with the interval.

### **Key Changes:**

1. **Signals (`model.py`, `signal_types.py`)**:
- Added a new `ModelSignals.RUN_ENDED` signal.
- Decorated `run_for` and `run_until` to automatically emit this signal when they finish executing.

2. **Base Recorder (`basedatarecorder.py`)**:
- Subscribed the recorder to the `RUN_ENDED` signal to automatically capture the final snapshot (`_on_run_ended`), removing the need for the manual `finalise()` method.
- Added `_last_collection` tracking to detect if the current timestep is identical to the previous one, passing an `is_overwrite` flag downstream.

### DIFF
--- a/tests/experimental/test_datarecorder.py
+++ b/tests/experimental/test_datarecorder.py
@@ -1087,3 +1087,12 @@ def test_run_ended(tmp_path, recorder_class):
     assert 4.0 in times
     assert df.loc[df["time"] == 4.0, "model_val"].iloc[0] == 0
     assert len(df) == 4
+
+    # Check for disabled dataset
+    model = MockModel()
+    recorder = DataRecorder(
+        model, config={"model_data": DatasetConfig(interval=2, enabled=False)}
+    )
+    model.run_for(3.0)
+    df = recorder.get_table_dataframe("model_data")
+    assert df.empty


codebreaker32:
Hi @quaquel 

Do you prefer this approach or should we wait for `RunConfiguration` idea to crystallise?

mesa/examples/basic/boltzmann_wealth_model/model.py
```diff
@@ -95,5 +95,31 @@ def gini(self):
...
+            agent.wealth += 10
+
+    model.schedule_event(trade, at=5.0)
+    model.run_until(10)
+
+    df2 = model.recorder.get_table_dataframe("model_data")
+    print(f"\nFinal Dataframe:\n {df2.to_string()}")
+    print(
+        f"\nUpdated Gini at time 5.0 after second run: {df2.loc[df2['time'] == 5.0, 'gini'].values[0]:.6f}\n"
+    )
```
quaquel: I assume this is for testing purposes?

codebreaker32: Yes. Reverted now.

quaquel:
Can you check the code coverage and if needed at tests? I'll merge later today or more likely tomorrow.

codebreaker32:
While writing tests, I found an interesting edge case in parquet backend, In old logic where we only pop from the in-memory buffer:

```python
if is_overwrite:
    while buffer and buffer[-1]["time"] == time:
        buffer.pop()

```

If a file flush happens right before the overwrite (e.g., calling `get_table_dataframe`), the buffer is empty. The `while` loop skips, and it blindly appends the new row, missing the fact that the old row is already saved in the physical file and causing duplicate data.

It might be propagated or reflected in user's own custom backend as well. I am convinced that it is a delicate approach. Do you think these type of edge cases can be handled by run_configs easily?

quaquel:
> If a file flush happens right before the overwrite (e.g., calling get_table_dataframe), the buffer is empty. The while loop skips, and it blindly appends the new row, missing the fact that the old row is already saved in the physical file and causing duplicate data.

> It might be propagated or reflected in user's own custom backend as well. I am convinced that it is a delicate approach. Do you think these type of edge cases can be handled by run_configs easily?

That's an interesting edge case. In my current conception of `run_config`, we would not need the overwrite as used in this PR, so it would avoid the edge case. We would not need it because run_ended is only emitted once, and it would leave the model in a state where any subsequent `run_x` calls would fail.

On the edge case for the current approach, what about never flushing the last item in the buffer?

codebreaker32:
> On the edge case for the current approach, what about never flushing the last item in the buffer?

Agree. However the trade-off for this approach, is that users would be strictly required to fetch their data through get_table_dataframe() to see the terminal state. If they bypass the recorder API and load the raw .parquet file directly from the disk, that final timestep will be missing

quaquel:
> Agree. However the trade-off for this approach, is that users would be strictly required to fetch their data through get_table_dataframe() to see the terminal state. If they bypass the recorder API and load the raw .parquet file directly from the disk, that final timestep will be missing

Fair point. I am unfortunately not familiar with parquet, but it seems the real fix will require some check of the file and see if something needs to be overwritten. The other option is to at least document this for parquet.

codebreaker32:
I have already implemented following approach: if the in-memory buffer is empty, it falls back to opening the physical .parquet file, filters out the old timestep (df = df[df["time"] != time]), and rewrites the file before appending the new data.
Are you comfortable with this approach of explicitly manipulating the .parquet file for this edge case, or would you prefer we stick to the simpler buffer logic and just document the quirk?

quaquel:
As indicated, I have little knowledge regarding parquet, so I'll trust your judgment.

quaquel (approved):

PR #3434, opened 2026-03-04
Fix incorrect warnings and exception types in `model.py` and `datacollection.py`

codebyNJ:
## Summary
- Remove duplicate `self.time: float = 0.0` assignment in `Model.__init__`
- Fix inverted warning messages in `Model._advance_time()` and `Model.run_until()` — the condition checks `until <= self.time` but the message said "is larger than" (should say "is not larger than")
- Fix `DataCollector.get_model_vars_dataframe()` and `get_agent_vars_dataframe()` — they used `raise UserWarning(...)` which crashes the caller instead of issuing a warning. Changed to `warnings.warn()` with a graceful `return pd.DataFrame()`
- Replace bare `raise Exception(...)` with `raise ValueError(...)` in `add_table_row()` and `get_table_dataframe()` for better exception specificity

### DIFF
--- a/mesa/exceptions.py
+++ b/mesa/exceptions.py
@@ -67,7 +67,7 @@ class ConnectionMissingException(SpaceException):
         )
 
 
-class TableMissingException(MesaException, ValueError):
+class TableMissingException(MesaException, ValueError):  # noqa: N818
     """Raised when attempting to access a table that does not exist in the DataCollector."""
 
     def __init__(self, table_name):


mesa/datacollection.py
```diff
@@ -407,27 +407,30 @@ def add_table_row(self, table_name, row, ignore_missing=False):
                             if False, throw an error if any columns are missing
         """
         if table_name not in self.tables:
-            raise Exception("Table does not exist.")
+            raise ValueError(f"Table '{table_name}' does not exist.")
```
quaquel: I am wondering whether we should not use a custom exception here instead?

mesa/datacollection.py
```diff
@@ -492,5 +498,5 @@ def get_table_dataframe(self, table_name):
             table_name: The name of the table to convert.
         """
         if table_name not in self.tables:
-            raise Exception("No such table.")
+            raise ValueError(f"Table '{table_name}' does not exist.")
```
quaquel: here the same custom exception might be used.

codebyNJ:
> One suggestion, but looks fine otherwise.

I have updated the PR with custom exceptions.

mesa/exceptions.py
```diff
@@ -67,6 +67,19 @@ def __init__(self, cell, other):
         )
 
 
+class TableMissingException(MesaException):
```
quaquel:
```suggestion
class TableMissingException(MesaException, ValueException):
```

PR #3438, opened 2026-03-05
Fix checkout of untrusted code in benchmark.yml

jackiekazil:
This was managed here for security purposes:  https://github.com/mesa/mesa/security/advisories/GHSA-3j55-5q6x-2h48

Already approved.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,7 +1,7 @@
 name: Performance benchmarks
 
 on:
-  pull_request_target:
+  pull_request:
     types: [opened, ready_for_review, labeled]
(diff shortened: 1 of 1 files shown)


EwoutH:
@m-y-mo did you fully validate this solution kept existing functionality intact? Apparently our benchmarks broke, could you look into it?

abhinavk0220:
Thanks for the quick fix! The two-workflow split approach looks correct.

EwoutH:
It might look correct, but it doesn’t work.

m-y-mo:
@EwoutH Please check and see if this works: https://github.com/mesa/mesa/pull/3464

PR #3446, opened 2026-03-05
replace global functools.cache with per-cell neighborhood caches

falloficaruss:
> [!NOTE]

### Pre-PR Checklist
- [ ] This PR is a bug fix, not a new feature or enhancement.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ ci:
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.14.14
+  rev: v0.15.4
   hooks:
     # Run the linter.
     - id: ruff-check
@@ -32,7 +32,7 @@ repos:
 
 # Prose linting with Vale
 - repo: https://github.com/errata-ai/vale
-  rev: v3.13.0
+  rev: v3.13.1
   hooks:
     - id: vale
       name: Vale
--- a/benchmarks/configurations.py
+++ b/benchmarks/configurations.py
@@ -7,6 +7,11 @@ from mesa.examples import (
     SugarscapeG1mt,
     WolfSheep,
 )
+from mesa.examples.advanced.sugarscape_g1mt.model import SugarScapeScenario
+from mesa.examples.advanced.wolf_sheep.model import WolfSheepScenario
+from mesa.examples.basic.boid_flockers.model import BoidsScenario
+from mesa.examples.basic.boltzmann_wealth_model.model import BoltzmannScenario
+from mesa.examples.basic.schelling.model import SchellingScenario
 
 configurations = {
     # BoltzmannWealth Model Configurations
@@ -15,6 +20,7 @@ configurations = {
             "seeds": 50,
             "replications": 5,
             "steps": 125,
+            "scenario_class": BoltzmannScenario,
             "parameters": {
                 "n": 100,
                 "width": 10,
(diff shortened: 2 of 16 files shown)


falloficaruss:
@quaquel Not sure why benchmark results could not be commented. In my local testing performance is roughly in the same area (in fact better is some cases). Can you take a look?

quaquel:
There was a security fix to the benchmarking, but it seems it has broken something. @jackiekazil, @EwoutH any idea what is going on? I might have time tomorrow to take a look as well.

quaquel:
here are the results:

| Model           | Size   | Init time [95% CI]         | Run time [95% CI]        | 
 |-----------------|--------|----------------------------|--------------------------|
| BoltzmannWealth | small  | 🟢 -16.4% [-16.9%, -16.0%] | 🔵 +0.6% [+0.4%, +0.8%]  |
| BoltzmannWealth | large  | 🟢 -25.0% [-27.3%, -22.4%] | 🔴 +8.2% [+5.5%, +10.9%] |
| Schelling       | small  | 🟢 -25.9% [-26.1%, -25.7%] | 🔵 -2.9% [-3.0%, -2.7%]  |
| Schelling       | large  | 🟢 -18.0% [-18.7%, -17.3%] | 🔵 -0.6% [-1.3%, +0.2%]  |
| WolfSheep       | small  | 🟢 -16.0% [-16.4%, -15.7%] | 🔵 +0.5% [+0.3%, +0.7%]  |
| WolfSheep       | large  | 🟢 -18.9% [-20.0%, -17.9%] | 🔵 -3.1% [-3.9%, -2.2%]  |
| SugarscapeG1mt  | small  | 🟢 -34.1% [-34.3%, -33.9%] | 🟢 -3.7% [-3.9%, -3.5%]  |
| SugarscapeG1mt  | large  | 🟢 -41.0% [-41.8%, -40.3%] | 🔵 -0.9% [-1.2%, -0.6%]  |
| BoidFlockers    | small  | 🔵 -2.3% [-2.7%, -1.9%]    | 🔵 -0.2% [-0.3%, -0.1%]  |
| BoidFlockers    | large  | 🔵 -1.4% [-2.1%, -0.8%]    | 🔵 +0.3% [+0.1%, +0.5%]  |

falloficaruss:
@quaquel what are your thoughts about the performance?

quaquel:
Disregarding performance for now. Startup times are measured in nanoseconds, so they don't matter much in the grand scheme of things. 

I am still not convinced by the need for this PR. If you remove one connection somewhere, any higher-order neighborhoods are also invalidated. I never developed that part fully, so dynamic grids are not actively supported by Mesa at present. Since higher-order neighborhoods are invalidated, it makes sense to clear the cache fully across the entire space. 

I suggest we first explore, through pathfinding or discussion, the idea of dynamic grids, and then reason backward to what this might imply for the caching strategy.

falloficaruss:
Thanks, that makes sense. I agree we should discuss dynamic grids first and only then decide cache behavior.

To make this concrete, can we align on two points:

- After runtime topology changes, do we want radius>1 neighborhood queries to be guaranteed correct immediately?
- If not, do we want to explicitly say radius>1 is best-effort (or unsupported) for dynamic updates?

jackiekazil:
> There was a security fix to the benchmarking... 
@quaquel I catching up on various other things. I believe this has been resolved by ewouth. If this is another issue that is still pending, please let me know.

quaquel:
> To make this concrete, can we align on two points:
>
> After runtime topology changes, do we want radius>1 neighborhood queries to be guaranteed correct immediately?
> If not, do we want to explicitly say radius>1 is best-effort (or unsupported) for dynamic updates?

Let's open a discussion on support for dynamic grids. Put foward your thoughts and then we can align there first. I am closing this in the meantime.

PR #3450, opened 2026-03-06
ENH: Accept Scenario as class or instance in Model init

### LINKED ISSUE #3435: Accept Scenario as class or instance in Model init
### Problem
The current pattern for wiring Scenarios into Models creates boilerplate in every model:

```python
class PdGrid(mesa.Model):
    def __init__(self, scenario=None):
        if scenario is None:
            scenario = PrisonersDilemmaScenario()
```

The call site is verbose for the default case:

```python
model = PdGrid(scenario=PrisonersDilemmaScenario(rng=42))
```

### Proposal
Allow `scenario` to accept either a **class** (instantiated automatically with defaults) or an **instance** (used directly). The class serves as the default in the signature:

```python
class PdGrid(mesa.Model):
    def __init__(self, scenario=PrisonersDilemmaScenario, rng=None):
        super().__init__(scenario=scenario, rng=rng)
            (self.scenario.width, self.scenario.height), ...
        )
```

### Call sites

```python
# Defaults
model = PdGrid()
model = PdGrid(rng=42)

# Custom parameters
model = PdGrid(scenario=PrisonersDilemmaScenario(width=100, rng=42))

# Batch runs
scenarios = [PrisonersDilemmaScenario(width=w, rng=r) for w in [50, 100] for r in seeds]
models = [PdGrid(scenario=s) for s in scenarios]
```

### Implementation

The change is in `Model.__init__`:

```python
if isinstance(scenario, type) and issubclass(scenario, Scenario):
    scenario = scenario(rng=rng)
elif scenario is None:
    scenario = Scenario(rng=rng)
# else: already an instance, use directly
```

### Benefits

- No `if scenario is None` boilerplate in every model
- The default argument documents which Scenario type the model expects
- IDE autocomplete and `help()` work (no `**kwargs` passing)
- Simple models that don’t need a custom Scenario just omit the parameter entirely
- SolaraViz can inspect the default to discover parameter names for sliders
- `batch_run` can inspect the default to construct Scenarios from parameter dicts

Nithurshen:
### Pre-PR Checklist
- [X] I linked an issue/discussion where a maintainer approved this enhancement/feature for implementation.

### Approval Link

Closes #3435

### Summary

This PR updates the `Model` class initialization to accept a `Scenario` class directly, automatically instantiating it with the appropriate default or provided `rng`. It maintains full backwards compatibility by continuing to support pre-instantiated `Scenario` objects or `None`.

### Motive

The current pattern for wiring Scenarios into Models creates unnecessary boilerplate code in every model subclass (e.g., manually checking `if scenario is None:`). By allowing a `Scenario` class to be passed directly, developers can use the class as a default argument in their model's signature. This improves code readability, enables proper IDE autocomplete, and allows tools like SolaraViz or `batch_run` to easily inspect default parameters.

### Implementation

1) `mesa/model.py`: 
  * Updated the `Model.__init__` type hints to accept a Scenario class (`scenario: S | type[S] | None = None`).
  * Modified the initialization logic to check if `scenario` is a class (`isinstance(scenario, type) and issubclass(scenario, Scenario)`). If so, it instantiates the scenario using the resolved `rng`. 
  * It continues to safely handle pre-existing `Scenario` instances and `None` fallbacks.
2) `tests/test_model.py`: Added a new test, `test_model_scenario_initialization`, which verifies the behavior when passing `None`, a pre-configured `Scenario` instance, and a custom `Scenario` class.

### Usage Examples

Before:
```python
class PdGrid(mesa.Model):
    def __init__(self, scenario=None):
        if scenario is None:
            scenario = PrisonersDilemmaScenario()
```

After:
```python
class PdGrid(mesa.Model):
    def __init__(self, scenario=PrisonersDilemmaScenario, rng=None):
        super().__init__(scenario=scenario, rng=rng)
        
model = PdGrid()
model = PdGrid(rng=42)
model = PdGrid(scenario=PrisonersDilemmaScenario(width=100, rng=42))
```

### Additional Notes

All existing tests and the newly added `test_model_scenario_initialization` pass locally.

### DIFF
--- a/.github/workflows/benchmarks-comment.yml
+++ b/.github/workflows/benchmarks-comment.yml
@@ -1,0 +1,37 @@
+name: Comment benchmark results
+
+on:
+  workflow_run:
+    workflows: ["Performance benchmarks"]
+    types:
+      - completed
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  comment:
+    if: github.event.workflow_run.conclusion == 'success'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download benchmark results
+        uses: actions/download-artifact@v8
+        with:
+          name: benchmark-results
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Comment PR
+        uses: actions/github-script@v8
+        with:
+          script: |
+            const fs = require('fs');
+            const prNumber = parseInt(fs.readFileSync('pr_number.txt', 'utf8').trim());
+            const output = fs.readFileSync('benchmark_output.txt', 'utf8');
+            await github.rest.issues.createComment({
+              issue_number: prNumber,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: 'Performance benchmarks:\n\n' + output
+            });
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -9,19 +9,17 @@ on:
       - '**.py'
       - '.github/workflows/benchmarks.yml'
 
+permissions:
+  contents: read
+
 jobs:
   run-benchmarks:
-    permissions:
-      contents: read
     if: >
       github.event.action == 'labeled' && contains(github.event.pull_request.labels.*.name, 'trigger-benchmarks') ||
       github.event.action == 'opened' ||
       github.event.action == 'ready_for_review'
     runs-on: ubuntu-latest
-    outputs:
-      output: ${{steps.compare_timings.outputs.timing_comparison}}
     steps:
-      # Python and dependency setup
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
@@ -32,7 +30,8 @@ jobs:
         run: pip install uv
       - name: Install dependencies
         run: uv pip install --system numpy pandas tqdm tabulate matplotlib solara networkx scipy
-      # Benchmarks on the mesa main branch
+
+      # Benchmarks on main
       - name: Checkout main branch
         uses: actions/checkout@v6
         with:
@@ -43,7 +42,8 @@ jobs:
       - name: Run benchmarks on main branch
         working-directory: benchmarks
         run: python global_benchmark.py
-      # Upload benchmarks, checkout PR branch, download benchmarks
+
+      # Upload main benchmarks, then checkout PR
       - name: Upload benchmark results
         uses: actions/upload-artifact@v7
         with:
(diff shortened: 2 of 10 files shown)


mesa/model.py
```diff
@@ -133,20 +133,18 @@ def __init__(
         # Strong references to active EventGenerators (prevent GC)
         self._event_generators: set[EventGenerator] = set()
 
-        # check if `scenario` is provided
-        # and if so, whether rng is the same or not
-        if scenario is not None:
-            if rng is not None and (scenario.rng != rng):
+        # Handle scenario if it is a class, instance, or None
+        if isinstance(scenario, type) and issubclass(scenario, Scenario):
```
quaquel, line 137: Did we not agree to raise probably a TypeError if scenario is not a subclass?

mesa/model.py
```diff
@@ -133,20 +133,18 @@ def __init__(
...
+            scenario = scenario(rng=rng)
+        elif scenario is None:
+            scenario = Scenario(rng=rng)  # type: ignore[assignment]
+        else:
+            # It's an instance, check if rng matches
+            if rng is not None and scenario.rng != rng:
                 raise ValueError("rng and scenario.rng must be the same")
-            else:
-                rng = scenario.rng
+            rng = scenario.rng
```
EwoutH, line 145: @quaquel how would you like to handle this long term? Always put rng in Scenario, keep it as something separate, something else?

tests/test_model.py
EwoutH, line 1: I think I would put the tests in `tests/experimental/test_scenarios.py`

mesa/model.py
quaquel: Always put it in Scenario.

mesa/model.py
EwoutH: At some point [you said](https://github.com/mesa/mesa/pull/3103#issuecomment-3732475778): 

> I don't want to force users to subclass Scenario to make any of this work.

It that still your current view?

quaquel: Both can be true: use the default Scenario class for just the rng and pass the rest as keyword arguments. But yes, it's a fair question. Looking at the open PRs on simplifying the examples, I like the cleaner code with scenario subclasses.

mesa/model.py
EwoutH: So we can deprecate/remove inputting `rng` directly? Only allow `rng` to be passed as part of a Scenario?

quaquel: yes, but keep that out of this PR. I hope to write up some stuff on replacing batchrunner soon and this will be included in that.

EwoutH:
Thanks for the effort.

@quaquel indicated that he wanted to give a stab to solve this more integrally, including resolving `rng` handling. If that's still his intention, I would prefer waiting how that would develop.

quaquel:
closing this in favor of #3493

PR #3459, opened 2026-03-06
Fix benchmark workflow permissions for fork PRs

EwoutH:
### Summary
Fixes the 403 "Resource not accessible by integration" error when the benchmark workflow tries to comment on PRs from forks.

### Bug / Issue
After switching the benchmark workflow from `pull_request_target` to `pull_request`, the comment step fails with a 403 for fork PRs. The `pull_request` event restricts the `GITHUB_TOKEN` to read-only for forks, regardless of declared `permissions`. Splitting into two jobs within the same workflow doesn't help (both jobs share the same restricted token.)

### Implementation
Split the single workflow into two files:

1. **`benchmarks.yml`:** Runs on `pull_request` with read-only permissions. Executes the benchmarks on both main and the PR branch, then uploads the comparison output and PR number as artifacts.
2. **`benchmarks-comment.yml`:** Triggers on `workflow_run` completion. Since `workflow_run` always runs in the base repo context, it has write permissions. Downloads the artifacts and posts the comment.

This follows [GitHub's recommended pattern](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run) for workflows that need write access on fork PRs.

As a side benefit, the base64 encoding workaround for passing the comparison output is no longer needed — results are passed as plain text through artifacts.

### Testing
- Verified on a fork PR that the benchmark job completes and uploads artifacts: https://github.com/EwoutH/mesa/pull/26
- Verified that the `workflow_run` job picks up the artifacts and posts the comment successfully.

### Additional Notes
- The `workflow_run` job only runs when the benchmark workflow succeeds (`conclusion == 'success'`).
- If we ever want to move back to a single-workflow approach, `pull_request_target` would work but requires careful handling of the checkout step to avoid running untrusted code with elevated permissions.

### DIFF
--- a/.github/workflows/benchmarks-comment.yml
+++ b/.github/workflows/benchmarks-comment.yml
@@ -1,0 +1,37 @@
+name: Comment benchmark results
+
+on:
+  workflow_run:
+    workflows: ["Performance benchmarks"]
+    types:
+      - completed
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  comment:
+    if: github.event.workflow_run.conclusion == 'success'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download benchmark results
+        uses: actions/download-artifact@v8
+        with:
+          name: benchmark-results
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Comment PR
+        uses: actions/github-script@v8
+        with:
+          script: |
+            const fs = require('fs');
+            const prNumber = parseInt(fs.readFileSync('pr_number.txt', 'utf8').trim());
+            const output = fs.readFileSync('benchmark_output.txt', 'utf8');
+            await github.rest.issues.createComment({
+              issue_number: prNumber,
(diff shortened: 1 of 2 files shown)


quaquel:
I triggered the benchmarks but no result is uploaded. Not sure if this is the correct way to test this?

EwoutH:
No it will only work after it's merged. I tested on my local fork:
- https://github.com/EwoutH/mesa/pull/26

This is a GitHub limitation by design: the `workflow_run` workflow is always loaded from the target branch (in our case `main`), not from the PR branch. This is a security measure to prevent anyone from modifying CI configuration in a PR to gain elevated permissions, which is exactly the kind of exploit that the `pull_request_target` to `pull_request` + `workflow_run` migration is meant to address.

That's also why @jackiekazil's https://github.com/mesa/mesa/pull/3438 seemed to work: It was still using the correct (working) config from `main`.

jackiekazil:
Thank you @EwoutH, @quaquel

EwoutH:
Validated: https://github.com/mesa/mesa/pull/3461#issuecomment-4012309003

PR #3461, opened 2026-03-06
Add Actions: event-driven timed agent behavior (v2)

EwoutH:
### Summary
Adds a minimal, subclassable `Action` system to `mesa.experimental.actions`, enabling agents to perform actions that take time, can be interrupted with progress tracking, and can be resumed.

As discussed in #3304, follows-up initial attempt #3308.

### Motive
Mesa 3.5 introduced event scheduling (`schedule_event`, `run_for`, etc.), but agents still lack a built-in concept of *doing something over time*. There's no way to know if an agent is busy, interrupt what it's doing, get partial credit for incomplete work, or resume an interrupted task. This has been discussed extensively in #2526, #2529, #2538, #2858, and #3304.

This PR introduces the minimal foundation those proposals build on.

### Implementation
**`Action` class** (`mesa/experimental/actions/actions.py`):

- Subclassable with `on_start()`, `on_resume()`, `on_complete()`, `on_interrupt(progress)` hooks. All default to `pass` (or `on_start()` for `on_resume`), so users only override what they need
- `name` property defaults to the class name (e.g. `Forage`), can be passed via `__init__`, set per-instance, or overridden in subclasses
- Duration and priority accept callables for state-dependent values, resolved at start time
- Interrupted actions remember progress and schedule only remaining duration on resume
- `ActionState` enum: PENDING → ACTIVE → COMPLETED / INTERRUPTED
- `interrupt()` respects the `interruptible` flag and returns bool; `cancel()` always succeeds
- Live-computed query properties: `progress`, `remaining_time`, `elapsed_time`, `is_resumable`

**Agent integration** (additions to `mesa/agent.py`):
- `current_action`, `start_action()`, `interrupt_for()`, `cancel_action()`, `is_busy` on base Agent
- Inert when unused — zero cost for agents that don't use actions
- `Agent.remove()` silently cancels any scheduled action event (see _Agent removal_ for rationale)

#### Creating actions

```python
from mesa.experimental.actions import Action, ActionState

# Subclass — override hooks for reusable actions
class Forage(Action):
    def __init__(self, sheep):
        super().__init__(sheep, duration=5.0)

    def on_start(self):
        print(f"Sheep {self.agent.unique_id} starts foraging")

    def on_resume(self):
        print(f"Sheep {self.agent.unique_id} resumes foraging")

    def on_complete(self):
        self.agent.energy += 30

    def on_interrupt(self, progress):
        self.agent.energy += 30 * progress

forage = Forage(sheep)

# Minimal subclass — only override what you need
class Rest(Action):
    def on_complete(self):
        self.agent.energy = 100
```

#### Name

```python
# Defaults to class name
Forage(sheep).name    # "Forage"
Action(agent).name    # "Action"

# Pass via __init__
Action(agent, duration=3.0, name="dig").name  # "dig"

# Override per-instance
action = Action(agent, duration=3.0)
action.name = "dig"
action.name           # "dig"

# Override in subclass
class Dig(Action):
    @property
    def name(self):
        return f"dig({self.agent.unique_id})"
```

#### Duration and priority

```python
# Fixed values
Action(agent, duration=5.0)
Action(agent, duration=5.0, priority=10.0)

# Callable — resolved once when start() is called
Action(agent, duration=lambda a: 10.0 / a.speed)
Action(agent, priority=lambda a: a.threat_level * 2)

# Non-interruptible (cannot be interrupted, only cancelled)
Action(agent, duration=3.0, interruptible=False)
```

#### Starting and running actions

```python
# Start an action — schedules a completion event
agent.start_action(action)

# Check if the agent is doing something
agent.is_busy              # True
agent.current_action       # the Action instance
agent.current_action.state # ActionState.ACTIVE

# Let time pass — action completes when duration elapses
model.run_for(5)           # on_complete() fires, agent.is_busy becomes False
```

#### Interruption

```python
# Interrupt current action and start a new one
# Returns True if successful, False if current action is non-interruptible
success = agent.interrupt_for(new_action)

# Cancel current action (always succeeds, ignores interruptible flag)
# Useful for forced stops. Returns False only if agent is idle.
agent.cancel_action()

# Both call on_interrupt(progress) on the interrupted action.
# progress is the raw time fraction: 0.0 to 1.0
```

#### Querying action state

```python
# All properties are live-computed while active — no stale values
action.state          # ActionState.PENDING / ACTIVE / COMPLETED / INTERRUPTED
action.progress       # 0.0 to 1.0, live time fraction completed
action.remaining_time # duration * (1 - progress), live
action.elapsed_time   # duration * progress, live
action.is_resumable   # True if INTERRUPTED and progress < 1.0
action.name           # class name by default, e.g. "Forage"
```

#### Resuming interrupted actions

```python
# An interrupted action remembers its progress
agent.start_action(forage)
model.run_for(3)           # 60% of a 5.0 duration action
agent.cancel_action()      # on_interrupt(0.6) fires

forage.progress            # 0.6
forage.remaining_time      # 2.0
forage.is_resumable        # True

# Resume — schedules completion for remaining duration only
agent.start_action(forage) # on_resume() fires (NOT on_start)
model.run_for(2)           # completes, on_complete() fires
```

#### on_start vs on_resume

```python
class BuildWall(Action):
    def on_start(self):
        self.agent.reserve_materials()  # Only on first start
        print("Starting construction")

    def on_resume(self):
        print("Resuming construction")  # Skip material reservation

    def on_complete(self):
        self.agent.wall_built = True

# If on_resume is NOT overridden, it calls on_start by default.
# Override on_resume when first-start and resumption need different behavior.
```

#### Agent removal

```python
# remove() silently cancels the scheduled event without firing on_interrupt.
# This is safe: no callbacks touch agent state during teardown.
agent.remove()

# If your action needs cleanup (release resources, dequeue, etc.),
# call cancel_action() before remove():
agent.cancel_action()  # Fires on_interrupt — cleanup runs
agent.remove()
```

#### Sheep: forage, flee, resume

```python
class Forage(Action):
    def __init__(self, sheep):
        super().__init__(sheep, duration=5.0)

    def on_complete(self):
        self.agent.energy += 30

    def on_interrupt(self, progress):
        self.agent.energy += 30 * progress


class Flee(Action):
    def __init__(self, sheep):
        super().__init__(sheep, duration=2.0, interruptible=False)

    def on_complete(self):
        pass  # Survived

    def on_interrupt(self, progress):
        self.agent.alive = False  # Didn't escape


class Sheep(Agent):
    def __init__(self, model):
        super().__init__(model)
        self.energy = 50.0
        self.alive = True
        self._forage_action = Forage(self)
        self.start_action(self._forage_action)

    def spot_predator(self):
        self.interrupt_for(Flee(self))

    def resume_foraging(self):
        if self._forage_action.is_resumable:
            self.start_action(self._forage_action)
        else:
            self._forage_action = Forage(self)
            self.start_action(self._forage_action)
```

#### Call center: worker handles calls, takes breaks

```python
class HandleCall(Action):
    def __init__(self, employee, call):
        duration = employee.model.rng.exponential(8.0)
        super().__init__(employee, duration=duration)
        self.call = call

    def on_complete(self):
        self.agent.calls_handled += 1
        self.call.resolve()

    def on_interrupt(self, progress):
        self.call.requeue()  # Put caller back in line


class TakeBreak(Action):
    def __init__(self, employee):
        super().__init__(employee, duration=15.0, interruptible=False)

    def on_complete(self):
        self.agent.energy = 100
```

#### Quick minimal actions

```python
# Minimal subclass for simple effects
class Move(Action):
    def on_complete(self):
        self.agent.move()

agent.start_action(Move(agent, duration=1.0))

# State-dependent duration via callable
class Travel(Action):
    def __init__(self, agent):
        super().__init__(agent, duration=lambda a: a.distance_to_target / a.speed)

    def on_complete(self):
        self.agent.arrive()
```

### Design choices
**Subclassable with hooks, no inline callbacks.** An Action is a proper class you subclass and override `on_start()`, `on_resume()`, `on_complete()`, and `on_interrupt(progress)`. This follows quaquel's recommendation from the review — it keeps actions clean, self-contained, and avoids having two ways to do the same thing. Even simple actions are just a few lines as a subclass.

```python
class Forage(Action):
    def __init__(self, sheep):
        super().__init__(sheep, duration=5.0)

    def on_complete(self):
        self.agent.energy += 30

    def on_interrupt(self, progress):
        self.agent.energy += 30 * progress
```

**All hooks default to `pass`.** Users only need to override the specific hooks they care about. A minimal action only needs `on_complete`. This keeps the boilerplate to a minimum while maintaining the single-approach design.

**Separate `on_start` and `on_resume`.** First start and resumption are semantically different — `on_start` might reserve resources or initialize state that shouldn't happen twice. `on_resume` defaults to calling `on_start`, so users who don't care get the same behavior. Override `on_resume` when the distinction matters.

**Separate `on_complete` and `on_interrupt` instead of a single `on_effect`.** The earlier draft (#3308) used one callback with a completion float. That conflates *why* the action ended with *how much was done* — you can't reliably tell "completed with weird reward curve" from "interrupted at 70%." Separate hooks mean the user always knows why their code was called.

**No reward curves in the base class.** Reward curves (mapping time progress to effective completion) encode a specific assumption about how partial work translates to partial value. That's domain logic. The base class gives you raw `progress` (time fraction, 0–1) and you decide what to do with it. Proportional reward, all-or-nothing, diminishing returns — it's a few lines in your `on_interrupt`, not a framework concept.

**Live-computed progress.** `progress`, `remaining_time`, and `elapsed_time` are properties that compute live from the model clock while the action is active. No stale values — you can query progress mid-execution for visualization, conditional logic, or other agents inspecting what this agent is doing.

**Resume support through re-startable actions.** An interrupted action remembers its progress and can be passed back to `start_action()` to continue from where it left off — only the remaining duration is scheduled. This covers the common "interrupt for something urgent, then go back to what you were doing" pattern without needing an action queue.

**Callable duration and priority.** Both accept `callable(agent) -> float`, resolved once at first start (not re-evaluated on resume). A wolf's chase duration depends on distance; a flee priority depends on threat proximity. This is common enough to support from day one rather than forcing users to pre-compute values.

**`name` passable via `__init__`.** Following quaquel's feedback, the name can be set directly in the constructor (`Action(agent, name="dig")`), via instance assignment, or by overriding the property in subclasses. Defaults to the class name.

**On base Agent, not a subclass.** `current_action`, `start_action()`, `interrupt_for()`, `cancel_action()`, and `is_busy` live directly on Agent. They're inert when unused (`current_action = None`). This avoids a migration when the API stabilizes and keeps the cost at one attribute initialized to None.

**Silent cleanup on Agent.remove().** When an agent is removed, its action's scheduled event is cancelled but `on_interrupt` does *not* fire. The agent is being destroyed, not making a behavioral decision — callbacks touching agent state during teardown can cause errors. Models that need action cleanup (releasing resources, dequeuing) can call `cancel_action()` before `remove()` to opt in.

### Additional Notes
This is deliberately minimal. Not included (future work):

- Action queue / chaining (`action.then(next_action)`)
- Preconditions and available action repertoire
- `evaluate()` / `select_action()` behavioral loop
- Integration with ActionSpace constraints (#2858)
- Integration with ContinuousStates (#2529)

See #3304 for the full roadmap. This PR provides the foundation those features build on.

### DIFF
--- a/mesa/experimental/actions/actions.py
+++ b/mesa/experimental/actions/actions.py
@@ -5,11 +5,10 @@ Mesa's event scheduling system for precise timing and supports interruption
 with progress tracking and optional resumption.
 
 Actions are subclassable: override on_start(), on_complete(), and
-on_interrupt() to define behavior. For simple cases, pass callables directly.
+on_interrupt() to define behavior.
 
 Example::
 
-    # Subclass approach (complex actions)
     class Forage(Action):
         def __init__(self, sheep):
             super().__init__(sheep, duration=5.0)
@@ -21,10 +20,6 @@ Example::
             self.agent.energy += 30 * progress
 
     sheep.start_action(Forage(sheep))
-
-    # Inline approach (simple actions)
-    action = Action(agent, duration=2.0, on_complete=lambda: agent.do_thing())
-    agent.start_action(action)
 """
 
 from __future__ import annotations
@@ -55,7 +50,7 @@ class Action:
     completion timing.
 
     Subclass and override on_start/on_complete/on_interrupt for complex
-    behavior, or pass callables for simple cases.
+    behavior. All hooks default to doing nothing (pass).
 
     Attributes:
         agent: The agent performing this action.
@@ -81,12 +76,9 @@ class Action:
         agent: Agent,
         duration: float | Callable[[Agent], float] = 1.0,
         *,
+        name: str | None = None,
         priority: float | Callable[[Agent], float] = 0.0,
         interruptible: bool = True,
-        on_start: Callable[[], None] | None = None,
-        on_complete: Callable[[], None] | None = None,
-        on_interrupt: Callable[[float], None] | None = None,
-        on_resume: Callable[[], None] | None = None,
     ) -> None:
         """Initialize an Action.
 
@@ -95,24 +87,16 @@ class Action:
             duration: Time to complete. Either a float or a callable
                 that receives the agent and returns a float. Resolved
                 when start() is called.
+            name: Human-readable name. Defaults to the class name.
             priority: Importance level for interruption decisions. Either
                 a float or a callable that receives the agent and returns
                 a float. Resolved when start() is called.
             interruptible: If False, interrupt() will fail and return False.
-            on_start: Optional callback, called on first start.
-                Ignored if the subclass overrides on_start().
-            on_complete: Optional callback, called when the action finishes.
-                Ignored if the subclass overrides on_complete().
-            on_interrupt: Optional callback(progress), called when interrupted.
-                Ignored if the subclass overrides on_interrupt().
-            on_resume: Optional callback, called when resuming after
-                interruption. Ignored if the subclass overrides on_resume().
-                Defaults to calling on_start if not provided.
         """
         self.agent = agent
         self.model = agent.model
         self.interruptible = interruptible
-        self._name: str | None = None
+        self._name: str | None = name
 
         # Store raw values (may be callables, resolved at start)
         self._duration_spec = duration
@@ -121,12 +105,6 @@ class Action:
         # Resolved values (set in start())
         self.duration: float = 0.0
         self.priority: float = 0.0
-
-        # Optional inline callbacks
-        self._on_start_fn = on_start
-        self._on_complete_fn = on_complete
-        self._on_interrupt_fn = on_interrupt
-        self._on_resume_fn = on_resume
 
         # Lifecycle state
         self.state: ActionState = ActionState.PENDING
@@ -140,8 +118,8 @@ class Action:
     def name(self) -> str:
         """Human-readable name. Returns the class name by default.
 
-        Can be set per-instance (``action.name = "dig"``) or
-        overridden in subclasses.
+        Can be set via __init__(name=...), per-instance assignment
+        (``action.name = "dig"``), or overridden in subclasses.
         """
         return self._name if self._name is not None else self.__class__.__name__
 
@@ -192,8 +170,6 @@ class Action:
         Override for setup logic (e.g., logging, animation triggers,
         resource reservation). Not called on resume — see on_resume().
         """
-        if self._on_start_fn is not None:
-            self._on_start_fn()
 
     def on_resume(self) -> None:
         """Called when the action resumes after interruption.
@@ -204,10 +180,7 @@ class Action:
 
         Default implementation calls on_start().
         """
-        if self._on_resume_fn is not None:
-            self._on_resume_fn()
-        else:
-            self.on_start()
+        self.on_start()
 
     def on_complete(self) -> None:
         """Called when the action finishes normally.
@@ -215,8 +188,6 @@ class Action:
         Override to apply the action's full effect (e.g., gaining
         energy, completing a transaction).
         """
-        if self._on_complete_fn is not None:
-            self._on_complete_fn()
 
     def on_interrupt(self, progress: float) -> None:
         """Called when the action is interrupted before completion.
@@ -234,8 +205,6 @@ class Action:
             a non-interruptible action that receives on_interrupt was
             necessarily cancelled, not interrupted.
         """
-        if self._on_interrupt_fn is not None:
-            self._on_interrupt_fn(progress)
 
     # --- Execution (called by Agent, not typically by users) ---
 
--- a/tests/experimental/test_actions.py
+++ b/tests/experimental/test_actions.py
@@ -66,6 +66,12 @@ class TestName:
         action.name = "custom_name"
         assert action.name == "custom_name"
 
+    def test_name_via_init(self):
+        """Name can be passed via __init__."""
+        _model, agent = make_model_and_agent()
+        action = Action(agent, name="my_action")
+        assert action.name == "my_action"
+
 
 # --- Basic lifecycle ---
 
@@ -237,34 +243,49 @@ class TestStartResume:
     def test_default_on_resume_calls_on_start(self):
         """When on_resume is not overridden, it delegates to on_start."""
         model, agent = make_model_and_agent()
-        started = []
 
-        action = Action(
-            agent,
-            duration=10.0,
-            on_start=lambda: started.append("start"),
-        )
+        class StartTracker(Action):
+            def __init__(self, agent):
+                super().__init__(agent, duration=10.0)
+                self.started = []
+
+            def on_start(self):
+                self.started.append("start")
+
+            def on_interrupt(self, progress):
+                pass
+
+        action = StartTracker(agent)
 
         agent.start_action(action)
-        assert started == ["start"]
+        assert action.started == ["start"]
 
         model.run_for(3)
         agent.cancel_action()
 
         agent.start_action(action)
         # Default on_resume calls on_start
-        assert started == ["start", "start"]
+        assert action.started == ["start", "start"]
 
-    def test_inline_on_resume_overrides_default(self):
+    def test_overridden_on_resume_prevents_on_start(self):
+        """When on_resume is overridden, on_start is not called on resume."""
         model, agent = make_model_and_agent()
-        log = []
 
-        action = Action(
-            agent,
-            duration=10.0,
-            on_start=lambda: log.append("start"),
-            on_resume=lambda: log.append("resume"),
-        )
+        class ResumeTracker(Action):
+            def __init__(self, agent):
+                super().__init__(agent, duration=10.0)
+                self.log = []
+
+            def on_start(self):
+                self.log.append("start")
+
+            def on_resume(self):
+                self.log.append("resume")
+
+            def on_interrupt(self, progress):
+                pass
+
+        action = ResumeTracker(agent)
 
         agent.start_action(action)
         model.run_for(3)
@@ -272,7 +293,7 @@ class TestStartResume:
 
         agent.start_action(action)
 
-        assert log == ["start", "resume"]
+        assert action.log == ["start", "resume"]
 
     def test_multiple_resume_cycles(self):
         model, agent = make_model_and_agent()
@@ -591,53 +612,59 @@ class TestCallableDurationPriority:
         assert call_count == 1  # Not called again
 
 
-# --- Inline callbacks ---
+# --- Subclass callbacks ---
 
 
-class TestInlineCallbacks:
-    def test_inline_on_complete(self):
+class TestSubclassCallbacks:
+    def test_subclass_on_complete(self):
         model, agent = make_model_and_agent()
         agent.energy = 50
 
-        action = Action(
-            agent,
-            duration=3.0,
-            on_complete=lambda: setattr(agent, "energy", agent.energy + 30),
-        )
+        class GainEnergy(Action):
+            def on_complete(self):
+                self.agent.energy += 30
+
+        action = GainEnergy(agent, duration=3.0)
 
         agent.start_action(action)
         model.run_for(3)
 
         assert agent.energy == 80
 
-    def test_inline_on_interrupt(self):
+    def test_subclass_on_interrupt(self):
         model, agent = make_model_and_agent()
-        received_progress = []
 
-        action = Action(
-            agent,
-            duration=10.0,
-            on_interrupt=lambda p: received_progress.append(p),
-        )
+        class ProgressTracker(Action):
+            def __init__(self, agent):
+                super().__init__(agent, duration=10.0)
+                self.received_progress = []
+
+            def on_interrupt(self, progress):
+                self.received_progress.append(progress)
+
+        action = ProgressTracker(agent)
 
         agent.start_action(action)
         model.run_for(2)
         agent.cancel_action()
 
-        assert received_progress == [pytest.approx(0.2)]
+        assert action.received_progress == [pytest.approx(0.2)]
 
-    def test_inline_on_start(self):
+    def test_subclass_on_start(self):
         _model, agent = make_model_and_agent()
-        started = []
 
-        action = Action(
-            agent,
-            duration=5.0,
-            on_start=lambda: started.append(True),
-        )
+        class StartTracker(Action):
+            def __init__(self, agent):
+                super().__init__(agent, duration=5.0)
+                self.started = False
+
+            def on_start(self):
+                self.started = True
+
+        action = StartTracker(agent)
 
         agent.start_action(action)
-        assert started == [True]
+        assert action.started is True
 
 
 # --- Agent removal ---
@@ -676,16 +703,19 @@ class TestResumeDetection:
 
     def test_on_start_can_detect_resume(self):
         model, agent = make_model_and_agent()
-        start_types = []
 
         class DetectResumeAction(Action):
+            def __init__(self, agent):
+                super().__init__(agent, duration=10.0)
+                self.start_types = []
+
             def on_start(self):
-                start_types.append("resume" if self.progress > 0 else "first")
+                self.start_types.append("resume" if self.progress > 0 else "first")
 
             def on_interrupt(self, progress):
                 pass
 
-        action = DetectResumeAction(agent, duration=10.0)
+        action = DetectResumeAction(agent)
 
         agent.start_action(action)
         model.run_for(3)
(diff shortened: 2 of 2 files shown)


quaquel:
Conceptually, I like the look of this. Will try to dig into the specifics as soon as possible.

EwoutH:
Thanks.

One other thing I was considering is giving `Action` dataclass-like properties (a bit similar we do with `Scenario` and `Schedule`).

Then instead of:
```Python
class Forage(Action):
    def __init__(self, sheep):
        super().__init__(sheep, duration=5.0)

    def on_complete(self):
        self.agent.energy += 30
```
You can just do:
```Python
class Forage(Action):
    duration = 5.0

    def on_complete(self):
        self.agent.energy += 30
```

EwoutH:
@mesa/maintainers I would love some reviews. Especially on conceptual and API level. Are we building the right thing, is it useful for a broad set of users, is it a good building block to build future components on, do we expose the right hooks (both user API and to Mesa itself), etc.

mesa/experimental/actions/actions.py
```diff
@@ -0,0 +1,390 @@
...
+        self,
+        agent: Agent,
+        duration: float | Callable[[Agent], float] = 1.0,
+        *,
+        priority: float | Callable[[Agent], float] = 0.0,
+        interruptible: bool = True,
+        on_start: Callable[[], None] | None = None,
+        on_complete: Callable[[], None] | None = None,
+        on_interrupt: Callable[[float], None] | None = None,
+        on_resume: Callable[[], None] | None = None,
```
quaquel: I am not sure about this. I prefer subclassing over passing a bunch of functions here. Having a small, lightweight custom action subclass seems clearer to me.

mesa/experimental/actions/actions.py
```diff
@@ -0,0 +1,390 @@
...
+            on_interrupt: Optional callback(progress), called when interrupted.
+                Ignored if the subclass overrides on_interrupt().
+            on_resume: Optional callback, called when resuming after
+                interruption. Ignored if the subclass overrides on_resume().
+                Defaults to calling on_start if not provided.
+        """
+        self.agent = agent
+        self.model = agent.model
+        self.interruptible = interruptible
+        self._name: str | None = None
```
quaquel: why can't this be passed via the init?

mesa/experimental/actions/actions.py
```diff
@@ -0,0 +1,390 @@
...
+        return self._progress
+
+    @property
+    def remaining_time(self) -> float:
+        """Time remaining until completion.
+
+        Computed live while active. For interrupted actions, returns
+        the time that would remain if resumed.
+        """
+        return self.duration * (1.0 - self.progress)
```
quaquel, line 153: this seems a complicated route to calculate this. We know self.start_time, self.duration, and self.model.time. Why not calculate this directly? Or even use event.scheduled_time?

mesa/experimental/actions/actions.py
```diff
@@ -0,0 +1,390 @@
...
+
+    @property
+    def elapsed_time(self) -> float:
```
quaquel, line 156: this seems a complicated route to calculate this. We know self.start_time, self.duration, and self.model.time. Why not calculate this directly? Or even use event.scheduled_time?

mesa/agent.py
```diff
@@ -62,6 +63,7 @@ def __init__(self, model: M, *args, **kwargs) -> None:
 
         self.model: M = model
         self.unique_id = None
+        self.current_action: Action | None = None
```
quaquel, line 66: I am not sure I like editing the base class for this. With a lot of the previous experimental features, we just maintained a custom agent that had the extra behavior. I realize that with the various agent subclasses we are getting this becomes more complicated.

EwoutH:
@quaquel thanks for the review. Before I dive into the details, there are no overarching, high level, design or conceptual point we need to address at this stage?

mesa/agent.py
EwoutH: I see your point. I can provide a Mixin, if that’s preferred.

In this case however, we’re not breaking anything, just adding a single variable and some methods. So personally I think in this case it might be okay.

quaquel:
Not really, other than subclassing version passing functions as per one of my comments on the code.

mesa/experimental/actions/actions.py
EwoutH: In many cases you only need a `on_complete` and `duration`. In that case subclassing seems like overkill.

mesa/agent.py
quaquel: I am not sure. In my view, this reveals a wider issue. Object-oriented stuff is fine, but if you want to compose objects, it starts to break down. Multiple inheritance is hard to reason about, so less than ideal. But that is all we have in Python. 

Remember, a similar problem popped up when adding the scheduling stuff to Model. Again, ideally, you want to contain this in a separate object that just contains the parts for scheduling, but not the other parts. But then you need to either use multiple inheritance or use pass-through methods. Neither of which is ideal. 

Technically, the problem here cannot be solved with a pure mixin. A mixin only adds methods, not state. But here you need 1 state attribute: current_action as well. Still, I am inclined to favour this approach at least while this is in experimental.

mesa/experimental/actions/actions.py
quaquel: I understand that, but a small lightweight class seems clearer to me. Just look at the code below:

Compare

```python
my_action = Action(agent, duration=10, on_complete=lambda agent:agent.do_something, on_interupt=lamdba agent.handle_interuption)

class MyAction(Action):
    def on_complete(self):
        self.agent.do_something()

    def on_interupt(self):
        self.agent.do_something()

my_action = MyAction(agent, duration=10)
```

I define the custom action once, and then whenever I use it, I can just focus on the scheduling of the action
instead of all the other stuff. Also, the current implementation allows for both, so we have 2 ways of doing things, which is less than ideal. I would suggest using pass by default on all `on_x`  methods so a default action does nothing. And users only need to implement the specific methods that they need.

mesa/experimental/actions/actions.py
EwoutH: I couldn't find a simpler way to be honest. For now it works and passes tests.

mesa/experimental/actions/actions.py
EwoutH: idem dito

mesa/experimental/actions/actions.py
EwoutH: Good catch, now it can (fe4ac15).

mesa/experimental/actions/actions.py
EwoutH: Fair, and we can always add syntatic suger later. For now this pattern is good.

Implemented (fe4ac15).

mesa/agent.py
EwoutH: I agree, we need a better integral approach to this.

Let's split that discussion from this PR. For this PR, I think the current approach is fine. We're in 4.0 dev anyways, and had extensive discussions about this feature and API.

EwoutH:
From my side this PR is ready to merge.

mesa/experimental/actions/actions.py
quaquel:
```suggestion
        return self._event.time - self.model.time
```

quaquel:
```suggestion
        return self.time - self._start_time
```

how will this react if not yet started?

mesa/experimental/actions/actions.py
quaquel:
```suggestion
        return self.time - self._start_time
```

How will this react if not yet started?

mesa/experimental/actions/actions.py
EwoutH: We need fallbacks for when there's no event (not started, interrupted, completed).

We can do it like this, but I'm not sure that's better:

```suggestion
        if self.state is ActionState.ACTIVE and self._event is not None:
            return self._event.time - self.model.time
        # Not active (pending, interrupted, completed): use stored progress
        return self.duration * (1.0 - self._progress)
```

mesa/experimental/actions/actions.py
EwoutH: For interrupted actions we need the frozen progress, and before start `_start_time` is `-1`.

We could do it like this:
```Python
        if self.state is ActionState.ACTIVE and self._start_time >= 0:
            return self.duration * self._progress + (self.model.time - self._start_time)
        # Not active: use stored progress
        return self.duration * self._progress
```

EwoutH:
Maybe one thing: The name `Action` itself.

_Action_ was taken from the RL world, which is used to talk in terms of _Actions_, _Agents_, _Environments_, _Rewards_ and _States_. Lot's of our discussions were RL-inspired, so it makes sense some of the terminology comes from there. However, we might want to align it more with ABM or broader simulation.

In simulation literature, I think the closer terms are often _activity_ or _process_: DES is commonly described in terms of events, activities, and processes, and process-interaction frameworks like SimPy make interruption/preemption first-class.

Since this PR models something durative that can be started, interrupted, resumed, and completed, `Activity` may be the most precise term. `Process` also fits, but feels heavier. `Action` is still defensible, especially if we're considering we're also talking about thinks of `ActionSpaces` (which is a RL-concept, and makes more sense then `ActivitySpace` or `ProcessSpace` (also typing `ssS` is weird).

CC @mesa/maintainers 

<hr>

One other follow-up is that we might want to split by interruption cause / termination reason. There's a difference between:
- voluntary cancel (Agent decides to do something else)
- forced cancel (another Agent or the Environment decides that this action won't/can't be done anymore)

And maybe this is a continuous scale conceptually (what's really "forcing" someone? Aren't we all determinisic?).

Practical example:
- Sheep is eating grass
- Second sheep starts eating same grass
- First sheep has eaten the grass (which is now gone)
- What happens to the second sheep?

This case points to resource claiming, (forced) resource transfer, research denial, etc. See https://github.com/mesa/mesa/discussions/3304#discussioncomment-15836530.

EwoutH:
I think interruption cause can be added later in a backward-compatible way, so it doesn’t need to block this PR. The safest path would be to keep `on_interrupt(progress)` working as-is and add cause separately, e.g. via an `InterruptCause` enum stored on the action (`last_interrupt_cause`) or via an optional second argument with a compatibility shim. The parts that would be harder to change later are semantics like whether `cancel()` is resumable or whether `Agent.remove()` should fire interruption hooks.

quaquel:
I favour activity. I would not go with process, because we don't actively support process interaction modeling (a devs formalism).

I am not sure we need to add causes. The resource claiming example, I would approach quite differently and potentially even without using Actions/Activities.

EwoutH:
Merging as is. It’s experimental. Let’s figure stuff out as we go.

EwoutH:
The development of Actions and the discussions around it exposed many concepts that might (or might not) be useful for core Mesa.

A few future directions to explore, in no particular order:

1. **Interruption causes.** The current `on_interrupt(progress)` doesn't distinguish *why* the action was interrupted — voluntary switch, forced cancellation, resource disappearing, agent removal by another agent, etc. Adding an `InterruptCause` enum (or similar) would let hooks respond differently. This can be done backward-compatibly (optional second argument or attribute on the action), but the semantics of `cancel()` vs `interrupt()` vs agent removal should be pinned down before people start building on the current API.
2. **Action queues and chaining.** Right now agents can only have one action at a time. A natural next step is an action queue so agents can plan sequences: `forage → rest → forage`. This also enables `reschedule_on_interrupt` behavior (interrupted actions get re-queued with remaining duration). The queue design needs thought around re-validation — should queued actions re-check preconditions when dequeued?
3. **Resources.** As sketched in #3304, a `Resource` primitive (shared object with limited capacity and a waiting queue) would cover a huge class of models: factory machines, hospital beds, phone lines, charging stations. The natural fit is that Resources hold Actions, not Agents — an Action already carries the agent, duration, and hooks. This is probably the highest-value next addition for DES-style modeling.
4. **Behavioral repertoire and evaluation loop.** The bigger vision from #3304 involves agents having a set of `available_actions` with preconditions and a `select_action()` method that picks what to do next. This is where different behavioral frameworks (#2538) — BDI, needs-based, utility maximization, RL — would plug in as different selection strategies. Given @quaquel's (reasonable) skepticism about whether this can be made generic enough, it might be better to first collect 3-4 concrete use cases and see what patterns emerge before designing the abstraction.
5. **Continuous States (#2529).** Values that change smoothly over time (energy depleting, hunger increasing) with threshold-triggered re-evaluation are a natural complement to Actions. A `ContinuousState` crossing a threshold could trigger `evaluate()`, causing the agent to reconsider its current action. This builds straightforwardly on the event system and observables.
7. **ActionSpace (#2858).** A concept like an ActionSpace could define constraints on what's physically or practically possible. It would sit between precondition filtering and action selection, constraining or modifying available actions. Not needed for the core action system but a natural extension once the behavioral loop exists.

Remember Mesa is an ABM library focussing on giving modellers a framework with blocks they can use or not. We're trying to provide building blocks that help lots of modelers in a significant way.

But first, let's build some example models!

PR #3462, opened 2026-03-06
Add static dependency injection to `@computed_property` for `@emit` support

codebreaker32:
### Summary

This PR adds static dependency injection to the `@computed_property` decorator via a new `dependencies` keyword argument. This allows derived states to manually subscribe to specific signals, explicitly supporting events generated by the `@emit` decorator which are otherwise invisible to the reactive system.

### Motive

See #3227

### Implementation

* **Decorator Expansion:** Updated `@computed_property` to accept a `dependencies` kwarg, which takes a list of tuples containing the target dependencies name and the specific signal types (or pass nothing for `ALL` sentinel).
* **State Initialization:** Modified `ComputedState.__init__` to parse the `dependencies` list upon creation.
* **Lifecycle Wiring:** For each dependency provided, the `ComputedState` actively registers itself with the target using `owner.observe(obs_name, signal_type, self._set_dirty)`. This bypasses the need for `__get__` auto-discovery and directly injects the cache invalidation trigger into the emitter's internal `subscribers` dictionary.

### Usage Examples

```python
from mesa.experimental.mesa_signals import HasEmitters, computed_property, emit, ModelSignals

class MyModel(mesa.Model):
    def __init__(self):
        super().__init__()
        self.agent_count = 0

    @emit("model", ModelSignals.AGENT_ADDED)
    def add_agent(self):
        self.agent_count += 1

    # Statically bridging the built-in signal to the computed property
    @computed_property(dependencies=[("model", ModelSignals.AGENT_ADDED)])
    def population_status(self):
        return "Populated" if self.agent_count > 0 else "Empty"
```

### DIFF
--- a/mesa/experimental/mesa_signals/core.py
+++ b/mesa/experimental/mesa_signals/core.py
@@ -20,7 +20,7 @@ from __future__ import annotations
 import functools
 import weakref
 from collections import defaultdict, namedtuple
-from collections.abc import Callable, Generator, Iterable
+from collections.abc import Callable, Generator, Iterable, MutableSequence
 from typing import Any, ClassVar, Literal
 
 from mesa.experimental.mesa_signals.signals_util import (
@@ -66,11 +66,15 @@ class BaseObservable:
     def __get__(self, instance: HasEmitters, owner):  # noqa: D105
         value = getattr(instance, self.private_name)
 
-        # fixme this makes signaling list part of computed
         if CURRENT_COMPUTED is not None:
-            # there is a computed dependent on this Observable, so let's add
-            # this Observable as a parent
-            CURRENT_COMPUTED._add_parent(instance, self.public_name, value)
+            if isinstance(value, MutableSequence):
+                # Pass None for mutable sequences to prevent memory leaks and
+                # ensure in-place mutations bypass the `old != new` check.
+                CURRENT_COMPUTED._add_parent(instance, self.public_name, None)
+            else:
+                # there is a computed dependent on this Observable, so let's add
+                # this Observable as a parent
+                CURRENT_COMPUTED._add_parent(instance, self.public_name, value)
 
             # fixme, this can be done more cleanly
             #  problem here is that we cannot use self (i.e., the observable), we need to add the instance as well
--- a/tests/experimental/test_mesa_signals.py
+++ b/tests/experimental/test_mesa_signals.py
@@ -891,6 +891,40 @@ def test_ObservableList_slice_delitem():
     )
 
 
+def test_observable_list_mutation():
+    """Test that in-place mutations of ObservableList trigger computed property updates."""
+
+    class ListAgent(Agent, HasEmitters):
+        inventory = ObservableList()
+
+        def __init__(self, model):
+            super().__init__(model)
+            self.inventory = [1, 2]
+
+        @computed_property
+        def inventory_size(self):
+            return len(self.inventory)
+
+    model = Model(rng=42)
+    agent = ListAgent(model)
+
+    # Initial access builds the dependency graph and caches the value
+    assert agent.inventory_size == 2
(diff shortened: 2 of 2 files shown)


mesa/experimental/mesa_signals/core.py
```diff
@@ -193,64 +218,76 @@ class ComputedProperty(property):
     signal_types = ObservableSignals
 
 
-def computed_property(func: Callable) -> property:
+def computed_property(func: Callable | None = None, *, dependencies=None) -> property:
     """Decorator to create a computed property.
```
quaquel, line 242: please expand the docstring to describe the arguments

mesa/experimental/mesa_signals/core.py
```diff
@@ -193,64 +218,76 @@ class ComputedProperty(property):
     signal_types = ObservableSignals
 
 
-def computed_property(func: Callable) -> property:
+def computed_property(func: Callable | None = None, *, dependencies=None) -> property:
```
quaquel: this seems strange, should the `*` not come before the kwargs?

mesa/experimental/mesa_signals/core.py
```diff
@@ -193,64 +218,76 @@ class ComputedProperty(property):
...
 
-        if state.is_dirty:
-            changed = False
-
-            # Check if parents actually changed
-            if not state.parents:
-                changed = True
+            if not hasattr(self, key):
+                state = ComputedState(
+                    self, computation_func.__name__, computation_func, dependencies
```
quaquel:
```suggestion
                    self, computation_func.__name__, computation_func, dependencies=dependencies
```

mesa/experimental/mesa_signals/core.py
```diff
@@ -146,9 +146,20 @@ def __set__(self, instance: HasEmitters, value):  # noqa D103
...
+        "func",
+        "is_dirty",
+        "name",
+        "owner",
+        "parents",
+        "value",
+    ]
+
+    def __init__(
+        self, owner: HasEmitters, name: str, func: Callable, dependencies=None
```
quaquel: what about typing dependencies?

mesa/experimental/mesa_signals/core.py
codebreaker32: This prevents if an user accidentally pass a list of dependencies as the first positional argument (where the decorator expects the function). Specifically I read that this is necessary to separate the decorator and the decorator factory. However I ain't an expert in typing so I am open to any further refactor.

quaquel: I agree on the use of the `*`, but you have now changed func from an argument into a keyword argument as well. 

So, my question was not about the typing but the signature itself.

codebreaker32: I did this explicitly to support both ways i.e `@computed_property` and `@computed_property(dependencies=[..]`. If I won't make func a kwarg, this will happen
```plaintext
>       @computed_property(dependencies=[("test", "test_signal")])
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       TypeError: computed_property() missing 1 required positional argument: 'func'
```

I also mentioned it at the end of `decorator`

```python
   if func is None:
        # Called with arguments: @computed_property(dependencies=...)
        return decorator
    else:
        # Called without arguments: @computed_property
        return decorator(func)
```

mesa/experimental/mesa_signals/core.py
codebreaker32: Done

mesa/experimental/mesa_signals/core.py
codebreaker32: Done

PR #3467, opened 2026-03-07
fix: allow FixedAgent removal via public cell setter

hillhack:
## Fix
Closes: https://github.com/mesa/mesa/issues/3466

The cleanup logic was moved out of the `remove()` method and into the **cell property descriptor**, where it properly belongs.

Previously, `FixedAgent.remove()` had to bypass the public `cell` API and modify private state directly. This created fragile behavior and inconsistent removal logic.

With this fix:

- The property descriptor now correctly handles cleanup.
- `FixedAgent` no longer needs to manually modify private attributes.
- Clearing an agent’s cell works consistently through the public API.

Without this change, any attempt to manually clear a `FixedAgent`’s cell outside of the `remove()` method would still trigger `ValueError` crashes.

### DIFF
--- a/tests/discrete_space/test_fixed_agent_removal.py
+++ b/tests/discrete_space/test_fixed_agent_removal.py
@@ -1,6 +1,7 @@
 """Tests for FixedAgent removal functionality and property consistency."""
 
 import pytest
+
 from mesa import Model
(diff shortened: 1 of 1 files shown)


quaquel:
We have discussed this issue in detail. I remain unconvinced by the need for this so I am closing this PR.

PR #3470, opened 2026-03-07
perf: cache is_empty in Cell using empty property

Rishav23av:
Closes part of #3379

## What
`Cell.is_empty` was recomputing `len(self._agents) == 0` on every call.
The `empty` property already maintains this boolean as a cache (via
`_empty` slot for plain cells, or numpy property layer for grid cells).
This PR makes `is_empty` use that cache directly.

## Changes
- `__init__`: initialize `self.empty = True`
- `add_agent`: flip cache only when needed, after append
- `remove_agent`: update cache with `not bool(self._agents)` directly
- `is_empty`: return `self.empty` instead of recomputing `len()`

## Benchmark
5000 agents, 100x100 grid, 500 movement steps (from #3379):

| | Before | After |
|---|---|---|
| Total runtime | ~63s | ~7.8s |
| `is_empty` cumtime | ~11.1s | ~0.6s |

~88% reduction in `is_empty` time.

## Notes
- All 46 discrete_space tests pass
- Single focused change as suggested in #3388 review
- No API changes


Rishav23av:
CI is passing on all platforms. Benchmark results on Apple Silicon (M-series), Python 3.12, using the script from #3379 (5000 agents, 100x100 grid, 500 steps):

| Metric          | Before | After |
|-----------------|--------|-------|
| Total runtime   | ~63s   | ~7.8s |
| is_empty cumtime| ~11.1s | ~0.6s |

~88% reduction in `is_empty` time. Happy to run benchmarks on additional models (Schelling, BoltzmannWealth) if that would help validate no regressions.

mesa/discrete_space/cell.py
```diff
@@ -166,12 +166,12 @@ def remove_agent(self, agent: CellAgent) -> None:
...
             raise AgentMissingException(agent, self.coordinate) from e
 
-        self.empty = self.is_empty
+        self.empty = not bool(self._agents)
 
     @property
     def is_empty(self) -> bool:
         """Returns a bool of the contents of a cell."""
-        return len(self._agents) == 0
+        return self.empty
```
souro26, line 174: This makes _empty the source of truth for cell occupancy. Previously `is_empty` was derived directly from _agents, which guaranteed correctness regardless of how _agents changed. With this change the invariant `_empty == (len(_agents) == 0)` has to hold everywhere

mesa/discrete_space/cell.py
```diff
@@ -146,13 +147,12 @@ def add_agent(self, agent: CellAgent) -> None:
...
-        n = len(self._agents)
-        self.empty = False
-
-        if self.capacity is not None and n >= self.capacity:
+        if self.capacity is not None and len(self._agents) >= self.capacity:
             raise CellFullException(self.coordinate)
 
         self._agents.append(agent)
+        if self.empty:
+            self.empty = False
```
souro26, line 155: _empty is now acting like a cache for whether the cell has agents. That means it always needs to stay in sync with _agents. Since _agents is the real source of truth, any change to it that doesn't update _empty could make this incorrect.

mesa/discrete_space/cell.py
Rishav23av: I checked all usages of _agents in the codebase  it's only ever mutated in add_agent (via .append) and remove_agent (via .remove), both of which update the cache correctly. All other references are read-only. So the invariant empty == (len(_agents) == 0) is guaranteed by the existing structure

mesa/discrete_space/cell.py
Rishav23av: Agreed  _empty is acting as a cache Since _agents is only ever mutated through add_agent and remove_agent and both update empty correctly the cache stays in sync. Happy to add a docstring comment making this explicit if that would help reviewers

souro26:
Also len(list) in CPython is already O(1), so replacing `len(self._agents) == 0` with a cached boolean shouldn't by itself explain an 8x runtime difference. My guess is that the benchmark loop is calling `is_empty` extremely frequently, so even a small per-call reduction accumulates in that hot loop.

Since len() is already constant time, the change mainly trades a derived check for a cached state, which introduces the invariant `_empty == (len(_agents) == 0)` that now has to be maintained everywhere _agents is mutated.

Rishav23av:
You're right that len(list) is O(1) in CPython, so the 8x improvement isn't solely from this change. The baseline I compared against was from the issue description (~63s), which was filed against an older version of Mesa. The current main branch is likely already faster due to other improvements. I should have run a proper A/B comparison against current main without my change. I'll do that now and report back with accurate numbers.

Rishav23av:
@souro26 
 You were right to question this. I ran a proper A/B comparison against current main:

main (no fix): 2.780s
with fix: 3.021s

The fix is actually ~8% slower on current main. The 8x improvement I reported earlier was incorrectly comparing against the benchmark numbers from the original issue (#3379), which were measured against an older version of Mesa. I'm going to close this PR as the change doesn't provide a net benefit against the current codebase

Issue #3469, opened 2026-03-07
document the duplicate coordinate


Krishsharma179:
Summary
In the current system in discrete_space.py in add_cell there is a probability that the user might initialize the new cell at the coordinate where there is already a cell then the previous cell will be overwritten by the new cell silently

Implementation
Just document this explicitly

```
Warning:
            Coordinate Collision: If a cell already exists at the specified
            coordinates (e.g., cell1), it will be overwritten silently by the
            new cell (cell2).

            Ensure the target coordinates are vacant before calling this method.
        
```

Rishav23av:
@Krishsharma179  can i work on this

Krishsharma179:
@Rishav23av  Thanks !!! for the interest Rishav but i have the implementation and i will send a PR in some time

quaquel:
I am not sure whether this is a problem. It might however be documented explicitly. 

Why might it not be a problem?  Dynamically changing   Discrete spaces is conceptually possible but not actively supported by Mesa. So the default use is to instantiate a space with cells and then leave it as is afterwards. Thus, it is unlikely that this condition might be triggered.

If we want to do something with it, I am not sure what the right solution would be. Raising an exception is quite extreme. It forces a user to first remove the old cell before adding a new one. A warning, however, also seems strange, because it suggests that what someone is doing is somehow not correct.

Krishsharma179:
@quaquel Thanks for the thoughtful explanation! 

I understand that this case might have a lower chance of occurring, but it can still happen especially for users with less experience. 

I agree that raising an error can be extreme, but a **Warning** could be a good middle ground:
- If a person is doing it intentionally, they can ignore the warning.
- If it's a mistake, the warning alerts them before silent data loss occurs.

That said, I'm happy to proceed with **documentation** instead if you prefer that approach. Just want to ensure users are aware of this behavior to avoid debugging headaches later!

quaquel:
I suggest just documenting this explicitly for now.

DashamiJituri:
Hi, I’d like to work on documenting this behavior in the relevant method to make it clearer for users.

I plan to add a concise note in the docstring explaining that existing cells at the same coordinates will be overwritten.

Please let me know if this approach looks good.

Rishav23av:
@DashamiJituri PR #3473 addresses this issue. Please read the issue and if someone open a pr  it before commenting.
@Krishsharma179,@quaquel you can go ahead and close this issue

DashamiJituri:
[...]

PR #3473, opened 2026-03-07
document the duplicate coordinate 

Krishsharma179:
### Summary
In the current system in discrete_space.py in add_cell  there is a probability that the user might initialize the new cell at the coordinate where there is already a cell then the previous cell will be overwritten by the new cell silently
### Bug / Issue
#3469 This is the link to the related issue 

### Implementation
Just document this explicitly
```
Warning:
            Coordinate Collision: If a cell already exists at the specified
            coordinates (e.g., cell1), it will be overwritten silently by the
            new cell (cell2).

            Ensure the target coordinates are vacant before calling this method.
        
 ```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ ci:
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.14.14
+  rev: v0.15.4
   hooks:
     # Run the linter.
     - id: ruff-check
@@ -32,7 +32,7 @@ repos:
 
 # Prose linting with Vale
 - repo: https://github.com/errata-ai/vale
-  rev: v3.13.0
+  rev: v3.13.1
   hooks:
     - id: vale
       name: Vale
--- a/mesa/datacollection.py
+++ b/mesa/datacollection.py
@@ -38,6 +38,7 @@ from copy import deepcopy
 from functools import partial
 
 from mesa import Agent
+from mesa.exceptions import TableMissingException
 
 with contextlib.suppress(ImportError):
     import pandas as pd
@@ -407,7 +408,7 @@ class DataCollector:
                             if False, throw an error if any columns are missing
         """
         if table_name not in self.tables:
-            raise Exception("Table does not exist.")
+            raise TableMissingException(table_name)
 
         for column in self.tables[table_name]:
             if column in row:
@@ -415,7 +416,7 @@ class DataCollector:
             elif ignore_missing:
                 self.tables[table_name][column].append(None)
             else:
-                raise Exception("Could not insert row with missing column")
+                raise ValueError(f"Could not insert row with missing column '{column}'")
 
     def get_model_vars_dataframe(self):
         """Create a pandas DataFrame from the model variables.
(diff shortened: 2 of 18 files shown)

### CODE REVIEW

**Comment by Krishsharma179** [comment]
Hi @quaquel , please check this out when you get a chance!

**Comment by souro26** [comment]
I’m not sure this should be treated as a bug. Right now `add_cell` just assigns directly into _cells, which means adding a cell at an existing coordinate replaces the old one. That is an intentional dictionary style semantic rather than something accidental. 

Changing this to raise a ValueError would shift the behavior from replacement to a hard failure. That’s a significant behavioral change, especially since the current implementation doesn’t enforce any uniqueness checks and simply lets the new cell overwrite the old one. Also, modifying existing tests to enforce a new behavior is not the best way to handle this in my opinion. 

Given the [comment](https://github.com/mesa/mesa/issues/3469#issuecomment-4016768846) that dynamically modifying DiscreteSpace isn’t really a primary supported workflow, I think enforcing validation here is the wrong direction. It is better to either document the overwrite behavior explicitly or discuss whether replacement vs. validation should actually be the intended API before locking it in with an exception and tests.

**Comment by Krishsharma179** [comment]
@quaquel  please check

**Comment by Krishsharma179** [comment]
> I’m not sure this should be treated as a bug. Right now `add_cell` just assigns directly into _cells, which means adding a cell at an existing coordinate replaces the old one. That is an intentional dictionary style semantic rather than something accidental.
> 
> Changing this to raise a ValueError would shift the behavior from replacement to a hard failure. That’s a significant behavioral change, especially since the current implementation doesn’t enforce any uniqueness checks and simply lets the new cell overwrite the old one. Also, modifying existing tests to enforce a new behavior is not the best way to handle this in my opinion.
> 
> Given the [comment](https://github.com/mesa/mesa/issues/3469#issuecomment-4016768846) that dynamically modifying DiscreteSpace isn’t really a primary supported workflow, I think enforcing validation here is the wrong direction. It is better to either document the overwrite behavior explicitly or discuss whether replacement vs. validation should actually be the intended API before locking it in with an exception and tests.

Hey @souro26  thanks for the feedback I understood that raising an error will be too extreme for this so I  document this behavior explicitly

**Comment by quaquel** [comment]
Can you update the original description so it correctly reflects the actual implementation?

File: mesa/discrete_space/discrete_space.py [hunk 1]
```diff
@@ -105,6 +105,15 @@ def add_cell(self, cell: T):
             if you rely on getting neighborhoods of cells with a radius higher than 1, these might not be cleared
             correctly if you are adding or removing cells and connections at runtime.
 
+        Warning:
+            Coordinate Collision: If a cell already exists at the specified
+            coordinates (e.g., cell1), it will be overwritten silently by the
+            new cell (cell2).
+
+            - The original cell will vanish without any notification or error.
+            - All agents contained within the original cell will be permanently lost.
```
quaquel: not true, please remove

mesa/discrete_space/discrete_space.py
```diff
@@ -105,6 +105,15 @@ def add_cell(self, cell: T):
             if you rely on getting neighborhoods of cells with a radius higher than 1, these might not be cleared
             correctly if you are adding or removing cells and connections at runtime.
 
+        Warning:
+            Coordinate Collision: If a cell already exists at the specified
+            coordinates (e.g., cell1), it will be overwritten silently by the
+            new cell (cell2).
+
+            - The original cell will vanish without any notification or error.
```
quaquel: not true perse. It has just been removed from the grid, but might stick around as a ghost cell depending on what holds references to it.

Krishsharma179:
ok I will do it

Krishsharma179:
@quaquel Done! Updated the documentation to reflect the actual 
behavior. Thanks for the clarification!

souro26:
Please update the PR title and description

PR #3476, opened 2026-03-07
Visualization: add `viz_dims` projection for nD `ContinuousSpace` (Matplotlib + Altair)

DemetersSon83:
## Summary

This PR makes Mesa’s Matplotlib space drawing utilities (`mesa.visualization.mpl_space_drawing`) compatible with ND (e.g., 3D) `ContinuousSpace` by projecting agent positions and axis bounds using `space.viz_dims` (defaulting to `(0, 1)` for backward compatibility).

In practical terms: ND models can now be visualized in Matplotlib by choosing the 2D projection plane (XY / XZ / YZ, etc.) without crashing and without misleading axis limits.

## Motivation / Problem

Mesa’s ND `ContinuousSpace` introduces (or enables) agent positions with more than two coordinates. However, the current Matplotlib drawing path implicitly assumes a 2D position:

- It attempts to unpack `agent.position` into `(x, y)` (which fails for 3D/ND and prevents visualization entirely).
- It computes axis limits using 2D-only bounds (effectively “hard-coding” dims 0 and 1), which makes projection switching inconsistent or incorrect when rendering other planes.

As a result, ND models are significantly harder to debug/interpret—especially when the “interesting structure” exists outside the XY plane.

This PR fixes the Matplotlib path to behave consistently with the projection approach used elsewhere: select two dimensions from the ND position based on `viz_dims`.

## What changes

1. **Matplotlib projection of ND positions**  
   When collecting agent coordinates for plotting, we project agent positions onto the `(i, j)` indices given by `space.viz_dims` (fallback `(0, 1)`).

2. **Axis bounds match the selected projection**  
   When drawing the continuous space, we compute `x_min/x_max` and `y_min/y_max` from `space.dimensions[i]` and `space.dimensions[j]` rather than assuming dims 0/1.

3. **Regression tests**  
   Adds focused tests to ensure:
   - ND projection does not regress back to 2D-only assumptions,
   - axis limits reflect the selected `viz_dims`, and
   - default behavior remains unchanged for standard 2D spaces.

IMG_0981

## Why this is useful (evidence)

I’m including a small, reproducible Matplotlib report demonstrating both utility and correctness of the ND visualization/projection.

### 1) Utility: “projection choice changes interpretation”

The report uses a simple “layered airspace” toy model: agents move in XY but belong to two distinct altitude layers in Z. In XY projection, the layers overlap visually; in XZ/YZ projections, the separation is obvious (report, page 1).

A quantitative metric makes the same point: with a near-miss radius of **0.08**, the final step shows:

- **XY (2D projection): 599** pairs within radius  
- **XYZ (true 3D): 244** pairs within radius  

That’s about a **2.5× overestimate** if you only look at XY. This directly demonstrates why `viz_dims` is not just “nice to have” but necessary for correct interpretation/debugging in ND models (report, page 2; run summary).

### 2) Correctness sanity check: MSD scales with dimension

The report includes a mean-squared displacement experiment showing that MSD grows linearly with step count and scales with dimension (2D vs 3D), matching the expected theoretical trend (report, page 3).

### 3) Performance: projection overhead is negligible

A small microbenchmark of Matplotlib agent data collection shows comparable cost across 2D vs 3D projections (roughly ~0.01–0.016 seconds per call in the run shown), suggesting that indexing two dimensions out of an ND vector does not introduce meaningful overhead (report, page 4).

## Backward compatibility

- Existing 2D `ContinuousSpace` visualization remains unchanged because the default `viz_dims` is `(0, 1)`.
- The changes only affect ND continuous-space visualization and only in the Matplotlib drawing utilities.

## Request for review

I’d love feedback on:

- whether `viz_dims` is the right minimal interface for ND→2D projection in Matplotlib,
- whether the chosen behavior (projection + bounds) matches maintainers’ expectations,
- and whether you’d prefer the demo/report kept as an example, moved into docs, or omitted from the PR (keeping only tests + core fix).

## Report link
[nd_viz_report.pdf](https://github.com/user-attachments/files/25816481/nd_viz_report.pdf)

## Summary of changes

This PR adds first-class support for visualizing **n-dimensional (e.g., 3D)** `ContinuousSpace` in Mesa’s **visualization layer** by introducing a simple and consistent **2D projection mechanism** driven by `viz_dims=(i, j)`. Default behavior remains unchanged (`viz_dims=(0, 1)`), so 2D models continue to render exactly as before.


DemetersSon83:
This PR fixes a real incompatibility in the Matplotlib visualization utilities: they assume 2D agent positions and hard-code 2D bounds, so ND ContinuousSpace can’t be plotted (and even when it can, non-XY projections don’t get correct bounds).

The change is intentionally minimal: project positions and bounds using space.viz_dims (default (0,1)), matching the projection approach already used by renderers.

I attached a small Matplotlib report showing why it matters: in a simple 3D “layered airspace” model, the XY projection overestimates close-pairs by ~2.5× relative to true 3D at the same radius (599 vs 244 at r=0.08), and switching viz_dims reveals the correct structure immediately.

EwoutH:
Thanks for the PR! I agree we need support and a solution for this.

@Sahil-Chhoker and @quaquel this is now implemented and defined by the user in the continuous space. Is that the right place, or should it be handled in SolaraViz?

quaquel:
In my view, it should be handled fully at the visualization side. Probably in the space renderer and via some `setup_x` method.

DemetersSon83:
Good afternoon,

I'm glad that you think the PR will be useful. Please let me know if there are any updates/changes that you need me to make.

quaquel:
@DemetersSon83, The key issue, I guess, is to handle everything on the visualization side. So, can you update the PR in light of this feedback or do you need me to be more specific?

EwoutH:
@Sahil-Chhoker are you able to give a few pointers to @DemetersSon83 where this could/should be modified in SolaraViz?

DemetersSon83:
Hello. Yes some direction on how to do that would be helpful.

DemetersSon83:
Good morning! Just wanted to follow up on my inquiry above. It's not exactly clear how you want me to modify this PR. Some guidance would be greatly appreciated.

EwoutH:
The core point is that this is a welcome feature, but should not be addressed in `mesa/experimental/continuous_space` like it currently is, but in `mesa/visualization`.

quaquel:
To add, my hunch is that it would go into `ContinuousSpaceDrawer` and/or both the matplotlib backend class and the altair backend class.

DemetersSon83:
Ok, I think I've successfully updated this PR to meet your request. Please let me know if you need anything else!

DemetersSon83:
Good morning. This failed again in Codecov. Will that pose a significant issue to resolving this? I am unable to convert this PR back to a draft for some reason so I can update the `tests/test_model.py` or `tests/test_model.py` files to correct the issue.

quaquel:
I have changed it back to draft.

We do care about coverage. You can check codecov to see what code is covered and what is not.

EwoutH:
Maybe rebase your PR. It seems like the diff comes from changes unrelated to your PR.

quaquel:
Can you update the PR description to reflect the final implementation? Once that is done, this is ready to be merged.

DemetersSon83:
> Can you update the PR description to reflect the final implementation? Once that is done, this is ready to be merged.

Ok, done. Please let me know if the descriptive changes are sufficient.

DemetersSon83:
> > Can you update the PR description to reflect the final implementation? Once that is done, this is ready to be merged.
> 
> 
> 
> Ok, done. Please let me know if the descriptive changes are sufficient.

Hello! Is there anything else you need me to change before this can be merged?

PR #3477, opened 2026-03-07
Fix double stepping of MetaAgent component agents

tejutejas05:
Problem

When using MetaAgent, both the MetaAgent and its constituent agents are registered in `model.agents`.  
This causes the scheduler to step the component agents twice:

1. Once through the MetaAgent
2. Once directly through the AgentSet scheduler

 Solution

Skip agents marked with `is_component` inside `AgentSet.do()`.

This ensures component agents are only stepped through their parent MetaAgent, preventing duplicate execution.

 Changes

- Added a check in "AgentSet.do()" to skip agents with the "is_component" flag.

### DIFF
--- a/mesa/examples/advanced/alliance_formation/app.py
+++ b/mesa/examples/advanced/alliance_formation/app.py
@@ -3,7 +3,10 @@ import networkx as nx
 import solara
 from matplotlib.figure import Figure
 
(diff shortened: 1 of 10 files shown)


quaquel:
Changes to agentset are not acceptable for fixing this bug. MetaAgent is experimental so issues should be fixed there.

PR #3482, opened 2026-03-08
Add diffing engine for dependency autodiscovery in `mesa_signals/core.py`

codebreaker32:
### Summary

This PR addresses: 
>simplify and enhance performance of @computed_property autodiscovery of dependencies. Currently, on every call, it's fully rebuilt after first disposing of all parents. But subscribing is not cheap. So, it might be better to diff the dependencies and update only the changed ones. This also moves the subscription responsibility back into the @computable_property, which is conceptually cleaner

from #3227

### Motivation

Currently, when a `@computed_property` re-evaluates, it eagerly calls `unobserve` on every parent in its graph, executes the function, and then calls `observe` on every discovered dependency from scratch. For a stable graph where the code path doesn't branch, this results in a costly and redundant "destroy and rebuild" cycle on every recalculation.

We replaced the rebuild cycle with a "diffing" strategy using Python set math.

### Key Changes

1. Removed `ComputedState._remove_parents()`
2. Implemented a diffing engine in `@computed_property`


codebreaker32:
With this and #3462, We can comfortably close #3227

mesa/experimental/mesa_signals/core.py
```diff
@@ -231,18 +222,37 @@ def wrapper(self: HasEmitters):
...
+                }
+
+                # Unsubscribe from removed dependencies
+                for p, attr in old_deps - new_deps:
+                    p.unobserve(attr, ALL, state._set_dirty)
+
+                # Subscribe to newly discovered dependencies
+                for p, attr in new_deps - old_deps:
+                    p.observe(attr, ALL, state._set_dirty)
+
```
quaquel: how does this interact with `BaseObservable.__get__` and the auto register mechanism there?

If func is dependent on BaseObservable, on visit, BaseObservable will call `CURRENT_COMPUTED._add_parent`. Registering itself (potentially again).

codebreaker32: That's the catch! The secret here is that `_add_parent` no longer subscribes to anything.

We stripped the `parent.observe(...)` call entirely, making it purely passive. When `__get__` triggers it, it simply updates the local dictionary (self.parents[parent][name] = value). The actual `observe()` call only happens once during the set diffing at the very end.

That said, the confusion is totally justified because the name implies an active subscription. Do you recommend naming `_add_parent` to `_record_access` or something similar?

quaquel: > That's the catch! The secret here is that _add_parent no longer subscribes to anything.

I overlooked removing that line. Thanks for pointing it out. 

> That said, the confusion is totally justified because the name implies an active subscription. Do you recommend naming _add_parent to _record_access or something similar?

Yes, a renaming would make sense to ensure the method name follows the implementation.

codebreaker32:
Hi @quaquel 
Is there anything else you would like me to change?

quaquel:
Sorry, I am rather busy at the moment. I had hoped to handle a couple of PRs today, but other stuff cropped up.

codebreaker32:
> Sorry, I am rather busy at the moment. I had hoped to handle a couple of PRs today, but other stuff cropped up.

Its perfectly fine, I was just checking up!

quaquel:
@codebreaker32, I guess this still warrants a review? 

I quickly scanned it and it looks fine to me, but if we still want to merge this,  I'll give it a closer look as well.

codebreaker32:
> @codebreaker32, I guess this still warrants a review?
> 
> I quickly scanned it and it looks fine to me, but if we still want to merge this, I'll give it a closer look as well.

Yes and I have also given it a final look, LGTM so far!

quaquel (approved):
LGTM

PR #3486, opened 2026-03-09
Fixes a cache-invalidation bug for `SignalingList`

### LINKED ISSUE #3481: Resolve a `FIXME` in `mesa_signals/core.py`
I propose a 1-line fix to resolve this `fixme`
https://github.com/mesa/mesa/blob/4af8e24349b86b9e91a0cb1a1486099a3fdf11a5/mesa/experimental/mesa_signals/core.py#L69-L74

Currently, mutable lists(such as SignalingList) fail the `current_val != old_val` cache-invalidation check 

https://github.com/mesa/mesa/blob/4af8e24349b86b9e91a0cb1a1486099a3fdf11a5/mesa/experimental/mesa_signals/core.py#L225-L229

because they compare the exact same memory address against itself. By explicitly passing None into the dependency graph for mutable sequences, we  can fix this

```python
    def __get__(self, instance: HasEmitters, owner):
        value = getattr(instance, self.private_name)

        if CURRENT_COMPUTED is not None:
            # Do not store mutable sequences in the computed graph.
            value = value if not isinstance(value, MutableSequence) else None
            ...
```

codebreaker32:
### Summary
Fixes a cache-invalidation bug for `SignalingList`. See #3481 

### Motivation
Because mutable lists mutate in-place, the reactive graph's `current_val != old_val` check compares the list's memory address against itself. It always evaluates to `False`, causing the `@computed_property` to return a stale cache. Furthermore, storing the actual list object in `ComputedState` traps massive data structures in memory.

### Implementation
Passed `None` to the dependency tracker whenever the accessed value is a `MutableSequence`. This keeps the list out of the `ComputedState` and ensures in-place mutations successfully bypass the `old != new` cache check. Immutable standard observables remain unaffected.

Closes #3481

### DIFF
--- a/mesa/time/events.py
+++ b/mesa/time/events.py
@@ -238,6 +238,7 @@ class EventGenerator:
         self.priority = priority
 
         self._active: bool = False
+        self._paused: bool = False
         self._current_event: Event | None = None
         self._execution_count: int = 0
 
@@ -276,7 +277,7 @@ class EventGenerator:
 
     def _execute_and_reschedule(self) -> None:
         """Execute the function and schedule the next event."""
-        if not self._active:
+        if not self._active or self._paused:
             return
 
         # Check weakref HERE (execution time), not in property getter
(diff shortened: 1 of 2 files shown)


quaquel:
Could you add a test that illustrates the bug as well? Otherwise, this looks good.

codebreaker32:
> Could you add a test that illustrates the bug as well? Otherwise, this looks good.

Good call, I was unintentionally returning that same `None`. Thanks :)

PR #3491, opened 2026-03-09
Optimise `_remove_agent` in Continuous Space

### LINKED ISSUE #3490: Optimise `_remove_agent` in Continuous Space
Currently, `_remove_agent` is incredibly slow for large models. When an agent is removed, the code uses a for loop to shift every single subsequent agent down by one index in the dictionaries, and then does an $O(N)$ array slice to shift the spatial data up. Because agents in a continuous space do not need to be stored in any specific order, we can make this removal $O(1)$ using the classic "Swap and Pop" algorithm.  We simply take the last agent in the array, move it into the slot of the removed agent, and pop the end of the array.

https://github.com/mesa/mesa/blob/d8f5f1c86e1edcbb40b62cc8d5dfb67621f97c50/mesa/experimental/continuous_space/continuous_space.py#L140-L166

Or Am I missing something? Please let me know.

codebreaker32:
### Summary

This PR optimizes the internal agent removal process in `ContinuousSpace` by replacing the $O(N)$ array shifting logic with an $O(1)$ "swap-with-last" algorithm. This eliminates costly `for` loops and NumPy array slice shifting, drastically improving performance for models with high agent churn.

### Motive

See #3490

### Implementation

This PR aligns `ContinuousSpace` with the performance standards recently established in Mesa's data collection modules. It implements the exact same "swap-with-last" logic used by `NumpyAgentDataSet`.

### Additional Notes

As Mesa 4.0 moves toward the "Stacked Spaces" architecture (where Continuous Agents may have primary role), optimizing base operations can be seen as prerequisites effort before we start introducing major changes

Closes #3490

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ ci:
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.14.14
+  rev: v0.15.4
   hooks:
     # Run the linter.
     - id: ruff-check
@@ -32,7 +32,7 @@ repos:
 
 # Prose linting with Vale
 - repo: https://github.com/errata-ai/vale
-  rev: v3.13.0
+  rev: v3.13.1
   hooks:
     - id: vale
       name: Vale
--- a/benchmarks/configurations.py
+++ b/benchmarks/configurations.py
@@ -7,6 +7,11 @@ from mesa.examples import (
     SugarscapeG1mt,
     WolfSheep,
 )
+from mesa.examples.advanced.sugarscape_g1mt.model import SugarScapeScenario
+from mesa.examples.advanced.wolf_sheep.model import WolfSheepScenario
+from mesa.examples.basic.boid_flockers.model import BoidsScenario
+from mesa.examples.basic.boltzmann_wealth_model.model import BoltzmannScenario
+from mesa.examples.basic.schelling.model import SchellingScenario
 
 configurations = {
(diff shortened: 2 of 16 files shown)


mesa/experimental/continuous_space/continuous_space.py
```diff
@@ -143,25 +143,26 @@ def _remove_agent(self, agent: Agent) -> None:
...
+            # Swap in numpy array
+            self._agent_positions[index] = self._agent_positions[last_index]
+
+            # Update dictionary mappings for the swapped agent
+            self._agent_to_index[last_agent] = index
+            self._index_to_agent[index] = last_agent
+
+        # Pop the last elements
+        self.active_agents.pop()
+        self._index_to_agent.pop(last_index, None)
```
quaquel, line 163: I am fine with this change, but I wonder whether a further improvement might not be possible. 

At the moment, on every access of agent.position, the agent has to look up its index. This seems wasteful. In principle, an agent can be assigned an index when added to a space and just keep this index for its lifetime. This does mean that on the space side, the indexing becomes a bit more difficult because we need to maintain a numpy array with the currently active indices. But it might offer speedups. Thoughts are welcome.

codebreaker32: >we need to maintain a numpy array with the currently active indices

How about using `_mesa_index`? I noticed we haven't using it actively. Any specific reason? Also see my latest commit.

quaquel: My hunch is that this index was added at some point while developing the code, but was not used in the final merged PR.

PR #3493, opened 2026-03-09
Updating Scenario in preparation for replacing batch runner

quaquel:
This PR updates the `Scenario` class and how it is used inside `Model`. It is inspired by ideas in #3483. 

1. Scenario becomes the sole place where rng and stdlib random are defined. `Model` just uses these. 
2. `Model.__init__` is updated to take either a Scenario subclass or instance. (supersedes #3450)
3. Scenario is now fully frozen after instantiation without a complicated ref back to `model`. 
4. Scenarios can spawn replications. This builds on #3195 but updates the idea for use inside Scenario.  See also [the numpy docs](https://numpy.org/devdocs/reference/random/parallel.html#seedsequence-spawn) for more details on how this all works. 
5. I added two class methods for creating scenarios from either a numpy array or a pandas dataframe.
6. I updated all relevant examples to use the new machinery. 

So what are the API implications from this?

```python

# this now works and uses the default values from your custom scenario
model = MyModel(scenario=MyScenario)

# rng will be passed to MyScenario(rng=rng)
model = MyModel(scenario=MyScenario, rng=42) 

# running 10 replications of a given scenario.
# the rng is different for all 10 and derived via numpy.random.default_rng(42).spawn(10)
scenario = MyScenario(a=1, b=2, rng-42)
for replication in scenario.spawn_replications(n=10):
    model = MyModel(scenario=replication)

# scenario is frozen
my_scenario = MyScenario(a=1, b=2)
my_scenario.a = 4  # raises a TypeError - inline with e.g. frozen dataclasses and tuples

# just using the base scenario still works as before
# custom scenarios just exist to specify defaults
scenario = Scenario(rng=42, a=1, b=2)


```

### DIFF
--- a/mesa/examples/advanced/wolf_sheep/agents.py
+++ b/mesa/examples/advanced/wolf_sheep/agents.py
@@ -67,24 +67,38 @@ class Sheep(Animal):
 
     def move(self):
         """Move towards a cell where there isn't a wolf, and preferably with grown grass."""
-        cells_without_wolves = self.cell.neighborhood.select(
-            lambda cell: not any(isinstance(obj, Wolf) for obj in cell.agents)
-        )
+        cells_without_wolves = []
+        cells_with_grass = []
+
+        for cell in self.cell.neighborhood:
+            has_wolf = False
+            has_grass = False
+
+            for obj in cell.agents:
+                # If there's a wolf, we can early exit
+                if isinstance(obj, Wolf):
+                    has_wolf = True
+                    break
+                elif isinstance(obj, GrassPatch) and obj.fully_grown:
+                    has_grass = True
+
+            # Prefer cells without wolves
+            if not has_wolf:
+                cells_without_wolves.append(cell)
+
+                # Among safe cells, pick those with grown grass
+                if has_grass:
+                    cells_with_grass.append(cell)
+
         # If all surrounding cells have wolves, stay put
         if len(cells_without_wolves) == 0:
             return
 
-        # Among safe cells, prefer those with grown grass
-        cells_with_grass = cells_without_wolves.select(
-            lambda cell: any(
-                isinstance(obj, GrassPatch) and obj.fully_grown for obj in cell.agents
-            )
-        )
         # Move to a cell with grass if available, otherwise move to any safe cell
         target_cells = (
             cells_with_grass if len(cells_with_grass) > 0 else cells_without_wolves
         )
-        self.cell = target_cells.select_random_cell()
+        self.cell = self.random.choice(target_cells)
 
 
 class Wolf(Animal):
--- a/mesa/experimental/continuous_space/continuous_space.py
+++ b/mesa/experimental/continuous_space/continuous_space.py
@@ -97,7 +97,6 @@ class ContinuousSpace:
 
         #  a mapping from agents to index and vice versa
         self._index_to_agent: dict[int, Agent] = {}
-        self._agent_to_index: dict[Agent, int | None] = {}
 
     @property
     def agents(self) -> AgentSet:
@@ -127,7 +126,7 @@ class ContinuousSpace:
                 ]
             )
 
-        self._agent_to_index[agent] = index
+        agent._mesa_index = index
         self._index_to_agent[index] = agent
 
         # we want to maintain a view rather than a copy on the active agents and positions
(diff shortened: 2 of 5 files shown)


mesa/experimental/scenarios/scenario.py
```diff
@@ -33,24 +29,32 @@ class MyScenario(Scenario):
         scenario = MyScenario(rng=42, cop_vision=10)  # Override defaults
 
     Attributes:
-        model: The model instance to which this scenario belongs
         scenario_id: A unique identifier for this scenario, auto-generated starting from 0
-        rng: Random number generator or seed value
+        experiment_id: Identifies the design point (e.g., row in a QMC sample matrix)
+        replication_id: Identifies the stochastic replication within a design point
```
EwoutH, line 37: Could you elaborate on why we now need 3 different id variables? Are there other options?

mesa/experimental/scenarios/scenario.py
```diff
@@ -60,54 +64,131 @@ def __init_subclass__(cls):
...
         """Return dict representation of the scenario."""
-        return {**self.__dict__, "model": self.model, "_scenario_id": self._scenario_id}
+        return {
+            **self.__dict__,
+            "scenario_id": self.scenario_id,
+            "replication_id": self.replication_id,
+            "initial_rng_state": self.initial_rng_state,
+        }
+
+    def spawn_replications(self, n: int) -> list["Scenario"]:
```
EwoutH: Would this make sense as a Class method? Like we do with Agent.create_agents() I believe?

EwoutH (commented):
Thanks for working on this. I think you’re looking in the right direction.

mesa/experimental/scenarios/scenario.py
quaquel: Based on my experience with the workbench, there are no clean other options. 

1. We need scenario_id as a link to the original experiment matrix (e.g., from numpy.qmc, or salib)
2. We need a replication_id for reproduction. Only by knowing this can you reproduce the exact results for a given experiment.
3. We need a high entropy starting RNG value for high quality random number generation (and no 42 is not a high entropy starting value because this has 30 `0` in front of it.

I also fail to see what the problem is. Scenario_id and replication_id are in principle generated by Mesa itself so the user need not worry about them at all.

mesa/experimental/scenarios/scenario.py
quaquel: No it cannot be a class method. `spawn_replication` takes a base scenario instance with the scenario-specific user-provided arguments and creates _n_ versions of it, each with a unique, independent, seeded numpy generator.

mesa/examples/advanced/alliance_formation/model.py
```diff
@@ -24,7 +24,7 @@ class MultiLevelAllianceModel(mesa.Model):
     Model for simulating multi-level alliances among agents.
     """
 
-    def __init__(self, scenario: AllianceScenario = None):
+    def __init__(self, scenario: AllianceScenario = AllianceScenario):
```
EwoutH, line 27: I understand how this assignment works (you need both the typing and the actual return), but it doesn't read that elegent.

It does spark the question again for me if we need to reconsider making the Scenario class a class-level attribute of the Model.

> Option 2 moves the "which scenario class does this model use?" information out of the `__init__` signature and into a class-level attribute. The idea is that the scenario type is a property of the model class itself, not something you'd typically vary call-by-call.
> 
> Here's how it would look in practice:
> 
> ```python
> # In Model base class
> class Model(HasEmitters):
>     scenario_class: ClassVar[type[Scenario]] = Scenario
> 
>     def __init__(self, *, scenario=None, rng=None, **kwargs):
>         if scenario is None:
>             scenario = self.scenario_class(rng=rng)
>         elif isinstance(scenario, type) and issubclass(scenario, Scenario):
>             scenario = scenario(rng=rng)
>         elif isinstance(scenario, Scenario):
>             if rng is not None:
>                 raise ValueError("Pass either rng or scenario instance, not both.")
>         
>         self.scenario = scenario
>         self.rng = scenario.rng
>         self.random = random.Random(scenario._stdlib_seed)
>         ...
> ```
> 
> Then each example model just declares its scenario class and has a clean `__init__`:
> 
> ```python
> class WolfSheep(Model):
>     scenario_class = WolfSheepScenario
> 
>     def __init__(self, scenario=None):
>         super().__init__(scenario=scenario)
>         # model setup using self.scenario.grass, self.scenario.initial_sheep, etc.
>         ...
> ```
> 
> User-facing API becomes:
> 
> ```python
> # Uses WolfSheepScenario defaults automatically
> model = WolfSheep()
> 
> # Pass a configured instance
> model = WolfSheep(scenario=WolfSheepScenario(rng=42, initial_sheep=200))
> 
> # Just override rng, still uses WolfSheepScenario
> model = WolfSheep(rng=42)
> 
> # Replications work naturally
> base = WolfSheepScenario(rng=42)
> for replication in base.spawn_replications(10):
>     model = WolfSheep(scenario=replication)
> ```
> 
> **Advantages:**
> 
> The `__init__` signature is clean everywhere — no `scenario: WolfSheepScenario = WolfSheepScenario` pattern. The relationship between model and scenario is declared once at the class level, which feels right since it's a structural relationship (a WolfSheep model always uses WolfSheepScenario). It's also easy to discover via `WolfSheep.scenario_class`.
> 
> **The SolaraViz question is the key concern though.** If SolaraViz needs to know which scenario class to instantiate on reset, it can read `model.scenario_class` or `type(model).scenario_class` instead of introspecting the `__init__` signature. That's arguably more robust — signature introspection is fragile, while a class attribute is explicit. The visualization code would look something like:
> 
> ```python
> # In SolaraViz reset logic
> scenario_cls = model.scenario_class
> new_scenario = scenario_cls(rng=rng_from_slider, **scenario_params_from_sliders)
> new_model = type(model)(scenario=new_scenario, **model_params_from_sliders)
> ```

quaquel: The typing is actually incomplete and should be `scenario: AllianceScenario|type[Scenario]`. I don't like the class attribute approach. Moreover, to go down such a route just because of type hints that you don't like seems not that compelling a reason.

PR #3494, opened 2026-03-09
Fix SolaraViz reset crash on raw Solara components

### LINKED ISSUE #3453: SolaraViz DX Improvements: Reset bug, Class Instantiation errors, and Warning spam
# SolaraViz Reset button crashes when model_params contains raw Solara components

### Bug Summary
Clicking the "Reset" button in SolaraViz causes a crash with `TypeError: 'ValueElement' object cannot be interpreted as an integer` when `model_params` contains raw Solara components (like `solara.SliderInt()`) instead of Mesa's wrapper classes.

### Root Cause
In `mesa/visualization/solara_viz.py`, the `do_reset()` function attempts to re-instantiate the model by extracting user parameters. When a raw Solara component (yielding a `reacton.core.ValueElement`) is passed in the `model_params` dictionary without Mesa's Slider wrapper, the functions `split_model_params()` and `check_param_is_fixed()` incorrectly treat it as a "fixed" dictionary parameter.

During reset, `do_reset()` then passes this raw UI Component object directly to the model's `__init__` kwargs instead of extracting its `.value`, causing the crash when the model constructor expects a primitive type (int, float, etc.).

### Steps to Reproduce
```python
from mesa import Model
import solara

class MyModel(Model):
    def __init__(self, num_agents):
        super().__init__()
        self.num_agents = num_agents  # Expects an integer

# Problematic: passing raw Solara component
model_params = {
    "num_agents": solara.SliderInt(value=10, min=1, max=100)  # Should use mesa.visualization.Slider
}

model = MyModel(10)
page = SolaraViz(model, components=[], model_params=model_params)
# Initial load works fine, but clicking "Reset" button crashes
```

### Expected Behavior
One of the following:

1. **Defensive extraction**: `do_reset()` should detect `ValueElement` objects and extract their `.value` property before passing to the model constructor
2. **Early validation**: Raise a clear error during `SolaraViz` initialization if `model_params` contains unsupported types, with a helpful message directing users to use Mesa's `Slider`, `Checkbox`, etc. wrapper classes

### Actual Behavior
```
TypeError: 'ValueElement' object cannot be interpreted as an integer
```

The error occurs deep in the model instantiation, making it unclear that the issue is with the parameter extraction logic.

### Environment
- Mesa version: 4.0.0.dev0
- Python version: 3.14.3

### Proposed Fix
I'm willing to submit a PR to fix this. The fix would involve:

- Adding type checking in `split_model_params()`/`do_reset()` to extract `.value` from `ValueElement` objects
- OR adding validation in `SolaraViz.__init__()` to reject raw Solara components with a clear error message

Which approach would the maintainers prefer?

---

**Note**: This is part 1 of a series of SolaraViz DX improvements.

Tejasv-Singh:
### Summary
Fixes a crash in [SolaraViz](cci:1://file:///home/hada_linux/Projects/ML/mesa/mesa_upstream/mesa/visualization/solara_viz.py:56:0-227:51) when raw Solara components are passed in [model_params](cci:1://file:///home/hada_linux/Projects/ML/mesa/mesa_upstream/mesa/visualization/solara_viz.py:761:0-777:49) instead of Mesa's wrapper classes. It adds early validation to raise a clear `TypeError` explaining how to correctly use the components.

### Bug / Issue
Fixes #3453

When a raw Solara component (yielding a `reacton.core.ValueElement`) is passed in the [model_params](cci:1://file:///home/hada_linux/Projects/ML/mesa/mesa_upstream/mesa/visualization/solara_viz.py:761:0-777:49) dictionary without Mesa's wrapper classes (e.g., `solara.SliderInt()` instead of `mesa.visualization.Slider`), it gets incorrectly treated as a "fixed" dictionary parameter. 

During the reset process, the [do_reset()](cci:1://file:///home/hada_linux/Projects/ML/mesa/mesa_upstream/mesa/visualization/solara_viz.py:574:4-595:26) function passes this raw UI Component object directly to the model's [__init__](cci:1://file:///home/hada_linux/Projects/ML/mesa/mesa_upstream/tests/visualization/test_solara_viz.py:355:8-360:44) kwargs. The model constructor expects a primitive type but gets a `ValueElement` instead, resulting in a crash with the error: `TypeError: 'ValueElement' object cannot be interpreted as an integer`.

### Implementation
- Added an early validation check in `SolaraViz.__init__()` ([mesa/visualization/solara_viz.py](cci:7://file:///home/hada_linux/Projects/ML/mesa/mesa_upstream/mesa/visualization/solara_viz.py:0:0-0:0)) that iterates over [model_params](cci:1://file:///home/hada_linux/Projects/ML/mesa/mesa_upstream/mesa/visualization/solara_viz.py:761:0-777:49).
- If any value passed is an instance of `reacton.core.ValueElement`, a clear `TypeError` is immediately raised directing the user to use wrappers from `mesa.visualization` like `Slider`, `Checkbox`, etc.

### Testing
- Added a unit text [test_raw_solara_component_in_model_params](cci:1://file:///home/hada_linux/Projects/ML/mesa/mesa_upstream/tests/visualization/test_solara_viz.py:393:0-411:86) in [tests/visualization/test_solara_viz.py](cci:7://file:///home/hada_linux/Projects/ML/mesa/mesa_upstream/tests/visualization/test_solara_viz.py:0:0-0:0).
- The test deliberately passes a raw `solara.SliderInt()` to [model_params](cci:1://file:///home/hada_linux/Projects/ML/mesa/mesa_upstream/mesa/visualization/solara_viz.py:761:0-777:49) and asserts that the defensive `TypeError` is raised. 

### Additional Notes
This change creates a better Developer Experience (DX) by "failing fast" and explicitly educating the user, rather than failing deep inside the model instantiation step with a cryptic error.

Do let me know your thoughts.


Sahil-Chhoker:
@Tejasv-Singh Thanks for this PR. Guardrails are great, but in this case you're handling only `ValueElement`. I think it would be better to do it the other way around, allow the values defined in `mesa` to pass while blocking the unwanted ones.

satharyasin-hub:
Thanks for the implementation! I went through the PR and the approach of adding early validation in `SolaraViz.__init__()` makes sense for improving the developer experience.

One potential improvement could be to avoid checking specifically for `ValueElement` and instead validate against the supported Mesa wrapper classes (such as `Slider`, `Checkbox`, etc.). This could make the validation more future-proof if additional Solara component types appear.

Additionally, it might be useful to add a short note in the error message suggesting the correct wrapper usage (for example `mesa.visualization.Slider` instead of `solara.SliderInt`) so users can quickly fix the issue.

Overall the approach looks solid and the added test case is helpful in preventing regressions.

Tejasv-Singh:
Thank you so much @Sahil-Chhoker and @satharyasin-hub for the thoughtful feedback, that's a really good point.

You're right that is a much cleaner and more future-proof approach than specifically blocking `ValueElement`. 

I'll update the implementation to validate against Mesa's supported param types (`Slider`, `Checkbox`, `Choice`, `NumberInput`, `StaticText`, and primitives like `int`, `float`, `bool`, `str`) and raise a clear `TypeError` with a helpful message if anything outside that set is passed. I'll also update the test accordingly.

Will push the updated commits shortly!

Tejasv-Singh:
Hey @Sahil-Chhoker, I've pushed the updated implementation addressing all the feedback. Here's a summary of what changed:

**Switched to a pure allowlist approach:**
- Replaced the `ValueElement` blocklist with `_SUPPORTED_MODEL_PARAM_TYPES = (UserParam, dict, int, float, bool, str)`
- `UserParam` covers `Slider` and all future Mesa wrapper classes automatically, making it fully future-proof
- No dependency on Solara internals anywhere in the validation logic

**Tests updated:**
- `test_raw_solara_component_in_model_params`, confirms raw Solara components are rejected
- `test_unsupported_type_rejected`, confirms the allowlist catches other unsupported types too (`list`, custom objects, `None`)
- `test_valid_model_params_types_accepted`, confirms all valid Mesa types pass without error

Would love a re-review when you get a chance, thanks so much! :smile: :raised_hands:

Tejasv-Singh:
Hey @EwoutH, the PR has been approved by @Sahil-Chhoker, all checks are passing, and there are no conflicts with the base branch. Could you take a look and merge when you get a chance? Also, would it be possible to remove the `GSoC: Not ready` label since the requested changes have been addressed? Thank you so much! 🙏

PR #3496, opened 2026-03-10
align AgentDataSet with NumpyAgentDataSet by supporting agent class input

souro26:
## Summary

This change allows AgentDataSet to accept an agent class. When a class is passed, agents are retrieved using model.agents_by_type. This brings it in line with how NumpyAgentDataSet works.

Fixes part of #3341 

## Motivation

AgentDataSet and NumpyAgentDataSet currently use different inputs (AgentSet vs agent class). This makes their APIs inconsistent and harder to use together. This change makes both datasets follow the same pattern.

## Implementation

- AgentDataSet now accepts either an AgentSet or an agent class
- When given a class, agents are taken from model.agents_by_type
- Existing behavior using AgentSet is unchanged

### DIFF
--- a/mesa/experimental/data_collection/dataset.py
+++ b/mesa/experimental/data_collection/dataset.py
@@ -5,10 +5,9 @@ from __future__ import annotations
 import abc
 import operator
 from contextlib import suppress
-from typing import TYPE_CHECKING, Any, Protocol, runtime_checkable, Literal
+from typing import TYPE_CHECKING, Any, Literal, Protocol, runtime_checkable
 
 import numpy as np
-
 
 from mesa.agent import Agent
 from mesa.agentset import AbstractAgentSet
@@ -130,7 +129,6 @@ class AgentDataSet[A: Agent](BaseDataSet):
         fields: str | list[str] | None = None,
     ):
         """Init. of AgentDataSet."""
-
         if args:
             if len(args) > 1:
                 raise TypeError("too many positional arguments")
@@ -202,7 +200,7 @@ class AgentDataSet[A: Agent](BaseDataSet):
             self._is_dirty = False
 
         return self._cache
-    
+
     def _get_agents(self):
         if self._mode == "set":
             return self._agents or []
--- a/mesa/experimental/data_collection/dataset.py
+++ b/mesa/experimental/data_collection/dataset.py
@@ -5,9 +5,10 @@ from __future__ import annotations
 import abc
 import operator
 from contextlib import suppress
-from typing import TYPE_CHECKING, Any, Protocol, runtime_checkable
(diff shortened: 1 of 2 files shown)


mesa/experimental/data_collection/dataset.py
```diff
@@ -543,13 +544,19 @@ def create_dataset(
...
         *,
         use_dirty_flag: bool = False,
     ) -> AgentDataSet:
-        """Track the specified fields for the agents in the AgentSet."""
+        """Track the specified fields for the agents.
+
+        The first argument can be either an AgentSet or an Agent class.
+        If an Agent class is passed, the registry resolves the corresponding agents from the model.
+        """
+        if isinstance(agents, type) and issubclass(agents, Agent):
```
quaquel: this won't work, use self.model.agents_by_type instead.

mesa/experimental/data_collection/dataset.py
```diff
@@ -514,8 +514,9 @@ def record(
 class DataRegistry:
     """A registry for data sets."""
 
-    def __init__(self):
+    def __init__(self, model):
         """Initialize the registry."""
+        self.model = model
```
quaquel: I am not sure I like this circular reference from odel to registry and back. It is only needed for track_agents and because you want to support agent types here. How valuable is that really?

mesa/experimental/data_collection/dataset.py
```diff
@@ -619,7 +626,7 @@ def __iter__(self):
 
     model = BoltzmannWealth()
     model.test = 5
-    agent_data = AgentDataSet("wealth", model.agents, "wealth")
+    agent_data = AgentDataSet("wealth", model.agents, fields="wealth")
     # model_data = ModelDataSet("gini", model, "test", gini=model.compute_gini)
     data = []
     for _ in range(5):
```
quaquel, line 679: what is this doing here? It's probably a left over from some other PR.

quaquel:
I am not sure this PR achieves its stated aims. I wanted to primarily change AgentDataSet, not track_agents.

souro26:
my idea here was to keep the change at the registry level so users could pass either an AgentSet or an Agent class without changing the dataset api. since most usage goes through DataRegistry anyway, i thought that it was simplest way to line up AgentDataSet and NumpyAgentDataSet from the outside.

looking at it again, this doesn’t really align the datasets themselves, it just hides the difference at the entry point. it also ends up pulling the model into the registry just to resolve agent classes, which isn’t great either. i'll move this towards updating AgentDataSet to take an agent class directly and keep the registry thin.

souro26:
@quaquel Apologies for the delay, I was wrapping up my proposal. I've updated the PR to align AgentDataSet with NumpyAgentDataSet and cleaned up the implementation and tests. Everything should be ready for review now, thanks.

PR #3497, opened 2026-03-10
Improve Slider validation for value range and step size

Shree600:
The current implementation of Slider validates only whether value, min, and max are provided. However, it does not ensure that these parameters form a logically valid slider configuration. Sliders can be created with invalid configurations such as a value outside the defined range, a minimum value greater than the maximum, or a non-positive step size. These situations may lead to inconsistent behavior or errors in visualization components that rely on the slider.
This pull request improves the validation logic by ensuring that:
The minimum value is less than the maximum value.
The slider value lies within the defined range.
The step size is a positive number.
If any of these conditions are violated, a ValueError is raised during initialization.

**### Test Examples**

**Valid slider configuration**

A slider is created with a value that lies within the range, and with a positive step size.
The slider initializes normally and works as expected.

**Value outside the range**

A slider is created where the value is greater than the maximum allowed value.
Previously this was accepted, but after the change it raises a ValueError.

**Invalid range**

A slider is created where the minimum value is greater than the maximum value.
Previously this configuration was allowed, but now it raises a ValueError.

**Zero step size**

A slider is created with a step size of zero.
Previously this configuration was allowed, but now initialization fails with a ValueError.

**Negative step size**

A slider is created with a negative step value.
Previously this was accepted, but the new validation correctly raises a ValueError.

### DIFF
--- a/mesa/visualization/user_param.py
+++ b/mesa/visualization/user_param.py
@@ -56,7 +56,7 @@ class Slider(UserParam):
         # Validate option type to make sure values are supplied properly
         valid = (
             self.value is not None
-            and self.min is not None 
+            and self.min is not None
             and self.max is not None
             and self.min < self.max
             and self.min <= self.value <= self.max


quaquel:
This is breaking various tests. Please ensure tests are passing.

quaquel:
Closing this as stale.

PR #3498, opened 2026-03-10
Safe file loading with error  handling

riddhismita06-cloud:
## Summary
Improves robustness of pickle file loading by introducing a `load_pickle()` helper with structured error handling.

## Problem
The original code had two identical `open()` blocks with no error handling. If either file was missing or corrupted, the script would crash with a raw Python traceback — unhelpful for users and noisy in CI pipelines.

## Changes
- **DRY refactor:** Duplicate `open()` blocks replaced with a single reusable `load_pickle(filename: str) -> dict` function
- **`FileNotFoundError`:** Prints `Error: File '<path>' not found.` to `stderr` and exits with code `1`
- **Catch-all exception:** Handles corrupted or unreadable files with a descriptive message
- **Graceful exit:** Uses `sys.exit(1)` so calling scripts/CI can detect failure via exit code

## Before
```python
with open(f"{filename1}.pickle", "rb") as handle:
    timings_1 = pickle.load(handle)

with open(f"{filename2}.pickle", "rb") as handle:
    timings_2 = pickle.load(handle)
```

## After
```python
def load_pickle(filename: str) -> dict:
    path = f"{filename}.pickle"
    try:
        with open(path, "rb") as handle:
            return pickle.load(handle)  # noqa: S301
    except FileNotFoundError:
        print(f"Error: File '{path}' not found.", file=sys.stderr)
        sys.exit(1)
    except Exception as e:
        print(f"Error loading '{path}': {e}", file=sys.stderr)
        sys.exit(1)

timings_1 = load_pickle(filename1)
timings_2 = load_pickle(filename2)
```

## Testing
- [ ] Verified normal execution with valid pickle files works as before
- [ ] Confirmed a missing file prints a clean error and exits with code `1`
- [ ] Confirmed a corrupted pickle file is caught and reported cleanly

## Notes
No behaviour change for the happy path — this only affects error scenarios.

### DIFF
--- a/benchmarks/compare_timings.py
+++ b/benchmarks/compare_timings.py
@@ -99,14 +99,16 @@ def performance_emoji(lower: float, upper: float, threshold: float = 3.0) -> str
 
 
 # Iterate over the models and sizes, perform analysis, and populate the DataFrame
-rows=[]
+rows = []
 for model, size in timings_1:
     model_name = model.__name__
 
     if (model, size) not in timings_2:
-        print(f"Warning: ({model.__name__}, {size}) found in file1 but not in file2 — skipping.", file=sys.stderr)
+        print(
+            f"Warning: ({model.__name__}, {size}) found in file1 but not in file2 — skipping.",
+            file=sys.stderr,
+        )
         continue
-
 
     # Calculate percentage change and confidence interval for init times
     (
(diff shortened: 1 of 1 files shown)


benchmarks/compare_timings.py
```diff
@@ -1,13 +1,58 @@
...
+    parser.add_argument(
+        "--threshold",
+        type=float,
+        default=3.0,
+        help="Percentage threshold for flagging significant change (default: 3.0)",
+    )
+    return parser.parse_args()
+
+
+def load_pickle(filename: str) -> dict:
```
souro26, line 43: load_pickle() terminates the process using sys.exit(1) when an error occurs. This couples a utility function to CLI behavior and makes it harder to reuse or test. A better approach is to raise the exception and let the CLI entrypoint decide how to handle termination. Helper functions should generally not exit the program directly.

quaquel:
This adds a lot of code for something that has been working fine for several years. I fail to see the point.

PR #3499, opened 2026-03-10
fix: replace pd.concat loop with list collection to fix O(n²) performance

riddhismita06-cloud:
## Summary
Replaces `pd.concat` inside a loop with a list-based row collection, fixing an O(n²) performance issue.

## Problem
Calling `pd.concat` on every loop iteration creates a full copy of the DataFrame each time. For large benchmark sets this becomes extremely slow and memory-intensive.

## Before
```python
results_df = pd.DataFrame()

for model, size in timings_1:
    row = pd.DataFrame({...})
    results_df = pd.concat([results_df, row], ignore_index=True)
```

## After
```python
rows = []

for model, size in timings_1:
    rows.append({...})

results_df = pd.DataFrame(rows, columns=["Model", "Size", "Init time [95% CI]", "Run time [95% CI]"])
```

## Why
Collecting rows in a list and calling `pd.DataFrame` once at the end is O(n), compared to O(n²) for the previous approach. No change to output or behaviour.

### DIFF
--- a/benchmarks/compare_timings.py
+++ b/benchmarks/compare_timings.py
@@ -99,14 +99,16 @@ def performance_emoji(lower: float, upper: float, threshold: float = 3.0) -> str
 
 
 # Iterate over the models and sizes, perform analysis, and populate the DataFrame
-rows=[]
+rows = []
 for model, size in timings_1:
     model_name = model.__name__
 
     if (model, size) not in timings_2:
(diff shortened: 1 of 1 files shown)


souro26:
The PR includes several unrelated changes beyond what is mentioned in the description. It would be better to keep each PR focused on a specific change, by working on a separate branch instead of main.

quaquel:
I am closing this PR. Feel free to submit a clean one with only the one change.

PR #3500, opened 2026-03-10
fix: prevent IndexError and add clear error message in select_random_empty_cell() when grid is full

BEASTSHRIRAM:
### Summary
Fixes a crash in `Grid.select_random_empty_cell()` when the grid is completely full by wrapping `random.choice()` in a try-catch block and raising a clear `ValueError` instead of a cryptic `IndexError`.

### Bug / Issue
Fixes #3014

When a user's agent-based model fills a grid to 100% capacity and calls `select_random_empty_cell()`, the method crashes with an unhelpful error:

```
IndexError: Cannot choose from an empty sequence
```

This happens because the heuristic tries 50 random cell selections (all fail on a full grid), then falls back to `np.argwhere(self.property_layers["empty"])` which returns an empty array. When `random.choice()` is called on an empty array, it raises `IndexError`, leaving users confused about what went wrong.

### Implementation
Wrapped `self.random.choice(empty_coords)` in a try-except block in `Grid.select_random_empty_cell()` (mesa/discrete_space/grid.py) to catch `IndexError` and raise a clear `ValueError` with proper exception chaining.

**Code change:**
```python
empty_coords = np.argwhere(self.property_layers["empty"])
try:
except IndexError as e:
    raise ValueError(
        "Grid is completely full. No empty cells available. "
        "Cannot select a random empty cell."
    ) from e
return self._cells[tuple(random_coord)]
```

This approach follows the EAFP (Easier to Ask for Forgiveness than Permission) principle, is more Pythonic, and provides better exception chaining for debugging.

### Testing
Added unit test `test_infinite_loop_on_full_grid()` in test_discrete_space.py that fills a 2×2 grid completely and asserts that the new `ValueError` with the descriptive message is raised.

Verified test passes with the implementation.

### Additional Notes
This change creates a better Developer Experience (DX) by "failing fast" with an explicit, actionable error message rather than crashing deep in the code with a cryptic exception. Users immediately understand that their grid is full and can adjust their model parameters accordingly (more cells, faster agent movement, etc.).

No breaking API changes — only the error type changed, which improves clarity without reducing functionality.

### DIFF
--- a/mesa/discrete_space/discrete_space.py
+++ b/mesa/discrete_space/discrete_space.py
@@ -105,6 +105,12 @@ class DiscreteSpace[T: Cell](ABC):
             if you rely on getting neighborhoods of cells with a radius higher than 1, these might not be cleared
             correctly if you are adding or removing cells and connections at runtime.
 
+        Warning:
+            Coordinate Collision: If a cell already exists at the specified
+            coordinates (e.g., cell1), it will be overwritten silently by the
+            new cell (cell2).
+
+            Ensure the target coordinates are vacant before calling this method.
         """
         self.__dict__.pop("all_cells", None)
         self._cells[cell.coordinate] = cell
--- a/mesa/experimental/mesa_signals/core.py
+++ b/mesa/experimental/mesa_signals/core.py
@@ -150,9 +150,32 @@ class Observable(BaseObservable):
 class ComputedState:
     """Internal class to hold the state of a computed property for a specific instance."""
 
-    __slots__ = ["__weakref__", "func", "is_dirty", "name", "owner", "parents", "value"]
+    __slots__ = [
+        "__weakref__",
+        "dependencies",
+        "func",
+        "is_dirty",
+        "name",
+        "owner",
+        "parents",
+        "value",
+    ]
 
-    def __init__(self, owner: HasEmitters, name: str, func: Callable):
+    def __init__(
+        self,
+        owner: HasEmitters,
+        name: str,
+        func: Callable,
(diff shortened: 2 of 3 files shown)


BEASTSHRIRAM:
Hi! I was going through the project setup and noticed that Docker is mentioned in readme. I use Docker quite often, but I couldn’t find a docker-compose.yml file in the repository.
Would you like me to create one for the project? I’d be happy to work on setting it up if needed.

mesa/discrete_space/grid.py
```diff
@@ -299,6 +299,11 @@ def select_random_empty_cell(self) -> T:  # noqa
                     return cell
 
         empty_coords = np.argwhere(self.property_layers["empty"])
+        if len(empty_coords) == 0:
```
quaquel: why not wrap self.random.choice in a try catch and then raise the value error from the index error?

BEASTSHRIRAM: Thanks for the feedback! Implementing now.

mesa/discrete_space/grid.py
```diff
@@ -299,7 +299,13 @@ def select_random_empty_cell(self) -> T:  # noqa
                     return cell
 
         empty_coords = np.argwhere(self.property_layers["empty"])
-        random_coord = self.random.choice(empty_coords)
+        try:
+            random_coord = self.random.choice(empty_coords)
+        except IndexError as e:
+            raise ValueError(
+                "Grid is completely full no empty cells available. "
```
```suggestion
                "Grid is completely full. No empty cells available. "
```

BEASTSHRIRAM:
Done! ThankYou @quaquel

PR #3503, opened 2026-03-10
Optimise Wolf-sheep performance

PietroMondini:
### Approval Link

https://github.com/mesa/mesa/issues/2565

### Summary

Optimizes the movement logic in the wolf-sheep example by replacing redundant CellCollection.select() passes with a single manual loop using an early exit strategy.

### Motive

The previous implementation performed two separate .select() passes over the neighborhood on every agent step — one to filter wolf-free cells, and one to find grass. Each pass iterated over all agents in all neighboring cells. In dense simulations with many agents, this becomes a significant bottleneck.

### Implementation

Replaced the two **CellCollection.select()** calls in **Sheep.move()** with a single manual loop over the neighborhood.
Added **has_wolf / has_grass** flags in order to break immediately upon finding a wolf, skipping remaining agents in that cell, and check both wolf and grass in the same loop.

### DIFF
--- a/.github/workflows/benchmarks-comment.yml
+++ b/.github/workflows/benchmarks-comment.yml
@@ -1,0 +1,37 @@
+name: Comment benchmark results
+
+on:
+  workflow_run:
+    workflows: ["Performance benchmarks"]
+    types:
+      - completed
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  comment:
+    if: github.event.workflow_run.conclusion == 'success'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download benchmark results
+        uses: actions/download-artifact@v8
+        with:
+          name: benchmark-results
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Comment PR
+        uses: actions/github-script@v8
+        with:
+          script: |
+            const fs = require('fs');
+            const prNumber = parseInt(fs.readFileSync('pr_number.txt', 'utf8').trim());
+            const output = fs.readFileSync('benchmark_output.txt', 'utf8');
+            await github.rest.issues.createComment({
+              issue_number: prNumber,
+              owner: context.repo.owner,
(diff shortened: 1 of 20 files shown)


mesa/examples/advanced/wolf_sheep/agents.py
```diff
@@ -67,24 +67,38 @@ def feed(self):
...
+                    has_grass = True
+
+            # Prefer cells without wolves
+            if not has_wolf:
+                cells_without_wolves.append(cell)
+
+                # Among safe cells, pick those with grown grass
+                if has_grass:
+                    cells_with_grass.append(cell)
+
```
quaquel, line 92: I have been looking at this code myself before as well and agree with your improvement. I am in principle fine with merging this. However, I am wondering whether a further improvement might not be possible: what about implementing grass as a property layer rather than as agent?

PietroMondini: Thank you for the feedback! Honestly it also feels more conceptually correct to me, since grass is more a passive property of a cell rather than an active agent. Moving it to a _PropertyLayer_ would eliminate the need to iterate over _cell.agents_ entirely when checking for grass.
I've been looking into implementing this and i have one question: should the grass regrowth still be handled via _model.schedule_event()_, or it'll be better with a different approach?

quaquel: I would still ike to use scheduling, but if we use a property layer, there is no natural place for the scheduling. It would have to be done inside the sheep when its eating a patch. From an encapsulation point of view, this seems a bit strange.

PietroMondini: A potential solution could be to keep the scheduling part inside the _GrassPatch_ agent and move the state to the model into a _fully_grown_ **PropertyLayer**.
So GrassPatch becomes a pure behavior agent with no state of his own. That's what i'm currently implementing.

Would it make sense to merge the current early exit optimization as-is and open a separate issue to track this work?

quaquel: that's fine with me.

PR #3508, opened 2026-03-10
Fix #3505: Add available_cells, fix VoronoiGrid capacity overwrite

### LINKED ISSUE #3505: capacity parameter in grid constructors not honored by most things
**Describe the bug**
When instantiating a grid object (e.g., `OrthogonalVonNeumannGrid`) there is a `capacity` parameter that ostensibly should limit the number of agents that each cell in that grid can accommodate. But this does not appear to be used, at least in many of the places where this limit could be exceeded.

**Expected behavior**
I expected an "overfull; move_to operation denied" error to arise when I did this:

```
m = Model()
grid = OrthogonalVonNeumannGrid((2,2),torus=False, capacity=1)
some_cell = grid.all_cells.cells[0]
bob = CellAgent(m)
julie = CellAgent(m)
bob.move_to(some_cell)
julie.move_to(some_cell)     <-- I expected this to throw an error.
print(some_cell)

output:
Cell((0, 0), [<mesa.discrete_space.cell_agent.CellAgent object at 0x7787172616d0>, 
<mesa.discrete_space.cell_agent.CellAgent object at 0x7787172611d0>])
```

Instead, the cell happily accepts two agents, even though it is in a grid whose `capacity` is set to 1.

**To Reproduce**
See code in previous section.

**Additional context**
If `capacity` isn't enforced, then why does it exist? One would think its sole purpose would be to act as a guardrail against attempts like in that sample code.

If this behavior is intended, then at the very least the docstring should have some "just kidding, it doesn't actually enforce this" language. At present it reads:

```
Type:           OrthogonalVonNeumannGrid
...
Args:
   ...
    capacity: capacity of the grid cell
  ...
```

DipayanDasgupta:
<html>
<body>
<html><head></head><body><h2>Summary</h2>
<p>Addresses the capacity concerns raised in #3505. While <code>CellFullException</code> is already correctly raised by <code>move_to</code> on the current <code>main</code> branch (the original reporter was likely on an older release), this PR fixes two real architectural gaps and one silent bug that the investigation uncovered.</p>
<hr>
<h2>Changes</h2>
<h3>1. <code>Grid.available_cells</code> property (new) — <code>mesa/discrete_space/grid.py</code></h3>
<p><code>Grid</code> already has <code>empties</code> and <code>select_random_empty_cell()</code>, but <code>cell.empty</code> flips to <code>False</code> the moment <strong>any</strong> agent enters — even in a cell with <code>capacity=5</code>. This makes the capacity parameter nearly impossible to use proactively.</p>
<p>The new <code>available_cells</code> property returns a <code>CellCollection</code> of all cells where <code>not cell.is_full</code>, i.e. cells that still have room regardless of whether they are completely empty or partially occupied.</p>
<pre><code class="language-python"># Before: no way to find a partially-filled but non-full cell
# After:
free_cell = grid.select_random_available_cell()
agent.move_to(free_cell)

# or inspect the full collection:
print(len(list(grid.available_cells)))  # cells that still have room
</code></pre>
<h3>2. <code>Grid.select_random_available_cell()</code> method (new) — <code>mesa/discrete_space/grid.py</code></h3>
<p>Companion method to <code>available_cells</code>. Uses the same two-phase O(1) heuristic as <code>select_random_empty_cell()</code> (50 random attempts, then explicit-list fallback). Raises <code>IndexError</code> with a clear message if every cell is at capacity.</p>
<h3>3. <code>VoronoiGrid._build_cell_polygons</code> bugfix — <code>mesa/discrete_space/voronoi.py</code></h3>
<p><code>_build_cell_polygons</code> was unconditionally overwriting every cell's capacity with a value derived from the polygon area via <code>capacity_function</code>, <strong>silently discarding</strong> any explicit <code>capacity</code> argument passed by the user.</p>
<pre><code class="language-python"># Before this fix:
grid = VoronoiGrid(points, capacity=1)
# → every cell's capacity was overwritten with round_float(polygon_area)
# → CellFullException was never raised, user-provided limit silently ignored

# After:
grid = VoronoiGrid(points, capacity=1)
# → capacity=1 is preserved; CellFullException fires correctly
</code></pre>
<p>The fix is a single conditional: <code>capacity_function</code> is only applied when <code>self.capacity is None</code>.</p>
<hr>
<h2>Tests</h2>
<p>Added <code>tests/discrete_space/test_grid_capacity.py</code> with 35 tests across 5 sections:</p>

Section | What is tested
-- | --
1 · CellFullException | move_to, direct .cell=, capacity=None, vacate-and-refill — parametrized across Moore, VonNeumann, HexGrid
2 · available_cells | Empty grid, full cell excluded, partial fill included, dynamic update after vacate
3 · select_random_available_cell | Non-full result, 30-sample consistency with available_cells, IndexError on full grid
4 · VoronoiGrid bugfix | Explicit capacity preserved, area function used when capacity=None, runtime enforcement, custom capacity_function
5 · Cell.is_full | Unlimited capacity, 0→1→2→1 transitions, direct add_agent raises

<p>All 35 tests pass on the current <code>main</code> branch (verified locally with <code>pytest tests/discrete_space/test_grid_capacity.py -v</code>).</p>
<hr>

<p>Closes #3505</p></body></html>
</body>
</html>

### DIFF
--- a/mesa/discrete_space/grid.py
+++ b/mesa/discrete_space/grid.py
@@ -24,6 +24,7 @@ import numpy as np
 from scipy.spatial import KDTree
 
 from mesa.discrete_space import Cell, DiscreteSpace
+from mesa.discrete_space.cell_collection import CellCollection
 
 T = TypeVar("T", bound=Cell)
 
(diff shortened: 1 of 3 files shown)


quaquel:
This is not the correct fix for the bug. See my comment in the issue.

DipayanDasgupta:
Excellent catch @quaquel. That perfectly explains why I couldn't reproduce the bug on my local `main` branch (where it's correctly passed as `capacity=capacity`) but the user was seeing it on `3.5.0`. 
I definitely overdid this by confusing a new feature with a straightforward bug report. Sorry for adding noise to the issue.I'll step back so you can push the fix for the backport.
That being said, during my investigation I did notice two things I think could be valuable for 4.0:1. There is currently no way to proactively query partially filled cells (e.g. `Grid.available_cells`), making capacity hard to use defensively.
2. `VoronoiGrid` unconditionally overwrites explicit `capacity` arguments with its area function.
 Would you be open to me submitting these additions as a separate, dedicated **Enhancement** PR for 4.0?

quaquel:
Atomic PR's on either issue are welcome. But consider first opening issues on them and ensure agreement on how to approach them before opening PRs.

PR #3514, opened 2026-03-11
Fix `DataCollector` `model_reporters` list-shape validation to avoid `IndexError`

### LINKED ISSUE #3513: `DataCollector` accepts malformed list-style model_reporters ([func]) and crashes with `IndexError` during `collect()`
**Describe the bug**
`DataCollector` accepts malformed list-style `model_reporters` entries with missing parameter lists (e.g., `[func]`) during validation, then crashes during `collect()` with a low-level `IndexError` instead of a clear validation error.

**Expected behavior**
Malformed list-style model reporters should be rejected consistently with a descriptive `ValueError`, e.g. indicating expected format `[function, [param1, param2]]`.

**To Reproduce**
```python
from mesa import Model

def f(model):
    return 1

m = Model()
dc = DataCollector(model_reporters={"bad": [f]})  # malformed: missing params list
dc.collect(m)  # crashes
```

Actual Result:
```text
IndexError: list index out of range
```

Relevant code path:
- Validation is incomplete: `mesa/datacollection.py` (`_validate_model_reporter`)
- Crash occurs at `reporter[0](*reporter[1])` in `collect()`

**Additional context**
This is a validation consistency bug in `DataCollector`:
  - Invalid format `["not_callable", [1, 2]]` already raises a clear `ValueError`.
  - Invalid format `[callable_only]` currently bypasses validation and fails later with `IndexError`.

falloficaruss:
### Summary
Fixes a `DataCollector` validation gap where malformed list-style `model_reporters` (for example `[func]`) bypass validation and later crash in `collect()` with `IndexError`: list index out of range.
Impact: users get an internal low-level exception instead of a clear actionable validation error.

### Bug / Issue
Closes #3513 

Context:
- `model_reporters` supports list format: `[function, [params...]]`.
- Current validation only checks that the list is non-empty and the first item is callable.
- A malformed reporter like `[func]` passes validation, then fails at collection time when code accesses `reporter[1]`.

Expected:
- Malformed list-style reporters should raise a clear `ValueError` describing expected format.

Actual (before fix):
- `IndexError`: list index out of range raised from `collect()`.

### Implementation
Updated mesa/datacollection.py in two places:

1. _validate_model_reporter(...)
- Strengthened list reporter validation to require:
   - exactly 2 elements,
   - callable first element,
   - second element is a params container (list or tuple).
2. `collect(...)`
   - Added defensive validation in the list branch before calling `reporter[0](*reporter[1])`.
   - Raises the same clear `ValueError` message if format is invalid.

This ensures malformed list reporters fail with a user-facing validation error instead of an internal `IndexError`.
### Testing
Added regression tests in `tests/test_datacollector.py`:
- `test_function_error_missing_params_list`
   - verifies `[lambda m: ...]` raises `ValueError` with informative message.
- `test_function_error_invalid_params_type`
   - verifies `[lambda m, x: ..., 1]` raises `ValueError` with informative message.

### Additional Notes
- This PR intentionally focuses on malformed list-shape handling and clearer exceptions.

---

### Context & motivation
I was exploring `DataCollector` and testing edge cases around reporter validation when I noticed an inconsistency in how malformed `model_reporters` lists were handled. A clearly invalid format like `["not_callable", [1, 2]]` already raised a clean `ValueError`, but another malformed case like `[func]` slipped through validation and later failed with a low-level `IndexError` during `collect()`. I wanted to fix this because it felt like a small but important usability bug: users should get a clear validation error that points to what they configured wrong, rather than an internal indexing failure.

### What I learned
Working on this helped me understand the different reporter shapes supported by `DataCollector` and how validation and collection interact. What was more subtle than I expected was that fixing the obvious bug was only part of the work; I also had to think about exception consistency, test coverage, and whether the validation change might unintentionally narrow accepted input types. That made the review process more interesting, because it was not just about making the error go away, but about making the behavior clearer and safer without changing more than necessary.

### Learning repo

🔗 My learning repo: https://github.com/falloficarus22/GSoC-learning-space.git
🔗 Relevant model(s): This is a bug fix in `DataCollector` validation and does not require a new feature example model.

### Readiness checks
- [x] This PR addresses an agreed-upon problem (linked issue or discussion with maintainer approval), **or** is a small/trivial fix
- [x] Another GSoC contributor has reviewed this PR: @souro26 

### DIFF
--- a/mesa/discrete_space/voronoi.py
+++ b/mesa/discrete_space/voronoi.py
@@ -12,7 +12,7 @@ Useful for models requiring irregular but mathematically meaningful spatial
 divisions, like territories, service areas, or natural regions.
 """
 
-from collections.abc import Sequence
+from collections.abc import Callable, Sequence
 from itertools import combinations
 from random import Random
 
@@ -185,10 +185,9 @@ class VoronoiGrid(DiscreteSpace):
     def __init__(
         self,
         centroids_coordinates: Sequence[Sequence[float]],
-        capacity: float | None = None,
+        capacity: int | Callable | None = None,
         random: Random | None = None,
         cell_klass: type[Cell] = Cell,
-        capacity_function: callable = round_float,
     ) -> None:
         """A Voronoi Tessellation Grid.
 
@@ -204,8 +203,19 @@ class VoronoiGrid(DiscreteSpace):
             capacity_function (Callable): function to compute (int) capacity according to (float) area
 
         """
-        super().__init__(capacity=capacity, random=random, cell_klass=cell_klass)
+        # Separate callable capacity from numeric capacity before passing to base class
+        if callable(capacity):
+            capacity_function = capacity
+            numeric_capacity = None
+        else:
+            capacity_function = None
+            numeric_capacity = capacity
+
+        super().__init__(
+            capacity=numeric_capacity, random=random, cell_klass=cell_klass
+        )
         self.centroids_coordinates = centroids_coordinates
+        self.capacity_function = capacity_function
         self._validate_parameters()
 
         # Build KD-tree for fast nearest-centroid lookup
@@ -226,7 +236,6 @@ class VoronoiGrid(DiscreteSpace):
         self.regions = None
         self.triangulation = None
         self.voronoi_coordinates = None
-        self.capacity_function = capacity_function
 
         self._connect_cells()
         self._build_cell_polygons()
@@ -262,6 +271,7 @@ class VoronoiGrid(DiscreteSpace):
     def _validate_parameters(self) -> None:
         if self.capacity is not None and not isinstance(self.capacity, float | int):
             raise ValueError("Capacity must be a number or None.")
+
         if not isinstance(self.centroids_coordinates, Sequence) or not isinstance(
             self.centroids_coordinates[0], Sequence
         ):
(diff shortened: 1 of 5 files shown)


mesa/datacollection.py
```diff
@@ -381,6 +385,15 @@ def collect(self, model):
                     )
                 # Check if function with arguments
                 elif isinstance(reporter, list):
+                    if (
+                        len(reporter) != 2
```
souro26: there is already validation for the reporter list format in `_validate_model_reporter` earlier in collect, so this additional check might be redundant.

falloficaruss: Validation only runs when `self._validated` is False (first collect cycle). After that, users can still mutate `model_reporters` (or load altered state), and without the guard you can still get a raw `IndexError`. The extra check preserves a clear `ValueError` contract instead of leaking internal indexing errors.

souro26: My bad, i overlooked that. Thanks for this.

falloficarus22 (commented):
### Self Review

- Fixes a real usability bug - users now get clear `ValueError` instead of cryptic `IndexError`
- Dual validation (in `_validate_model_reporter` and `collect`) handles edge case where reporters are mutated after first validation
- Tests verify both malformed cases mentioned in the PR
- Error messages are clear and actionable

falloficaruss:
@EwoutH ready for review

EwoutH:
Isn’t this just user error? Is the complexity it adds proportional to the improvement in user experience?

falloficaruss:
> Isn’t this just user error? Is the complexity it adds proportional to the improvement in user experience?

I don’t think this is only user error in the usual sense. The reporter list format is part of the documented API, and malformed input was producing a low-level internal `IndexError` rather than a clear validation error. My goal was to make that failure mode consistent and actionable. In this case the added complexity seemed small relative to the benefit of replacing an internal crash with a predictable API-level error.

Issue #2565, opened 2024-12-22
Performance optimization opportunities for wolf-sheep movement 


EwoutH:
The current movement implementation in the wolf-sheep model performs redundant checks when agents search their neighborhood for valid moves, after https://github.com/projectmesa/mesa/pull/2503. This could be performance optimized. Two promising optimization approaches:

1. Early exit strategy
Instead of checking every cell in the neighborhood for agent types, we can break out of loops early once a valid cell is found. For example, when a sheep checks for wolves, it can immediately skip a cell once a wolf is detected rather than continuing to check other agents.

2. PropertyLayer-based vectorization 
We can leverage Mesa's PropertyLayer to track wolf and grass positions as boolean layers. By having wolves and grass patches update their position in the property layer during movement, we can perform vectorized operations to find valid moves instead of iterating through agent lists. This would allow expressing the entire movement logic using fast numpy operations.

Trade-offs:
- Early exits: Simpler implementation but more limited gains
- PropertyLayer: More complex but potentially much faster for large grids, at cost of extra memory

Worth benchmarking both approaches against current implementation with different model parameters.

ricor07:
Hello, I'd like to work on the project by fixing this issue. Can you assign this to me? Thanks

EwoutH:
Awesome! Let me know if you have any questions.

Sahil-Chhoker:
@EwoutH I looked into this issue and tried optimizing the code for movement using property layers but I couldn't get that much of a performance boost. Is my approach wrong, and If yes how can it be improved?
What I have done is I centralized the movement logic inside `Animal` class using three property layers each for grass, sheep and wolf, something like this:
```python
def move(self):
        """Generic movement logic using property layers."""
        current_pos = self.cell.coordinate
        x, y = current_pos
        w = self.model.grid.width
        h = self.model.grid.height

        # Get neighbor positions using array operations
        neighbors = np.array(
            [((x + 1) % w, y), ((x - 1) % w, y), (x, (y + 1) % h), (x, (y - 1) % h)]
        )

        # Get target layer based on agent type
        target_layer = (
            self.model.sheep_layer if isinstance(self, Wolf) else self.model.grass_layer
        )
        avoid_layer = self.model.wolf_layer if isinstance(self, Sheep) else None

        # Create masks for valid moves
        if avoid_layer is not None:
            valid_mask = np.array(
                [avoid_layer.data[tuple(pos)] == 0 for pos in neighbors]
            )
[...]

PR #3522, opened 2026-03-12
Grass and Wolves as a PropertyLayer

PietroMondini:
### Approval Link

https://github.com/mesa/mesa/issues/2565

### Summary

Introduces boolean **PropertyLayer** instances to enable vectorized NumPy lookups for wolf and grass state, eliminating per-cell agent iteration entirely.

### Implementation

Introduced two boolean **PropertyLayer** instances on the model (_wolf_layer, grass_layer_) to track wolf positions and grass state as NumPy arrays. The **GrassPatch** agent is refactored to hold no state of its own _fully_grown_ is removed and its value is read and written directly from _self.model.grass_layer.data[x, y]._ **GrassPatch** retains ownership of the scheduling logic via _get_eaten()_ and _regrow()_, preserving encapsulation. This allows _Sheep.move()_ to replace per-cell agent iteration with a single vectorized NumPy lookup over neighbor coordinates, and Sheep.feed() to check grass state via a direct array index instead of iterating over cell.agents.

### Context & motivation

I was looking for open issues to contribute to Mesa and getting my hands on it, issue #2565 caught my attention, and i thought  that with my past experience in ABM and numpy i could optimize agent interactions.

### What I learned

I learned how to manage the interactions between agents, and with the model, optimizing them. I'm also trying to implement a next step enhancement using PropertyLayers to represent positions of Grass and Wolves.
To do so i've learned how PropertyLayer works and how to keep them in sync during the simulation, and to optimize using vectorization.

### Learning repo

🔗 My learning repo:  https://github.com/PietroMondini/GSoC-learning-space
🔗 Relevant model(s): https://github.com/mesa/mesa/tree/main/mesa/examples/advanced/wolf_sheep

### Readiness checks
- [x]  This PR addresses an agreed-upon problem (linked issue or discussion with maintainer approval), **or** is a small/trivial fix
- [x] Another GSoC contributor has reviewed this PR: @souro26 

### DIFF
--- a/mesa/examples/advanced/wolf_sheep/agents.py
+++ b/mesa/examples/advanced/wolf_sheep/agents.py
@@ -72,12 +72,12 @@ class Sheep(Animal):
 
     def move(self):
         """Move towards a cell where there isn't a wolf, and preferably with grown grass."""
-        neighbors = np.array(list(self.cell.neighborhood))
+        neighbors = self.cell.neighborhood
         coords = np.array([cell.coordinate for cell in neighbors])
         xs, ys = coords[:, 0], coords[:, 1]
 
-        has_wolf = np.asarray(self.model.grid.wolves.data)[xs, ys].astype(bool)
-        has_grass = np.asarray(self.model.grid.grass.data)[xs, ys].astype(bool)
+        has_wolf = self.model.grid.wolves[xs, ys].astype(bool)
+        has_grass = self.model.grid.grass[xs, ys].astype(bool)
 
         safe_cells = ~has_wolf
         safe_grass_cells = safe_cells & has_grass
@@ -88,7 +88,8 @@ class Sheep(Animal):
 
         # Move to a cell with grass if available, otherwise move to any safe cell
         target_mask = safe_grass_cells if safe_grass_cells.any() else safe_cells
-        self.cell = self.random.choice(neighbors[target_mask])
+        valid_cells = [cell for cell, valid in zip(neighbors, target_mask) if valid]
+        self.cell = self.random.choice(valid_cells)
 
 
 class Wolf(Animal):
@@ -123,6 +124,9 @@ class Wolf(Animal):
         # Mark the cell as occupied by a wolf
         self.model.grid.wolves.data[self.cell.coordinate] = True
 
+    def remove(self):
+        self.model.grid.wolves.data[self.cell.coordinate] = False
+        super().remove()
 
 class GrassPatch(FixedAgent):
     """A patch of grass that grows at a fixed rate and can be eaten by sheep."""
--- a/mesa/examples/advanced/wolf_sheep/agents.py
+++ b/mesa/examples/advanced/wolf_sheep/agents.py
@@ -1,3 +1,5 @@
+import numpy as np
+
 from mesa.discrete_space import CellAgent, FixedAgent
 
 
@@ -70,30 +72,23 @@ class Sheep(Animal):
 
     def move(self):
         """Move towards a cell where there isn't a wolf, and preferably with grown grass."""
-        cells_without_wolves = []
-        cells_with_grass = []
+        neighbors = np.array(list(self.cell.neighborhood))
+        coords = np.array([cell.coordinate for cell in neighbors])
+        xs, ys = coords[:, 0], coords[:, 1]
 
-        for cell in self.cell.neighborhood:
-            has_wolf = self.model.grid.wolves.data[self.cell.coordinate]
-            has_grass = self.model.grid.grass.data[self.cell.coordinate]
+        has_wolf = np.asarray(self.model.grid.wolves.data)[xs, ys].astype(bool)
+        has_grass = np.asarray(self.model.grid.grass.data)[xs, ys].astype(bool)
 
-            # Prefer cells without wolves
-            if not has_wolf:
-                cells_without_wolves.append(cell)
-
-                # Among safe cells, pick those with grown grass
-                if has_grass:
-                    cells_with_grass.append(cell)
+        safe_cells = ~has_wolf
+        safe_grass_cells = safe_cells & has_grass
 
         # If all surrounding cells have wolves, stay put
-        if len(cells_without_wolves) == 0:
+        if not safe_cells.any():
             return
 
         # Move to a cell with grass if available, otherwise move to any safe cell
-        target_cells = (
-            cells_with_grass if len(cells_with_grass) > 0 else cells_without_wolves
-        )
-        self.cell = self.random.choice(target_cells)
+        target_mask = (safe_grass_cells if safe_grass_cells.any() else safe_cells)
+        self.cell = self.random.choice(neighbors[target_mask])
 
 
 class Wolf(Animal):
@@ -127,7 +122,6 @@ class Wolf(Animal):
 
         # Mark the cell as occupied by a wolf
         self.model.grid.wolves.data[self.cell.coordinate] = True
-
 
 class GrassPatch(FixedAgent):
     """A patch of grass that grows at a fixed rate and can be eaten by sheep."""
--- a/mesa/examples/advanced/wolf_sheep/agents.py
+++ b/mesa/examples/advanced/wolf_sheep/agents.py
@@ -1,4 +1,4 @@
-from mesa.discrete_space import CellAgent, FixedAgent
+from mesa.discrete_space import CellAgent, FixedAgent, CellCollection
 
 
 class Animal(CellAgent):
@@ -36,6 +36,9 @@ class Animal(CellAgent):
     def feed(self):
         """Abstract method to be implemented by subclasses."""
 
+    def move(self):
+        """Abstract method to be implemented by subclasses."""
+
     def step(self):
         """Execute one step of the animal's behavior."""
         # Move to random neighboring cell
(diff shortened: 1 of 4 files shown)


mesa/examples/advanced/wolf_sheep/agents.py
```diff
@@ -71,16 +74,8 @@ def move(self):
         cells_with_grass = []
 
         for cell in self.cell.neighborhood:
```
quaquel: yoiu can create a neighborhood mask fully vectorizing this loop.

PietroMondini: i hadn't thought of this approach, i've updated the implementation to fully vectorize the neighborhood loop as you suggested.

mesa/examples/advanced/wolf_sheep/agents.py
```diff
@@ -61,49 +66,38 @@ def feed(self):
...
+        if grass_patch.is_fully_grown():
             self.energy += self.energy_from_food
             grass_patch.get_eaten()
 
     def move(self):
         """Move towards a cell where there isn't a wolf, and preferably with grown grass."""
-        cells_without_wolves = []
-        cells_with_grass = []
+        neighbors = np.array(list(self.cell.neighborhood))
+        coords = np.array([cell.coordinate for cell in neighbors])
```
quaquel: You can get a neighborhood mask from the space itself:

`self.model.grid.get_neighborhood_mask(self.cell.coordinate)`

PietroMondini: I've implemented it, but i'm getting huge delays in the computation time

   ```
 def move(self):
        """Move towards a cell where there isn't a wolf, and preferably with grown grass."""
        neighbors_mask = self.model.grid.get_neighborhood_mask(self.cell.coordinate, include_center=False)

        safe_cells = ~self.model.grid.wolves.astype(bool) & neighbors_mask
        safe_grass_cells = safe_cells & self.model.grid.grass.astype(bool)

        # If all surrounding cells have wolves, stay put
        if not safe_cells.any():
            return

        # Move to a cell with grass if available, otherwise move to any safe cell
        target_mask = (safe_grass_cells if safe_grass_cells.any() else safe_cells)

        safe_cells_coords = np.argwhere(target_mask)
        x, y = self.random.choice(safe_cells_coords)
        self.cell = self.model.grid[x, y]
```

File: mesa/examples/advanced/wolf_sheep/agents.py [hunk 3]
```diff
@@ -61,49 +66,38 @@ def feed(self):
...
+        xs, ys = coords[:, 0], coords[:, 1]
 
-        for cell in self.cell.neighborhood:
-            has_wolf = False
-            has_grass = False
+        has_wolf = np.asarray(self.model.grid.wolves.data)[xs, ys].astype(bool)
```
souro26: wolves.data and grass.data are already numpy arrays, so np.asarray() might not be necessary here.

mesa/examples/advanced/wolf_sheep/agents.py
```diff
@@ -61,49 +66,38 @@ def feed(self):
...
+        target_mask = safe_grass_cells if safe_grass_cells.any() else safe_cells
+        self.cell = self.random.choice(neighbors[target_mask])
 
 
 class Wolf(Animal):
     """A wolf that walks around, reproduces (asexually) and eats sheep."""
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.model.grid.wolves.data[self.cell.coordinate] = True
```
souro26: Here the wolves layer is set when a wolf is created, and it’s also updated in move(). I might be missing it, but where is the layer cleared when a wolf dies?

PietroMondini: You're right i have to override the Agent.remove() method, ty

mesa/examples/advanced/wolf_sheep/agents.py
PietroMondini: Yes, using just _grid.wolves_ it works, but _grid.wolves.data_ is a memoryview object. Ty for the feedback

mesa/examples/advanced/wolf_sheep/agents.py
PietroMondini: I think the major issue is that the method returns a full (width × height) boolean array for every agent on every step. Switching back to extracting only the 4 neighbor coordinates and indexing the PropertyLayer directly keeps the allocation at O(4) per agent per step regardless of grid size, which should be much more efficient. 

Do you have any suggestions on how to make the mask approach more efficient?

PietroMondini:
This is the best solution I've come up with so far, balancing the logic of the system with performance. By keeping **GrassPatch** as a behavior agent while moving the state to a **PropertyLayer**, encapsulation is preserve, grass regrowth scheduling stays where it belongs.
Happy to get feedback on this

mesa/examples/advanced/wolf_sheep/agents.py
```diff
@@ -61,49 +66,39 @@ def feed(self):
...
+        if grass_patch.is_fully_grown():
             self.energy += self.energy_from_food
             grass_patch.get_eaten()
 
     def move(self):
         """Move towards a cell where there isn't a wolf, and preferably with grown grass."""
-        cells_without_wolves = []
-        cells_with_grass = []
+        neighbors = self.cell.neighborhood
+        coords = np.array([cell.coordinate for cell in neighbors])
```
souro26: This replaces a simple neighbor check with array-based logic per agent. Since each agent only looks at a small fixed neighborhood, this increases work instead of reducing it. PropertyLayers help with fast indexed access, but the neighbor iteration itself can stay as a simple loop, reading values directly from the layer.

mesa/examples/advanced/wolf_sheep/agents.py
```diff
@@ -61,49 +66,39 @@ def feed(self):
...
+        xs, ys = coords[:, 0], coords[:, 1]
 
-        for cell in self.cell.neighborhood:
-            has_wolf = False
-            has_grass = False
+        has_wolf = self.model.grid.wolves[xs, ys].astype(bool)
```
souro26: This allocates arrays and applies masking for every agent at every step. Accessing the layer per neighbor will avoids this.

mesa/examples/advanced/wolf_sheep/agents.py
```diff
@@ -61,49 +66,39 @@ def feed(self):
...
+        valid_cells = [cell for cell, valid in zip(neighbors, target_mask) if valid]
+        self.cell = self.random.choice(valid_cells)
 
 
 class Wolf(Animal):
     """A wolf that walks around, reproduces (asexually) and eats sheep."""
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.model.grid.wolves.data[self.cell.coordinate] = True
```
souro26: State is stored in the grid, so correctness depends on updating it in every path. Missing an update will silently desynchronize the model state.

PietroMondini: Yep, it was only missing one update on the removal of a wolf, as you noted before, i've double checked but i think now the property layers are properly update where needed.

wolves:
- `__init__`: set to `True` on spawn 
- `move()`:  set to `False` before move,  set to `True` after 
- `remove()`:  set to `False` on death 

grass:
- `__init__`: set to `True` or 'False' based on the countdown value
- `regrow()`: set to `True` when it grows back
- `get_eaten()`: set to `False` when eaten

mesa/examples/advanced/wolf_sheep/agents.py
PietroMondini: I totally agree, building a coordinate array per agent per step added overhead without any benefit given the fixed small neighborhood size.

mesa/examples/advanced/wolf_sheep/agents.py
PietroMondini: That's actually what i'm going back to, i've tried different paths but they're all more expensive. Each neighbor is now checked with a direct scalar read (`neighbor.wolves`, `neighbor.grass`), so there's no intermediate array created at any point in `move()`.

souro26:
Can you please rebase the branch onto main? Right now there are several unrelated changes.

souro26:
Thanks for this, i noted one small thing, otherwise it looks good to me.

mesa/examples/advanced/wolf_sheep/agents.py
```diff
@@ -136,19 +142,23 @@ def __init__(self, model, countdown, grass_regrowth_time, cell):
...
 
         # Schedule initial growth if not fully grown
-        if not self.fully_grown:
+        if not self.cell.grass:
             self.model.schedule_event(self.regrow, after=countdown)
 
     def regrow(self):
         """Regrow the grass."""
-        self.fully_grown = True
+        self.cell.grass = True
```
souro26, line 155: state is now stored on cell.grass and cell.wolves and updated in multiple places. A helper method can avoid inconsistencies.

PietroMondini: The state is still stored in a single place, the PropertyLayer's NumPy array. I found out that when a layer is attached to the grid, Mesa automatically generates a property accessor on each cell with the same name, so `cell.wolves` directly reads/writes `grid.wolves.data[cell.coordinate]` under the hood.
Basically it's just a simpler and more concise way to access the same data.

PietroMondini: Thanks for pointing it out, it gave me the chance to clarify those changes
.

souro26: my bad, i didn't realize that. thanks for the clarification.

quaquel:
I am closing this PR. I prefer to keep wolf sheep as is for now.

PR #3523, opened 2026-03-12
Fix incorrect iterable snapshot in BatchContext.capture

satharyasin-hub:
Thanks for opening a PR! Please click the `Preview` tab and select a PR template:

- [🐛 Bug fix](?expand=1&template=bug.md)
- [🛠 Feature/enhancement](?expand=1&template=feature.md)

Feature/enhancement PRs require prior maintainer approval in an issue or discussion before they are accepted.

BatchContext.capture attempts to snapshot attribute values using list(),
which incorrectly converts non-list iterables such as strings and sets.

This patch restricts list copying to actual list instances, preserving
the original type for other values and preventing incorrect aggregation
comparisons during batching.

### DIFF
--- a/mesa/experimental/mesa_signals/batching.py
+++ b/mesa/experimental/mesa_signals/batching.py
@@ -185,7 +185,7 @@ class _BatchContext:
             # Only capture if not already snapshotted before mutation
             if name not in self._captured_values:
                 current_value = getattr(signal.owner, name, None)
-                
+
                 if isinstance(current_value, list):
                     self._captured_values[name] = list(current_value)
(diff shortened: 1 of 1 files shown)


quaquel:
The basic claim of this PR is not correct. There is an instance check, so only lists are copied. This is directly related to the fact that at the moment only SignallingList is available for mutable collections.

satharyasin-hub:
Thank you for the clarification. I understand now that mutable collections are handled through SignalingList, so the snapshot behavior is already safe. I appreciate the feedback.

PR #3370, opened 2026-02-24
Fix: Pin starlette<1.0 to unblock CI / solara tests

Nithurshen:
### Summary
Pins `starlette<0.45.0` in the project dependencies to resolve a breaking CI failure caused by the latest Starlette release.

### Bug / Issue
Closes #3369 

Starlette 0.45.0 entirely removed the deprecated `on_startup` parameter. `solara` currently relies on this parameter during server initialization, causing an immediate `TypeError` that crashes the `pytest` test collection step. 

### Implementation
* Added `starlette<0.45.0` to `pyproject.toml` optional dependencies. This is a temporary constraint to unblock Mesa's CI until `solara` updates its framework to use the modern `lifespan` context manager protocol.

### Testing
* Ran `pytest tests/` locally to ensure the test suite successfully initializes and runs without the Starlette initialization error.

PR #3524, opened 2026-03-12
cell capacity bugfix

### LINKED ISSUE #3505: capacity parameter in grid constructors not honored by most things

quaquel:
### Summary
This is a small update to Cell and Grid to ensure capacity is set properly. In Cell, we explicitly force keyword arguments. In Grid, we pass capacity as a keyword argument instead of as an argument. All other discrete spaces already handle this correctly.

### Bug / Issue
Closes #3505.

### Implementation
* Cell now enforces explicit use of keyword arguments
* Grid passes capacity as a keyword argument.

### Testing
An additional test is added to check that capacity is properly set on a cell in a (subclass) of Grid.

### DIFF
--- a/tests/discrete_space/test_discrete_space.py
+++ b/tests/discrete_space/test_discrete_space.py
@@ -88,7 +88,7 @@ def test_orthogonal_grid_neumann():
     grid = OrthogonalVonNeumannGrid(
         (width, height), torus=False, capacity=2, random=random.Random(42)
     )
-    assert grid._cells[(0, 0, 0)].capacity == 2
+    assert grid._cells[(0, 0)].capacity == 2
 
 
 def test_orthogonal_grid_neumann_3d():


quaquel:
examples are failing since this merges into 3.5 and various examples have already been updated to 4.0
pre is failing because of the solara issue that was merged (#3370).

PR #3528, opened 2026-03-13
Remove ContinuousSpaceAgent.pos property (Fixes #3399)

### LINKED ISSUE #3399: `ContinuousSpaceAgent.pos` assignment does not update position

satharyasin-hub:
This PR removes the deprecated `ContinuousSpaceAgent.pos` property.

Maintainer feedback in issue #3399 requested that `agent.pos` be fully removed rather than kept as an alias to `position`.

This change deletes the `pos` property and its setter from
`continuous_space_agents.py`, keeping `position` as the single source of truth for agent positions.

closes #3399.

### DIFF
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -6,7 +6,6 @@ import pytest
 from mesa.agent import Agent, AgentSet
 from mesa.experimental.devs.simulator import DEVSimulator
 from mesa.model import Model
-from mesa.time import Schedule
 
 
 def test_model_set_up():
@@ -77,7 +76,7 @@ def test_rng(rng=23):
     )
 
 
-def etest_reset_rng(newseed=42):
+def test_reset_rng(newseed=42):
     """Test resetting the random seed on the model."""
     model = Model(rng=5)
     old_rng = model._rng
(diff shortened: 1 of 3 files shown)


satharyasin-hub:
Thanks for the review and merge! I appreciate the feedback and guidance during the process.
Glad to contribute to Mesa and help remove the deprecated `ContinuousSpaceAgent.pos` property.
Looking forward to contributing more improvements.

PR #3531, opened 2026-03-13
Update benchmarks to use the new Scenario API

quaquel:
### Summary
Update benchmarks to use the new Scenario API, replacing scenario_class + parameters + seeds with a scenario instance and replications (used to be called seeds) /iterations (used to be called replications) keys. global_benchmark.py now uses spawn_replications() for scenario replications and removes stale code.

### Motive
With the new Scenario API, the code can be simplified, and it served as a nice test case for the new API.

### Implementation
* replace `seed` with `spawn_replications`
* Move parameters into the appropriate `Scenario` subclass
* Rename `replications` with `iterations`. Still not sure about this label. It is the number of repeats of a given scenario + replications. 
* removed some stale code and comments.

### DIFF
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -85,30 +85,6 @@ def test_rng(rng=23):
     )
 
 
-# def test_reset_rng(newseed=42):
-#     """Test resetting the random seed on the model."""
-#     model = Model(rng=5)
-#     old_rng = model._rng
-#
-#     model.reset_rng(rng=6)
-#     new_rng = model._rng
-#
-#     assert old_rng != new_rng
-#
-#     old_rng = new_rng
-#     model.reset_rng()
-#     new_rng = model.rng.bit_generator.state
-#
-#     assert old_rng == new_rng
-#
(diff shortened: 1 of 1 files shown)


tests/test_model.py
```diff
@@ -85,28 +85,28 @@ def test_rng(rng=23):
...
-
-    assert old_rng == new_rng
-
-    model = Model(rng=np.random.MT19937(42))
-    old_rng = model._rng
-    model.reset_rng()
-    new_rng = model.rng.bit_generator.state
-
-    assert np.all(old_rng["state"]["key"] == new_rng["state"]["key"])
+# def test_reset_rng(newseed=42):
```
EwoutH: Is this test not needed anymore? In that case, it can be removed. Or was this commented out for testing?

EwoutH (approved):

tests/test_model.py
quaquel: I thought this test was removed in #3493, but it wasn't.

PR #3433, opened 2026-03-04
Fix non-deterministic meta-agent membership assignment (#3184)

codebyNJ:
### Pre-PR Checklist
- [x] This PR is a bug fix, not a new feature or enhancement.

### Summary
Fixes #3184 — `agent.meta_agent` was set non-deterministically when an agent belongs to multiple meta-agents.

### Bug / Issue
Three locations in `meta_agent.py` blindly set `agent.meta_agent = self` without checking existing memberships:

- `MetaAgent.__init__` (line 315) — overwrites any existing assignment
- `add_constituting_agents` (line 391) — same issue
- `create_meta_agent` path 1 (line 246) — iterates over `a.meta_agents` (`set()`) which has non-deterministic order

Different runs produce different `agent.meta_agent` pointers for the same model, breaking reproducibility.

### Implementation
All three locations now pick the meta-agent with the **lowest `unique_id`**:

```python
agent.meta_agent = sorted(
    agent.meta_agents, key=lambda x: x.unique_id or 0
)[0]
```

Consistent with the existing pattern in `remove_constituting_agents` (lines 404-409).

### Testing
All 20 meta-agent tests pass. No new tests needed — fix makes existing behavior deterministic without changing semantics.

### Additional Notes
None.

---

### Context & motivation
I found this bug while reading through #3184 and tracing how overlapping membership works in meta-agents. `agent.meta_agent` was just being set to `self` everywhere without checking if the agent already belonged to another meta-agent. The pointer ends up depending on creation order, which is not predictable. Reproducibility matters for any simulation framework — same inputs should give same outputs.

### What I learned
The interesting part was that `remove_constituting_agents` already had the correct deterministic pattern (sorting by `unique_id`) but `__init__` and `add_constituting_agents` didn't follow it. I also learned that iterating over a Python `set()` is non-deterministic across runs, so even the `create_meta_agent` path needed sorting. The fix was about consistency — applying the same pattern that already existed in one place to the others that were missing it.

### Learning repo
My learning repo: https://github.com/codebyNJ/GSoC-learning-space

### Readiness checks
- [x] This PR addresses an agreed-upon problem (linked issue #3184)
- [x] I have read the [contributing guide](https://github.com/mesa/mesa/blob/main/CONTRIBUTING.md) and [deprecation policy](https://github.com/mesa/mesa/blob/main/CONTRIBUTING.md#deprecation-policy)
- [x] I have performed a self-review
- [x] Tests pass locally (`pytest --cov=mesa tests/`)
[...]

PR #3538, opened 2026-03-14
Add missing mesa-internal attributes to `mesa_primitives` blocklist in `create_meta_agent`

codebyNJ:
## Pre-PR Checklist

## Summary

Fixes `add_attributes` in `create_meta_agent` so existing meta-agent attributes are not overwritten by values inferred from constituting agents. Uses `hasattr()` checks to preserve existing attributes—both internal Mesa attributes and explicitly-provided ones.

## Bug / Issue

When `assume_constituting_agent_attributes=True`, the previous implementation unconditionally copied all public non-callable attributes from constituent agents onto the meta-agent, silently overwriting:

1. **Mesa internal attributes** (cell, current_action, meta_agent, meta_agents)
2. **Explicitly-passed meta_attributes** (as reported in #3472)

Expected behavior: don't overwrite attributes that already exist on the meta-agent.
Actual behavior: any existing attribute could be silently overwritten without warning.

Related context: issues #3433, #3506, #3472 all identified meta-agent attribute problems.

## Implementation

Updated `add_attributes` helper inside `create_meta_agent` in `mesa/experimental/meta_agents/meta_agent.py`:

- When building meta_attributes from agents: Check `not hasattr(meta_agent_instance, name)` before copying
- When applying meta_attributes to meta-agent: Check `if not hasattr(meta_agent_instance, key)` before calling `setattr()`

**Why this is better than a blocklist:**
- More robust and future-proof
- No manual blocklist maintenance needed as Mesa evolves
- Automatically protects any existing attribute (internal or explicit)
- Solves both attribute overwrite issues (#3538 and #3472) with one approach

## Testing

Added `test_assume_attributes_does_not_overwrite_existing_attributes`:
- Verifies explicit `meta_attributes` are preserved when adding agents with `assume_constituting_agent_attributes=True`
- Verifies missing attributes are still copied from constituents

## Additional Notes

This change keeps behavior predictable: existing attributes are preserved, inferred values only fill missing keys.

No API changes are required for current users.

This complements determinism-related fixes (#3433, #3506) and does not replace them.


tpike3:
As this is clearly a bug and there are some related issues as identified in #3472, instead of both #3538 and #3472 why dont we just ensure the meta-agent doesn't overwrite attributes it already has.

This approach 

```
 mesa_primitives = [
            "unique_id",
            "model",
            "pos",
            "name",
            "random",
            "rng",
            "cell",
            "current_action",
            "meta_agent",
            "meta_agents",
        ]
```

is very brittle, and so issues identified in #3472 and #3538, plus the current brittleness can all be solved with not overwriting existing meta-agent attributes.

falloficaruss:
> As this is clearly a bug and there are some related issues as identified in #3472, instead of both #3538 and #3472 why dont we just ensure the meta-agent doesn't overwrite attributes it already has.
> 
> This approach
> 
> ```
>  mesa_primitives = [
>             "unique_id",
>             "model",
>             "pos",
>             "name",
>             "random",
>             "rng",
>             "cell",
>             "current_action",
>             "meta_agent",
>             "meta_agents",
>         ]
> ```
> 
> is very brittle, and so issues identified in #3472 and #3538, plus the current brittleness can all be solved with not overwriting existing meta-agent attributes.

This is a better direction for the attribute side. It seems more robust than extending the mesa_primitives blocklist, and it also looks like it would cover both the internal-attribute corruption in this PR and the explicit-attribute overwrite case in #3472. I do think methods are still a separate question, since inferred method-name collisions are not addressed by the same rule, but for attributes this feels like the cleaner approach.

codebyNJ:
@tpike3 @falloficarus22 I have updated the PR.

tpike3 (changes requested):
Thanks @codebyNJ , this looks good to me, can you fix the test conflicts and then I will merge.

codebyNJ:
> Thanks @codebyNJ , this looks good to me, can you fix the test conflicts and then I will merge.

Welcome @tpike3. I have fixed the merge conflict issues in the test files.

tpike3 (approved):

tpike3:
@codebyNJ Could you look at the Warehouse Model failure. The warehouse model is the other meta-agent example so we need to understand why this code broke it. https://github.com/mesa/mesa-examples/tree/main/examples/warehouse

codebyNJ:
sure @tpike3 i will look into warehouse model and find out the reason why this code broke.

codebyNJ:
@tpike3 Fixed the warehouse failure. The issue: `hasattr()` check prevented explicit `meta_attributes` (like `cell`) from overriding class defaults. 

**Solution:** Always apply explicit meta_attributes (user intent takes precedence).

tpike3:
@codebyNJ   Thank you very much for the PR! Due to the major refactor design of meta-agents for Google Summer of Code, I am closing all meta-agent PRs. 

Thank you again for your excellent work and interest, we look forward to your continuing contributions.

PR #3539, opened 2026-03-14
docs: standardize run instructions across Mesa examples

DashamiJituri:
This PR standardizes the **"How to Run" instructions** across multiple Mesa example READMEs.

## Changes

* Removed installation steps from individual examples.
* Simplified run instructions to assume **Mesa is already installed**.
* Updated examples to use a consistent command:
         solara run app.py

Following maintainer feedback, this consolidates the documentation updates into a single PR and ensures consistent run instructions across examples.

### DIFF
--- a/mesa/examples/advanced/alliance_formation/Readme.md
+++ b/mesa/examples/advanced/alliance_formation/Readme.md
@@ -36,8 +36,13 @@ An example of the bilateral shapley value in another model:
 
 ## How to Run
 
-Assuming Mesa is already installed, run the example with:
+Install Mesa with recommended dependencies:
+
+pip install "mesa[rec]"
+
+Then run the example:
 
 solara run app.py
 
-Then open the displayed local URL in your browser.
+Open the displayed local URL in your browser.
+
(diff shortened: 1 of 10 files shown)


DashamiJituri:
> these READMEs now tell users only to assume “Mesa is already installed,” but the examples they document still depend on optional visualization/network extras that are not part of the base mesa install. For example, `boid_flockers/Readme.md` now omits any install guidance, while its app imports `mesa.visualization` and `matplotlib` in `boid_flockers/app.py`, and the package metadata still shows those under optional viz extras in `pyproject.toml`. The same problem exists for networked examples like `virus_on_network/Readme.md` versus `virus_on_network/app.py`. As written, a user who installed plain mesa can follow these new instructions and immediately hit missing-module errors for solara, matplotlib, or networkx.

Thanks for pointing this out — that's a good catch.

You're right that several examples rely on optional visualization or network dependencies (e.g., `solara`, `matplotlib`, `networkx`) which are not included in the base `mesa` installation. With the current wording, a user installing only `mesa` could encounter missing-module errors when running these examples.

To address this, I’ll update the **How to Run** sections to recommend installing Mesa with the appropriate extras (e.g., `pip install "mesa[rec]"`) so the visualization dependencies required by the examples are available before running `solara run app.py`.

I’ll push an update to the PR shortly to reflect this.

DashamiJituri:
Thank you for the review and for merging this! I really appreciate the feedback throughout the process — it helped me understand the documentation structure and dependency handling much better. Looking forward to contributing more to Mesa.

Issue #3536, opened 2026-03-14
GSoC contributors: PR requirements


EwoutH:
_This applies across all Mesa repos (mesa, mesa-examples, mesa-llm)._

### The situation
Mesa is maintained by a small group of volunteers. We want GSoC to be a genuinely valuable learning experience for you, and we want your contributions to make Mesa better. But we have limited capacity — every PR we review is time we're not spending on development, mentoring, or our own lives.

This is an experiment for us too. We're trying to find a process that maximizes your learning, produces quality contributions, and is sustainable for our maintainers. We need your help making it work. The single most important thing you can do: **invest in understanding before you invest in PRs.**

A few well-understood contributions are worth far more than many shallow ones — for you and for us. When you deeply understand a problem, your PR is easier to review, more likely to be merged, and you learn more from the process. When you don't, the review becomes a back-and-forth that drains everyone's time and energy.

_See also the [Google Summer of Code 2026](https://github.com/mesa/mesa/wiki/Google-Summer-of-Code-2026) guide._

### Step 1: Build models first
Before opening PRs, fork the [GSoC-learning-space](https://github.com/mesa/GSoC-learning-space) template repo and start building models. This is the foundation for everything else. See the template README for why this matters and how to use it.

Your learning repo is not graded and maintainers won't monitor it. But when you open a PR, you'll link to it, and it will be obvious whether you've done the work.

### Step 2: Engage in discussions
Read [open issues](https://github.com/mesa/mesa/issues), [discussions](https://github.com/mesa/mesa/discussions), and [recent release notes](https://github.com/mesa/mesa/releases). Before working on something, make sure there's agreement that it's worth doing — comment on the issue or discussion, or open a new one. Don't start coding before direction is clear. See our [contributing guide](https://github.com/mesa/mesa/blob/main/CONTRIBUTING.md#when-to-open-what) for when to open what.

If you're unsure whether something is a good use of your time, ask. We'd much rather spend five minutes aligning on an idea than three hours reviewing a PR that goes in the wrong direction.

### Step 3: PR requirements
Once you've built understanding and aligned on what to work on, here's what each PR needs.

#### Use the normal PR template
Fill out the existing bugfix or feature PR template as usual. Don't skip sections.

#### Add the GSoC checklist below your PR description
Copy the block below and paste it at the bottom of your PR description. Fill in the text fields and check the boxes.

~~~markdown
---

## GSoC contributor checklist

### Context & motivation
[...]

PR #3542, opened 2026-03-15
Enhancement #3541: Add Grid.available_cells and select_random_availab…

### LINKED ISSUE #3541: Enhancement: Add Grid.available_cells and select_random_available_cell() for partial-capacity queries
## Problem

`Grid` has `empties` and `select_random_empty_cell()`, but `cell.empty` 
flips to `False` the moment **any** agent enters a cell — even one with 
`capacity=5`. This makes the `capacity` parameter nearly impossible to 
use proactively:
```python
grid = OrthogonalMooreGrid((10, 10), capacity=3)
# Fill every cell with 2 agents...
# Now I want a cell that still has room. How?
# grid.empties → empty (no cells are fully empty)
# There is no built-in way to find cells where len(agents) < capacity
```

## Proposed Solution

Add two methods to `Grid`, mirroring the existing `empties` / 
`select_random_empty_cell()` pattern:
```python
@property
def available_cells(self) -> CellCollection[T]:
    """Cells where not cell.is_full (i.e. len(agents) < capacity)."""
    return self.all_cells.select(lambda cell: not cell.is_full)

def select_random_available_cell(self) -> T:
    """O(1) heuristic random pick from available_cells.
    Raises IndexError if all cells are at capacity."""
```

`Cell.is_full` already exists — this just exposes it at the grid level.

## Notes

- `capacity=None` cells are always available (is_full is always False), 
  so this degrades gracefully for unlimited grids.
- The heuristic (50 random attempts → explicit list fallback) matches 
  the existing `select_random_empty_cell` implementation exactly.
- I have a working implementation and full test suite ready if the 
  approach is agreed upon.
```

### PR DESCRIPTION
**Pull request by DipayanDasgupta**
## Summary

Closes #3541

Adds two new methods to `Grid` that make the `capacity` parameter 
usable proactively. The existing `empties` / `select_random_empty_cell()` 
API only tracks cells with zero agents. Once any agent enters a cell, 
`cell.empty` is `False` — even if `capacity=5` and 4 slots remain.

### `Grid.available_cells` (new property)
Returns a `CellCollection` of all cells where `not cell.is_full`.

### `Grid.select_random_available_cell()` (new method)
Uses the same two-phase O(1) heuristic as `select_random_empty_cell()`.
Raises `IndexError` with a clear message if all cells are at capacity.
```python
# Place agent in any non-full cell
agent.move_to(grid.select_random_available_cell())

# Count cells that still have room
len(list(grid.available_cells))
```

## Tests — `tests/discrete_space/test_grid_capacity.py`

27 tests parametrized across `OrthogonalMooreGrid`, 
`OrthogonalVonNeumannGrid`, and `HexGrid`:

| Section | Coverage |
|---|---|
| CellFullException | `move_to`, `.cell=`, `capacity=None`, vacate-and-refill |
| `available_cells` | Empty grid, full excluded, partial fill, dynamic update |
| `select_random_available_cell` | Non-full result, 30-sample consistency, `IndexError` |
| `Cell.is_full` | Unlimited, 0→1→2→1 transitions, direct `add_agent` |

---

### Context & motivation
While experimenting with different grid types and capacity settings I kept
running into the same limitation: `grid.empties` and `select_random_empty_cell()`
become useless the moment any agent steps into a cell — even when capacity is
explicitly set to 5 or 10. There was no built-in way to ask the grid "give me
a cell that still has free slots", which felt like a fundamental missing piece
for any model where agents share cells up to a limit. I opened issue #3541 and
implemented the fix in this PR.

### What I learned
`cell.empty` and `cell.is_full` answer two completely different questions, and
conflating them is a real source of bugs in capacity-bounded models. My first
implementation mirrored the `empty` property layer with a `full` layer — it
looked clean but profiling showed it added measurable overhead in move-heavy
models (numpy writes on every `add_agent` / `remove_agent` call). Dropping
the persistent layer and computing `not_full_cells` lazily on demand was both
simpler to reason about and faster. Good reminder that less state is often better.

### Learning repo
🔗 Relevant model(s): https://github.com/DipayanDasgupta/GSoC-learning-space/tree/main/models/boltzmann_wealth

### Readiness checks
- [done] This PR addresses an agreed-upon problem (linked issue #3541, maintainer-approved)
- [done] I have read the contributing guide and deprecation policy
- [] I have performed a self-review (comments posted in Files changed tab)
- [done ] Another GSoC contributor has reviewed this PR: @codebreaker32
- [done] Tests pass locally (`pytest tests/discrete_space/`)
- [doone] Code is formatted (`ruff check . --fix`)
- [done] Docstrings included inline; no separate docs needed for this internal API

### DIFF
--- a/mesa/discrete_space/grid.py
+++ b/mesa/discrete_space/grid.py
@@ -331,7 +331,7 @@ class Grid(DiscreteSpace[T]):
         Example::
 
             # Place an agent in any non-full cell
-            agent.move_to(grid.select_random_not_full_cell())
+            agent.move_to(grid.select_random_cell_with_capacity())
 
             # Count how many cells still have room
             len(list(grid.available_cells))
--- a/tests/discrete_space/test_discrete_space.py
+++ b/tests/discrete_space/test_discrete_space.py
@@ -1426,7 +1426,7 @@ def test_select_random_available_cell_not_full(factory) -> None:
     for cell in grid._celllist[:half]:
         for _ in range(2):
             make_agent(model).move_to(cell)
-    chosen = grid.select_random_not_full_cell()
+    chosen = grid.select_random_cell_with_capacity()
     assert not chosen.is_full
 
 
@@ -1438,7 +1438,7 @@ def test_select_random_available_cell_raises_when_all_full(factory) -> None:
     for cell in grid._celllist:
         make_agent(model).move_to(cell)
     with pytest.raises(IndexError, match="No available cells"):
-        grid.select_random_not_full_cell()
+        grid.select_random_cell_with_capacity()
 
 
 @pytest.mark.parametrize("factory", GRID_FACTORIES)
@@ -1451,7 +1451,7 @@ def test_select_random_available_cell_consistent_with_available_cells(factory) -
             make_agent(model).move_to(cell)
     available_set = set(grid.cells_with_capacity)
     for _ in range(30):
-        chosen = grid.select_random_not_full_cell()
+        chosen = grid.select_random_cell_with_capacity()
         assert chosen in available_set
 
 


quaquel:
I like the idea of this PR, but the performance hit is less than ideal. Any idea what explains this and how we might implement this differently to avoid the performance hit?

DipayanDasgupta:
@quaquel — investigated locally. The regression was caused by the original full property layer implementation, which wrote to a numpy array on every add_agent and remove_agent call. In models where agents move frequently, that's 2 extra numpy writes per agent per step — directly in the hot path.
That layer has been removed entirely. not_full_cells now computes lazily via select(lambda cell: not cell.is_full) with no persistent state. select_random_not_full_cell retains the 50-sample O(1) fast path for the common case; the fallback builds an explicit list only when needed.
Benchmarked locally on a 20×20 grid with 500 agents running 200 steps:

main: 0.117s
this branch: 0.107s (8.5% faster than main, no regression)

The current implementation adds no overhead to the move path.

DipayanDasgupta:
@quaquel @EwoutH  What are the next steps you'd recommend to take to get this into a mergeable position , any ideas , leads?

EwoutH:
@DipayanDasgupta could you go through our GSoC PR requirements and check if it follows all, doing a proper self-review?

- https://github.com/mesa/mesa/issues/3536

After that, @codebreaker32 are you willing to review this?

codebreaker32:
> After that, @codebreaker32 are you willing to review this?

Sure I'll do it :)

DipayanDasgupta:
> @DipayanDasgupta could you go through our GSoC PR requirements and check if it follows all, doing a proper self-review?
> 
> * [GSoC contributors: PR requirements #3536](https://github.com/mesa/mesa/issues/3536)
> 
> After that, @codebreaker32 are you willing to review this?

sure @EwoutH  will do

DipayanDasgupta:
@EwoutH @codebreaker32  I am new to the GSOC readiness check guide but I cloned the learning repo and tried to do the steps laid out , please give me the feedback for the same so I can better meet your requirements in the next PR. Thanks for your time. I have updated the PR description.

EwoutH:
Thanks, you're effort is appreciated!

The purpose of the learning repo is to serve as a tool for you to develop your own skills. An added benefit of you doing that in public is that we can have some context if we need to.

docs/tutorials/0_first_model.ipynb
codebreaker32, line 1: Please revert changes of this file

tests/discrete_space/test_grid_capacity.py
codebreaker32, line 1: Add test under the respective test file, here `test_discrete_space`

mesa/discrete_space/grid.py
```diff
@@ -308,6 +309,77 @@ def select_random_empty_cell(self) -> T:  # noqa
...
+        Uses the same two-phase heuristic as :meth:`select_random_empty_cell`:
+
+        1. **Fast path** — attempt up to 50 random samples from the cell list
+           (O(1) average when the grid is not densely occupied).
+        2. **Fallback** — if the fast path fails, build an explicit list of
+           available cells and sample from that (O(n) worst case).
+
+        This mirrors the performance characteristics documented in the
+        ``select_random_empty_cell`` implementation (see Agents.jl discussion
+        at https://github.com/JuliaDynamics/Agents.jl/pull/541).
```
codebreaker32: Keep these as inline comments

mesa/discrete_space/grid.py
```diff
@@ -308,6 +309,77 @@ def select_random_empty_cell(self) -> T:  # noqa
...
+
+        Example::
+
+            # Place an agent in any non-full cell
+            agent.move_to(grid.select_random_available_cell())
+
+            # Count how many cells still have room
+            len(list(grid.available_cells))
+        """
+        return self.all_cells.select(lambda cell: not cell.is_full)
```
codebreaker32, line 341:
```suggestion
        if self.capacity is None:
            return self.all_cells
        return self.all_cells.select(lambda cell: not cell.is_full)
```

mesa/discrete_space/grid.py
```diff
@@ -308,6 +309,77 @@ def select_random_empty_cell(self) -> T:  # noqa
...
+            T: A randomly chosen cell where ``not cell.is_full``.
+
+        Raises:
+            IndexError: If every cell in the grid is at full capacity.
+
+        Example::
+
+            # Safe placement that respects capacity limits
+            free_cell = grid.select_random_available_cell()
+            agent.move_to(free_cell)
```
codebreaker32, line 353: Use the correct name

mesa/discrete_space/grid.py
```diff
@@ -308,6 +309,77 @@ def select_random_empty_cell(self) -> T:  # noqa
...
+           (O(1) average when the grid is not densely occupied).
+        2. **Fallback** — if the fast path fails, build an explicit list of
+           available cells and sample from that (O(n) worst case).
+
+        This mirrors the performance characteristics documented in the
+        ``select_random_empty_cell`` implementation (see Agents.jl discussion
+        at https://github.com/JuliaDynamics/Agents.jl/pull/541).
+
+        Returns:
+            T: A randomly chosen cell where ``not cell.is_full``.
```
codebreaker32: Use Cell instead of T

codebreaker32:
Looks fine to me code wise with some small suggestions and benchmarks are also stable

codebreaker32:
I went through the original issue where @quaquel suggested
>Implementation wise, I would suggest adding a full propertly layer analogous to the empty property layer.

I'd also suggest this for better fallback performance

And are we good to go with "not_full_cell", I personally don't prefer because it is considered an anti-pattern for properties in python.

DipayanDasgupta:
@EwoutH — all @codebreaker32 review points have been addressed. The diff is now clean: only grid.py and test_discrete_space.py changed. Benchmarks are stable. Ready for your final review.

DipayanDasgupta:
> I went through the original issue where @quaquel suggested
> 
> > Implementation wise, I would suggest adding a full propertly layer analogous to the empty property layer.
> 
> I'd also suggest this for better fallback performance
> 
> And are we good to go with "not_full_cell", I personally don't prefer because it is considered an anti-pattern for properties in python.

@codebreaker32 — re the full property layer: quaquel suggested it before we had benchmark data. After implementing it, the CI benchmarks showed +13.5% run time regression on BoltzmannWealth large (every add_agent/remove_agent call writes to a numpy array — directly in the move hot path). I removed it and the latest benchmarks are all within noise. The lazy select() approach is both simpler and faster. Happy to defer to @quaquel on this if he still prefers the layer after seeing the numbers.

DipayanDasgupta:
@codebreaker32 — good catch on the naming. Renamed not_full_cells → cells_with_capacity which reads positively and clearly. On the property layer: the benchmark data already in this PR shows it causes measurable regression — leaving it as lazy computation. @quaquel  @EwoutH happy to hear your view on both naming and the layer decision given the numbers.

codebreaker32:
> After implementing it, the CI benchmarks showed +13.5% run time regression on BoltzmannWealth large (every add_agent/remove_agent call writes to a numpy array — directly in the move hot path). I removed it and the latest benchmarks are all within nois

I saw your previous commits but I couldn't find the property_layer implementation for this, Could you point it out?

DipayanDasgupta:
@codebreaker32 here's the full property layer implementation from commit 038fc0e: https://github.com/DipayanDasgupta/mesa/commit/038fc0e
The key overhead was in cell.py self.full = self.is_full was called at the end of both add_agent and remove_agent. Since is_full computes len(self._agents) >= self.capacity, and then the result is written to the numpy property layer via the full setter, every single agent move triggered 2 numpy array writes. In BoltzmannWealth large where agents move every step, that's the worst possible place to add work.The fallback in select_random_not_full_cell also used np.argwhere(property_layers["full"] == False) which scans the entire array, no faster than the current list comprehension, and adds the cost of maintaining the layer in exchange for nothing. Current approach has zero per-move overhead. The property layer would only make sense if cells_with_capacity were called extremely frequently in a tight loop, which isn't the typical use case.

mesa/discrete_space/grid.py
```diff
@@ -308,6 +309,70 @@ def select_random_empty_cell(self) -> T:  # noqa
...
+            # Place an agent in any non-full cell
+            agent.move_to(grid.select_random_not_full_cell())
+
+            # Count how many cells still have room
+            len(list(grid.available_cells))
+        """
+        if self.capacity is None:
+            return self.all_cells
+        if self.capacity is None:
+            return self.all_cells
```
quaquel: why twice?

mesa/discrete_space/grid.py
```diff
@@ -308,6 +309,70 @@ def select_random_empty_cell(self) -> T:  # noqa
...
+        return self.all_cells.select(lambda cell: not cell.is_full)
```
quaquel, line 341: it should be possible to do this simpler? we can do the negation on the entire layer and use the indices of this to get the cells. Sometihing roughly like

```python
[self._cells[i, j] for i,j in np.argwhere(not self.is_full)]
```

Not sure what is faster, but might be worth testing.

mesa/discrete_space/grid.py
```diff
@@ -308,6 +309,70 @@ def select_random_empty_cell(self) -> T:  # noqa
...
+        # Fallback: build explicit list, O(n) worst case (mirrors select_random_empty_cell
+        # heuristic, see https://github.com/JuliaDynamics/Agents.jl/pull/541).
+
+        if self._try_random:
+            for _ in range(50):
+                cell = random.choice(cells)
+                if not cell.is_full:
+                    return cell
+
+        not_full = [cell for cell in cells if not cell.is_full]
```
quaquel: why not use the cells_with_capacity property here?

mesa/discrete_space/grid.py
quaquel: this might not be possilby given that you removed the property layer for runtime reasons.

mesa/discrete_space/grid.py
codebreaker32: I think he pushed it once by himself and then committed the suggestion from GitHub UI(https://github.com/mesa/mesa/pull/3542/changes/09b0902f92062fc36aafa25383792a8ce4f01d67) as well

DipayanDasgupta: Fixed — I had pushed the same check manually and then accidentally committed the GitHub UI suggestion on top, creating the duplicate. Removed in latest commit.

DipayanDasgupta:
@quaquel fixed both removed the duplicate if self.capacity is None block (caused by committing the GitHub suggestion on top of my own identical change), and updated the fallback in select_random_not_full_cell to use self.cells_with_capacity directly. The stale  comment from codebreaker32 is already fixed the example now shows select_random_not_full_cell().

mesa/discrete_space/grid.py
DipayanDasgupta: Updated in latest commit the fallback now delegates to list(self.cells_with_capacity) directly.

mesa/discrete_space/grid.py
DipayanDasgupta: @quaquel @codebreaker32  — agreed that without the property layer the numpy approach isn't possible. The current implementation uses self.cells_with_capacity for both the fallback and the property, keeping things consistent. The fast path (50 random samples) handles the common case with zero overhead

mesa/discrete_space/grid.py
```diff
@@ -308,6 +309,68 @@ def select_random_empty_cell(self) -> T:  # noqa
...
+            agent.move_to(grid.select_random_not_full_cell())
+
+            # Count how many cells still have room
+            len(list(grid.available_cells))
+        """
+        if self.capacity is None:
+            return self.all_cells
+        return self.all_cells.select(lambda cell: not cell.is_full)
+
+    def select_random_not_full_cell(self) -> Cell:
```
```suggestion
    def select_random_cell_with_capacity(self) -> Cell:
```

tests/discrete_space/test_discrete_space.py
```diff
@@ -1254,6 +1255,236 @@ def test_network_missing_layout_node():
...
+@pytest.mark.parametrize("factory", GRID_FACTORIES)
+def test_select_random_available_cell_not_full(factory) -> None:
+    """Returned cell must have remaining capacity."""
+    model = make_model()
+    grid = factory(2)
+    half = len(grid._celllist) // 2
+    for cell in grid._celllist[:half]:
+        for _ in range(2):
+            make_agent(model).move_to(cell)
+    chosen = grid.select_random_not_full_cell()
```
quaquel:
```suggestion
    chosen = grid.select_random_cell_with_capacity()
```

tests/discrete_space/test_discrete_space.py
```diff
@@ -1254,6 +1255,236 @@ def test_network_missing_layout_node():
...
+
+@pytest.mark.parametrize("factory", GRID_FACTORIES)
+def test_select_random_available_cell_raises_when_all_full(factory) -> None:
+    """IndexError must be raised when every cell is at capacity."""
+    model = make_model()
+    grid = factory(1)
+    for cell in grid._celllist:
+        make_agent(model).move_to(cell)
+    with pytest.raises(IndexError, match="No available cells"):
+        grid.select_random_not_full_cell()
```
quaquel:
```suggestion
        grid.select_random_cell_with_capacity()
```

tests/discrete_space/test_discrete_space.py
```diff
@@ -1254,6 +1255,236 @@ def test_network_missing_layout_node():
...
+def test_select_random_available_cell_consistent_with_available_cells(factory) -> None:
+    """Every result from select_random_available_cell must be in available_cells."""
+    model = make_model()
+    grid = factory(3)
+    for cell in grid._celllist[::3]:
+        for _ in range(2):
+            make_agent(model).move_to(cell)
+    available_set = set(grid.cells_with_capacity)
+    for _ in range(30):
+        chosen = grid.select_random_not_full_cell()
```
quaquel:
```suggestion
        chosen = grid.select_random_cell_with_capacity()
```

mesa/discrete_space/grid.py
```diff
@@ -308,6 +309,68 @@ def select_random_empty_cell(self) -> T:  # noqa
...
+        available regardless of how many agents it holds.
+
+        Returns:
+            CellCollection[T]: All cells where ``len(agents) < capacity``
+            (or all cells when capacity is None).
+
+        Example::
+
+            # Place an agent in any non-full cell
+            agent.move_to(grid.select_random_not_full_cell())
```
```suggestion
            agent.move_to(grid.select_random_cell_with_capacity())
```

PR #3544, opened 2026-03-15
Fix #3543: VoronoiGrid no longer overwrites explicit capacity with area function

### LINKED ISSUE #3543: Bug: VoronoiGrid silently overwrites user-provided capacity with area-derived value
## Bug

`VoronoiGrid._build_cell_polygons` unconditionally overwrites every 
cell's `capacity` with a value derived from polygon area via 
`capacity_function`, silently discarding any explicit `capacity` 
argument the user passed to `__init__`.

## Reproduction
```python
from mesa.discrete_space.voronoi import VoronoiGrid

points = [[0,0],[1,0],[0.5,1],[1.5,1],[0,2],[1,2]]
grid = VoronoiGrid(points, capacity=1)

# Expected: cell.capacity == 1 for all cells
# Actual:   cell.capacity == round_float(polygon_area) — ignores capacity=1

for cell in grid._cells.values():
    print(cell.capacity)  # prints area-derived int, NOT 1
```

As a result, `CellFullException` is never raised for VoronoiGrid even 
when an explicit capacity is set, because the user's value is overwritten 
before any agents are placed.

## Root Cause

In `_build_cell_polygons` (last line of the method):
```python
# Current — always overwrites:
self._cells[region].capacity = self.capacity_function(polygon_area)

# Fix — only apply area function when user didn't provide a capacity:
if self.capacity is None:
    self._cells[region].capacity = self.capacity_function(polygon_area)
```

## Fix

Single conditional guard. When `capacity` is explicitly provided, 
preserve it. When `capacity=None` (the default), continue using 
`capacity_function` as before — no change to existing behaviour.

I have a working fix and tests ready if confirmed as a bug.

DipayanDasgupta:
## Summary

Closes #3543

`VoronoiGrid._build_cell_polygons` was unconditionally overwriting every
cell's `capacity` with a value derived from polygon area via
`capacity_function`, silently discarding any explicit `capacity` argument
passed by the user to `__init__`.

## Root cause
```python
# Before — always overwrites, even when user passed capacity=1:
self._cells[region].capacity = self.capacity_function(polygon_area)
```

## Fix

Single conditional guard — one line change:
```python
if self.capacity is None:
    self._cells[region].capacity = self.capacity_function(polygon_area)
```

When `capacity` is explicitly provided, it is preserved. When
`capacity=None` (the default), `capacity_function` is applied as before —
no change to existing behaviour.

## Tests

Added `tests/discrete_space/test_voronoi_capacity.py` with 5 tests:

- Explicit `capacity=5` preserved across all cells
- `capacity=None` still produces area-derived int capacities
- `CellFullException` fires at runtime with explicit `capacity=1`
- Custom `capacity_function` is ignored when capacity is explicit
- Default `round_float` produces non-negative int capacities

### DIFF
--- a/tests/discrete_space/test_discrete_space.py
+++ b/tests/discrete_space/test_discrete_space.py
@@ -1254,56 +1254,42 @@ def test_network_missing_layout_node():
         Network(g, layout=partial_layout, random=rng)
 
 
-# ---------------------------------------------------------------------------
-# VoronoiGrid capacity fix — Issue #3543
-# ---------------------------------------------------------------------------
-
-
 def test_voronoi_default_no_capacity() -> None:
-    """Default VoronoiGrid (no capacity, no function) leaves cells with None capacity."""
+    """Default VoronoiGrid (capacity=None) leaves cells with None capacity."""
     centroids = [[0.0, 0.0], [1.0, 0.0], [0.5, 1.0], [1.5, 1.0], [0.0, 2.0], [1.0, 2.0]]
     grid = VoronoiGrid(centroids, random=random.Random(42))
     for cell in grid._cells.values():
         assert cell.capacity is None
 
 
-def test_voronoi_explicit_capacity_is_preserved() -> None:
-    """Explicit capacity must not be overwritten by the area function."""
+def test_voronoi_int_capacity_applied_to_all_cells() -> None:
+    """Integer capacity is applied uniformly to every cell."""
     centroids = [[0.0, 0.0], [1.0, 0.0], [0.5, 1.0], [1.5, 1.0], [0.0, 2.0], [1.0, 2.0]]
     grid = VoronoiGrid(centroids, capacity=5, random=random.Random(42))
     for cell in grid._cells.values():
         assert cell.capacity == 5
 
 
-def test_voronoi_capacity_function_used_when_no_capacity() -> None:
-    """capacity_function is applied per-cell when capacity=None."""
+def test_voronoi_callable_capacity_derives_from_area() -> None:
+    """A callable capacity is called per-cell with polygon area and must return int."""
     centroids = [[0.0, 0.0], [1.0, 0.0], [0.5, 1.0], [1.5, 1.0], [0.0, 2.0], [1.0, 2.0]]
-    grid = VoronoiGrid(
-        centroids,
-        capacity=None,
-        capacity_function=round_float,
-        random=random.Random(42),
-    )
+    grid = VoronoiGrid(centroids, capacity=round_float, random=random.Random(42))
     for cell in grid._cells.values():
         assert cell.capacity is not None
         assert isinstance(cell.capacity, int)
         assert cell.capacity >= 0
 
 
-def test_voronoi_raises_when_both_capacity_and_function_provided() -> None:
-    """Providing both capacity and capacity_function must raise ValueError."""
+def test_voronoi_callable_capacity_custom_function() -> None:
+    """A custom callable capacity is applied correctly."""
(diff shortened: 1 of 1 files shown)


mesa/discrete_space/voronoi.py
```diff
@@ -293,4 +293,8 @@ def _build_cell_polygons(self):
             self._cells[region].properties["polygon"] = polygon
             polygon_area = self._compute_polygon_area(polygon)
             self._cells[region].properties["area"] = polygon_area
-            self._cells[region].capacity = self.capacity_function(polygon_area)
+            # Fix: only derive capacity from area when the user did NOT supply one.
```
quaquel: I disagree with this fix. None can also mean that there is no capacity. Rather, use capacity if provided or None. Only switch to capacity function if specified and capacity is None. If both are provided raise an appropriate exception.

DipayanDasgupta:
Thanks for the clarification @quaquel — updated. Now: uses capacity directly if provided, applies capacity_function only when capacity=None, and raises ValueError if both a non-default capacity_function and an explicit capacity are passed simultaneously.

mesa/discrete_space/voronoi.py
```diff
@@ -228,6 +228,12 @@ def __init__(
         self.voronoi_coordinates = None
         self.capacity_function = capacity_function
 
+        if capacity is not None and capacity_function is not round_float:
```
quaquel: This is not correct, I now cannot pass my own callable. 

Also, I am wondering whether we should change the default behavior of this class. That is, in `__init__`, change capacity_function to be `None`. The default round_found function makes no sense to me.

quaquel:
1. Do not add new test files; find the relevant tests in the appropriate file and update/expand those.
2. Tests are currently failing. Please ensure all tests pass.

DipayanDasgupta:
@quaquel 

Fixed AttributeError by moving `self.capacity_function = ...` before `_validate_parameters()`
 Replaced `callable | None` → `Callable | None` (fixes windows 3.14 CI)
Moved all Voronoi capacity tests into `test_discrete_space.py` (deleted separate file)
Default is now `capacity_function: Callable | None = None`
Ruff + pre-commit clean
Ready for another look whenever you have time, thank you for your reviews and feedback.

mesa/discrete_space/voronoi.py
```diff
@@ -188,7 +188,7 @@ def __init__(
         capacity: float | None = None,
         random: Random | None = None,
         cell_klass: type[Cell] = Cell,
-        capacity_function: callable = round_float,
+        capacity_function: Callable | None = None,
```
quaquel: sorry, one more thought. What about simplifying it even further. Let capacity take an interger, a callable or None and just drop the function. Makes stuff even simpler. But does require an update of the docstring.

tests/discrete_space/test_discrete_space.py
```diff
@@ -1251,3 +1252,64 @@ def test_network_missing_layout_node():
         SpaceException, match="is missing from the provided layout dictionary"
     ):
         Network(g, layout=partial_layout, random=rng)
+
+
+# ---------------------------------------------------------------------------
+# VoronoiGrid capacity fix — Issue #3543
+# ---------------------------------------------------------------------------
```
quaquel:
```suggestion
```

just remove this.

DipayanDasgupta: Done

DipayanDasgupta:
@quaquel — simplified as suggested. capacity now accepts int | Callable | None: pass an int for fixed per-cell limit, a callable for area-derived limits, or None for unlimited. capacity_function parameter removed entirely. Docstring updated.

PR #3545, opened 2026-03-15
add rescale helper to scenario 

souro26:
### Approval Link
https://github.com/mesa/mesa/discussions/3483#discussioncomment-16131083

### Summary
Add a small `rescale(samples, ranges)` helper for mapping samples from [0,1] to user defined parameter ranges.

Discussed in #3483 

### Motive
When generating experiments via Scenario.from_ndarray, users often sample parameters on [0,1] and then need to transform them into meaningful parameter ranges. This helper provides that transformation in a small, vectorized utility.

### Implementation
Adds a vectorized rescale function in scenario.py. Includes basic validation and tests covering scaling correctness, dimension mismatches, and edge cases.

### Usage Examples
````python
samples = qmc.LatinHypercube(d=3).random(100)

ranges = np.array([
    [0, 10],
    [5, 20],
    [-1, 1]
])

scaled = rescale(samples, ranges)

scenarios = Scenario.from_ndarray(scaled, ["a", "b", "c"], rng=42)
````
---

### Context & motivation
In #3483 , @quaquel suggested adding a small helper to rescale QMC samples from [0,1] to parameter ranges. Since I was already participating in that discussion around the experiment workflow and scenarios, I picked this up and implemented the helper so sampled parameters can be directly passed into `Scenario.from_ndarray`.

### What I learned
Most of the learning here actually came from the discussion rather than the implementation. Reading through the design ideas around scenarios, sampling, and experiment execution helped me understand how these pieces are expected to fit together in the new workflow.

### Readiness checks
- [x] This PR addresses an agreed-upon problem
- [ ] Another GSoC contributor has reviewed this PR: @

### DIFF
--- a/mesa/experimental/scenarios/scenario.py
+++ b/mesa/experimental/scenarios/scenario.py
@@ -21,8 +21,7 @@ def rescale_samples(
     *,
     inplace: bool = False,
 ) -> np.ndarray:
-    """
-    Rescale samples from the unit interval [0, 1] to parameter ranges.
+    """Rescale samples from the unit interval [0, 1] to parameter ranges.
 
     Parameters
     ----------
@@ -35,17 +34,16 @@ def rescale_samples(
         If False (default), a new array containing the rescaled samples
         is returned.
 
-    Returns
+    Returns:
     -------
     ndarray (n, d)
         Rescaled samples.
 
-    Notes
+    Notes:
     -----
     The rescaling is performed using NumPy broadcasting. If ``inplace=True``,
     the original ``samples`` array is overwritten.
     """
-
     samples = np.asarray(samples)
     ranges = np.asarray(ranges)
 
--- a/tests/experimental/test_scenarios.py
+++ b/tests/experimental/test_scenarios.py
@@ -309,7 +309,7 @@ def test_rescale_bounds_mapping():
 
     expected = np.array([[5, 2]])
     assert np.allclose(scaled, expected)
-    
+
 
 def test_rescale_inplace():
     """Check that inplace=True modifies the original array."""

### CODE REVIEW

**Comment by EwoutH** [comment]
Thanks for the checklist. Can you do a self-review though the GitHub UI?

**Comment by souro26** [comment]
Yes I will do it right away

File: mesa/experimental/scenarios/scenario.py [hunk 1]
```diff
@@ -15,6 +15,38 @@
 RNGLike = np.random.Generator | np.random.BitGenerator
 
 
+def rescale(samples: np.ndarray, ranges: np.ndarray) -> np.ndarray:
```
souro26: I kept this as a small standalone helper instead of adding it to Scenario because the rescaling happens before scenarios are created. It’s just a preprocessing step for the sample matrix.

mesa/experimental/scenarios/scenario.py
```diff
@@ -15,6 +15,38 @@
...
+            Parameter ranges as [[min, max], ...].
+
+    Returns:
+        ndarray (n, d)
+            Rescaled samples.
+    """
+    samples = np.asarray(samples)
+    ranges = np.asarray(ranges)
+
+    if samples.ndim != 2:
```
souro26: I added a few simple checks here to catch common mistakes like mismatched dimensions between the samples and the ranges.

mesa/experimental/scenarios/scenario.py
```diff
@@ -15,6 +15,38 @@
...
+    if ranges.ndim != 2 or ranges.shape[1] != 2:
+        raise ValueError("ranges must have shape (d,2)")
+
+    if samples.shape[1] != ranges.shape[0]:
+        raise ValueError("dimension mismatch between samples and ranges")
+
+    mins = ranges[:, 0]
+    scale = ranges[:, 1] - mins
+
+    return samples * scale + mins
```
souro26, line 58: This ues numpy broadcasting so the whole matrix can be scaled at once instead of looping over rows or columns.

tests/experimental/test_scenarios.py
```diff
@@ -220,3 +221,109 @@ def test_scenario_from():
     # check if parameter names matches number of columns in numpy array
     with pytest.raises(ValueError):
         Scenario.from_ndarray(samples, parameter_names=[], rng=42)
+
+
+def test_rescale_basic():
```
souro26, line 226: I added tests for normal scaling, negative ranges, invalid inputs, and larger sample matrices to make sure the helper behaves as expected.

mesa/experimental/scenarios/scenario.py
quaquel: I agree with this choice. 

```suggestion
def rescale_samples(samples: np.ndarray, ranges: np.ndarray) -> np.ndarray:
```

mesa/experimental/scenarios/scenario.py
quaquel: Can you motivate all these checks and what exception to do you get if you don't have the checks and have a wrongly shaped parameter? 

I am in general, not in favour of extensive input validation. Here, it is even less relevant because this is called early on just after sampling. So, a small error does not cost you a lot of time.

mesa/experimental/scenarios/scenario.py
quaquel: This creates a new numpy array. What about making it an in-place operation that overwrites the original numpy array? If we do this, we should document it clearly..... Alternatively, we might have a kwarg for controlling this.

mesa/experimental/scenarios/scenario.py
souro26: I originally added those checks to catch obvious shape mistakes early and give clearer error messages. But numpy broadcasting already raises errors immediately if the shapes are incompatible, so I'll remove the explicit checks.

mesa/experimental/scenarios/scenario.py
souro26: The current version returns a new array. The original samples are unchanged, which avoids unexpected mutation of the input array, but it makes another array. Making it always in-place would avoid the extra allocation but would overwrite the original samples.

I think the keyword argument you suggested would allow both behaviors: the default could return a new array, while something like inplace=True would scale the array directly when users want to avoid the extra allocation.

tests/experimental/test_scenarios.py
```diff
@@ -220,3 +221,101 @@ def test_scenario_from():
...
+    samples = np.array([[0.0, 1.0]])
+    ranges = np.array([[5, 10], [-2, 2]])
+
+    scaled = rescale_samples(samples, ranges)
+
+    expected = np.array([[5, 2]])
+    assert np.allclose(scaled, expected)
+
+
+def test_rescale_inplace():
```
PietroMondini, line 314: If you want to be extra, you could also add a complementary test asserting that inplace=False (default) does not modify the original array.

EwoutH:
Thanks for the PR and the reviews.

I must say, as I see this now this is a very narrow feature that's relatively easy to do it yourself if you need to. But maybe I'm missing something. What's the use case and audience?

And could have it been build in `from_ndarray()`, since that seems to be the primary use case?

quaquel:
@EwoutH, please check #3483.

EwoutH:
This one right?

- https://github.com/mesa/mesa/discussions/3483#discussioncomment-16131083

Sorry, so many stuff going on.

But it seems it was well thought out.

PR #3550, opened 2026-03-15
Replace np.vstack with geometric array growth in ContinuousSpace

### LINKED ISSUE #3548: Replace `np.vstack` in `ContinuousSpace` by array growth approach inspired from `NumpyAgentDataset`
In Continuous Space, we currently hardcodes a 20% growth fraction inside its `_add_agent` method.

https://github.com/mesa/mesa/blob/2b23c16b7f13f3fdcf9c11c90853d859d1acedec/mesa/experimental/continuous_space/continuous_space.py#L116-L127

However, `NumpyAgentDataSet` uses geometric doubling, controlled by class-level constants,
It then uses `growth = max(int(current_size * (self._GROWTH_FACTOR - 1)), self._MIN_GROWTH)` to calculate the new size.

https://github.com/mesa/mesa/blob/2b23c16b7f13f3fdcf9c11c90853d859d1acedec/mesa/experimental/data_collection/dataset.py#L359-L367

Can we use this approach in Continuous Space as well?

Austin-dev-dot:
Fixes #3548

The `_add_agent` method used `np.vstack` with a hardcoded 20% growth fraction to resize the positions array. This reconstructs the entire array on every resize. Replaced it with the same geometric doubling approach used by `NumpyAgentDataSet`:

- Added class-level `_GROWTH_FACTOR = 2.0` and `_MIN_GROWTH = 100`
- Growth is calculated as `max(int(current_size * (_GROWTH_FACTOR - 1)), _MIN_GROWTH)`
- Uses `np.empty` + slice copy instead of `np.vstack`

Also added a test that verifies the growth strategy gives sufficient capacity and preserves existing positions after resize.

### DIFF
--- a/tests/experimental/test_continuous_space.py
+++ b/tests/experimental/test_continuous_space.py
@@ -520,4 +520,3 @@ def test_array_growth_strategy():
         np.testing.assert_array_equal(
             space._agent_positions[a._mesa_index], [float(i), float(i)]
         )
-


quaquel:
This is premature. As indicated in the issue, I'd like to see some microbenchmarks first.

PR #3551, opened 2026-03-16
Add FutureWarning for deprecated DataCollector and update migration guide

### LINKED ISSUE #3547: Add deprecation warnings to DataCollector
Hi!

While reviewing the Mesa 4 development tracking issue (#3132), I noticed the task **"Add deprecation warnings to DataCollector."** It does not appear that there is a dedicated issue for this task yet, so I’m opening this issue to discuss the implementation.

Since `DataCollector` is planned to be removed in Mesa 4.0, it may be helpful to emit a deprecation warning when it is used, so users are aware that the class will be removed in a future version.

One possible approach would be adding a `DeprecationWarning` using Python’s `warnings` module in `mesa/datacollection.py`, for example when `DataCollector` is instantiated.

Before starting implementation, I wanted to confirm:

* Whether this task is still planned for Mesa 4.0
* If there are specific locations where the warning should be added
* Whether there is a preferred pattern for deprecation warnings in Mesa

If this approach sounds reasonable, I would be happy to work on the implementation and open a PR.

Thanks!

satharyasin-hub:
Summary

This PR adds a FutureWarning when DataCollector is instantiated to notify users that the class is deprecated and will be removed in Mesa 4.0.

The warning directs users to the new data collection system (DataRegistry, DataSet, and DataRecorder) and links to the migration guide for guidance on migrating existing code.

In addition, the migration guide has been updated with a section describing the deprecation and providing a simple example of how users can transition from DataCollector to the new system.

Changes

Added a FutureWarning in DataCollector.__init__() using Python's warnings module

Included a link to the migration guide in the warning message

Added a new DataCollector Deprecated section to docs/migration_guide.md

Provided example code showing the old DataCollector usage and the new data collection system

Related Issue

Closes #3547

GSoC contributor checklist
Context & motivation

While reviewing the Mesa 4 development tracking issue (#3132), I noticed the task to add a deprecation warning for DataCollector. Since the new data collection system (DataRegistry, DataSet, and DataRecorder) has already been introduced and DataCollector is planned to be removed in Mesa 4.0, adding a warning helps notify users about the upcoming change and guide them toward the new system.

I opened issue #3547 to discuss the implementation approach before starting the work.

What I learned

While working on this change, I explored how DataCollector works internally and how Mesa collects model, agent, and agent-type data during simulations. I also learned how deprecation warnings are implemented in Python using the warnings module and how these warnings interact with the existing test suite and example models that still use DataCollector.

This helped me better understand Mesa’s deprecation policy and how changes should be introduced gradually to avoid breaking existing user code.

Learning repo

🔗 My learning repo:
https://github.com/satharyasin-hub/GSoC-learning-space

🔗 Relevant model(s):
https://github.com/satharyasin-hub/GSoC-learning-space/tree/main/models/my_first_model/schelling_model

Readiness checks

 This PR addresses an agreed-upon problem (linked issue or discussion with maintainer approval)

 I have read the contributing guide and deprecation policy

 I have performed a self-review

 Tests pass locally (pytest --cov=mesa tests/)

 Code is formatted (ruff check . --fix)

 Documentation / migration guide updated

### DIFF
--- a/docs/migration_guide.md
+++ b/docs/migration_guide.md
@@ -4,15 +4,88 @@ This guide contains breaking changes between major Mesa versions and how to reso
 Non-breaking changes aren't included, for those see our [Release history](https://github.com/mesa/mesa/releases).
 
 ## Mesa 3.5.0
+### Event scheduling and time advancement
+Mesa 3.5 introduces public methods for event scheduling and time advancement directly on `Model`, replacing the need for `Simulator` classes.
+
+#### Time-based advancement replaces step loops
+```python
+# Old
+for _ in range(10):
+    model.step()
+
+# New
+model.run_for(10)    # Functionally equivalent for standard ABMs
+model.run_until(10)  # You can now also run until a specific time
+```
+`run_for(1)` produces identical results to `step()` for traditional models.
+
+#### One-off event scheduling
+
+```python
+# Schedule a single event at or after a specific time
+model.schedule_event(callback, at=50.0)    # Absolute time
+model.schedule_event(callback, after=5.0)  # Relative time
+
+# Cancel if needed
+event = model.schedule_event(callback, at=100.0)
+event.cancel()
+```
+
+#### Recurring event scheduling
+```python
+from mesa.time import Schedule
+
+# Schedule an event every 10 time units
+model.schedule_recurring(func, Schedule(interval=10))  # By default, starting after 1 interval (t=10 in this case)
+model.schedule_recurring(func, Schedule(interval=10, start=0))  # Start immediately or at any other time
+
+# Save the event and stop it when needed
+gen = model.schedule_recurring(func, Schedule(interval=5.0))
+gen.stop()
+
+# Limit executions
+model.schedule_recurring(func, Schedule(interval=1.0, count=10))
+```
+
+#### Replacing Simulator classes
+The experimental Simulator classes are now also deprecated.
+```python
+# Old - ABMSimulator
+from mesa.experimental.devs.simulator import ABMSimulator
+simulator = ABMSimulator()
+simulator.setup(model)
+simulator.run_for(100)
+
+# New
+model.run_for(100)
+```
+
+```python
+# Old - DEVSimulator
+from mesa.experimental.devs.simulator import DEVSimulator
+simulator = DEVSimulator()
+simulator.setup(model)
+simulator.schedule_event_absolute(callback, time=10.5)
+simulator.run_for(50)
+
+# New
+model.schedule_event(callback, at=10.5)
+model.run_for(50)
+```
+
+Mesa 3.5 doesn't introduce any breaking changes. Mesa 4 will clean up many deprecated components and thus will break unmodified models.
+
+* Ref: [Discussion #2921](https://github.com/mesa/mesa/discussions/2921), [PR #3266](https://github.com/projectmesa/mesa/pull/3266)
+
 ### DataCollector Deprecated
 
 `DataCollector` is deprecated and will be removed in Mesa 4.0.
 
 Users should migrate to the new data collection system:
 
-- `DataRegistry`
-- `DataSet`
-- `DataRecorder`
+* `DataRegistry`
+* `DataSet`
+* `DataRecorder`
 
 The legacy `DataCollector` class will continue to work for now but will emit a `PendingDeprecationWarning` to notify users about the upcoming removal.
 
@@ -21,9 +94,9 @@ The legacy `DataCollector` class will continue to work for now but will emit a `
 Old (DataCollector) → New (DataRegistry + DataRecorder)
 
 Key Concepts:
-- DataRegistry → defines WHAT to collect
-- DatasetConfig → defines WHEN to collect
-- DataRecorder → defines HOW to store
+* DataRegistry → defines WHAT to collect
+* DatasetConfig → defines WHEN to collect
+* DataRecorder → defines HOW to store
 
 ### Note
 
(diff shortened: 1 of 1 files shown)


docs/migration_guide.md
```diff
@@ -4,6 +4,37 @@ This guide contains breaking changes between major Mesa versions and how to reso
...
+)
+```
+
+```python
+# New
+# Use the new data collection system instead
+from mesa.datacollection import DataRegistry
+```
+
+- Ref: [Issue #3547](https://github.com/projectmesa/mesa/issues/3547), [PR #3550](https://github.com/projectmesa/mesa/pull/3550)
```
codebreaker32: Wouldn't it make more sense to attach the links for PR where Dataregistry and Datarecorder were introduced?

docs/migration_guide.md
```diff
@@ -4,6 +4,37 @@ This guide contains breaking changes between major Mesa versions and how to reso
...
+
+datacollector = DataCollector(
+    model_reporters={"AgentCount": lambda m: len(m.agents)}
+)
+```
+
+```python
+# New
+# Use the new data collection system instead
+from mesa.datacollection import DataRegistry
```
codebreaker32: 1. You'll need to import DataRecorder
2. DataRegistry is already attached to Model
3. Wrong module, its in experimental section
4. There is no usage example

satharyasin-hub:
It looks like the CI failures are caused by the new FutureWarning emitted when DataCollector is instantiated during the tests.

Since many existing tests create DataCollector instances, the warning appears during test execution and pytest treats it as a failure.

Would you prefer that the tests be updated to explicitly expect the FutureWarning using pytest.warns, or should the warning be filtered in the test configuration?

codebreaker32:
Tests also need to be updated as you now raise warning

codebreaker32:
I'd advise you to go through #3145 and #3156

satharyasin-hub:
Thanks for the review and suggestions!

I'll update the migration guide to:
- include the correct module path
- add a proper usage example
- reference DataRecorder as well

I'll also update the tests to expect the FutureWarning raised by DataCollector.

quaquel:
As I warned in #3547, this would break all tests. I am at this point not sure whether this PR is coming in at the right time. 

We want to remove the datacollector, but we also know it is tighly integrated into a lot of things. So, the best output of a pathfinding PR like this is to provide is with insight into everything that breaks. Next, this PR might just be closed, while we open discussions on how to first fix everything that will be broken by removing the data collector. 

From the top of my head we have at least
* solaraviz; this is the big one. How do we want to move forward with the UI without the data collector? I am not sure I want to just replace a lock-in on the data collector with a lock in on the data recorder. Instead, I would like to explore possiblities for streaming data to the UI and make the UI responsible for any necessary accumulation.
* batchrunning; this is the trivial one since the batch runner will also be removed. See #3483. 

But there might be more.

satharyasin-hub:
Thanks for the clarification and for the detailed explanation.

That makes sense — I can see how tightly integrated `DataCollector`
is across the codebase and how introducing the warning currently
breaks many tests.

I'm glad the PR helped identify some of the areas that would be
affected by the migration.

Happy to leave this PR as a reference for the pathfinding work,
or close it if that is the preferred next step while the migration
strategy is discussed.

EwoutH:
> As I warned in #3547, this would break all tests. I am at this point not sure whether this PR is coming in at the right time.

One trick is to use `PendingDeprecationWarning`. Our CI is currently configured to show `PendingDeprecationWarning`s but not fail the tests on them. So then you can fix them one by one.

See our [deprecation policy](https://github.com/mesa/mesa/blob/main/CONTRIBUTING.md#deprecation-policy).

satharyasin-hub:
Thanks for the feedback!

I'll update the migration guide example to include:
- the correct module import
- DataRecorder usage
- a proper usage example

Pushing an update shortly.

quaquel:
As indicated, the key point to doing this is so we know where the problems are when we remove it. Just adding the warning is trivial. The value is in the issues that follow from it.

satharyasin-hub:
Thanks, that makes sense.

I understand that the main value of this PR is identifying where
DataCollector is tightly integrated.

From the warnings, I can already see usage across:
- tests
- examples
- visualization (solara viz)
- batch running

Would you prefer that I:
1) Open follow-up issues documenting these areas, or
2) Start working on one specific area (e.g. examples or tests) to migrate?

Happy to continue in the direction that is most useful.

EwoutH:
> The value is in the issues that follow from it.

But let's add the `PendingDeprecationWarning` in one PR, then fix stuff in follow-up PRs, and then once everything is resolved open a PR to update to a `FutureWarning`.

quaquel:
Yes, but I'd like to have an overview of those issues.

docs/migration_guide.md
satharyasin-hub: Thanks for the feedback!

I've updated the example to:
- Use the correct experimental module
- Include a complete working example using DataRecorder
- Fix import issues and ensure it runs properly

Let me know if anything else should be improved.

docs/migration_guide.md
```diff
@@ -4,9 +4,119 @@ This guide contains breaking changes between major Mesa versions and how to reso
 Non-breaking changes aren't included, for those see our [Release history](https://github.com/mesa/mesa/releases).
 
 ## Mesa 3.5.0
-### Event scheduling and time advancement
-Mesa 3.5 introduces public methods for event scheduling and time advancement directly on `Model`, replacing the need for `Simulator` classes.
+### DataCollector Deprecated
```
EwoutH, line 80: Putting it under 3.5 I think you're implying we're going/needing to backport it to 3.5.x, is that your intention?

quaquel: We have to backport the warning at some point since I hope to be able to remove the datacollector before the release of 4.

EwoutH:
@B2prakash this is just a summary (in some sense useful), not a review. Try to take a critical look at it yourself.

docs/migration_guide.md
satharyasin-hub: Thanks for the clarification!

My intention was to introduce the deprecation warning early so users are informed ahead of Mesa 4.0, so I placed it under 3.5.x.

Given that we plan to backport the warning, I believe this placement makes sense, but I’m happy to adjust if there’s a preferred structure.

codebreaker32 (commented):
Update the tutorials, This is not how it should be done, You can go through the original PRs or docstrings to understand the usage.

docs/migration_guide.md
```diff
@@ -117,75 +190,6 @@ if __name__ == "__main__":
     df = model.recorder.get_table_dataframe("model_data")
     print(df)
 ```
-#### Time-based advancement replaces step loops
```
EwoutH, line 10: Why was this removed?

satharyasin-hub: Thanks for pointing this out!

I realized I unintentionally replaced the existing “Event scheduling and time advancement” section while adding the deprecation note. I’ve now restored the original content and added the DataCollector deprecation as a separate section.

Please let me know if this structure looks correct.

mesa/datacollection.py
```diff
@@ -116,6 +116,14 @@ def __init__(
...
               use attribute names for the agent reporter, as it will be faster.
         """
+        warnings.warn(
+            "DataCollector is deprecated and will be removed in Mesa 4.0. "
+            "Use the new data collection system (DataRegistry, DataSet, DataRecorder) instead. "
+            "See: https://mesa.readthedocs.io/latest/migration_guide.html",
+            PendingDeprecationWarning,
+            stacklevel=2,
+        )
+
```
TheMS1314, line 126: stacklevel=2 is the right choice here — it makes the warning point to the caller's code where DataCollector is instantiated rather than to this line inside the library, which gives users a more useful traceback. One thing worth considering: PendingDeprecationWarning is silently ignored by Python by default unless the user explicitly enables it. So most users running standard Python won't see this warning at all without -W default::PendingDeprecationWarning. Might be worth adding a note in the migration guide about how to surface it.

docs/migration_guide.md
```diff
@@ -77,6 +77,120 @@ Mesa 3.5 doesn't introduce any breaking changes. Mesa 4 will clean up many depre
...
+        config = {
+            "model_data": DatasetConfig(interval=1)
+        }
+
+        # 3️⃣ HOW to store
+        self.recorder = DataRecorder(self, config=config)
+
+    @property
+    def agent_count(self):
+        return len(self.agents_list)
```
TheMS1314, line 173: The note suggesting recorder.collect() should be called manually in step() conflicts with the actual implementation — BaseDataRecorder subscribes to model.time observable signals automatically via _subscribe_to_model(), so collection is triggered without manual calls. The collect() method's own docstring says "Normally you don't need to call this." Removing this note would make the example more accurate for users migrating from DataCollector.

TheMS1314:
Went through this PR as part of a peer review swap. The core change is clean and the decision to use PendingDeprecationWarning over FutureWarning makes sense given how tightly integrated DataCollector is — surfacing the breakage points gradually is smarter than breaking everything at once. The migration example is the most useful part for end users and it's in good shape now. Looking forward to seeing the follow-up PRs that address SolaraViz and batch runner dependencies.
Also leaving this here — I'm a GSoC 2026 candidate on mesa-examples, PR #443 extends the Emperor's Dilemma model if you get a chance to review!

codebreaker32:
> The note about calling recorder.collect() manually is important and easy to miss without it

Please don't make such remarks, this is entirely wrong. Once again, *I'd suggest to go through original PRs/codebase or docstrings*

TheMS1314:
> > The note about calling recorder.collect() manually is important and easy to miss without it
> 
> Please don't make such remarks, this is entirely wrong. Once again, _I'd suggest to go through original PRs/codebase or docstrings_

Thanks for the correction! I'll go through the original PRs and docstrings to understand the correct behavior. I'll update my review comment accordingly.

TheMS1314:
> > > The note about calling recorder.collect() manually is important and easy to miss without it
> > 
> > 
> > Please don't make such remarks, this is entirely wrong. Once again, _I'd suggest to go through original PRs/codebase or docstrings_
> 
> Thanks for the correction! I'll go through the original PRs and docstrings to understand the correct behavior. I'll update my review comment accordingly.

Thanks for pushing me to read the original PRs — went through #3145 and basedatarecorder.py directly. You're right, collection is signal-driven automatically via _subscribe_to_model() and the collect() docstring itself says "Normally you don't need to call this." Updated my comment accordingly. Should have done that before commenting.

codebreaker32:
Hi @satharyasin-hub 
Are you still interested in taking this further?

codebreaker32:
Should I take this on as @satharyasin-hub looks inactive? @quaquel

quaquel:
@codebreaker32 please if you have the time.

PR #3558, opened 2026-03-16
docs: auto-detect mesa version in conf.py

### LINKED ISSUE #3507: Make users see stable (V3.5) documentation + code by default
When reading the documentation and looking up the source code on GitHub, I was pretty confused between the observed behavior and the documentation (e.g. batch_run() not being available or documented). Turns out, I was having V3.5 installed and looked at the documentation and the code for the unreleased V4.

Having the main branch stable at the latest release and having a separate dev branch should solve the problems. I have never worked with readthedocs, but the "latest" documentation seems to look at the main branch.

Additionally display the version in the documentation so you know if you are reading the right version.

(Side-Quest: Having different tagged major+minor versions available at readthedocs would also be awesome.)

Sorry, too little time right now :( For switching switching the purpose of the main branch I guess I have to little permissions.

**Additional context**

shalini-s7:
Closes #3507

## What this PR does
- Replaces hardcoded version (`"0.5"`) in docs/conf.py with 
  automatic version detection using 'importlib.metadata'
- The docs will now always display the correct installed version 
  of Mesa, so users can confirm they are reading the right version

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,4 +1,3 @@
-
 #!/usr/bin/env python3
 #
 # Mesa documentation build configuration file, created by


quaquel:
I agree this is an improvement, but how does this solve the issue?

EwoutH:
Where is this number used and display by ReadTheDocs exactly?

quaquel:
@EwoutH it can be used in the version drop-down menu, but this change alone is nowhere near enough to close the issue.

I'll try to put in a PR later this week that properly addresses the issue on the configuration side, but it will also require some manual stuff on the readthedocs website itself.  Have been doing some comparisons with other libraries I maintain, so I know how to fix it. 

In the meantime, I am closing this PR because it does not solve the problem.

shalini-s7:
Thanks for the feedback! I understand this is only a partial fix. Happy to learn and contribute more in the future!

PR #3560, opened 2026-03-17
fix: avoid eager imports in mesa/__init__ using lazy loading

DashamiJituri:
This PR replaces eager imports in `mesa/__init__.py` with lazy loading to avoid importing optional dependencies at package import time.

## Changes
- Removed direct imports of `discrete_space`, `experimental`, and `time` from `__init__.py`
- Added `__getattr__` to lazily load these modules when accessed
- Preserved existing API via `__all__`

## Motivation
Previously, importing `mesa` would eagerly import submodules that rely on optional dependencies (e.g., visualization or network-related libraries), which could lead to import errors for users without those dependencies installed.

## Result
- Reduces unnecessary imports at package initialization
- Prevents import-time errors due to missing optional dependencies
- Maintains backward compatibility for existing import patterns

### DIFF
--- a/mesa/__init__.py
+++ b/mesa/__init__.py
@@ -27,14 +27,18 @@ __copyright__ = f"Copyright {_this_year} Mesa Team"
 
 # Lazy import to avoid importing optional dependencies at package import time
 
+
 def __getattr__(name):
     if name == "discrete_space":
         import mesa.discrete_space as discrete_space
+
         return discrete_space
     elif name == "experimental":
         import mesa.experimental as experimental
+
         return experimental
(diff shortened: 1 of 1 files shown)


quaquel:
Let's first agree on a way forward in #2343 and do some benchmarking. Closing this.

DashamiJituri:
> Let's first agree on a way forward in #2343 and do some benchmarking. Closing this.

Thanks for the clarification!

That makes sense — I understand that this should be aligned with the broader discussion in #2343 first.

I’ll follow the discussion there and would be happy to contribute once the direction is decided.

PR #3565, opened 2026-03-18
Version docs

### LINKED ISSUE #3507: Make users see stable (V3.5) documentation + code by default

quaquel:
### Summary
Enable correct version switching in readthedocs

### Bug / Issue
Closes #3507.

### Implementation
This is largely a configuration issue. I updated `conf.py` to autodetect the version and release of Mesa that is being built. I also updated the configuration for the version switcher to show the version number. I added a JSON (see below - but this points to stable, so it won't work and cannot be tested for now). 

### Testing
See the render of read the docs

## discussion point
There are 2 ways to control which versions are built. Option 1 is via the readthedocs website. Here you can select GitHub tags for which you want to build docs. This is the easiest approach, but it requires someone to update this. I don't have the login rights to do it. I use this approach on the ema_workbench and pynetlogo, two other libraries I maintain. 

Option 2 is via a JSON file in the repo. This can be incorporated into the release cycle and the release CI. I have not used this before, but we can explore this if we prefer this approach.

Thoughts and preferences are particularly welcome from @EwoutH, @jackiekazil, and @tpike3.

### DIFF
--- a/mesa/datacollection.py
+++ b/mesa/datacollection.py
@@ -146,6 +146,20 @@ class DataCollector:
             for name, columns in tables.items():
                 self._new_table(name, columns)
 
+    @staticmethod
+    def _check_list_reporter(name, reporter):
+        """Validate that a list-style reporter has the form [callable, [params]]."""
+        if len(reporter) != 2 or not callable(reporter[0]):
+            raise ValueError(
+                f"Reporter '{name}' must use the format [function, [param1, param2]]. "
+                f"Got: {reporter!r}"
+            )
+        if not isinstance(reporter[1], (list, tuple)):
+            raise ValueError(
+                f"Reporter '{name}' must use the format [function, [param1, param2]]. "
+                f"The second element must be a list or tuple of parameters, got: {reporter[1]!r}"
+            )
+
     def _validate_model_reporter(self, name, reporter, model):
         """Validate model reporter and handle validation results appropriately.
 
@@ -160,9 +174,8 @@ class DataCollector:
             TypeError: If reporter type is not supported
             RuntimeError: If reporter execution fails
         """
-        self._validated = True  # put the change of signal firstly avoid losing efficacy
+        self._validated = True
 
-        # Type 1: Lambda function or partial
         if isinstance(reporter, (types.LambdaType, partial)):
             try:
                 reporter(model)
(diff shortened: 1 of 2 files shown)


docs/_static/switcher.json
```diff
@@ -0,0 +1,42 @@
...
+  },
+  {
+    "name": "stable",
+    "version": "stable",
+    "url": "https://mesa.readthedocs.io/en/stable/"
+  },
+  {
+    "name": "3.5.1",
+    "version": "v3.5.1",
+    "url": "https://mesa.readthedocs.io/en/v3.5.1/"
```
EwoutH, line 20: This seems like another thing to manually update every release. Is there any way this can be automated?

quaquel: I don't know. Some of it might be handled in the release CI or the build CI. Basically, on release, run a simple script that updates the switcher JSON. 

Unfortunately, neither approach (website, or JSON) can easily be automated by default. The JSON approach, as used here, at least holds some potential. The website requires manual configuration always. As far as I have been able to discover, these two approaches are the only options.

quaquel: @EwoutH , see the last commit of what is required. I added a job to the release yaml. This job runs after a successful PyPi publish and only on stable tags (so no alpha and pre-release). The job checks out main, updates `switcher.json`, and pushes it to main.  I am not sure about this approach. I don't like automatic commits, but it at least shows a direction. I am more inclined to include it in the release guidelines, analogous to how we handle the release notes.

jackiekazil: I am not a fan of auto commits either. Nor am I a fan of manual. 
But given the choice between two things I don't like, in this case, I lean towards the manual operation unless someone can come up with another idea.

quaquel: Doing this manually should be quite easy. It's just adding one item to the JSON file when creating a new release.

EwoutH:
Maybe just documenting it in our release checklist is enough? https://github.com/mesa/mesa/blob/main/CONTRIBUTING.md#releases

quaquel:
> Maybe just documenting it in our release checklist is enough? https://github.com/mesa/mesa/blob/main/CONTRIBUTING.md#releases

Yes, that is my inclination as well. I'll try to update this PR accordingly later today.

quaquel:
I am merging this. We have to check after the next stable release to see if everything is working as intended.

PR #3591, opened 2026-03-21
Add agent-level action lifecycle hooks

EwoutH:
### Summary
Adds three hooks to `Agent` that fire after the corresponding `Action` hooks, giving agents a centralized place for behavioral decision logic.

Quick prototype to see how https://github.com/mesa/mesa/discussions/3304#discussioncomment-16242923 would look. No tests etc. yet.

### Motive
Currently, every `Action.on_complete` and `Action.on_interrupt` must include "what next?" logic, leading to duplicated decision code across actions. This couples action _effects_ with agent-level behavioral _decisions_.

With agent-level hooks, actions stay single-purpose (defining their own effects), while the agent handles decisions in one place:

```python
class Forage(Action):
    def on_complete(self):
        self.agent.energy += 30  # Just the effect

class Sheep(Agent):
    def on_action_complete(self, action):
        if self.energy < 30:
            self.start_action(Rest(self))
        else:
            self.start_action(Forage(self))
```

### Implementation
**`agent.py`** — Three new no-op methods on `Agent`:
- `on_action_start(action)` — called after `Action.on_start`/`on_resume`
- `on_action_complete(action)` — called after `Action.on_complete`
- `on_action_interrupt(action, progress)` — called after `Action.on_interrupt`

**`actions.py`** — One line added in each of:
- `Action.start()` — calls `agent.on_action_start` (skipped for instantaneous actions, which go through `_do_complete` instead)
- `Action._do_complete()` — calls `agent.on_action_complete`
- `Action.interrupt()` and `Action.cancel()` — call `agent.on_action_interrupt`

### Usage Examples
**Before:** decision logic duplicated in every action:
```python
class Forage(Action):
    def on_complete(self):
        self.agent.energy += 30
        # decision logic here...
        if self.agent.sees_predator():
            self.agent.start_action(Flee(self.agent))
        elif self.agent.energy > 80:
            self.agent.start_action(Rest(self.agent))
        else:
            self.agent.start_action(Forage(self.agent))

class Rest(Action):
    def on_complete(self):
        self.agent.energy = min(100, self.agent.energy + 20)
        # ...duplicated here
        if self.agent.sees_predator():
            self.agent.start_action(Flee(self.agent))
        elif self.agent.energy > 80:
            self.agent.start_action(Rest(self.agent))
        else:
            self.agent.start_action(Forage(self.agent))

class Flee(Action):
    def on_complete(self):
        self.agent.move_away_from_predator()
        # ...and here
        if self.agent.sees_predator():
            self.agent.start_action(Flee(self.agent))
        elif self.agent.energy > 80:
            self.agent.start_action(Rest(self.agent))
        else:
            self.agent.start_action(Forage(self.agent))
```

**After:** actions are single-purpose, decisions live in one place:
```python
class Forage(Action):
    def on_complete(self):
        self.agent.energy += 30

class Rest(Action):
    def on_complete(self):
        self.agent.energy = min(100, self.agent.energy + 20)

class Flee(Action):
    def on_complete(self):
        self.agent.move_away_from_predator()

class Sheep(Agent):
    def on_action_complete(self, action):
        self.decide_next()

    def on_action_interrupt(self, action, progress):
        self.decide_next()

    def decide_next(self):
        if self.sees_predator():
            self.start_action(Flee(self))
        elif self.energy < 30:
            self.start_action(Rest(self))
        else:
            self.start_action(Forage(self))
```

Adding a fourth action (e.g. `Drink`) requires zero changes to decision logic. Changing the decision logic requires updating only one method.

All hooks default to `pass`, so existing code is unaffected.

### Additional Notes
- Hooks fire *after* the Action's own hooks, so `Action.on_complete` runs before `Agent.on_action_complete`.
- Zero-duration actions fire `on_action_complete` but not `on_action_start`, avoiding double-firing in the same instant.
- This is purely additive, you can still put all logic in `Action.on_complete` if you want.
- Ref: https://github.com/mesa/mesa/discussions/3304#discussioncomment-16242923

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -299,3 +299,13 @@ class Agent[M: Model]:
     def is_busy(self) -> bool:
         """Whether the agent is currently performing an action."""
         return self.current_action is not None
+
+    # --- Action lifecycle hooks (override in subclasses) ---
+    def on_action_start(self, action: Action) -> None:
+        """Called after Action.on_start/on_resume. Override for agent-level responses."""
+
+    def on_action_complete(self, action: Action) -> None:
+        """Called after Action.on_complete. Override to decide the agent's next action."""
+
+    def on_action_interrupt(self, action: Action, progress: float) -> None:
+        """Called after Action.on_interrupt. Override to handle interruption at agent level."""
--- a/mesa/experimental/actions/actions.py
+++ b/mesa/experimental/actions/actions.py
@@ -263,6 +263,7 @@ class Action:
 
         # Schedule completion event for remaining duration
         self._event = self.model.schedule_event(self._do_complete, after=remaining)
+        self.agent.on_action_start(self)
         return self
 
     def interrupt(self) -> bool:
@@ -291,6 +292,7 @@ class Action:
             self.agent.current_action = None
 
         self.on_interrupt(self._progress)
+        self.agent.on_action_interrupt(self, self._progress)
         return True
 
     def cancel(self) -> bool:
@@ -315,6 +317,7 @@ class Action:
             self.agent.current_action = None
 
         self.on_interrupt(self._progress)
+        self.agent.on_action_interrupt(self, self._progress)
         return True
 
     # --- Internal ---
@@ -349,6 +352,7 @@ class Action:
             self.agent.current_action = None
 
         self.on_complete()
+        self.agent.on_action_complete(self)
 
     def __repr__(self) -> str:
         """Return string representation."""


quaquel:
I find it confusing that you speak of firing. Signals are emitted or fired, but here the methods are simply called directly. 


> This couples action _effects_ with agent-level behavioral _decisions|_.

Yes, this is a very important and useful distinction. 

I don't fully agree, however, with your example where you put all agent decision logic in your actions. I can easily envision using the current actions merely as a way of scheduling an activity that lasts for a given duration and has a fixed effect while ending with a callback to the owning agent (see below). This PR mainly formalizes this.

```python
class Forage(Action):
    def on_complete(self):
        self.agent.energy += 30
        self.agent.action_completed()

```

I am, however, a bit concerned about the api design. We now have `Action.on_x` and `Agent.on_action_x`. Namewise, these are quite close together, which might cause confusion. It also suggests we might have to move the agent action stuff into a `HasActions` class as discussed before.

Vanya-kapoor:
@quaquel 
On the naming confusion between `Action.on_x` and `Agent.on_action_x `— one option would be a` HasActions` mixin that groups `on_action_complete`, `on_action_interrupt`,` on_action_start together`, making the agent-side hooks clearly separate from the action-side hooks. This also keeps the base Agent class clean for agents that don't use Actions at all.

codebreaker32:
Great work on Actions @EwoutH

I’ve been tackling this exact problem, separating an action's physical effect from the agent's decision of "what to do next", while prototyping a Behavioral Framework proposal. 

While I agree with @EwoutH’s goal of removing decision boilerplate from `Action.on_complete`, I share @quaquel's concerns about the API design. Adding direct method calls to the base `Agent` class creates a few architectural friction points: It keeps the Action tightly coupled to the Agent's API. If an agent wants to react to an action completing, it is forced to implement a specific method name.

To solve this without cluttering the agent, I actually leaned exactly into the `HasActions`(In my case I called it TaskManager) concept that @quaquel mentioned, combined with signals

Instead of putting hooks on the Agent, I built a `TaskManager` (which essentially acts as a `HasActions` component). Here is the data flow:

1. The Action (`Task`) handles its own physical effect in `on_complete` (e.g., `self.agent.energy += 30`). 
2. The Action notifies the `TaskManager` that it is done.
3. The `TaskManager` emits a signal (`TaskSignals.COMPLETED`).
4. A `DecisionSystem`(which is agent's brain in my proposal) simply observes that signal to trigger the next decision.

Here is what that looks like in practice.

```python
class BehavioralAgent(Agent, HasEmitters):
    def __init__(self, model):
        ...
        self.task_manager = TaskManager(self) 
        self.decision_system = DecisionSystem(self, self.task_manager)
        self.observe("task_manager", TaskSignals.COMPLETED, self._on_action_done)

    def _on_action_done(self, **kwargs):
        # Action is done. Evaluate the environment and do the next thing!
        self.decision_system.evaluate() 
```

Instead of adding on_action_complete, on_action_interrupt, etc., to the base Agent class, what if the Action class simply utilize the Signals to emit `ActionSignals.COMPLETED`?Curious to hear your thoughts!

EwoutH:
@quaquel what do you think about Aman’s approach?

quaquel:
I like the richer conceptual design. I am less sure about the specific use of signals. 

For me, signals are used for asynchronous communication. But if an agent starts an action, and it then completes, to me the agent's response to this is synchronous not asynchronous. Would have to see a more worked out conceptual diagram or so to provide more feedback.

This does not mean that actions should/could not fire signals on start, interupt and completion. But these would be for, for example, statistics gathering.

codebreaker32:
> Would have to see a more worked out conceptual diagram or so to provide more feedback.

I have discussed a preliminary design here https://github.com/mesa/mesa/discussions/3428 However I need to update it with some of my later findings, new Action arch and your current feedbacks.

quaquel:
I am closing this since it is being replaced with all the new action stuff.

PR #3621, opened 2026-03-23
Add join strategy hook for existing meta-agents

yassinekolsi:
## Summary
- add an optional join-strategy hook when multiple existing meta-agents of the same class are available
- keep default behavior unchanged when no selector is provided
- add a regression test to validate custom selection logic

## Why
There is a TODO in `create_meta_agent` about allowing user control over how agents join an existing meta-agent. This adds a minimal, opt-in selector without changing current defaults.

## Scope
- Atomic change: meta-agent selection only
- Files changed: 2

## Validation
- `python -m pytest tests/experimental/test_meta_agents.py -q`

## Overlap check
- No existing open PR found for this specific join-strategy hook.

### DIFF
--- a/tests/examples/test_examples_viz.py
+++ b/tests/examples/test_examples_viz.py
@@ -215,13 +215,7 @@ def test_virus_on_network_model(solara_test, page_session: playwright.sync_api.P
     """Test virus on network model behavior and visualization."""
     from mesa.examples.basic.virus_on_network.model import State  # noqa: PLC0415
 
-    model = VirusOnNetwork(
-        rng=42,
-        virus_spread_chance=1.0,
-        virus_check_frequency=1.0,
-        recovery_chance=1.0,
-        gain_resistance_chance=1.0,
-    )
+    model = VirusOnNetwork(rng=42)
 
     def agent_portrayal(agent):
         node_color_dict = {
--- a/tests/examples/test_examples_viz.py
+++ b/tests/examples/test_examples_viz.py
@@ -215,7 +215,13 @@ def test_virus_on_network_model(solara_test, page_session: playwright.sync_api.P
     """Test virus on network model behavior and visualization."""
     from mesa.examples.basic.virus_on_network.model import State  # noqa: PLC0415
 
-    model = VirusOnNetwork(rng=42)
+    model = VirusOnNetwork(
+        rng=42,
+        virus_spread_chance=1.0,
+        virus_check_frequency=1.0,
+        recovery_chance=1.0,
+        gain_resistance_chance=1.0,
+    )
 
     def agent_portrayal(agent):
         node_color_dict = {
--- a/tests/experimental/test_meta_agents.py
+++ b/tests/experimental/test_meta_agents.py
@@ -191,7 +191,7 @@ def test_create_meta_agent_custom_join_strategy_requires_existing(setup_agents):
 
     with pytest.raises(
         ValueError,
-        match="select_existing_meta_agent must return one of the existing meta-agents.",
+        match=r"select_existing_meta_agent must return one of the existing meta-agents\.",
     ):
         create_meta_agent(
             model,
--- a/tests/experimental/test_meta_agents.py
+++ b/tests/experimental/test_meta_agents.py
@@ -142,7 +142,7 @@ def test_create_meta_agent_custom_join_strategy(setup_agents):
     """Test selecting an existing meta-agent with a custom join strategy."""
     model, agents = setup_agents
 
-    meta_agent1 = create_meta_agent(
+    _meta_agent1 = create_meta_agent(
         model,
         "MetaAgentClass",
         [agents[0]],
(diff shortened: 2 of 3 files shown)


falloficaruss (commented):
The new callback receives `existing_meta_agents` in a nondeterministic order, because that list is built by iterating each agent’s `meta_agents` set and preserving discovery order. `agent.meta_agents` is a set, so a selector that uses positional logic like “pick the first candidate” will behave unpredictably across runs. Since this hook is explicitly about resolving ambiguity, I think the PR should sort candidates before passing them to `select_existing_meta_agent`, probably by `unique_id`, just like the default branch already does.

falloficaruss (commented):
The new tests cover the happy path for custom selection and the invalid-return case, but I think we still need one regression test for the default path. In the ambiguous case where multiple same-class meta-agents already exist, `create_meta_agent()` used to fall back to the lowest-unique_id candidate. Since this PR adds an opt-in hook rather than changing the default behavior, it would be good to add a test showing that when `select_existing_meta_agent` is not provided, the function still chooses the same default meta-agent as before. That would make the backward-compatibility story much clearer.

yassinekolsi:
Addressed the review feedback in commit 07920818.

- Sorted existing_meta_agents by unique_id before passing candidates to select_existing_meta_agent, so selector input order is deterministic even though agent.meta_agents is a set.
- Added regression test test_create_meta_agent_default_join_strategy to confirm default behavior is unchanged when no selector is provided (ambiguous case still joins the lowest-unique_id candidate).

Validation:
- .venv/bin/python -m pytest tests/experimental/test_meta_agents.py -q
- Result: 21 passed

yassinekolsi:
Conflict resolution update pushed in 024e8da3.

- Merged latest origin/main into this branch.
- Resolved the conflict in tests/experimental/test_meta_agents.py by keeping both sets of regression coverage:
  - default join-strategy regression test for create_meta_agent (lowest unique_id when no selector)
  - MetaAgent.remove() stale-reference cleanup tests from main
- Kept the deterministic selector-input fix (sorted by unique_id before select_existing_meta_agent callback).

Validation run:
- .venv/bin/python -m pytest tests/experimental/test_meta_agents.py -q
- Result: 23 passed

tests/experimental/test_meta_agents.py
```diff
@@ -138,6 +138,72 @@ def test_add_agents_to_existing_meta_agent(setup_agents):
...
+    )
+    create_meta_agent(
+        model,
+        "MetaAgentClass",
+        [agents[1]],
+        Agent,
+    )
+
+    def pick_non_existing(_existing_meta_agents, _agents):
+        return MetaAgent(model, set())
```
falloficaruss: The helper returns `MetaAgent(model, set())` just to force a bad value. That works, but it also mutates the model during the test. A plain sentinel object would make the test cleaner.

yassinekolsi: wait i'll fix it rn

falloficaruss:
This looks much cleaner now. For the original problem, the core question is not just “which existing meta-agent do we pick,” but “what kind of joining model should Mesa expose?” A callback solves the immediate ambiguity, but it leaves the user to invent the whole policy contract themselves each time.

A stronger design would be a small policy-based interface. For example, `create_meta_agent()` could accept something like `join_strategy=` with built-in options such as:

- `lowest_unique_id`
- `highest_unique_id`
- `create_new`
- `error_on_ambiguity`

And then, optionally, still allow a custom function for advanced users. That gives:

- a clean default
- reproducible built-in behavior
- less burden on users
- easier testing and documentation

I think that is better than only exposing a callback, because most users probably do not want to write a function just to say “fail if ambiguous” or “always create a new one”.

An even better long-term direction, if you follow the broader meta-agent discussion, is to separate two layers:

- Membership decision layer:
This decides whether an agent should join at all.

- Candidate resolution layer:
This decides which existing group to join when more than one valid group exists.

Right now this PR only addresses the second layer. That is useful, but it is still a narrow patch. The broader design would likely want both:

- a policy/hook for “can this agent join?”
- a policy/hook for “which group does it join?”

What are your thoughts?

tests/examples/test_examples_viz.py
```diff
@@ -215,7 +215,13 @@ def test_virus_on_network_model(solara_test, page_session: playwright.sync_api.P
...
     from mesa.examples.basic.virus_on_network.model import State  # noqa: PLC0415
 
-    model = VirusOnNetwork(rng=42)
+    model = VirusOnNetwork(
+        rng=42,
+        virus_spread_chance=1.0,
+        virus_check_frequency=1.0,
+        recovery_chance=1.0,
+        gain_resistance_chance=1.0,
+    )
```
falloficaruss: This is out of scope of the PR

yassinekolsi:
@falloficarus22, I completely agree with that long-term vision. Having a policy-based interface with built-in string presets (like lowest_unique_id or error_on_ambiguity) would definitely make for a much cleaner UX. I also agree that separating the 'can it join?' layer from the 'which group does it join?' layer is the right architectural move for meta-agents moving forward.

That being said, I want to be mindful of scope creep for this specific PR. This PR was originally scoped just to resolve the immediate ambiguity (the existing TODO) by providing the underlying callback mechanism.

Since we finally have the callback working cleanly and tested, what do you think about merging this as the foundational layer? From there, we can open a new issue to track building the join_strategy= policy interface and the two-layer architecture on top of it. I'd be happy to help draft up that issue so we don't lose these ideas. Let me know what you think!

tests/examples/test_examples_viz.py
yassinekolsi: the CI build was failing randomly on this specific test across different runs, so I just stabilized it to get the checks to pass so we could focus on the meta-agent stuff,you're right though,it is out of scope.I'll revert this change right now to keep the PR clean. I can open a quick separate PR to fix that flaky test later if you want

falloficaruss:
@tpike thoughts?

tpike3:
I think narrow is fine, and that was a very good discussion.  Could you take a look at the conflicts @yassinekolsi

yassinekolsi:
changes can be cleanly merged!

tpike3:
@yassinekolsi  Thank you very much for the PR! Due to the major refactor design of meta-agents for Google Summer of Code, I am closing all meta-agent PRs. 

Thank you again for your interest in Mesa and please keep contributing.

PR #3628, opened 2026-03-24
Fix #3597: skip missing measures in PlotMatplotlib, add regression test

DipayanDasgupta:
## Summary
Closes #3597 (partial — visualization fix)

`PlotMatplotlib` called `df.loc[:, m]` unconditionally for every key
in `measure`, crashing with `KeyError` when a metric wasn't in the
dataframe. When `grass=False`, the DataCollector never registers a
"Grass" reporter, so requesting it in the plot measure dict raises
`KeyError`.
```python
# Before — crashes when measure key missing from dataframe:
for m, color in measure.items():

# After — safe guard:
for m, color in measure.items():
    if m in df.columns:
```

The guard applies to all three measure types (`str`, `dict`,
`list|tuple`) — any model with conditionally-collected metrics benefits,
not just WolfSheep.

## Tests added
`test_plot_matplotlib_missing_column_skipped` — exercises the
`df.columns` guard directly, confirms no `KeyError` when a measure
key is missing from the dataframe.

---

### Context & motivation
After fixing the agent-level crash in WolfSheep with `grass=False`, the visualizer still crashed with a `KeyError` on "Grass". The DataCollector is intentionally dynamic — reporters are only registered conditionally — but `PlotMatplotlib` wasn't respecting that. It assumed every key in the measure dict existed in the dataframe, which isn't guaranteed.

### What I learned
Visualization components shouldn't assume a metric exists just because it was requested. The DataCollector is dynamic by design and the visualization layer needs to verify against the actual collected data before trying to plot it.

### Learning repo
🔗 Relevant model: https://github.com/DipayanDasgupta/GSoC-learning-space/tree/main/models/wolf_sheep_investigation

### Readiness checks
- [x] Addresses agreed-upon problem (issue #3597)
- [x] Self-review performed
- [ ] Peer review: @codebreaker32
- [x] Tests pass locally
- [x] Code formatted (`ruff check . --fix`)

### DIFF
--- a/mesa/examples/advanced/wolf_sheep/agents.py
+++ b/mesa/examples/advanced/wolf_sheep/agents.py
@@ -59,9 +59,9 @@ class Sheep(Animal):
     def feed(self):
         """If possible, eat grass at current location."""
         grass_patch = next(
-            obj for obj in self.cell.agents if isinstance(obj, GrassPatch)
+            (obj for obj in self.cell.agents if isinstance(obj, GrassPatch)), None
         )
-        if grass_patch.fully_grown:
+        if grass_patch is not None and grass_patch.fully_grown:
             self.energy += self.energy_from_food
             grass_patch.get_eaten()
 
--- a/mesa/experimental/continuous_space/continuous_space.py
+++ b/mesa/experimental/continuous_space/continuous_space.py
@@ -76,6 +76,7 @@ class ContinuousSpace:
 
         self.dimensions: np.array = np.asanyarray(dimensions)
         self.ndims: int = self.dimensions.shape[0]
+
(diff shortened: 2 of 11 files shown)


DipayanDasgupta:
@tpike3 Merge conflict resolved and pre-commit.ci noqa directives restored. All 12 CI checks (build, coverage, examples, docs) are passing — only the required pre-commit.ci check was failing due to missing # noqa: PLC0415 on inline imports, which this commit fixes. Would you be able to give this a final review? The fix is a minimal guard (if m in df.columns) across all three measure types in PlotMatplotlib, with a regression test confirming no KeyError when a metric is absent from the dataframe.

quaquel:
closing this because of conflicts. Feel free to reopen and solve the merge conflicts.

PR #3657, opened 2026-03-30
test: stabilize action interruption and viz tests

SumeetDUTTA:
## Summary
- replace direct `_event` mutation in action interruption tests with `model.run_for(...)` plus public `interrupt()`/`cancel()` paths
- add retry logic for Playwright screenshots in examples viz tests to handle DOM re-render detachment

## Overview
This PR improves test stability in two places.

In the experimental Actions interruption tests, it replaces direct `_event` mutation with real model time advancement and public interruption/cancellation paths. This keeps the scheduler state accurate and avoids leaving stale scheduled events behind.

In the examples visualization tests, it adds retry logic around Playwright screenshots so transient Solara DOM re-renders do not fail the test with `Element is not attached to the DOM`.

## Motivation
These changes make the test suite more reliable and better aligned with Mesa's public APIs.

The action tests now exercise real scheduler behavior instead of bypassing it through private state mutation. The visualization test helper now tolerates harmless frontend rerenders that were causing flaky failures in CI.

## Testing
- `python -m pytest tests/experimental/test_actions_interruption.py`
- targeted verification for `tests/examples/test_examples_viz.py -k pd_grid_model`

### DIFF
--- a/tests/examples/test_examples_viz.py
+++ b/tests/examples/test_examples_viz.py
@@ -28,6 +28,25 @@ from mesa.visualization.components.matplotlib_components import (
 )
 
 
+def take_stable_screenshot(
+    locator: playwright.sync_api.Locator, retries: int = 3
+) -> bytes:
+    """Retry element screenshots when the DOM swaps the target node."""
+    last_error = None
+
+    for _ in range(retries):
+        locator.wait_for(state="visible")
+        try:
+            return locator.screenshot()
+        except playwright.sync_api.Error as error:
+            if "Element is not attached to the DOM" not in str(error):
+                raise
+            last_error = error
+
+    assert last_error is not None
+    raise last_error
+
+
 def run_model_test(
     model,
     agent_portrayal,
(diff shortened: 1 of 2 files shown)


quaquel:
test_examples_viz has been updated. Please check how this fix compares to the update and see if this PR is still relevant. If so, please explain why.

SumeetDUTTA:
I compared my change against the latest branch state at **2d0b1ae** and reviewed **tests/examples/test_examples_viz.py**.

Against main, my patch still adds the screenshot retry guard for the DOM-detach race and routes the screenshot call sites through it.

Against **test/test-stability-fixes**, the file content is identical, so this specific fix is already included there.

Given that, if this PR targets **test/test-stability-fixes**, it is redundant and can be closed; it is only still relevant if re-targeted to main.

tests/examples/test_examples_viz.py
```diff
@@ -28,6 +28,25 @@
...
+def take_stable_screenshot(
+    locator: playwright.sync_api.Locator, retries: int = 3
+) -> bytes:
+    """Retry element screenshots when the DOM swaps the target node."""
+    last_error = None
+
+    for _ in range(retries):
+        locator.wait_for(state="visible")
+        try:
+            return locator.screenshot()
```
falloficaruss, line 40: `take_stable_screenshot()` only retries around `locator.screenshot()`, but the preceding `locator.wait_for(state="visible")` can hit the same DOM-detach race. If the Solara rerender swaps the `` before `wait_for()` completes, the helper still fails on the first attempt, so the flake is not fully covered. I’d wrap both calls in the retry path or catch the same detach failure around `wait_for()` too.

falloficaruss (commented):
This new suite mostly exercises `Action.start()` directly and duplicates scenarios already covered in `tests/experimental/test_actions.py`. Because it bypasses the public `Agent.start_action()` / `cancel_action()` path, it won’t catch regressions in `current_action` bookkeeping or the agent-facing API, which is the behavior users actually rely on. I’d fold the useful cases into the existing file and keep the tests on the public path.

quaquel:
test are failing so I am closing this.

PR #3669, opened 2026-03-31
Fix agent portrayal warning in space_renderer

Imen-source:
This PR fixes an issue in `space_renderer.py` where agent portrayal could trigger warnings when missing or malformed.

The fix adds a minimal safeguard to ensure portrayal is handled safely while preserving existing behavior.

Only `space_renderer.py` is modified to keep the change focused and minimal.


souro26:
Please only add the relevant changes in the PR. There are a lot of unrelated changes.

falloficaruss:
@quaquel This PR can be closed.

PR #3678, opened 2026-04-01
Pathfinding for Stacked Space

codebreaker32:
Pathfinding for adding stacked spaces architecture to Mesa. This builds upon the foundational ideas discussed in #2585 and #3458.

1. It introduces a bidirectional sync mechanism allowing agents to seamlessly operate across a continuous physical space (the primary space) and discrete patches (like `OrthogonalMooreGrid`), solving the multi-space routing challenge.
2. It introduces an unified agent `SpatialAgent` that automatically pushes updates from its continuous `position` downward to all registered discrete spaces.
3. Which example model can be used to best evaluate it?
4. Any suggestions on API ergonomics? Architecture? Naming?

Please see [Architectural Diagram](https://github.com/mesa/mesa/discussions/3458#discussioncomment-16129090)

### DIFF
--- a/mesa/discrete_space/discrete_space.py
+++ b/mesa/discrete_space/discrete_space.py
@@ -19,7 +19,7 @@ import warnings
 from abc import ABC, abstractmethod
 from functools import cached_property
 from random import Random
-from typing import TypeVar
+from typing import Any, TypeVar
 
 import numpy as np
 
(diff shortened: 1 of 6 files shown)


EwoutH:
Thanks for the PR. Always appreciate a bit of pathfinding.

I think we need to get the primitives in place first. See https://github.com/mesa/mesa/discussions/3458#discussioncomment-16508056.

PR #3287, opened 2026-02-12
poc: implement Locatable protocol for spatial agents

codebyNJ:
## Summary

Proof of concept for the `Locatable` protocol tracked in #3132 (Space Architecture section).

Follows @quaquel's composition-over-inheritance vision from [#2585](https://github.com/projectmesa/mesa/discussions/2585#discussioncomment-11418207):
- `Locatable` protocol (like `HasCellProtocol`) defines the interface
- `HasPosition` mixin (like `HasCell`) provides writable `position` via composition
- Base `Agent` does NOT have `position` -- agents opt in by mixing in `HasPosition`
- `self.position = new_pos` works (agent-centric API per [PR #3043 comment](https://github.com/mesa/mesa/pull/3043#issuecomment-3700129146))

### What satisfies Locatable?

| Class | Locatable? | How |
|-------|-----------|-----|
| `Agent` (base) | No | By design -- composition, not inheritance |
| `Agent + HasPosition` | Yes | Mixin provides writable position |
| `CellAgent` | Yes | Derives from `cell.coordinate` |
| `FixedAgent` | Yes | Derives from `cell.coordinate` |
| `ContinuousSpaceAgent` | Yes | Existing numpy-backed property |
| `Cell` | Yes | Alias for `coordinate` |

### Files changed

| File | Change |
|------|--------|
| `mesa/protocols.py` (new) | `Locatable` protocol, `HasPosition` mixin, `PositionLike` type |
| `mesa/discrete_space/cell.py` | Added `position` property |
| `mesa/discrete_space/cell_agent.py` | Added `position` to `CellAgent` and `FixedAgent` |
| `mesa/agent.py` | No changes to base Agent (intentional) |
| `mesa/__init__.py` | Exports `Locatable`, `HasPosition`, `PositionLike` |

### Design alignment

This POC mirrors the existing `HasCellProtocol`/`HasCell` pattern already in Mesa:

- HasCellProtocol (Protocol)  <-->  Locatable (Protocol)
- HasCell (mixin)             <-->  HasPosition (mixin)
- CellAgent(Agent, HasCell)   <-->  MyAgent(Agent, HasPosition0

PR #3693, opened 2026-04-04
add Locatable protocol

souro26:
### Summary
Adds a `Locatable` protocol and `HasPosition` mixin to formalize the `.position` interface across agents and cells in Mesa. Closes the task in #3132 .

### Motive
`.position` exists on agents and cells but there's no formal contract for it. Stacked spaces #3458 needs this as a foundation.

### Implementation
- Added `mesa/protocols.py` with `Locatable` protocol and `HasPosition` mixin
- Added `position` property to `CellAgent` and `FixedAgent` 
- Removed redundant `HasPositionProtocol` from `continuous_space_agents.py`

### Usage Examples
```python
from mesa import Agent, Model
from mesa.protocols import Locatable, HasPosition
import numpy as np

class MyAgent(Agent, HasPosition):
    pass

model = Model()
agent = MyAgent(model)
agent.position = np.array([3.0, 4.0])

def distance(a: Locatable, b: Locatable) -> float:
    return float(np.linalg.norm(np.array(a.position) - np.array(b.position)))

cell = Cell((3, 4))
print(distance(agent, cell))  # agent and cell, same function
```


EwoutH:
Thanks for the PR. I skimmed through it, on the surface it looks like a clean implementation.

@quaquel and/or @codebreaker32 could you review? Including if we're even building the right thing, at the right place, using the right constructs, etc. (see also https://github.com/mesa/mesa/discussions/3458#discussioncomment-16508056)

codebreaker32:
I would be curious to know how it differs from #3287 because you acknowledged #3287 in few places so why we need this PR? If both are serving same purpose, I think we should resolve that one first

souro26:
#3287 sets up Locatable and HasPosition, but it is opt in. In this PR, agents like CellAgent and FixedAgent already follow it, so `.position` behaves the same way for them instead of each type handling it differently.

codebreaker32:
Let's wait for @quaquel to chime in.

codebreaker32 (commented):
I took a close look at both PRs, and they actually appear to serve the exact same purpose. In #3287, CellAgent and FixedAgent already natively follow the Locatable protocol via duck typing. There are tests in #3287 that explicitly prove this. You can read more about Protocols here https://typing.python.org/en/latest/spec/protocol.html

EwoutH:
From my quick glance this PR seems a lot cleaner and more atomic than #3287. What do you think?

Anyway @quaquel owns this part of Mesa currently, so let's indeed wait for his perspective.

codebreaker32:
> From my quick glance this PR seems a lot cleaner and more atomic than #3287. What do you think?

I am fine either way

quaquel:
I am closing this. I hope to pick up the spaces stuff again in the coming months and prefer to take a clean look. 


I like the PR as such, but I am not sure about how to organize the protocols in MESA. I am more inclined to keep Locatable inside the spaces, but need to think this through.

PR #3697, opened 2026-04-07
Fix: CellAgent inheritance initialization with MetaAgent

codebyNJ:
### Pre-PR Checklist

### Summary
Fixes `MetaAgent` initialization order to properly support inheritance from descriptor-based parent classes like `CellAgent`. Spatial attributes (`cell`, `pos`) are now correctly initialized instead of remaining `None`.

### Bug / Issue
When creating a meta-agent inheriting from `CellAgent`, spatial attributes were not set correctly because parent class initialization happened before explicit attributes were applied. The descriptor-based `_mesa_cell` attribute requires strict initialization ordering.

**Expected:** Meta-agent inheriting from `CellAgent` has `cell` and `pos` set to intended values.  
**Actual:** Spatial attributes remained `None` after initialization.

### Implementation
Modified `MetaAgent.__init__()` in `mesa/experimental/meta_agents/meta_agent.py`:

1. Added `initial_attributes` parameter (optional, defaults to `None`)
2. Apply attributes **before** `super().__init__(model)` using `object.__setattr__()`
3. Updated `create_meta_agent()` Path 3 to pass `initial_attributes=meta_attributes`

**Key change:** Initial attributes are set before parent class initialization to ensure descriptors work correctly.

### Testing
- All 20 existing meta-agent tests pass ✓
- No breaking changes, fully backward compatible
- `initial_attributes` parameter is optional

### Additional Notes
- Minimal, focused fix (initialization order only)
- No API changes, backward compatible
- Solves warehouse model spatial meta-agent pattern
- Complements #3538 (explicit meta_attributes safety)

### DIFF
--- a/tests/experimental/test_meta_agents.py
+++ b/tests/experimental/test_meta_agents.py
@@ -3,21 +3,14 @@
 import pytest
 
 from mesa import Agent, Model
+from mesa.discrete_space.cell_agent import CellAgent
+from mesa.discrete_space.grid import OrthogonalMooreGrid
 from mesa.experimental.meta_agents.meta_agent import (
     MetaAgent,
     create_meta_agent,
     evaluate_combination,
     find_combinations,
 )
-
-# Optional imports for CellAgent tests
-try:
-    from mesa.discrete_space.cell_agent import CellAgent
-    from mesa.discrete_space.grid import OrthogonalMooreGrid
-
-    HAS_CELLAGENT = True
-except ImportError:
-    HAS_CELLAGENT = False
 
 
 class CustomAgent(Agent):
@@ -417,9 +410,6 @@ def test_create_meta_agent_repeated_instances_with_descriptor_parent():
     Verifies that initial_attributes are applied correctly for the second and subsequent
     instances of the same meta-agent class.
     """
-    if not HAS_CELLAGENT:
-        pytest.skip("CellAgent or OrthogonalMooreGrid not available")
-
     model = Model()
     grid = OrthogonalMooreGrid((10, 10), random=model.random)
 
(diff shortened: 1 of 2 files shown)


codebyNJ:
Sorry the #3696 had some extra commit history therefore I have raised this clean PR.

falloficaruss (commented):
The fix only covers new meta-agent classes, not later instances of the same class. In `meta_agent.py`, Path 2 still does `agent_class(model, agents)` and only applies `meta_attributes` afterward. That means the exact `CellAgent`/descriptor ordering bug still reproduces when you create a second meta-agent of an already-registered class with no overlapping constituents. The new `initial_attributes` plumbing is only used in Path 3, so this PR doesn’t fully fix the reported behavior for repeated `create_meta_agent()` calls with the same class name.

codebyNJ:
> The fix only covers new meta-agent classes, not later instances of the same class. In `meta_agent.py`, Path 2 still does `agent_class(model, agents)` and only applies `meta_attributes` afterward. That means the exact `CellAgent`/descriptor ordering bug still reproduces when you create a second meta-agent of an already-registered class with no overlapping constituents. The new `initial_attributes` plumbing is only used in Path 3, so this PR doesn’t fully fix the reported behavior for repeated `create_meta_agent()` calls with the same class name.

Thanks for noticing the issue @falloficarus22. I have fixed it. You can review it again now.

tests/experimental/test_meta_agents.py
```diff
@@ -399,3 +401,54 @@ def test_meta_agent_remove_with_multiple_memberships():
...
+
+def test_create_meta_agent_repeated_instances_with_descriptor_parent():
+    """Test creating multiple instances of same meta-agent class with descriptor-based parent.
+
+    This tests Path 2 (existing meta-agent class instantiation) with a descriptor-based parent.
+    Verifies that initial_attributes are applied correctly for the second and subsequent
+    instances of the same meta-agent class.
+    """
+    model = Model()
+    grid = OrthogonalMooreGrid((10, 10), random=model.random)
```
falloficaruss, line 414: The new test uses `OrthogonalMooreGrid((10, 10), random=model.random)` and indexes `grid[2, 2]` and `grid[5, 5]`. This matches expected usage, but if `OrthogonalMooreGrid` ever changes its indexing or constructor signature, this test will be fragile. It’s fine for now, just a note.

codebyNJ: Yaa makes sense tho. I want some confirmation from @tpike3 and suggestions does this match his vision towards meta agents.

tpike3 (approved):
LGTM! Thanks, that was a good bug to find. 

Base don the future updates of meta-agents I am ok with the test being brittle for now.

PR #3698, opened 2026-04-08
fix: raise clear TypeError when SolaraViz receives a model class instead of an instance

### LINKED ISSUE #3454: SolaraViz crashes with cryptic errors when passed a Model class instead of instance
# SolaraViz should validate that model is an instance, not a class

### Bug Summary
When users accidentally pass a Model *class* instead of an *instance* to SolaraViz (e.g., `SolaraViz(MyModel, ...)`), it causes cryptic UI failures deep within the Solara/React components instead of raising a clear, immediate error.

### Root Cause
In `mesa/visualization/solara_viz.py`, `SolaraViz` expects `model: Model | solara.Reactive[Model]` as its first parameter. However, there's no validation to ensure the provided argument is an *instance* rather than a *class type*.

When a class is passed:
1. Solara blindly converts it to `solara.use_reactive(model)`
2. Execution proceeds down the component tree
3. Eventually, code tries to call `.step()` or access instance attributes on the class definition itself
4. This results in confusing tracebacks like `TypeError: step() missing 1 required positional argument: 'self'` or `AttributeError: type object 'MyModel' has no attribute 'schedule'`

This is especially problematic for users migrating from the old Tornado-based `ModularServer`, which accepted the Model *class* (e.g., `ModularServer(MyModel, ...)`).

### Steps to Reproduce
```python
from mesa import Model, Agent

class MyModel(Model):
    def __init__(self):
        super().__init__()
        self.running = True
    
    def step(self):
        print("Step called")

# Wrong: passing the class instead of an instance
page = SolaraViz(
    MyModel,  # Should be: MyModel()
    components=[],
    model_params={}
)

# Attempting to run this causes cryptic errors deep in the rendering pipeline
```

### Expected Behavior
`SolaraViz.__init__()` should immediately validate the `model` parameter and raise a clear, actionable error:
```python
TypeError: SolaraViz requires an initialized model instance, not a model class.
Did you mean: SolaraViz(MyModel(), ...) instead of SolaraViz(MyModel, ...)?
```

### Actual Behavior
The error manifests deep in the component tree with confusing tracebacks like:
```
TypeError: step() missing 1 required positional argument: 'self'
```

or
```
AttributeError: type object 'MyModel' has no attribute 'schedule'
```

These errors don't indicate the actual problem (passing a class instead of an instance), wasting significant debugging time.

### Environment
- Mesa version: 4.0.0.dev0
- Python version: 3.14.3

### Proposed Fix
I'm willing to submit a PR to fix this. The fix would add validation in `SolaraViz.__init__()`:
```python
def __init__(self, model, ...):
    # Validate model is an instance, not a class
    if isinstance(model, type):
        raise TypeError(
            "SolaraViz requires an initialized model instance, not a model class. "
            f"Did you mean: SolaraViz({model.__name__}(), ...) instead of SolaraViz({model.__name__}, ...)?"
        )
    # ... rest of initialization
```

This provides an immediate, clear error message that guides users to the correct usage.

---

**Note**: This is part 2 of 3 in a series of SolaraViz DX improvements:
- Part 2: Class instantiation validation (this issue)
- Part 3: will update

Tejasv-Singh:
## Summary
`SolaraViz` crashes with cryptic errors when passed a Model class instead of an instance. This PR adds immediate validation that raises a clear, actionable `TypeError` pointing the user to the fix.

## Bug / Issue
Closes #3454

Users migrating from `ModularServer` (which accepted a class) commonly write `SolaraViz(MyModel, ...)` instead of `SolaraViz(MyModel(), ...)`. Previously this surfaced deep in Solara's rendering pipeline as `TypeError: step() missing 1 required positional argument: 'self'` or `AttributeError: type object 'MyModel' has no attribute 'schedule'`, with no indication of the actual cause.

## Implementation
Added validation at the top of the `SolaraViz` component body (before the reactive conversion) in `mesa/visualization/solara_viz.py`. Two cases are handled:

- `isinstance(model, type)` : class passed directly
- `isinstance(model, solara.Reactive) and isinstance(model.value, type)` : class wrapped in a reactive

Both raise a `TypeError` with a hint: `Did you mean: SolaraViz(MyModel(), ...) instead of SolaraViz(MyModel, ...)?`

## Testing
Added `tests/visualization/test_solara_viz_class_validation.py` with 5 tests covering both error cases, the error message content, and the happy paths (valid instance and valid reactive instance). All 5 new tests pass, and all 17 existing tests in `test_solara_viz.py` continue to pass.

> **Note for reviewers:** `@solara.component` is lazy, the function body runs only at render time. Tests therefore use `solara.render(element, handle_error=False)` to trigger the component and catch the error correctly.

## Additional Notes
This is Part 2 of a series of SolaraViz DX improvements. Part 1 is #3453.


souro26:
Looks good to me, just keep the tests in the relevant test file instead of a new one.

Tejasv-Singh:
Thanks for the review, @souro26! :raised_hands:  Good call,  I've moved all 5 tests into the existing `test_solara_viz.py` and deleted the separate file. The change is now pushed. Do let me know if there's anything else that I can do :raised_hands:

souro26 (approved):

tests/visualization/test_solara_viz.py
tpike3, line 1: Thanks @Tejasv-Singh --- This is could but could you reduce the number of tests; 5 tests for this one validation seems a bit much

Tejasv-Singh: Thanks for the feedback! Consolidated the 5 tests down to 2 following the same pattern as the _validate_model_params tests in the file.

tpike3 (approved):
LGTM

Issue #3700, opened 2026-04-10
stale double KDTree rebuild in remove_cell + lazy KDTree rebuild


Krishsharma179:
Description
When removing a spatial cell dynamically from a `network` grid using `remove_cell`, the framework forces a full rebuild of the `scipy.spatial.KDTree` twice. Furthermore, the first rebuild operates on a `stale` list that still contains the deleted "ghost" cell, wasting computational cycles on invalid data

The Bug
In `mesa.discreate_space.network.py` the `remove_cell` 
implementation looks like this:
```
    def remove_cell(self, cell: Cell):
        """Remove a cell from the space."""
        super().remove_cell(cell)
        self.G.remove_node(cell.coordinate)
        
        # BUG: This rebuilds the entire KDTree BEFORE the cell is removed 
        # from the _kdtree_cells list. The tree is built including the dead cell.
        self._rebuild_kdtree() 

        if cell._position is not None:
            self._kdtree_cells.remove(cell)
            
            # CORRECT: Rebuilds the tree properly a second time.
            self._rebuild_kdtree()
```
It removes from the graph node but since the __rebuild function look at the kdtree_cell list the graph is built on the stale data
Kdtree implementation
````
def _rebuild_kdtree(self) -> None:
        """Rebuild the KD-Tree."""
        if self._kdtree_cells:
            positions = np.array([c._position for c in self._kdtree_cells])
            self._kdtree = KDTree(positions)
        else:
            self._kdtree = None
````
so the proposed solution will be :
```
    def remove_cell(self, cell: Cell):
        super().remove_cell(cell)
        self.G.remove_node(cell.coordinate)

        # no building of kdtree on the stale data 
        
       # if the cell postion is not none then only we build the kdtree with the updated data
        if cell._position is not None:
            self._kdtree_cells.remove(cell)
            self._rebuild_kdtree()

  ```

The Optimal Solution: Lazy KDTree Rebuilding
Even with the bug fixed, calling `remove_cell` in side a loop of 1,000 cells will force the `KDTree` to rebuild from scratch 1,000 times, freezing the simulation.

We can solve this easily using Lazy Evaluation. By setting a dirty flag and delaying the heavy math until the user actually queries the space, we ensure the tree only builds once per batch of changes.

1. Add a flag to `__init__`
 ```
        self._kdtree_cells = []
        self._kdtree_dirty = False
```
2. Just set the flag in `add_cell`/ `remove_cell`

```
[...]

PR #3701, opened 2026-04-10
perf:lazy KDTree rebuild + tests

Krishsharma179:
### Pre-PR Checklist

### Summary
Primary: fixes stale/redundant rebuild behavior in `remove_cell` .
Secondary: lazy rebuild is the implementation strategy to avoid repeated unnecessary rebuilds.

Fixes a stale/redundant KDTree rebuild issue in `NetworkGrid` where `remove_cell()` could trigger an eager rebuild on outdated `_kdtree_cells` data. As part of the fix, `KDTree` rebuilds are now deferred (lazy) until `find_nearest_cell()` is actually called, eliminating redundant work during bulk mutations.

### Bug / Issue
#3700 In the previous implementation,` add_cell()` and `remove_cell()` both triggered eager `KDTree` rebuilds. In `remove_cell()`, this could result in a rebuild on stale or already-modified `_kdtree_cells` data, causing a double/redundant rebuild. 
### Implementation
1.Added `self._kdtree_dirty` = False flag in `__init__`.
2.`_rebuild_kdtree()` now resets `_kdtree_dirty = False ` after every rebuild.
3.`remove_cell()`: Removes from `_kdtree_cells` and sets `_kdtree_dirty = True`. Eager rebuild removed, eliminating the previous stale/redundant rebuild on already-modified `_kdtree_cells` data.
4.`add_cell()`: For spatial cells, appends to `_kdtree_cells`  and sets `_kdtree_dirty = True` .
5.`find_nearest_cell()` checks the dirty flag and triggers a rebuild on demand before querying.

### Testing
`tests/discrete_space/test_spatial_lookups.py`
Three new tests added using `unittest.mock.patch.object(..., wraps=...)` to spy on `_rebuild_kdtree` call count:

`test_network_lazy_rebuild_deferred_until_query`  verifies no rebuild occurs during mutation-only phase; rebuild triggered only on first`find_nearest_cell(` call.
`test_network_lazy_rebuild_batches_mutations_to_single_rebuild`  verifies multiple mutations result in exactly one rebuild on query, not one per mutation.
`test_network_non_spatial_cell_mutation_does_not_dirty_kdtree ` verifies that adding/removing non-spatial cells does not set the dirty flag or trigger any rebuild
### Additional Notes
This change is purely a fix to rebuild timing. No public API changes. Graph topology updates remain eager and unaffected. The lazy rebuild is an internal optimization that emerges naturally from fixing the stale rebuild path in `remove_cell()`.

### DIFF
--- a/mesa/discrete_space/network.py
+++ b/mesa/discrete_space/network.py
@@ -105,7 +105,7 @@ class Network(DiscreteSpace[Cell]):
             self._kdtree = KDTree(positions)
         else:
             self._kdtree = None
-        self._kdtree_dirty = False  #set to false after rebuilding  every time
+        self._kdtree_dirty = False  # set to false after rebuilding  every time
 
     def _connect_cells(self) -> None:
         for cell in self.all_cells:
@@ -134,8 +134,7 @@ class Network(DiscreteSpace[Cell]):
         Raises:
             ValueError: If network is not spatial
         """
-        
-        #build the kd-tree once if the _kdtree_dirty flag is true 
+        # build the kd-tree once if the _kdtree_dirty flag is true
 
         if self._kdtree_dirty:
             self._rebuild_kdtree()
--- a/mesa/discrete_space/network.py
+++ b/mesa/discrete_space/network.py
@@ -105,7 +105,7 @@ class Network(DiscreteSpace[Cell]):
             self._kdtree = KDTree(positions)
         else:
             self._kdtree = None
-        self._kdtree_dirty = False  #set to false after rebuilding  everytime
+        self._kdtree_dirty = False  #set to false after rebuilding  every time
 
     def _connect_cells(self) -> None:
         for cell in self.all_cells:
--- a/mesa/discrete_space/network.py
+++ b/mesa/discrete_space/network.py
@@ -68,6 +68,7 @@ class Network(DiscreteSpace[Cell]):
             )
 
         self._kdtree_cells = []
+        self._kdtree_dirty = False  # True when KDTree needs a rebuild.
         positions_for_tree = []
 
         # Create cells and gather KD-Tree data simultaneously
@@ -104,6 +105,7 @@ class Network(DiscreteSpace[Cell]):
             self._kdtree = KDTree(positions)
         else:
             self._kdtree = None
+        self._kdtree_dirty = False  #set to false after rebuilding  everytime
 
     def _connect_cells(self) -> None:
         for cell in self.all_cells:
@@ -118,6 +120,11 @@ class Network(DiscreteSpace[Cell]):
 
         Only works for spatial networks (networks with node positions).
 
+        Note:
+            The KD-Tree index is rebuilt lazily. If cells were added or removed
+            since the previous spatial query, this method rebuilds the KD-Tree
+            before performing the nearest-neighbor lookup.
+
         Args:
             position: Physical position [x, y]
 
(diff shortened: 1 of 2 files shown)


mesa/discrete_space/network.py
```diff
@@ -140,17 +152,16 @@ def add_cell(self, cell: Cell):
...
         """Remove a cell from the space."""
         super().remove_cell(cell)
         self.G.remove_node(cell.coordinate)
-        self._rebuild_kdtree()
 
-        if cell._position is not None:
+        if cell._position is not None and cell in self._kdtree_cells:
             self._kdtree_cells.remove(cell)
-            self._rebuild_kdtree()
+            self._kdtree_dirty = True
```
Krishsharma179, line 165: Removed the eager `self._rebuild_kdtree()` call here. Previously,` _rebuild_kdtree()` was called before removing the cell from `_kdtree_cells`, causing it to rebuild on stale data. Now we mark `_kdtree_dirty = True` and defer the rebuild until `find_nearest_cell()` is actually called.

mesa/discrete_space/network.py
```diff
@@ -127,6 +134,11 @@ def find_nearest_cell(self, position: np.ndarray) -> Cell:
         Raises:
             ValueError: If network is not spatial
         """
+        # build the kd-tree once if the _kdtree_dirty flag is true
+
+        if self._kdtree_dirty:
+            self._rebuild_kdtree()
+
```
Krishsharma179, line 141: This is the core of the lazy rebuild strategy. Instead of rebuilding `KDTree` on every `add_cell()` or `remove_cell()` , we defer the rebuild here only when `find_nearest_cell()` is actually called and `_kdtree_dirty` is True. This ensures at most one rebuild per query, regardless of how many mutations happened before it.

mesa/discrete_space/network.py
```diff
@@ -68,6 +68,7 @@ def __init__(
             )
 
         self._kdtree_cells = []
+        self._kdtree_dirty = False  # True when KDTree needs a rebuild.
```
Krishsharma179, line 71: Dirty flag initialized as False at construction no mutations have occurred yet. Set to True on any spatial `add_cell()` or `remove_cell()` call, and reset to False after `_rebuild_kdtree()` completes.

mesa/discrete_space/network.py
```diff
@@ -104,6 +105,7 @@ def _rebuild_kdtree(self) -> None:
             self._kdtree = KDTree(positions)
         else:
             self._kdtree = None
+        self._kdtree_dirty = False  # set to false after rebuilding  every time
```
Krishsharma179, line 108: `_kdtree_dirty` reset to False after rebuild completes, so subsequent calls to `find_nearest_cell()` skip the rebuild until the next mutation occur

Krishsharma179 (commented):
This is the self review of my PR It Introduces lazy `KDTree` rebuild via `_kdtree_dirty` flag. Removes eager `_rebuild_kdtree()` calls from `add_cell()` and `remove_cell()` , fixing stale rebuild risk in the remove flow. Rebuild now deferred until `find_nearest_cell()` is called. Three new tests added. All 14 CI checks passin

codebreaker32 (approved):
Thanks for this optimisation. LGTM!

Krishsharma179:
@codebreaker32 Thank you for the review!

quaquel (approved):

PR #3703, opened 2026-04-12
fixed:_get_interval and added the tests

### LINKED ISSUE #3702: Callable Schedule interval can return 0 in EventGenerator, bypassing validation
**Describe the bug**
`Schedule` validates fixed numeric intervals with `interval <= 0` (must be strictly > 0), but `EventGenerator._get_interval()` only rejects callable interval results when `< 0`.  
This means a callable interval returning `0` is accepted, which is inconsistent with fixed-interval validation and can schedule events at the same simulation time repeatedly.

**Expected behavior**
Callable and non-callable intervals should follow the same rule: interval must be strictly greater than `0`.  
If a callable returns `0` (or negative), a `ValueError` should be raised.

**To Reproduce**
1. Create a model.
2. Create a `Schedule` with a callable interval that returns `0`.
3. Start an `EventGenerator` using that schedule.
4. Observe that no `ValueError` is raised for interval `0`.

Minimal repro:
```python
from mesa.time.events import EventGenerator, EventList, Schedule


class DummyModel:
	def __init__(self):
		self.time = 0.0
		self._event_list = EventList()
		self._event_generators = set()


def noop():
	pass


# Case 1: fixed interval=0 is correctly rejected.
try:
	Schedule(interval=0)
	print("fixed interval=0 -> NOT rejected (unexpected)")
except ValueError as exc:
	print(f"fixed interval=0 -> rejected (expected): {exc}")


# Case 2: callable interval returning 0 currently passes validation.
schedule = Schedule(interval=lambda _m: 0)
generator = EventGenerator(DummyModel(), noop, schedule)
interval = generator._get_interval()
print(f"callable interval returns: {interval}")

if interval == 0:
	print("BUG VERIFIED: callable interval can return 0")
else:
	print("No bug reproduced")
```
Current code:
```python
def _get_interval(self) -> float | int:
        """Get the next interval value."""
        if callable(self.schedule.interval):
            interval = self.schedule.interval(self.model)
            if interval < 0:
                raise ValueError(f"Interval must be > 0, got {interval}")
            return interval
        return self.schedule.interval
```
proposed solution :
```python
def _get_interval(self) -> float | int:
        """Get the next interval value."""
        if callable(self.schedule.interval):
            interval = self.schedule.interval(self.model)
            if interval <= 0:  #changed here
                raise ValueError(f"Interval must be strictly > 0, got {interval}")
            return interval
        return self.schedule.interval
```
I’ll be opening a PR shortly; the implementation is ready, and any feedback or suggestions are welcome

Krishsharma179:
> [!NOTE]

### Pre-PR Checklist

### Summary
Fixes inconsistent interval validation in `EventGenerator` for callable schedules.  
Previously, callable intervals could return `0` without error, while fixed numeric intervals already rejected `<= 0`.  
This could lead to repeated rescheduling at the same simulation timestamp and a potential infinite loop in event processing.

### Bug / Issue
Closes #3702 

`Schedule.__post_init__` enforces `interval > 0` for non-callable values, but `EventGenerator._get_interval()` only partially validated callable results.  
Expected behavior: callable and non-callable intervals should follow the same rule (`> 0`).  
Actual behavior: a callable returning `0` was accepted.

### Implementation
- Updated `EventGenerator._get_interval()` in `mesa/time/events.py`:
  - validation now raises `ValueError` when callable interval returns `<= 0`
  - error message clarifies interval must be strictly greater than zero
- Updated docstring for `_get_interval()` to document raised `ValueError`.

### Testing
Added regression tests in `tests/time/test_events.py`:
- `test_callable_interval_zero_raises_in_get_interval`
  - directly verifies `_get_interval()` raises `ValueError` for callable returning `0`
- `test_callable_interval_zero_raises`
  - verifies runtime path raises `ValueError` during model execution/rescheduling

These complement the existing negative-interval test.

### Additional Notes
- Scope is intentionally minimal and limited to validation consistency for callable intervals.
- This change prevents zero-interval rescheduling from creating a potential infinite same-timestamp event loop.

### DIFF
--- a/mesa/time/events.py
+++ b/mesa/time/events.py
@@ -260,11 +260,16 @@ class EventGenerator:
         return self._current_event.time
 
     def _get_interval(self) -> float | int:
-        """Get the next interval value."""
+        """Return the next interval value from the schedule.
+
+        Raises:
+            ValueError: If a callable interval returns a negative value or zero.
+            
+        """
         if callable(self.schedule.interval):
             interval = self.schedule.interval(self.model)
-            if interval < 0:
-                raise ValueError(f"Interval must be > 0, got {interval}")
+            if interval <= 0:
+                raise ValueError(f"Interval must be strictly > 0, got {interval}")
             return interval
         return self.schedule.interval
 
--- a/mesa/time/events.py
+++ b/mesa/time/events.py
@@ -264,7 +264,7 @@ class EventGenerator:
 
         Raises:
             ValueError: If a callable interval returns a negative value or zero.
-            
+
         """
         if callable(self.schedule.interval):
             interval = self.schedule.interval(self.model)
--- a/tests/time/test_events.py
+++ b/tests/time/test_events.py
@@ -613,6 +613,23 @@ class TestEventGeneratorExecution:
         with pytest.raises(ValueError):
             model.run_for(3)
 
+    def test_callable_interval_zero_raises_in_get_interval(self, setup):
+        """Test if callable interval raises ValueError immediately in _get_interval."""
+        model, fn = setup
+        gen = EventGenerator(model, fn, Schedule(start=1, interval=lambda m: 0))
+
+        with pytest.raises(ValueError):
+            gen._get_interval()
+
+    def test_callable_interval_zero_raises(self, setup):
+        """Test if callable interval raises exception if return is zero."""
+        model, fn = setup
+        gen = EventGenerator(model, fn, Schedule(start=1, interval=lambda m: 0))
+        gen.start()
+
+        with pytest.raises(ValueError):
+            model.run_for(3)
+
     def test_functools_partial(self, setup):
         """Test using functools.partial for arguments."""
         model, fn = setup


mesa/time/events.py
```diff
@@ -260,11 +260,16 @@ def next_scheduled_time(self) -> float | None:
...
+        Raises:
+            ValueError: If a callable interval returns a negative value or zero.
+
+        """
         if callable(self.schedule.interval):
             interval = self.schedule.interval(self.model)
-            if interval < 0:
-                raise ValueError(f"Interval must be > 0, got {interval}")
+            if interval <= 0:
+                raise ValueError(f"Interval must be strictly > 0, got {interval}")
```
Krishsharma179, line 272: This change fixes the previous behavior where a callable interval returning 0 could bypass validation and reschedule the model at the same simulation time repeatedly

tests/time/test_events.py
```diff
@@ -613,6 +613,23 @@ def test_callable_interval_negative_raises(self, setup):
         with pytest.raises(ValueError):
             model.run_for(3)
 
+    def test_callable_interval_zero_raises_in_get_interval(self, setup):
+        """Test if callable interval raises ValueError immediately in _get_interval."""
+        model, fn = setup
+        gen = EventGenerator(model, fn, Schedule(start=1, interval=lambda m: 0))
+
+        with pytest.raises(ValueError):
+            gen._get_interval()
```
Krishsharma179, line 622: This test I added to see if the __get_interval raises error if the interval function returns zero

tests/time/test_events.py
```diff
@@ -613,6 +613,23 @@ def test_callable_interval_negative_raises(self, setup):
...
+
+    def test_callable_interval_zero_raises(self, setup):
+        """Test if callable interval raises exception if return is zero."""
+        model, fn = setup
+        gen = EventGenerator(model, fn, Schedule(start=1, interval=lambda m: 0))
+        gen.start()
+
+        with pytest.raises(ValueError):
+            model.run_for(3)
+
```
Krishsharma179, line 632: verifies runtime path raises `ValueError` during model execution/rescheduling

Krishsharma179 (commented):
This fix makes callable interval validation consistent with fixed intervals, and the added regression tests cover both the direct validation path and the runtime scheduling path. This also prevents zero-interval rescheduling from repeatedly firing at the same simulation time.

Krishsharma179:
@codebreaker32 Please do a  peer-review of this Pr ,thanks

quaquel:
I disagree on the bug and so on the fix. We can discuss further in the issue if you have good arguments.

PR #3671, opened 2026-03-31
Fix ValueError in get_k_nearest_agents when k >= number of agents

Vartika222:
### Pre-PR Checklist
- [x] This PR is a bug fix, not a new feature or enhancement.

### Summary
`get_k_nearest_agents` crashed when `k >= n` (number of agents) or the space was empty, due to `np.argpartition` receiving an out-of-range index.

### Bug / Issue
Two problems:
- No guard against `k >= n` or empty space, causing `np.argpartition` to raise a `ValueError`
- Off-by-one: `np.argpartition` takes a zero-based index so it should be `k-1`, not `k`

### Implementation
- Early return for empty space or `k <= 0`
- Clamp `k` to `n` when `k > n`, with a `UserWarning` so the caller knows
- Fix `np.argpartition(dists, k-1)`
- Update docstring to document the `k > n` behaviour

### Testing
Four new tests: `k > n` (checks warning too), empty space, `k=0`, and `k` exactly equal to population. All existing tests pass.

### Additional Notes
Also removed an unrelated line that had crept into `test_examples_viz.py`, and aligned new tests to use `Model(rng=42)` / `model.random` per reviewer feedback.

PR #3717, opened 2026-05-18
Fix get_k_nearest_agents crash when k >= number of agents

AshNicolus:
## What

`get_k_nearest_agents` raises `ValueError` from `np.argpartition` whenever
`k >= number of agents in the space`, because `np.argpartition(arr, kth)`
requires `kth < arr.size`.

Additionally, `agent_positions` was declared as a bare type annotation in
`__init__` and never assigned — any distance query on an empty space raised
`AttributeError`. This was caught while writing the test for the above.

## Fix

- Guard in `get_k_nearest_agents`: if `k >= n`, return all agents directly.
  If the space is empty, return `([], np.array([]))`.
- Initialise `agent_positions` to an empty slice in `__init__` so distance
  methods work on a freshly-created space with no agents.

## Test

Added `test_get_k_nearest_agents_k_exceeds_count` covering:
- `k` equal to agent count
- `k` greater than agent count  
- empty space

### DIFF
--- a/mesa/experimental/continuous_space/continuous_space.py
+++ b/mesa/experimental/continuous_space/continuous_space.py
@@ -88,7 +88,9 @@ class ContinuousSpace:
         self._agent_positions: np.array = np.empty(
             (n_agents, self.dimensions.shape[0]), dtype=float
         )
-        self.agent_positions: np.array = self._agent_positions[0:0]  # empty view, updated in _add_agent
+        self.agent_positions: np.array = self._agent_positions[
+            0:0
+        ]  # empty view, updated in _add_agent
 
         # the list of agents in the space
         self.active_agents = []
--- a/mesa/experimental/continuous_space/continuous_space.py
+++ b/mesa/experimental/continuous_space/continuous_space.py
@@ -88,9 +88,7 @@ class ContinuousSpace:
         self._agent_positions: np.array = np.empty(
             (n_agents, self.dimensions.shape[0]), dtype=float
         )
-        self.agent_positions: (
-            np.array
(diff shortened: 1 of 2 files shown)


quaquel:
I agree that this needs fixing, but I am not convinced by the proposed fix. Is it not more sensible to address this at the root in `calculate_distances`?

mesa/experimental/continuous_space/continuous_space.py
```diff
@@ -253,13 +253,20 @@ def get_k_nearest_agents(
...
             earlier an agent is inserted the higher it will rank.
+            If k exceeds the number of agents, all agents are returned.
 
         """
         dists, agents = self.calculate_distances(point)
+        n = len(dists)
+
+        if n == 0:
+            return [], np.array([])
+
```
quaquel, line 264: this is something that in my view should be handled inside `self.calculate_distances`

mesa/experimental/continuous_space/continuous_space.py
```diff
@@ -88,9 +88,9 @@ def __init__(
         self._agent_positions: np.array = np.empty(
             (n_agents, self.dimensions.shape[0]), dtype=float
         )
-        self.agent_positions: (
-            np.array
-        )  # a view on _agent_positions containing all active positions
+        self.agent_positions: np.array = self._agent_positions[
+            0:0
+        ]  # empty view, updated in _add_agent
```
quaquel, line 93: why?

quaquel:
This PR also overlaps with #3671. Earlier PRs generally take precedence.

PR #3722, opened 2026-05-22
fix: remove trailing comma in switcher.json

### LINKED ISSUE #3720: Docs Version selector overlaps / renders incorrectly
During inspection and reading of the docs of mesa i found that in this [mesadocs](https://mesa.readthedocs.io/latest/) when choosing the version something incosistent happening like a box coming but options are not appearing so i would like to adress this issue also wanted to fix this issue here is the attachemnt.

Image

i have a idea of giving here a dropdown so that user could choose various version also include a option below it which will show older versions of mesa its a added functionality. Love to have if you have any query and any suggestion if you have.

dietcoke-17:
Fixes #3720 
This PR fixes the version selector dropdown that was showing a blank box instead of displaying available versions.
The **switcher.json** file had a comma after the last JSON object, which caused invalid JSON and prevented the version dropdown from rendering correctly. 

Key Fix: Removed the comma from  @**switcher.json**

### DIFF
--- a/docs/_static/switcher.json
+++ b/docs/_static/switcher.json
@@ -18,6 +18,6 @@
     "name": "3.5.1",
     "version": "v3.5.1",
     "url": "https://mesa.readthedocs.io/en/v3.5.1/"
-  },
+  }
 
 ]
--- a/docs/_static/switcher.json
+++ b/docs/_static/switcher.json
@@ -12,7 +12,7 @@
   {
     "name": "2.4.0",
     "version": "v2.4.0",
-    "url": "https://mesa.readthedocs.io/en/v2.4.0/"
+    "url": "https://mesa.readthedocs.io/v2.4.0/"
   },
   {
     "name": "3.5.1",
--- a/docs/_static/switcher.json
+++ b/docs/_static/switcher.json
@@ -17,7 +17,7 @@
   {
     "name": "3.5.1",
     "version": "v3.5.1",
-    "url": "https://mesa.readthedocs.io/en/v3.5.1/"
+    "url": "https://mesa.readthedocs.io/v3.5.1/"
   }
 
 ]


codebreaker32:
Could you elaborate over the bug and how is removing comma gonna solve this?

dietcoke-17:
Let me elaborate this 
The switcher.json file contains a trailing comma after the last JSON object, which violates the JSON specification. This makes the entire file invalid JSON. When the version dropdown JavaScript code tries to parse this file using JSON.parse():
The parser encounters invalid JSON and throws an error then no data is returned to the dropdown and the dropdown renders blank instead of showing available versions

This is backed by RFC 8259, the official JSON specification, which forbids a comma after the final element of an array or object. You can use this to cross check wheter the JSON is in correct format or not and it will show key error "https://www.jsonbeautify.org/en". 
And here are the references i used:
1. https://www.rfc-editor.org/rfc/rfc8259.html
2. https://www.ecma-international.org/publications-and-standards/standards/ecma-404/

codebreaker32:
So I tested this in #3723 and its working but I encountered 404 after clicking on versions other than stable and latest. Any ideas? @EwoutH @quaquel

quaquel:
I am not sure about this. We currently have a dropdown box in the top left, and a chooser in the bottom right. It makes little sense to support 2 selectors. The bottom right one is tied to releases (see #3565).

codebreaker32:
I went through #3565 and found the root cause for the 404s. The URLs for v3.5.1 and v2.4.0 in `switcher.json` incorrectly include the `/en/` prefix.

While stable and latest successfully redirect the `/en/` paths automatically, the version tags fail to do so.
Current: https://mesa.readthedocs.io/en/v2.4.0/
Correct: https://mesa.readthedocs.io/v2.4.0/

I figured this out after checking via the version chooser in the bottom-left.

docs/_static/switcher.json
codebreaker32:
```suggestion
    "url": "https://mesa.readthedocs.io/v2.4.0/"
```

docs/_static/switcher.json
```diff
@@ -18,6 +18,6 @@
     "name": "3.5.1",
     "version": "v3.5.1",
     "url": "https://mesa.readthedocs.io/en/v3.5.1/"
```
codebreaker32:
```suggestion
    "url": "https://mesa.readthedocs.io/v3.5.1/"
```

codebreaker32 (approved):
With these changes, it will be ready to merge!

tpike3:
Apologies I am a little late to this party. To make sure I am understanding this fixes the dropdown in the upper left if someone changes the document view (e.g. latest or stable in the bottom right), correct?  and thanks @codebreaker32 for figuring out the `/en/` issue and @dietcoke-17  for working on this issue.

My question would be why are we why are we showing 2.4? Shouldn't it be latest (current github version and stable v 3.5.1)? 

And to @quaquel do we need two? Can we just use bottom right or is this a feature of our read the docs layout?

quaquel:
> And to @quaquel do we need two? Can we just use bottom right or is this a feature of our read the docs layout?

My personal opinion is that we should keep only 1 and then I prefer the bottom right solution. Curious to hear what others think.

dietcoke-17:
@quaquel & @tpike3  i think we could also go with the two the bottom right is the stable one this is just a added feature on top of it as a reader it will be easy to navigate and catch the attention at the top and if anyone wants to get info about the older version and approach he also could get those using this one and currently many use normally a dropdown version to see older version so it would be helpful for the newbies to navigate.

tpike3:
I am not overly passionate, but I am good with one in the bottom right. 

I am passionate that we get rid of 2.4 though.

quaquel:
> I am passionate that we get rid of 2.4 though.

Why would we remove this documentation? Unfortunately, there are still people using this.

dietcoke-17:
ok we can go with the bottom one only and should not remove the old documentation if there are active users in it. should i close this pr and the issue?

tpike3:
Closing, supersed by #3738

PR #3725, opened 2026-05-30
Make CellCollection.select_random_cell empty-safe with a default parameter

AshNicolus:
## What

`CellCollection.select_random_cell()` calls `self.random.choice(self.cells)`
directly, which raises a bare `IndexError` with no context when the collection
is empty.

The sibling method `select_random_agent()` already handles this gracefully via a
`default` parameter (returning the default when empty, or raising `LookupError`
otherwise). This PR brings `select_random_cell()` in line with that existing API
so the two methods behave consistently.

## Change

- `select_random_cell(default=RAISES)`: returns `default` when the collection is
  empty, or raises `LookupError` with a clear message if no default is given.
  Non-empty behavior is unchanged.

This mirrors the existing `select_random_agent` signature and semantics exactly —
no new patterns introduced.

## Test

Added `test_select_random_cell_empty_safe`, matching the existing
`test_select_random_agent_empty_safe`:
- raises `LookupError` on an empty collection
- returns the provided `default` when empty
- still returns a valid cell for a non-empty collection

All 88 tests in `tests/discrete_space/test_discrete_space.py` pass; `ruff` clean.

### DIFF
--- a/mesa/discrete_space/cell_collection.py
+++ b/mesa/discrete_space/cell_collection.py
@@ -98,8 +98,24 @@ class CellCollection[T: Cell]:
     def agents(self) -> Iterable[CellAgent]:  # noqa
         return itertools.chain.from_iterable(self._cells.values())
 
-    def select_random_cell(self) -> T:
-        """Select a random cell."""
+    def select_random_cell(self, default=RAISES) -> T | None:
+        """Select a random cell from the collection.
+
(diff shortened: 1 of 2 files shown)


quaquel (approved):

quaquel:
This makes sense to me and looks good. Thanks for the PR.

PR #3726, opened 2026-05-30
Fix AttributeError when accessing Cell.empty before adding/removing agents

AshNicolus:
## What

Accessing `cell.empty` on a freshly created `Cell` raises:

    AttributeError: 'Cell' object has no attribute '_empty'

`_empty` is declared in `Cell.__slots__` and the `empty` property getter returns
`self._empty`, but `Cell.__init__` never initializes it. The attribute only comes
into existence after the first `add_agent`/`remove_agent` call (which assign
`self.empty`).

This affects every cell that is **not** a grid cell — i.e. cells in `Network`
and `VoronoiGrid`. `GridCell` is unaffected because the grid attaches an `"empty"`
property-layer accessor that overrides the attribute.

## Reproduction

```python
import random

c = Cell((0,), random=random.Random())
c.empty          # AttributeError: 'Cell' object has no attribute '_empty'

### DIFF
--- a/mesa/discrete_space/cell.py
+++ b/mesa/discrete_space/cell.py
@@ -104,6 +104,7 @@ class Cell:
         self._agents: list[
             CellAgent
         ] = []  # TODO:: change to AgentSet or weakrefs? (neither is very performant, )
+        self._empty: bool = True  # a freshly created cell holds no agents
         self.capacity: int | None = capacity
         self.properties: dict[
             Coordinate, object
--- a/tests/discrete_space/test_discrete_space.py
+++ b/tests/discrete_space/test_discrete_space.py
@@ -668,6 +668,30 @@ def test_cell():
         cell_zero.add_agent(CellAgent(model))
 
 
+def test_cell_empty_attribute_initialized():
+    """A freshly created Cell exposes `empty` without first adding/removing an agent.
+
+    Cells that are not grid cells (e.g. those in Network and VoronoiGrid) have no
+    `empty` property layer overriding the attribute, so reading `cell.empty` must
+    work straight after construction. Regression test for the uninitialized
+    `_empty` slot which raised AttributeError.
+    """
+    model = Model()
+    cell = Cell((0,), capacity=None, random=random.Random())
+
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by quaquel** [comment]
>  The attribute only comes into existence after the first add_agent/remove_agent call (which assign self.empty).

As you indicate, the `empty` attribute is controlled via properties that are assigned in the grid class when the `empty` property layer is created. See [here](https://github.com/mesa/mesa/blob/30ae6c2a4da0f1640d284641c7322e849838983f/mesa/discrete_space/grid.py#L213). Moreover, this PR also is related to #3072. Basically, we have 2 mechanisms for tracking empty cells: the `_empties` attribute and the `Grid.empty` property layer. As part of the larger overhaul of spaces, we plan to make the empty property layer work across all discrete spaces, not just grids. 

Can you elaborate on the use case where you ran into the error that is fixed by this PR.

**Comment by AshNicolus** [comment]
Thanks, that context helps. I didn't know about the overhaul or #3072.

To be honest, I didn't hit this in an actual model. I found it while going through the discrete_space code looking at how the different spaces line up. The thing that caught my eye is that `Cell.empty` is a public property with a getter and setter, and add_agent/remove_agent keep it updated. On grid cells it works because of the property-layer accessor you linked, but on a plain Cell (so any cell in a Network or VoronoiGrid) reading `.empty` before you've added or removed an agent throws AttributeError, since `_empty` is in `__slots__` but never set in `__init__`.

```python
import random, networkx as nx
from mesa.discrete_space import Network

net = Network(nx.path_graph(3), random=random.Random(0))
net._cells[0].empty   # AttributeError: 'Cell' object has no attribute '_empty'
```

So this is really just stopping an existing public property from crashing on non-grid spaces. It doesn't get into the `_empties` vs property-layer side of things.

Since you're planning to unify the empty property layer across all spaces anyway, I'm fine either way. If the overhaul is going to replace this, feel free to close it. If a small fix is useful in the meantime, this is about as minimal as it gets. Up to you.

quaquel (approved):
Let's merge this now so at least the bug is gone.

PR #3727, opened 2026-05-31
Fix TypeError in meta-agent sorting with non-integer unique_id

### LINKED ISSUE #3563: TypeError in remove_constituting_agents sorting with non-integer unique_id

AshNicolus:
Closes #3563

## What

`meta_agent.py` keeps each agent's backward-compatibility `meta_agent` attribute
up to date by sorting the agent's meta-agents by `unique_id`. Several sites used:

```python
sorted(meta_agents, key=lambda x: x.unique_id or 0)
```

The `or 0` injects an `int` into the sort key whenever a `unique_id` is `None` or
falsy. As soon as another `unique_id` is non-integer (a model using string or UUID
ids), the sort compares e.g. `0 < "alpha"` and raises:

    TypeError: '<' not supported between instances of 'int' and 'str'

## Reproduction

```python
from mesa import Model
from mesa.agent import Agent
from mesa.experimental.meta_agents.meta_agent import MetaAgent

m = Model(rng=2)
a = Agent(m)
ma1 = MetaAgent(m, {a}, name="MA1")
ma2 = MetaAgent(m, {a}, name="MA2")
ma1.unique_id = "alpha"
ma2.unique_id = None       # `None or 0` -> 0, then compared against "alpha"
ma1.add_constituting_agents({a})   # TypeError
```

## Fix

Replaced the `unique_id or 0` lambdas (and one bare `unique_id` lambda) with a
shared helper:

```python
def _unique_id_sort_key(agent):
    return (agent.unique_id is not None, agent.unique_id)
```

This is type-stable and `None`-safe: a `None` id sorts first (treated as lowest,
matching the previous intent) without injecting an incompatible `int`. Integer
ordering is unchanged, and homogeneous string/UUID ids now sort correctly.

Note: genuinely mixed incomparable id types (e.g. a `str` and a `UUID` in the same
set) remain unsortable by design — "lowest" is undefined there, and forcing it via
`str()` would corrupt integer ordering.

## Tests

Added regression tests for string ids, UUID ids, and a `None` id. Full experimental
suite passes under `-Werror`; ruff clean.

### DIFF
--- a/tests/experimental/test_meta_agents.py
+++ b/tests/experimental/test_meta_agents.py
@@ -1,5 +1,7 @@
 """Tests for the meta_agents module."""
 
+import uuid
+
 import pytest
 
 from mesa import Agent, Model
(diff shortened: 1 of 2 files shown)


tests/experimental/test_meta_agents.py
```diff
@@ -1,5 +1,7 @@
 """Tests for the meta_agents module."""
 
+import uuid
```
tpike3: Can you remove this and update the associated tests, please? It is not really needed, based on the associated tests.

tpike3:
Thanks @AshNicolus! I made one comment, although meta-agents is going through a significant rewrite, it is still good to fix the bugs and ensure we do not include them in the update.

tests/experimental/test_meta_agents.py
AshNicolus: Done. I removed the import uuid and the UUID test. The string and None tests already cover the non-integer case so coverage is the same. Thanks for the review, and good to know about the rewrite.

tpike3 (approved):
Thanks @AshNicolus! LGTM

PR #3731, opened 2026-06-04
Preserve read_only property layers across Grid pickling and deepcopy

### LINKED ISSUE #3730: Grid property layers lose read_only after pickling
**Describe the bug**

Read-only property layers on a `Grid` become writable again after the grid is pickled and unpickled. The `read_only=True` setting is not preserved across the round trip, so write protection that worked before pickling is silently lost.

The cause is in `mesa/discrete_space/grid.py`. `_attach_property_layer` creates a getter-only `property` when `read_only=True`, so writes raise `AttributeError`. But `Grid.__setstate__` rebuilds every property layer with both a getter and a setter, ignoring the original `read_only` flag, so the restriction disappears after unpickling.

**Expected behavior**

After unpickling, writing to a read-only property layer should still raise `AttributeError`, exactly as it does before pickling. The `read_only` flag should survive the pickle round trip.

**To Reproduce**

```python
import pickle
from mesa import Model

grid = OrthogonalMooreGrid(dimensions=(3, 3), random=Model().random)
grid.create_property_layer("protected", default_value=0.0, read_only=True)

# Before pickle: read-only is enforced
cell = grid._cells[(0, 0)]
try:
    cell.protected = 99.0
    print("before pickle: read-only NOT enforced")
except AttributeError:
    print("before pickle: read-only enforced (correct)")

# After pickle: read-only is gone
grid2 = pickle.loads(pickle.dumps(grid))
cell2 = grid2._cells[(0, 0)]
cell2.protected = 99.0          # expected to raise AttributeError, but it succeeds
print("after pickle: value =", cell2.protected, "(read-only lost)")
```

Output:

```text
before pickle: read-only enforced (correct)
after pickle: value = 99.0 (read-only lost)
```

**Additional context**

Mesa 4.0.0a0 (current `main`), Python 3.12. Platform-independent, since this is a pure logic bug.

`Grid.__setstate__` does not know which layers were read-only because that information is not carried in the pickled state. A fix would record the set of read-only layer names in `__getstate__` (or recompute it) and have `__setstate__` rebuild each accessor with a getter only when the layer was read-only, mirroring what `_attach_property_layer` already does.

AshNicolus:
Closes #3730

## What

Read-only property layers on a `Grid` became writable again after the grid was
pickled (or deepcopied). `Grid.__setstate__` rebuilt every property layer with both
a getter and a setter, ignoring the original `read_only` flag, so the restriction
that `_attach_property_layer` set up was silently lost on the round trip.

## Reproduction

```python
import pickle
from mesa import Model

grid = OrthogonalMooreGrid(dimensions=(3, 3), random=Model().random)
grid.create_property_layer("protected", default_value=0.0, read_only=True)

grid2 = pickle.loads(pickle.dumps(grid))
grid2._cells[(0, 0)].protected = 99.0   # expected AttributeError, but it succeeded
```

## Fix

The read-only status is now carried through the pickle round trip:

- `__getstate__` records the set of read-only layers (those whose cell-class
  accessor has `fset is None`).
- `__setstate__` rebuilds each accessor as getter-only when the layer was read-only,
  and getter+setter otherwise, mirroring `_attach_property_layer`.

The change is confined to the two pickle methods. `copy.deepcopy` uses the same
methods, so it is fixed too.

## Test

Added `test_read_only_property_layer_survives_pickle_and_deepcopy`: after both
`deepcopy` and `pickle` round trips, a read-only layer still raises `AttributeError`
on write, while a read-write layer remains writable.

Full `tests/discrete_space/` suite passes under `-Werror`; ruff clean.

### DIFF
--- a/mesa/discrete_space/grid.py
+++ b/mesa/discrete_space/grid.py
@@ -396,23 +396,34 @@ class Grid(DiscreteSpace[T]):
         """Custom __getstate__ for handling dynamic GridCell class and property_layer accessors."""
         state = super().__getstate__()
         state = {k: v for k, v in state.items() if k != "cell_klass"}
+        # Record which property_layers are read-only (their accessor has no setter)
+        # so the read-only constraint survives the pickle round trip.
+        state["_read_only_property_layers"] = {
+            name
+            for name in self.property_layers
+            if getattr(self.cell_klass, name).fset is None
+        }
         return state
 
     def __setstate__(self, state: dict[str, Any]) -> None:
         """Restore state and re-attach property_layer accessors to the cell class."""
+        read_only_layers = state.pop("_read_only_property_layers", set())
         super().__setstate__(state)
         for name, array in self.property_layers.items():
-            setattr(
-                self.cell_klass,
-                name,
-                property(
+            if name in read_only_layers:
+                accessor = property(
+                    lambda self_cell, a=array: a[self_cell.coordinate],
+                    doc=f"property_layer '{name}'",
+                )
+            else:
+                accessor = property(
                     lambda self_cell, a=array: a[self_cell.coordinate],
                     lambda self_cell, v, a=array: a.__setitem__(
                         self_cell.coordinate, v
                     ),
                     doc=f"property_layer '{name}'",
-                ),
-            )
+                )
+            setattr(self.cell_klass, name, accessor)
 
 
 class OrthogonalMooreGrid(Grid[T]):
--- a/mesa/discrete_space/grid.py
+++ b/mesa/discrete_space/grid.py
@@ -110,6 +110,9 @@ class Grid(DiscreteSpace[T]):
             {"property_layers": set(), "__slots__": ()},
         )
         self.property_layers: dict[str, np.ndarray] = {}
+        # Track which property layers are read-only so the constraint survives
+        # pickling and deepcopy (see __getstate__/__setstate__).
+        self._read_only_layers: set[str] = set()
 
         # we register the pickle_gridcell helper function
         copyreg.pickle(self.cell_klass, pickle_gridcell)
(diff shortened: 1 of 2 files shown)


quaquel:
The fix seems fine. However, I am wondering whether it would not be a good idea to store on the space which layers are read only and which ones are not.

codebreaker32:
Thanks @AshNicolus for the PR and bug report. I agree with @quaquel that making the state explicit feels cleaner than inferring it via `.fset` reflection during `__getstate__`. We can just initialize a `self._read_only_layers = set()` in `Grid.__init__`, add the layer name to it when `read_only=True` is passed, and then read from that set in `__setstate__`

AshNicolus:
Good call, done. Switched to an explicit `self._read_only_layers` set: initialized in `Grid.__init__`, the layer name is added in `_attach_property_layer` when `read_only=True`, discarded in `remove_property_layer`, and read in `__setstate__`. `__getstate__` is back to its original form since the set is part of `__dict__` and pickles along with everything else. No more `.fset` reflection.

codebreaker32 (approved):
LGTM!

dietcoke-17:
hey @AshNicolus 1 minor Q: do we need any additional safeguards in __setstate__ if _read_only_layers contains a layer name that is no longer present in property_layers, or is that state impossible given the current lifecycle of property layers?

AshNicolus:
@dietcoke-17 Good question. No extra safeguard is needed, for two reasons:

1. The lifecycle keeps `_read_only_layers` a subset of `property_layers`. `remove_property_layer` discards the name from `_read_only_layers` as well, so the two never drift apart.

2. Even if a stale name were somehow present, `__setstate__` is driven by `property_layers`, not by `_read_only_layers`. The loop iterates `self.property_layers.items()` and only does a membership check (`if name in read_only_layers`), so a name in `_read_only_layers` that isn't in `property_layers` is simply never looked at. No `KeyError`, no effect.

I confirmed this: adding a bogus name to `_read_only_layers` and then pickling round-trips fine, the bogus name does not appear in `property_layers`, and read-write layers stay writable.

dietcoke-17 (commented):
cool lgtm me too

quaquel (approved):

PR #3734, opened 2026-06-08
fix: return descriptor on class-level Observable access (#3732)

### LINKED ISSUE #3732: Observable raises AttributeError on class-level access (descriptor protocol)
**Describe the bug**

Accessing an `Observable` (or any `BaseObservable`) attribute on the class instead of an instance raises `AttributeError` instead of returning the descriptor.

`BaseObservable.__get__` does `getattr(instance, self.private_name)` unconditionally. When the attribute is accessed on the class (for example `MyAgent.value`), Python calls `__get__` with `instance=None`, so this becomes `getattr(None, "_value")` and raises `AttributeError: 'NoneType' object has no attribute '_value'`.

Standard descriptors (such as the built-in `property`) return the descriptor object itself when `instance is None`. Because `Observable` does not, normal class-level introspection silently breaks too: `hasattr(MyAgent, "value")` returns `False`, even though every instance has the attribute.

**Expected behavior**

Class-level access should return the `Observable` descriptor (the same convention `property` follows), and `hasattr(cls, name)` should be `True`.

**To Reproduce**

```python
from mesa import Agent, Model
from mesa.experimental.mesa_signals import Observable, HasEmitters


class MyAgent(Agent, HasEmitters):
    value = Observable()

    def __init__(self, model):
        super().__init__(model)
        self.value = 10


m = Model(rng=42)
a = MyAgent(m)
print(a.value)                      # 10  (instance access works)

MyAgent.value                       # AttributeError: 'NoneType' object has no attribute '_value'
print(hasattr(MyAgent, "value"))    # False  (should be True)
```

Output:

```text
10
AttributeError: 'NoneType' object has no attribute '_value'
```

(`hasattr` swallows the `AttributeError`, so it returns `False` rather than raising.)

**Additional context**

Mesa 4.0.0a0 (current `main`), Python 3.12. Platform-independent.

The fix is the standard descriptor-protocol guard: at the top of `BaseObservable.__get__`, return `self` when `instance is None`. `Observable` inherits this `__get__`, so the single guard covers both classes.

gaoflow:
Fixes #3732.

`BaseObservable.__get__` calls `getattr(instance, self.private_name)` unconditionally. On class-level access (e.g. `MyAgent.value`) Python calls `__get__` with `instance=None`, so this becomes `getattr(None, "_value")` and raises `AttributeError` instead of returning the descriptor. As a side effect `hasattr(MyAgent, "value")` silently returns `False`, even though every instance has the attribute.

Standard descriptors (including the built-in `property`) return the descriptor itself when `instance is None`. This adds that guard at the top of `BaseObservable.__get__`:

```python
class MyAgent(Agent, HasEmitters):
    value = Observable()

MyAgent.value               # before: AttributeError -> after: the Observable descriptor
hasattr(MyAgent, "value")   # before: False -> after: True
```

`Observable` and `ObservableList` both inherit this `__get__` (neither overrides it), so the single base-class guard fixes class-level access for all observable descriptors at once.

Added `test_observable_class_level_access`, which checks class-level access and `hasattr` for both `Observable` and `ObservableList`. It fails on `main` (the `AttributeError`) and passes here, and the rest of `test_mesa_signals.py` stays green.

Thanks to @AshNicolus for the clear report and diagnosis.

### DIFF
--- a/mesa/experimental/mesa_signals/core.py
+++ b/mesa/experimental/mesa_signals/core.py
@@ -64,6 +64,10 @@ class BaseObservable:
         self.fallback_value = fallback_value
 
     def __get__(self, instance: HasEmitters, owner):  # noqa: D105
+        if instance is None:
+            # class-level access (``Cls.attr``) returns the descriptor itself,
+            # matching ``property`` and the standard descriptor protocol
+            return self
         value = getattr(instance, self.private_name)
 
         if CURRENT_COMPUTED is not None:
--- a/tests/experimental/test_mesa_signals.py
+++ b/tests/experimental/test_mesa_signals.py
@@ -42,6 +42,37 @@ def test_observables():
     handler.assert_called_once()
 
 
+def test_observable_class_level_access():
+    """Observables follow the descriptor protocol on class-level access.
+
+    Accessing an Observable/ObservableList on the class rather than an
+    instance should return the descriptor itself, like ``property``, instead
+    of raising ``AttributeError`` (gh #3732).
+    """
+
+    class MyAgent(Agent, HasEmitters):
+        value = Observable()
+        items = ObservableList()
+
+        def __init__(self, model):
+            super().__init__(model)
+            self.value = 10
+
+    model = Model(rng=42)
+    agent = MyAgent(model)
+
+    # instance access is unchanged
+    assert agent.value == 10
+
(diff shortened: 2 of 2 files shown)


quaquel:
identical to #3733

PR #3738, opened 2026-06-09
docs: show only stable version in switcher

dietcoke-17:
This PR is for the docs to point only to the stable version present in mesa currently that is this one https://mesa.readthedocs.io/stable/ only and all other versions are removed as it the docs will guide to the only to the stable version as recommended by @tpike3 so i have removed all other version from docs/_static/switcher.json and used only the stable version. I think this solves the problem

### DIFF
--- a/docs/_static/switcher.json
+++ b/docs/_static/switcher.json
@@ -1,7 +1,12 @@
 [
     {
-    "name": "stable",
-    "version": "stable",
-    "url": "https://mesa.readthedocs.io/en/stable/"
-  }
+        "name": "latest",
+        "version": "latest",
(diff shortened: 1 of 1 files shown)


tpike3:
@dietcoke-17  Appreciating this contradicts what I said yesterday, can you instead do latest and v3.5.1. Looking at the read the docs and really thinkin through the user experience, this would make the docs the explicit last release (v.3.5.1) so there is less version confusion and the current repo. I will merge soonest.

dietcoke-17:
@tpike3 i have updated as you said

tpike3 (approved):

tpike3:
Thanks @dietcoke-17

PR #3741, opened 2026-06-10
Add informative __repr__ and __str__ to AgentSet and GroupBy

### LINKED ISSUE #3735: Add informative __repr__ to AgentSet and GroupBy

The core collection objects have no informative `__repr__`, so in a notebook or debugger they show only the default object address:

```python
>>> repr(model.agents)
'<mesa.agentset._HardKeyAgentSet object at 0x...>'
>>> repr(model.agents.select(lambda a: a.wealth > 0))
'<mesa.agentset.AgentSet object at 0x...>'
>>> repr(model.agents.groupby("type"))
'<mesa.agentset.GroupBy object at 0x...>'
```

This is also inconsistent with the rest of the API: `Cell` already defines a useful `__repr__` (`Cell((0, 0), [...])`), but `AgentSet`, `_HardKeyAgentSet`, and `GroupBy` do not. Inspecting agents interactively is a very common operation, so the missing repr is friction in exactly the place it is felt most.

Add a concise, informative `__repr__` to the AgentSet family and to `GroupBy`, for example:

```python
>>> model.agents
AgentSet (200 agents)
>>> model.agents.groupby("type")
GroupBy(3 groups: {'wolf': 50, 'sheep': 150, 'grass': 400})
```

Implementing `__repr__` once on `AbstractAgentSet` covers both `AgentSet` and `_HardKeyAgentSet`. `GroupBy` gets its own showing the number of groups and their sizes. The exact wording is open to discussion (for example, whether the `AgentSet` repr should include a per-type breakdown or just the count).

Yes, happy to open a PR once the approach and the repr format are approved.

**Additional context**

Self-contained and backward compatible: it only changes `repr` output, no behavioral change. Mesa 4.0.0a0 / current `main`. `Model.__repr__` is similarly missing and could be a follow-up, but this issue is scoped to the collection objects to keep it focused.

AshNicolus:
Closes #3735

## What

The core collection objects had no informative `__repr__`/`__str__`, so in a REPL,
notebook, or debugger they showed only the default `<... object at 0x...>`. `Cell`
already defines a useful `__repr__`, so this also removes an API inconsistency.

## Change

Per the discussion in #3735:

- `AbstractAgentSet.__repr__` → `<AgentSet (200 agents)>` (uses the real class name,
  so `model.agents` shows `<_HardKeyAgentSet (200 agents)>`).
- `AbstractAgentSet.__str__` → `AgentSet with 200 agents`.
- `GroupBy.__repr__` → `GroupBy(3 groups)`.

All are count-only and O(1) (`len()` is constant-time for both AgentSet types and
GroupBy), so they add no overhead. The per-type breakdown was dropped to avoid
iterating all agents on every call, and `_repr_html_` is left for a future issue,
both as agreed on the issue.

## Test

Added `test_agentset_repr_and_str` and `test_groupby_repr`. Full non-viz suite
(508 tests) passes under `-Werror`; ruff clean.

### DIFF
--- a/mesa/agentset.py
+++ b/mesa/agentset.py
@@ -958,5 +958,6 @@ class GroupBy:
         return len(self.groups)
 
     def __repr__(self) -> str:
-        """Return a concise representation showing the number of groups."""
-        return f"{type(self).__name__}({len(self)} groups)"
+        """Return a representation showing each group's identifier and size."""
+        sizes = {name: len(group) for name, group in self.groups.items()}
+        return f"{type(self).__name__}({len(self)} groups: {sizes})"
--- a/mesa/agentset.py
+++ b/mesa/agentset.py
@@ -55,6 +55,14 @@ class AbstractAgentSet[A: Agent](ABC, MutableSet[A]):
         """Update the AbstractAgentSet A with new set of agents."""
         ...
 
+    def __repr__(self) -> str:
+        """Return a concise, programmer-facing representation of the AbstractAgentSet."""
+        return f"<{type(self).__name__} ({len(self)} agents)>"
+
+    def __str__(self) -> str:
+        """Return a user-facing description of the AbstractAgentSet."""
+        return f"{type(self).__name__} with {len(self)} agents"
+
     def select(
         self,
         filter_func: Callable[[A], bool] | None = None,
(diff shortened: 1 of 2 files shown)


quaquel:
In groupby, you went with just the number of groups but not the composition of each group (or at least the identifiers for the groups). Why and would it not make sense to at least include the group identifiers?

AshNicolus:
Good point. I kept it minimal at first because of the cost concern from #3735, but that was about the AgentSet per-type breakdown, which needs a full pass over the agents. It doesn't apply to GroupBy since the groups are already split, so len(group) per group is cheap.

Updated the repr to include the identifiers and sizes:

GroupBy(2 groups: {'Wolf': 2, 'Sheep': 1})

Test updated too.

quaquel (approved):

PR #3743, opened 2026-06-11
fixing the version switcher in the docs

dietcoke-17:
this actually fix the version switcher in the read the docs , confy.py in docs folder actually generates the docs and in the switcher json link it previously has /en which was creating the problem so its a fix to that problem and followed up by #3738

### DIFF
--- a/.github/workflows/update-switcher.yml
+++ b/.github/workflows/update-switcher.yml
@@ -1,0 +1,42 @@
+name: Update docs version switcher
+
+on:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  update-switcher:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          ref: main
+      - name: Update switcher.json
+        shell: python
+        env:
+          NEW_TAG: ${{ github.ref_name }}
+        run: |
+          import json, os
+          from pathlib import Path
+          tag = os.environ["NEW_TAG"]
+          path = Path("docs/_static/switcher.json")
+          entries = json.loads(path.read_text())
+          if not any(e.get("version") == tag for e in entries):
+              entries.append({"name": tag, "version": tag, "url": f"https://mesa.readthedocs.io/{tag}/"})
+          path.write_text(json.dumps(entries, indent=2) + "\n")
+      - name: Open PR
+        uses: peter-evans/create-pull-request@v8
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: "docs: add ${{ github.ref_name }} to version switcher"
+          branch: "docs/switcher-${{ github.ref_name }}"
+          delete-branch: true
+          title: "docs: add ${{ github.ref_name }} to version switcher"
+          body: "Adds `${{ github.ref_name }}` to `docs/_static/switcher.json`."
+          base: main
+          labels: documentation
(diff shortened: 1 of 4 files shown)


dietcoke-17:
{178D1365-CCD8-4570-A9C3-E9FD046C70A1}
i have checked the pydata sphinx theme they have a similar dropdown if we want keep this or totally can drop this idea any of the things works which direction should we go once you confirm i can trim it down there

tpike3:
@dietcoke-17  Great thanks for tracking this down. Can you fix the conflicts please?  

To be clear my understanding is there is an organic read the docs that stores the previous versions so users can look through the history and this will now allow that.  correct?

dietcoke-17:
Yes I can fix the conflicts actually but regarding the history i will let you know how we can do that is there any possible way

dietcoke-17:
Hey @tpike3 i got it this pr actually solves that problem 
{155EAF1E-0E49-4804-AF0B-4A4F9B9CDD2C}
as we can see in the image and here is the full picture whats happening the below read the docs is by the addon documentation hosted by read the docs which serve similar type and the upper one is a more cleaner format both of them stores the version that we want to show and **yes your thoughts are perfectly align to this that "there is an organic read the docs that stores the previous versions so users can look through the history and this will now allow that."** so its just the fix that the version switcher was broken it basically fix that

tpike3:
@dietcoke-17 I think this needs more a permanent fix that is part of the our github actions. This right now seems like a very brittle fix that would have to be updated each time we do a release. In addition, right now if you go to the drop down in the upper left on the generated docs form this PR https://mesa--3743.org.readthedocs.build/3743/  the options then disappear.

I would go to just one version selector upper left or lower right, and there must be a way to make a new release a part of the readthedocs so then the versions are available to users. 

@EwoutH put together a very nice step by step guide on how we do a release. https://github.com/mesa/mesa/blob/main/CONTRIBUTING.md#releases. 

Would you be able to update this PR so it updates the workflow and read the docs so as we do a release it generates  the the drop down as your pydata theme picture above?

EwoutH (approved):
Diff looks good, thanks.

Will let merge to another maintainer.

dietcoke-17:
@tpike3 sorry i didnt have that bigger picture & surely working on this and will make sure to fix this problem in this PR

dietcoke-17:
@tpike3 i have updated as you have diretected and updated the workflow so whenever their is a new update the workflow will make sure with new release it generates the the drop down as well as update with new version

tpike3:
@dietcoke-17 Thanks. Looking at the read the docs from the PR, https://mesa--3743.org.readthedocs.build/3743/, it still does not seem to be having the desired impact. the switcher in the upper left is only in latest. While stable and 3.5.1 does not have it. 

I do not want to send you chasing for improvements beyond your original PR. So,  if you want you can go back to the original simpler PR, with the goal that their is only the switcher in the bottom right. Or, you can continue to pursue this more automated version. Entirely up to you. Either way we appreciate the work you are putting into this!

dietcoke-17:
@tpike3 as currently in my local in both latest and v3.5.1 both were properly working and the dropdown working smoothly and as you hav directed in the pr #3738 directed only two version due to a mistake in the confy.py file the stable version remained unchanged and also i have mentioned it eariler now both latest and v3.5.1 works smoothly we can see in the readthe docs build you might contradict with the fact that whenever you click on the v3.5.1 version it actually redirect you to the version which is currently present without this merged pr after this pr the fix will be automatically can be seen i think this fixes the issue i have checked every possible angle why it wont fail.

tpike3:
Thanks @dietcoke-17

PR #3744, opened 2026-06-12
Remove unused assignments in test_copying_discrete_spaces

### LINKED ISSUE #3737: Variable shadowing in test_copying_discrete_spaces makes pre-loop assertions dead code
**Description**
In tests/discrete_space/test_discrete_space.py, the test_copying_discrete_spaces function assigns c1 and c2 before a loop, but the loop immediately overwrites them with the same variable names:

```python
c1 = grid[(5, 5)].connections
c2 = grid_copy[(5, 5)].connections

for c1, c2 in zip(grid.all_cells, grid_copy.all_cells):
    ...
```
The two lines before the loop are dead code - they compute values that are never used. The same pattern appears twice more in the function for HexGrid.

**Expected Behaviour**
Either the pre-loop assignments are intentional and the loop variables should be renamed (e.g. cell1, cell2), or the pre-loop lines should be removed.

**To Reproduce**
Read the function, no test failure occurs because the dead code doesn't cause an error, it just silently does nothing.

Devansh-567:
## Description

Fixes #3737

The variables `c1` and `c2` in `test_copying_discrete_spaces` were assigned before the loop but immediately shadowed by the loop variables:

```python
c1 = grid[(5, 5)].connections
c2 = grid_copy[(5, 5)].connections

for c1, c2 in zip(grid.all_cells, grid_copy.all_cells):
```

As noted in the issue discussion, these assignments were never used because the loop iterates over all cells, including `(5, 5)`.

This PR removes the unused assignments from the `OrthogonalMooreGrid` and `HexGrid` test blocks, eliminating dead code without changing test behavior.

Kindly have a look at this, and if anything has to be changed do let me know, after the merging I will close #3737

### DIFF
--- a/tests/discrete_space/test_discrete_space.py
+++ b/tests/discrete_space/test_discrete_space.py
@@ -1103,9 +1103,6 @@ def test_copying_discrete_spaces():  # noqa: D103
         grid = OrthogonalMooreGrid((100, 100), random=random.Random(42))
         grid_copy = copy.deepcopy(grid)
 
-        c1 = grid[(5, 5)].connections
-        c2 = grid_copy[(5, 5)].connections
-
         for c1, c2 in zip(grid.all_cells, grid_copy.all_cells):
             for k, v in c1.connections.items():
                 assert v.coordinate == c2.connections[k].coordinate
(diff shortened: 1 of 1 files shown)


codebreaker32 (approved):
LGTM! Thanks @Devansh-567

Devansh-567:
The failing check (build / build (ubuntu, 3.14, --pre)) is unrelated to this PR. It's caused by nest_asyncio not being available in the solara/ipyvuetify dependency chain on Python 3.14 pre-release. This is an environment/dependency issue that exists on main independently of this change.

dietcoke-17 (approved):

quaquel (approved):

PR #3747, opened 2026-06-15
fix(visualization): emit deprecation warning only once per session

### LINKED ISSUE #3455: Agent portrayal deprecation warning floods console (emitted per agent per render)
# Agent portrayal deprecation warning floods console (emitted per agent per render)

### Bug Summary
When using the traditional dictionary-based return from `agent_portrayal` functions (as shown in most tutorials/documentation), the console is flooded with hundreds of identical `FutureWarning` messages, one for each agent on each render cycle, making it nearly impossible to see other debug output.

### Root Cause
In `mesa/visualization/space_renderer.py` and Altair backend components, the renderer emits a deprecation warning every time `agent_portrayal` returns a dictionary instead of an `AgentPortrayalStyle` instance:
```
FutureWarning: Returning a dict from agent_portrayal is deprecated and will be removed in Mesa 4.0. Please return an AgentPortrayalStyle instance instead.
```

Because this check happens inside the visualization rendering loop (executed for every agent on every frame), a simple model with 100 agents stepping 10 times generates **1,000+ identical warnings**, completely cluttering the terminal and obscuring other important output.

### Steps to Reproduce
```python
from mesa import Model, Agent

class MyAgent(Agent):

class MyModel(Model):
    def __init__(self):
        super().__init__()
        self.grid = MultiGrid(10, 10, True)
        for i in range(50):
            agent = MyAgent(i, self)
            self.grid.place_agent(agent, (i % 10, i // 10))

# Traditional dict-based portrayal (still documented in many tutorials)
def agent_portrayal(agent):
    return {
        "color": "blue",
        "size": 50,
    }

model = MyModel()
page = SolaraViz(
    model,
    components=[make_space_component(agent_portrayal)],
)
# Running this floods the console with hundreds of identical FutureWarning messages
```

### Expected Behavior
One of the following approaches:

1. **Silent auto-conversion**: Automatically wrap dictionaries in `AgentPortrayalStyle` internally without warnings (most user-friendly during transition period)
2. **Warn once per session**: Use `warnings.simplefilter('once')` or a flag to ensure the deprecation message only appears once
3. **Warn once per function**: Track which `agent_portrayal` functions have been warned about and only warn on first use

### Actual Behavior
The warning is emitted for **every agent on every render cycle**, resulting in:
- Terminal completely flooded with repetitive warnings
- Other important debug output or errors become invisible
- Degraded developer experience, especially for newcomers following older tutorials

Example output:
```
FutureWarning: Returning a dict from agent_portrayal is deprecated and will be removed in Mesa 4.0...
FutureWarning: Returning a dict from agent_portrayal is deprecated and will be removed in Mesa 4.0...
FutureWarning: Returning a dict from agent_portrayal is deprecated and will be removed in Mesa 4.0...
[... repeated hundreds of times ...]
```

### Environment
- Mesa version: 4.0.0.dev0
- Python version: 3.14.3

### Proposed Fix
I'm willing to submit a PR to fix this. The most user-friendly approach would be:

1. **Short-term** (for Mesa 3.x → 4.0 transition): Silently auto-wrap dictionaries in `AgentPortrayalStyle`, emitting the warning only once per session using:
```python
   warnings.warn(..., FutureWarning, stacklevel=2)
   warnings.simplefilter('once', FutureWarning)
```

2. **Long-term** (Mesa 4.0): Remove dict support entirely as planned

This balances backward compatibility with a clean migration path without degrading the developer experience.

---

**Note**: This is part 3 of 3 in a series of SolaraViz DX improvements:
- Part 2: #3454  - Class instantiation validation
- Part 3: Deprecation warning spam (this issue)

tianming-1996:
## Summary

- Wrap legacy dict-based `agent_portrayal` to emit `FutureWarning` only **once** instead of once per agent per render cycle
- Before: 100 agents × 10 frames = **1000+** identical warnings flooding console
- After: **1 warning** emitted, then silenced for remainder of session

## Changes

| File | Change |
|------|--------|
| `mesa/visualization/space_renderer.py` | Added `_wrap_legacy_portrayal()` wrapper with module-level flag; modified `setup_agents()` to use wrapper |

## Approach

The fix uses a simple module-level flag (`_warned_about_dict_portrayal`) to track whether the deprecation warning has been emitted. On the first dict return, the warning is shown. Subsequent calls return the dict silently without re-warning.

This is the most conservative approach - it preserves all existing behavior (dict still works, warning still appears) while eliminating the console flood that made debugging difficult.

## Test Plan

- [ ] Manual verification with a model rendering 100+ agents (confirm only 1 warning appears)
- [ ] Verify existing tests still pass

Fixes #3455

### DIFF
--- a/mesa/visualization/backends/altair_backend.py
+++ b/mesa/visualization/backends/altair_backend.py
@@ -14,6 +14,9 @@ from mesa.discrete_space import (
     OrthogonalVonNeumannGrid,
 )
 from mesa.visualization.backends.abstract_renderer import AbstractRenderer
+from mesa.visualization.components.portrayal_components import (
+    warn_if_legacy_agent_portrayal,
+)
 
 OrthogonalGrid = OrthogonalMooreGrid | OrthogonalVonNeumannGrid
 HexGrid = mesa.discrete_space.HexGrid
@@ -99,16 +102,7 @@ class AltairBackend(AbstractRenderer):
             aps: AgentPortrayalStyle
 
             if isinstance(portray_input, dict):
-                warnings.warn(
-                    (
-                        "Returning a dict from agent_portrayal is deprecated and will be removed in Mesa 4.0. "
-                        "Please return an AgentPortrayalStyle instance instead. "
-                        "For more information, refer to the migration guide: "
-                        "https://mesa.readthedocs.io/latest/migration_guide.html#defining-portrayal-components"
-                    ),
-                    FutureWarning,
-                    stacklevel=2,
-                )
+                warn_if_legacy_agent_portrayal()
                 dict_data = portray_input.copy()
                 agent_x, agent_y = self._get_agent_pos(agent, space)
 
--- a/mesa/visualization/backends/matplotlib_backend.py
+++ b/mesa/visualization/backends/matplotlib_backend.py
@@ -18,6 +18,9 @@ from mesa.discrete_space import (
     OrthogonalVonNeumannGrid,
 )
 from mesa.visualization.backends.abstract_renderer import AbstractRenderer
+from mesa.visualization.components.portrayal_components import (
+    warn_if_legacy_agent_portrayal,
+)
 
 OrthogonalGrid = OrthogonalMooreGrid | OrthogonalVonNeumannGrid
 HexGrid = mesa.discrete_space.HexGrid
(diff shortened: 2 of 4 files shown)


tianming-1996:
Hi @rht @EwoutH, this PR is ready for review. It fixes issue #3455 - the deprecation warning was being emitted once per agent per render cycle, flooding the console with thousands of identical warnings.

The fix uses a simple module-level state dict to track whether the warning has been shown, so it only emits once per session instead of once per agent.

All checks pass except  which is failing due to a CI environment issue (missing `nest_asyncio` module), unrelated to this code change.

Would appreciate a review when you have time!

quaquel:
I am sorry but I fail to understand this PR. Which specific warning are you trying to handle? The code only adds a new warning which is emitted once via the module level flag. I fail to see however, which other warning is being prevented by this.

PS. please refrain from pinging maintainers. We all get emails about PRs and get to it whenever we have time.

quaquel:
superseded by #3803

PR #3749, opened 2026-06-17
Fix broken file links in boid_flockers example.

Aanviksh281:
Fix broken file links in boid_flockers example documentation.

The file list in the boid_flockers example page was displaying model.py, agents.py and app.py as non functional relative links.

The PR replaces those with working GitHub URLs pointing to the correct source files in the repository.


dietcoke-17 (changes requested):
no not like this see how other examples files were written like it should not mandatorly point something it is a wrong type of written format

dietcoke-17:
you can check #3750 i have updated there

PR #3750, opened 2026-06-17
fixing the files section as other examples

dietcoke-17:
This is a documentation fix which basically fix the format as of other readme.md the file were showing pointing somwhere which should not be there


Aanviksh281:
Got it thanks.

tpike3 (approved):
LGTM

PR #3753, opened 2026-06-19
Add a RunConfiguration class for parameter sweeps

quaquel:
This PR adds a new `RunConfiguration` class. This class encapsulates the logic of instantiating a model, running it for a single scenario, and extracting user specifiable output from the run. This PR is part of the larger project of replacing the batch runner with a more flexible and powerful alternative. For detailed background, see #3483, and specifically https://github.com/mesa/mesa/discussions/3483#discussioncomment-17323416

The `RunConfiguration` class is designed to be subclassable, but it makes some default assumptions and provides keyword arguments to control its basic behavior to cover the most common use cases. 
* The class using `model.run_until` with a user specifiable end time
* It extracts all or the user specified output from the data_recorder. The attribute to which the recorder is assigned defaults to `data_recorder` but this is currently controllable via a keyword argument
* additional arguments and keyword arguments necessary to instantiate a model can also be passed to `RunConfiguration`. 

User subclassing can be done by overriding any of the three methods:
* `instantiate_model` - handles model instantiation
* `run_model` - handles running the model
* `extract_output` - handles the extraction and any post processing the user wants to do. 

It is likely that parts of this class will be refined in future PRs. Most notably, extract_output is currently a basic implementation which is likely to be expanded when working out the storage solution for parameter sweeps.

### DIFF
--- a/mesa/experimental/scenarios/runner.py
+++ b/mesa/experimental/scenarios/runner.py
@@ -1,0 +1,78 @@
+"""Classes for running parameter sweeps over scenarios."""
+from typing import Any
+
+import pandas as pd
+
+from mesa import Model
+from mesa.experimental.scenarios import Scenario
+
+
+class RunConfiguration:
+    """Defines how a single Scenario is executed and what is extracted from it.
+
+    Can be used as is for simple use cases or subclassed by overriding one or more of the following
+    methods
+
+    - ``instantiate_model`` — construct a Model from a Scenario (default:
+      ``model_class(*model_args, scenario=scenario, *model_kwargs)``).
+    - ``run_model`` — advance the model. Default delegates to ``model.run_until`` based on the ``until`` attribute.
+      Override for alternative run control
+    - ``extract_output`` — return a dict with outcome names as key and dataframes as values
+
+    Stopping is the model's responsibility. ``RunConfiguration`` only chooses
+    which run primitive to call.
+
+    """
+    def __init__(self, model_class: type[Model], until:float | int, model_args:None| list[Any]=None,
+                 model_kwargs:None|dict[str, Any]=None, outcomes: None|str|list[str]=None):
+        """Initialize a RunConfiguration object.
+
+        Args:
+            model_class: the model class to instantiate
+            until: until which time to run the model
+            model_args: any additional model arguments
+            model_kwargs: any additional model keyword arguments
+            outcomes: the outcomes to extract. If None, extract all outcomes.
+        """
+        super().__init__()
+
+        if not (isinstance(model_class, type) and issubclass(model_class, Model)):
+            raise TypeError("model_class must be a subclass of Model")
+        if not isinstance(until, (int, float)):
+            raise TypeError("until must be an int or float")
+        if until<=0:
+            raise ValueError("until must be positive")
+
+        self.model_class = model_class
+        self.model_args = [] if model_args is None else model_args
+        self.model_kwargs = {} if model_kwargs is None else model_kwargs
+        self.until = until
+
+        if isinstance(outcomes, str):
+            outcomes = [outcomes]
+        self.outcomes = outcomes
+
+    def instantiate_model(self, scenario) -> Model:
+        """Instantiate the model."""
+        return self.model_class(*self.model_args, scenario=scenario, **self.model_kwargs)
+
+    def run_model(self, model: Model):
+        """Run the model."""
+        model.run_until(self.until)
+
+    def extract_output(self, model: Model) -> dict[str, pd.DataFrame]:
+        """Extract output from model."""
+        # fixme:: this code assumed that the recorder is assigned to model.data_recorder
+        #   this is probably a a convention that needs to be pinned down explicitly on the model class
+        if self.outcomes is None:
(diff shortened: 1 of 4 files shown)


EwoutH:
Make sense I think. The default use-case (for a single sweep) would be creating a single RunConfiguration, and then varying multiple Scenarios within that right?

So if I'm doing two experiments (that each answer a different subquestion in my research), I probably have two RunConfigurations, with each a (larger) set of scenario's, that can be either created manually or using some sort of (external library for) sampling?

(some use cases with minimal API examples might be useful here)

quaquel:
> Make sense I think. The default use-case (for a single sweep) would be creating a single RunConfiguration, and then varying multiple Scenarios within that right?

Yes

> So if I'm doing two experiments (that each answer a different subquestion in my research), I probably have two RunConfigurations, with each a (larger) set of scenario's, that can be either created manually or using some sort of (external library for) sampling?

Whether you need separate RunConfigurations will depend on the specifics. It is only needed if how you run the model or how you extract data is different

> (some use cases with minimal API examples might be useful here)

I did not want to repeat everything here so see https://github.com/mesa/mesa/discussions/3483#discussioncomment-17323416

quaquel:
If it is helps, the overall API will look roughly like this


```python
# generate samples with scipy.qmc or salib

scenarios = MyScenario.from_dataframe(samples, rng=42, replications=10)
run_config = RunConfig(MyModel, until=150)

with ProcessPoolExecutor() as executor:
    results = run_scenarios(scenarios, run_config, executor=executor, store=InMemoryStore())

```

tpike3:
@quaquel This is really cool! I am just catching up and reading #3483 this draws a really nice boundary on what Mesa does and doesn't do. 

It will take me a day or 2, but I have some minor thoughts on the code (e.g., should we consider this edge case or not etc), but this is really great!

codebreaker32:
Looks great to me so far! Eagerly waiting for `Store` protocol 😀

quaquel:
> Eagerly waiting for Store protocol 

I hope to open a draft PR somewhere this week with `run_scenarios`, the `Store` protocol, and a working implementation of `InMemoryStore`.

mesa/experimental/scenarios/runner.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+class RunConfiguration:
+    """Defines how a single Scenario is executed and what is extracted from it.
+
+    Can be used as is for simple use cases or subclassed by overriding one or more of the following
+    methods
+
+    - ``instantiate_model`` — construct a Model from a Scenario (default:
+      ``model_class(*model_args, scenario=scenario, *model_kwargs)``).
```
tpike3: add missing * to model_kwargs so `**model_kwargs`

mesa/experimental/scenarios/runner.py
```diff
@@ -0,0 +1,99 @@
...
+
+        if isinstance(outcomes, str):
+            outcomes = [outcomes]
+        self.outcomes = outcomes
+
+    def instantiate_model(self, scenario: Scenario) -> Model:
+        """Instantiate the model."""
+        return self.model_class(
+            *self.model_args, scenario=scenario, **self.model_kwargs
+        )
```
tpike3: maybe because I crave tracebacks due to issues I have had, I would offer wrapping this function in a try/ except so users can see specifics if they are not passing in the parameters they think they are. As LLMs are very verbose, here is an overly verbose example

```
def instantiate_model(self, scenario: Scenario) -> Model:
    """Instantiate the model."""
    try:
        return self.model_class(
            *self.model_args, scenario=scenario, **self.model_kwargs
        )
    except TypeError as e:
        # Construct a highly descriptive error message
        class_name = getattr(self.model_class, '__name__', str(self.model_class))
        
        error_msg = (
            f"Failed to instantiate the model class '{class_name}'.\n"
            f"Please check your model_args and model_kwargs.\n"
            f" - Passed args: {self.model_args}\n"
            f" - Passed kwargs: {self.model_kwargs}\n"
            f" - Injected kwarg: {{'scenario': {scenario}}}\n"
            f"--> Original Python Error: {e}"
        )
        
        # Re-raise the exception with the new message, preserving the traceback
        raise TypeError(error_msg) from e
```

mesa/experimental/scenarios/runner.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def run_model(self, model: Model):
```
tpike3: Would add the type annotation 

` def run_model(self, model: Model) -> None:`

tpike3:
A few nitpicks and a suggestion, I hope they are useful

mesa/experimental/scenarios/runner.py
quaquel: I agree with this idea so I will address it. Note that all instantiation, running, and retrieving exceptions are also caught at the `run_scenarios` level. See #3756 for details.

quaquel:
> A few nitpicks and a suggestion, I hope they are useful

Thanks, those are useful and have all been addressed.

tpike3 (approved):
LGTM!

PR #3754, opened 2026-06-20
Experimental `ContinuousState` Descriptor with Tensor backend

codebreaker32:
This PR introduces an experimental Entity-Component-System (ECS) backend for modeling continuously evolving agent properties using a columnar NumPy architecture.

#### Architecture

* **`StateTensor`**: A columnar memory backend using parallel NumPy arrays to track values, rates, bounds, and limits for agent attributes. Updates and projection calculations are handled via vectorized operations or $O(1)$ scalar updates.
* **`ContinuousScheduler`**: A priority-based scheduler that maintains a single master event pointed at the earliest projected crossing time across the entire tensor, re-targeting dynamically as writes occur.
* **`ContinuousState`**: A descriptor that maps agent attributes directly to the tensor. It utilizes caching (`_cache_attr`, `_lazy_cache_attr`) on the agent instance to bypass Python's dynamic lookup overhead during reads and writes.

#### Thresholding

Thresholds can be attached to attributes in two modes (`eager` and `lazy`), with newly defined zero-crossing mechanics:

1. **Eager Mode (Predictive)**: Triggers are checked on `__set__`. The descriptor safely iterates over and synchronizes *all* tensor rows (`indices`) assigned to a multi-threshold attribute, re-projecting the earliest event to the `ContinuousScheduler`.
2. **Lazy Mode (Deferred & Cached)**: Triggers are evaluated exclusively on `__get__`. To prevent excessive evaluations, the descriptor utilizes a `last_read_attr` cache to track the value since the previous read. If a crossing occurs, execution is pushed to an auto-flushing deferred queue (`after=1e-9`) to execute safely at the end of the current step.

#### Example Model Updates

* **Sugarscape**: Refactored the example to utilize `ContinuousState` for sugar and spice metabolism.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -103,7 +103,7 @@ class Agent[M: Model]:
         # ensures models are also removed from datasets
         for dataset in self._datasets:
             self.model.data_registry[dataset].remove_agent(self)
-        
+
         indices = getattr(self, "_continuous_indices", {})
         for idx_list in indices.values():
             for idx in idx_list:
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -103,6 +103,12 @@ class Agent[M: Model]:
         # ensures models are also removed from datasets
         for dataset in self._datasets:
             self.model.data_registry[dataset].remove_agent(self)
+        
+        indices = getattr(self, "_continuous_indices", {})
+        for idx_list in indices.values():
+            for idx in idx_list:
+                if hasattr(self.model, "state_tensor"):
+                    self.model.state_tensor.remove(idx)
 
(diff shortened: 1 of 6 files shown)


codebreaker32:
A lot of work is yet to be done. Some of them are:

1. Linking it with Observables(or mesa_signals)
2. Profiling the approaches
3. Maybe removing some dead codes or overabstraction
4. Adding comments and docstrings

I have opened this PR mainly to check for benchmarks and some early reviews(once I update the code with comments and docstrings)

codebreaker32:
Will continue in next PR

PR #3756, opened 2026-06-22
Add run_scenarios and storage 

### LINKED ISSUE #3762: Handling of replication_id=None in Scenario
How should one interpret `replication_id=None` in a scenario instance? At the moment, `replication_id` is a keyword argument on `Scenario` that defaults to `None`. When one uses `spawn_replications` or `from_dataframe`, and passes a number of replications, these are correctly used in the returned scenario instances. 

However, while working on #3756, I ran into a problem in case `replication_id=None`. We want to use (scenario_id, replication_id) as a unique key to track runs while doing a parameter sweep. If we want to then use this in e.g., a data frame, as our multi index, this creates all kinds of problems. First, the multi index will become a float because None is turned into NaN (which is a float, not an int). Moreover, You cannot do easy index based lookups if you set some rows to None/NaN. So, how should we handle this?

The overarching question here is whether to keep a parent child relationship between a base scenario and the replications that are spawned from it. 

Options
1. Modify `spawn_replications` to return the original scenario + n-1 additional spawned replications. In this way, if `replication_id` is not passed to the init of `Scenario`, we can use 0 instead of None. This behavior, however, differs from how `spawn` behaves on a numpy random number generator. 
2. Use an integer placeholder such as -1 instead of None. This is simple, avoids all problems but needs to be carefully documented and explained.
3. Modify from_dataframe to always set `replication_id` even in `replications=None`. This is a partial fix that will cover the most common code path when doing parameter sweeps, but it does not solve the underlying root problem.

quaquel:
This PR builds on #3753 by adding `run_scenarios`, a protocol for storage (`Store`, `Writer`, and `Reference`), and an implementation of the protocol for in memory storage (`InMemoryStore`, `InMemoryWriter`, and `InMemoryReference`). 

The main focus here is hashing out the protocol. For background, see https://github.com/mesa/mesa/discussions/3483#discussioncomment-17323416. 

The overarching aim is to replace batch runner with the following basic API

```python
# generate samples with scipy.qmc or salib

scenarios = MyScenario.from_dataframe(samples, rng=42, replications=10)
run_config = RunConfig(MyModel, until=150)

with ProcessPoolExecutor() as executor:
    results = run_scenarios(scenarios, run_config, executor=executor, store=InMemoryStore())
```

The main challenge is to design all of this such that we can support sequential local running (executor=None), parallel local running (`ProcessPoolExecutor`), and HPC (`mpi4py.futures.MPIPoolExecutor`). This brings with it a wide range of design considerations. 

Some key considerations are
* We want to minimize the amount of data being exchanged between the root process and the worker processes.
* We want to avoid creating many small files (unless you want to be kicked of the HPC by the system administrators)
* We want to avoid crashing an ensemble run if any individual scenario fails for whatever reason. Instead, we want to log the scenario/replication combination that failed and its cause and move on

To satisfy these and other considerations, this PR adds a bunch of things

1. A `Store` protocol that specifies the basic interface of any storage solution. Store always runs root side.
2. A `Writer` protocol that specifies how output from `RunConfiguration.extract_output` is to be send back to the root process (consideration 1). Writer always runs worker side.
3. A `Reference` protocol that specifies the fields in a message being send from the worker to the root
4. Example implementations of all three protocols are provided for in memory storage. So we have `InMemoryStore`, `InMemoryWriter`, and `InMemoryReference`. A future PR will add a disk based implementation of all three protocols while accounting for consideration 2. Note that `InMemoryStore` is just a simple dict based backend and nothing more fancy. It's here now just to clarify how it all works. 
5. There is an implementation of `run_scenarios` with support for sequential running. This implementation also handles the safe running of scenarios with graceful handling of failing scenarios (consideration 3). A future PR will add support for `ProcessPoolExecutor` and `MPIPoolExecutor` which have the same API so can be handled in the same way. 

The sequence diagram shown below shows how it all fits together in practice. 

Untitled diagram-2026-06-25-141239


quaquel:
I have done some more work on this to simplify and streamline it. I can really benefit from feedback at this point, so @tpike3, @EwoutH, and @codebreaker32, please chime in. To help clarify what is going on, please check the sequence diagram below. 

It shows the flow of information and the basic protocols as specified by this PR. The main thing not yet fully settled is what happens if a model uses a file based DataRecorder. I want to leave that to a future PR. 

Untitled diagram-2026-06-25-141239

mesa/experimental/scenarios/store.py
```diff
@@ -0,0 +1,222 @@
...
+            raise ScenarioNotReadyException(run_id)
+        if record.status == Status.FAILED:
+            raise ScenarioFailedException(run_id, record.failure)
+        return record.output
+
+    def write_scenarios(self, scenarios: list[Scenario]) -> None:
+        """Record the full ensemble of scenarios before dispatch."""
+        for scenario in scenarios:
+            key = RunId(scenario.scenario_id, scenario.replication_id)
+            self._runs[key] = RunRecord(scenario=scenario)
```
codebreaker32, line 188: In your design notes, you mentioned in open items
>write_design idempotency — "initialize only missing points as pending," not clobber; required for resume.

Should we enforce it by adding a small check(`if key not in self._runs`)?

mesa/experimental/scenarios/store.py
```diff
@@ -0,0 +1,222 @@
...
+        record.failure = failure
+
+    def status(self) -> pd.DataFrame:
+        """One row per design point: pending / succeeded / failed."""
+        idx = pd.MultiIndex.from_tuples(
+            [astuple(run_id) for run_id in self._runs],
+            names=[f.name for f in fields(RunId)],
+        )
+        return pd.DataFrame(
+            [r.status for r in self._runs.values()],
```
codebreaker32: `status()` populates the `DataFrame` column with the raw Status Enum objects. This breaks  Pandas string filtering. If a user types `df[df['status'] == 'FAILED']`, it will silently return an empty `DataFrame` because the string doesn't match the Enum object.
Extracting the `.value` is a better UX in my opinion.

codebreaker32 (commented):
First off, incredible work on this. Translating our discussions into this concrete `Store` -> `Writer` -> `Reference` protocol is a great leap forward for Mesa. I appreciate putting together that sequence diagram, it makes the process boundary and data flow instantly clear. I would try to find some more time and play with this a bit more. 

>The main thing not yet fully settled is what happens if a model uses a file based DataRecorder. I want to leave that to a future PR.

Agree. I would also be happy to help up there :)

mesa/experimental/scenarios/store.py
quaquel: This only becomes an issue once we add support for resuming runs from a `Store` object. So for now, I think the current code is fine.

mesa/experimental/scenarios/store.py
quaquel: good catch, thanks

quaquel:
@codebreaker32, thanks for the feedback. 

You have Claude to thank for the sequence diagram. But it proved really useful while I was trying to clean up and simplify the work flow. 

> Agree. I would also be happy to help up there :)

Your input will be much appreciated. It is likely that adding a `FileBasedStore` will require some careful deliberation and might affect some of the existing `DataRecorder` classes. The main challenge is that you don't want to save each individual run into its own separate small file. In particular on shared file systems as used on HPC, that is a quick route to a ban. These file systems don't like it when you create 100k files. So, the best solution is to have a writer on each worker that takes the data from a single run and sends it to a file. In that way you only create as many files as you have workers. So if you have 128 workers (2 nodes on my university HPC), you only create 128 files instead of 100k files. 

Another realization I had, is that, if we store the data in long form, with scenario_id, replication_id, and tilmestep as a combined key, we can just append rows to a table. A groupby operation allows you to quickly extract the individual time series, even if they have varying length. This solves the jagged run length problem we discussed a while back.

codebreaker32:
>So, the best solution is to have a writer on each worker that takes the data from a single run and sends it to a file. In that way you only create as many files as you have workers. So if you have 128 workers (2 nodes on my university HPC), you only create 128 files instead of 100k files.

The main implementation hurdle we will have to figure out for that future PR is how the Writer object dynamically identifies its worker ID, whether through `mpi4py` ranks or standard `multiprocessing` process IDs, so it knows exactly which file handle to own.


> Another realization I had, is that, if we store the data in long form, with scenario_id, replication_id, and tilmestep as a combined key, we can just append rows to a table. A groupby operation allows you to quickly extract the individual time series, even if they have varying length. This solves the jagged run length problem we discussed a while back.

it is also the exact format that `.parquet` files are heavily optimized for under the hood. Because Parquet uses columnar dictionary encoding, repeating the same `scenario_id` and `replication_id` down a column for hundreds of rows takes up virtually zero bytes of disk space, meaning the output will compress beautifully

quaquel:
> The main implementation hurdle we will have to figure out for that future PR is how the Writer object dynamically identifies its worker ID, whether through mpi4py ranks or standard multiprocessing process IDs, so it knows exactly which file handle to own.

The easiest will be to go with the PID because this works for bot multiprocessing and MPI. 

> it is also the exact format that .parquet files are heavily optimized for under the hood. Because Parquet uses columnar dictionary encoding, repeating the same scenario_id and replication_id down a column for hundreds of rows takes up virtually zero bytes of disk space, meaning the output will compress beautifully

That is good to know and an extra reason to favor parquet (perhaps even for making InMemoryStore persistent on disk.

codebreaker32 (approved):
Looks Good To Me!

quaquel:
Thanks! I'll give @EwoutH and @tpike3 a chance to chime in as well before merging it. In the meantime, I'll start on the next step: parallel processing.

quaquel:
For the record, I pushed an update that back ports some changes identified in the next PR for parallelization support as well as forcing replication_id to always be an int (see #3762).

EwoutH:
Bit of a hectic time. Will try to review this weekend.

EwoutH:
Thanks for this PR. You can see you built this earlier (with the EMAworkbench) it's impressive!

Few questions:

1. Scope: Previously I think you stated that running (different model configurations) was out of scope and should be handled by external packages. What changed? Or was that only the configuration of scenario's, and is this just providing a convenient, exposed interface?
2. Storage/serialization: is the plan for the disk backend to delegate to Parquet rather than a custom format? `pyarrow.dataset` already handles the "one logical dataset, many partitions" case that consideration 2 is about, would that be useful? And in the parallel case, do workers write their own partitions (then consolidate) or send everything back to root for writing? The `Writer` seam supports both, the disk backend has to pick a lane I think.
3. Resumption/restart: right now `write_scenarios` marks everything `PENDING` and the loop runs all of it. It doesn't skip completed runs, so a sweep that dies at 9,000/10,000 restarts from zero. And do we want caching and/or automated retries?
4. Store abstraction: would it be useful keeping `Store` as the pure `Protocol` and adding an `AbstractStore(ABC)` for the derived methods? `succeeded`/`failed`/`pending` are identical one-liners for any record-keyed store, and `status()` is ~10 lines that'd be copied verbatim into every backend.
6. Is `InMemoryStore` the right *default* when `store=None`, given it's illustrative and holds every run's dataframes in memory? Especially if we consider checkpoints and retries?

quaquel:
Thanks for the comments and suggestions. Below my responses. 

> Scope: Previously I think you stated that running (different model configurations) was out of scope and should be handled by external packages. What changed? Or was that only the configuration of scenario's, and is this just providing a convenient, exposed interface?

The generation of samples (e.g., Salib, numpy qmc) and the execution (e.g., ProcessPoolExecutor, MPIPoolExecutor) are out of scope. This series of PRs provides the connection between those libraries and Mesa. Below is a quick example from my testing that shows how it fits together. As you can see, this code uses `rescale_samples`, `run_scenario`, and `RunConfiguration` from mesa in combination with numpy, pandas, and concurrent.futures. 

```python
    samples = sp.stats.qmc.LatinHypercube(4, rng=42).random(50)
    samples = rescale_samples(samples, np.array([[0.5,1],[0.05, 0.5],[0.2, 0.8],[1,5]]))
    samples[0, :] = np.floor(samples[3, :]) # radius needs to be an integer, this means it will be between 1-4.
    samples = pd.DataFrame(samples, columns=["density", "minoriy_pc", "homophily", "radius"])
    samples.radius = samples.radius.astype(int)

    scenarios = SchellingScenario.from_dataframe(samples, replications=2)

    with ProcessPoolExecutor(max_workers=4) as executor:
        results = run_scenarios(scenarios, RunConfiguration(Schelling, until=100), executor=executor)

    print(results)

```


> Storage/serialization: is the plan for the disk backend to delegate to Parquet rather than a custom format? pyarrow.dataset already handles the "one logical dataset, many partitions" case that consideration 2 is about, would that be useful? And in the parallel case, do workers write their own partitions (then consolidate) or send everything back to root for writing? The Writer seam supports both, the disk backend has to pick a lane I think.

I am about to start on OnDiskStore and I am inclined to provide a reference implementation using parquet but I will check pyarrow as well. In the parallel case, each worker will have its own file. On a big MPI cluster, you don't want to send large amounts of data over the network. 

> Resumption/restart: right now write_scenarios marks everything PENDING and the loop runs all of it. It doesn't skip completed runs, so a sweep that dies at 9,000/10,000 restarts from zero. And do we want caching and/or automated retries?

This will be handled by a later PR. As you can see in #3768, It is one of the last items on the to do list. 

> Store abstraction: would it be useful keeping Store as the pure Protocol and adding an AbstractStore(ABC) for the derived methods? succeeded/failed/pending are identical one-liners for any record-keyed store, and status() is ~10 lines that'd be copied verbatim into every backend.

I'll have to see now that I am about to start on an `OnDiskStore` implementation. If there is overlap between both the `InMemoryStore` and the `OnDiskStore`, I'll factor out the overlap into an ABC. 

> Is InMemoryStore the right default when store=None, given it's illustrative and holds every run's dataframes in memory? Especially if we consider checkpoints and retries?

For smal local debug runs, I think it is a fine default. Moreover, an `InMemoryStore` default is tricky because you need a file name and directory.

mesa/experimental/scenarios/exceptions.py
```diff
@@ -0,0 +1,188 @@
...
+            run_id: identifier of the run that failed during execution
+        """
+        super().__init__(f"Model run failed for {run_id}")
+        self.run_id = run_id
+
+
+class OutcomeExtractionException(RunStageException):
+    """Raised when extracting outcomes from a finished model fails."""
+
+    origin = FailureOrigin.EXTRACTING
```
edgeofchaos42: I would make this 
`origin = FailureOrigin = FailureOrigin.Extracting`

for consistency

edgeofchaos42 (approved):
LGTM! This is very awesome! Just one nitpick

tpike3 (approved):
LGTM

PR #3759, opened 2026-06-23
Added __repr__ method to Agent class

### LINKED ISSUE #3758: Agent is missing __repr__ inconsistent with every other core Mesa class
**SUMMARY**
`Agent` defines `__str__` but not `__repr__`. This means `repr(agent)` falls back to `object.__repr__`, returning a useless memory address that changes every run.

This makes agents invisible in debuggers, logging output, and interactive shells.

**Current behaviour**
```python
class Wolf(Agent):
    pass

model = Model()
wolf = Wolf(model)

print(str(wolf))   # Wolf, agent_id = 0        
print(repr(wolf))    
```

**Expected behaviour**

```python
print(repr(wolf))  # Wolf(unique_id=0) 
```
**This is an internal inconsistency**
Every other core Mesa class already defines `__repr__`. `Agent` is the only one that was skipped. The pattern is already established in the codebase. `Agent` was one core object where it was never defined.
**Precedent**
PR #2195 added `__repr__` to `EventList` for exactly this reason. Same fix.

**I'd like to submit a PR for this**

dietcoke-17:
**Summary**
Adds `__repr__` method to the `Agent` class to provide meaningful string representations in debuggers, logs, and interactive shells.

**Problem**
The `Agent` class defined `__str__` but not `__repr__`, causing `repr(agent)` to fall back to `object.__repr__` which returns an unhelpful memory address that changes every run. This makes agents invisible in debuggers and logging output.

**Solution**
Implemented `__repr__` to return a descriptive string following the format: `ClassName(unique_id=<value>)`

**Consistency**
This change brings Agent in line with other core Mesa classes (EventList, Model, etc.) which already define __repr__. The pattern established in PR #2195 is now applied to Agent.

**Testing**
Added three new tests to verify:

test_agent_str(): Validates the existing __str__ method
test_agent_repr(): Validates the new __repr__ method
test_agent_repr_subclass(): Ensures __repr__ correctly uses the subclass name (e.g., Wolf not Agent)

**Fixes**
Closes #3758

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -205,9 +205,9 @@ class Agent[M: Model]:
 
         return AgentSet(agents, random=model.random)
 
-    def __str__(self) -> str:
-        """Return a human-readable string representation of the agent."""
-        return f"{self.__class__.__name__}, agent_id = {self.unique_id}"
+    def __repr__(self) -> str:
+        """Return an unambiguous string representation of the agent."""
+        return f"{self.__class__.__name__}(unique_id={self.unique_id})"
 
     @property
     def random(self) -> Random:
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -205,6 +205,10 @@ class Agent[M: Model]:
 
         return AgentSet(agents, random=model.random)
 
+    def __str__(self) -> str:
+        """Return a human-readable string representation of the agent."""
+        return f"{self.__class__.__name__}, agent_id = {self.unique_id}"
+
     def __repr__(self) -> str:
         """Return an unambiguous string representation of the agent."""
         return f"{self.__class__.__name__}(unique_id={self.unique_id})"
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -37,7 +37,7 @@ class Agent[M: Model]:
     """
 
     _datasets: ClassVar = set()
-    _repr_excluded_fields = {"model", "current_action"}
+    _repr_excluded_fields: ClassVar[set[str]] = {"model", "current_action"}
 
     def __init_subclass__(cls, **kwargs):
         """Called when DatasetTrackedAgent is subclassed."""
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -212,21 +212,22 @@ class Agent[M: Model]:
     def __repr__(self) -> str:
         """Return an unambiguous string representation including agent state."""
         # Mesa default fields to exclude
-        mesa_fields = {'model', 'unique_id', 'current_action'}
-        
+        mesa_fields = {"model", "unique_id", "current_action"}
+
         # Get user-defined attributes (exclude private and Mesa fields)
         user_attrs = {
-            k: v for k, v in self.__dict__.items()
-            if not k.startswith('_') and k not in mesa_fields
+            k: v
+            for k, v in self.__dict__.items()
+            if not k.startswith("_") and k not in mesa_fields
         }
-        
+
         # Build the repr string
         if user_attrs:
             attr_str = ", ".join(f"{k}={v!r}" for k, v in user_attrs.items())
             return f"<{self.__class__.__name__} id={self.unique_id} {attr_str}>"
         else:
             return f"<{self.__class__.__name__} id={self.unique_id}>"
-    
+
     @property
     def random(self) -> Random:
         """Return a seeded stdlib rng."""
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -210,9 +210,23 @@ class Agent[M: Model]:
         return f"{self.__class__.__name__}, agent_id = {self.unique_id}"
 
     def __repr__(self) -> str:
-        """Return an unambiguous string representation of the agent."""
-        return f"{self.__class__.__name__}(unique_id={self.unique_id})"
-
+        """Return an unambiguous string representation including agent state."""
+        # Mesa default fields to exclude
+        mesa_fields = {'model', 'unique_id', 'current_action'}
+        
+        # Get user-defined attributes (exclude private and Mesa fields)
+        user_attrs = {
+            k: v for k, v in self.__dict__.items()
+            if not k.startswith('_') and k not in mesa_fields
+        }
+        
+        # Build the repr string
+        if user_attrs:
+            attr_str = ", ".join(f"{k}={v!r}" for k, v in user_attrs.items())
+            return f"<{self.__class__.__name__} id={self.unique_id} {attr_str}>"
+        else:
+            return f"<{self.__class__.__name__} id={self.unique_id}>"
+    
     @property
     def random(self) -> Random:
         """Return a seeded stdlib rng."""
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -37,6 +37,7 @@ class Agent[M: Model]:
     """
 
     _datasets: ClassVar = set()
+    _repr_excluded_fields = {"model", "current_action"}
 
     def __init_subclass__(cls, **kwargs):
         """Called when DatasetTrackedAgent is subclassed."""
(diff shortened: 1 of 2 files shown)


mesa/agent.py
```diff
@@ -209,6 +209,10 @@ def __str__(self) -> str:
         """Return a human-readable string representation of the agent."""
         return f"{self.__class__.__name__}, agent_id = {self.unique_id}"
 
+    def __repr__(self) -> str:
+        """Return an unambiguous string representation of the agent."""
+        return f"{self.__class__.__name__}(unique_id={self.unique_id})"
+
```
quaquel: I am not sure about this. It suggests that you can initialize the agent with unique_id which is not correct at all. 

Also, more broadly, should we not add these to all agent classes in mesa (so cell agent shows the cell field as well)?

dietcoke-17:
@quaquel Thanks for the feedback! quick clarifications required:

**On the format:**
You're right — `Agent(unique_id=0)` is misleading because `unique_id` can't actually be passed to `__init__`. 

What format would be clearer? which should i prefer:
- `Agent[id=0]` (brackets instead of parens to signal it's not a constructor)
- `<Agent id=0>` (angle brackets, like AgentSet)
- `Agent@0` (symbol notation)
- Something else would be better 

**Broader scope**: 
You're right — we should add `__repr__` to all agent classes consistently, not just Agent. Should I expand this PR to include  `CellAgent` (show `cell` field) or separtely we would adress this in another PR which would be easier for you i will move in that direction.

EwoutH:
Thanks for the PR.

I think it's worth checking whether `__str__` and `__repr__` should return (meaningfully) different things here. A common practice is to define only `__repr__` unless you need a distinct human-friendly format. Since `str()` falls back to `__repr__` anyway, one method covers both. If they're returning essentially the same info, dropping `__str__` and keeping just `__repr__` would be cleaner.

quaquel:
> What format would be clearer? which should i prefer:

I would stay close to agentset, but @EwoutH does raise a relevant point as well. So curious to hear your thoughts on this.

dietcoke-17:
Thanks @EwoutH @quaquel for the query
Looking at Mesa's design philosophy from contributing.md and the JOSS paper, what i had the picture on my mind that mesa is designed to serve both beginner and advanced programmers with one clean API, not separate ones.
 
For `__str__` vs `__repr__`, I propose we follow that principle:
I think we should drop `__str__`, keep only `__repr__`** and why due to the following consideration from my side:-
1.Beginners can see clear, context-rich output
2.Developers will see unambiguous, searchable output
3.One method, two use cases consistent with Mesa's philosophy of unified APIs
4.Follows Python best practice that Ewouth sir has already mentioned that `str()` naturally falls back to `__repr__`

**This is the same approach Mesa used for SolaraViz and AgentSet one good API that serves both audiences, not duplication.**

dietcoke-17:
@EwoutH @quaquel should i go with __repr__ and make changes in the pr according to that or should we stick to our __str__?

quaquel:
I like it when `str` returns a human readable string while `repr` is more technical. For something like `Agent` that might be overkill unless `repr` does some clever parsing of `__dict__` so you get more insight into the state of the agent. Hope this helps.

dietcoke-17:
Thank You @quaquel

dietcoke-17:
I want to clarify your suggestion about "clever parsing of __dict__".

You said `__repr__` might be worth doing if it shows more insight into the agent's state. I have three options:

**Option A: Just unique_id (current PR)**
```python
repr(wolf)  # → Wolf(unique_id=0)
```
Simple, but not much state info.

**Option B: Show all agent attributes from __dict__**
```python
repr(wolf)  # → <Wolf id=0 wealth=100 energy=50 state='active'>
```
Shows full agent state. Flexible across subclasses.

**Option C: Make it base implementation + override-friendly**
```python
# Base class shows just id:
repr(agent)  # → <Agent id=0>

# Subclasses can override to show their fields:
class Wolf(Agent):
    def __repr__(self):
        return f"<Wolf id={self.unique_id} wealth={self.wealth}>"

repr(wolf)  # → <Wolf id=0 wealth=100>
```
Gives subclasses control over what to show.

Which direction makes sense? And should this apply to all agent classes (Agent, CellAgent, etc.) or start with Agent and iterate?

I want to make sure this is worth the complexity before implementing.

quaquel:
What I had in mind is a bit different. Your idea forces the user to implement a dunder method to get useful information. However, we can already handle this by including `__dict__` inside how we implement `__repr__`. We just need to filter out any default mesa fields and probably anything with a leading underscore.

dietcoke-17:
This implements  `__repr__`  method that automatically shows user-defined attributes by parsing __ dict __ filtered to exclude:
- Mesa internal fields (model, unique_id, current_action)
- Private attributes (starting with _)

Keeps `__str__` unchanged for human-readable output.
No user override needed; works automatically for all Agent subclasses.

Example output:
  repr(wolf)  # → <Wolf id=0 wealth=100 energy=50>
  str(wolf)   # → Wolf, agent_id = 0

Implemented smart approach — Auto-parse `__dict__` to show user-defined attributes:

-Filters out Mesa internal fields (model, unique_id, current_action)
-Filters out private attributes (starting with _)
-No user override needed — works automatically for all sub-classes

what changed in which file
mesa/agent.py:

Added `__repr__` method below existing `__str__`
Shows class name, id, and all custom attributes
Uses angle brackets <ClassName id=...> format (consistent with AgentSet)

tests/test_agent.py:

Added test_agent_repr() — tests repr with custom attributes
Added test_agent_repr_basic() — tests repr with no custom attributes
Added test_agent_repr_subclass() — tests subclass name appears correctly
```
class Wolf(Agent):
    def __init__(self, model):
        super().__init__(model)
        self.wealth = 100
        self.energy = 50

wolf = Wolf(model)
```
# Before:
repr(wolf)  # → <mesa.agent.Wolf object at 0x7f3b2c1d4a90>  ❌

# After:
repr(wolf)  # → <Wolf id=0 wealth=100 energy=50>  ✅
str(wolf)   # → Wolf, agent_id = 0  ✅

mesa/agent.py
```diff
@@ -209,6 +209,25 @@ def __str__(self) -> str:
         """Return a human-readable string representation of the agent."""
         return f"{self.__class__.__name__}, agent_id = {self.unique_id}"
 
+    def __repr__(self) -> str:
+        """Return an unambiguous string representation including agent state."""
+        # Mesa default fields to exclude
+        mesa_fields = {"model", "unique_id", "current_action"}
```
quaquel: We now hard code these fields in this method. Is there a more generic, potentially user extendable way of handling this? 

Also, I would always include unique_id, so I suggest removing it from this set.

mesa/agent.py
```diff
@@ -209,6 +209,25 @@ def __str__(self) -> str:
...
+        # Get user-defined attributes (exclude private and Mesa fields)
+        user_attrs = {
+            k: v
+            for k, v in self.__dict__.items()
+            if not k.startswith("_") and k not in mesa_fields
+        }
+
+        # Build the repr string
+        if user_attrs:
+            attr_str = ", ".join(f"{k}={v!r}" for k, v in user_attrs.items())
```
quaquel, line 226: what does `!r! do?

quaquel (commented):
It looks fine but I have some clarifying questions and a minor suggestion.

mesa/agent.py
dietcoke-17: The !r calls repr() on the value. It ensures strings are quoted and also 
f"wealth={wealth}, name={name}"
→ "wealth=100, name=wolf"   --here name looks like code
but
f"wealth={wealth!r}, name={name!r}"
→ "wealth=100, name='wolf'"   --string is quoted
It also ensures that when someone reads the __repr__ output, they immediately understand:
-What type each value is
-Whether it's a string, number, None, etc.
-No ambiguity

dietcoke-17:
Added comprehensive coverage tests for __repr__ filtering logic

- Tests filtering of Mesa internal fields (model, current_action)
- Tests filtering of private attributes (starting with _)
- Tests repr formatting with various types (strings, numbers, None, lists)

Also while adding new test found out that there are some missing test needed to be added in the agent which will increse the overall codecove score overall

quaquel (approved):
thanks

PR #3764, opened 2026-06-29
Refactor: extract dependency-tracked evaluation from `computed_property` into `ComputedState.evaluate()`

codebreaker32:
### Summary
Extracts the dependency-tracking evaluation logic out of `computed_property`'s `wrapper` closure into a new `ComputedState.evaluate()` method. No behavioral change for existing `computed_property` users, this is a pure refactor that gives `ComputedState` a public, reusable evaluation primitive instead of keeping that mechanism locked inside one decorator's closure.

### Motive
`computed_property`'s `wrapper` inlines a fairly involved block: enter the dependency-tracking frame (`CURRENT_COMPUTED`, `PROCESSING_SIGNALS`), call the user function, diff old vs. new parent Observables, subscribe/unsubscribe accordingly, and clear `is_dirty`. `ComputedState` already owns all the relevant data (`value`, `parents`, `is_dirty`, `func`), but had no method that could actually run this evaluation, the logic only existed inline inside the decorator, so it was unreachable from anywhere else.

That's a real limitation: dependency-tracked reactive evaluation is a generally useful primitive beyond simple computed properties, for example, any feature that needs to re-evaluate a function whenever the Observables it reads change (a derived rate, a derived rule, a cached lookup keyed on agent state) would currently have no way to reuse Mesa's tracking machinery short of re-implementing the same save/restore-globals + diff-and-resubscribe dance by hand. Making `evaluate()` a method on `ComputedState` turns this from a decorator-only trick into something any future caller can construct and drive directly.

### Implementation
- Added `ComputedState.evaluate()`, which performs exactly the block that used to be inlined in `wrapper`: enters the tracking frame, calls `self.func(self.owner)`, diffs `parents` before/after, resubscribes/unsubscribes, and clears `is_dirty` on success.
- `computed_property`'s `wrapper` now calls `state.evaluate()` when its own staleness check (`changed`) determines a recompute is needed, and keeps the existing unconditional `state.is_dirty = False` afterward — preserving the original structure exactly (this line is a no-op when `evaluate()` ran, since `evaluate()` already cleared it; it's the only thing that clears `is_dirty` when `changed` was `False`).
- No change to `wrapper`'s own staleness check (the "did any parent's value actually change" loop), that policy stays where it is; only the "how do we evaluate and re-track" mechanism moved.

### Usage Examples
Before, this evaluation mechanism only existed inline in the decorator and wasn't reusable:
```python
# old: only reachable through @computed_property
@computed_property
def area(self):
    return self.width * self.height
```

After, the same dependency-tracked evaluation is available directly on `ComputedState` for any caller that needs reactive, automatic-dependency-tracking evaluation outside the property-decorator pattern, for instance, something that derives a per-agent rate or value from other Observables and needs it kept in sync without manually wiring up subscriptions:

```python
comp_state = ComputedState(owner=agent, name="rate", func=lambda a: a.acceleration)
...
if comp_state.is_dirty:
    comp_state.evaluate()   # tracks `acceleration` as a dependency automatically
value = comp_state.value
```

### Additional Notes

**This PR touches only `core.py` and is self-contained; it doesn't change any existing public behavior.**

### DIFF
--- a/mesa/experimental/mesa_signals/core.py
+++ b/mesa/experimental/mesa_signals/core.py
@@ -225,6 +225,57 @@ class ComputedState:
         except KeyError:
             self.parents[parent] = {name: current_value}
 
+    def evaluate(self) -> Any:
+        """Recompute the value by calling self.func, tracking dependencies as it runs.
+
+        Any Observable read during the call is recorded as a dependency, and
+        subscriptions are updated so this state is marked dirty again when
+        one of them changes. On success, value is updated and is_dirty is
+        cleared.
+
+        Raises:
+            Exception: Any exception raised by self.func propagates
+                unchanged. parents is restored to its prior state and
+                is_dirty is left set, so the caller may retry later.
+
+        Returns:
+            The newly computed value.
+        """
+        global CURRENT_COMPUTED  # noqa: PLW0603
+
+        old_computed = CURRENT_COMPUTED
+        # Save and reset PROCESSING_SIGNALS so that `self.func` accessing an
+        # Observable that is itself derived from this same ComputedState
+        # (e.g. a rate function reading the ContinuousState it drives) does
+        # not falsely trigger a cyclical-dependency error.
+        old_processing = set(PROCESSING_SIGNALS)
+        PROCESSING_SIGNALS.clear()
+
+        CURRENT_COMPUTED = self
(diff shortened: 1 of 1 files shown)


quaquel (approved):
makes good sense

PR #3765, opened 2026-06-29
Fix incorrect parameter documentation in VoronoiGrid

eeshsaxena:
**Summary**

The `VoronoiGrid.__init__` docstring (`mesa/discrete_space/voronoi.py`) documented a `capacity_function` parameter, but `capacity_function` is a local variable inside `__init__`, not an argument. The actual `capacity` parameter is typed `int | Callable | None` and already covers that behaviour.

This drops the nonexistent `capacity_function` entry and documents that `capacity` may be an int or a callable computing a cell's capacity from its polygon area.

Documentation-only change.


dietcoke-17 (approved):
Good catch, LGTM

quaquel (approved):

PR #3766, opened 2026-06-29
Add `ContinuousState` and `Threshold` for event-driven continuous-time dynamics

codebreaker32:
### Summary
Adds `mesa/experimental/states/` — `ContinuousState` and `Threshold`, giving Mesa agents exact, event-driven continuous-time kinematics instead of step-by-step polling.

### Motive
Agents with continuously-changing quantities — speed, position, energy, inventory drain, anything with a rate — have had two bad options in Mesa: poll every tick (cost scales with step count, accuracy scales with step size), or hand-roll time-projection math per model. Neither generalizes.

This module gives a declarative alternative: state a quantity's rate of change once, and get its exact value at any future time plus a callback the instant it crosses a threshold — computed once analytically, not approximated by sampling. A transit agent whose speed and position must be queryable at arbitrary times, without the model stepping every second just to check "have we reached the brake point yet," is the motivating case; a tram example exercising it end-to-end follows in a separate PR.

### Implementation
- **`ContinuousState`** (a `BaseObservable` descriptor): tracks `base_value`/`last_time`/`current_rate` and extrapolates `value(t) = base + rate·dt`. `rate` is either a constant float or a callable `f(agent) -> float` evaluated reactively through the `mesa_signals` `ComputedState` diffing engine, so a rate change snapshots the current value as the new baseline and the trajectory stays mathematically continuous across the switch.
  - **Chained states**: when a state's rate is itself another `ContinuousState` — `position`'s rate is `speed`, which is itself changing under acceleration — `_find_chained_rate` detects it from the recorded dependencies and adds the next-order term, `0.5·second_order_rate·dt²`, so position integrates exactly through an acceleration phase instead of drifting between reads.
  - Declaring the state auto-generates a shadow `{name}_rate` Observable, with a `__set_name__` guard that raises on namespace collisions (checked against both `__dict__` and `__slots__`).
  - Assigning inside a computed context raises on cyclical dependencies, and an init-ordering race (a rate lambda reading an Observable whose backing store doesn't exist yet, mid-`Agent.__init__`) is narrowly detected and treated as "at rest," while genuine typos in the rate lambda still raise.

- **`Threshold`**: fires a named agent callback when a `ContinuousState` crosses a value, with the crossing time solved analytically rather than polled. The solver branches on whether the state is chained — plain linear solve for a constant-rate state, exact quadratic root solve (`0.5·a·t² + v·t + (x₀−limit) = 0`) for one chained off another `ContinuousState`. `direction` (`"rising"` / `"falling"` / `"crossing"`) is enforced against the velocity *at the root*, so tangent touches bounce rather than fire, and a zero-time root while already sitting on the limit (the mid-callback case, where `brake()` sets acceleration while `position == limit`) is rejected instead of immediately re-firing.
  - **Limits are declarative.** `limit` is a fixed float *or* an `Observable` on the agent. When it's an Observable, `bind()` subscribes to the limit's `CHANGED` signal alongside the state's, so a per-instance runtime target — a tram's brake point depends on its own route, not a class-wide constant — is set by plain assignment and recalculates itself. No imperative `set_limit()`/`rearm()` calls, and re-arming across repeated cycles falls out of the same mechanism.
  - Each threshold owns exactly one live entry on Mesa's native event queue via `model.schedule_event`, cancelling and re-scheduling as the trajectory shifts; a "never crosses" result cancels and parks the projected time at `inf`.
  - Thresholds bind lazily — the first `ContinuousState` access on an instance walks the MRO for `_continuous_thresholds` and wires them up, so no scheduler or registration step is required.

### Usage
```python
from mesa.experimental.mesa_signals import HasEmitters, Observable
from mesa.experimental.states import ContinuousState, Threshold


class Tram(Agent, HasEmitters):
    acceleration = Observable(fallback_value=0.0)
    speed = ContinuousState(fallback_value=0.0, rate=lambda a: a.acceleration)
    position = ContinuousState(fallback_value=0.0, rate=lambda a: a.speed)  # chained: position' = speed
    brake_point = Observable(fallback_value=float("inf"))

    # fixed limit -- same cruise speed for every tram
    _cruise = Threshold(state=speed, limit=15.0, callback="start_coasting", direction="rising")

    # Observable limit -- per-instance, recalculates whenever brake_point is assigned
    _brake_point = Threshold(state=position, limit=brake_point, callback="brake", direction="rising")

    _stop = Threshold(state=speed, limit=0.0, callback="arrive_at_station", direction="falling")

    def depart(self):
        self.brake_point = self.next_station - self.braking_distance()  # re-arms automatically
        self.acceleration = 2.0

    def brake(self):
        # fires with an exact crossing time whether the tram is still
        # accelerating, coasting, or already braking when it reaches brake_point
        self.acceleration = -3.0
```

### Additional Notes
A future direction worth flagging now: `_find_chained_rate` looks one link deep and adds a single quadratic term. It works because the chain is exactly two `ContinuousState`s long (`position → speed → acceleration`, a discrete `Observable`). A natural generalization is an **`ODEState`**: instead of a single scalar rate, it would hold a rate *function* `f(t, y) -> dy/dt` (or a small system of them) stepped forward with a proper numerical integrator (RK4 or an adaptive method) rather than a closed-form polynomial per order. That would let arbitrary chains, or genuinely nonlinear dynamics such as drag proportional to `speed²`, be expressed directly — at the cost of `Threshold` losing its exact analytical crossing time and needing bisection/interpolation on the integrator's output instead of a polynomial root. Worth keeping in mind as a separate, later addition.


codebreaker32:
I have spent some good time building over @quaquel, @EwoutH and @tpike3 feedbacks(Thanks for it). Hope this implementation aligns with what you all have in mind.
I am currently working on Tram Model as suggested, will push the model by tomorrow to have better understanding

codebreaker32:
I am also open to integrate `ContinuousScheduler` into `mesa.time.events`

codebreaker32:
Example Run of Tram Model
```plaintext
Route: [0, 200, 400, 600, 800, 1000, 1200, 1400, 1600, 1800, 2000, 2200, 2400, 2600, 2800, 3000, 3200, 3400, 3600, 3800]
cruise_speed=15.0 m/s, acceleration=2.0 m/s^2, deceleration=3.0 m/s^2, braking_distance=37.50m

[t=0.00] Tram 1 departing (pos=0.00) towards station at 200.00m. Accelerating at 2.0 m/s^2.
[t=7.50] Tram 1 reached cruise speed 15.0 m/s. Coasting.
[t=14.58] Tram 1 at brake point (pos=162.50). Applying brakes at -3.0 m/s^2.
[t=19.58] Tram 1 arrived at station (pos=200.00).
[t=24.58] Tram 1 departing (pos=200.00) towards station at 400.00m. Accelerating at 2.0 m/s^2.
[t=32.08] Tram 1 reached cruise speed 15.0 m/s. Coasting.
[t=39.17] Tram 1 at brake point (pos=362.50). Applying brakes at -3.0 m/s^2.
[t=44.17] Tram 1 arrived at station (pos=400.00).
[t=49.17] Tram 1 departing (pos=400.00) towards station at 600.00m. Accelerating at 2.0 m/s^2.
[t=56.67] Tram 1 reached cruise speed 15.0 m/s. Coasting.
[t=63.75] Tram 1 at brake point (pos=562.50). Applying brakes at -3.0 m/s^2.
[t=68.75] Tram 1 arrived at station (pos=600.00).
[t=73.75] Tram 1 departing (pos=600.00) towards station at 800.00m. Accelerating at 2.0 m/s^2.
[t=81.25] Tram 1 reached cruise speed 15.0 m/s. Coasting.
[t=88.33] Tram 1 at brake point (pos=762.50). Applying brakes at -3.0 m/s^2.
[t=93.33] Tram 1 arrived at station (pos=800.00).
[t=98.33] Tram 1 departing (pos=800.00) towards station at 1000.00m. Accelerating at 2.0 m/s^2.
[t=105.83] Tram 1 reached cruise speed 15.0 m/s. Coasting.
[t=112.92] Tram 1 at brake point (pos=962.50). Applying brakes at -3.0 m/s^2.
[t=117.92] Tram 1 arrived at station (pos=1000.00).
[t=122.92] Tram 1 departing (pos=1000.00) towards station at 1200.00m. Accelerating at 2.0 m/s^2.
[t=130.42] Tram 1 reached cruise speed 15.0 m/s. Coasting.
[t=137.50] Tram 1 at brake point (pos=1162.50). Applying brakes at -3.0 m/s^2.
[t=142.50] Tram 1 arrived at station (pos=1200.00).
[t=147.50] Tram 1 departing (pos=1200.00) towards station at 1400.00m. Accelerating at 2.0 m/s^2.
[t=155.00] Tram 1 reached cruise speed 15.0 m/s. Coasting.
[t=162.08] Tram 1 at brake point (pos=1362.50). Applying brakes at -3.0 m/s^2.
[t=167.08] Tram 1 arrived at station (pos=1400.00).
[t=172.08] Tram 1 departing (pos=1400.00) towards station at 1600.00m. Accelerating at 2.0 m/s^2.
[t=179.58] Tram 1 reached cruise speed 15.0 m/s. Coasting.
[t=186.67] Tram 1 at brake point (pos=1562.50). Applying brakes at -3.0 m/s^2.
[t=191.67] Tram 1 arrived at station (pos=1600.00).
[t=196.67] Tram 1 departing (pos=1600.00) towards station at 1800.00m. Accelerating at 2.0 m/s^2.
[t=204.17] Tram 1 reached cruise speed 15.0 m/s. Coasting.
[t=211.25] Tram 1 at brake point (pos=1762.50). Applying brakes at -3.0 m/s^2.
[t=216.25] Tram 1 arrived at station (pos=1800.00).
[t=221.25] Tram 1 departing (pos=1800.00) towards station at 2000.00m. Accelerating at 2.0 m/s^2.
[t=228.75] Tram 1 reached cruise speed 15.0 m/s. Coasting.
[t=235.83] Tram 1 at brake point (pos=1962.50). Applying brakes at -3.0 m/s^2.
[t=240.83] Tram 1 arrived at station (pos=2000.00).
[t=245.83] Tram 1 departing (pos=2000.00) towards station at 2200.00m. Accelerating at 2.0 m/s^2.
[t=253.33] Tram 1 reached cruise speed 15.0 m/s. Coasting.
[t=260.42] Tram 1 at brake point (pos=2162.50). Applying brakes at -3.0 m/s^2.
[t=265.42] Tram 1 arrived at station (pos=2200.00).
[t=270.42] Tram 1 departing (pos=2200.00) towards station at 2400.00m. Accelerating at 2.0 m/s^2.
[t=277.92] Tram 1 reached cruise speed 15.0 m/s. Coasting.
[t=285.00] Tram 1 at brake point (pos=2362.50). Applying brakes at -3.0 m/s^2.
[t=290.00] Tram 1 arrived at station (pos=2400.00).
[t=295.00] Tram 1 departing (pos=2400.00) towards station at 2600.00m. Accelerating at 2.0 m/s^2.
[t=302.50] Tram 1 reached cruise speed 15.0 m/s. Coasting.
[t=309.58] Tram 1 at brake point (pos=2562.50). Applying brakes at -3.0 m/s^2.
[t=314.58] Tram 1 arrived at station (pos=2600.00).
[t=319.58] Tram 1 departing (pos=2600.00) towards station at 2800.00m. Accelerating at 2.0 m/s^2.
[t=327.08] Tram 1 reached cruise speed 15.0 m/s. Coasting.
[t=334.17] Tram 1 at brake point (pos=2762.50). Applying brakes at -3.0 m/s^2.
[t=339.17] Tram 1 arrived at station (pos=2800.00).
[t=344.17] Tram 1 departing (pos=2800.00) towards station at 3000.00m. Accelerating at 2.0 m/s^2.
[t=351.67] Tram 1 reached cruise speed 15.0 m/s. Coasting.
[t=358.75] Tram 1 at brake point (pos=2962.50). Applying brakes at -3.0 m/s^2.
[t=363.75] Tram 1 arrived at station (pos=3000.00).
[t=368.75] Tram 1 departing (pos=3000.00) towards station at 3200.00m. Accelerating at 2.0 m/s^2.
[t=376.25] Tram 1 reached cruise speed 15.0 m/s. Coasting.
[t=383.33] Tram 1 at brake point (pos=3162.50). Applying brakes at -3.0 m/s^2.
[t=388.33] Tram 1 arrived at station (pos=3200.00).
[t=393.33] Tram 1 departing (pos=3200.00) towards station at 3400.00m. Accelerating at 2.0 m/s^2.
[t=400.83] Tram 1 reached cruise speed 15.0 m/s. Coasting.
[t=407.92] Tram 1 at brake point (pos=3362.50). Applying brakes at -3.0 m/s^2.
[t=412.92] Tram 1 arrived at station (pos=3400.00).
[t=417.92] Tram 1 departing (pos=3400.00) towards station at 3600.00m. Accelerating at 2.0 m/s^2.
[t=425.42] Tram 1 reached cruise speed 15.0 m/s. Coasting.
[t=432.50] Tram 1 at brake point (pos=3562.50). Applying brakes at -3.0 m/s^2.
[t=437.50] Tram 1 arrived at station (pos=3600.00).
[t=442.50] Tram 1 departing (pos=3600.00) towards station at 3800.00m. Accelerating at 2.0 m/s^2.
[t=450.00] Tram 1 reached cruise speed 15.0 m/s. Coasting.
[t=457.08] Tram 1 at brake point (pos=3762.50). Applying brakes at -3.0 m/s^2.
[t=462.08] Tram 1 arrived at station (pos=3800.00).
[t=462.08] Tram 1 has completed its route.

Final time:     2000.00
Final position: 3800.00 m
```

codebreaker32:
I would appreciate some guidance @quaquel

quaquel:
@codebreaker32, I had a quick first pass but need some time to look at it in detail. I'll try to respond tomorrow the latest.

mesa/experimental/states/state.py
```diff
@@ -0,0 +1,585 @@
...
+    def _snapshot_trajectory(self, instance: HasEmitters) -> None:
+        """Commit the current extrapolated value as the new baseline when the rate changes."""
+        # Call order matters:
+        #   1. Extrapolate using the *old* cached rate(s) — prevents retroactive time-travel.
+        #   2. Lock in the new baseline and timestamp.
+        #   3. Evaluate and cache the *new* rates.
+        #   4. Notify Threshold observers that the trajectory has shifted.
+
+        state = self._get_state(instance)
+        current_time = getattr(instance.model, "_time", 0.0)
```
quaquel: is there no clean way to get time from the model?

mesa/experimental/states/state.py
```diff
@@ -0,0 +1,585 @@
...
+                    valid_times.append((t, v))
+        else:
+            # Quadratic Case
+            A = 0.5 * a
+            B = v
+            C = current_value - limit
+            discriminant = B**2 - 4 * A * C
+
+            # Allow microscopic negative discriminants due to float inaccuracies near tangent touches
+            if discriminant >= -1e-12:
```
quaquel: I think numpy has a standard field for this that you might reuse.

mesa/experimental/states/state.py
```diff
@@ -0,0 +1,585 @@
...
+                ObservableSignals.CHANGED,
+                old=old_value,
+                new=float(value),
+            )
+            core.PROCESSING_SIGNALS.discard(
+                _hashable_signal(instance, self.public_name)
+            )
+
+
+class ContinuousScheduler:
```
quaquel: I guess this is here because of agent interaction causing many canceled events. But is that really within scope for this PR?

mesa/experimental/states/tram_model.py
```diff
@@ -0,0 +1,200 @@
...
+            f"[t={self.model.time:.2f}] Tram {self.unique_id} departing "
+            f"(pos={self.position:.2f}) towards station at {target:.2f}m. "
+            f"Accelerating at {self.acceleration_rate:.1f} m/s^2."
+        )
+
+        brake_at = target - self.braking_distance()
+        Tram._brake_point.set_limit(self, brake_at)
+        Tram._cruise.rearm(self)
+        Tram._stop.rearm(self)
+
```
quaquel: what is going on here?

mesa/experimental/states/state.py
codebreaker32: If we switch to `instance.model.time`, here is the problem:

1. During initialization and trajectory recalculations, `ContinuousState` evaluates its diffing dependencies (like `lambda a: a.speed`). During this evaluation, it calls `__get__`, which now accesses the public `model.time`.

2. `ComputedState` engine is watching. It sees `ContinuousState` look at `model.time` and It registers `model.time` as a hard dependency.

3. Now, whenever `model.time` advances. Because it is a tracked dependency, this emits a `CHANGED` signal
4. This signal wakes up every `ContinuousState` in the model, forcing them to run `_snapshot_trajectory` on every single microscopic change in `time`. This completely destroys the continuous analytical parabolas, shattering them into hundreds of microscopic piecewise segments.


If we want to access it anyway we have to temporarily suppress `CURRENT_COMPUTED`, which I don't think is a good idea(also because it is a global variable. Thoughts?

mesa/experimental/states/state.py
codebreaker32: Yes, Thanks.

mesa/experimental/states/state.py
codebreaker32: Fair Enough, I will open separate PR for it.

mesa/experimental/states/tram_model.py
codebreaker32: This block handles resetting the thresholds for the next station segment. Because `Threshold` is a class-level descriptor, it uses a per-instance `fired` flag to prevent double-triggering.

* **`set_limit()`**: Assigns a new dynamic spatial target for `_brake_point` and automatically rearms it for the new segment.
* **`rearm()`**: Wakes up the `_cruise` and `_stop` thresholds. Since they fired and went dormant during the previous segment, they must be explicitly reset to watch for the next one.
 
 
Going dormant prevents infinite loop edge cases (e.g., preventing the brake callback from firing repeatedly while the tram's position continues past the brake line).

mesa/experimental/states/state.py
quaquel: Clear explanation. What it seems we need is a way of peaking at an observable without registering as a dependency. Does that make sense to you?

codebreaker32: Yes, similar to `suppress()`.

codebreaker32:
`Threshold.limit` now accepts an `Observable` in addition to a plain float. When it does, `bind()` subscribes to the limit's `CHANGED` signal so the projected crossing time updates automatically whenever the limit changes, the same way it already responds to the watched state changing. This removes the need for `rearm()`, `set_limit()`, `limit_attr`, and `fired_attr`, which only existed to manually compensate for the limit being static. In `tram_model.py`, the brake point becomes a plain `Observable` on the tram; `depart()` just sets it alongside `acceleration`, and the threshold does the rest.

codebreaker32:
Would appreciate what @quaquel think about APIs now? Following our discussion, `tram_model.py` no longer accesses the private descriptor nor method

mesa/experimental/states/state.py
```diff
@@ -0,0 +1,605 @@
...
+        self.fallback_value = fallback_value
+        self._rate_input = rate
+
+    def __set_name__(self, owner: type[HasEmitters], name: str) -> None:
+        """Bind the descriptor to the class and create shadow observables."""
+        super().__set_name__(owner, name)
+        self.private_state_name = f"_{name}_continuous_state"
+        self.rate_name = f"{name}_rate"
+
+        if self.rate_name in owner.__dict__:
```
quaquel: what about also checking `__slots__`

mesa/experimental/states/state.py
```diff
@@ -0,0 +1,605 @@
...
+                        for t_name in klass.__dict__["_continuous_thresholds"]:
+                            thresh = getattr(cls, t_name, None)
+                            if thresh is not None and hasattr(thresh, "bind"):
+                                thresh.bind(instance)
+
+            time = (
+                getattr(instance.model, "_time", 0.0)
+                if hasattr(instance, "model")
+                else 0.0
+            )
```
quaquel: have you had a chance to think about a peek style method as a more elegant solution for this? Alternatively, we might open an issue about it so at least we don't forget it.

mesa/experimental/states/tram_model.py
```diff
@@ -0,0 +1,186 @@
...
+            Chained off speed (position' = speed).
+        brake_point (Observable): Position at which braking begins for the
+            current segment, m. Updated at the start of each departure;
+            the _brake_point threshold reacts to changes automatically.
+    """
+
+    acceleration = Observable(fallback_value=0.0)
+    speed = ContinuousState(fallback_value=0.0, rate=lambda a: a.acceleration)
+    position = ContinuousState(fallback_value=0.0, rate=lambda a: a.speed)
+    brake_point = Observable(fallback_value=float("inf"))
```
quaquel: can accelerate and brake_point be normal attributes or must they be Observable

mesa/experimental/states/state.py
codebreaker32: Good catch, I missed it.

mesa/experimental/states/state.py
codebreaker32: I have something in mind, I will open a PR pretty soon

mesa/experimental/states/tram_model.py
codebreaker32: `acceleration` needs to be `Observable` because `speed`'s rate lambda reads it. A plain attribute would be read as a normal Python attribute and never trigger `recalculate()` when it changes.

`brake_point` needs to be `Observable` for the same reason in this design, `Threshold` subscribes to its `CHANGED` signal. A plain attribute doesn't emit signals, so changing it wouldn't trigger `recalculate()`. If you wanted `brake_point` to be a plain attribute, you'd need a different mechanism to notify the threshold when it changes, which is exactly what we were trying to get rid of.

codebreaker32:
For the idea of making `peek()` an instance method, I have a quick architectural question: If we attach it to `HasEmitters`, how do we safely read `model.time`? Since a standard Model doesn't inherit `HasEmitters`, wouldn't calling `instance.model.peek('time')` throw an `AttributeError`? Because of that, should we keep it as a standalone `peek(obj, name)` utility so it works on any object, or is there a way to ensure the model always has access to it?

quaquel:
Unless I am mistaken, Model subclasses `HasEmitters`, so if you have a `peek` on `HasEmitters`, analogues to `observe`, it should work: `model.peek("time")` would be the resulting call. Which is slightly different from `model.time.peek` which would be virtually impossible to make work.

codebreaker32:
> Unless I am mistaken, Model subclasses `HasEmitters`

My bad, yes its working fine.

jackiekazil:
I did a review the PR. I like the PR, and the example is good clean illustration. 

Some questions... 

- Have all of @quaquel's comments been resolved? (It is uncertain to me.) 
- Is there a reason the tests haven't been fixed? Looks like it maybe the result of Threshold.set_limit().

mesa/experimental/states/state.py
```diff
@@ -0,0 +1,605 @@
...
+                ObservableSignals.CHANGED,
+                old=old_value,
+                new=float(value),
+            )
+            core.PROCESSING_SIGNALS.discard(
+                _hashable_signal(instance, self.public_name)
+            )
+
+
+# class ContinuousScheduler:
```
jackiekazil: Is there a reason why this is commented out instead of deleted?

mesa/experimental/states/state.py
```diff
@@ -0,0 +1,605 @@
...
+    def execute(self, instance: Agent) -> None:
+        """Fire the callback when a crossing is reached.
+
+        Args:
+            instance (Agent): The instance whose threshold was triggered.
+        """
+        setattr(instance, self.time_attr, math.inf)
+        setattr(instance, self.event_attr, None)
+
+        callback_method = getattr(instance, self.callback)
```
jackiekazil, line 491: If I understand this correctly the callback_method() can call agent.remove(), but what happens to pending threshold events in the queue associated with that after death? Are those cleared out?  If not, they could fire on a removed agent, potentially triggering callbacks that corrupt model state (e.g. scheduling new events, incrementing indices). Would it be worth explicitly clearing out all sibling threshold events here before calling the callback?

mesa/experimental/states/state.py
```diff
@@ -0,0 +1,605 @@
...
+                    return descriptor.get_rate(parent)
+
+        return 0.0
+
+    def _extrapolate(self, state: dict[str, Any], dt: float) -> float:
+        """Project the value forward by dt using the cached base and rates."""
+        return (
+            state["base_value"]
+            + state["current_rate"] * dt
+            + 0.5 * state["second_order_rate"] * dt * dt
```
jackiekazil, line 157: What happens when there is an additional level to the chain? Would this fail silently?

mesa/experimental/states/state.py
codebreaker32: I wanted to keep it until I get the final picture. Its safe to delete now

mesa/experimental/states/state.py
codebreaker32: We actually get this safety for free because of the native `weakref` architecture in `events.py`.When `Event` schedules a callback, it explicitly wraps it in a weak reference to prevent memory leaks. During `Threshold.bind()`, we bind the `_trigger` function directly to the agent instance. If `callback_method()` calls `agent.remove()` and the agent is subsequently garbage collected, its `_trigger` functions are destroyed with it. When a pending sibling event eventually pops from the queue, it detects that the `weakref` evaluates to `None` and silently drops the event, exactly as designed. We can't safely clear sibling events inside `execute()` because `execute()` doesn't know if the callback resulted in death (doing so unconditionally would break sequential thresholds, like a tram hitting a brake point).

codebreaker32: You might also want to see https://github.com/mesa/mesa/pull/3755#issuecomment-4761473901

mesa/experimental/states/state.py
codebreaker32: Yes. The solution to this is an `ODEState`(yet to be implemented) but pitched in https://github.com/mesa/mesa/discussions/3721#discussioncomment-17299785

codebreaker32:
> * Have all of @quaquel's comments been resolved? (It is uncertain to me.)

So far, yes.

> * Is there a reason the tests haven't been fixed? Looks like it maybe the result of Threshold.set_limit().

I was waiting for `peek` PR to be merged then Ig it slipped my mind. Fixed now.

mesa/experimental/states/state.py
jackiekazil: thank you

codebreaker32:
I have fixed the ruff checks, Now the next thing is to decide upon tram_model example
1. Should I include it in this PR, if yes? where should I put it?
2. Or should I open a separate PR for it?

jackiekazil (dismissed):
LGTM.... everything discussed was covered. All related tests are passing. There are tests unrelated to this that are failing due to upstream dependencies, but that shouldn't hold this up.

jackiekazil:
> I have fixed the ruff checks, Now the next thing is to decide upon tram_model example
> 
> 1. Should I include it in this PR, if yes? where should I put it?
> 2. Or should I open a separate PR for it?

^^ I missed this comment. 

RE: tram_model -- yes or no? 
Looking at our examples in our examples folder -- I didn't think that we can cleanly extend one of our current examples. It sounds like we are in agreement here (/cc @quaquel, re: https://github.com/mesa/mesa/pull/3755#issuecomment-4791917101). 

RE: tram_model PR 
Let's remove it from this PR and put it into another PR, then we can merge this.

codebreaker32:
> RE: tram_model PR
> Let's remove it from this PR and put it into another PR, then we can merge this.

Done.

jackiekazil:
Was new PR for the example posted? Can you link it?

codebreaker32:
Tests and Description updated! Good to go from my end 👍

jackiekazil (approved):
LGTM (Looks good to me)

PR #3767, opened 2026-07-01
Support for running scenarios in parallel

quaquel:
This PR builds on #3756 by adding support for running scenarios with a parallel executors such as `ProcessPoolExecutor` and `MPIPoolExecutor`. 

The main things it adds are 
* Fills in the the branch if executor is not None
* Handles failures where the entire pool crashes, turning all scenarios that have not finished into a new state: ABORTED
* Various other minor changes as knock on effects from these two points.

I have tested this already successfully with a process pool, I will test this with MPI asap.


mesa/experimental/scenarios/runner.py
```diff
@@ -102,11 +107,164 @@ def extract_output(self, model: Model) -> dict[str, pd.DataFrame]:
...
+                        None,
+                        FailureInfo(
+                            origin=FailureOrigin.WRITING,
+                            exception_type=type(e).__name__,
+                            message=str(e),
+                            traceback="".join(traceback.format_exception(e)),
+                        ),
+                    )
+                _record(scenario, result)
+        except BrokenExecutor as e:
```
tpike3, line 258: I am pretty rusty on parallel executions, but don't you need to cancel all your futures here, as other runs may be still be failing on other processors

quaquel: A broken executor will result in the crash of the entire pool, so no need to cancel the futures. What we have to do, however, is mark all the outstanding scenarios as being aborted, which is exactly what this code does.

tpike3 (approved):
LGTM

PR #3769, opened 2026-07-03
Scenario persistence

quaquel:
As part of my ongoing work to replace the batch runner, this PR changes the way in which Scenario instances are pickled/unpickled and can be safely stored to disk. This PR is needed in preparation for storing parameter sweeps to disk. 

Some background: the original implementation relied on the initial state dict of the numpy bit generator. This works, but this state dict might not cleanly be saved in e.g., a JSON or CSV file because it relies on numpy dtypes. Due to numerical precision, integers might be lost in a round trip. This in turn will invalidate any attempt to exactly replicate a given run. 

After some more in depth digging, this PR implements a much cleaner approach to achieve the same end. It takes the state information from the underlying seed sequence and the class of the bit generator. The advantage of this is that the state information of the seed sequence are simple integers and thus will survive a round trip to e.g., a JSON or CSV file 

This PR branches from the parallel running branch (#3767) so the diff might seem larger than is actually the case. It also means that #3767 should be merged prior to this one.


tpike3 (approved):
I see no issues, LGTM

PR #3773, opened 2026-07-10
Fix HexGrid neighbor offset selection for even/odd rows

somtri:
### Pre-PR Checklist

### Summary
`HexGrid._connect_cells_2d` picks the wrong offset table for every row (even and odd are swapped), so `cell.connections` disagrees with the true hexagonal neighbors implied by the grid's own physical-position geometry. This affects every cell in every `HexGrid`.

### Bug / Issue
No existing issue for this one. I found it while reading through the discrete_space code.

Repro:

```python
from mesa.discrete_space import HexGrid
import random

grid = HexGrid((10, 10), torus=False, random=random.Random(42))
cell = grid._cells[(4, 4)]
print(sorted(c.coordinate for c in cell.connections.values()))
# currently: [(3, 4), (4, 3), (4, 5), (5, 3), (5, 4), (5, 5)]
# but cell.position (set in _init_hex_geometry, same file) puts the true
# nearest neighbors at [(3, 3), (3, 4), (3, 5), (4, 3), (4, 5), (5, 4)]
```

I checked this by computing the true nearest neighbors directly from the grid's own `x = sqrt(3)*(col + 0.5*(row%2))`, `y = 1.5*row` position formula, independent of `_connect_cells_2d`. Every cell in a 10x10 grid, torus and non-torus, disagrees with its geometrically-true neighbors under the current code.

### Implementation
One-line fix in `_connect_cells_2d`: the ternary had `odd_offsets` and `even_offsets` backwards relative to row parity.

`test_hexgrid` and the HexGrid block in `test_cell_neighborhood` asserted the buggy neighbor coordinates and counts as correct, so I fixed those expected values against the same independent position-based geometry rather than against the fixed code's own output. I also added `test_hexgrid_connectivity_matches_geometry`, which checks every cell's connections against its nearest neighbors by physical position, for both torus and non-torus grids, so this exact bug can't come back unnoticed.

### Testing
- `pytest --cov=mesa tests/`: 682 passed (681 baseline + 1 new test), 86% coverage, no regression from main
- `ruff format --diff` / `ruff check`: clean
- Confirmed the new regression test actually catches the bug: reverted the one-line fix temporarily, confirmed `test_hexgrid_connectivity_matches_geometry` fails against the unfixed code, then re-applied the fix

### Additional Notes
None.

### DIFF
--- a/mesa/discrete_space/grid.py
+++ b/mesa/discrete_space/grid.py
@@ -589,7 +589,7 @@ class HexGrid(Grid[T]):
 
         for cell in self.all_cells:
             i = cell.coordinate[1]
-            offsets = even_offsets if i % 2 else odd_offsets
+            offsets = odd_offsets if i % 2 else even_offsets
             self._connect_single_cell_2d(cell, offsets=offsets)
 
     def _connect_cells_nd(self) -> None:
--- a/tests/discrete_space/test_discrete_space.py
+++ b/tests/discrete_space/test_discrete_space.py
@@ -406,7 +406,7 @@ def test_cell_neighborhood():
     grid = HexGrid(
         (width, height), torus=False, capacity=None, random=random.Random(42)
     )
-    for radius, n in zip(range(1, 4), [3, 7, 13]):
+    for radius, n in zip(range(1, 4), [2, 6, 11]):
         if radius == 1:
             neighborhood = grid._cells[(0, 0)].neighborhood
         else:
@@ -418,7 +418,7 @@ def test_cell_neighborhood():
     grid = HexGrid(
         (width, height), torus=False, capacity=None, random=random.Random(42)
     )
-    for radius, n in zip(range(1, 4), [4, 10, 17]):
+    for radius, n in zip(range(1, 4), [4, 9, 15]):
         if radius == 1:
             neighborhood = grid._cells[(1, 0)].neighborhood
         else:
@@ -437,15 +437,15 @@ def test_hexgrid():
     assert len(grid._cells) == width * height
 
     # first row
-    assert len(grid._cells[(0, 0)].connections.values()) == 3
+    assert len(grid._cells[(0, 0)].connections.values()) == 2
     for connection in grid._cells[(0, 0)].connections.values():
-        assert connection.coordinate in {(0, 1), (1, 0), (1, 1)}
+        assert connection.coordinate in {(0, 1), (1, 0)}
 
     # second row
     assert len(grid._cells[(1, 0)].connections.values()) == 4
     for connection in grid._cells[(1, 0)].connections.values():
         # fmt: off
-        assert connection.coordinate in {   (1, 1), (2, 1),
+        assert connection.coordinate in {   (0, 1), (1, 1),
                                          (0, 0),    (2, 0),}
         # fmt: on
 
@@ -453,9 +453,9 @@ def test_hexgrid():
     assert len(grid._cells[(5, 5)].connections.values()) == 6
     for connection in grid._cells[(5, 5)].connections.values():
         # fmt: off
-        assert connection.coordinate in {  (4, 4), (5, 4),
+        assert connection.coordinate in {  (5, 4), (6, 4),
                                          (4, 5), (6, 5),
-                                         (4, 6), (5, 6)}
+                                         (5, 6), (6, 6)}
 
         # fmt: on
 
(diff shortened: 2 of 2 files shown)


quaquel:
This code has been extensively tested and confirmed visually as well. I'll need more evidence before considering this PR.

somtri:
Makes sense, I'd want more evidence too. Here's what I checked.

I ran a brute-force comparison on a 10x10 non-torus HexGrid: for each interior cell (away from the boundary, where edge effects don't complicate things), I compute its true nearest neighbors directly from `cell.position` (Euclidean distance, independent of `.connections`), and compare that to what `.connections` actually returns.

On `main`: 64 out of 64 interior cells disagree with their true geometric neighbors.
With this PR's one-line fix: 0 out of 64 disagree.

Script (runnable as-is against either commit):

```python
import numpy as np
import random
from mesa.discrete_space import HexGrid

def true_neighbors(grid, coordinate, eps=1e-6):
    center = grid._cells[coordinate].position
    dists = [
        (round(np.linalg.norm(cell.position - center), 6), coord)
        for coord, cell in grid._cells.items()
        if coord != coordinate
    ]
    dmin = min(dists)[0]
    return {coord for d, coord in dists if abs(d - dmin) < eps}

grid = HexGrid((10, 10), torus=False, random=random.Random(42))
mismatches = sum(
    1 for coord, cell in grid._cells.items()
    if 1 <= coord[0] <= 8 and 1 <= coord[1] <= 8
    and {c.coordinate for c in cell.connections.values()} != true_neighbors(grid, coord)
)
print(mismatches, "/ 64 interior cells mismatched")
```

There's also `test_hex_grid_lookups` in `test_spatial_lookups.py`, untouched by this PR and already on main. It labels this grid's layout as "Odd-R Pointy-Topped" and asserts the exact position formula this fix relies on (e.g. `cell_1_0.position == [sqrt(3), 0.0]`). So the geometry I'm checking against isn't something I invented for this PR, it's the same geometry the codebase already commits to via `find_nearest_cell`.

One thing that might explain the "confirmed visually" result: `HexSpaceDrawer` builds the rendered hex tiling from cell positions, not from `.connections`, and both the correct and the buggy offset tables have 6 entries per interior cell. So a rendered grid looks like a proper hexagonal tiling either way, and a neighbor-count check passes either way too. The parity swap only shows up if you check which specific cells are connected against absolute coordinates, not whether it looks hexagonal or each cell has 6 neighbors.

Did the visual check compare specific neighbor coordinates against expected values, or mainly whether the tiling looked right and each cell had 6 neighbors? That distinction is exactly where this bug hides.

quaquel:
But why is the euclidian true nearest neighbor relevant. Hex spaces are a logical tiling, not a spatial one. 

When I implemented this, I ran extensive testing with agents and neighbors and agents hoping from cell to cell. This was assessed visually and by checking coordinates.

somtri:
Fair question, and I should correct something before answering it: this PR didn't just change `_connect_cells_2d`, it also rewrote the pinned coordinate assertions in `test_hexgrid` and `test_cell_neighborhood` (disclosed in the PR description, but it's the crux here so worth restating plainly). Before this PR, `test_hexgrid` asserted `(0,0)` has 3 connections including `(1,1)`; this PR changed that to 2, excluding `(1,1)`. So there was a coordinate-level check on main already. It just had different expected values than what this PR asserts.

Here's the one thing I think I can actually prove, independent of whether "spatial" is the right lens at all: two tests that already existed before this PR disagree with each other.

- `test_hex_grid_lookups` (untouched by this PR) asserts exact `cell.position` values and that `find_nearest_cell(cell.position) == cell` for every cell. Under those positions, `(0,0)` sits at `[0, 0]` and `(1,1)` sits at `[2.598, 1.5]`, a distance of 3.0. `(1,0)` and `(0,1)` are each about 1.73 (sqrt(3)) from `(0,0)`.
- The old, pre-PR `test_hexgrid` asserted `(0,0)` connects to `(1,1)`, a cell roughly twice as far away as the two actual nearest cells.

Both can't be right if "connection" means "nearest neighbor by position." That's a contradiction between two pre-existing tests, not an assumption I'm importing from outside the class.

I think you're making a different, legitimate point underneath this, though: maybe connection identity is the canonical definition for HexGrid, and position/`find_nearest_cell` is what should bend to match it, not the other way around. That's a real design call, and it's yours to make, not mine.

So, directly: should `HexGrid.connections` be defined by the offset-table logic on its own terms, independent of position? Or should it be defined to match nearest-neighbor-by-position, the way `_init_hex_geometry`/`find_nearest_cell` currently compute it? Whichever you pick, I'll close this out to match, including reverting the test changes if connectivity turns out to be what's actually canonical here.

quaquel:
I am fine with AI use, but please acknowledge if you do so and don't blindly copy paste responses. I am closing this PR. I need solid evidence of a bug not some AI generated stuff.

PR #3774, opened 2026-07-15
Add `peek()` to `HasEmitters` for non-reactive attribute access

codebreaker32:
### Summary

Adds a `peek` method to the `HasEmitters`. This allows developers to retrieve the value of an `Observable` without registering it in the reactive dependency graph.

### Motive

In the `mesa_signals` architecture, reading an `Observable` from within a `ComputedState` automatically logs it as a dependency. For components like we discussed in #3766 , we frequently need to read a state (e.g., `model.time`) without permanently binding to it or triggering a reactive signal cascade every time the clock ticks.

### Implementation

* Implemented `.peek()` directly on the `HasEmitters` mixin.
* The method temporarily sets `core.CURRENT_COMPUTED` to `None`, extracts the value via `getattr`, and securely restores the previous tracker state in a `finally` block.

### Usage Examples

```python
# Standard access
current_time = instance.model.time

# Peek access 
current_time = instance.model.peek("time", 0.0) 

```

### DIFF
--- a/mesa/experimental/mesa_signals/core.py
+++ b/mesa/experimental/mesa_signals/core.py
@@ -702,6 +702,17 @@ class HasEmitters:
             for st in current_signals:
                 subs_dict[(name, st)].append(ref)
 
+    def peek(self, name: str, default: Any = None) -> Any:
+        """Retrieve an attribute without registering it in the dependency graph."""
+        global CURRENT_COMPUTED  # noqa: PLW0603
+        prev = CURRENT_COMPUTED
+        CURRENT_COMPUTED = None
+
+        try:
+            return getattr(self, name, default)
+        finally:
+            CURRENT_COMPUTED = prev
+
 
 def descriptor_generator(
     cls,
--- a/tests/experimental/test_mesa_signals.py
+++ b/tests/experimental/test_mesa_signals.py
@@ -1037,3 +1037,66 @@ def test_clear_all_class_subscriptions():
     DummyAgent.clear_all_class_subscriptions("state")
     agent1.state = "inactive"
     assert len(handler_calls) == 1
+
+
+def test_peek():
+    """Test that peek safely retrieves Observables, computed properties, and standard attributes."""
+
+    class PeekAgent(Agent, HasEmitters):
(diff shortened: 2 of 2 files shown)


mesa/experimental/mesa_signals/core.py
```diff
@@ -702,6 +702,17 @@ def _register_observer(
...
                 subs_dict[(name, st)].append(ref)
 
+    def peek(self, name: str, default: Any = None) -> Any:
+        """Retrieve an attribute without registering it in the dependency graph."""
+        global CURRENT_COMPUTED  # noqa: PLW0603
+        prev = CURRENT_COMPUTED
+        CURRENT_COMPUTED = None
+
+        try:
+            return getattr(self, name, default)
```
quaquel, line 712: Just curious, but why don't you simply access the private attribute so you don't have to mess with the current computed stuff?

Looking at `BaseObservable`, the private name is always the public name with a leading underscore. This is also how you were accessing time. 

Separate issue: I assume you'll add tests? Once those are in, this can be merged.

codebreaker32: That was actually my first thought too But I realized that directly accessing `_{name}` is a bit brittle after running some smoke test.
First, it breaks on `@computed_property`, which stores its internal state under `_computed_{name}` rather than `_{name}`. 
Even if you correctly guessed the name _computed_speed, grabbing the private variable directly bypasses the reactive engine entirely. When you access a `@computed_property` via its public name, the framework's descriptor checks if `state.is_dirty` or if its parents have changed. If dependencies have changed, it explicitly calls `state.evaluate()` to calculate the fresh value before returning it.  If you access the private variable directly, you are just grabbing the internal `ComputedState` object, and its cached `.value` might be entirely stale. By using `.peek("speed")`, you route the request through the public descriptor, which natively executes the `is_dirty` evaluation logic, while still safely blinding the `CURRENT_COMPUTED` tracker

codebreaker32: Done with tests!

quaquel (approved):

PR #3775, opened 2026-07-16
Fix FixedAgent.remove() when agent has no cell

### LINKED ISSUE #3748: FixedAgent.remove() raises AttributeError when the agent has no cell
**Describe the bug**

```python
    def remove(self):
        super().remove()
        self._mesa_cell = None
```

**Expected behavior**

**To Reproduce**

```python
from mesa import Model

model = Model(rng=1)
```

Or via teardown:

```python
from mesa import Model

model = Model(rng=1)
```

```python
```

**Additional context**

```python
def remove(self):
    super().remove()
        self._mesa_cell = None
```

extreme-coder:
### Pre-PR Checklist

### Summary
`FixedAgent.remove()` crashed with `AttributeError` when called on an agent that was never placed on a cell. Because `Model.remove_all_agents()` calls `remove()` on every agent, a single unplaced `FixedAgent` made the whole model impossible to tear down

### Bug / Issue

Fixes #3748.

`FixedAgent.remove()` called `self.cell.remove_agent(self)` unconditionally. When the agent had never been assigned a cell, `self.cell` is `None`, so this raised `AttributeError: 'NoneType' object has no attribute 'remove_agent'`.

Expected: removing a `FixedAgent` with no cell should succeed as a no-op for the cell part, the same way `CellAgent.remove()` already behaves. This also is an internal inconsistency between the two agent types.

Reproduction:

from mesa import Model

model = Model(rng=1)
agent = FixedAgent(model)   # created but never placed on a cell
agent.remove()              # AttributeError before this fix

### Implementation

Guarded the cell removal in FixedAgent.remove() so it only runs when a cell is actually assigned:

```python
def remove(self):
    super().remove()
    # bypasses the FixedCell setter, which rejects reassignment
    if self.cell is not None:
        self.cell.remove_agent(self)
    self._mesa_cell = None
```

CellAgent.remove() solves the same problem by assigning self.cell = None, letting the HasCell setter no-op when the cell is already None. FixedAgent cannot reuse that path: the FixedCell setter raises ValueError("Cannot move agent in FixedCell") on any reassignment, which is why FixedAgent assigns _mesa_cell directly. Hence the explicit None guard rather than routing through the setter.

### Testing
Added two regression tests in tests/discrete_space/test_discrete_space.py:
- test_fixed_agent_removal_without_cell; remove() on an agent that was never placed
- test_remove_all_agents_with_unplaced_fixed_agent; the remove_all_agents() teardown path, with a mix of placed and unplaced fixed agents

Both tests fail before the change with the reported AttributeError and pass after. ruff check and ruff format --check are clean.

### Additional Notes
No API or signature changes

### DIFF
--- a/mesa/discrete_space/cell_agent.py
+++ b/mesa/discrete_space/cell_agent.py
@@ -115,7 +115,9 @@ class FixedAgent(Agent, FixedCell):
         """Remove the agent from the model."""
         super().remove()
 
-        self.cell.remove_agent(self)
+        # bypasses the FixedCell setter, which rejects reassignment
+        if self.cell is not None:
+            self.cell.remove_agent(self)
         self._mesa_cell = None
 
 
--- a/tests/discrete_space/test_discrete_space.py
+++ b/tests/discrete_space/test_discrete_space.py
@@ -1218,6 +1218,43 @@ def test_fixed_agent_removal_state():
     assert agent.cell is None
 
 
+def test_fixed_agent_removal_without_cell():
+    """Test that a FixedAgent that was never placed on a cell can be removed.
+
+    This is a regression test for issue #3748:
+    FixedAgent.remove() raises AttributeError when the agent has no cell.
+    """
+    model = Model()
+    agent = FixedAgent(model)
+
+    assert agent.cell is None
+
+    agent.remove()
+
+    assert agent.cell is None
+    assert agent not in model._all_agents
+
+
+def test_remove_all_agents_with_unplaced_fixed_agent():
+    """Test that a model with an unplaced FixedAgent can still be torn down.
+
+    This is a regression test for issue #3748: remove_all_agents() crashed on
+    any FixedAgent that was never assigned a cell.
+    """
+    model = Model()
+    cell = Cell((1,), capacity=None, random=random.Random())
+    placed_agent = FixedAgent(model)
+    placed_agent.cell = cell
+    unplaced_agent = FixedAgent(model)
+
+    model.remove_all_agents()
+
+    assert placed_agent not in cell.agents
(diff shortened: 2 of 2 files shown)


dietcoke-17 (commented):
First thing first as a opensource community you could have asked in the issue page that could you work on this issue or what are your thought regarding that issue and you should not have done PR sniping like this

tpike3:
@extreme-coder  Thank you for the PR. Please take a look at #3748 (issue). 

Our current view is that this is not a bug, but a way to help user understand if there are issues with their model. 

However, please feel free to weigh in on the discussion if your experience is different and may reveal a user dynamic we were not considering. 

Thank you again for the PR and your interest in contributing to Mesa,  as this is an active discussion in #3748. I am going to close this PR.

shipitdev:
> @extreme-coder Thank you for the PR. Please take a look at #3748 (issue).
> 
> Our current view is that this is not a bug, but a way to help user understand if there are issues with their model.
> 
> However, please feel free to weigh in on the discussion if your experience is different and may reveal a user dynamic we were not considering.
> 
> Thank you again for the PR and your interest in contributing to Mesa, as this is an active discussion in #3748. I am going to close this PR.

Following this with interest. One thought on the "helps users notice mistakes" right now that "help" is an AttributeError about NoneType, which doesn't actually tell a user what went wrong or how to fix it, they'd have to dig into Mesa's internals to connect it to "oh, I forgot to place this agent." i think if the goal is genuinely diagnostic, i'd a purpose-built exception (something like a clear "agent was never assigned a cell" message) serve that better than either silently no-opping or leaving the current AttributeError?

PR #3779, opened 2026-07-25
test: assert empties count after adding agents in test_empties_space

### LINKED ISSUE #3745: test_empties_space does not assert empties count after adding agents
## Description:
In tests/discrete_space/test_discrete_space.py, test_empties_space only checks that empties equals n on a fresh grid, but never verifies that empties is updated correctly after agents are added:
```python
assert len(grid.empties) == n  # checked

for i in range(8):
    grid._cells[i].add_agent(CellAgent(model))

# no assertion here, empties tracking after agent addition is never tested
```
This means the test would pass even if empties was completely broken after agent placement.
Expected behavior: After adding 8 agents to a 10-node network, len(grid.empties) should equal 2. This should be explicitly asserted.
Suggested fix:
```python
for i in range(8):
    grid._cells[i].add_agent(CellAgent(model))

assert len(grid.empties) == n - 8
```

lacebx:
The existing test_empties_space test only checked the initial empties count on a fresh grid, but never verified that empties updates correctly after agents are added to cells. This adds an explicit assertion that len(grid.empties) equals n minus 8 after adding 8 agents.

Closes #3745

### DIFF
--- a/tests/discrete_space/test_discrete_space.py
+++ b/tests/discrete_space/test_discrete_space.py
@@ -558,8 +558,7 @@ def test_empties_space():
     for i in range(8):
         grid._cells[i].add_agent(CellAgent(model))
 
-        assert len(grid.empties) == n - 8
-
+    assert len(grid.empties) == n - 8
 
 def test_cell_missing_exception():
     """Test that CellMissingException is raised when accessing non-existent cells."""
--- a/tests/discrete_space/test_discrete_space.py
+++ b/tests/discrete_space/test_discrete_space.py
@@ -558,6 +558,8 @@ def test_empties_space():
     for i in range(8):
         grid._cells[i].add_agent(CellAgent(model))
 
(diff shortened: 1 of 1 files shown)


tests/discrete_space/test_discrete_space.py
```diff
@@ -558,6 +558,8 @@ def test_empties_space():
     for i in range(8):
         grid._cells[i].add_agent(CellAgent(model))
 
+        assert len(grid.empties) == n - 8
```
falloficaruss:
```suggestion
    assert len(grid.empties) == n - 8
```

falloficaruss (commented):
Thank you for the PR!

The assertion is indented inside the loop which is causing the test failures. Moving it out would make this PR ready to merge.

quaquel:
superseded by #3793

lacebx (commented):
really was that simple. Thanks

PR #3780, opened 2026-07-29
DOC: warn about tuple-length ambiguity in Agent.create_agents

shipitdev:
create_agents silently splits any list/tuple/ndarray/Series argument whose length equals n across agents one-per-agent, even when theargument was intended as a single shared value (e.g. a coordinate tuple). Confirmed directly:

```python
    >>> MyAgent.create_agents(model, 2, pos=(10, 20))
    Agent 1: pos = 10
    Agent 2: pos = 20
``` 

instead of both agents sharing pos=(10, 20). This doesn't fix the underlying ambiguity (that's a separate, larger discussion, i'm going to open an issue about it and reference it here but documents the footgun and the workaround for now. Pure docstring change, no behavior change.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -127,7 +127,7 @@ class Agent[M: Model]:
 
         Returns:
             AgentSet containing the agents created.
- 
+
         Warning:
             A list, tuple, ndarray, or pandas Series argument whose length
             equals n is always treated as one value per agent, even if you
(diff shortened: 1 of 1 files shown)


tpike3 (approved):
Thanks @shipitdev --- adding the Warning is a nice and easy PR to integrate approach that immediately lets users know of this subtlety

PR #3785, opened 2026-08-05
Add examples/experimental group and move alliance_formation into it

codebreaker32:
### Summary
Adds a third example group, `mesa/examples/experimental/`, alongside `basic/` and
`advanced/`, and moves the entire `alliance_formation` example into it.

### Motive
`alliance_formation` is built on `mesa.experimental.meta_agents`. Everything under
`mesa/experimental/` explicitly carries no semver guarantees and may change without
notice, but the example demonstrating it lived in `examples/advanced/`, which reads
as core, maintained, stable-API territory.

That mismatch is misleading in both directions: users browsing the advanced examples
have no signal that this one rests on an unstable API, and the example is liable to
break on any `meta_agents` change, which makes it look like a regression in a core
example rather than expected churn in an experimental one.

Rather than special-case a single example, this splits the examples into three tiers
so the directory a user is browsing tells them what stability to expect.

This also gives future experimental-API examples an obvious home instead of accreting
into `advanced/` or `basic/`.

**No public API change.**

### Usage Examples

```python
# unchanged, the public export is the same
from mesa.examples import MultiLevelAllianceModel

# deep import path, if you need the scenario or app
from mesa.examples.experimental.alliance_formation.model import AllianceScenario

model = MultiLevelAllianceModel(scenario=AllianceScenario(n=50, rng=42))
model.run_model()


codebreaker32:
Hi @tpike3, I went ahead and implemented the refactor as we discussed. Would appreciate if you or @jackiekazil could give a quick review

tpike3 (approved):
Looks great thanks!

PR #3787, opened 2026-08-06
Remove spurious 'List' entry from find_combinations docstring

AshNicolus:
`find_combinations` documents an argument `List` in its `Args:` section that isn't a parameter of the function:

    Args:
        model: ...
        group: ...
        size: ...
        evaluation_func: ...
        filter_func: ...
        List: The function to filter combinations. Defaults to None.   <- not a parameter

It looks like a leftover, and its description just duplicates `filter_func`. This removes the stray line so the documented arguments match the actual signature (`model`, `group`, `size`, `evaluation_func`, `filter_func`).

Found with a quick docstring-vs-signature check; this was the only such mismatch in the codebase. Mesa 4.0.0a0 / current `main`.


tpike3 (approved):
LGTM -- Thanks @AshNicolus

PR #3789, opened 2026-08-07
added missing test for some lines in agent.py

dietcoke-17:
during the pr #3759 found some missing line in agent.py has missing tests for some lines given by codecov report locally so adding those tests 
### Changes
- **`test_agent_remove_cancels_current_action_silently`**: covers `remove()`'s silent-cancellation branch — an in-progress action is cancelled without firing `on_interrupt`/`on_complete`, since the agent is being destroyed, not making a behavioral choice.
- **`test_agent_remove_cleans_up_datasets`**: covers `remove()`'s dataset-cleanup loop — the agent is removed from every dataset registry entry it belongs to.
 These last two were previously untested — `test_agent_removal` only ever exercised removing an idle agent with no dataset membership.


dietcoke-17:
Pushed a follow-up commit — codecov/patch was failing at 95.24% because test_agent_remove_cancels_current_action_silently explicitly asserts on_complete/on_interrupt are not called (that's the point of the test), so those two callback bodies in _LongAction were never executed by anything else. Added two more tests exercising the normal paths instead (start_action() completing instantly, cancel_action() interrupting), which should bring patch back to the targeted percentage.

quaquel:
@codebreaker32, does this interact with any work you have been doing on the action framework? And if so, can this be merged or should it be closed in favor of work you are doing?

tests/test_agent.py
```diff
@@ -7,6 +7,7 @@
 import pytest
 
 from mesa.agent import Agent
+from mesa.experimental.actions import Action
```
codebreaker32: `tests/test_agent.py` is a core module and `mesa.experimental.actions` is in experimental section that can change without notice so better not to couple them, and we already have a test suite for actions in tests/experimental

tests/test_agent.py
```diff
@@ -308,3 +324,91 @@ def __init__(self, model):
     assert "count=42" in r
     assert "status=None" in r
     assert "items=[1, 2, 3]" in r
+
+
+def test_agent_remove_cancels_current_action_silently():
+    """Test that remove() cancels an in-progress action without firing on_interrupt."""
```
codebreaker32: Already covered under https://github.com/mesa/mesa/blob/27ef2e7f9af04389cf1123ebcc7d4691844eb5e6/tests/experimental/test_actions.py#L674-L675

tests/test_agent.py
```diff
@@ -308,3 +324,91 @@ def __init__(self, model):
...
+    # Silent cleanup: neither callback should have fired.
+    assert action.interrupted_progress == []
+    assert not action.completed
+
+
+def test_agent_remove_cleans_up_datasets():
+    """Test that remove() removes the agent from every dataset it belongs to."""
+
+    class _DatasetRegistryDouble:
+        def __init__(self):
```
codebreaker32: Why are you not using Dataregistry and instead a stub?

tests/test_agent.py
```diff
@@ -308,3 +324,91 @@ def __init__(self, model):
...
+    agent = DatasetAgent(model)
+    assert registry.added == [agent]
+
+    agent.remove()
+
+    assert registry.removed == [agent]
+
+
+def test_agent_start_action_completes_instantly_with_zero_duration():
+    """Test that a zero-duration action completes synchronously in start_action(), firing on_complete."""
```
codebreaker32: Already covered under https://github.com/mesa/mesa/blob/27ef2e7f9af04389cf1123ebcc7d4691844eb5e6/tests/experimental/test_actions.py#L111

tests/test_agent.py
```diff
@@ -308,3 +324,91 @@ def __init__(self, model):
...
+
+    agent.start_action(action)
+
+    assert action.completed
+    assert agent.current_action is None
+    assert not agent.is_busy
+
+
+def test_agent_cancel_action_fires_on_interrupt():
+    """Test that cancel_action() cancels the current action and fires on_interrupt."""
```
codebreaker32: Already covered in https://github.com/mesa/mesa/blob/27ef2e7f9af04389cf1123ebcc7d4691844eb5e6/tests/experimental/test_actions.py#L184

tests/test_agent.py
```diff
@@ -308,3 +324,91 @@ def __init__(self, model):
...
+
+    result = agent.cancel_action()
+
+    assert result is True
+    assert agent.current_action is None
+    assert not agent.is_busy
+    assert action.interrupted_progress != []
+
+
+def test_agent_advance_is_noop():
```
codebreaker32, line 333: Would appreciate its mention in description and we can keep it

codebreaker32 (changes requested):
No conflict so far with my ongoing work but I want to be a bit cautious as if I change the behavior, I will have to track it in different files

tests/test_agent.py
dietcoke-17: Replaced the Dataregistrydouble and stub in test_agent_remove_cleans_up_datasets with the real dataregistry/numpyagentdataset via model.data_registry.track_agents_numpy

dietcoke-17:
1.Removed the mesa.experimental.actions import and _LongAction helper — tests/test_agent.py is core and shouldn't couple to the experimental actions module.
2.Deleted the three duplicate tests (test_agent_remove_cancels_current_action_silently, test_agent_start_action_completes_instantly_with_zero_duration, test_agent_cancel_action_fires_on_interrupt) — verified each is already covered in tests/experimental/test_actions.py (test_remove_cancels_action_silently, test_instantaneous_action, test_cancel_ignores_interruptible_flag), exactly as the reviewer pointed out.
3.Replaced the _DatasetRegistryDouble stub in test_agent_remove_cleans_up_datasets with the real DataRegistry/NumpyAgentDataSet via model.data_registry.track_agents_numpy(...). This is safe to use directly (no new import needed) because mesa/model.py already hardwires Model.data_registry to a real DataRegistry unconditionally — Agent.remove()'s dataset cleanup path is inherently coupled to that class, so testing against the stub was hiding a real integration bug surface. The test now checks dataset.active_agents/len(dataset) (public API) before and after agent.remove().
4.Kept test_agent_advance_is_noop as requested — it doesn't touch Action at all so no changes were needed there.

quaquel (approved):

PR #3794, opened 2026-08-10
Fix FixedAgent.remove() when agent has no cell (#3748)

### LINKED ISSUE #3748: FixedAgent.remove() raises AttributeError when the agent has no cell

Dilraj07:
### Pre-PR Checklist

### Summary
Fixes `FixedAgent.remove()` raising `AttributeError` when called on an unplaced `FixedAgent` (`cell is None`).

### Bug / Issue
Fixes #3748.

Instantiating an unplaced `FixedAgent` (`agent = FixedAgent(model)`) and calling `agent.remove()` or `model.remove_all_agents()` raised `AttributeError: 'NoneType' object has no attribute 'remove_agent'`.

### Implementation
Guarded cell removal in `FixedAgent.remove()` (`mesa/discrete_space/cell_agent.py`) with `if self._mesa_cell is not None:`, matching `CellAgent.remove()` behavior.

### Testing
Added two regression tests in `tests/discrete_space/test_discrete_space.py` verifying `FixedAgent.remove()` and `model.remove_all_agents()` with unplaced agents.

Passed `pytest` and `ruff check`.

### DIFF
--- a/mesa/discrete_space/cell_agent.py
+++ b/mesa/discrete_space/cell_agent.py
@@ -115,8 +115,9 @@ class FixedAgent(Agent, FixedCell):
         """Remove the agent from the model."""
         super().remove()
 
-        self.cell.remove_agent(self)
-        self._mesa_cell = None
+        if self._mesa_cell is not None:
+            self._mesa_cell.remove_agent(self)
+            self._mesa_cell = None
 
 
 class Grid2DMovingAgent(CellAgent):
--- a/tests/discrete_space/test_discrete_space.py
+++ b/tests/discrete_space/test_discrete_space.py
@@ -1218,6 +1218,36 @@ def test_fixed_agent_removal_state():
     assert agent.cell is None
 
 
+def test_fixed_agent_removal_without_cell():
+    """Test that removing an unplaced FixedAgent succeeds without error."""
+    model = Model()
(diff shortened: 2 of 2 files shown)


AshNicolus:
LGTM. The guard is the right fix here: FixedAgent can't delegate removal to the cell setter the way CellAgent does, since FixedCell.cell raises on any reassignment (and would hit None.add_agent when unplaced), so clearing _mesa_cell directly behind an is-not-None guard is the correct approach. Tests cover both the unplaced single remove and the mixed placed/unplaced remove_all_agents path. Verified against the FixedCell setter and CellAgent.remove.

AshNicolus (approved):

quaquel:
Based on #3748 and and #3687, it seems we don't have alignment yet on the best way to address this issue. Therefore, I cannot accept this fix. Instead, I am closing it.

PR #3796, opened 2026-08-11
Add the tram route model to the experimental examples

codebreaker32:
Adds a Tram Route Model under `mesa/examples/experimental/`, as a worked demonstration of `mesa.experimental.states` — `ContinuousState` and `Threshold`( ref #3766 )

## What it models

A single tram runs an ordered route of stations. Leaving a stop it accelerates at a fixed rate, coasts once it reaches cruise speed, brakes at an analytically-computed point, arrives, dwells, and departs for the next stop.

## Why this example

Every transition is a `Threshold` crossing rather than a per-tick check:

```python
_cruise_threshold = Threshold(state=speed, limit=cruise_speed, callback="start_coasting", direction="rising")
_braking_threshold = Threshold(state=position, limit=brake_point, callback="brake", direction="rising")
_stop_threshold = Threshold(state=speed, limit=0.0, callback="arrive_at_station", direction="falling")
```

The two things it is meant to show:

- **Chained continuous state.** `speed' = acceleration` and `position' = speed`, so position extrapolates as a piecewise quadratic and is exact at any time between events — never a stale value waiting for the next tick.
- **Limits that are Observables, not constants.** `cruise_speed` and `brake_point` are `Observable`s, and `brake_point` is rewritten at the start of every segment. The thresholds observe those attributes and re-solve for the new crossing time themselves; nothing re-arms them.

The result is a model with no per-agent step logic. `TransitSystem.step` exists only to sample for the `DataCollector` behind the plots, delete it and the tram still runs the route with identical timings, driven entirely by the event queue.

The brake point is derived from the segment's *peak* speed, `min(cruise_speed, sqrt(2d / (1/a + 1/b)))`, not from cruise speed. On a short segment, or with weak acceleration, the tram never reaches cruise speed; assuming it does puts the brake point outside the segment, and a rising threshold that begins already past its limit never fires at all. This matters because the app exposes all four kinematic parameters as sliders.

## Visualization

Screenshot (327)
Screenshot (328)
Screenshot (329)

### DIFF
--- a/mesa/examples/README.md
+++ b/mesa/examples/README.md
@@ -49,3 +49,6 @@ The experimental examples are built on Mesa APIs that live under `mesa.experimen
 
 ### [Alliance Formation Model](examples/experimental/alliance_formation)
 Demonstrates dynamic meta-agent creation: agents form alliances through a game-theoretic process, and the resulting alliances become meta-agents that can themselves form higher-level alliances.
+
+### [Tram Route Model](examples/experimental/tram_model)
+Demonstrates continuous states and thresholds: a tram accelerates, coasts, brakes and dwells its way along a route, with every transition solved for analytically and scheduled on the event queue rather than polled for on each tick.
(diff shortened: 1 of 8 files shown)


codebreaker32:
The core logic resides in `model.py` and `agents.py` which is a roughly ~230 line diff

jackiekazil (approved):
This looks good. 
Thank you for breaking this into another PR & for adding the UI & the little tram icon is a nice touch.

jackiekazil:
Thank you @codebreaker32

PR #3799, opened 2026-08-14
feat(agentset): add select_random with uniform and weighted sampling …

### LINKED ISSUE #3746: Add random / weighted sampling to AgentSet

`AgentSet` has no built-in way to draw a random sample of agents, and no way at all to do a *weighted* selection.

- Uniform random subset: you currently have to chain `agents.shuffle().select(at_most=n)`, which is non-obvious (it works because `select` takes the first `n` after a shuffle).
- Weighted selection (pick agents with probability proportional to some attribute, e.g. fitness-proportional / roulette-wheel selection): there is no path through the AgentSet API at all. You have to drop down to raw primitives:

```python
agents = model.agents.to_list()
weights = [a.fitness for a in agents]
chosen = model.random.choices(agents, weights=weights, k=2)
```

Weighted and random selection are very common ABM operations (selection by fitness, preferential attachment, random matching, sampling a subset to act each step), so it would be useful to have a first-class, reproducible method for them.

A sampling method on `AgentSet`, for example:

```python
# uniform, without replacement (returns an AgentSet)
model.agents.select_random(3)

# weighted by an attribute
model.agents.select_random(3, weights="fitness")

# weighted by a callable
model.agents.select_random(3, weights=lambda a: a.fitness ** 2)

# with replacement
model.agents.select_random(3, replace=True)
```

It would use the model's seeded RNG (`self.random`) so results stay reproducible, and return an `AgentSet` (consistent with `select`/`shuffle`). `weights` could accept a sequence, an attribute name, or a callable.

Open design questions for maintainers:
- Method name: `select_random` vs `sample` (note `CellCollection` already has `select_random_agent`).
- Return type for `n == 1`: an `AgentSet` of one, or the agent itself.
- `weights` input format (sequence / attribute name / callable , or a subset of these).

Yes, happy to open a PR once the API (name, signature, return type) is agreed.

**Additional context**

Self-contained and additive (no change to existing behavior). It would build on `self.random` (`random.sample` for uniform without replacement, `random.choices` for weighted/with replacement), so it stays reproducible. Mesa 4.0.0a0 / current `main`.

Dilraj07:
## Summary
Closes #3746.

Adds a first-class `select_random()` method to `AbstractAgentSet` (and all inheriting `AgentSet` subclasses), enabling:
1. **$\mathcal{O}(k)$ Uniform Random Sampling:** Directly samples $k$ agents without the $O(N)$ overhead of full population shuffles (`shuffle().select()`).
2. **Weighted Selection:** Supports fitness-proportionate / roulette-wheel selection, softmax exploration, and weighted agent sampling.

---

## Changes Made
- **`mesa/agentset.py` (`AbstractAgentSet.select_random`)**:
  - Uniform without replacement: $\mathcal{O}(k)$ via `self.random.sample`.
  - Uniform & Weighted with replacement via `self.random.choices`.
  - Weighted without replacement via **Efraimidis & Spirakis (A-Res)** algorithm (`k_i = u_i^{1/w_i}` where $u_i \sim U(0, 1)$), preserving 100% reproducibility using `self.random`.
  - Flexible weights support: agent attribute name (`str`), callable/lambda (`Callable`), or numerical sequence (`Sequence`).
  - Supports discrete integer counts ($n \ge 1$) and fractional percentages ($n \in (0, 1]$).
  - Supports `inplace=True` and `inplace=False`.
- **`tests/test_agentset.py`**:
  - Added 10 unit tests covering uniform, fractional, with/without replacement, attribute/callable/sequence weights, inplace mutation, error edge cases, and a multi-generation Evolutionary ABM scenario.

---

### 1. Complexity Comparison ($k=5$ sampled)
| Population ($N$) | Sample Size ($k$) | `shuffle().select()` | `select_random()` | Speedup |
| :--- | :--- | :--- | :--- | :--- |
| **100** | 5 | 0.022 ms | 0.003 ms | **7.7x** |
| **1,000** | 5 | 0.302 ms | 0.005 ms | **64.8x** |
| **10,000** | 5 | 2.883 ms | 0.007 ms | **431.3x** |
| **100,000** | 5 | 38.884 ms | 0.007 ms | **5,537.4x** |

### 2. Verification Visuals
sampling_benchmark_and_distribution

image

---

### Context & motivation
In Agent-Based Modeling and evolutionary/probabilistic simulations, selecting agents proportional to fitness, traits, or connection weights is a fundamental operation. Previously, Mesa lacked weighted selection and required chaining `shuffle().select(at_most=k)` for uniform subsets, which forces an O(N) full shuffle and intermediate set allocation. This PR adds direct uniform and weighted sampling to `AbstractAgentSet`.

### What I learned
I deepened my understanding of Mesa's `AbstractAgentSet` design, weak reference handling in collections, and the Efraimidis-Spirakis algorithm for reproducible weighted reservoir sampling without replacement.

### Readiness checks
- [x] This PR addresses an agreed-upon problem (linked issue #3746)
- [x] I have performed a self-review of my code
- [x] Tests pass locally (`pytest tests/test_agentset.py`)

### DIFF
--- a/mesa/agentset.py
+++ b/mesa/agentset.py
@@ -24,6 +24,47 @@ if TYPE_CHECKING:
 _MISSING = object()
 
 
+def _resolve_weights(
+    agents: Sequence[Agent],
+    weights: str | Callable[[Agent], float] | Sequence[float],
+) -> list[float]:
+    """Extract and validate numerical weights for a sequence of agents.
+
+    Args:
+        agents: The sequence of agents to extract weights for.
+        weights: Agent attribute name (str), weight callable, or sequence of floats.
+
+    Returns:
+        list[float]: Validated list of non-negative weights with positive sum.
+
+    Raises:
+        TypeError: If weights is of an unsupported type.
+        ValueError: If sequence length does not match, if any weight is negative, or if total sum <= 0.
+    """
+    if isinstance(weights, str):
+        w = [getattr(agent, weights) for agent in agents]
+    elif callable(weights):
+        w = [weights(agent) for agent in agents]
+    elif isinstance(weights, Sequence):
+        if len(weights) != len(agents):
+            raise ValueError(
+                f"Length of weights ({len(weights)}) does not match AgentSet length ({len(agents)})"
+            )
+        w = list(weights)
+    else:
+        raise TypeError(
+            f"Unsupported weights type: {type(weights).__name__}. "
+            "Expected str, Callable, Sequence[float], or None."
+        )
+
+    if any(x < 0 for x in w):
+        raise ValueError("All weights must be non-negative.")
+    if sum(w) <= 0:
+        raise ValueError("Sum of weights must be strictly positive.")
+
+    return w
+
+
 class AbstractAgentSet[A: Agent](ABC, MutableSet[A]):
     """An abstract base collection class that represents an ordered set of agents within an agent-based model (ABM).
 
@@ -128,8 +169,8 @@ class AbstractAgentSet[A: Agent](ABC, MutableSet[A]):
 
         Args:
             n (int | float, optional): The number or fraction of agents to select. Defaults to 1.
-                - If an integer >= 1, at most that number of agents are selected.
-                - If a float between 0 and 1, at most that fraction of the agents are selected.
+                - If an integer >= 1, the number of agents to select.
+                - If a float in (0, 1], the fraction of agents to select.
             weights (str | Callable[[Agent], float] | Sequence[float] | None, optional): Weights for selection.
                 - If a str: uses the specified agent attribute name.
                 - If a Callable: calls the function on each agent to compute its weight.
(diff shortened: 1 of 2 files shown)


mesa/agentset.py
```diff
@@ -116,6 +116,95 @@ def agent_generator(
...
+            inplace (bool, optional): If True, modifies the current AbstractAgentSet; otherwise, returns a new AbstractAgentSet. Defaults to False.
+
+        Returns:
+            AbstractAgentSet: A new or updated AbstractAgentSet containing the sampled agents.
+
+        Raises:
+            ValueError: If weights are negative, total weight <= 0, or length of weights sequence does not match the AgentSet.
+            TypeError: If weights is of an unsupported type.
+        """
+        if len(self) == 0 or n <= 0:
```
quaquel: not sure about this.

I would argue that n <= 0 should raise a value error.

Not sure about the case of n being larger than len(self) and replacement being false.

mesa/agentset.py
```diff
@@ -116,6 +116,95 @@ def agent_generator(
...
+                raise TypeError(
+                    f"Unsupported weights type: {type(weights).__name__}. "
+                    "Expected str, Callable, Sequence[float], or None."
+                )
+
+            if any(x < 0 for x in w):
+                raise ValueError("All weights must be non-negative.")
+            if sum(w) <= 0:
+                raise ValueError("Sum of weights must be strictly positive.")
+
```
quaquel: I would consider moving this into a separate private helper method to keep the code simpler to read here.

quaquel (commented):
I added some comments. Also, what explains the clear performance difference?

mesa/agentset.py
Dilraj07: Thanks for pointing this out, I have updated the behavior to be consistent with standard Python sampling (`random.sample`, NumPy `np.random.choice`):
1. `n <= 0` (or fractional `n <= 0.0` / `n > 1.0`) now raises a `ValueError`.
2. `select_random` on an empty `AgentSet` raises a `ValueError`.
3. When `replace=False` and `n > len(self)`, it now raises a `ValueError` indicating that the sample size exceeds the population size, preventing silent errors in simulation workflows.

mesa/agentset.py
Dilraj07: Done! Extracted the weight resolution and validation logic into a private `_resolve_weights()` helper function to keep `select_random()` clean and readable.

Dilraj07:
> I added some comments. Also, what explains the clear performance difference?

Thanks for the review!

Regarding the performance difference:

**1. Microbenchmark (`select_random(k)` vs `shuffle().select(at_most=k)`):**
- `shuffle().select()` shuffles the entire population (`O(N)`) and allocates an intermediate `AgentSet` of size `N` before slicing `k` agents.
- `select_random(k)` uses Python's `random.sample()` under the hood, sampling only `k` agents directly in `O(k)` time without shuffling the rest or creating intermediate sets.
- Because of this, sampling `k=5` takes almost constant time (~0.007 ms) regardless of population size (100 vs 100,000 agents).

**2. CI Benchmark Bot:**
- The CI bot benchmarks standard example models (Schelling, Sugarscape, etc.), which don't call `select_random` yet. The small ±1–4% variations are just standard GitHub Actions runner noise.

quaquel:
Are there places in the examples were we could already use this code?

Dilraj07:
> Are there places in the examples were we could already use this code?

Most `mesa-examples` currently cache `.to_list()` outside their `step()` loops (e.g., `deffuant_weisbuch`). Since `select_random()` still uses `.to_list()` internally, replacing those with `select_random()` inside the loop would actually be slower by reallocating lists $N$ times. 

However, `select_random()` immediately improves two key areas:

**1. Fixing Deprecation Warnings (e.g., `bank_reserves`)**
Currently, `bank_reserves` uses `random.choice(self.cell.agents)`. Since `self.cell.agents` is an AgentSet, this triggers a sequence `__getitem__` deprecation warning. `select_random()` provides a clean, warning-free native alternative:
```python
neighbors = self.cell.agents.select(lambda a: a != self)
if len(neighbors) > 0:
    customer, = neighbors.select_random(1)
```

**2. Weighted Agent Sampling**
We lack evolutionary models in `mesa-examples` mostly because extracting agents and fitnesses to use `random.choices` was annoying. Now we can easily support these models directly via `self.agents.select_random(k, weights="fitness")`.

quaquel:
I meant the examples that ship with Mesa itself, but we can leave that for a future PR as well.

quaquel (approved):

Issue #3798, opened 2026-08-13
Tracking issue for Action preconditions, preemption and continuation


codebreaker32:
_This is a **tracking issue**, meant for short status updates rather than discussion. Rationale lives in [#3721](https://github.com/mesa/mesa/discussions/3721)._

Continues "Advanced Agent Behavior" of #3132, building on `Action` (#3461) and `ContinuousState` / `Threshold` (#3766).

An agent's brain wakes on exactly two triggers: its body became free (`on_idle`), or the world crossed a line it declared (`Threshold`, merged). Never a queue, never a poll. The engine never decides what an agent does next.

## 1. Event list - `mesa.time`
**Independent of the design below i.e from 2-5**. 

PR: #3800

Measured over 300 steps with 500 agents, the action layer triggers compaction once even at a 30% interrupt rate, while the `Threshold` reschedule pattern triggers it 150 times. The beneficiary is `mesa.experimental.states`, not the work in sections 2-5.

- [x] Wire up adaptive compaction. `compact()` was added in #3359 but nothing in the library calls it; a comment in `EventList.remove()` already promises it. A cancel-and-reschedule workload peaks at 239,731 heap entries, 99% tombstones. Wired up that peak drops to 4,000 and peak memory from 29.0 MB to 1.3 MB, with a 2-3x speedup depending on the machine.
- [x] Track the tombstone count. Needs a weak back-reference from `Event` to its list, since `Event.cancel()` is called directly on the event by `Action`, `Threshold` and `EventGenerator`, never through the list.
- [x] Make `__len__` O(1). It scans the whole heap today — roughly 30 µs at 1k events and 2 ms at 50k — and `is_empty()` inherits that.

## 2. Action primitives

PR: #3801 for the first two.

- [x] `requirements` on `Action`, checked at `start()` and completion only. Continuous revalidation would reintroduce activity scanning; mid-flight aborts belong to `Threshold` → `cancel_action()`.
- [x] `ActionState.FAILED` and `on_fail()`. A failed requirement skips `on_complete()`.
- [x] Give `priority` an effect (#3805). It is resolved at `start()` and read by nothing, so a priority-1 action currently preempts a priority-100 one.

## 3. Wake contract

- [ ] `Agent.on_idle(previous: Action | None)`.
[...]

PR #3800, opened 2026-08-14
Compact the event list when canceled events dominate

codebreaker32:
`EventList.compact()` was added in #3359 but nothing in the library ever called it, while a comment in `EventList.remove()` already said canceled events "may trigger adaptive compaction if they dominate the heap".

Cancellation uses the tombstone pattern to keep the heap invariant, so a model that repeatedly cancels and reschedules an event grows the heap without bound. A `Threshold` whose projected crossing time keeps moving is the clearest case: 2000 agents rescheduling over 800 steps peak at 239,731 heap entries, 99.2% of them tombstones.

Wire compaction up from `cancel()` rather than from `add_event` or `pop_event`. Cancelling is the only operation that raises the share of canceled events; pushing grows the heap and popping discards a tombstone, so both move the ratio the other way.

Tracking that share needs a back-reference, because `Event.cancel()` is called straight on the event by `Action`, `Threshold` and `EventGenerator` and never routes through the list. Events therefore hold a weak reference to their list, set in `add_event` and cleared on pop, since an event that has left the heap must not be counted against it.

The same count makes `__len__` O(1) instead of a scan of the whole heap, which took 34.7 us at 1k events and 2.6 ms at 50k. `is_empty()` and `peek_ahead()` both inherit that.

Measured on the cancel-and-reschedule workload:

```
 200 agents    0.12s -> 0.11s   peak heap  14,256 ->   399
1000 agents    1.66s -> 0.97s   peak heap  89,379 -> 2,001
2000 agents   15.37s -> 5.26s   peak heap 239,731 -> 4,000
```

Peak memory at 1000 agents drops from 29.0 MB to 1.3 MB, and `len()` from 2.6 ms to 0.12 us at 50k events.

Completes (1) from #3798

### DIFF
--- a/mesa/time/events.py
+++ b/mesa/time/events.py
@@ -8,6 +8,7 @@ simulation events in chronological order while respecting event priorities. Key 
 - Weak references to prevent memory leaks from canceled events
 - Efficient event insertion and removal using a heap queue
 - Support for event cancellation without breaking the heap structure
+- Adaptive compaction, so repeated cancellation does not grow the heap without bound
 
 The module contains three main components:
 - Priority: An enumeration defining event priority levels (HIGH, DEFAULT, LOW)
@@ -122,6 +123,9 @@ class Event:
         self.function_args = function_args if function_args else []
         self.function_kwargs = function_kwargs if function_kwargs else {}
 
+        # Set by EventList.add_event, so cancel() can update the list's count.
+        self._owner: ReferenceType[EventList] | None = None
+
     def execute(self) -> None:
         """Execute this event."""
         if not self._canceled:
@@ -131,10 +135,18 @@ class Event:
 
     def cancel(self) -> None:
         """Cancel this event."""
+        if self._canceled:
+            return
+
         self._canceled = True
         self.fn = None
         self.function_args = []
         self.function_kwargs = {}
+
+        if self._owner is not None:
+            owner = self._owner()
+            if owner is not None:
+                owner._note_cancellation()
 
     def __lt__(self, other: Event) -> bool:
         """Define a total ordering for events to be used by the heapq."""
@@ -151,12 +163,14 @@ class Event:
         fn = self.fn() if self.fn is not None else None
         state["_fn_strong"] = fn
         state["fn"] = None  # Don't pickle the weak reference
+        state["_owner"] = None  # Weak references cannot be pickled
         return state
 
     def __setstate__(self, state: dict[str, Any]) -> None:
         """Restore state after unpickling."""
         fn = state.pop("_fn_strong")
         self.__dict__.update(state)
+        self._owner = None
         # Recreate callable reference strategy.
         if fn is not None:
             self.fn = _create_callable_reference(fn)
@@ -402,13 +416,48 @@ class EventList:
     appropriate data structure. Events are sorted based on their time stamp, their priority, and their unique_id
     as a tie-breaker, guaranteeing a complete ordering.
 
+    Canceling an event leaves it on the heap so the heap invariant is not broken.
+    Models that repeatedly cancel and reschedule an event would therefore grow the
+    heap without bound, so the list counts the canceled events it holds and rebuilds
+    itself once they dominate. That count also makes len() an O(1) operation.
+
+    Attributes:
+        COMPACTION_RATIO (float): Fraction of canceled events above which the heap is rebuilt
+        COMPACTION_FLOOR (int): Minimum number of canceled events before compaction is considered
+
+    Notes:
+        An event belongs to one event list at a time. Adding the same event to a
+        second list hands ownership over, and the first list's count goes stale.
 
     """
+
+    COMPACTION_RATIO: float = 0.5
+    COMPACTION_FLOOR: int = 64
 
     def __init__(self):
         """Initialize an event list."""
         self._events: list[Event] = []
+        self._n_canceled: int = 0
+        # Built once and handed to every event, since creating it per push is costly.
+        self._self_ref: ReferenceType[EventList] = ref(self)
         heapify(self._events)
+
+    def __getstate__(self) -> dict[str, Any]:
+        """Prepare state for pickling."""
+        state = self.__dict__.copy()
+        state["_self_ref"] = None  # Weak references cannot be pickled
+        return state
+
+    def __setstate__(self, state: dict[str, Any]) -> None:
+        """Restore state after unpickling.
+
+        Events drop their weak reference to this list when pickled, so it has to
+        be re-established or later cancellations would go uncounted.
+        """
+        self.__dict__.update(state)
+        self._self_ref = ref(self)
+        for event in self._events:
+            event._owner = self._self_ref
 
     def add_event(self, event: Event):
         """Add the event to the event list.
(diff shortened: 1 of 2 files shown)


codebreaker32:
We can keep this PR open until the maintainers reach on consensus if it needs to be added.

mesa/time/events.py
```diff
@@ -402,23 +416,76 @@ class EventList:
...
+        COMPACTION_FLOOR (int): Minimum number of canceled events before compaction is considered
+
+    Notes:
+        An event belongs to one event list at a time. Adding the same event to a
+        second list hands ownership over, and the first list's count goes stale.
 
     """
 
+    COMPACTION_RATIO: float = 0.5
+    COMPACTION_FLOOR: int = 64
```
quaquel: what motivates these values?

mesa/time/events.py
```diff
@@ -402,23 +416,76 @@ class EventList:
...
+
     def __init__(self):
         """Initialize an event list."""
         self._events: list[Event] = []
+        self._n_canceled: int = 0
+        # Built once and handed to every event, since creating it per push is costly.
+        self._self_ref: ReferenceType[EventList] = ref(self)
```
quaquel, line 441: do we need a weak ref for this?

mesa/time/events.py
```diff
@@ -402,23 +416,76 @@ class EventList:
...
     def add_event(self, event: Event):
         """Add the event to the event list.
 
         Args:
             event (Event): The event to be added
 
         """
+        event._owner = self._self_ref
+        if event.CANCELED:
+            self._n_canceled += 1
```
quaquel: why add an event that is already canceled?

mesa/time/events.py
```diff
@@ -402,23 +416,76 @@ class EventList:
...
         heappush(self._events, event)
 
+    def _note_cancellation(self) -> None:
```
quaquel: Not sure about the name of this. It is basically a on_cancelation handler which check if compacting is needed.

mesa/time/events.py
```diff
@@ -449,8 +516,12 @@ def pop_event(self) -> Event:
             event = heappop(self._events)
 
             if not event.CANCELED:
+                # No longer on the heap, so a later cancel() must not be counted here.
+                event._owner = None
```
quaquel, line 524: Should a cancel on an executed event not raise something instead?

quaquel (approved):
Looks fine to me but I have a few clarifying questions.

mesa/time/events.py
codebreaker32: I swept both on a workload where cancellations accumulate instead of draining: 800 agents over 250 steps, each one cancelling its pending event and rescheduling past the end of the run, so no tombstone is ever popped. This is the `Threshold` shape from #3766, where a projected crossing time keeps moving forward. Each cell is median time, peak heap, and calls to `compact()`:

| FLOOR | ratio 0.1 | ratio 0.25 | ratio 0.5 | ratio 0.75 | ratio 0.9 |
|---|---|---|---|---|---|
| 0 | 1.17s 888 2238x | 0.87s 1,066 746x | 1.02s 800 249x | 1.05s 2,564 83x | 1.05s 7,408 27x |
| 64 | 1.12s 888 2238x | 0.82s 1,066 746x | 1.06s 800 249x | 1.07s 2,564 83x | 1.08s 7,408 27x |
| 1024 | 1.02s 1,800 194x | 1.06s 1,800 194x | 1.10s 1,800 194x | 1.05s 2,564 83x | 1.05s 7,408 27x |

Without compaction the same run takes 1.22s and peaks at 200,000 heap entries.

mesa/time/events.py
codebreaker32: Yes, it has to be weak. `Event._owner` points back at the list while the list holds the event, so a strong reference is a cycle, and worse than that, any event a model keeps a handle on would pin the entire list.

mesa/time/events.py
codebreaker32: I missed it and the branch was defensive, and checking it again showed it does not even do its job. It bumps `_n_canceled` without testing the compaction trigger, and an event canceled before it is added has no owner to notify, so neither path checks. 
I would rather remove the branch and raise. `cancel()` is terminal, so a canceled event can never run and adding one to a list is always a caller bug. The only place the library re-adds an event is `_advance_time` pushing back one it just popped, where nothing can cancel in between, so raising is safe. Thoughts?

mesa/time/events.py
codebreaker32: Agreed. I'll rename it.

mesa/time/events.py
codebreaker32: It cannot raise, because `EventGenerator` does exactly that during normal shutdown.

`_current_event` points at the next scheduled event and is only replaced *after* the callback returns, so while your function is running it still points at the event executing right now. `stop()` cancels `_current_event`. That means calling `generator.stop()` from inside the callback cancels an event that has already executed. The dead-weakref path does the same with no user code involved: `_execute_and_reschedule` resolves the callback weakref, finds it dead, and calls `self.stop()` on itself. I verified both against this branch.
Raising would turn both into errors.

mesa/time/events.py
quaquel: I would go with raising then

mesa/time/events.py
quaquel: I am not sure I fully follow. Looking at this table, a ratio of 0.25 in general seems to be faster than 0.5.

mesa/time/events.py
quaquel: clear, I also misread the code.

quaquel (approved):

mesa/time/events.py
codebreaker32: I have messed up something while benchmarking locally, I have had different results which inspired me and it came up differently when I posted. I will recheck and get back to you.

codebreaker32: Rechecked with 25 reps:

| ratio | peak | `L/(1-r)` | compactions | median | in `compact()` |
|---|---|---|---|---|---|
| 0.1 | 889 | 889 | 2,238 | 1.363s | 0.459s |
| 0.25 | 1,067 | 1,067 | 746 | 1.104s | 0.172s |
| 0.5 | 1,599 | 1,600 | 249 | 1.269s | 0.103s |
| 0.75 | 3,197 | 3,200 | 83 | 1.273s | 0.042s |
| 0.9 | 7,991 | 8,000 | 27 | 1.302s | 0.034s |
| none | 200,001 | - | 0 | 1.324s | - |

the peak column differs from the one I posted because that was sampled once per step, at the end, and at 0.5 compaction fires on the last agent of every step, so I was catching the heap just after a rebuild. On every push it is 1,599, not 800. 
You are right that 0.25 is faster. Paired and alternating it wins 21 of 30, and the total is a U with its minimum there.

I took 0.5 from the dynamic array analogy, rebuild once the dead outnumber the live so that each rebuild at least halves the heap. That minimises rebuild cost, which the last column confirms it does, but rebuild cost is not the term that dominates. The ratio is really a bound on dead weight, the heap peaking at `L/(1-r)`, so 1.33x the live set at 0.25 against 2x at 0.5. 0.25 is the better of the two on both counts.

(with `L` live events, `c` cancel-and-reschedules leave the heap holding `L + c` of which `c` are dead, and compaction fires once `c > r(L + c)`, that is `c > rL/(1-r)`, so the heap at that moment is `L + c = L/(1-r)`.)

The floor I can defend less well, and 64 was a guess. Its only visible effect is that a small model carries up to 64 dead events before compacting, raising the peak by that much. I am happy either way, keep it as a cheap guard or drop it.

quaquel: what would the performance be if we drop it

codebreaker32: dropping the floor is free above 192 live events, and below that it trades peak heap for rebuild frequency. Performance wise regression was <+15%

codebreaker32:
All of @quaquel's suggestions have been addressed and I have dropped `COMPACTION_FLOOR` for now.

quaquel (approved):

jackiekazil:
LGTM. I am going to go ahead and merge.

PR #3801, opened 2026-08-14
Add requirements and a `FAILED` state to `Action`

codebreaker32:
### Summary
`Action` gains two requirement lists and a failure path. `requirements` gates the start, `completion_requirements` gates the effect, and a predicate that does not hold moves the action to `ActionState.FAILED` and fires `on_fail()` instead of letting it complete.

### Motive
An action can only succeed today. `Forage` schedules its completion the moment it starts, and five time units later `on_complete()` fires and the sheep gains energy whether or not the conditions still hold. The workaround is to re-check inside `on_complete()` and return early, which repeats the same guard in every action and still leaves the action `COMPLETED` with nothing accomplished.

Declaring the condition moves the check into the framework and gives the failure its own state, so a model can tell "ate" from "got there and could not".

### Implementation
Both lists accept a single `callable(agent) -> bool` or an iterable of them, mirroring the way `duration` and `priority` already accept callables resolved against the agent, and each is normalised into a list the action owns. Subclasses can assign to either after `super().__init__()` instead of passing them in.

`requirements` is checked at `start()`, before duration and priority are resolved, so a failing action never fires `on_start()` or `on_resume()` and never schedules a completion event.

`completion_requirements` is checked in `_do_complete()` before the effect is applied, and is empty by default. Nothing is checked at completion unless a model asks for it, which makes claim-at-start the default path.

They are separate lists rather than one list checked twice because the two questions are different, and separating them leaves the choice to the model developer. A rival resource should be claimed at the start, and the decisions that follow, take what remains, go elsewhere, or wait, belong to the agent rather than to a boolean gate. What the completion list is for is conditions nobody can reserve: a market still open when a trade lands, a counterparty still solvent at settlement.

Neither list is re-checked in between. Continuous revalidation would mean rescanning every active action whenever the world changes, which is activity scanning by another name. An action that must abort the moment a condition breaks belongs to a `Threshold` (#3766) calling `agent.cancel_action()`.

Failure moves the action to `FAILED` and fires `on_fail()` in place of `on_start()` or `on_complete()`. `_fail()` clears `agent.current_action` before firing, so the agent is already idle when the hook runs and can start something else from inside it. On a completion failure `progress` still reads `1.0`, because the full duration did elapse and only the effect is withheld.

`interrupt_for()` changes behaviour. It returned `True` unconditionally once the interruption succeeded; it now returns `not new_action.has_failed`. It deliberately does not roll the interruption back: the old action is already `INTERRUPTED` and the agent is left idle, since whether to resume it is the model's decision. The two `False` cases therefore differ in what they leave behind, a refused interruption changes nothing and a failed requirement does not.

### Usage Examples
Claiming a rival resource at the start, which is where contention belongs:

```python
class Graze(Action):
    def __init__(self, sheep, patch):
        super().__init__(sheep, duration=3.0, requirements=lambda a: patch.servings > 0)
        self.patch = patch

    def on_start(self):
        self.patch.servings -= 1  # claimed, so nobody else can take it

    def on_complete(self):
        self.agent.energy += 30

    def on_fail(self):
        self.agent.energy -= 2  # walked there for nothing
        self.agent.look_elsewhere()
```

Two sheep starting on a patch with one serving: the first claims it and grazes unimpeded, the second never starts and decides what to do next from `on_fail()`.

Gating an effect on something nobody can reserve:

```python
Trade(trader, duration=5.0, completion_requirements=lambda a: market.open)
```

A trade that starts at 15:58 and lands at 16:03 ends `FAILED`. Checking at the start would have reported the market open, which is true and useless, and there is no claim to take at 15:58 that would keep it open.

### Additional Notes
A requirement that fails on **resume** is terminal. An action interrupted at 40% whose requirement then breaks moves to `FAILED`, keeps its 0.4 progress, and cannot be restarted even once the condition is repaired; the model has to build a fresh action and lose the partial work. The alternative, sending a failed resume back to `INTERRUPTED` so it stays resumable, was rejected because `start()` would then have to accept `FAILED`, and `resuming = self.state is ActionState.INTERRUPTED` would misclassify it, calling `on_start()` instead of `on_resume()` and re-resolving duration. If transient requirements turn out to be the common case, this is the piece to revisit.

Requirements are callables, so a lambda makes a model unpicklable. That is not new, `duration=lambda a: ...` already has the same effect; named functions and bound methods pickle fine.

The third item of section 2 in #3798, giving `priority` an effect, is not here. It is #3805, which adds `Agent.should_interrupt(current, incoming)` and rewrites the same three lines of `interrupt_for()`.

### DIFF
--- a/mesa/experimental/actions/actions.py
+++ b/mesa/experimental/actions/actions.py
@@ -4,8 +4,8 @@ An Action represents something an agent does over time. It integrates with
 Mesa's event scheduling system for precise timing and supports interruption
 with progress tracking and optional resumption.
 
-Actions are subclassable: override on_start(), on_complete(), and
-on_interrupt() to define behavior.
+Actions are subclassable: override on_start(), on_complete(),
+on_interrupt(), and on_fail() to define behavior.
 
 Example::
 
@@ -24,7 +24,7 @@ Example::
 
 from __future__ import annotations
 
-from collections.abc import Callable
+from collections.abc import Callable, Iterable
 from enum import IntEnum, auto
 from typing import TYPE_CHECKING
 
@@ -40,6 +40,7 @@ class ActionState(IntEnum):
     ACTIVE = auto()
     COMPLETED = auto()
     INTERRUPTED = auto()
+    FAILED = auto()
 
 
 class Action:
@@ -61,7 +62,10 @@ class Action:
         priority: Importance level. Higher = more important. May be a
             callable(agent) -> float, resolved at start time.
         interruptible: Whether higher-priority actions can preempt this.
-        state: Current lifecycle state (PENDING, ACTIVE, COMPLETED, INTERRUPTED).
+        requirements: Predicates the world must satisfy for the action to
+            run. Each is a callable(agent) -> bool.
+        state: Current lifecycle state (PENDING, ACTIVE, COMPLETED,
+            INTERRUPTED, FAILED).
         progress: Time fraction completed, 0.0 to 1.0. Computed live
             while the action is active.
 
@@ -79,6 +83,9 @@ class Action:
         name: str | None = None,
         priority: float | Callable[[Agent], float] = 0.0,
         interruptible: bool = True,
+        requirements: Callable[[Agent], bool]
+        | Iterable[Callable[[Agent], bool]]
+        | None = None,
     ) -> None:
         """Initialize an Action.
 
@@ -92,11 +99,19 @@ class Action:
                 a float or a callable that receives the agent and returns
                 a float. Resolved when start() is called.
             interruptible: If False, interrupt() will fail and return False.
+            requirements: A single callable(agent) -> bool, or an iterable
+                of them.
         """
         self.agent = agent
         self.model = agent.model
         self.interruptible = interruptible
         self._name: str | None = name
+
+        if requirements is None:
+            requirements = []
+        elif callable(requirements):
+            requirements = [requirements]
+        self.requirements: list[Callable[[Agent], bool]] = list(requirements)
 
         # Store raw values (may be callables, resolved at start)
         self._duration_spec = duration
@@ -162,6 +177,11 @@ class Action:
         """Whether this action can be resumed (interrupted, not completed)."""
         return self.state is ActionState.INTERRUPTED and self._progress < 1.0
 
+    @property
+    def has_failed(self) -> bool:
+        """Whether a requirement did not hold, at start or at completion."""
+        return self.state is ActionState.FAILED
+
     # --- Lifecycle methods (override in subclasses) ---
 
     def on_start(self) -> None:
@@ -206,6 +226,19 @@ class Action:
             necessarily cancelled, not interrupted.
         """
 
+    def on_fail(self) -> None:
+        """Called when a requirement does not hold.
+
+        Override to release anything the action
+        reserved, or to record the attempt.
+
+        Notes:
+            Check self.progress to tell the two failures apart: it is 1.0
+            when the action ran its full duration and only then found the
+            requirement broken, and below 1.0 when the action never got to
+            run this attempt.
+        """
+
     # --- Execution (called by Agent, not typically by users) ---
 
     def start(self) -> Action:
@@ -215,11 +248,16 @@ class Action:
         starts from progress=0. On resume (INTERRUPTED): continues from
         existing progress with remaining duration.
 
+        If a requirement does not hold the action moves to FAILED instead,
+        firing on_fail() rather than on_start() or on_resume(). Inspect
+        state after starting rather than assuming the action is ACTIVE.
+
         Returns:
             Self, for chaining.
 
         Raises:
             ValueError: If the action is not in PENDING or INTERRUPTED state.
+                A FAILED action cannot be restarted; create a new one.
         """
         resuming = self.state is ActionState.INTERRUPTED
 
@@ -228,6 +266,12 @@ class Action:
                 f"Cannot start action in {self.state.name} state. "
                 f"Only PENDING or INTERRUPTED actions can be started."
             )
+
+        # Gate entry before duration and priority are resolved, so a failing
+        # action never fires on_start() nor schedules a completion event.
+        if not self._requirements_met():
+            self._fail()
+            return self
 
         # Resolve callables on first start only
         if not resuming:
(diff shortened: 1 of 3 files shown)


quaquel:
> They are checked in exactly two places. At start() the check gates entry before duration and priority are resolved, so a failing action never fires on_start() or on_resume() and never schedules a completion event. At _do_complete() the check runs again before the effect is applied.

I am not sure about this and I am not convinced by the grazing example. Typically, you would claim the resource at the start of the action, rather than check at the end if the resource is still available. So, if you arive at a station and there is not enough resource remaining, the action would be to claim just the remaining resource, move to another station, or wait until enough of the resource is again available. 

Also, requirements at start need not be the same as requirements on completion. Can this be handled?

codebreaker32:
Sorry, I missed it yesterday and apologies for the wrong example.

Your three responses, take what remains, go elsewhere, or wait, all need to know how much is there. A requirement returns True or False, so it cannot express "ten were wanted, three are left, take the three". In writing that example I presented `requirements` as a way to acquire a resource. It is not one, and it should not become one. Acquisition needs a quantity and a decision, and both belong to the agent, in `on_start` or before it, where it can query the station, take what is actually there, and choose between your three outcomes.

What a boolean gate is good for is preconditions that are genuinely binary, where the only sensible response is not to start at all: the agent holds a permit, the vehicle has a driver, the door is unlocked. Nothing to claim, no quantity to negotiate.
That also leaves the completion check with much less to justify it, since the grazing example was the justification.
What I think still needs it is state that nobody can reserve. A trade takes five minutes and starts at 15:58, the market closes at 16:00, and the trade lands at 16:03. Checking at the start says the market is open, which is true and useless, and there is no claim to take at 15:58 that would keep it open, because opening hours are not a resource. A settlement that requires the counterparty still solvent when the money moves has the same shape. In those cases the start check answers the wrong question.
If you would rather handle that with a `Threshold` that cancels the action the moment the condition breaks, then the completion check has nothing left to do.

On your second question, yes:
```python
Action(agent, requirements=[...], completion_requirements=[...])
```

Independent lists rather than one list checked twice, with the completion list empty by default so nothing is checked at the end unless a model asks for it.

I will replace the example with a claim-based one either way. It makes the boundary obvious, and the current one teaches the wrong pattern.

quaquel:
If we separate on_start requirements and on_completion requirements, it leaves it to the model developer to decide and gives them also a lot more flexibility.

codebreaker32:
Yes and using `Threshold` for other cases you mentioned. I will try to build an example demonstrating it

mesa/experimental/actions/actions.py
```diff
@@ -79,6 +95,12 @@ def __init__(
         name: str | None = None,
         priority: float | Callable[[Agent], float] = 0.0,
         interruptible: bool = True,
+        requirements: Callable[[Agent], bool]
```
quaquel: should this not be `start_requirements`

mesa/experimental/actions/actions.py
```diff
@@ -92,12 +114,21 @@ def __init__(
...
+                of them. All must hold for the action to start.
+            completion_requirements: The same, checked instead when the action
+                completes, gating the effect rather than the attempt.
         """
         self.agent = agent
         self.model = agent.model
         self.interruptible = interruptible
         self._name: str | None = name
 
+        self.requirements: list[Callable[[Agent], bool]] = _as_predicates(requirements)
```
quaquel: same here

quaquel (commented):
I have a few minor points at the code level.

Also, we now have start and completion requirements. But what about an interrupt? How would that work?

codebreaker32:
Renamed, they are `start_requirements` and `completion_requirements` now.

On interrupt:

1. `start()` checks `start_requirements`. They hold, so the action goes `ACTIVE`, `on_start()` fires, and a completion event is scheduled for the full duration.
2. `interrupt()` freezes progress, cancels that completion event, moves the action to `INTERRUPTED` and fires `on_interrupt(progress)`. Nothing is checked. The action is not deciding anything here, it is being stopped.
3. `start()` again re-checks `start_requirements`, because resume goes through the same entry point rather than a separate one. If they hold, the action goes `ACTIVE`, `on_resume()` fires, and a completion event is scheduled for the remaining duration only. If they do not, the action moves to `FAILED` and `on_fail()` fires, with `on_resume()` never called.
4. When that event lands, `completion_requirements` decide whether the effect applies: `COMPLETED` and `on_complete()`, or `FAILED` and `on_fail()`.

`completion_requirements` are never consulted at step 2 or 3. They gate the effect, and an interrupted action has no effect to gate yet.

For eg. A technician on a eight hour shift. The technician has to be on shift to work on it, and the overhaul only counts if the machine is still in service when it lands:

```python
class Repair(Action):
    def __init__(self, tech, machine):
        super().__init__(
            tech,
            duration=8.0,
            start_requirements=lambda a: a.on_shift,
            completion_requirements=lambda a: machine.in_service,
        )
```

An emergency callout preempts it three hours in, and runs an hour past the end of the shift:

```
t=0.0  repair started
t=3.0  repair paused at 38%
t=3.0  emergency callout
t=9.0  repair failed (38% done)      # resume re-checks on_shift, shift ended at 8
```

And separately, the same overhaul running undisturbed while the machine is written off on the bench:

```
t=0.0  repair started
t=8.0  repair failed (100% done)     # completion requirement withheld the effect
```

One consequence worth noting from that first trace. The 38% is lost: `FAILED` is terminal, and there is no public way to construct an Action that starts part-done, so a model wanting to pick the overhaul up next morning starts from zero. I am inclined to leave that as is, though. By the same argument as before, an agent should not be attempting a resume it cannot perform, the gate is a safety net, not control flow, so the stranded progress is a model deciding badly rather than the framework losing information it should have kept. Thoughts?

Separately, I also have had a question. `cancel_action()` from #3461 does not cancel: it force-interrupts, ignoring the `interruptible` flag, and leaves the action `INTERRUPTED` and resumable. There is no `CANCELLED` state. That read fine when there was no terminal state to confuse it with, but `FAILED` gives us one.
Is that worth renaming? The obvious candidate, `interrupt_action()`, matches the state but hides the thing that makes the method useful: `interrupt_for()` is refused by a non-interruptible action while this one overrides it, so two methods called "interrupt" would behave differently on the same flag, `force_interrupt()` keeps both halves.

quaquel (approved):

EwoutH:
Thanks for working on this. Interesting stuff.

Actions do get more powerful but also more complicated this way. A documentation/tutorial effort might be useful.

codebreaker32:
Right, and that is indeed point 5 of #3798. I am a bit preoccupied currently but I will land it by the end of this week.

PR #3802, opened 2026-08-15
Sync agent cell reference in Cell.add_agent and Cell.remove_agent

### LINKED ISSUE #3790: discrete_space: Synchronize Cell-Agent References in Cell.add_agent and Cell.remove_agent
### PR Summary

`discrete_space`: Synchronize Cell-Agent References in `Cell.add_agent` and `Cell.remove_agent`

## Problem Statement

In Mesa 3.0's `discrete_space` module, spatial occupation is represented by a bidirectional relationship between `Cell` and `CellAgent`:
1. `cell.agents`: The list of agents occupying the cell.
2. `agent.cell`: The reference pointing to the agent's current cell.

When model builders use `agent.move_to(cell)` or `agent.cell = cell`, both references are updated consistently.

### The Bug / State Corruption

When model builders call `cell.add_agent(agent)` or `cell.remove_agent(agent)` directly:
- `cell.add_agent(agent)` adds `agent` to `cell._agents`, but **fails to update `agent._mesa_cell`** or remove `agent` from its previous cell. The agent exists in multiple cells simultaneously!
- `cell.remove_agent(agent)` removes `agent` from `cell._agents`, but **fails to set `agent._mesa_cell = None`**. The agent enters a "phantom state" where `agent.cell` points to a cell that has no record of it.

Later in the simulation, when the phantom agent attempts to move (`agent.move_to(new_cell)` or `agent.move("north")`), Mesa calls `old_cell.remove_agent(self)`. Since `old_cell` no longer contains the agent, Mesa crashes with `AgentMissingException: Agent 1 is not in cell (0, 0)`.

---

## Root Cause Analysis

In `mesa/discrete_space/cell.py`:

```python
def add_agent(self, agent: CellAgent) -> None:
    ...
    self._agents.append(agent)  # Only mutates local list

def remove_agent(self, agent: CellAgent) -> None:
    ...
    self._agents.remove(agent)  # Only mutates local list
```

`Cell.add_agent()` and `Cell.remove_agent()` perform unidirectional mutation on `self._agents` without synchronizing `agent._mesa_cell`. This breaks the fundamental spatial invariant:

$$\text{agent} \in \text{cell.agents} \iff \text{agent.cell} == \text{cell}$$

---

### Strategy: Atomic Two-Way Synchronization

Update `Cell.add_agent()` and `Cell.remove_agent()` so that mutating a cell's agent list automatically synchronizes the agent's internal cell reference (`agent._mesa_cell`):

1. **`add_agent(agent)`:** If `agent` belongs to a different cell (`old_cell`), remove it from `old_cell`, append to `self._agents`, and set `agent._mesa_cell = self`.
2. **`remove_agent(agent)`:** Remove `agent` from `self._agents`, and if `agent._mesa_cell is self`, set `agent._mesa_cell = None`.

---

NoiceHax:
Cell.add_agent and Cell.remove_agent only updated Cell._agents and left agent.cell alone, so the two sides could disagree. After cell.add_agent(agent) the agent still reported its cell as None, and after cell.remove_agent(agent) it still pointed at the cell it had just left. The next move_to then tried to detach the agent from a cell it was no longer in and raised AgentMissingException. Adding an agent to a second cell also left it registered in the first one.

Both methods now own the back reference. add_agent detaches the agent from its current cell and points _mesa_cell at itself, remove_agent clears it, and HasCell.cell delegates to them instead of assigning _mesa_cell on its own.

add_agent also set self.empty = False before the capacity check, so a failed add on a capacity 0 cell marked the cell as occupied. The check now runs first, so a CellFullException leaves the agent where it was.

Added regression tests for each case. Fixes #3790.

### DIFF
--- a/mesa/discrete_space/cell.py
+++ b/mesa/discrete_space/cell.py
@@ -143,20 +143,30 @@ class Cell:
     def add_agent(self, agent: CellAgent) -> None:
         """Adds an agent to the cell.
 
+        The agent is removed from the cell it currently occupies, if any, and
+        its `cell` reference is updated to point to this cell.
+
         Args:
             agent (CellAgent): agent to add to this Cell
 
         """
-        n = len(self._agents)
-        self.empty = False
-
-        if self.capacity is not None and n >= self.capacity:
+        # capacity is checked before anything is mutated, so a CellFullException
+        # leaves both this cell and the agent untouched
+        if self.capacity is not None and len(self._agents) >= self.capacity:
             raise CellFullException(self.coordinate)
 
+        old_cell = agent._mesa_cell
+        if old_cell is not None and old_cell is not self:
+            old_cell.remove_agent(agent)
+
         self._agents.append(agent)
+        self.empty = False
+        agent._mesa_cell = self
 
     def remove_agent(self, agent: CellAgent) -> None:
         """Removes an agent from the cell.
+
+        The agent's `cell` reference is cleared as well.
 
         Args:
             agent (CellAgent): agent to remove from this cell
@@ -168,6 +178,8 @@ class Cell:
             raise AgentMissingException(agent, self.coordinate) from e
 
         self.empty = self.is_empty
+        if agent._mesa_cell is self:
+            agent._mesa_cell = None
 
     @property
     def is_empty(self) -> bool:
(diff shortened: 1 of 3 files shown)


quaquel:
Let's first agree in #3790, and I am fairly sure that were previous discussions and even PRs on this.

PR #3803, opened 2026-08-15
Warn once per call for dict agent_portrayal

### LINKED ISSUE #3455: Agent portrayal deprecation warning floods console (emitted per agent per render)

NoiceHax:
Fixes #3455.

The FutureWarning that asks you to stop returning a dict from `agent_portrayal` sits inside the per agent loop in `collect_agent_data`. With a few thousand agents you get a few thousand copies of the same message, and under Solara that repeats on every render, so the console fills up and drawing gets slower for no reason.

This adds a local flag before each of the three loops (the matplotlib backend, the altair backend, and `mpl_space_drawing.collect_agent_data`) and only warns the first time a dict portrayal turns up in that call. The dict to `AgentPortrayalStyle` conversion is unchanged, so nothing about the drawn output moves. I left the per agent `UserWarning` about ignored keys alone, since different agents can return different extra keys and folding those together would drop information.

Added tests that a multi agent dict portrayal produces exactly one FutureWarning at each of the three sites. `tests/visualization` passes with `-Werror`.

### DIFF
--- a/mesa/visualization/backends/altair_backend.py
+++ b/mesa/visualization/backends/altair_backend.py
@@ -94,21 +94,26 @@ class AltairBackend(AbstractRenderer):
             "4": "triangle-right",
         }
 
+        # Warn only once per call instead of once per agent
+        dict_portrayal_warned = False
+
         for agent in space.agents:
             portray_input = agent_portrayal(agent)
             aps: AgentPortrayalStyle
 
             if isinstance(portray_input, dict):
-                warnings.warn(
-                    (
-                        "Returning a dict from agent_portrayal is deprecated and will be removed in Mesa 4.0. "
-                        "Please return an AgentPortrayalStyle instance instead. "
-                        "For more information, refer to the migration guide: "
-                        "https://mesa.readthedocs.io/latest/migration_guide.html#defining-portrayal-components"
-                    ),
-                    FutureWarning,
-                    stacklevel=2,
-                )
+                if not dict_portrayal_warned:
+                    warnings.warn(
+                        (
+                            "Returning a dict from agent_portrayal is deprecated and will be removed in Mesa 4.0. "
+                            "Please return an AgentPortrayalStyle instance instead. "
+                            "For more information, refer to the migration guide: "
+                            "https://mesa.readthedocs.io/latest/migration_guide.html#defining-portrayal-components"
+                        ),
+                        FutureWarning,
+                        stacklevel=2,
+                    )
+                    dict_portrayal_warned = True
                 dict_data = portray_input.copy()
                 agent_x, agent_y = self._get_agent_pos(agent, space)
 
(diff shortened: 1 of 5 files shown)


quaquel (approved):
Looks fine to me

PR #3805, opened 2026-08-16
Add `Agent.should_interrupt` so `Action.priority` decides preemption

codebreaker32:
### Summary
`Agent.interrupt_for()` now consults a new hook, `Agent.should_interrupt(current, incoming)`, before preempting. The default is `current.interruptible and incoming.priority >= current.priority`, which gives `Action.priority` its first effect. Section 4 of #3798; it also closes the last bullet of section 2.

### Motive
`Action.priority` is resolved at `start()` and read by nothing. `git grep` on `main` finds one assignment and no reads, so a priority-1 action preempts a priority-100 one and the only lever a model has is the binary `interruptible` flag. The 4.0.0a0 notes list "callable duration and priority for state-dependent values" as a feature; the priority half is inert.

Preemption is also the agent's decision. Whether a sheep abandons foraging when a predator appears is behavior, so it belongs on `Agent` rather than hard-coded in `interrupt_for`.

### Implementation
`should_interrupt` lives on `Agent` next to `start_action` and `interrupt_for`. It is consulted only when the agent is busy, and only decides whether to *attempt* the interruption: returning `True` still goes through `Action.interrupt()`, which refuses a non-interruptible action as before. The hook cannot override the `interruptible` flag; `cancel_action()` is still the way to force it.

The default uses `>=` rather than `>` because priorities default to `0.0`, and equal-priority preemption is how `interrupt_for` has always behaved. `>` would have broken three tests already on `main` (`test_interrupt_for_replaces_action`, `test_sheep_forage_flee_resume`, `test_worker_interrupted_resumes_task`); with `>=` nothing existing changes.

The incoming action is still `PENDING` when the hook runs, and priority was only resolved inside `start()`, so a callable priority would have read `0.0` when the hook compared it. `Action._resolve_priority()` now resolves the spec at most once, called from both `start()` and `interrupt_for()`. A callable spec is invoked exactly once regardless of which path reaches it first, and a refused action keeps its resolved priority for a later `start_action`.

### Usage Examples
```python
class Sheep(Agent):
    def should_interrupt(self, current, incoming):
        if current.name == "Flee":
            return False  # never stop fleeing, whatever the priority
        return super().should_interrupt(current, incoming)


sheep.start_action(Forage(sheep, priority=1.0))
sheep.interrupt_for(Flee(sheep, priority=10.0))    # True, forage interrupted
sheep.interrupt_for(Forage(sheep, priority=1.0))   # False, flee continues
```

Without the override, the second `interrupt_for` would already be refused by the default on priority alone; the override makes it hold even if a caller passes a higher number.

### Additional Notes
`interrupt_for` with an action that cannot be started, one belonging to a different agent or already `COMPLETED`, interrupts the current action and then raises from `start_action`, leaving the agent idle. That predates this PR and is unchanged. The hook now runs first, so a lower-priority bad action is at least refused before the interruption happens; validating the incoming action up front would fix it fully and is a small separate change.

Where the hook lives is the open placement question from #3798: `on_idle` and `should_interrupt` on core `Agent`, or the whole action cluster behind an experimental mixin. It sits on `Agent` here because `start_action`, `interrupt_for` and `cancel_action` already do, and moving all of them together later is one change rather than several.


quaquel:
I like this PR. However, it does raise a bigger question: should all the action stuff reside inside the main Agent class or is there a way we can make this work through composition. The larger the base Agent class becomes, the steeper the learning curve becomes for a new user. If instead we can decompose the various parts into separate subclasses or something like that, it becomes easier for a user to compose their base agent class from the pieces that are needed for their model without having to worry about what they don't need. 

I approved #3801. So, I can merge this if you want me to. That will clean up this diff.

jackiekazil:
RE: Discussion of example to illustrate this PR and #3801 

I see two options -- 

1. Wolf Sheep (as mentioned)
2. Tram example - adding an advanced version that is a multi-tram transportation example 

Wolf Sheep seems to be a better fit to me. (But I could be talked into an advanced Tram example.)

codebreaker32:
> it does raise a bigger question: should all the action stuff reside inside the main Agent class or is there a way we can make this work through composition

You are right, and I have also raised this question as you can see in PR description. The current situation is odd on its own terms: `current_action`, `start_action`, `interrupt_for`, `cancel_action`, and `is_busy` have sat on the stable `Agent`, while `Action` itself is experimental - so the stable base class carries API whose semantics may still change (this PR changing preemption behavior is a case in point).

I think moving the whole cluster behind a mixin, would fix both things at once: `Agent` stays lean, and the experimental surface lives in the experimental package.

codebreaker32:
> RE: Discussion of example to illustrate this PR and #3801
> 
> I see two options --
> 
> 1. Wolf Sheep (as mentioned)
> 2. Tram example - adding an advanced version that is a multi-tram transportation example
> 
> Wolf Sheep seems to be a better fit to me. (But I could be talked into an advanced Tram example.)

I am also inclined towards Wolf Sheep. Meanwhile I was also experimenting with meerkat clan that may look interested:

- An alarm is `interrupt_for` as a broadcast, and each forager's `should_interrupt` weighs its dig against the alarm's urgency (`priority`) - meerkats really do ignore mild alarms when prey is nearly out (graded alarm calls, [Manser 2001](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC1088882/)).
- Sentinel duty rotates through `interrupt_for` too: the best-fed meerkat is asked to take the mound ([Clutton-Brock et al. 1999](https://www.science.org/doi/10.1126/science.284.5420.1640)) and a digger with a scorpion nearly out refuses - and refusals have a consequence, because an unguarded clan gets no early warning.
- Daylight and prey-still-there are `start_requirements`, re-checked when an interrupted dig resumes; prey escaping underground is the `completion_requirements` case; and a callable `duration` sizes tomorrow's dig to the sand already moved.

jackiekazil:
**RE: Mix-ins** - @quaquel @codebreaker32 thank you bringing it up. I really like this idea. Should we do this now or separate PR? I am thinking now since this PR has a clear impact on agent.py now. Thoughts? 

**RE: Example** - Conceptually I like the meerkat clan example, however we really try to keep the examples in core lean and not add new ones. The examples in the main repo are utilitarian. If you feel that wolf-sheep can't handle it, then less discuss merkat more. Another option is to put it in https://github.com/mesa/mesa-examples, but we would still need the illustrative example in the core mesa repo.

quaquel:
> RE: Mix-ins - @quaquel @codebreaker32 thank you bringing it up. I really like this idea. Should we do this now or separate PR? I am thinking now since this PR has a clear impact on agent.py now. Thoughts?

It's fine as a future PR in my view. 

> RE: Example - Conceptually I like the meerkat clan example, however we really try to keep the examples in core lean and not add new ones. The examples in the main repo are utilitarian. If you feel that wolf-sheep can't handle it, then less discuss merkat more.

Does this all need to be covered by examples in the main Repo? I like the meerkat example, but we might first add it to mesa-examples instead. I prefer not to change the sheep wolf example. It's a classic ABM so perfect to demonstrate standard mesa. Alternatively, we seem to have experimental as a new folder in examples. I don't know what I think of that but we could do meerkat there also.

codebreaker32:
> **RE: Mix-ins** - @quaquel @codebreaker32 thank you bringing it up. I really like this idea. Should we do this now or separate PR? I am thinking now since this PR has a clear impact on agent.py now. Thoughts?

I also think that future PR would be better

jackiekazil (approved):
I am good with merging this in, while we discuss the example further. 
@quaquel feel free to merge when you are ready.

jackiekazil:
RE: Mix-in - all agreed on future mix-in 

RE: Example -  
> Does this all need to be covered by examples in the main Repo? 
* That was the goal with the examples in the example folder -- all features are illustrated so people have a clear up to date example of how the feature works.
* I agree with not changing the wolf sheep example. I was hesitant as well. 
* RE: Experimental example folder -- that is exactly what it was added for. 
* **Agree -- let's put a meerkat example the experimental folder.** (I would like to keep the core repo examples as slim as possible, but also understand that might not be possible without bloating the content of the example.)

quaquel:
There is no consensus yet on how to address this. It's the third PR in a week that I am closing.

codebreaker32:
I think you mistakenly closed this PR. Isn't it @quaquel ?

quaquel (approved):

PR #3807, opened 2026-08-18
Update discord invite link

wang-boyu:
The current discord invite link will expire, and this PR replaces that with a new link that doesn't. It also updates two matrix chat links in the code of conduct doc to discord chat.

Screenshot 2026-08-18 at 2 17 54 PM

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -71,8 +71,10 @@ representative at an online or offline event.
 ## Enforcement
 
 Instances of abusive, harassing, or otherwise unacceptable behavior may be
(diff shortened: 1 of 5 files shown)


jackiekazil:
LGTM.

jackiekazil (approved):

dietcoke-17 (commented):
Just a quick follow up question that, is this link wont be restricting the number of new members joining means limiting upto a certain number of members then again in that condition we have to update this manually

wang-boyu:
Yes I did select "no limit" for the max number of users, if I remember correctly. It's strange that Discord doesn't provide a way to easily validate that after an invite link is created.

shyama7004 (approved):
lgtm

Issue #3016, opened 2025-12-26
[ENH] Support overlapping hierarchies by allowing agents to belong to multiple MetaAgents


Nithurshen:
**What's the problem this feature will solve?**

Currently, the `MetaAgent` implementation uses a single reference: `agent.meta_agent = self`. This creates a 1-to-1 relationship between a constituent agent and its meta-agent. The current implementation overwrites previous meta-agent references, making overlapping hierarchies impossible to model.

**Describe the solution you'd like**

I would like to transition `agent.meta_agent` from a single object reference to a set.
1) In `MetaAgent.__init__` and `add_constituting_agents`, the logic should append the meta agent to this set.
2) The `remove_constituting_agents` method should use discard on the set.
3) This ensures that an agent can be tracked by multiple higher-level entities without losing references.

**Additional context**

This addresses the `# TODO: Make a set for meta_agents` found in `mesa/meta.py`.

Nithurshen:
@EwoutH @quaquel
I would like to work on this. Am I good to go?

EwoutH:
@tpike3 is the designer of meta agents.

Nithurshen:
@EwoutH, we will wait for their say on this.

sakshi-thigale:
Hi — I’ve been looking into MetaAgent and the TODO in mesa/meta.py.

Supporting overlapping hierarchies via agent.meta_agents: set makes sense conceptually. Before starting, I just want to confirm whether there are any existing assumptions in activation/scheduling that rely on a strict 1-to-1 relationship.

If the intended direction is to allow true multi-membership, I’d be happy to take this on and include tests to ensure no regressions.

PR #3811, opened 2026-08-20
Meta-Agents Rewrite

falloficaruss:
> [!NOTE]
>This PR is a part of GSoC 2026 project [*"Hypergraph-Based Meta-Agents for Mesa"* 
](https://summerofcode.withgoogle.com/programs/2026/projects/8Rhd5XVJ)
Related: #3016, #3403 https://github.com/mesa/mesa-examples/pull/483
> A detailed write-up is available [here](https://gist.github.com/falloficaruss/50188788e106cb766d4764fbbfe3d491)

### Summary

A complete rewrite of Mesa's `meta-agent` capability. Meta-agents are agents composed of other agents (alliances, households, teams, org units) with overlapping, typed, and nested memberships. This PR replaces the old experimental implementation where each group object held its own member set and membership state was duplicated on member agents with a _**backend-authoritative architecture**_: all memberships live as `(member, group, relation)` triplets in a single store on the model, and every read/write goes through one `MetaAgents` manager. The module is moved out from `mesa.experimental.meta_agents` to a top-level `mesa.meta_agents` package (the old API is removed outright, consistent with the Mesa 4.0 breaking-change window).

### Motive

The old implementation scattered membership state across live objects (`meta_agent.constituting_agents`, `agent.meta_agent`, `agent.meta_agents`), which made overlapping and nested hierarchies a bolt-on rather than a structural property, gave no single source of truth for invariants, and offered no indexed queries. The rewrite follows @tpike3's direction: a hypergraph-style conceptual basis, a canonical incidence backend, and a facade-style public API making overlap, typed relations, and multi-level containment first-class while keeping groups as ordinary Mesa agents that participate in scheduling, datacollection, and visualization.

### Implementation

Three layers under `mesa/meta_agents/`:

**`MembershipBackend` (`backend.py`)** :
- Memberships as `(member_id, group_id, relation)` triplets with bidirectional `_by_agent` / `_by_group` indexes
- Identity normalized to `unique_id`, so the store never holds live-object references (no dangling pointers when agents die)
- Idempotent add/remove, `replace_relation`, bulk add, cascade `remove_agent` / `remove_group`, and `assert_invariants` for tests

**`MetaAgents` manager (`meta_agents_api.py`)** : the public API, installed once per model (`model.meta_agents = MetaAgents(model)`) and the intended only write path:
- `create(...)` builds a live group agent (dynamic subclass of user-supplied or existing model classes, with `meta_attributes` / `meta_methods`) and records edges in bulk
- `add_member` / `remove_member` accept live objects, `unique_id`s, or group names
- `members_of` / `groups_of` return live `AgentSet`s, optionally filtered by relation
- `query_memberships` returns an immutable `MembershipView` of resolved `MembershipEdge`(s)
- `deactivate` (detach all edges, keep the object) vs `dissolve` (detach and remove from the model)
- `at_level(n, root=...)` - BFS containment depth with shallowest-path semantics; cycle-safe
- Subscribes to `ModelSignals.AGENT_REMOVED` so memberships are deactivated automatically when an agent leaves the model

**Internal `MetaAgent` + discovery helpers (`meta_agent.py`)** a thin `Agent` subclass bound to the manager (its `remove()` routes through `dissolve`); `find_combinations` / `evaluate_combination` are kept as pure, non-mutating discovery helpers.

Key design decisions:
- Groups cache nothing — every read resolves live from the backend, so overlap and multi-level nesting (groups of groups) cost nothing structurally
- Typed relation labels (default `"member"`) allow several roles between the same agent–group pair
- All query results use deterministic id-stable ordering

### Usage Examples

```python
from mesa import Agent, Model
from mesa.meta_agents import MetaAgents

model = Model()
model.meta_agents = MetaAgents(model)

alice, bob = Agent(model), Agent(model)
team = model.meta_agents.create("Team", [alice, bob], Agent)

model.meta_agents.add_member("Team", Agent(model))
model.meta_agents.members_of("Team")        # AgentSet
model.meta_agents.groups_of(alice)          # AgentSet
model.meta_agents.query_memberships(alice)  # MembershipView
model.meta_agents.at_level(1, root=team)    # containment depth
model.meta_agents.dissolve(team)
```

The in-tree alliance formation model (moved experimental -> advanced) exercises the full surface: agents form alliances via a bilateral Shapley evaluation, alliances become meta-agents that form higher-level alliances, with overlapping memberships tracked by the manager and rendered as a multi-level network.
Additional Notes
- Docs: new docs/apis/meta_agents.md page (added to the API index) and a migration-guide section with an old→new mapping for the removed helpers (create_meta_agent, agent.meta_agent, constituting_* mutators, etc.)
- Tests: 44 tests covering the backend (indexes, cascades, invariants, multi-relation), the facade (create/add/remove/query/dissolve/deactivate/at_level incl. cycles and overlap), multi-level scenarios, example-level behavior, and the import namespace; old experimental test suites (~600 lines) removed with the old implementation
- Benchmarks: MultiLevelAllianceModel small/large configurations added to benchmarks/configurations.py
- Breaking change: mesa.experimental.meta_agents is gone with no shim; the warehouse example in mesa-examples still uses the old factory and is addressed in a companion PR to `mesa-examples`: mesa/mesa-examples#483
- MetaAgents currently observes signals from mesa.experimental.mesa_signals; the dependency boundary needs to be resolved before Mesa 4 release

### Future Scope & Limitations

- make `create` always create a new group with an explicit join path (today it is join-or-create when a same-class group exists; the alliance example works around this with unique class names)
- **Scale**: incremental live-entity index (membership lookups currently scan `model.agents` per call)
- **Docs**: expansion of the how-to page and add a tutorial


benchmarks/configurations.py
```diff
@@ -126,4 +128,18 @@
...
+            "replications": 25,
+            "iterations": 3,
+            "steps": 10,
+            "scenario": AllianceScenario(n=50, mean=0.5, std_dev=0.1, rng=42),
+        },
+        "large": {
+            "replications": 10,
+            "iterations": 3,
+            "steps": 20,
+            "scenario": AllianceScenario(n=120, mean=0.5, std_dev=0.1, rng=42),
```
tpike3: I think number of agents is what matters here not the number of steps it typically stops forming alliances in a small number of steps, I would keep steps as 10 and change n to 200.

docs/apis/meta_agents.md
```diff
@@ -0,0 +1,56 @@
...
+from mesa.meta_agents import MetaAgents
+
+model = Model()
+model.meta_agents = MetaAgents(model)
+
+alice = Agent(model)
+bob = Agent(model)
+team = model.meta_agents.create("Team", [alice, bob], Agent)
+
+model.meta_agents.add_member("Team", Agent(model))
```
tpike3: Can you add comments here that refers back to the features previously discussed similar to the PR write up (e.g., create, add, etc) for example
```
model.meta_agents.at_level(1) # walk nested levels
model.meta_agents.at_level(1, root=team) # walk nested levels and meta-agent types
```

mesa/meta_agents/meta_agent.py
```diff
@@ -0,0 +1,254 @@
...
+    model,
+    evaluation_func: Callable[[tuple[Agent, ...]], float] | None,
+) -> tuple[tuple[Agent, ...], float] | None:
+    """Evaluate a candidate meta-agent group with a user-supplied function."""
+    if evaluation_func is None:
+        return None
+    return candidate_group, evaluation_func(candidate_group)
+
+
+def find_combinations(
```
tpike3: I think this would probably be cleaner in the meta_agents_api as an @staticmethod. This way all public functions are in the api

mesa/meta_agents/meta_agent.py
```diff
@@ -0,0 +1,254 @@
...
+) -> tuple[type[Agent], ...]:
+    """Normalize user-provided Mesa base classes for dynamic meta-agent classes."""
+    if mesa_agent_type is None:
+        return (Agent,)
+    if isinstance(mesa_agent_type, tuple):
+        return mesa_agent_type
+    return (mesa_agent_type,)
+
+
+def evaluate_combination(
```
tpike3: I think this would probably be cleaner in the meta_agents_api as an @staticmethod. This way all public functions are in the api

mesa/meta_agents/meta_agents_api.py
```diff
@@ -0,0 +1,366 @@
...
+        args = signal.additional_kwargs.get("args") or ()
+        if not args:
+            return
+        self.deactivate(args[0])
+
+    def _entity_id(self, entity: Hashable) -> Hashable:
+        """Return the backend identity for a live entity or hashable external id."""
+        return getattr(entity, "unique_id", entity)
+
+    def _live_entity_lookup(self) -> dict[Hashable, Any]:
```
tpike3, line 91: This is going to be pretty expensive to build for every meta-agent look up. Can you cache the lookup it and then only rebuild with an add or remove call? Then add a TODO to do future optimizations, with bulk adds and removes?

mesa/meta_agents/meta_agents_api.py
```diff
@@ -0,0 +1,366 @@
...
+
+from __future__ import annotations
+
+from collections import deque
+from collections.abc import Callable, Hashable, Iterable
+from dataclasses import dataclass
+from typing import Any
+
+from mesa.agent import Agent, AgentSet
+from mesa.experimental.mesa_signals import ModelSignals
```
tpike3: Would be better if we could not tie to an experimental feature.

mesa/meta_agents/meta_agents_api.py
```diff
@@ -0,0 +1,366 @@
...
+        )
+
+    def _detach_entity(self, entity: Hashable) -> MembershipView:
+        """Remove all incident memberships for one entity."""
+        snapshot = self.query_memberships(entity)
+        self.backend.remove_agent(entity)
+        self.backend.remove_group(entity)
+        return snapshot
+
+    def create(
```
tpike3, line 164: I think for this we need to be very explicit on what happens in create as there is a lot of options and people will intuitively apply their use case

Here is an example
```
 """
  Create a meta-agent group and record its memberships.
  When agents already belong to an existing group with the same `new_agent_class`,
    that group is reused and the agents are added to it. To force a new group, use a unique class name (e.g., append a timestamp or UUID). For an example see the alliance_formation model.
    
    Args:
        new_agent_class: The group's class name. Groups with the same name are reused
            if any member already belongs to a group with that class.
        agents: Initial members.
        mesa_agent_type: Base class for the group agent.
        meta_attributes: Attributes to set on the group.
        meta_methods: Methods to bind to the group.
        relation: Membership relation label (default "member").
        memberships: Alternative to `agents`; list of (member, relation) tuples.
    
    Returns:
        The group agent (new or existing).
    
    Examples:
        Create a new team with two members:
        >>> team = model.meta_agents.create("Team", [alice, bob], Agent)
        
        Add carol to the same team (reuses the "Team" group):
        >>> team2 = model.meta_agents.create("Team", [carol], Agent)
        >>> assert team is team2 

         This is the same as using add carol using add_member 
(more explicit):
      >>> model.meta_agents.add_member("Team", carol)
      >>> assert carol in model.meta_agents.members_of("Team")
        
        Force a new group by using a unique name:
        >>> team_a = model.meta_agents.create("Team_2026_A", [...], Agent)
        >>> team_b = model.meta_agents.create("Team_2026_B", [...], Agent)
    """
```

mesa/meta_agents/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+
+from .meta_agent import evaluate_combination, find_combinations
+from .meta_agents_api import MembershipEdge, MembershipView, MetaAgents
+
+__all__ = [
+    "MembershipEdge",
+    "MembershipView",
+    "MetaAgents",
+    "evaluate_combination",
+    "find_combinations",
```
tpike3: Would change based on moving evaluate_combinations and move_combinations to meta-agents_api

mesa/meta_agents/meta_agents_api.py
tpike3: Hmmm thinking on this more, why are you going through every agent and not using the membership backend applied to whatever local action the user is trying to accomplish? 

A TODO would then be a staticmethod in the api to make a cache_build option so if they know their model is going to be constantly referencing the same sets of meta agents they can cache it so it doesn't have to constantly be rebuilt. However this would be more involved so is definitely a future iteration.

What do you think?

tpike3 (changes requested):
@falloficaruss This is really some great work! The membership backend of storing the triplets, ensuring there can only be one, and the assert invariance makes meta-agents much better. 


I think the biggest challenge is the lookup bottleneck in the meta_agents_api. So take a look at that, but unless I am missing something I don't think it is leveraging membership backend like it could be. 

Overall though great work in moving meta-agents forward.

mesa/meta_agents/meta_agent.py
falloficaruss: makes sense

falloficaruss: I have also removed model param from both the functions since we no longer need them in new meta agents

mesa/meta_agents/meta_agents_api.py
falloficaruss: the backend already scopes which ids each operation resolves. `members_of` only resolves ids from `backend.agents_of(group)`, etc. The real problem is mapping those ids back to live objects. Mesa core has no unique_id -> agent index (`model.agents` is a flat set). So some lookup structure is unavoidable.

falloficaruss: But we can fix this by maintaining it incrementally instead of rebuilding it per call.

falloficaruss:
> I think the biggest challenge is the lookup bottleneck in the meta_agents_api. So take a look at that, but unless I am missing something I don't think it is leveraging membership backend like it could be.

I have an initial idea of how we could handle this but would need some more thought. We should do this after this PR.

mesa/meta_agents/meta_agents_api.py
```diff
@@ -0,0 +1,495 @@
...
+        self,
+        new_agent_class: str,
+        agents: Iterable[Any],
+        mesa_agent_type: type[Agent] | None,
```
tpike3: I would make this optional too so it defaults to Agent so 

 mesa_agent_type: type[Agent] | None = None

tpike3 (approved):
Excellent work for GSoC!!!

Issue #3816, opened 2026-08-25
AgentSet.set should support assigning per-agent values, mirroring get


AshNicolus:
**What's the problem this feature will solve?**

AgentSet.get already retrieves values per agent (and supports multiple attributes
at once), but AgentSet.set can only broadcast a single value to every agent.
There's no first-class way to assign a different value to each agent from a
sequence. The common pattern of reading an attribute, computing a new value per
agent (often vectorized with numpy), and writing it back currently needs a
manual loop:

    new_energies = [e * decay for e in model.agents.get("energy")]
    for agent, e in zip(model.agents, new_energies):
        agent.energy = e

It's an asymmetry with get, and the scatter-back step is common enough in models
to be worth a first-class method.

**Describe the solution you'd like**

Let set assign per-agent values from a sequence, as the natural mirror of get:

    model.agents.set("energy", new_energies, per_agent=True)

Because get and set iterate the AgentSet in the same order, a
get -> transform -> set round-trip stays consistent. A length mismatch between
the sequence and the AgentSet would raise a ValueError. The default stays
broadcast, so this is fully backward-compatible.

Open question on the API: a per_agent flag on set (as above), or a separate
method like set_each / scatter? I lean toward the flag since it keeps one obvious
entry point next to get, but I'm happy to go either way.

**Are you planning to open a PR for this?**

Yes. I have a working prototype and will open a PR with tests once the API
direction is approved here.

**Additional context**

Relates to the get / set / agg attribute-access methods on AbstractAgentSet. The
main thing to settle is the API shape (flag vs separate method); the
implementation itself is small.

quaquel:
do we need the `per_agent` keyword argument. In principle, the right behavior can be derived from the value that is passed. I realize this can be tricky. What if you want to assign the same string to all agents, or what if the length of a sequence is identical to the number of agents? So I can see value in being explicit. 

Concept wise, I agree that adding something like this makes good sense. I am inclined to add the behavior to `set` rather than adding yet another method to `AgentSet`. 

Since we model AgentSet on pandas/polars data frames. How is this kind of stuff handled there?

AshNicolus:
Good question. I checked the data-frame libraries, and mesa already has a precedent internally. How each handles `col = value`:

| value passed | pandas | polars | mesa `create_agents` | proposed `set` |
|---|---|---|---|---|
| scalar (int/float/bool) | broadcast | broadcast (`pl.lit`) | broadcast | broadcast |
[...]

PR #3817, opened 2026-08-25
Add element-wise assignment to AgentSet.set (derived from value)

AshNicolus:
Implements element-wise assignment on `AgentSet.set`, closing #3816. Behavior is derived from the value (no flag), following the value-based rule pandas uses:

- A `list` / `tuple` / `numpy.ndarray` / `pandas.Series` is assigned element-wise and must have the same length as the AgentSet; a length mismatch raises `ValueError`.
- Any other value (a scalar, a string, or a non-sequence) is broadcast to every agent.

The value-resolution logic is extracted into a shared private helper (`_resolve_per_agent_values`) with a `strict` flag. `set` uses `strict=True` (raise on mismatch). `create_agents` is refactored onto the same helper with `strict=False`, so its behavior is unchanged (a mismatched-length sequence is still broadcast).

Behavior change (documented in the docstring, release-note-worthy):
- `set("attr", seq)` where `len(seq) == len(agentset)` now assigns element-wise instead of broadcasting the sequence to every agent.

On `create_agents`: I first switched it to also raise on a length mismatch (per the discussion on #3816), but that breaks the VirusAntibody example in mesa-examples, which passes a length-3 vector to broadcast across 20 agents. Since aligning it needs a coordinated mesa-examples update, I kept `create_agents`'s current broadcast behavior here (via `strict=False`) and left making it raise as a follow-up. The helper's `strict` flag makes that a one-line change once mesa-examples is updated.

Tests cover element-wise assignment (list/tuple/ndarray), the get/set round-trip, scalar and string broadcast, and the mismatch `ValueError` for `set`. Verified locally: full suite passes (761), ruff check and format clean.


AshNicolus:
Quick heads-up on a change from what we discussed: the examples CI caught that flipping `create_agents` to raise on a length mismatch breaks the VirusAntibody example in mesa-examples, which passes a length-3 vector to broadcast across 20 agents. So I kept `create_agents` broadcasting for now (it's refactored onto the shared helper with `strict=False`, behavior unchanged) and scoped the raise change to `set` only, which is new and has no downstream users.

Making `create_agents` also raise is a one-line flip on the helper (`strict=True`), but it needs a matching mesa-examples update to land without breaking CI, so I'd do that as a separate follow-up. CI is green now.

quaquel:
> Making create_agents also raise is a one-line flip on the helper (strict=True), but it needs a matching mesa-examples update to land without breaking CI, so I'd do that as a separate follow-up. CI is green now.

Breaking mesa-examples is fine in principle. The typical work-flow is to break it here, fix it next there. 

I am also open to merging this, but would love to have clear plan for next steps.

AshNicolus:
Thanks. Here's the plan I'd suggest:

1. This PR: `set` raises on a length mismatch, `create_agents` unchanged (still broadcasts, via `strict=False` on the shared helper). Green and ready to merge as-is, and it closes #3816.

2. Follow-up in core: flip `create_agents` to `strict=True` so it also raises on a mismatch and the two line up. One-line change now that the helper is in place.

3. Paired mesa-examples fix: update VirusAntibody to broadcast its length-3 vector explicitly (e.g. `[vec] * n`) so it works under the stricter `create_agents`.

If you'd rather not do two rounds, I can fold step 2 into this PR now and open the mesa-examples fix right after, so the break-here-fix-there happens back to back. Either sequencing works for me, just say which you prefer.

quaquel:
I'll merge this and you can cary on with 2 and 3.

quaquel (approved):

PR #3819, opened 2026-08-26
Fix signal docstring examples and add doctest coverage

falloficaruss:
Follow-up to #3818.

## Summary

- Fix signal docstring examples to import `ObservableSignals`.
- Replace weakly referenced inline lambdas with named handlers.
- Define handlers for the string-based compatibility examples.
- Add isolated doctest coverage for the signal type docstrings.

### DIFF
--- a/tests/experimental/test_mesa_signals.py
+++ b/tests/experimental/test_mesa_signals.py
@@ -1,5 +1,6 @@
 """Tests for mesa_signals."""
 
+import doctest
 from unittest.mock import Mock, patch
 
 import pytest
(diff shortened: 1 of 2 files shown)


tests/experimental/test_mesa_signals.py
```diff
@@ -15,10 +16,17 @@
...
 )
 from mesa.experimental.mesa_signals.signals_util import Message, _AllSentinel
 
 
+def test_signal_type_docstrings():
+    """Execute signal type examples in an isolated namespace."""
+    result = doctest.testmod(signal_types, globs={})
+    assert result.failed == 0
+
+
```
quaquel, line 29: this is fine for now, but we might want to handle this for the entire repo.

quaquel (approved):

PR #3822, opened 2026-08-28
fix(discrete_space): guard cell removal in FixedAgent.remove() when cell is None

### LINKED ISSUE #3748: FixedAgent.remove() raises AttributeError when the agent has no cell

sundeep8967:
## Description
`FixedAgent.remove()` raised an `AttributeError` when the agent had not been assigned to a cell (`self.cell is None`), because it called `self.cell.remove_agent(self)` unconditionally. This caused `Model.remove_all_agents()` to crash whenever unplaced `FixedAgent` instances existed.

## Solution
Added a guard in `FixedAgent.remove()` to check if `self._mesa_cell is not None` before invoking `remove_agent()`, mirroring `CellAgent.remove()`. Added a unit test verifying clean removal of unplaced `FixedAgent` instances.

Fixes #3748

### DIFF
--- a/mesa/discrete_space/cell_agent.py
+++ b/mesa/discrete_space/cell_agent.py
@@ -115,8 +115,9 @@ class FixedAgent(Agent, FixedCell):
         """Remove the agent from the model."""
         super().remove()
 
-        self.cell.remove_agent(self)
-        self._mesa_cell = None
+        if self._mesa_cell is not None:
+            self._mesa_cell.remove_agent(self)
+            self._mesa_cell = None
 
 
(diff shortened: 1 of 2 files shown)


quaquel:
see #3794, closing this.

PR #3824, opened 2026-08-30
Make create_agents raise on length-mismatched sequences

AshNicolus:
Follow-up to #3817 (step 2 of the plan there). Switches `create_agents` onto the shared `_resolve_per_agent_values` helper with `strict=True`, so it now raises `ValueError` on a length-mismatched sequence argument instead of silently broadcasting the whole sequence to every agent. This aligns `create_agents` with `AgentSet.set`, which already raises.

Behavior change (documented in the docstring, release-note-worthy):
- `create_agents(model, n, seq)` where `len(seq) != n` now raises `ValueError` instead of broadcasting `seq` to every agent.

As expected from the discussion on #3817, this breaks the VirusAntibody example in mesa-examples (it passes a length-3 vector to broadcast across 20 agents). Per the break-here-fix-there workflow, a matching mesa-examples PR follows to fix it (broadcasting the vector explicitly as `[vec] * n`, which works under both the old and new behavior).

Verified locally: core suite passes, ruff check and format clean. The `build / examples` check will be red until the mesa-examples fix lands.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -129,17 +129,19 @@ class Agent[M: Model]:
             AgentSet containing the agents created.
 
         Warning:
-            A list, tuple, ndarray, or pandas Series argument whose length
-            equals n is always treated as one value per agent, even if you
-            intended it as a single shared value. This is especially easy
-            to hit with coordinate tuples: create_agents(model, 2, pos=(10,
-            20)) does NOT give both agents pos=(10, 20); it gives agent 0
-            pos=10 and agent 1 pos=20, since the tuple's length (2) matches
-            n (2).
+            A list, tuple, ndarray, or pandas Series argument is treated as one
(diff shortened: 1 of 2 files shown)


AshNicolus:
For the mesa-examples side (step 3): the only thing that breaks is VirusAntibody, which passes a single length-3 `dna` list to `VirusAgent.create_agents` for `initial_viruses` (~20) agents, relying on the old broadcast behavior.

The fix is a one-liner in its `model.py`, broadcasting the shared value explicitly:

```python
dna=[dna] * self.initial_viruses
```

That keeps the existing behavior (every virus starts with the same dna) and works under both the old and new `create_agents`, so it can merge independently. I'll put that up as a matching mesa-examples PR to turn the examples check green.

quaquel (approved):

PR #3828, opened 2026-09-01
performance(meta_agents): cache live entity lookup using model lifecycle hooks

shipitdev:
> [!NOTE]
> Opened as a **Draft PR** for maintainer review alongside the proposal in Discussions/Issues.

### Pre-PR Checklist
- [x] I linked an issue/discussion where a maintainer approved this enhancement/feature for implementation.

### Approval Link
Addressing in-code `TODO(perf)` in `mesa/meta_agents/meta_agents_api.py:93` following the GSoC Meta-Agents rewrite in #3811.

---

### Summary
Implements an event-driven live entity cache in `mesa.meta_agents.MetaAgents` using model lifecycle hooks (`_register_agent_added_hook` and `_register_agent_removed_hook`). This avoids rebuilding an $O(N)$ lookup dictionary over every model agent on each meta-agent query.

---

### Motive & Benchmarks
Previously, `MetaAgents._live_entity_lookup()` iterated over `self.model.agents` to construct a new ID-to-agent dictionary on every call. High-level methods like `members_of()`, `groups_of()`, and `_resolve_view()` called this repeatedly within a single query:

* **Before (Uncached):** 1,000 `members_of()` queries across 5,000 agents took **~0.39s**.
* **After (Cached):** 1,000 `members_of()` queries across 5,000 agents took **~0.001s** (a **300x+ speedup**).

---

### Implementation Details
1. **`mesa/model.py`:**
   * Added `_agent_added_hooks` and `Model._register_agent_added_hook(hook)` to pair with `_register_agent_removed_hook`.
   * `register_agent()` invokes `_agent_added_hooks` synchronously upon registration.

2. **`mesa/meta_agents/meta_agents_api.py`:**
   * Initialized `self._live_entity_cache` in `MetaAgents.__init__`.
   * Subscribed to `_on_agent_added` and `_on_agent_removed` hooks to keep the cache in sync as agents enter or leave the model.
   * `_live_entity_lookup()` returns `self._live_entity_cache` directly in $O(1)$ time.

---

### Tests
* Added `test_agent_added_hook` in `tests/test_model.py`.
* Added `test_live_entity_cache_lifecycle_synchronization` in `tests/test_meta_agents_api.py`.
* Full test suite passing locally (`107/107 passed`), `ruff check` clean.

### DIFF
--- a/mesa/meta_agents/meta_agents_api.py
+++ b/mesa/meta_agents/meta_agents_api.py
@@ -78,10 +78,28 @@ class MetaAgents:
         self.backend = backend or MembershipBackend()
         model.meta_agents = self
 
-        self.model._register_agent_removed_hook(self._on_agent_removed)
+        self._live_entity_cache: dict[Hashable, Any] = {}
+        for entity in self.model.agents:
+            entity_id = getattr(entity, "unique_id", None)
+            if entity_id is not None:
+                self._live_entity_cache[entity_id] = entity
 
-    def _on_agent_removed(self, agent) -> None:
-        """Deactivate memberships when a live agent leaves the model."""
+        if hasattr(self.model, "_register_agent_added_hook"):
+            self.model._register_agent_added_hook(self._on_agent_added)
+        if hasattr(self.model, "_register_agent_removed_hook"):
+            self.model._register_agent_removed_hook(self._on_agent_removed)
+
+    def _on_agent_added(self, agent: Any) -> None:
+        """Cache live agent object when added to the model."""
+        entity_id = getattr(agent, "unique_id", None)
+        if entity_id is not None:
+            self._live_entity_cache[entity_id] = agent
+
+    def _on_agent_removed(self, agent: Any) -> None:
+        """Deactivate memberships and purge cache when a live agent leaves the model."""
+        entity_id = getattr(agent, "unique_id", None)
+        if entity_id is not None:
+            self._live_entity_cache.pop(entity_id, None)
         self.deactivate(agent)
 
     def _entity_id(self, entity: Hashable) -> Hashable:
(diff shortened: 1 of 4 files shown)


tpike3:
Thanks @shipitdev! 

After a quick glance, just one thing to consider as it is still in draft is how the cache relates to the membership backend. With the cache I feel like we are creating a second record keeping mechanism, which is not a good idea. However the tracking of the relationships is the hard part of meta-agents. So let me know what I am missing.

@falloficaruss please also take a look.

shipitdev:
> Thanks @shipitdev!
> 
> After a quick glance, just one thing to consider as it is still in draft is how the cache relates to the membership backend. With the cache I feel like we are creating a second record keeping mechanism, which is not a good idea. However the tracking of the relationships is the hard part of meta-agents. So let me know what I am missing.
> 
> @falloficaruss please also take a look.

Hi @tpike3 , thanks for reviewing this. I had already considered this. 

actually the cache isn't tracking any relationships i know its the backend job and its entirely there. Here's with the cache its just a simple ID to object kind of phone book. so when the backend returns "agent 42 and 55 are in this group", mesa needs to find the actual python objects for those ids previously, `_live_entity_lookup()` searched through every agent in the model to do that, and it did it on every single query. 

The cache now just keeps the phone book ready and updates it automatically when agents are added or removed so no relationship data is duplicated :)

falloficaruss:
@tpike3 This cache is not a second record-keeping mechanism. The two structures hold different data.

The cache is just a _unique_id -> live object_ index over model.agents (a "phone book" as @shipitdev said). This seems to somewhat closer to what was discussed in #3811 

For the bookkeeping concern I would suggest moving the index into the core: `Model` already eagerly maintains `_all_agents` and `_agents_by_type` in `register_agent`/`deregister_agent`, so an `_agents_by_id` dict updated in those same two spots would give every module an _O(1)_ id->agent lookup. `_live_entity_lookup()` would just return `model._agents_by_id`, the added hook could be dropped again, and downstream consumers (e.g., the warehouse robot->parent lookups in `mesa-examples`) get it for free

quaquel:
> For the bookkeeping concern I would suggest moving the index into the core: 

I am not in favor of this. Mesa-agents is still experimental. So, ideally it should not change anything in stable. For actions, we are also looking into how to more cleanly separate it again from the stable core.

tpike3:
Hmm @shipitdev  and @falloficaruss .. based on @quaquel comments can you give me 2 or so weeks to think on this. My priority right now is getting another meta-agents example (emergent firms) and getting the associated paper and presentation complete. That should also help me think through this

shipitdev:
Hi all, completely agree with @quaquel on not wanting to muddy with core on experimental features.

And @tpike3, please take all the time you need!

i see your immediate priority is the "emergent firms" example. I’ve been deep-diving into the new incidence-tensor architecture and the meta-agents refactor lately. If you need any help writing the code or setting up that example so you can focus on your paper, just let me know. I'd love to help out :)

tpike3:
@shipitdev  and @falloficaruss take a look at #3841. As I was doing emergent firms, the bottleneck in `_live_entity_lookup` was causing significant slowdowns. Those changes resulted in a significant speedup, and by caching in the API there is no reason to touch `model`. As meta-agents becomes more solidified then we can look at further optimization where it is completely interwoven in the base mesa modules (if needed).

No issues pivoting to this PR, the goal would be if you can find any other ways to optimize the functioning.

tpike3:
Closing this in favor of #3841

PR #3830, opened 2026-09-01
CI: Start testing Python 3.15

EwoutH:
Python has released it final 3.15 release candidate, 3.15.0rc2. This seems to be a good moment to see what breaks and what not. Most likely we will have some upstream dependencies who are not ready yet.

https://www.python.org/downloads/release/python-3150rc2/


EwoutH:
PyArrow doesn't have wheels on PyPI yet, see https://github.com/apache/arrow/issues/48172#issuecomment-5470323973.

PR #3833, opened 2026-09-04
Add Agent.on_idle wake dispatched as a deferred low-priority event

codebreaker32:
### Summary
Adds `Agent.on_idle(previous)`: called whenever the agent's action has ended — completion, interruption, cancellation, or failure — and nothing has replaced it. Dispatched as a zero-delay `Priority.LOW` event. Completes section 3 of #3798.

### Motive
After #3801 and #3805 an action can fail or be preempted, but nothing tells the agent afterwards; models must poll or schedule around it. `on_idle` is the "body became free" half of the wake contract: one place to chain the next action or resume an interrupted one. The default is a no-op, and nothing is resumed automatically.

### Implementation
- Every ending funnels through a new `Action._release_agent()`, which clears the agent's slot and schedules the wake. `Agent.remove()` bypasses it and cancels a pending wake — a removed agent never wakes.
- Dispatch is deferred rather than synchronous from `_do_complete`: a zero-duration action started from `on_idle` cannot recurse, and every agent finishing at time *t* decides after all completions at *t* have run.
- At most one wake per agent per model time; endings that pile up coalesce, with `previous` reporting the latest. A wake that finds the agent busy again (`interrupt_for` refilled the slot synchronously) is skipped.
- Eleven new lifecycle tests (`TestOnIdle`); `test_start_failure_schedules_nothing` sharpened into `test_start_failure_schedules_no_completion`.

### Usage Examples
```python
class Worker(Agent):
    def on_idle(self, previous):
        self.start_action(Task(self, duration=1.0))
```
Each completion wakes the worker once, after all other completions at that time, and the next task begins, no external driver.

### Additional Notes
Docs and the task-driven example are section 5 of #3798 and follow separately. `on_idle` sits on `Agent` next to the existing action cluster, pending the mixin question raised in #3805.

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -17,8 +17,10 @@ import pandas as pd
 if TYPE_CHECKING:
     from mesa.experimental.actions import Action
     from mesa.model import Model
+    from mesa.time import Event
 
 from mesa.agentset import AgentSet, _resolve_per_agent_values
+from mesa.time import Priority
 
 
 class Agent[M: Model]:
@@ -65,6 +67,9 @@ class Agent[M: Model]:
         self.model: M = model
         self.unique_id = None
         self.current_action: Action | None = None
+        self._wake_event: Event | None = None
+        self._wake_previous: Action | None = None
+        self._last_wake_time: float = float("-inf")
         self.model.register_agent(self)
 
         for dataset in self._datasets:
@@ -77,7 +82,8 @@ class Agent[M: Model]:
         scheduled completion event is cancelled silently. The action's
         on_interrupt() callback is NOT fired, because the agent is being
         destroyed — not making a behavioral decision. The action moves
-        to no defined end state; it is simply abandoned.
+        to no defined end state; it is simply abandoned. A pending
+        on_idle wake is cancelled as well: a removed agent never wakes.
 
         If your action holds external resources (e.g., a Resource slot,
         a reservation, a lock), override Agent.remove() and call
@@ -97,6 +103,11 @@ class Agent[M: Model]:
         if self.current_action is not None:
             self.current_action._cancel_event()  # Silent cleanup, no callback
             self.current_action = None
+
+        if self._wake_event is not None:
+            self._wake_event.cancel()
+            self._wake_event = None
+            self._wake_previous = None
 
         with contextlib.suppress(KeyError):
             self.model.deregister_agent(self)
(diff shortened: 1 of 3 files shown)


codebreaker32:
tests are failing, see https://github.com/widgetti/solara/pull/1199

quaquel (approved):
This looks fine to me. Just one clarifying question: why schedule the wake instead of calling it synchronously?

codebreaker32:
Three failure modes pushed me to defer it:

1. **Zero-duration actions recurse.** A duration-0 action completes inside `start()`. If `_do_complete` called `on_idle` synchronously, an agent that chains another duration-0 action from `on_idle` would nest `start() → _do_complete() → on_idle() → start() → …` without bound. Deferred, the once-per-time guard collapses the chain to a single wake per `model.time` (`test_zero_duration_chain_wakes_once_per_time` covers this — it hangs rather than fails if the guard breaks).

2. **`interrupt_for()` would break.** Interruption releases the slot mid-call: synchronous dispatch would run `on_idle` between `current.interrupt()` and `start_action(incoming)`, and if `on_idle` starts anything there, the incoming action then raises "already performing". Deferring, plus skipping the wake when the slot is already refilled by the time it fires, keeps `interrupt_for` atomic (`test_no_wake_when_interrupt_for_refills_the_slot`).

3. **Ordering at a timestamp.** As a `Priority.LOW` event, every agent finishing at time *t* decides after *all* completions at *t*. Synchronously, decisions would interleave with completions in heap-insertion order, so agents would decide against a half-updated world (`test_wake_runs_after_all_completions_at_that_time`).

jackiekazil:
Agents won't get a second wake at the same time step with a "once per time" guard on. This makes sense, but will this fail silently?

Maybe a docstring for hint to users?
Add it to the debugger?
Thoughts?

jackiekazil (approved):

jackiekazil (approved):
Looks good to me.

codebreaker32:
The `on_idle` docstring already carries the hint:

> At most one wake fires per agent per model time, so a zero-duration action started here cannot re-trigger the hook in the same instant.

The suppression does have to be quiet: the only agent that hits it is one chaining zero-duration actions within a single instant, and the alternative to stopping quietly is spinning forever without time advancing.

A debug aid is cheap, though, I can emit a `DEBUG`-level log line whenever the guard suppresses a wake, either in this PR or with the docs one. Preference?

jackiekazil:
Ty for the docstring. I missed it. 
I think in the PR is better.

jackiekazil:
LGTM. Thank you!

PR #3835, opened 2026-09-05
fix(visualization): reject positional-only parameters in _check_model_params

### LINKED ISSUE #2494: SolaraViz docs

sundeep8967:
## Description
Adds explicit validation for `inspect.Parameter.POSITIONAL_ONLY` in `_check_model_params` alongside `VAR_POSITIONAL`.

## Motivation and Context
Fixes #2494

SolaraViz instantiates models internally using `type(model.value)(**kwargs)`. If a model constructor defines positional-only parameters (e.g., `def __init__(self, width, /, height):`), attempting to instantiate it with keyword arguments raises a runtime `TypeError: __init__() got some positional-only arguments passed as keyword arguments`.

This change ensures `_check_model_params` catches positional-only parameters early during parameter validation and raises a descriptive `ValueError` explaining the keyword-only requirement for Solara visualization.

## How Has This Been Tested?
Added unit tests in `tests/visualization/test_solara_viz.py` testing both positional-only (`/`) and variable-positional (`*args`) constructors.

### DIFF
--- a/mesa/visualization/solara_viz.py
+++ b/mesa/visualization/solara_viz.py
@@ -752,12 +752,12 @@ def _check_model_params(model_or_func, model_params):
 
     model_parameters = inspect.signature(init_func).parameters
 
-    has_var_positional = any(
-        param.kind == inspect.Parameter.VAR_POSITIONAL
+    has_positional_only = any(
+        param.kind in (inspect.Parameter.POSITIONAL_ONLY, inspect.Parameter.VAR_POSITIONAL)
         for param in model_parameters.values()
     )
 
-    if has_var_positional:
+    if has_positional_only:
         raise ValueError(
             "Mesa's visualization requires the use of keyword arguments to ensure the parameters are passed to Solara correctly. Please ensure all model parameters are of form param=value"
         )
--- a/mesa/visualization/solara_viz.py
+++ b/mesa/visualization/solara_viz.py
@@ -753,7 +753,8 @@ def _check_model_params(model_or_func, model_params):
     model_parameters = inspect.signature(init_func).parameters
 
     has_positional_only = any(
-        param.kind in (inspect.Parameter.POSITIONAL_ONLY, inspect.Parameter.VAR_POSITIONAL)
+        param.kind
+        in (inspect.Parameter.POSITIONAL_ONLY, inspect.Parameter.VAR_POSITIONAL)
         for param in model_parameters.values()
     )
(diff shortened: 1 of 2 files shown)


dietcoke-17 (changes requested):
Thanks for tackling this, @sundeep8967 — nice catch on #2494, and the fix itself is spot on.

**What's working well**
- The logic is correct: `POSITIONAL_ONLY` and `VAR_POSITIONAL` are the only two `inspect.Parameter` kinds that can't be passed as keyword arguments, so this fully closes the gap.
- Test coverage is thorough — separate cases for the `/`-syntax and `*args` forms, each checked against the right error message.
- Clear PR description with the linked issue and reasoning — makes this easy to review.

**To get this merge-ready, one change:**
1. `has_positional_only` is a bit misleading now that it also flags `VAR_POSITIONAL` — that's not actually "positional-only" in the `inspect` sense. Renaming it to something like `has_non_keyword_param` would keep future readers from misreading what the check does.

**Optional (not blocking):**
- Pulling `(inspect.Parameter.POSITIONAL_ONLY, inspect.Parameter.VAR_POSITIONAL)` into a small named constant would save the next person a lookup.
- Naming the offending parameter in the error message (e.g. "Parameter 'width' must be passed as a keyword argument") would make this easier to debug for anyone who hits it.

Once the rename's in, I'm good to approve. Nice work!

sundeep8967:
Thanks @dietcoke-17, updated as requested:
- Renamed `has_positional_only` to `has_non_keyword_param` to accurately reflect the check across both `POSITIONAL_ONLY` and `VAR_POSITIONAL`.
- Extracted `_NON_KEYWORD_PARAM_KINDS = (inspect.Parameter.POSITIONAL_ONLY, inspect.Parameter.VAR_POSITIONAL)` at the module level for clarity and reuse.

The remaining test failure in the `examples` job (`VirusAntibodyModel`) is an upstream issue in mesa-examples and unrelated to these visualization changes.

quaquel:
Code wise, this looks good to me. Can you check what's going on with the code coverage? Ideally, we don't want to loss coverage due to new PRs.

sundeep8967:
Added unit tests passing model instances to `_check_model_params` for both positional-only and var-positional cases so both branches in `_check_model_params` are covered.

tpike3 (approved):
LGTM thanks @sundeep8967

PR #3838, opened 2026-09-05
adding a small helper function to validate MetaAgent type in add&remove member funcs

shipitdev:
hey guys, 

was building some models with the new incidence-tensor backend and ran into a small ux trap. `add_member(group, member)` accepts two hashables, which makes it super easy to accidentally invert them (e.g. passing the team as the member). because the tensor just accepts the triplets, it fails silently and completely corrupts the model logic later down the line.

i just added a quick isinstance check in the facade so it throws a clear `TypeError` if you accidentally pass a base agent as the group. 

here is a min reproduction script showing how easy it is to flip them, and how it completely breaks the query logic later on:

### Reproduction script

```python
from mesa import Model, Agent
from mesa.meta_agents import MetaAgents

class SimpleModel(Model):
    def __init__(self):
        super().__init__()
        self.meta = MetaAgents(self)

m = SimpleModel()
team = m.meta.create(new_agent_class="Team", agents=[])
agent = Agent(m)

# inverted arguments
m.meta.add_member(agent, team) 

# Later in the simulation, we try to iterate over the team:
team_members = m.meta.members_of(team)
print(f"Team has {len(team_members)} members")

# The tensor silently corrupted the relationship:
agent_members = m.meta.members_of(agent)
print(f"Agent has {len(agent_members)} members: {agent_members}")
```

### Output before the Fix:
```python
Team has 0 members
Agent has 1 members: <AgentSet with 1 agents>

``` 
(The simulation happily continues running with empty teams, and individual agents thinking they "contain" a team).


mesa/meta_agents/meta_agents_api.py
```diff
@@ -278,6 +279,11 @@ def add_member(
         member = lookup.get(self._entity_id(member), member)
         group = self._resolve_group(group)
 
+        if not isinstance(group, MetaAgent):
+            raise TypeError(
+                f"Expected group to be a MetaAgent instance or valid group ID, but got {type(group).__name__}. "
+            )
+
```
falloficaruss: `remove_member()` has the same (group, member) argument contract, but it does not receive this validation and can still silently accept inverted arguments. Could we extract this into a small shared helper and call it after group resolution in both `add_member()` and `remove_member()`?
This also makes the `MetaAgent`-only group contract explicit in one place. Please add regression tests for inverted `add_member` and `remove_member`, while confirming that a live group object, its `unique_id`, and its name still work.

shipitdev:
Thank you for reviewing this @falloficaruss :)

i've added a small helper function and called them at both the sides as you had suggested. i've also added the regression tests and rebased and squashed everything into a single commit for an easy review

mesa/meta_agents/meta_agents_api.py
```diff
@@ -127,6 +128,13 @@ def _resolve_group(self, group: Hashable) -> Any:
             raise ValueError(f"Ambiguous group name {group!r}")
         return group
 
+    def _validate_meta_agent(self, group: Any) -> None:
+        """Validate that the resolved group is a MetaAgent instance."""
+        if not isinstance(group, MetaAgent):
+            raise TypeError(
+                f"Expected group to be a MetaAgent instance or valid group ID, but got {type(group).__name__}."
+            )
```
falloficaruss, line 144: `_resolve_group()` passes unrecognized non-string IDs through unchanged, so a group that used to be supplied as a raw hashable ID now raises `TypeError` instead of recording an edge. If "group must resolve to a live MetaAgent" is the intended contract, could we document it in the add_member/remove_member docstrings (and docs/apis/meta_agents.md), and raise `ValueError` when an ID simply fails to resolve. Consistent with the existing path reserving `TypeError` for wrong-type arguments? As written, the message says "or valid group ID" while an unresolved int can never reach the write. A "did you swap group and member?" hint would also make the targeted trap self-explanatory.

tests/test_meta_agents_api.py
```diff
@@ -166,6 +166,48 @@ def test_add_and_remove_member_by_unique_id():
...
+    model = Model()
+    meta_agents = MetaAgents(model)
+    alice = Agent(model)
+    group = meta_agents.create("Team", [])
+
+    with pytest.raises(TypeError, match="Expected group to be a MetaAgent instance"):
+        meta_agents.add_member(alice, group)
+
+    with pytest.raises(TypeError, match="Expected group to be a MetaAgent instance"):
+        meta_agents.remove_member(alice, group)
```
falloficaruss, line 186: This covers both methods well. Two small additions would lock the behavior in: assert that `backend.as_triplets()` is unchanged after each rejected call (guards the validate-before-write ordering), and add one case with a `MetaAgent` in the member position, to prove the guard only constrains the group argument.

shipitdev:
Thanks @falloficaruss, i've added the docs and swap hint, and unknown group IDs now raise ValueError in both methods. 

The tests check that rejected calls leave memberships unchanged and that a MetaAgent can be a member. members_of() keeps its existing behavior. Also rebased.

PR #3840, opened 2026-09-06
Fix Grid.create_property_layer() silently overwriting Cell attributes

### LINKED ISSUE #3839: Property layers can break
`Grid.create_property_layer()` / `Grid.add_property_layer()` are supposed to reject a
property-layer name that collides with an existing attribute, so that attaching a
layer can never silently break cell-level access. The collision check in
`Grid._attach_property_layer()` (`mesa/discrete_space/grid.py:184-202`) only checks
`type(self)` — the **Grid**'s own class — and `self.cell_klass.__slots__`. It never
checks the actual attributes defined on `self.cell_klass` (the `Cell`/`GridCell`
subclass), which is the class that gets mutated via `setattr(self.cell_klass, name, accessor)`
a few lines later.

As a result, creating a property layer named `is_empty`, `is_full`, `neighborhood`,
or `position` — all real `@property` accessors on `Cell` (`mesa/discrete_space/cell.py`)
— silently overwrites them with a static, array-backed accessor, with no error or
warning. Since `DiscreteSpace.empties` and `Grid.cells_with_capacity` depend on
`cell.is_empty` / `cell.is_full` being computed live from the cell's actual agent
list, this corrupts occupancy and capacity queries for the rest of the model's life.

**Expected behavior**

Calling `grid.create_property_layer("is_full", ...)` (or `is_empty`, `neighborhood`,
`position`) should raise a `ValueError`, exactly as it already does today for
Grid-level attribute names like `"capacity"` and `"width"`:

```python
    grid.create_property_layer("capacity", 1, dtype=int)
```

**To Reproduce**

```python
import mesa
from mesa.discrete_space import OrthogonalMooreGrid, CellAgent

class MyModel(mesa.Model):
    def __init__(self):
        super().__init__(rng=1)
        self.grid = OrthogonalMooreGrid((3, 3), torus=False, capacity=1, random=self.random)

m = MyModel()
print("empties before:", len(m.grid.empties))
m.grid.create_property_layer("is_empty", default_value=False, dtype=bool)
print("empties after:", len(m.grid.empties))

m2 = MyModel()
m2.grid.create_property_layer("is_full", default_value=False, dtype=bool)
c2 = m2.grid._cells[(1, 1)]
CellAgent(m2).cell = c2
print("is_full (should be True):", c2.is_full)
print("cells_with_capacity (should be 8):", len(list(m2.grid.cells_with_capacity)))
```

Output on current `main` (`e5c3b38d`):

```
empties before: 9
empties after: 0
is_full (should be True): False
cells_with_capacity (should be 8): 9
```

No exception is raised anywhere in this run — the corruption is silent.

Image

**Additional context**

- File/lines: `mesa/discrete_space/grid.py:184-202` (`Grid._attach_property_layer`).
- Existing tests (`tests/discrete_space/test_discrete_space.py:896-919`) only cover
  collisions with Grid-level attributes (`capacity`, `width`), not Cell-level
  accessors, so this gap has no test coverage today.
- Not limited to the four names above: any custom `cell_klass` with its own
  `@property`/method hits the same silent corruption if a property layer is later
  created with a matching name, since `cell_klass` is never consulted by the check.
- Related but distinct: #3072 ("Make `cell.empty` read only") is about direct
  assignment to `cell.empty`, not this name-collision path.
- Impact: any model using `grid.empties`, `select_random_empty_cell()`, or
  `cells_with_capacity` after creating a colliding property layer will get silently
  wrong occupancy/capacity results (agents placed over capacity, or a
  `LookupError` from `random.choice(list(self.empties))` on a grid that isn't
  actually full).

---

## Proposed fix

Extend the collision check in `Grid._attach_property_layer()` to also inspect
`self.cell_klass.__mro__`, not just `type(self)` (the Grid) and the cell's
`__slots__`. The one name that must stay allowed is `"empty"`, since
`Grid.__init__` itself installs a property layer under that name to back
`Cell.empty`:

```python
# mesa/discrete_space/grid.py, inside _attach_property_layer

cell_attrs = set(
    chain.from_iterable(c.__dict__ for c in self.cell_klass.__mro__)
)
if name != "empty" and name in cell_attrs:
    raise ValueError(
        f"property_layer '{name}' conflicts with an existing Cell attribute."
    )
```

This mirrors the existing `slots` check immediately below it and keeps the same
`ValueError` contract callers already rely on for `capacity`/`width` collisions.

### Suggested regression test

Add to `tests/discrete_space/test_discrete_space.py`, next to the existing
`capacity`/`width` conflict test:

```python
@pytest.mark.parametrize(
    "reserved", ["is_empty", "is_full", "neighborhood", "position"]
)
def test_property_layer_rejects_cell_attribute_collision(reserved):
    grid = OrthogonalMooreGrid((3, 3), torus=False, capacity=1, random=random.Random(1))
        grid.create_property_layer(reserved, default_value=False, dtype=bool)

    # occupancy must still be computed correctly after the rejection
    assert len(grid.empties) == 9
```

0xnithinmys:
## Summary
`Grid._attach_property_layer()` only checked name collisions against the Grid's own class and the Cell's `__slots__`, so creating a property layer named `is_empty`, `is_full`, `neighborhood`, or `position` silently overwrote those live `Cell` properties instead of raising `ValueError`. Since `DiscreteSpace.empties` and `Grid.cells_with_capacity` depend on them, this silently corrupted occupancy/capacity queries with no error.

Fixes #3839.

## Fix
Also reject names colliding with any attribute on `self.cell_klass.__mro__`, with an exception for `"empty"` (which `Grid.__init__` itself installs as a property layer over `Cell.empty`).

## Test plan
- Added `test_property_layer_rejects_cell_attribute_collision` (parametrized over `is_empty`/`is_full`/`neighborhood`/`position`) and `test_property_layer_empty_bootstrap_is_not_blocked`
- Full suite passes: 805 passed
- `ruff check` / `ruff format --check`: clean


dietcoke-17 (commented):
Nice fix — this closes a real gap (mesa#3839). `_attach_property_layer()` previously only checked new names against Grid's own attributes and `cell_klass.__slots__`. Since `is_empty`/`is_full` are `@property` methods living in the class `__dict__` rather than `__slots__`, naming a layer `is_empty` would silently shadow the live property with a static array — no error, just quietly wrong `DiscreteSpace.empties` / `Grid.cells_with_capacity` from then on. Walking `cell_klass.__mro__` and collecting `__dict__` keys correctly catches properties, plain attributes, and slots alike.

Tests are great: `test_property_layer_rejects_cell_attribute_collision` covers both property- and plain-attribute collisions and checks real post-rejection behavior (`grid.empties`, `cell.is_full`, `cells_with_capacity`), not just the exception. `test_property_layer_empty_bootstrap_is_not_blocked` is a good, minimal guard on the internal bootstrap.

mesa/discrete_space/grid.py
```diff
@@ -200,6 +200,14 @@ def _attach_property_layer(
             raise ValueError(
                 f"property_layer name '{name}' clashes with existing slot '{name}'."
             )
+
+        cell_attrs = set(
+            chain.from_iterable(c.__dict__ for c in self.cell_klass.__mro__)
```
dietcoke-17, line 201: naming the actual class (`self.cell_klass.__name__`) instead of the generic "Cell" would help when `cell_klass` is a custom subclass a few levels removed.

mesa/discrete_space/grid.py
```diff
@@ -200,6 +200,14 @@ def _attach_property_layer(
...
+        )
+        if name != "empty" and name in cell_attrs:
```
dietcoke-17, line 203: **`name != "empty"` exception** — this is doing real work (it's exempting the layer `Grid.__init__` installs over `Cell.empty`), but nothing in the code says so. Could you add a one-line comment there? It reads like a leftover special case otherwise, and it's exactly the kind of line a future cleanup deletes without realizing it breaks the bootstrap.

dietcoke-17 (changes requested):

mesa/discrete_space/grid.py
```diff
@@ -200,6 +200,17 @@ def _attach_property_layer(
             raise ValueError(
                 f"property_layer name '{name}' clashes with existing slot '{name}'."
             )
+
+        cell_attrs = set(
```
quaquel, line 200: we rely heavily on slots as well, should they not be included in this?

0xnithinmys: Slots are in __dict__, so this already covers them  removed the redundant __slots__ check.

quaquel (approved):

PR #3841, opened 2026-09-06
meta-agent optimization

tpike3:
### Summary
The meta-agent api module currently looks through all agents multiple times via `_live_entity_lookup` this creates a significant bottleneck. This PR focuses on using the membership store as the reference that the meta-agent-apis uses, creating significant speed up. Related to #3828 

### Implementation
Add `triplets_for` and `all_entity_id` to `backend` to provide more membership store reference options
Update to `meta-agents-api` to use a mixture of caching, membership store look ups and fallback lookup to `model` if needed

### DIFF
--- a/mesa/meta_agents/meta_agents_api.py
+++ b/mesa/meta_agents/meta_agents_api.py
@@ -138,7 +138,8 @@ class MetaAgents:
                 return entity
         warnings.warn(
             f"Entity with id {entity_id!r} not found among model agents. "
-            f"Returning the raw id as a fallback.",
+            f"Entity reference not found in model agents."
+            f"Please pass valid Agent objects or IDs.",
             UserWarning,
             stacklevel=2,
         )
--- a/mesa/meta_agents/meta_agents_api.py
+++ b/mesa/meta_agents/meta_agents_api.py
@@ -78,39 +78,80 @@ class MetaAgents:
         self.backend = backend or MembershipBackend()
         model.meta_agents = self
 
+        # Cached mapping from backend entity ids to live model objects.
+        # Populated incrementally by mutation methods; seeded once on first
+        # access via _ensure_cache() for pre-existing backend state.
+        self._id_to_entity: dict[Hashable, Any] = {}
+        self._cache_seeded: bool = False
+
         self.model._register_agent_removed_hook(self._on_agent_removed)
 
     def _on_agent_removed(self, agent) -> None:
         """Deactivate memberships when a live agent leaves the model."""
         self.deactivate(agent)
+        entity_id = self._entity_id(agent)
+        self._id_to_entity.pop(entity_id, None)
 
     def _entity_id(self, entity: Hashable) -> Hashable:
         """Return the backend identity for a live entity or hashable external id."""
         return getattr(entity, "unique_id", entity)
 
-    def _live_entity_lookup(self) -> dict[Hashable, Any]:
-        """Build a lookup from backend ids back to live model objects."""
-        # TODO(perf): This rebuilds an O(N) mapping over every model agent on
-        # each call. Cache the lookup instead and only rebuild it on agent add
-        # or remove calls (e.g. via the model's agent lifecycle hooks, seeding
-        # once in ``__init__``). Future optimization: bulk adds and removes.
-        lookup: dict[Hashable, Any] = {}
+    def _cache_entity(self, entity: Any) -> None:
+        """Store a live entity in the id-to-object cache."""
+        entity_id = getattr(entity, "unique_id", None)
+        if entity_id is not None:
+            self._id_to_entity[entity_id] = entity
+
+    def _ensure_cache(self) -> None:
+        """One-time seed of the cache for any pre-existing backend state.
+
+        Only scans the model agents whose ids are already known to the
+        backend, making this O(N_model) once rather than on every call.
+        After the seed, all future entries come from mutation methods.
+        """
+        if self._cache_seeded:
+            return
+        self._cache_seeded = True
+        needed = self.backend.all_entity_ids() - self._id_to_entity.keys()
+        if not needed:
+            return
         for entity in self.model.agents:
             entity_id = getattr(entity, "unique_id", None)
-            if entity_id is not None:
-                lookup[entity_id] = entity
-        return lookup
+            if entity_id is not None and entity_id in needed:
+                self._id_to_entity[entity_id] = entity
+                needed.discard(entity_id)
+                if not needed:
+                    break
+
+    def _resolve_id(self, entity_id: Hashable) -> Any:
+        """Look up a cached entity, falling back to a model scan on miss.
+
+        On a cache miss the model agents are scanned for the specific id.
+        The result is cached so subsequent lookups are O(1).
+        Returns the raw *entity_id* when no live object is found.
+        """
+        self._ensure_cache()
+        hit = self._id_to_entity.get(entity_id)
+        if hit is not None:
+            return hit
+        # Fallback: scan model agents for this specific id.
+        for entity in self.model.agents:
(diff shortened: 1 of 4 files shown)


shipitdev (commented):
this looks great, especially since we want to avoid touching the core `Model` right now. 

the `triplets_for` optimization is a huge win. i verified it locally with 10k agents and query times dropped from ~25ms to ~0.02ms. totally fixes the bottleneck.

although i did have one question since this is a smart workaround while the feature is experimental, what happens when it becomes stable? will we eventually just bake the lifecycle hooks directly into the core model so the cache updates itself automatically?

mesa/meta_agents/meta_agents_api.py
```diff
@@ -78,39 +78,80 @@ def __init__(self, model: Any, backend: MembershipBackend | None = None) -> None
...
+        hit = self._id_to_entity.get(entity_id)
+        if hit is not None:
+            return hit
+        # Fallback: scan model agents for this specific id.
+        for entity in self.model.agents:
+            eid = getattr(entity, "unique_id", None)
+            if eid == entity_id:
+                self._id_to_entity[eid] = entity
+                return entity
+        return entity_id
```
shipitdev: i noticed one small edge case while testing that a user might accidentally trigger

if someone checks for an ID that doesn't exist (like an external string ID), the cache misses and it scans all agents in the model. but because it doesn't cache that 'not found' result, querying it in a loop forces it to scan the whole model every single time.

we can fix this by just caching the negative result at the bottom of `_resolve_id`

```suggestion
        self._id_to_entity[entity_id] = entity_id
        return entity_id

Comment (shipitdev) [inline mesa/meta_agents/meta_agents_api.py:143]: we can fix this by just caching the negative result at the bottom of `_resolve_id`

```suggestion
        self._id_to_entity[entity_id] = entity_id
        return entity_id

**Review by shipitdev — COMMENTED** [review]
just one small nitpick, i noticed one small edge case while testing that a user might accidentally trigger

if someone checks for an ID that doesn't exist (like an external string ID), the cache misses and it scans all agents in the model. but because it doesn't cache that 'not found' result, querying it in a loop forces it to scan the whole model every single time.

File: mesa/meta_agents/meta_agents_api.py (hunk 1 above)
Comment (tpike3) [reply mesa/meta_agents/meta_agents_api.py:146]: Hmmm.. this may be more than a nitpick. The challenge with meta-agents is really keeping track of all the changes as efficiently as possible while the simulation changes, which essentially becomes a reference store (`backend`) and a look up and action process (`api`). If I am understanding correctly, in this case, we would be adding a non-standard reference to the cache because that is what the user is using, but not how meta-agents module track agents or their connections. Correct? 

Instead of adding more and more references, it may be better to add a warning, so users know they need to reference their agents differently. This also begs the larger question of how the cache should evolve over time and to your other comments, as meta-agents develops and mesa core (e.g. model and agent) stabilizes then we can figure out how to seamlessly integrate the membership store (`backend`) and reference (`model`?) , but this a ways off. 

Current, question is how do we handle these edge cases? I am concerned about adding whatever reference is used to the cache, because that would imply the user is not understanding how mesa/their code is working which can introduce other bugs or false interpretation of results. However, I could be misunderstanding, what are you thoughts?

Comment (shipitdev) [reply mesa/meta_agents/meta_agents_api.py:146]: Correct. your point is pretty solid. i dont think that you're misunderstanding at all. if we just blindly cache wahtever bad reference they throw at the api we're essentially just hiding their bugs and collecting garbage (currently with this change) i can see how that leads to a false interpretation of their results because they wouldn't even realize they're using mesa wrong. i agree on not collecting the garbage in our cache.

to answer your question on how we should handle these edge cases currently: a more robust approach would be to just raise a `UserWarning` (or even a `ValueError` if we want the api strictly enforced). if the code screams at them the second they pass a bad reference, it instantly tells them they're using the api incorrectly and naturally prevents them from writing loops that degrade performance. that way the user can know how to use our code efficiently.

for now, throwing a `UserWarning("Entity reference not found in model agents. Please ensure you are passing valid Agent objects or IDs.")` inside that fallback seems like a good idea to me as well.

if it helps, i could go through the user behavior and edge cases around the backend, note them down, and give some thought to the most efficient ways we could handle them. maybe we just strictly enforce these rules across the board if that aligns more with what you want, so we can eventually progress to stabilizing this later (even if it's a long way off like you said). what you think?

Comment (shipitdev) [reply mesa/meta_agents/meta_agents_api.py:146]: also worth mentioning this whole discussion actually aligns perfectly with my pr #3838. over there, accidentally inverting arguments (passing an agent as a group), and because the backend accepted any hashable, it failed silently and completely corrupted the model logic. 

there i added a quick `isinstance` check to throw a `TypeError` rather than just catching errors later, but the whole point is exactly the same: enforcing strict boundaries so users don't silently build on top of bad references and broken code.

**Comment by falloficaruss** [comment]
For now, I think the right tradeoff is to leave unresolved IDs uncached and keep this PR focused on the indexed backend queries and caching live entities reached through the MetaAgents API.

**Comment by tpike3** [comment]
Test failure unrelated to PR --- `FAILED test_examples.py::test_model_steps[VirusAntibodyModel] - ValueError: sequence of length 3 does not match the number of agents (20)`

File: mesa/meta_agents/meta_agents_api.py [hunk 2]
```diff
@@ -78,39 +79,82 @@ def __init__(self, model: Any, backend: MembershipBackend | None = None) -> None
...
+        hit = self._id_to_entity.get(entity_id)
+        if hit is not None:
+            return hit
+        # Fallback: scan model agents for this specific id.
+        for entity in self.model.agents:
+            eid = getattr(entity, "unique_id", None)
+            if eid == entity_id:
+                self._id_to_entity[eid] = entity
+                return entity
+        if entity_id in self.backend.all_entity_ids():
```
falloficaruss, line 139: It runs on every miss and `all_entity_ids()` rebuilds the full set each time, even when the id isn't known to the backend at all. would a simple `has_entity check` on the backend feel more natural here? Then `all_entity_ids()` could stay as the one-time seed in `_ensure_cache` and this check stays cheap. Curious about your thoughts

mesa/meta_agents/meta_agents_api.py
```diff
@@ -78,39 +79,82 @@ def __init__(self, model: Any, backend: MembershipBackend | None = None) -> None
...
+            warnings.warn(
+                f"Entity with id {entity_id!r} not found among model agents. "
+                f"Please pass valid Agent objects or IDs.",
+                UserWarning,
+                stacklevel=2,
+            )
```
falloficaruss, line 145: If someone does `add_member(group, "ext")`, "ext" gets recorded in the backend, and then the query_memberships call inside `add_member` comes back through here and warns even though external hashable IDs seem to be supported (`_entity_id` and `backend._to_id` both mention them). So a valid reference ends up sounding like a mistake. Should we only warn when an id that used to point at a live agent stops resolving, or is the intent that external IDs shouldn't go through the public API? Either way, a small test would pin the behavior as intentional. Let me know if I am missing anything.

tpike3:
So chewing on this, looking at the set up, also accounting for the goal to have actions separate from core. See @quaquel  comments in #3828. 

The real issue is meta-agents uses ids for management and through this has some additional functions so users can effectively name meta-agents and look them up or track them. However, Mesa core tracks by object (the pointer to the memory location).    

So I really see 2 paths forward - 
1 - use objects vs ids in the triplets in `backend`
2 - do a new pr to `model` that adds `agents_by_id` to model similar to `agents_by_type`

@quaquel what are your thoughts on option 2?

tpike3:
Closing this PR. Doing an update to meta-agent to use the object pointer instead of id. See #3875

PR #3845, opened 2026-09-08
Prevent zero weight agents from filling weighted samples

### LINKED ISSUE #3844: Weighted sampling without replacement can select agents with zero weight
**Describe the bug**

`AgentSet.select_random()` can select agents whose weight is zero when `replace=False` and the requested sample size exceeds the number of agents with positive weights. The weighted sampling algorithm assigns zero weight entries a key of zero, then includes those entries to fill the requested sample.

**Expected behavior**

An agent with zero weight should not be selected by weighted sampling. If the requested sample cannot be formed from agents with positive weights without replacement, the method should raise a clear `ValueError`.

**To Reproduce**

```python
from random import Random

from mesa import Agent, Model

model = Model()
agents = [Agent(model) for _ in range(3)]
agentset = AgentSet(agents, random=Random(7))

sampled = agentset.select_random(
    2,
    weights=[1.0, 0.0, 0.0],
    replace=False,
)

print([agents.index(agent) for agent in sampled])
```

This prints `[0, 1]`, so an agent with zero weight is included.

**Additional context**

The issue affects weighted sampling without replacement. Sampling with replacement already excludes zero weight entries because it uses `random.choices()`.

SahilKumar75:
> [!NOTE]
> Use this template for bug fixes only. For enhancements and new features, use the feature template and get maintainer approval in an issue or discussion before opening a PR.

### Pre PR Checklist

### Summary

Prevents weighted sampling without replacement from selecting agents whose weight is zero when too few agents have positive weights.

### Bug / Issue

Closes #3844.

When the requested sample size exceeded the number of positive weights, the sampling algorithm used zero weight entries to fill the result. Those agents should have no chance of selection.

### Implementation

Counts positive weights before weighted sampling without replacement. The method now raises a clear `ValueError` when the requested sample cannot be formed from agents with positive weights.

### Testing

Added a regression test using one positive weight and two zero weights. The test verifies that requesting two agents raises the expected error.

Validation completed:

* `python3 -m pytest tests/test_agentset.py -q`: 40 passed
* `ruff check mesa/agentset.py tests/test_agentset.py`: passed
* `ruff format --check mesa/agentset.py tests/test_agentset.py`: passed

### Additional Notes

Sampling with replacement is unchanged.


dietcoke-17 (changes requested):
Thanks for tracking this down and fixing it, @SahilKumar75 — nice work. Confirmed this is a real bug: zero-weight agents were getting a sort key of zero instead of being excluded, so they'd fill in when there weren't enough positive-weight agents to satisfy the sample size. Great Work and 

A few things I'd like addressed before merging:
1.Naming: positive_weights holds a count, not the weights themselves. Suggest positive_weight_count to avoid confusion.
2.A mixed case with some zero and some positive weights, where positive count is sufficient — confirms the zero-weight agent is still never selected and no false-positive error fires. Right now we only test the failure path.

mesa/agentset.py
```diff
@@ -260,6 +260,13 @@ def select_random(
             if replace:
                 chosen = self.random.choices(items, weights=w, k=sample_size)
             else:
+                positive_weights = sum(weight > 0 for weight in w)
```
shipitdev: we could just do it all in a single pass like this:

```python
# Efraimidis & Spirakis (A-Res) algorithm for weighted sampling without replacement
keys = []
for agent, wi in zip(items, w):
    if wi > 0:
        u = self.random.random()
        key = u ** (1.0 / wi)
        keys.append((key, agent))

if sample_size > len(keys):
    raise ValueError(
        f"Sample size ({sample_size}) cannot exceed the number of "
        f"agents with positive weights ({len(keys)}) when "
        "replace=False."
    )

keys.sort(key=lambda x: x[0], reverse=True)
chosen = [agent for _, agent in keys[:sample_size]]
```
that way we wouldn't need sum(), and keys.sort() only has to sort the agents that actually have a chance of being picked.

shipitdev (commented):
thank you @SahilKumar75 for reporting and raising a PR

looking at the current implementation, using sum() adds the overhead of iterating through weights list twice, and then still keeping zero-weight agents in keys which forces the keys.sort() to sort thru all the zero weight entries which i think are unnecessary. a more efficient and cleaner way would be to just remove the else block entirely since the zero-weight agents have a 0% chance of being selected anyway, we dont even need to store them in keys in the first place. i believe this was implemented in #3799 (cc @quaquel) i just want to get your thoughts on this.

mesa/agentset.py
shipitdev: @dietcoke-17 please let me know your verdict after reading this as well :) Thank you

SahilKumar75:
Thanks for the reviews. I updated the implementation to build sampling keys only for agents with positive weights, so validation and sorting now use the filtered list in one pass. The count is named positive_weight_count, and I added a mixed case regression test confirming that sampling succeeds when enough positive weights exist while the zero weight agent remains excluded. All 41 focused AgentSet tests pass, and the modified files pass Ruff lint and format checks.

quaquel (approved):

Issue #3836, opened 2026-09-05
Optimize random empty-cell selection for nearly full grids


Krishsharma179:
**Problem**

The current select_random_empty_cell() implementation makes up to 50 random attempts to find an empty cell. When the grid is mostly full, these attempts are likely to fail, after which np.argwhere() scans the entire grid to find empty cells.

This can be expensive for large, nearly-full grids.

current implemetation 
```
 if self._try_random:
            # Limit attempts to avoid infinite loops on full grids
            for _ in range(50):
                cell = random.choice(cells)
                if cell.is_empty:
                    return cell

        empty_coords = np.argwhere(self.property_layers["empty"])
```

**Solution**

Maintain a list of empty cells and update it whenever a cell is added or removed. This avoids scanning the entire grid each time an empty cell needs to be selected.

I have the implementation I will be sending a PR in a while

quaquel:
There is a long history on this issue so please check that first before opening any PR. 

In short, from memory, maintaining a list of empty cells creates a lot of overhead. This overhead is wasteful if the grid is largely empty. Only when the grid is nearly full does it start to make sense to identify all empty cells. In addition, there is still a plan to integrate this more cleanly with the empty property layer that already exists.

Krishsharma179:
@quaquel Agreed  maintaining the list would be wasteful if the grid is largely empty. My proposed approach is to just maintain a cheap counter of empty cells, and only start building/maintaining the full list once that counter drops below a threshold. 

The open question is: what should that threshold be?

BTW, I’ve opened a draft PR for this to get some early feedback on the approach.

PR #3846, opened 2026-09-09
enchanced the performance of random empty cell

Krishsharma179:
### Pre-PR Checklist
- [X] I linked an issue/discussion where a maintainer approved this enhancement/feature for implementation.

### Approval Link
#3836 

### Summary
Draft proposal for optimizing `Grid.select_random_empty_cell()` on dense grids, without adding always-on maintenance overhead addressing the concern  that a persistently maintained empty-cells structure would be wasteful on sparse grids.
### Motive
Currently, `select_random_empty_cell()` tries up to 50 random guesses, then falls back to `np.argwhere(self.property_layers["empty"])` a full O(n) scan of the grid every single call once guessing fails. On a nearly-full grid, this means repeated full scans on every call, which gets expensive as the grid approaches capacity. 
### Implementation
A lightweight `_empty_cell_count` integer is updated via the existing `property_layers["empty"]`  setter whenever a cell's emptiness changes O(1) per change, effectively free.
Once `_empty_cell_count` drops below a threshold (currently a placeholder not yet benchmarked, see Additional Notes), `_start_empty_cell_maintenance()` builds a `_empty_cells`  list + `_empty_cell_indices`  dict, enabling O(1) add/remove/random-pick from then on via swap-and-pop.
Below that threshold, no maintenance happens at all zero added overhead on the common case of a sparse grid, directly addressing the concern about wasted cost on largely-empty grids.

### Additional Notes
Opening as a draft, not merge-ready. The current `10%` threshold is a placeholder and hasn’t been empirically validated yet. I’m benchmarking across densities and grid sizes to find the actual break-even point. Opening early to get feedback on the approach before finalizing the value. @quaquel  flagging you since this addresses the original concern you raised.

### DIFF
--- a/mesa/discrete_space/grid.py
+++ b/mesa/discrete_space/grid.py
@@ -113,6 +113,10 @@ class Grid(DiscreteSpace[T]):
         # Track which property layers are read-only so the constraint survives
         # pickling and deepcopy (see __getstate__/__setstate__).
         self._read_only_layers: set[str] = set()
+        self._empty_cell_count = math.prod(self.dimensions)
+        self._empty_cells: list[T] = []
+        self._empty_cell_indices: dict[tuple[int, ...], int] = {}
+        self._maintain_empty_cells = False
 
         # we register the pickle_gridcell helper function
         copyreg.pickle(self.cell_klass, pickle_gridcell)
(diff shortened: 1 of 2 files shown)


quaquel:
As evidenced by the benchmarks, this direction is not resulting in a performance improvement.

codebreaker32 (commented):
Thanks for the PR but The threshold isn't the problem, the placement is (https://github.com/mesa/mesa/issues/3836#issuecomment-5599551107)

The hook sits in the generic property-layer setter, so every write to every
layer pays a string compare and every `empty` write also reads the array back.
That cost lands on the sparse grids the threshold was meant to protect.

Two structural issues too:

- `__setstate__` rebuilds accessors with a plain setter and no hook, while
  `_maintain_empty_cells` and the stale list restore as-is. After a deepcopy
  I filled every cell through the normal API and `select_random_empty_cell()`
  returned an occupied cell. No exception.
- `_maintain_empty_cells` is never set back to False. One dip below the
  threshold and the grid maintains forever.

In my opinion, this is an overkill. Also worth knowing #3340 stripped that setter down to a bare array write on purpose.

quaquel:
I am closing this PR and invite you to open a discussion. Here we can more freely explore design directions and assess performance implications. Once we agree in the discussion, a targeted PR can be opened that implements this.

PR #3848, opened 2026-09-10
Reject nonfinite weights before weighted sampling

### LINKED ISSUE #3847: Weighted sampling without replacement accepts nonfinite weights
**Describe the bug**

AgentSet.select_random() accepts NaN and positive infinity when sampling without replacement. These values do not define valid probabilities. NaN can remain in the generated sampling keys and affect which agent is selected, while positive infinity produces a key of one. The same inputs are rejected by the standard library path used when replacement is enabled.

**Expected behavior**

Weighted sampling should reject every nonfinite weight before selecting either sampling algorithm.

**To Reproduce**

Create two agents and call select_random(1, weights=[float("nan"), 1.0], replace=False). The call returns a sample instead of raising a validation error. Replacing NaN with positive infinity has the same problem.

This was reproduced against current main at commit a5dfcd6ad2bb7863a00d8735649d4c3ebaf014f8.

**Additional context**

A focused validation before the existing negative weight check makes replacement and nonreplacement behavior consistent. Regression coverage should include both NaN and positive infinity.

SahilKumar75:
> [!NOTE]
> Use this template for bug fixes only. For enhancements or new features, use the feature template and get maintainer approval in an issue or discussion before opening a PR.

### Pre PR Checklist

### Summary

Reject NaN and infinite weights before weighted sampling chooses an algorithm. Without this validation, sampling without replacement can silently select agents using invalid probability weights.

### Bug / Issue

Fixes #3847.

AgentSet.select_random() currently accepts NaN and positive infinity when replace=False, although the replacement path rejects them. This makes the result depend on invalid generated keys instead of a valid probability distribution.

### Implementation

Validate that every resolved weight is finite before checking negativity or total weight. The validation applies consistently to both replacement modes.

### Testing

Added parameterized regression coverage for NaN and positive infinity. All 41 AgentSet tests pass. Ruff formatting passes. Ruff lint passes apart from two existing UP038 findings in agentset.py that are also present on current main.

### Additional Notes

This validation is separate from PR #3845, which addresses zero weight sampling capacity.


mesa/agentset.py
```diff
@@ -61,6 +61,8 @@ def _resolve_weights(
             "Expected str, Callable, Sequence[float], or None."
         )
 
+    if any(not np.isfinite(x) for x in w):
```
quaquel: why use numpy here on each individual number. Might it not be faster (too be tested) to cast the weights to a numpy array and vectorize this check? Or alternatively compare with inf directly using python itself instead of numpy.

quaquel (changes requested):
minor question/change. Looks fine otherwise.

mesa/agentset.py
SahilKumar75: Switched to math.isfinite. Timed all three over 100 runs on a list of 100k floats: numpy called per element 8.2s, vectorized np.isfinite(np.asarray(w)).all() 0.34s, math.isfinite 0.15s. Since w is already a plain list, building the array costs more than the vectorized check saves, so plain Python wins at every size I tried. Pushed in b9ccd50.

quaquel (approved):

PR #3851, opened 2026-09-10
Move the action cluster from Agent to a HasActions mixin

codebreaker32:
### Summary
Moves the agent-side action API off core `Agent` into a `HasActions` mixin in `mesa.experimental.actions`. `Agent` shrinks by ~210 lines; models opt in with `class Wolf(HasActions, CellAgent)`. Public import path: `from mesa.experimental.actions import Action, HasActions`. Follows up on @quaquel's composition question in #3805, agreed there as a future PR and recorded as the open placement decision in #3798.

### Motive
Two problems with the status quo: every new hook grew the base `Agent` class ("the larger the base Agent class becomes, the steeper the learning curve"), and `start_action` & co. sat on *stable* `Agent` while `Action` itself is experimental. With the mixin, the whole cluster is experimental together, and models that don't use actions don't carry the API.

### Implementation
- `HasActions` lives in `actions.py` alongside `Action` and carries the cluster unchanged: `current_action`, `start_action`/`interrupt_for`/`cancel_action`, `should_interrupt`, `is_busy`, `on_idle` with the deferred-wake machinery (including the DEBUG log from #3833), and a `remove()` override that does action/wake cleanup before `super().remove()`. Cooperative `__init__`, mixin listed first.
- `Action.__init__` now rejects a mixin-less agent up front with an actionable `TypeError`, instead of failing later on a missing attribute.
- `Agent` keeps two deliberate traces: `"current_action"` stays in `_repr_excluded_fields` (a repr filter two existing tests rely on, not API surface).
- Tests run through `ActionAgent(HasActions, Agent)` as models will combine it; new tests pin the `TypeError` and composition with `CellAgent` on a real `OrthogonalMooreGrid`.

### Usage Examples
```python
from mesa.experimental.actions import Action, HasActions


class Graze(Action):
    def __init__(self, sheep):
        super().__init__(sheep, duration=5.0)

    def on_complete(self):
        self.agent.energy += 30


class Flee(Action):
    def __init__(self, sheep):
        super().__init__(sheep, duration=2.0, priority=10.0, interruptible=False)


class Sheep(HasActions, CellAgent):
    def on_idle(self, previous):
        if previous is not None and previous.is_resumable:
            self.start_action(previous)      # finish the interrupted graze
        else:
            self.start_action(Graze(self))


# A predator appears mid-graze: should_interrupt compares priorities,
# the graze yields with partial progress, and when the flee completes
# on_idle resumes the remainder.
sheep.interrupt_for(Flee(sheep))
```

### DIFF
--- a/mesa/agent.py
+++ b/mesa/agent.py
@@ -15,15 +15,9 @@ import numpy as np
 import pandas as pd
 
 if TYPE_CHECKING:
-    from mesa.experimental.actions import Action
     from mesa.model import Model
-    from mesa.time import Event
 
 from mesa.agentset import AgentSet, _resolve_per_agent_values
-from mesa.mesa_logging import create_module_logger
-from mesa.time import Priority
-
-_mesa_logger = create_module_logger()
 
 
 class Agent[M: Model]:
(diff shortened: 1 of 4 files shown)


quaquel (approved):

EwoutH:
Thanks for working on this.

We do now have the situation that agents have "opt-in" capability. Not sure that's the most intuitive concept. We should actively communicate this in our docs, release notes and tutorials I think.

Alternatives would be to have multiple Agents by default that have more capability. You could have:
- `Agent`
- `FullAgent`

With the latter having all (sensibel) opt-in capabilities.

Or the other way around:
- `BareAgent`
- `Agent`

codebreaker32:
Thanks for your thoughts on this. 

> We do now have the situation that agents have "opt-in" capability. Not sure that's the most intuitive concept

I think mesa already works this way, and actions are just a new capability following the existing pattern. Agents already pick their base class by what they need to do: `CellAgent`, `FixedAgent`, `ContinuousSpaceAgent`, `Grid2DMovingAgent`. And `CellAgent` is itself `class CellAgent(Agent, HasCell, BasicMovement)`, so it is already a composed class over two capability mixins. `HasActions` is the same idea rather than a new concept.

> Alternatives would be to have multiple Agents by default that have more capability. You could have:
> 
> * `Agent`
> * `FullAgent`
> 
> With the latter having all (sensibel) opt-in capabilities.

Worth considering, but I do not think the composition closes. All sensible capabilities cannot include space, because cell space and continuous space are mutually exclusive, an agent is one or the other. So `FullAgent` ends up meaning `Agent` + `HasActions`, and then someone who wants actions on a grid needs `FullCellAgent`, then `FullContinuousAgent`, and so on.

My one hesitation on `FullAgent` is that all this stuff is still experimental and may still move or be renamed, so it doubles the surface we change later.

quaquel:
As indicated in #3805, I believe it would be good to have a broader discussion on composition versus inheritance for agents inside MESA. For now, however, I am in favor of keeping experimental stuff out of the main agent classes.

PR #3853, opened 2026-09-12
Make select() raise on invalid float at_most instead of silently ignoring it

### LINKED ISSUE #3852: AgentSet.select(at_most=...) silently ignores NaN/negative/zero instead of raising
**`AbstractAgentSet.select()`'s `at_most` silently ignores invalid float values instead of raising.**

The fraction check is:

```python
if at_most <= 1.0 and isinstance(at_most, float):
    at_most = int(len(self) * at_most)
```

For `at_most=float('nan')`, `nan <= 1.0` is `False`, so this branch is skipped and the raw NaN is passed straight into the generator, where it's compared as `count >= at_most`. That comparison is always `False` for NaN, so the loop never breaks and every agent is selected, i.e. `at_most=nan` behaves like no limit at all instead of raising.

Repro:

```python
from random import Random

class Dummy: pass

aset = AgentSet([Dummy() for _ in range(10)], Random(0))
print(len(aset.select(at_most=float('nan'))))  # prints 10, not an error
```

Negative floats and `0.0` have a related but milder issue: they also skip the branch (`-0.5 <= 1.0` is actually `True` so they do get floored, e.g. `int(10 * -0.5) == -5`, and the generator's `count >= at_most` is then always `True` from the first agent, so it silently returns 0 selected instead of raising).

`select_random`'s `n` parameter already validates this exact range (`0.0 < n <= 1.0`) and raises `ValueError` otherwise, so `select`'s `at_most` doesn't follow the same convention as its sibling method.

I'll open a PR with a fix (validate the range up front, raise `ValueError` for anything outside `(0.0, 1.0]` including NaN).

RudraDudhat2509:
Fixes #3852

`at_most <= 1.0 and isinstance(at_most, float)` skips the fraction branch whenever the comparison is False, which is also what NaN does (every NaN comparison is False). So `at_most=float('nan')` fell through untouched, and the generator's `count >= at_most` check is then always False for NaN, so the loop never breaks and every agent gets selected instead of none. Negative floats and `0.0` had the same fallthrough problem, just landing on 0 selected instead of unlimited.

Fix validates the `(0.0, 1.0]` range up front and raises `ValueError` for anything outside it, matching how `select_random`'s `n` parameter already validates the same range on its own float path.

Verified against the pre-fix commit: `at_most=float('nan')` returned 10/10 agents before the fix, raises `ValueError` after. Ran the full `test_agentset.py`, 39 passed. Added 4 cases (nan, 0.0, -0.5, 1.5) to the existing `test_agentset` test. ruff check + format clean.


quaquel:
It's a very implausible bug that is fixed here but it might crop up in rare cases. Fix seems fine.

quaquel (approved):

PR #3712, opened 2026-05-14
Fix partial network edgecolors in matplotlib visualization

pragnyanramtha:
> [!NOTE]
> Use this template for bug fixes only. For enhancements/new features, use the feature template and get maintainer approval in an issue/discussion before opening a PR.

### Pre-PR Checklist
- [x] This PR is a bug fix, not a new feature or enhancement.

### Summary
Fixes a matplotlib network visualization crash when only some agent portrayals include `edgecolors`.

### Bug / Issue
Fixes #2691.

`draw_network` expects per-agent style arrays to stay aligned. When an agent portrayal omitted `edgecolors`, the collectors skipped that entry, so the edgecolor array could be shorter than the node collection. Mixed string and RGBA tuple edgecolor values also need a one-dimensional object array so Matplotlib can consume them consistently.

### Implementation
- Collect one `edgecolors` entry per agent whenever any agent supplies edgecolors.
- Use Matplotlib `"none"` for agents that omit edgecolors.
- Preserve normal NumPy conversion where possible, and fall back to a one-dimensional object array for mixed color specifications such as RGBA tuples plus `"none"`.
- Apply the same behavior to both `draw_network` and `MatplotlibBackend.collect_agent_data`.
- Add regression tests for partial edgecolors with string colors and RGBA tuple colors.

### Testing
- `uv run --extra viz --extra network --with pytest pytest -p no:cacheprovider tests/visualization/test_components_matplotlib.py::test_draw_network_with_partial_edgecolors tests/visualization/test_backends.py::test_matplotlib_backend_collects_partial_edgecolors -q` -> 4 passed
- `uv run --extra viz --extra network --with pytest pytest -p no:cacheprovider tests/visualization/test_components_matplotlib.py tests/visualization/test_backends.py -q` -> 30 passed
- `uv run --extra viz --extra network --with pytest --with ruff ruff check --no-cache mesa/visualization/mpl_space_drawing.py mesa/visualization/backends/matplotlib_backend.py tests/visualization/test_components_matplotlib.py tests/visualization/test_backends.py` -> passed
- `git diff --check` -> passed

No screenshots are attached because the regression was an exception before rendering completed; the automated Matplotlib tests cover the before/after failure mode.

### Additional Notes
This is scoped to matplotlib visualization style collection. It does not change public APIs, dependencies, schemas, or CI configuration.

PR #3856, opened 2026-09-13
Fix network crash error

### LINKED ISSUE #2691: Network Visualization error when Agent Portrayal has inconsistent fields

dietcoke-17:
Fixes #2691 and #3712 follow up

## Bug

`collect_agent_data` in both `mpl_space_drawing.py` and `MatplotlibBackend` only added `edgecolors` when an agent provided it.

If only some agents had `edgecolors`, the array length became smaller than other agent data arrays (`loc`, `marker`, `zorder`, etc.), causing an `IndexError` during boolean indexing.

## Fix

- Always add one `edgecolors` entry per agent.
- Use `None` when an agent does not specify a value, keeping all arrays aligned.

Also added a shared `_to_numpy_argument_array()` helper in `mpl_space_drawing.py` and reused it in `MatplotlibBackend` to avoid duplicate array handling logic.

The helper correctly handles:
- `None` values
- Mixed color formats
- Tuple markers like `(numsides, style, angle)`


Dilraj07:
Nice work consolidating the array handling into `_to_numpy_argument_array`! One quick thought: should the object-array fallback also apply to `c` (face colors), in case a portrayal mixes string colors (e.g. `"blue"`) and RGBA tuples?

dietcoke-17:
Good catch, and I checked — it's real (np.asarray(["blue", (1.0,0.0,0.0,1.0)]) raises the same ValueError). It's a separate pre-existing bug though, not something this PR introduces — happens on current main too.Happy to pick it up right after this merges since the fix is basically the same pattern.

quaquel:
How does this relate to #3712

dietcoke-17:
> How does this relate to #3712 

#3856 fixes the same root cause as #3712 (#2691) — collect_agent_data only appended edgecolors when an agent's portrayal set it, so that array could end up shorter than every other per-agent array (loc, marker, zorder), causing the boolean-index IndexError.
It's a fresh implementation rather than a continuation of that branch, but it builds on:
the diagnosis in #3712 — same bug, same root cause
@tpike3's review comment there asking to consolidate the duplicated array-conversion logic — this PR does that via a shared _to_numpy_argument_array() helper in mpl_space_drawing.py, imported by MatplotlibBackend instead of reimplementing
#3712 stalled after that review request with no follow-up.

quaquel (approved):

quaquel:
there is a conflict that must be resolved before this can be merged. Code wise it looks fine to me.

dietcoke-17:
Is it regarding the object array fallback asked by dilraj or any other conflicts?

quaquel:
it seems import related so trivial to fix. I can probably handle it if needed but would be good if you can take a quick lok.

dietcoke-17:
Found it and fixed it too — test_components_matplotlib.py had two PRs adding different names to the same mpl_space_drawing import list (_get_hexmesh from another PR, _to_numpy_argument_array from mine).

PR #3865, opened 2026-09-17
add agent_by_id

tpike3:
### Summary
Adds an `agents_by_id` property to `model` that provides O(1) lookup of any agent by its `unique_id`. The mapping is maintained automatically through `register_agent` / `deregister_agent`.

### Motive
Currently there is `model.agents` and `model.agents_by_type` , but there is no built-in way to resolve a single agent from its `unique_id` without a linear scan. This gap becomes a bottleneck for:

### Implementation

-  New internal store: `self._agents_by_id: dict[int, A] = {}` initialized alongside` _agents_by_type` and `_all_agents` in `model`
- Registration hook: `register_agent` inserts `agent.unique_id → agent` into `_agents_by_id` immediately after the ID is assigned.
- Deregistration hook: `deregister_agent` deletes the entry from `_agents_by_id` before removing the agent from `_all_agents`.
- `Model.agents_by_id` exposes the dict as a read-only-by-convention property, consistent with how agents_by_type is exposed. 

### Usage Examples
```
# Direct O(1) lookup
agent = model.agents_by_id[42]
# Safe lookup with default
agent = model.agents_by_id.get(some_id)
# Resolve a batch of stored IDs to live agent references
team = [model.agents_by_id[uid] for uid in stored_member_ids]
```

### Additional Notes
Dict insert/delete in register_agent / deregister_agent is O(1); no measurable impact on model setup or teardown.


quaquel:
I have always seen the agent_id as an internal thing and have been reluctant to add this style of lookup in the past. What practical problem related to meta agents is being solved by this?

API wise, I am fine. I'll try to look at the implementation after I understand the use case.

tpike3:
@quaquel Thanks! Honestly, I just wanted to get your thoughts in context. Going through the use cases I did not see a compelling case to keep ids as the main reference look up and the code is much cleaner without it. 

I am going to close this and the other PR and do a new PR with the update.

PR #3866, opened 2026-09-18
mixed-face-colors patch

dietcoke-17:
### Pre-PR Checklist
### Summary
Extends the object-array fallback added in #3856 to face colors (`c`), so mixing plain color-name strings and RGBA tuples across agents doesn't crash `collect_agent_data`.
 
### Bug
Raised by @Dilraj07 during review of #3856: `_to_numpy_argument_array` handles mixed color representations for `edgecolors`, but `c` (face color) can hit the exact same failure. `np.asarray(["blue", (1.0, 0.0, 0.0, 1.0)])` raises `ValueError: setting an array element with a sequence...` for the same reason the original `edgecolors` bug did — NumPy can't coerce a plain string next to a 4-tuple into one array.
 
This is a pre-existing bug, not introduced by #3856 — the same call would have failed on `main` before that PR too. Filed as a separate follow-up per discussion on #3856 to keep that PR scoped to #2691.
 
### Fix
- Factored the existing try/`np.asarray`/except `ValueError`/manual-object-array logic out of the `edgecolors` branch into a small `_coerce_or_object_array()` helper.
- Reused it for `c`, so mixed face colors fall back to a 1D object array the same way mixed edgecolors do.
- Uniform-length RGBA/RGB tuples are untouched — `np.asarray` already turns those into a valid `(N, 4)`/`(N, 3)` array, which is exactly what Matplotlib expects, so only genuinely mixed representations take the fallback path.
- `MatplotlibBackend` needs no changes: it already imports `_to_numpy_argument_array` from `mpl_space_drawing.py`, so it picks up the fix automatically.
### Testing
- New: `test_to_numpy_argument_array_mixed_face_colors` — unit test confirming a mixed `["blue", (1.0, 0.0, 0.0, 1.0)]` list round-trips correctly, and a uniform RGBA-tuple list still produces a proper `(N, 4)` array rather than being forced into an object array.
- New: `test_draw_network_with_mixed_face_colors` — end-to-end regression test through `draw_network` with agents split between a plain color name and an RGBA tuple.
- Also manually confirmed `MatplotlibBackend.collect_agent_data` handles the same mixed-color case correctly with zero changes to that file.

### Additional Notes
Scoped to matplotlib visualization style collection only — no public API, dependency, schema, or CI changes. Builds directly on the `_to_numpy_argument_array` helper introduced in #3856; please merge that first if it hasn't landed yet, since this branches from it.


tpike3 (approved):
LGTM thanks @dietcoke-17

PR #3868, opened 2026-09-18
Make select() raise on negative integer or boolean at_most (#3867)

### LINKED ISSUE #3867: Make AbstractAgentSet.select() and CellCollection.select() raise on negative integer or boolean at_most
### Describe the bug
Following PR #3853 and PR #3860 which added float fraction validation in `select()`, negative integers and boolean inputs are still not validated in `AbstractAgentSet.select()` and `CellCollection.select()`.

### To Reproduce
```python
# 1. Negative integer silently returns empty set instead of raising ValueError
model.agents.select(at_most=-1)  # returns 0 agents

# 2. Booleans are silently treated as 0 or 1 because isinstance(bool, int) is True
model.agents.select(at_most=False)  # returns 0 agents
model.agents.select(at_most=True)   # returns 1 agent

# 3. Non-numeric types crash late inside generator rather than failing eagerly
list(model.agents.select(at_most="5"))  # TypeError: '>=' not supported between instances of 'int' and 'str'
```

*(The same behavior reproduces on `CellCollection.select()`)*.

### Expected Behavior
- Passing an integer `< 0` should raise `ValueError("Integer at_most must be non-negative, got ...")`.
- Passing a boolean or non-number should raise `TypeError("at_most must be an integer or float, got ...")`, matching `AbstractAgentSet.select_random()`.

I have a fix with unit tests ready for both modules and can open a PR.

Dilraj07:
Fixes #3867

### Summary
Closes validation gaps in `AbstractAgentSet.select()` and `CellCollection.select()` by eagerly raising `TypeError` for booleans / non-numeric types, and `ValueError` for negative integers.

### Problem
1. **Negative integers (`at_most < 0`):** The termination check `count >= at_most` (`0 >= -1`) evaluated to `True` on the first item, silently returning an empty set instead of raising `ValueError`.
2. **Booleans (`at_most=True/False`):** Because `bool` subclasses `int` in Python (`isinstance(False, int) == True`), booleans were silently treated as `0` or `1`.
3. **Non-numeric types:** Caused late crashes inside generator iteration instead of failing eagerly at call time.

### Fix
Applied symmetrically to `AbstractAgentSet.select()` and `CellCollection.select()`:
- **Eager Type Check:** `if isinstance(at_most, bool) or not isinstance(at_most, (int, float)): raise TypeError(...)` (intercepts booleans before integer evaluation).
- **Negative Value Check:** `if isinstance(at_most, int) and at_most < 0: raise ValueError(...)` (`at_most=0` remains valid).
- **Docstrings:** Documented `TypeError` and updated `ValueError` conditions.

### Testing
- Added tests in `tests/test_agentset.py` and `tests/discrete_space/test_discrete_space.py` covering negative numbers, booleans, and non-numeric types.
- All 134 tests passed locally (`pytest tests/test_agentset.py tests/discrete_space/test_discrete_space.py`).
- Formatted and linted with `ruff check .` and `ruff format .`.

### Checklist
- [x] Fixes linked issue (#3867)
- [x] Self-review performed
- [x] Tests added and passing locally
- [x] Code formatted and linted (`ruff`)


Dilraj07:
Note on CI Failure: The failing build / examples check is an upstream issue in mesa-examples where VirusAntibodyModel triggers the sequence length validation added in #3824 (Agent.create_agents). This is affecting all open PRs in Mesa and is tracked / resolved in [mesa-examples#496](https://github.com/mesa/mesa-examples/pull/496). All core Mesa test matrices and coverage checks passed.

mesa/discrete_space/cell_collection.py
```diff
@@ -157,13 +157,24 @@ def select(
...
+            raise TypeError(
+                f"at_most must be an integer or float, got {type(at_most).__name__}."
+            )
+
+        if isinstance(at_most, int):
+            if at_most < 0:
+                raise ValueError(
+                    f"Integer at_most must be non-negative, got {at_most}."
+                )
+        elif at_most != float("inf"):
```
quaquel: We are adding lost of tests here. Can't we simplify this a lot, for example by first checking if its a number and converting it to a float before moving on?

Dilraj07: > We are adding lost of tests here. Can't we simplify this a lot, for example by first checking if its a number and converting it to a float before moving on?

Thanks for the suggestion. We can't convert to `float` early because the type itself carries meaning, documented in the existing docstring:

- `at_most=1` **(int)** selects **1 agent** (absolute count)
- `at_most=1.0` **(float)** selects **100% of agents** (fraction x total)

Converting `int` to `float` early would make `at_most=1` behave like `at_most=1.0`, silently returning all agents instead of one, which would be a regression.

That said, I've restructured the branches to be flatter (float path first, int in `elif`) which reduces one level of nesting. Let me know if you had a different shape in mind.

PR #3870, opened 2026-09-19
Handle all numpy int and float widths in NumpyJSONEncoder

### LINKED ISSUE #3869: NumpyJSONEncoder fails on numpy float32 and float16 values
**Describe the bug**

`NumpyJSONEncoder` in `mesa/experimental/data_collection/datarecorders.py` only converts `np.float64` among the float types, so serializing a `np.float32` or `np.float16` value raises `TypeError`. It lists every integer width explicitly but handles just one float width, so narrower floats fall through to the base encoder and fail.

This shows up when a `JSONDataRecorder` saves a `ModelDataSet` or `AgentDataSet` whose field holds a numpy float32/float16 value (for example an attribute derived from a float32 array). The dict and list snapshots keep the raw numpy scalars, and `save_to_json` then runs them through the encoder.

**To Reproduce**

```python
import json
import numpy as np
from mesa.experimental.data_collection.datarecorders import NumpyJSONEncoder

json.dumps({"x": np.float32(1.5)}, cls=NumpyJSONEncoder)
# TypeError: Object of type float32 is not JSON serializable
# np.float16 fails the same way; np.float64 works
```

So whether serialization succeeds depends on the width of the value, which is surprising.

**Expected behavior**

The encoder should convert every numpy float width (and every int width) to a native Python number, the way it already intends to for ints.

**Additional context**

Reproduced on current main. A small fix is to check against `np.integer` and `np.floating` rather than enumerating specific dtypes.

AshNicolus:
Closes #3869.

`NumpyJSONEncoder` listed every integer width but only `np.float64` among the floats, so serializing a `np.float32` or `np.float16` value raised `TypeError`. In practice that happens when a `JSONDataRecorder` saves a `ModelDataSet` or `AgentDataSet` whose field holds a numpy float32/float16 scalar, since `save_to_json` runs those dict and list snapshots through the encoder.

Repro on main:

```python
import json, numpy as np
from mesa.experimental.data_collection.datarecorders import NumpyJSONEncoder
json.dumps({"x": np.float32(1.5)}, cls=NumpyJSONEncoder)  # TypeError
```

Fix: check against `np.integer` and `np.floating` instead of enumerating specific dtypes, so all int and float widths are covered (int8 through int64, the unsigned ints, and float16/float32/float64). Behavior for float64, ints, bool, and ndarray is unchanged.

Also extended `test_json_recorder_numpy_encoder_types` to cover float16, float32, and a couple more int widths, which the previous test missed.

Verified locally: the data-collection recorder tests pass and ruff check and format are clean.


tpike3 (approved):
LGTM! Thanks @AshNicolus

PR #3874, opened 2026-09-19
Honor stop and pause called from an EventGenerator callback

### LINKED ISSUE #3873: EventGenerator ignores stop() or pause() called from its own callback
**Describe the bug**

`EventGenerator._execute_and_reschedule` runs the user callback and then schedules the next event, but it does not re-check whether the callback stopped or paused the generator. So calling `gen.stop()` or `gen.pause()` from inside the generator's own callback is only partly honored: a next event still gets scheduled, and the generator is left with a stale `_current_event`, so `next_scheduled_time` reports a time even though the generator is stopped.

**To Reproduce**

```python
from mesa import Model
from mesa.time import Schedule

model = Model()
calls = []
gen = None
def cb():
    calls.append(model.time)
    if len(calls) == 1:
        gen.stop()

gen = model.schedule_recurring(cb, Schedule(interval=1.0, start=1.0))
model.run_for(5)

print(gen.is_active)            # False
print(gen.next_scheduled_time)  # 2.0, expected None
```

`pause()` from the callback has the same problem: it reschedules a next event instead of staying paused until `resume()`.

**Expected behavior**

If the callback stops or pauses the generator, no next event should be scheduled, and `next_scheduled_time` should be `None` for a stopped generator.

**Additional context**

Reproduced on current main. The reschedule happens because `_execute_and_reschedule` does not re-check `_active`/`_paused` after running the callback.

AshNicolus:
Closes #3873.

`EventGenerator._execute_and_reschedule` ran the callback and then scheduled the next event without re-checking whether the callback had stopped or paused the generator. So `gen.stop()` or `gen.pause()` called from inside the generator's own callback was only partly honored: a next event was still scheduled, and `next_scheduled_time` kept reporting a time on a stopped generator.

Repro on main:

```python
def cb():
    calls.append(model.time)
    gen.stop()
gen = model.schedule_recurring(cb, Schedule(interval=1.0, start=1.0))
model.run_for(5)
gen.next_scheduled_time   # 2.0, expected None
```

Fix: after running the callback, return early if the generator is no longer active or is paused, before scheduling the next event.

Added tests for stopping and pausing from within the callback.

Verified locally: the event tests pass and ruff check and format are clean.


quaquel (approved):

EwoutH:
Thanks for catching this!

PR #3876, opened 2026-09-20
Preserve nested mutable DataCollector values

SahilKumar75:
> [!NOTE]
> Use this template for bug fixes only. For enhancements or new features, use the feature template and get maintainer approval in an issue or discussion before opening a PR.

### Pre PR Checklist

### Summary

Preserve historical DataCollector values when an agent reporter or agent type reporter returns a tuple containing mutable objects.

### Bug / Issue

DataCollector skips copying values it considers immutable. The shortcut currently includes every tuple, even though a tuple can contain mutable objects such as lists.

After collection, mutating a nested list therefore changes the value stored in an earlier historical record. The expected behavior is for collected records to remain unchanged.

No separate issue was opened because this is a clear, scoped bug fix and the contribution guide permits a direct pull request for this case.

### Implementation

Remove tuple from the immutable value shortcut in both agent reporter collection paths. This allows `deepcopy` to inspect tuple contents and copy nested mutable objects.

Add regression coverage for both regular agent reporters and agent type reporters.

### Testing

`python -m pytest tests/test_datacollector.py -q`

Result: 49 passed.

`python -m pytest -q`

Result: 838 passed.

`pre-commit run --files mesa/datacollection.py tests/test_datacollector.py`

Result: all hooks passed.

### Additional Notes

This does not change the reporter API. It only ensures that nested mutable values receive the same historical isolation already provided for other mutable reporter results.


mesa/datacollection.py
```diff
@@ -340,7 +340,7 @@ def _record_agenttype(self, model, agent_type):
         """Record agent-type data in a mapping of functions and agents."""
         rep_funcs = self.agenttype_reporters[agent_type].values()
         # Immutable types that don't need deepcopy
-        python_immutable_types = (str, int, bool, float, tuple)
+        python_immutable_types = (str, int, bool, float)
```
dietcoke-17: A short comment at the tuple removal (why an "immutable" type still needs copying) would save the next reader some head-scratching

mesa/datacollection.py
```diff
@@ -319,7 +319,7 @@ def _record_agents(self, model):
         """Record agents data in a mapping of functions and agents."""
         rep_funcs = self.agent_reporters.values()
         # Immutable types that don't need deepcopy
-        python_immutable_types = (str, int, bool, float, tuple)
+        python_immutable_types = (str, int, bool, float)
```
dietcoke-17: python_immutable_types is now duplicated identically in two functions — this bug is a live example of that duplication causing drift. Worth hoisting to a shared constant.

dietcoke-17:
Not related to this fix, but noticed while reading: add_table_row() appends row values with no copy, so a mutable value written into a table keeps a live reference after being recorded — same category of issue this PR fixes for agent/agenttype reporters, just a different path (tables vs the reporter dicts).

mesa/datacollection.py
SahilKumar75: Done in 1a25bd7. Both copies are gone; there is now one module level `_PYTHON_IMMUTABLE_TYPES` that both record paths use, so the two cannot drift again.

mesa/datacollection.py
SahilKumar75: Added, on the constant itself rather than at the call sites: it notes that a tuple is only shallowly immutable, so a tuple holding a list still needs a deepcopy for collected records to stay frozen.

SahilKumar75:
Pushed 1a25bd7 for both review comments: the duplicated immutable type tuple is now one module level constant, and it carries a comment explaining why tuple is not in it.

Re `add_table_row()`: agreed, that is the same class of bug on a different path, and it is real. Keeping it out of this PR so the fix stays scoped to the reporter paths, since a table row fix changes copy behavior for every existing table writer and deserves its own discussion. Happy to open a separate issue for it.

Local results after the change: `tests/test_datacollector.py` 49 passed, full suite 838 passed, ruff check and format clean.

The `examples` job failure is not from this branch. It fails the same way on `main` (for example run 35594755774), `VirusAntibodyModel` raising `ValueError: sequence of length 3 does not match the number of agents (20)` from `agentset.py`, which no code here touches.

dietcoke-17:
> Pushed 1a25bd7 for both review comments: the duplicated immutable type tuple is now one module level constant, and it carries a comment explaining why tuple is not in it.
> 
> Re `add_table_row()`: agreed, that is the same class of bug on a different path, and it is real. Keeping it out of this PR so the fix stays scoped to the reporter paths, since a table row fix changes copy behavior for every existing table writer and deserves its own discussion. Happy to open a separate issue for it.
> 
> Local results after the change: `tests/test_datacollector.py` 49 passed, full suite 838 passed, ruff check and format clean.
> 
> The `examples` job failure is not from this branch. It fails the same way on `main` (for example run 35594755774), `VirusAntibodyModel` raising `ValueError: sequence of length 3 does not match the number of agents (20)` from `agentset.py`, which no code here touches.

I will look at the table row() and dont worry about that I just pointed out its worth fixing well I have created the issue and will look into it

